PR #1160, opened 2021-01-23
Refactor regression test to expose / remove unused Datadog.Trace.AspNet code

zacharycmontoya:
The Log4Net.SerializationException regression test does the bare minimum to simulate the IIS PreStartInit phase to make sure that initializing a Tracer with `LogsInjectionEnabled=true` doesn't crash with a `SerializationException`. Refactor to remove a code path that is no longer used in the `Datadog.Trace.AspNet`.

@DataDog/apm-dotnet


src/Datadog.Trace.AspNet/HttpApplicationStartup.cs
```diff
@@ -1,35 +0,0 @@
-using System.Web;
-using Datadog.Trace.AspNet;
-
-[assembly: PreApplicationStartMethod(typeof(HttpApplicationStartup), "Register")]
```
kevingosse, line 4: I thought this was needed to load the TracingHttpModule, but looks like it works without it

kevingosse (approved):

src/Datadog.Trace.AspNet/HttpApplicationStartup.cs
lucaspimentel: It's deprecated code from when we used to distribute it as a nuget package.

lucaspimentel (approved):

src/Datadog.Trace.AspNet/HttpApplicationStartup.cs
lucaspimentel: We now add it to `applicationHost.config` during installation. Remember the bug when we left out the `v` in `runtimeVersionv4.0`? 😅 

https://github.com/DataDog/dd-trace-dotnet/blob/4d7cbd6710bf6a66bbae4c7005b153cfd2f97f90/src/WindowsInstaller/install.cmd#L6

PR #1162, opened 2021-01-25
Add support for synchronous HttpClient.Send()

andrewlock:
Adds support for synchronous `Send()` method on `HttpClientHandler`, introduced in .NET 5.0. 

Essentially identical to `SendAsync()` instrumentation

@DataDog/apm-dotnet

### DIFF
--- a/integrations.json
+++ b/integrations.json
@@ -73,6 +73,30 @@
           "type": "Datadog.Trace.ClrProfiler.AutoInstrumentation.HttpClientHandler.HttpClientHandlerIntegration",
           "action": "CallTargetModification"
         }
+      },
+      {
+        "caller": {},
+        "target": {
+          "assembly": "System.Net.Http",
+          "type": "System.Net.Http.HttpClientHandler",
+          "method": "Send",
+          "signature_types": [
+            "System.Net.Http.HttpResponseMessage",
+            "System.Net.Http.HttpRequestMessage",
+            "System.Threading.CancellationToken"
+          ],
+          "minimum_major": 5,
+          "minimum_minor": 0,
+          "minimum_patch": 0,
+          "maximum_major": 5,
+          "maximum_minor": 65535,
+          "maximum_patch": 65535
+        },
+        "wrapper": {
+          "assembly": "Datadog.Trace.ClrProfiler.Managed, Version=1.22.0.0, Culture=neutral, PublicKeyToken=def86d061d0d2eeb",
+          "type": "Datadog.Trace.ClrProfiler.AutoInstrumentation.HttpClientHandler.HttpClientHandlerSyncIntegration",
+          "action": "CallTargetModification"
+        }
       }
     ]
   },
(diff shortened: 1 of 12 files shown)


src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/HttpClientHandler/HttpClientHandlerIntegration.cs
```diff
@@ -37,21 +33,7 @@ public class HttpClientHandlerIntegration
...
-
-                    // add distributed tracing headers to the HTTP request
-                    SpanContextPropagator.Instance.Inject(scope.Span.Context, new HttpHeadersCollection(requestMessage.Headers));
-
-                    return new CallTargetState(scope);
-                }
-            }
-
-            return CallTargetState.GetDefault();
+            return HttpClientHandlerCommon.OnMethodBegin(instance, requestMessage, cancellationToken);
```
tonyredondo, line 36: I think is great to have this common methods!

I was thinking also in move `HttpClientHandler`, `SocketsHttpHandler` and `WinHttpHandler` to a new `Http` parent folder (`AutoInstrumentation\Http\[name]`). Something similar to what I'm doing with the `AdoNet` integrations. (I also want to create a new `Testing` folder for `NUnit` and `XUnit` integrations.

WDYT?

andrewlock: Yes, I agree, I think that would make a lot of sense 👍

My concern is whether this would cause breaking issues when there's a mismatch between manual and automatic instrumentation? I don't have a good sense of when those happen yet (as changing an `internal` class caused issues in my testing!)

tonyredondo: Well all instrumentations within `AutoInstrumentation` folder is CallTarget and that is not GA yet. So we are in a good place to make all changes we want. For CallSite integrations you have to look into `Integrations` folder. I don't think this PR has a CallSite integration isn't ? I would expect it in `Integrations\HttpMessageHandlerIntegration.cs`

andrewlock: Ah, oops, didn't notice both were required. In which case yes, that makes sense to update the call targets layout I think. 

I assume I should also add the CallSite integration? Or would we rather only introduce this for CallTarget, given that's what everything's moving to? I'm happy either way, but it probably makes sense to add it to CallSite too I'd guess?

andrewlock: Urgh, the integration tests are genuinely failing in CI, but passing when I run locally, and who knows why. Great 😩

tonyredondo (approved):
LGTM!

zacharycmontoya (approved):
LGTM!

PR #1150, opened 2021-01-20
Standardise log levels

andrewlock:
Initial cleanup of log levels before implementing remainder of [the logging RFC](https://github.com/DataDog/architecture/blob/master/rfcs/apm/integrations/tracer-logging/rfc.md#rate-limits)

- Use only the allowed levels: `Debug`, `Warning`, `Error`
- Update default/minimum log level accordingly
- Update some log levels based on [suggestions in RFC](https://github.com/DataDog/architecture/blob/master/rfcs/apm/integrations/tracer-logging/rfc.md#log-levels).
- Replace `string.Format()` style log templates (`"{0},{1}"`) with named arguments (`"{Name},{Value}"`). Doesn't make much difference atm, as we're only writing to file, but is better practice, and will be better if we do structured logging down the line instead.

In a follow up PR I'll encapsulate the `ILogger` to restrict to using only the allowed levels, and add rate limiting, as defined in the RFC

@DataDog/apm-dotnet

PR #1163, opened 2021-01-25
Replace log messages using string.Format style with structured log style

andrewlock:
Initial cleanup of log levels before implementing remainder of the logging RFC

Replaces string.Format() style log templates (`"{0},{1}"`) with named arguments (`"{Name},{Value}"`). Doesn't make much difference atm, as we're only writing to file, but is better practice, and will be better if we do structured logging down the line instead.

Extracted from https://github.com/DataDog/dd-trace-dotnet/pull/1150 as there concerns around following the RFC in that PR, and this change is independent

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace.Ci.Shared/CIEnvironmentValues.cs
+++ b/src/Datadog.Trace.Ci.Shared/CIEnvironmentValues.cs
@@ -117,7 +117,7 @@ namespace Datadog.Trace.Ci
             }
             catch (Exception ex)
             {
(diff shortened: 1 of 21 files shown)


zacharycmontoya (approved):

lucaspimentel (approved):

andrewlock:
Integration tests failing due to out of drive space, and all safe changes, so merging

PR #1167, opened 2021-01-26
Try detecting multiple initializations

kevingosse:


kevingosse:
Doesn't seem to crash :(

PR #1170, opened 2021-01-27
Fix WinHttpHandlerIntegration

andrewlock:
This PR fixes the CallTarget `WinHttpHandlerIntegration` and adds extra integration tests

The current `HttpMessageHandler` previously only tested two scenarios:

* `HttpClient` tracing with the default configuration (using `new HttpClient()`)
* A custom `HttpClientHandler`

However, the `HttpClient` stack can be quite complicated. For example, the following scenarios were not previously tested:

* `HttpClient` with [the `DOTNET_SYSTEM_NET_HTTP_USESOCKETSHTTPHANDLER`  switch](https://docs.microsoft.com/dotnet/api/system.net.http.socketshttphandler) disabled and calling `new HttpClient()`
* `SocketsHttpHandler` without the `HttpClientHandler`, i.e. calling `new HttpClient(new SocketsHttpHandler())`
* `WinHttpHandler` without the `HttpClientHandler`, i.e. calling `new HttpClient(new WinHttpHandler())`
* `CurlHandler` (not instrumented or tested)

This PR expands the testing matrix to cover these additional scenarios.

We currently have CallSite instrumentation for `SocketsHttpHandler` and `HttpClientHandler`, and CallTarget instrumentation for both these and the `WinHttpHandler`. 

> It is worth noting that both  `SocketsHttpHandler` and `WinHttpHandler` CallTarget instrumentation [are currently disabled by default](https://github.com/DataDog/dd-trace-dotnet/blob/a32ebc2604aef5be79c7c4840b7d3bc962146063/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/WinHttpHandler/WinHttpHandlerIntegration.cs#L97), so if a user uses `new HttpClient(new WinHttpHandler())` they will **not** be instrumented by default. ~~Is that intentional?~~ As mentioned by @tonyredondo , yes it is, but we should potentially consider enabling them by default for CallTarget

As part of this testing, discovered that the `WinHttpHandler` CallTarget tracing was not working, as it's referencing the wrong assembly. This PR fixes that. 

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/WinHttpHandler/WinHttpHandlerIntegration.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/WinHttpHandler/WinHttpHandlerIntegration.cs
@@ -39,15 +39,11 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.WinHttpHandler
         public static CallTargetState OnMethodBegin<TTarget, TRequest>(TTarget instance, TRequest requestMessage, CancellationToken cancellationToken)
             where TRequest : IHttpRequestMessage
         {
-            Log.Warning("WinHttpHandler started");
             if (IsTracingEnabled(requestMessage.Headers))
             {
-                Log.Warning("WinHttpHandler enabled");
                 Scope scope = ScopeFactory.CreateOutboundHttpScope(Tracer.Instance, requestMessage.Method.Method, requestMessage.RequestUri, IntegrationId, out HttpTags tags);
                 if (scope != null)
                 {
-                    Log.Warning("WinHttpHandler scope created");
-
                     tags.HttpClientHandlerType = instance.GetType().FullName;
 
                     // add distributed tracing headers to the HTTP request
--- a/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/WinHttpHandler/WinHttpHandlerIntegration.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/WinHttpHandler/WinHttpHandlerIntegration.cs
@@ -25,7 +25,6 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.WinHttpHandler
         private const string IntegrationName = nameof(IntegrationIds.HttpMessageHandler);
         private static readonly IntegrationInfo IntegrationId = IntegrationRegistry.GetIntegrationInfo(IntegrationName);
         private static readonly IntegrationInfo WinHttpHandlerIntegrationId = IntegrationRegistry.GetIntegrationInfo(nameof(IntegrationIds.WinHttpHandler));
-        private static readonly Vendors.Serilog.ILogger Log = Datadog.Trace.Logging.DatadogLogging.GetLogger(typeof(WinHttpHandlerIntegration));
 
         /// <summary>
         /// OnMethodBegin callback
--- a/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/WinHttpHandler/WinHttpHandlerIntegration.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/WinHttpHandler/WinHttpHandlerIntegration.cs
@@ -12,7 +12,7 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.WinHttpHandler
     /// System.Net.Http.WinHttpHandler calltarget instrumentation
     /// </summary>
     [InstrumentMethod(
-        AssemblyName = "System.Net.Http",
+        AssemblyName = "System.Net.Http.WinHttpHandler",
         TypeName = "System.Net.Http.WinHttpHandler",
         MethodName = "SendAsync",
         ReturnTypeName = ClrNames.HttpResponseMessageTask,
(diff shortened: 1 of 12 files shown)


src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/WinHttpHandler/WinHttpHandlerIntegration.cs
```diff
@@ -38,11 +39,15 @@ public class WinHttpHandlerIntegration
         public static CallTargetState OnMethodBegin<TTarget, TRequest>(TTarget instance, TRequest requestMessage, CancellationToken cancellationToken)
             where TRequest : IHttpRequestMessage
         {
+            Log.Warning("WinHttpHandler started");
```
tonyredondo, line 42: Seems like these logs are for debugging.

andrewlock: oops! you caught me

tonyredondo:
>It is worth noting that the WinHttpHandler CallTarget instrumentation is currently disabled by default, so if a user uses new HttpClient(new WinHttpHandler()) they will not be instrumented by default. Is that intentional?

Yes the same with the [SocketsHttpHandlerIntegration](https://github.com/DataDog/dd-trace-dotnet/blob/a32ebc2604aef5be79c7c4840b7d3bc962146063/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/SocketsHttpHandler/SocketsHttpHandlerIntegration.cs#L97), this inherits from the [CallSite integration](https://github.com/DataDog/dd-trace-dotnet/blob/a32ebc2604aef5be79c7c4840b7d3bc962146063/src/Datadog.Trace.ClrProfiler.Managed/Integrations/HttpMessageHandlerIntegration.cs#L290). I didn't want to change this behavior until we discuss it with the team. But personally I vote for enabling it by default.

tonyredondo:
> As part of this testing, discovered that the WinHttpHandler CallTarget tracing was not working, as it's referencing the wrong assembly. This PR fixes that.

Thanks for that catch!

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/WinHttpHandler/WinHttpHandlerIntegration.cs
andrewlock: fixed in https://github.com/DataDog/dd-trace-dotnet/pull/1170/commits/e5d6c21afb0e573e74d328dc65ad6c51dd34094b

andrewlock:
> Yes the same with the [SocketsHttpHandlerIntegration](https://github.com/DataDog/dd-trace-dotnet/blob/a32ebc2604aef5be79c7c4840b7d3bc962146063/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/SocketsHttpHandler/SocketsHttpHandlerIntegration.cs#L97)

Oops, for some reason, I thought the `SocketsHttpHandlerIntegration` was _on_ by default, my bad

> ...personally I vote for enabling it by default.

I agree 👍

tonyredondo:
> CurlHandler (not instrumented or tested)

Yep that one is missing, also the HttpWebRequest for older framework versions.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/WinHttpHandler/WinHttpHandlerIntegration.cs
```diff
@@ -25,6 +25,7 @@ public class WinHttpHandlerIntegration
         private const string IntegrationName = nameof(IntegrationIds.HttpMessageHandler);
         private static readonly IntegrationInfo IntegrationId = IntegrationRegistry.GetIntegrationInfo(IntegrationName);
         private static readonly IntegrationInfo WinHttpHandlerIntegrationId = IntegrationRegistry.GetIntegrationInfo(nameof(IntegrationIds.WinHttpHandler));
+        private static readonly Vendors.Serilog.ILogger Log = Datadog.Trace.Logging.DatadogLogging.GetLogger(typeof(WinHttpHandlerIntegration));
```
tonyredondo, line 28: nit: I think we can remove this instance if we are not going to use it.

andrewlock:
> also the HttpWebRequest for older framework versions

We have CallSite integration for `WebRequest` currently though right? just not CallTarget?

tonyredondo (approved):
LGTM!

tonyredondo:
> > also the HttpWebRequest for older framework versions
> 
> We have CallSite integration for `WebRequest` currently though right? just not CallTarget?

Yes, that migration is pending.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/WinHttpHandler/WinHttpHandlerIntegration.cs
andrewlock: Fixed in https://github.com/DataDog/dd-trace-dotnet/pull/1170/commits/5a3a3aa6ec6cc39e36ecd61366207fda020f57be

Datadog.Trace.sln
```diff
@@ -406,6 +406,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "LargePayload", "test\test-a
 EndProject
 Project("{778DAE3C-4631-46EA-AA77-85C1314464D9}") = "Samples.SqlServer.Vb", "test\test-applications\integrations\dependency-libs\Samples.SqlServer.Vb\Samples.SqlServer.Vb.vbproj", "{CDE0C4D2-4B75-4ED0-988D-08E08B23B895}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Samples.WebRequest", "test\test-applications\integrations\Samples.WebRequest\Samples.WebRequest.csproj", "{15BA9E32-CE6A-4D36-B81E-B8CD713E42FD}"
```
zacharycmontoya, line 409: Good catch! I can't believe that got missed in a recent migration 😅

zacharycmontoya (approved):
Looks good!

PR #1173, opened 2021-01-28
Refactor ADO.NET integrations to use the assembly integration attribute

tonyredondo:
This PR is a refactor for integrations that follows the standard ADO.NET base classes/interfaces. Allowing to create new ADO.NET integrations faster.

**Changes made in this PR:**

- [x] Changes the `InstrumentMethodAttribute` to enable declaration support at Assembly level.
- [x] Changes the PrepareRelease tool to add support to the new version of the `InstrumentMethodAttribute`. 
- [x] Adds the ADO.NET CallTarget integration in a generic way to be used for all specific ADO.NET target libraries.
- [x] Adds multiple `InstrumentMethodAttribute` sub types for each ADO.NET method that need to be instrumented.
- [x] Refactor the SqlClient ADO.NET integration to use these new features.

**Advantages:**

To create a new ADO.NET integration using this mechanism we should create two files:

- **[VendorName]Constants.cs**: Here we should define a new type that inherits from `IAdoNetClientData` with the required specific data from the client library. [Example](https://github.com/DataDog/dd-trace-dotnet/blob/tony/calltarget-assembly-attribute/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AdoNet/SqlClient/SqlClientConstants.cs)
- **[VendorName]Definitions.cs**: Here we should define all `[assembly: ....]` attributes to define all methods that we want to be intercepted in the target library. [Example](https://github.com/DataDog/dd-trace-dotnet/blob/tony/calltarget-assembly-attribute/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AdoNet/SqlClient/SqlClientDefinitions.cs)

@DataDog/apm-dotnet


src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AdoNet/AdoNetClientInstrumentMethodAttribute.cs
```diff
@@ -0,0 +1,125 @@
...
+        protected IAdoNetClientData AdoNetClientData { get; }
+
+        internal class CommandExecuteNonQueryAsyncAttribute : AdoNetClientInstrumentMethodAttribute
+        {
+            public CommandExecuteNonQueryAsyncAttribute(Type adoNetClientDataType)
+                : base(adoNetClientDataType)
+            {
+                Method = AdoNetConstants.MethodNames.ExecuteNonQueryAsync;
+                ReturnTypeName = AdoNetConstants.TypeNames.IntTaskType;
+                ParametersTypesNames = new[] { ClrNames.CancellationToken };
```
lucaspimentel, line 27: Not directly related to this PR, but this property should be called `ParameterTypeNames` (only `Names` needs to be plural).

lucaspimentel: Also `MethodName` and `CallTargetTypeName`, for consistency.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AdoNet/AdoNetConstants.cs
```diff
@@ -13,6 +14,9 @@ public static class TypeNames
 
             public const string DbDataReaderType = "System.Data.Common.DbDataReader";
             public const string DbDataReaderTaskType = "System.Threading.Tasks.Task`1<System.Data.Common.DbDataReader>";
+
+            public const string IntTaskType = "System.Threading.Tasks.Task`1<System.Int32>";
```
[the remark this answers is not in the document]
tonyredondo: done in a93295c

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AdoNet/AdoNetClientInstrumentMethodAttribute.cs
tonyredondo: done in a93295c. CallTargetClass has been changed to CallTargetClassType.

lucaspimentel: Oh yeah, that one is a `Type`, not a `string`. Thanks 😉 
Maybe `CallTargetType` then?

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AdoNet/AdoNetConstants.cs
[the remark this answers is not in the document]
lucaspimentel: thanks

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AdoNet/AdoNetClientInstrumentMethodAttribute.cs
tonyredondo: 😂 ok, I change it, I don't have a strong opinion.

lucaspimentel: thanks!

lucaspimentel (approved):

build/tools/PrepareRelease/GenerateIntegrationDefinitions.cs
```diff
@@ -19,16 +19,35 @@ public static void Run(params string[] outputDirectories)
...
             assemblies.Add(typeof(Instrumentation).Assembly);
 
-            // find all methods in Datadog.Trace.ClrProfiler.Managed.dll with [InstrumentMethod]
+            // Extract all InstrumentMethodAttribute at assembly scope level
+            var assemblyInstrumentMethodAttributes = from assembly in assemblies
+                                                     let attributes = assembly.GetCustomAttributes(inherit: false)
+                                                            .Select(a => a as InstrumentMethodAttribute)
+                                                            .Where(a => a != null).ToList()
+                                                     from attribute in attributes
+                                                     let callTargetClassCheck = attribute.CallTargetType ?? throw new NullReferenceException("The usage of InstrumentMethodAttribute in assembly scope must define the CallTargetClass property.")
```
zacharycmontoya, line 28: Can you include the type name of the attribute and maybe the target type / target method name strings in the exception message to make it easier to identify the attribute that was not constructed correctly?

zacharycmontoya: Or maybe we can just define a `ToString()` implementation and use that?

tonyredondo: Done in 7189594

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AdoNet/AdoNetClientInstrumentMethodAttribute.cs
```diff
@@ -0,0 +1,125 @@
...
+        }
+
+        protected IAdoNetClientData AdoNetClientData { get; }
+
+        internal class CommandExecuteNonQueryAsyncAttribute : AdoNetClientInstrumentMethodAttribute
+        {
+            public CommandExecuteNonQueryAsyncAttribute(Type adoNetClientDataType)
+                : base(adoNetClientDataType)
+            {
+                MethodName = AdoNetConstants.MethodNames.ExecuteNonQueryAsync;
```
zacharycmontoya, line 25: Instead of assigning these values in constructors, could the public properties just be overridden? This infrastructure seems like more work than is needed, but perhaps I don't have the full understanding

build/tools/PrepareRelease/GenerateIntegrationDefinitions.cs
zacharycmontoya: Thanks!

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AdoNet/AdoNetClientInstrumentMethodAttribute.cs
zacharycmontoya: Oh wait, you're trying to re-use one attribute to represent the same method but just have different assembly names and return types. Okay so we do need some sort of parameterization to make it simple like you're envisioning

tonyredondo: well this is only done once for all possible ADO.NET integrations, the usage of these attributes are in the SqlClientDefinitions.cs file.

zacharycmontoya: Got it. So I guess my only feedback here is, would it make sense to override the properties for the constants that do not depend on the `adoNetClientDataType` and then in the constructor set the ones that do? It's a small refactor that might might improve readability, in my opinion

tonyredondo: Oh sure, I'll do that

tonyredondo: do you prefer it like this?

image

tonyredondo: done in 15c7af7

zacharycmontoya: Sorry, I didn't see this in time! After seeing the effects of the changes, I don't think it improves the readability like I thought it did. If you prefer this way let's keep it, otherwise my slight preference is to revert this latest commit. Sorry to waste your time like that

zacharycmontoya (approved):

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AdoNet/AdoNetClientInstrumentMethodAttribute.cs
tonyredondo: don't worry that commit has been reverted. ;)

PR #1174, opened 2021-01-28
Ignore agent and port when in AAS (dogstatsd)

colin-higgins:
Make internal metrics AAS aware.

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace/Tracer.cs
+++ b/src/Datadog.Trace/Tracer.cs
@@ -667,12 +667,17 @@ namespace Datadog.Trace
                 }
 
                 var statsd = new DogStatsdService();
-                statsd.Configure(new StatsdConfig
+                if (AzureAppServices.Metadata.IsRelevant)
                 {
-                    StatsdServerName = settings.AgentUri.DnsSafeHost,
-                    StatsdPort = port,
-                    ConstantTags = constantTags.ToArray()
-                });
+                    statsd.Configure(new StatsdConfig
+                    {
+                        ConstantTags = constantTags.ToArray()
+                    });
+                }
+                else
+                {
+                    statsd.Configure(new StatsdConfig { StatsdServerName = settings.AgentUri.DnsSafeHost, StatsdPort = port, ConstantTags = constantTags.ToArray() });
+                }
 
                 return statsd;
             }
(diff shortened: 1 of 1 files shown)


src/Datadog.Trace/Tracer.cs
```diff
@@ -667,12 +667,19 @@ private static IDogStatsd CreateDogStatsdClient(TracerSettings settings, string
                 }
 
                 var statsd = new DogStatsdService();
-                statsd.Configure(new StatsdConfig
+                if (AzureAppServices.Metadata.IsRelevant)
```
lucaspimentel, line 670: Instead of checking for AAS, should we check if named pipes are enabled (regardless of environment)?

src/Datadog.Trace/Tracer.cs
```diff
@@ -667,12 +667,19 @@ private static IDogStatsd CreateDogStatsdClient(TracerSettings settings, string
...
+                    // Environment variables set by the Azure App Service extension are used internally.
+                    // Setting the server name will force UDP, when we need named pipes.
+                    statsd.Configure(new StatsdConfig
+                    {
+                        ConstantTags = constantTags.ToArray()
+                    });
+                }
+                else
+                {
+                    statsd.Configure(new StatsdConfig { StatsdServerName = settings.AgentUri.DnsSafeHost, StatsdPort = port, ConstantTags = constantTags.ToArray() });
```
lucaspimentel, line 681: For readability, can you split this into multiple lines like the previous code?

kevingosse (approved):

src/Datadog.Trace/Tracer.cs
colin-higgins: I thought about this, and it's subject to change in the future.
Right now, named pipes usage for dogstatsd is a combination the pipe variable being set, and the absence of the agent host variable.
I'd like to work with Olivier to make this more explicit for stats.

lucaspimentel: I agree it should be fixed in the dogstatsd client, but right now, if I wanted to enable named pipes on my machine (outside of AAS), wouldn't I still have the issue you are trying fixing?

colin-higgins: If you wanted runtime metrics, and also named pipes on your local machine, you would need to currently set `DD_AZURE_APP_SERVICES:true` plus all of the named pipes variables.
The named pipe is an undocumented and not officially supported feature, so I don't see it being used outside of AAS until we explicitly decide to expand support for it.

lucaspimentel: I see. Thanks for clarifying! I'll resolve this thread.

tonyredondo (approved):

PR #1176, opened 2021-01-29
Exponential Backoff Named Pipe Check

colin-higgins:
Reduce transient named pipe conflicts in AAS.
Reduces timespan where stats are sent into the abyss.

### netcore31
Tested cold startup.
Tested deployment over live app with concurrent requests which resulted in these logs for the trace agent as an example:
```
2021-01-29 17:15:56.721 +00:00 [INF] Using NamedPipeClientStreamFactory for trace transport, with pipe name datadogtrace-F243918E-6DAE-4F03-B0E3-2BAB4DB686B8 and timeout 100ms.
{ MachineName: ".", Process: "[27556 w3wp]", AppDomain: "[1 BareboneStats]", TracerVersion: "1.22.1.0" }
2021-01-29 17:15:57.031 +00:00 [INF] DATADOG TRACER CONFIGURATION - {"date":"2021-01-29T17:15:57.0224985+00:00","os_name":"Windows","os_version":"Microsoft Windows NT 10.0.14393.0","version":"1.22.1.0","platform":"x86","lang":".NET Core","lang_version":"3.1.11","env":"dd-stats-debug-2","enabled":true,"service":"BareboneStats","agent_url":"http://127.0.0.1:8126/","debug":false,"analytics_enabled":false,"sample_rate":null,"sampling_rules":null,"tags":[],"log_injection_enabled":false,"runtime_metrics_enabled":true,"disabled_integrations":[],"netstandard_enabled":false,"agent_reachable":true,"agent_error":""}
{ MachineName: ".", Process: "[27556 w3wp]", AppDomain: "[1 BareboneStats]", TracerVersion: "1.22.1.0" }
2021-01-29 17:16:00.402 +00:00 [INF] Recovering from previous C:\home\SiteExtensions\Datadog.Development.AzureAppServices\v100_5_3\Agent\datadog-trace-agent.exe shutdown. Ready for start.
{ MachineName: ".", AppDomain: "[1 BareboneStats]", TracerVersion: "1.22.1.0" }
2021-01-29 17:16:00.409 +00:00 [INF] Attempting to start C:\home\SiteExtensions\Datadog.Development.AzureAppServices\v100_5_3\Agent\datadog-trace-agent.exe.
{ MachineName: ".", AppDomain: "[1 BareboneStats]", TracerVersion: "1.22.1.0" }
2021-01-29 17:16:00.509 +00:00 [INF] Successfully started C:\home\SiteExtensions\Datadog.Development.AzureAppServices\v100_5_3\Agent\datadog-trace-agent.exe.
{ MachineName: ".", AppDomain: "[1 BareboneStats]", TracerVersion: "1.22.1.0" }
```

### net472

Tested cold start.
Tested deploy over the top of framework app with concurrent requests.
Interestingly I have not hit the same code path in framework. Possibly it's a more clean cut over?
Either way, it works.
```
2021-01-29 19:01:26.804 +00:00 [INF] Attempting to start C:\home\SiteExtensions\Datadog.Development.AzureAppServices\v100_5_4\Agent\datadog-trace-agent.exe.
{ MachineName: ".", AppDomain: "[2 /LM/W3SVC/1717539916/ROOT-1-132564204831634214]", TracerVersion: "1.22.1.0" }
2021-01-29 19:01:26.913 +00:00 [INF] Using NamedPipeClientStreamFactory for trace transport, with pipe name datadogtrace-10D77F83-F7C9-404F-ABC2-DF3B4389E212 and timeout 100ms.
{ MachineName: ".", Process: "[28092 w3wp]", AppDomain: "[2 /LM/W3SVC/1717539916/ROOT-1-132564204831634214]", TracerVersion: "1.22.1.0" }
2021-01-29 19:01:27.054 +00:00 [INF] DATADOG TRACER CONFIGURATION - {"date":"2021-01-29T19:01:27.0540606+00:00","os_name":"Windows","os_version":"Microsoft Windows NT 10.0.14393.0","version":"1.22.1.0","platform":"x86","lang":".NET Framework","lang_version":"4.8","env":"aas-stats-debug-framework","enabled":true,"service":"dd-aas-stats-debug-framework","agent_url":"http://127.0.0.1:8126/","debug":false,"analytics_enabled":false,"sample_rate":null,"sampling_rules":null,"tags":[],"log_injection_enabled":false,"runtime_metrics_enabled":true,"disabled_integrations":[],"netstandard_enabled":false,"agent_reachable":true,"agent_error":""}
{ MachineName: ".", Process: "[28092 w3wp]", AppDomain: "[2 /LM/W3SVC/1717539916/ROOT-1-132564204831634214]", TracerVersion: "1.22.1.0" }
2021-01-29 19:01:28.038 +00:00 [INF] Attempting to start C:\home\SiteExtensions\Datadog.Development.AzureAppServices\v100_5_4\Agent\dogstatsd.exe.
{ MachineName: ".", AppDomain: "[2 /LM/W3SVC/1717539916/ROOT-1-132564204831634214]", TracerVersion: "1.22.1.0" }
2021-01-29 19:01:28.507 +00:00 [INF] Successfully started C:\home\SiteExtensions\Datadog.Development.AzureAppServices\v100_5_4\Agent\datadog-trace-agent.exe.
{ MachineName: ".", AppDomain: "[2 /LM/W3SVC/1717539916/ROOT-1-132564204831634214]", TracerVersion: "1.22.1.0" }
2021-01-29 19:01:31.350 +00:00 [INF] Successfully started C:\home\SiteExtensions\Datadog.Development.AzureAppServices\v100_5_4\Agent\dogstatsd.exe.
{ MachineName: ".", AppDomain: "[2 /LM/W3SVC/1717539916/ROOT-1-132564204831634214]", TracerVersion: "1.22.1.0" }
```
@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace/TracingProcessManager.cs
+++ b/src/Datadog.Trace/TracingProcessManager.cs
@@ -43,7 +43,6 @@ namespace Datadog.Trace
         };
 
         private static readonly ILogger Log = DatadogLogging.For<TracingProcessManager>();
-        private static CancellationTokenSource _cancellationTokenSource;
 
         internal enum ProcessState
         {
@@ -73,8 +72,6 @@ namespace Datadog.Trace
                     Log.Warning("Directory for trace agent does not exist: {Directory}", TraceAgentMetadata.DirectoryPath);
                     return;
                 }
-
-                _cancellationTokenSource = new CancellationTokenSource();
 
                 Log.Debug("Starting child processes from process {ProcessName}, AppDomain {AppDomain}.", DomainMetadata.ProcessName, DomainMetadata.AppDomainName);
                 StartProcesses();
@@ -113,16 +110,12 @@ namespace Datadog.Trace
                 {
                     try
                     {
+                        // Indicate that there is a worker responsible for doing this
+                        // This acts as a lock
                         metadata.IsBeingManaged = true;
 
                         while (true)
                         {
-                            if (_cancellationTokenSource.Token.IsCancellationRequested)
-                            {
-                                Log.Debug("Shutdown triggered for keep alive {Process}.", path);
-                                return;
-                            }
-
                             if (metadata.SequentialFailures >= MaxFailures)
                             {
                                 Log.Error("Circuit breaker triggered for {Process}. Max retries reached ({ErrorCount}).", path, MaxFailures);
@@ -170,7 +163,7 @@ namespace Datadog.Trace
                                 }
                                 else if (metadata.ProcessState == ProcessState.Healthy || metadata.ProcessState == ProcessState.Faulted)
                                 {
-                                    // This means we have tried to start from this domain before and we're in a keep alive check
+                                    // This means we have tried to start from this domain before
                                     metadata.ProcessState = metadata.NamedPipeIsBound() ? ProcessState.Healthy : ProcessState.ReadyToStart;
                                 }
 
@@ -178,7 +171,11 @@ namespace Datadog.Trace
                                 {
                                     Log.Debug("Attempting to start {Process}.", path);
 
-                                    var startInfo = new ProcessStartInfo { FileName = path };
+                                    var startInfo = new ProcessStartInfo
+                                    {
+                                        FileName = path,
+                                        UseShellExecute = false
+                                    };
 
                                     if (!string.IsNullOrWhiteSpace(metadata.ProcessArguments))
                                     {
--- a/src/Datadog.Trace/TracingProcessManager.cs
+++ b/src/Datadog.Trace/TracingProcessManager.cs
@@ -144,7 +144,7 @@ namespace Datadog.Trace
 
                                         while (--attempts > 0)
                                         {
-                                            await Task.Delay((int)delay);
+                                            await Task.Delay((int)delay).ConfigureAwait(false);
 
                                             if (metadata.ProcessIsHealthy())
                                             {
@@ -208,7 +208,7 @@ namespace Datadog.Trace
                                             break;
                                         }
 
-                                        await Task.Delay(100);
+                                        await Task.Delay(100).ConfigureAwait(false);
                                         timeout -= 100;
                                     }
                                 }
@@ -224,12 +224,12 @@ namespace Datadog.Trace
                                 {
                                     metadata.SequentialFailures++;
                                     // Quicker retry in these cases
-                                    await Task.Delay(ExceptionRetryInterval);
+                                    await Task.Delay(ExceptionRetryInterval).ConfigureAwait(false);
                                 }
                                 else
                                 {
                                     // Delay for a reasonable amount of time before we check to see if the process is alive again.
-                                    await Task.Delay(KeepAliveInterval);
+                                    await Task.Delay(KeepAliveInterval).ConfigureAwait(false);
                                 }
                             }
                         }
--- a/src/Datadog.Trace/TracingProcessManager.cs
+++ b/src/Datadog.Trace/TracingProcessManager.cs
@@ -117,7 +117,7 @@ namespace Datadog.Trace
                         {
                             if (metadata.SequentialFailures >= MaxFailures)
                             {
-                                Log.Error("Circuit breaker triggered for {Process}. Max retries reached ({ErrorCount}).", path, MaxFailures);
+                                Log.Error("Maximum retries ({ErrorCount}) reached starting {Process}.", path, MaxFailures);
                                 metadata.ProcessState = ProcessState.Faulted;
                                 return;
                             }
@@ -202,7 +202,7 @@ namespace Datadog.Trace
 
                                         if (metadata.Process == null || metadata.Process.HasExited)
                                         {
-                                            Log.Error("{Process} has failed to start.", path);
+                                            Log.Error("Failed to start {Process}.", path);
                                             metadata.ProcessState = ProcessState.Faulted;
                                             break;
                                         }
@@ -215,7 +215,7 @@ namespace Datadog.Trace
                             catch (Exception ex)
                             {
                                 metadata.ProcessState = ProcessState.Faulted;
-                                Log.Error(ex, "Exception when trying to start an instance of {Process}.", path);
+                                Log.Error(ex, "Failed to start {Process}.", path);
                             }
                             finally
                             {
--- a/src/Datadog.Trace/TracingProcessManager.cs
+++ b/src/Datadog.Trace/TracingProcessManager.cs
@@ -17,11 +17,13 @@ namespace Datadog.Trace
     internal class TracingProcessManager
     {
         internal static readonly int KeepAliveInterval = 10_000;
-        internal static readonly int ExceptionRetryInterval = 1_000;
+        internal static readonly int ExceptionRetryInterval = 2_000;
+        internal static readonly int MaxFailures = 5;
 
         internal static readonly ProcessMetadata TraceAgentMetadata = new ProcessMetadata
         {
             Name = "datadog-trace-agent",
+            PipeName = EnvironmentHelpers.GetEnvironmentVariable(ConfigurationKeys.TracesPipeName),
             ProcessPath = EnvironmentHelpers.GetEnvironmentVariable(ConfigurationKeys.TraceAgentPath),
             ProcessArguments = EnvironmentHelpers.GetEnvironmentVariable(ConfigurationKeys.TraceAgentArgs),
         };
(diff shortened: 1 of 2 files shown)


src/Datadog.Trace/TracingProcessManager.cs
```diff
@@ -115,65 +123,112 @@ private static Task StartProcessWithKeepAlive(ProcessMetadata metadata)
...
+                                    else
+                                    {
+                                        // If no pipe is bound, kick it off
+                                        metadata.ProcessState = ProcessState.ReadyToStart;
+                                    }
                                 }
+                                else if (metadata.ProcessState == ProcessState.Healthy || metadata.ProcessState == ProcessState.Faulted)
+                                {
+                                    // This means we have tried to start from this domain before and we're in a keep alive check
+                                    if (!metadata.NamedPipeIsBound())
```
andrewlock, line 171: nit: I'd prefer to not have negation in the condition where possible, so `if (metadata.NamedPipeIsBound())`

src/Datadog.Trace/TracingProcessManager.cs
```diff
@@ -115,65 +123,112 @@ private static Task StartProcessWithKeepAlive(ProcessMetadata metadata)
...
+                                            delay = delay * 1.75d;
+                                        }
+
+                                        metadata.ProcessState = ProcessState.Healthy;
+                                    }
+                                    else
+                                    {
+                                        // If no pipe is bound, kick it off
+                                        metadata.ProcessState = ProcessState.ReadyToStart;
+                                    }
```
andrewlock, line 166: So if the pipe _is_ bound, and remains so, does that mean we'll lose 3+ seconds at the start? That seems like quite a lot? Or do we use the pipe if it's available?

src/Datadog.Trace/TracingProcessManager.cs
```diff
@@ -182,12 +237,15 @@ private static Task StartProcessWithKeepAlive(ProcessMetadata metadata)
                         metadata.IsBeingManaged = false;
                     }
                 },
-                _cancellationTokenSource.Token);
+                TaskCreationOptions.LongRunning);
```
andrewlock, line 240: Given the lambda uses `await`, doesn't that negate the utility of `TaskCreationOptions.LongRunning`, as the thread will be returned anyway?

andrewlock (approved):

src/Datadog.Trace/TracingProcessManager.cs
colin-higgins: We use the pipe if it's available.
Currently, if there is a conflict on the first try, we lose stats for ~10 seconds.
This conflict arises if a named pipe has not been cleaned up from a previous instance.
The first process start fails because the named pipe is not available to bind to, and the next retry doesn't come through for 10 seconds, even though the pipe might be cleared in 100ms.

This mechanism is to check the named pipe in quicker intervals on startup.
If a named pipe is bound, it will check in 50ms, then 87ms, then 153 ms, and so on until a max wait of 1436ms, which sums to approximately 3.28 seconds.

Another improvement here which is less obvious, the first failed attempt goes away.
Previously, we were deciding to start if the process was not running, but the named pipe remains around after process kill in testing, so the named pipe being bound is a more accurate timing check.
This means no more (or at least incredibly rare) transient error messages on startup.

andrewlock: Ah ok, I think the point I was missing is that there's (presumably) no way for a pipe to transition from bound -> unbound -> bound inside the `ProcessState.NotRunning` branch. You'll only ever see a bound -> unbound transition (as it's rebound in the `ProcessState.Healthy || ProcessState.Faulted` branch).

colin-higgins: Unless there are multiple processes or even app domains running. :)

andrewlock: ok... then if that _is_ a possibility, (bound -> unbound -> bound), then there's potential for a race condition here where you wait for 3s unnecessarily right? I'm not sure if a) if that's even avoidable without the long lived process approach, b) if it's a problem 🙂

colin-higgins: Yes, any separate process or app domain will run into this delay for this task.
This is totally acceptable to me, as it doesn't block the application and makes the agent processes much more robust.

tonyredondo (approved):

src/Datadog.Trace/TracingProcessManager.cs
```diff
@@ -104,91 +109,145 @@ private static Task StartProcessWithKeepAlive(ProcessMetadata metadata)
...
+                                        var delay = 50d;
+
+                                        while (--attempts > 0)
+                                        {
+                                            await Task.Delay((int)delay);
+
+                                            if (metadata.ProcessIsHealthy())
+                                            {
+                                                // Should result in a max delay of ~3.28 seconds before giving up
+                                                delay = delay * 1.75d;
```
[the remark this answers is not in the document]
colin-higgins: Experimentation in a reasonable delay structure for making sure the process is started if the previous one drops.
Do you think this should be specifically documented?

src/Datadog.Trace/TracingProcessManager.cs
```diff
@@ -104,91 +109,145 @@ private static Task StartProcessWithKeepAlive(ProcessMetadata metadata)
...
                         metadata.IsBeingManaged = true;
-                        var sequentialFailures = 0;
 
                         while (true)
                         {
-                            if (_cancellationTokenSource.Token.IsCancellationRequested)
+                            if (metadata.SequentialFailures >= MaxFailures)
                             {
-                                Log.Debug("Shutdown triggered for keep alive {Process}.", path);
+                                Log.Error("Circuit breaker triggered for {Process}. Max retries reached ({ErrorCount}).", path, MaxFailures);
```
lucaspimentel, line 120:
```suggestion
                                Log.Error("Maximum retries ({ErrorCount}) reached when starting {Process}.", MaxFailures, path);
```

src/Datadog.Trace/TracingProcessManager.cs
```diff
@@ -104,91 +109,145 @@ private static Task StartProcessWithKeepAlive(ProcessMetadata metadata)
...
+                                        {
+                                            metadata.SequentialFailures = 0;
+                                            metadata.ProcessState = ProcessState.Healthy;
+                                            Log.Information("Successfully started {Process}.", path);
+                                            break;
+                                        }
+
+                                        if (metadata.Process == null || metadata.Process.HasExited)
+                                        {
+                                            Log.Error("{Process} has failed to start.", path);
```
lucaspimentel, line 205: Small tweak for consistency with the `Successfully started {Process}` message above.
```suggestion
                                            Log.Error("Failed to start {Process}.", path);
```

src/Datadog.Trace/TracingProcessManager.cs
```diff
@@ -104,91 +109,145 @@ private static Task StartProcessWithKeepAlive(ProcessMetadata metadata)
...
+                                        await Task.Delay(100).ConfigureAwait(false);
+                                        timeout -= 100;
+                                    }
                                 }
                             }
                             catch (Exception ex)
                             {
-                                metadata.IsFaulted = true;
+                                metadata.ProcessState = ProcessState.Faulted;
                                 Log.Error(ex, "Exception when trying to start an instance of {Process}.", path);
```
lucaspimentel, line 218:
```suggestion
                                Log.Error(ex, "Failed to start {Process}.", path);
```

src/Datadog.Trace/TracingProcessManager.cs
```diff
@@ -261,6 +320,11 @@ public bool ProgramIsRunning()
                     return false;
                 }
             }
+
+            private bool NamedPipeIsBound()
```
lucaspimentel, line 323: Is this one line worth the additional method call that we can't inline?

lucaspimentel (commented):
LGTM. Some small tweaks to log messages.

src/Datadog.Trace/TracingProcessManager.cs
colin-higgins: I liked the readability here. In the usual circumstance, this code is called once every 10 seconds.
If you think it's important to combine them into one method I can.

zacharycmontoya (approved):

src/Datadog.Trace/TracingProcessManager.cs
lucaspimentel: Nope, it's good. Thanks for the context.

PR #1186, opened 2021-02-03
RabbitMQ CallTarget Integration

zacharycmontoya:
This PR migrates the `RabbitMQ` integration to CallTarget instrumentation.

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/RabbitMQ/BasicGetIntegration.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/RabbitMQ/BasicGetIntegration.cs
@@ -1,10 +1,5 @@
 using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
 using Datadog.Trace.ClrProfiler.CallTarget;
-using Datadog.Trace.Configuration;
 using Datadog.Trace.DuckTyping;
 using Datadog.Trace.Logging;
 using Datadog.Trace.Tagging;
--- a/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/RabbitMQ/BasicGetIntegration.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/RabbitMQ/BasicGetIntegration.cs
@@ -29,18 +29,6 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.RabbitMQ
 
         private static readonly Vendors.Serilog.ILogger Log = DatadogLogging.GetLogger(typeof(BasicGetIntegration));
 
-        private static Func<IDictionary<string, object>, string, IEnumerable<string>> headersGetter = ((carrier, key) =>
-        {
-            if (carrier.TryGetValue(key, out object value) && value is byte[] bytes)
-            {
-                return new[] { Encoding.UTF8.GetString(bytes) };
-            }
-            else
-            {
-                return Enumerable.Empty<string>();
-            }
-        });
-
         /// <summary>
         /// OnMethodBegin callback
         /// </summary>
@@ -83,7 +71,7 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.RabbitMQ
                 {
                     try
                     {
-                        propagatedContext = SpanContextPropagator.Instance.Extract(basicPropertiesHeaders, headersGetter);
+                        propagatedContext = SpanContextPropagator.Instance.Extract(basicPropertiesHeaders, ContextPropagation.HeadersGetter);
                     }
                     catch (Exception ex)
                     {
--- a/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/RabbitMQ/BasicGetIntegration.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/RabbitMQ/BasicGetIntegration.cs
@@ -1,0 +1,129 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Datadog.Trace.ClrProfiler.CallTarget;
+using Datadog.Trace.Configuration;
+using Datadog.Trace.DuckTyping;
+using Datadog.Trace.Logging;
+using Datadog.Trace.Tagging;
+
+namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.RabbitMQ
+{
+    /// <summary>
+    /// RabbitMQ.Client BasicGet calltarget instrumentation
+    /// </summary>
+    [InstrumentMethod(
+        Assembly = "RabbitMQ.Client",
+        Type = "RabbitMQ.Client.Impl.ModelBase",
+        Method = "BasicGet",
+        ReturnTypeName = "RabbitMQ.Client.BasicGetResult",
+        ParametersTypesNames = new[] { ClrNames.String, ClrNames.Bool },
+        MinimumVersion = "3.6.9",
+        MaximumVersion = "6.*.*",
+        IntegrationName = RabbitMQConstants.IntegrationName)]
+    public class BasicGetIntegration
+    {
+        private const string Command = "basic.get";
+
+        private static readonly Vendors.Serilog.ILogger Log = DatadogLogging.GetLogger(typeof(BasicGetIntegration));
+
+        private static Func<IDictionary<string, object>, string, IEnumerable<string>> headersGetter = ((carrier, key) =>
+        {
+            if (carrier.TryGetValue(key, out object value) && value is byte[] bytes)
+            {
+                return new[] { Encoding.UTF8.GetString(bytes) };
+            }
+            else
+            {
+                return Enumerable.Empty<string>();
+            }
+        });
+
+        /// <summary>
+        /// OnMethodBegin callback
+        /// </summary>
+        /// <typeparam name="TTarget">Type of the target</typeparam>
+        /// <param name="instance">Instance value, aka `this` of the instrumented method.</param>
+        /// <param name="queue">The queue name of the message</param>
+        /// <param name="autoAck">The original autoAck argument</param>
+        /// <returns>Calltarget state value</returns>
+        public static CallTargetState OnMethodBegin<TTarget>(TTarget instance, string queue, bool autoAck)
+        {
+            return new CallTargetState(scope: null, state: queue, startTime: DateTimeOffset.UtcNow);
+        }
+
+        /// <summary>
+        /// OnMethodEnd callback
+        /// </summary>
+        /// <typeparam name="TTarget">Type of the target</typeparam>
+        /// <typeparam name="TResult">Type of the BasicGetResult</typeparam>
+        /// <param name="instance">Instance value, aka `this` of the instrumented method.</param>
+        /// <param name="basicGetResult">BasicGetResult instance</param>
+        /// <param name="exception">Exception instance in case the original code threw an exception.</param>
+        /// <param name="state">Calltarget state value</param>
+        /// <returns>A default CallTargetReturn to satisfy the CallTarget contract</returns>
+        public static CallTargetReturn<TResult> OnMethodEnd<TTarget, TResult>(TTarget instance, TResult basicGetResult, Exception exception, CallTargetState state)
+            where TResult : IBasicGetResult, IDuckType
+        {
+            string queue = (string)state.State;
+            DateTimeOffset? startTime = state.StartTime;
+
+            SpanContext propagatedContext = null;
+            string messageSize = null;
+
+            if (basicGetResult.Instance != null)
+            {
+                messageSize = basicGetResult.Body?.Length.ToString();
+                var basicPropertiesHeaders = basicGetResult.BasicProperties?.Headers;
+
+                // try to extract propagated context values from headers
+                if (basicPropertiesHeaders != null)
+                {
+                    try
+                    {
+                        propagatedContext = SpanContextPropagator.Instance.Extract(basicPropertiesHeaders, headersGetter);
+                    }
+                    catch (Exception ex)
+                    {
+                        Log.Error(ex, "Error extracting propagated headers.");
+                    }
+                }
+            }
+
+            using (var scope = ScopeFactory.CreateRabbitMQScope(Tracer.Instance, out RabbitMQTags tags, Command, parentContext: propagatedContext, spanKind: SpanKinds.Consumer, queue: queue, startTime: startTime))
+            {
+                if (scope != null)
+                {
+                    string queueDisplayName = string.IsNullOrEmpty(queue) || !queue.StartsWith("amq.gen-") ? queue : "<generated>";
+                    scope.Span.ResourceName = $"{Command} {queueDisplayName}";
+
+                    if (tags != null && messageSize != null)
+                    {
+                        tags.MessageSize = messageSize;
+                    }
+
+                    if (exception != null)
+                    {
+                        scope.Span.SetException(exception);
+                    }
+                }
+            }
+
+            return new CallTargetReturn<TResult>(basicGetResult);
+        }
+
+        private readonly struct IntegrationState
+        {
+            public readonly Scope Scope;
+            public readonly RabbitMQTags Tags;
+
+            public IntegrationState(Scope scope, RabbitMQTags tags)
+            {
+                Scope = scope;
+                Tags = tags;
+            }
+        }
+    }
+}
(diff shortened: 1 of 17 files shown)


src/Datadog.Trace.ClrProfiler.Managed/ScopeFactory.cs
```diff
@@ -1,5 +1,6 @@
 using System;
 using System.Data;
+using Datadog.Trace.ClrProfiler.AutoInstrumentation.RabbitMQ;
```
zacharycmontoya, line 3: I figured we'd want to consolidate the Scope creation (for both CallSite and CallTarget) somewhere. Is this class the best candidate to hold the method?

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/RabbitMQ/BasicGetIntegration.cs
```diff
@@ -0,0 +1,117 @@
...
+        /// OnMethodBegin callback
+        /// </summary>
+        /// <typeparam name="TTarget">Type of the target</typeparam>
+        /// <param name="instance">Instance value, aka `this` of the instrumented method.</param>
+        /// <param name="queue">The queue name of the message</param>
+        /// <param name="autoAck">The original autoAck argument</param>
+        /// <returns>Calltarget state value</returns>
+        public static CallTargetState OnMethodBegin<TTarget>(TTarget instance, string queue, bool autoAck)
+        {
+            return new CallTargetState(scope: null, state: queue, startTime: DateTimeOffset.UtcNow);
```
zacharycmontoya, line 42: This is the one usage of the new `CallTargetState.StartTime` property. Since we must defer scope creation until the method returns (possible distributed parent context), we need to store the original start time somewhere, so I'm proposing we add it to the CallTargetState since this may not be the only case where we do this. I'd rather modify CallTargetState than mess with our Tracer's logic of initializing `SpanContext` objects

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/RabbitMQ/BasicPublishIntegration.cs
```diff
@@ -0,0 +1,96 @@
...
+        /// <param name="instance">Instance value, aka `this` of the instrumented method.</param>
+        /// <param name="exchange">Name of the exchange.</param>
+        /// <param name="routingKey">The routing key.</param>
+        /// <param name="mandatory">The mandatory routing flag.</param>
+        /// <param name="basicProperties">The message properties.</param>
+        /// <param name="body">The message body.</param>
+        /// <returns>Calltarget state value</returns>
+        public static CallTargetState OnMethodBegin<TTarget, TBasicProperties, TBody>(TTarget instance, string exchange, string routingKey, bool mandatory, TBasicProperties basicProperties, TBody body)
+            where TBasicProperties : IBasicProperties, IDuckType
+            where TBody : IBody // Versions < 6.0.0: TBody is byte[] // Versions >= 6.0.0: TBody is ReadOnlyMemory<byte>
```
zacharycmontoya, line 47: @tonyredondo in our CallSite instrumentation, due to the changing type of the body parameter, we have two BasicPublish integrations: One for versions < 6.0.0 and one integration for versions >= 6.0.0. Will it be performant to combine them this way in our CallTarget instrumentation? Or should I implement them separately again?

src/Datadog.Trace.ClrProfiler.Managed/ScopeFactory.cs
tonyredondo: I wonder why do you need this using in the ScopeFactory, it's because the RabbitMQ constants class?

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/RabbitMQ/BasicGetIntegration.cs
tonyredondo: It's ok for me to modify the CallTargetState struct with this new parameter, however we must be careful and only defer the creation of leaf spans in a distributed tracing, because any span created between OnMethodBegin and OnMethodEnd will have the wrong parent. I think it's ok in this case.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/RabbitMQ/BasicPublishIntegration.cs
tonyredondo: It's ok keeping both combined. First I think both versions of this library cannot be loaded to the same app and context. But also the caller in this case it's a delegate with the signature:

```csharp
internal delegate CallTargetState InvokeDelegate(TTarget instance, TArg1 arg1, TArg2 arg2, TArg3 arg3, TArg4 arg4, TArg5 arg5);
```

stored in a class with the signature:

```csharp
internal static class BeginMethodHandler<TIntegration, TTarget, TArg1, TArg2, TArg3, TArg4, TArg5>
```

Being `TArg5 == TBody` if the `TBody` type changes then the delegate will be stored in another class type and it will be isolated.

tonyredondo (commented):
One thing to consider and maybe doing it in another PR is increase the number of arguments of the calltarget fastpath, In this PR I'm seeing instrumentation of methods with 7 and 8 parameters. Currently the CallTarget machinery supports up to 6 parameters for the fastpath, the slow path uses an array, meaning casting and boxing = allocations.

Because this is an official instrumentation I think we should modify and support up to 8 parameters in the fast path.

For reference:
https://github.com/DataDog/dd-trace-dotnet/blob/master/src/Datadog.Trace.ClrProfiler.Native/calltarget_tokens.h#L161-L166

and

https://github.com/DataDog/dd-trace-dotnet/blob/master/src/Datadog.Trace.ClrProfiler.Managed/CallTarget/CallTargetInvoker.cs#L168-L198

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/RabbitMQ/BasicGetIntegration.cs
zacharycmontoya: Yes this is a leaf span, so this should be fine

src/Datadog.Trace.ClrProfiler.Managed/ScopeFactory.cs
zacharycmontoya: Yes, it's because of the RabbitMQ constants class

zacharycmontoya:
> One thing to consider and maybe doing it in another PR is increase the number of arguments of the calltarget fastpath, In this PR I'm seeing instrumentation of methods with 7 and 8 parameters. Currently the CallTarget machinery supports up to 6 parameters for the fastpath, the slow path uses an array, meaning casting and boxing = allocations.
> 
> Because this is an official instrumentation I think we should modify and support up to 8 parameters in the fast path.
> 
> For reference:
> https://github.com/DataDog/dd-trace-dotnet/blob/master/src/Datadog.Trace.ClrProfiler.Native/calltarget_tokens.h#L161-L166
> 
> and
> 
> https://github.com/DataDog/dd-trace-dotnet/blob/master/src/Datadog.Trace.ClrProfiler.Managed/CallTarget/CallTargetInvoker.cs#L168-L198

Yes you're right we should modify CallTarget to use the fast path here. Yes I think a separate PR would be good for that work

tonyredondo:
> > One thing to consider and maybe doing it in another PR is increase the number of arguments of the calltarget fastpath, In this PR I'm seeing instrumentation of methods with 7 and 8 parameters. Currently the CallTarget machinery supports up to 6 parameters for the fastpath, the slow path uses an array, meaning casting and boxing = allocations.
> > Because this is an official instrumentation I think we should modify and support up to 8 parameters in the fast path.
> > For reference:
> > https://github.com/DataDog/dd-trace-dotnet/blob/master/src/Datadog.Trace.ClrProfiler.Native/calltarget_tokens.h#L161-L166
> > and
> > https://github.com/DataDog/dd-trace-dotnet/blob/master/src/Datadog.Trace.ClrProfiler.Managed/CallTarget/CallTargetInvoker.cs#L168-L198
> 
> Yes you're right we should modify CallTarget to use the fast path here. Yes I think a separate PR would be good for that work

I'll create the ticket in the backlog, I can work on that next sprint.

tonyredondo (approved):

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/RabbitMQ/BasicPublishIntegration.cs
lucaspimentel: >  I think both versions of this library cannot be loaded to the same app and context

Ha, you'd be surprised by the things we've seen...

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/RabbitMQ/ExchangeDeclareIntegration.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        /// <summary>
+        /// OnMethodBegin callback
+        /// </summary>
+        /// <typeparam name="TTarget">Type of the target</typeparam>
+        /// <param name="instance">Instance value, aka `this` of the instrumented method.</param>
+        /// <param name="exchange">Name of the exchange.</param>
+        /// <param name="type">Type of the exchange.</param>
+        /// <param name="passive">The original passive setting</param>
+        /// <param name="durable">The original duable setting</param>
```
lucaspimentel, line 31:
```suggestion
        /// <param name="durable">The original durable setting</param>
```

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/RabbitMQ/QueueDeclareIntegration.cs
```diff
@@ -0,0 +1,55 @@
...
+        private const string Command = "queue.declare";
+
+        /// <summary>
+        /// OnMethodBegin callback
+        /// </summary>
+        /// <typeparam name="TTarget">Type of the target</typeparam>
+        /// <param name="instance">Instance value, aka `this` of the instrumented method.</param>
+        /// <param name="queue">Name of the queue.</param>
+        /// <param name="passive">The original passive setting</param>
+        /// <param name="durable">The original duable setting</param>
```
lucaspimentel, line 30:
```suggestion
        /// <param name="durable">The original durable setting</param>
```

src/Datadog.Trace.ClrProfiler.Managed/CallTarget/CallTargetState.cs
```diff
@@ -32,13 +35,29 @@ public CallTargetState(Scope scope, object state)
...
+        /// </summary>
+        /// <param name="scope">Scope instance</param>
+        /// <param name="state">Object state instance</param>
+        /// <param name="startTime">The intended start time of the scope, intended for scopes created in the OnMethodEnd handler</param>
+        public CallTargetState(Scope scope, object state, DateTimeOffset? startTime)
+        {
+            _previousScope = null;
+            _scope = scope;
+            _state = state;
+            _startTime = startTime;
```
lucaspimentel, line 52: Is there a reason why we don't chain the constructors?
Same comment for `CallTargetState(Scope, Scope, object)` below.

```suggestion
        public CallTargetState(Scope scope, object state, DateTimeOffset? startTime)
            : this(scope, state)
        {
            _startTime = startTime;
```

src/Datadog.Trace.ClrProfiler.Managed/ScopeFactory.cs
lucaspimentel: Not asking for any change in this PR, but I'm wondering if grouping scope creating for all the disparate instrumentations makes sense, or it this something that should be part of each integration? There are only a few that are shared like http and adonet clients, but the rest have nothing in common.

src/Datadog.Trace.ClrProfiler.Managed/ScopeFactory.cs
```diff
@@ -191,6 +192,48 @@ public static Scope CreateDbCommandScope(Tracer tracer, IDbCommand command)
             return scope;
         }
 
+        public static Scope CreateRabbitMQScope(Tracer tracer, out RabbitMQTags tags, string command, ISpanContext parentContext = null, string spanKind = SpanKinds.Client, DateTimeOffset? startTime = null, string queue = null, string exchange = null, string routingKey = null)
+        {
+            tags = null;
+
+            if (!tracer.Settings.IsIntegrationEnabled(RabbitMQConstants.IntegrationId))
+            {
+                // integration disabled, don't create a scope, skip this trace
```
lucaspimentel, line 201: I probably started this trend 😅 , but I see this comment copied everywhere and it should say `...skip this span`.
One day I'll create a PR with a quick search and replace...

src/Datadog.Trace.ClrProfiler.Managed/ScopeFactory.cs
```diff
@@ -191,6 +192,48 @@ public static Scope CreateDbCommandScope(Tracer tracer, IDbCommand command)
...
+                span.Type = SpanTypes.Queue;
+                span.ResourceName = command;
+                tags.Command = command;
+
+                tags.Queue = queue;
+                tags.Exchange = exchange;
+                tags.RoutingKey = routingKey;
+
+                tags.InstrumentationName = RabbitMQConstants.IntegrationName;
+                tags.SetAnalyticsSampleRate(RabbitMQConstants.IntegrationId, tracer.Settings, enabledWithGlobalSetting: false);
```
lucaspimentel, line 225: I think we want `enabledWithGlobalSetting: true` if `spanKind == SpanKinds.Server`.

src/Datadog.Trace.ClrProfiler.Managed/ScopeFactory.cs
```diff
@@ -191,6 +192,48 @@ public static Scope CreateDbCommandScope(Tracer tracer, IDbCommand command)
             return scope;
         }
 
+        public static Scope CreateRabbitMQScope(Tracer tracer, out RabbitMQTags tags, string command, ISpanContext parentContext = null, string spanKind = SpanKinds.Client, DateTimeOffset? startTime = null, string queue = null, string exchange = null, string routingKey = null)
```
lucaspimentel, line 195: Shouldn't we use `SpanKinds.Producer` and `SpanKinds.Consumer` for RabbitMQ? Too late to change now until Tracer 2.0?

lucaspimentel, line 195: Can we make `spanKind` not optional? I can see this leading to subtle bugs because we forgot to specify the span kind somewhere.

zacharycmontoya: Yeah we already use `SpanKinds.Producer` and `SpanKinds.Consumer`. Here's the spanKind for each operation:
- Consumers use `SpanKinds.Consumer`: BasicDeliver, BasicGet
- Producers use `SpanKinds.Producer`: BasicPublish
- Queue/Exchange operations default to `SpanKinds.Client`: ExchangeDeclare, QueueBind, QueueDeclare

So, nothing needs to change here, though I like your suggestion to make spanKind a required parameter.

zacharycmontoya: Yeah you're right, I'll make this a required parameter

zacharycmontoya: Resolved in 26f8024013e710203be90c3e9143cc6eb8e13d77

src/Datadog.Trace.ClrProfiler.Managed/ScopeFactory.cs
zacharycmontoya: You're right it should be for each integration. I was trying to figure out the best place to share the scope creation between CallSite and CallTarget. I'll move the RabbitMQ scope creation back to the original RabbitMQIntegration and just have the CallTarget instrumentation use that

src/Datadog.Trace.ClrProfiler.Managed/ScopeFactory.cs
lucaspimentel: Thanks. I should have looked at the code that calls into this. I just saw the default `SpanKinds.Client` and assumed we were only using client & server.

src/Datadog.Trace.ClrProfiler.Managed/CallTarget/CallTargetState.cs
zacharycmontoya: My only reason for not chaining them is that struct's have to have all their fields initialized before being returned to the caller, so keeping the entire set of fields initializations in one method makes it pretty clear that we're doing this (even though the compiler would verify that of course).

src/Datadog.Trace.ClrProfiler.Managed/ScopeFactory.cs
zacharycmontoya: Perhaps an exercise left for the reader? 😆

src/Datadog.Trace.ClrProfiler.Managed/ScopeFactory.cs
zacharycmontoya: Does this apply for queueing technologies? I can add it if `spanKind == SpanKinds.Consumer` but I still don't fully understand what the effect of this change is

lucaspimentel: Never mind. I checked our RFC and `false` is the correct setting. We should only enable analytics with the global setting in web frameworks spans (aspnet, mvc, webapi, etc). Resolving this comment.

lucaspimentel (approved):

PR #1192, opened 2021-02-04
Properly drain stdout/stderr in integration tests

kevingosse:
Previous code was using `.ReadToEnd()` to drain the output. It meant that stderr was only drained when stdout was closed (= when the process exited) which could lead to deadlock situations.

### DIFF
--- a/test/Datadog.Trace.ClrProfiler.IntegrationTests/ProcessHelper.cs
+++ b/test/Datadog.Trace.ClrProfiler.IntegrationTests/ProcessHelper.cs
@@ -1,0 +1,54 @@
+using System;
+using System.Diagnostics;
+using System.Text;
+using System.Threading;
+
+namespace Datadog.Trace.ClrProfiler.IntegrationTests
+{
+    /// <summary>
+    /// Drains the standard and error output of a process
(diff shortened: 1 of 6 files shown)


test/Datadog.Trace.ClrProfiler.IntegrationTests/ProcessHelper.cs
```diff
@@ -0,0 +1,54 @@
...
+            process.BeginErrorReadLine();
+        }
+
+        public string StandardOutput => _outputBuffer.ToString();
+
+        public string ErrorOutput => _errorBuffer.ToString();
+
+        public bool Drain(int timeout = Timeout.Infinite)
+        {
+            return _outputMutex.Wait(timeout / 2) && _errorMutex.Wait(timeout / 2);
```
tonyredondo, line 33: I think we should have a case when timeout is Timeout.Infinite, feels wrong the `timeout/2` in that case.

tonyredondo (approved):
LGTM, only a little nit.

andrewlock (approved):

lucaspimentel (approved):

PR #1203, opened 2021-02-06
Fix RequestReJIT deadlock.

tonyredondo:
This PR fixes the RequestReJIT deadlock by moving the call from the ModuleFinished callback to a custom not CLR thread.

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
+++ b/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
@@ -2842,16 +2842,13 @@ size_t CorProfiler::CallTarget_RequestRejitForModule(ModuleID module_id, ModuleM
            ", Signature=", caller.signature.str(),
            "]");
 
-      rejit_handler->EnqueueForRejit(module_id, methodDef);
-      
       enumIterator = ++enumIterator;
     }
   }
 
   // Request the ReJIT for all integrations found in the module.
   if (!vtMethodDefs.empty()) {
-    //Info("Requesting ReJIT for ", vtMethodDefs.size(), " methods.");
-    //this->info_->RequestReJIT((ULONG)vtMethodDefs.size(), vtModules.data(), vtMethodDefs.data());
+    this->rejit_handler->EnqueueForRejit(vtMethodDefs.size(), vtModules.data(), vtMethodDefs.data());
   }
 
   // We return the number of ReJIT requests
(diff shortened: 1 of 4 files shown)


src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -2840,15 +2841,17 @@ size_t CorProfiler::CallTarget_RequestRejitForModule(ModuleID module_id, ModuleM
...
            "]");
+
+      rejit_handler->EnqueueForRejit(module_id, methodDef);
       
       enumIterator = ++enumIterator;
     }
   }
 
   // Request the ReJIT for all integrations found in the module.
   if (!vtMethodDefs.empty()) {
```
kevingosse, line 2852: Don't forget to remove vtMethodDefs and vtModules in the final version ;)

tonyredondo: yeah, I'm working on that, also to accept the complete array so we don't suspend the runtime for each single method.

tonyredondo: the method was changed in: a029fef

kevingosse (approved):
LGTM

src/Datadog.Trace.ClrProfiler.Native/rejit_handler.h
```diff
@@ -119,6 +133,47 @@ class RejitHandler {
...
+
+ private:
+  static void enqueue_thread(RejitHandler* handler) {
+    auto queue = handler->rejit_queue_;
+    auto profilerInfo = handler->profilerInfo;
+
+    Info("Initializing ReJIT request thread.");
+    HRESULT hr = profilerInfo->InitializeCurrentThread();
+    if (FAILED(hr)) {
+      Warn("Call to InitializeCurrentThread fail.");
```
zacharycmontoya, line 156: Is there anything different we need to do when this fails? Probably the answer is no, but I thought I would ask

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -592,6 +592,7 @@ HRESULT STDMETHODCALLTYPE CorProfiler::Shutdown() {
   // to prevent it from unloading while in use
   std::lock_guard<std::mutex> guard(module_id_to_info_map_lock_);
 
+  // rejit_handler->Shutdown();
```
zacharycmontoya, line 595: nit: Can we delete this line?

zacharycmontoya (approved):
LGTM! Thank you for the fix and to @kevingosse for identifying the issue!

src/Datadog.Trace.ClrProfiler.Native/rejit_handler.h
tonyredondo: Well I don't think we can do anything in that case.

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
tonyredondo: I re-enabled the shutdown process in eb1e7ab

PR #1196, opened 2021-02-05
Change the folder structure to use an Http parent folder.

andrewlock:
As suggested in https://github.com/DataDog/dd-trace-dotnet/pull/1162#discussion_r563655744

Update integrations.json version (and fix version numbers broken in
https://github.com/DataDog/dd-trace-dotnet/commit/d72b04cb694fd6c2141a59de0baa59568cd8ad49)

@DataDog/apm-dotnet

PR #1204, opened 2021-02-08
Add calltarget support for WebRequest

kevingosse:
Add CallTarget instrumentation for WebRequest.

### DIFF
--- a/integrations.json
+++ b/integrations.json
@@ -1,5 +1,88 @@
 [
   {
+    "name": "AspNetMvc",
+    "method_replacements": [
+      {
+        "caller": {},
+        "target": {
+          "assembly": "System.Web.Mvc",
+          "type": "System.Web.Mvc.Async.AsyncControllerActionInvoker",
+          "method": "BeginInvokeAction",
+          "signature_types": [
+            "System.IAsyncResult",
+            "System.Web.Mvc.ControllerContext",
+            "System.String",
+            "System.AsyncCallback",
+            "System.Object"
+          ],
+          "minimum_major": 4,
+          "minimum_minor": 0,
+          "minimum_patch": 0,
+          "maximum_major": 5,
+          "maximum_minor": 65535,
+          "maximum_patch": 65535
+        },
+        "wrapper": {
+          "assembly": "Datadog.Trace.ClrProfiler.Managed, Version=1.23.0.0, Culture=neutral, PublicKeyToken=def86d061d0d2eeb",
+          "type": "Datadog.Trace.ClrProfiler.AutoInstrumentation.AspNet.AsyncControllerActionInvoker_BeginInvokeAction_Integration",
+          "action": "CallTargetModification"
+        }
+      },
+      {
+        "caller": {},
+        "target": {
+          "assembly": "System.Web.Mvc",
+          "type": "System.Web.Mvc.Async.AsyncControllerActionInvoker",
+          "method": "EndInvokeAction",
+          "signature_types": [
+            "System.Boolean",
+            "System.IAsyncResult"
+          ],
+          "minimum_major": 4,
+          "minimum_minor": 0,
+          "minimum_patch": 0,
+          "maximum_major": 5,
+          "maximum_minor": 65535,
+          "maximum_patch": 65535
+        },
+        "wrapper": {
+          "assembly": "Datadog.Trace.ClrProfiler.Managed, Version=1.23.0.0, Culture=neutral, PublicKeyToken=def86d061d0d2eeb",
+          "type": "Datadog.Trace.ClrProfiler.AutoInstrumentation.AspNet.AsyncControllerActionInvoker_EndInvokeAction_Integration",
+          "action": "CallTargetModification"
+        }
+      }
+    ]
+  },
+  {
+    "name": "AspNetWebApi2",
+    "method_replacements": [
+      {
+        "caller": {},
+        "target": {
+          "assembly": "System.Web.Http",
+          "type": "System.Web.Http.ApiController",
+          "method": "ExecuteAsync",
+          "signature_types": [
+            "System.Threading.Tasks.Task`1<System.Net.Http.HttpResponseMessage>",
+            "System.Web.Http.Controllers.HttpControllerContext",
+            "System.Threading.CancellationToken"
+          ],
+          "minimum_major": 5,
+          "minimum_minor": 0,
+          "minimum_patch": 0,
+          "maximum_major": 5,
+          "maximum_minor": 1,
+          "maximum_patch": 65535
+        },
+        "wrapper": {
+          "assembly": "Datadog.Trace.ClrProfiler.Managed, Version=1.23.0.0, Culture=neutral, PublicKeyToken=def86d061d0d2eeb",
+          "type": "Datadog.Trace.ClrProfiler.AutoInstrumentation.AspNet.ApiController_ExecuteAsync_Integration",
+          "action": "CallTargetModification"
+        }
+      }
+    ]
+  },
+  {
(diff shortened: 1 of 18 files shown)


src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/WebRequest/WebRequestGetResponseIntegration.cs
```diff
@@ -0,0 +1,69 @@
...
+    /// <summary>
+    /// CallTarget integration for WebRequest.GetResponse
+    /// </summary>
+    [InstrumentMethod(
+        AssemblyName = WebRequestCommon.NetFrameworkAssembly,
+        TypeName = WebRequestCommon.HttpWebRequestTypeName,
+        MethodName = MethodName,
+        ReturnTypeName = WebRequestCommon.WebResponseTypeName,
+        ParameterTypeNames = new string[0],
+        MinimumVersion = WebRequestCommon.Major2,
```
zacharycmontoya, line 17: We don't need to instrument assembly version `2.0.0.0` here. Since we're doing CallTarget, only the `4.0.0.0` will be loaded. This was a workaround in CallSite where older .NET Framework applications were built against an older framework so their assembly reference was `2.0.0.0` and we had to make sure to instrument with that information

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/WebRequest/WebRequestGetRequestStreamIntegration.cs
```diff
@@ -0,0 +1,62 @@
...
+    [InstrumentMethod(
+        AssemblyName = WebRequestCommon.NetCoreAssembly,
+        TypeName = WebRequestCommon.HttpWebRequestTypeName,
+        MethodName = MethodName,
+        ReturnTypeName = ClrNames.Stream,
+        ParameterTypeNames = new string[0],
+        MinimumVersion = WebRequestCommon.Major4,
+        MaximumVersion = WebRequestCommon.Major5,
+        IntegrationName = WebRequestCommon.IntegrationName)]
+    public class WebRequestGetRequestStreamIntegration
```
zacharycmontoya, line 28: nit: Can you add `Http` to the front of this type name and `WebRequestGetResponseIntegration`?

zacharycmontoya (approved):
I left a couple of small nits but overall LGTM!

andrewlock (commented):
In [a recent PR](https://github.com/DataDog/dd-trace-dotnet/pull/1196), we moved all the HTTP auto-instrumentations into an _http_ sub-folder - I think this one should be in there too?

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/WebRequest/WebRequestGetResponseIntegration.cs
kevingosse: Alright 👍

kevingosse:
> In [a recent PR](https://github.com/DataDog/dd-trace-dotnet/pull/1196), we moved all the HTTP auto-instrumentations into an _http_ sub-folder - I think this one should be in there too?

That's HttpClient stuff. WebRequest is a different pipeline (well, at least in netfx) so I'd rather keep them separate.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/WebRequest/HttpWebRequestGetRequestStreamIntegration.cs
```diff
@@ -0,0 +1,62 @@
...
+    [InstrumentMethod(
+        AssemblyName = WebRequestCommon.NetCoreAssembly,
+        TypeName = WebRequestCommon.HttpWebRequestTypeName,
+        MethodName = MethodName,
+        ReturnTypeName = ClrNames.Stream,
+        ParameterTypeNames = new string[0],
+        MinimumVersion = WebRequestCommon.Major4,
+        MaximumVersion = WebRequestCommon.Major5,
+        IntegrationName = WebRequestCommon.IntegrationName)]
+    public class HttpWebRequestGetRequestStreamIntegration
```
kevingosse, line 28: This is a mouthful, maybe we should consider using `_` in the names?

`HttpWebRequest_GetRequestStream_Integration`

zacharycmontoya: I'm on board with that change

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/WebRequest/WebRequestGetRequestStreamIntegration.cs
kevingosse: Done

integrations.json
```diff
@@ -1,4 +1,87 @@
 [
+  {
+    "name": "AspNetMvc",
```
zacharycmontoya, line 3: Can you regenerate this `integrations.json` file? I think this may have gotten mixed up with your AspNetMvc / WebApi PR?

kevingosse: Good catch.
Took me a good while to understand why this was happening. Turns out that I GAC'd the assemblies when I was working on the ASP.NET MVC stuff.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/WebRequest/HttpWebRequest_GetRequestStream_Integration.cs
```diff
@@ -0,0 +1,62 @@
...
+{
+    /// <summary>
+    /// CallTarget integration for HttpWebRequest.GetRequestStream
+    /// </summary>
+    [InstrumentMethod(
+        AssemblyName = WebRequestCommon.NetFrameworkAssembly,
+        TypeName = WebRequestCommon.HttpWebRequestTypeName,
+        MethodName = MethodName,
+        ReturnTypeName = ClrNames.Stream,
+        ParameterTypeNames = new string[0],
```
tonyredondo, line 15: If there is no parameters you can avoid this declaration.

tonyredondo (approved):
LGTM.

tonyredondo:
> > In [a recent PR](https://github.com/DataDog/dd-trace-dotnet/pull/1196), we moved all the HTTP auto-instrumentations into an _http_ sub-folder - I think this one should be in there too?
> 
> That's HttpClient stuff. WebRequest is a different pipeline (well, at least in netfx) so I'd rather keep them separate.
> I don't have a strong opinion on this though. @tonyredondo @zacharycmontoya what do you think?

I do prefer to move it inside the `Http` folder to keep all the http client side related stuff together. I think is ok that doesn't share anything with `HttpClient` implementation.

kevingosse:
> > > In [a recent PR](https://github.com/DataDog/dd-trace-dotnet/pull/1196), we moved all the HTTP auto-instrumentations into an _http_ sub-folder - I think this one should be in there too?
> > 
> > 
> > That's HttpClient stuff. WebRequest is a different pipeline (well, at least in netfx) so I'd rather keep them separate.
> > I don't have a strong opinion on this though. @tonyredondo @zacharycmontoya what do you think?
> 
> I do prefer to move it inside the `Http` folder to keep all the http client side related stuff together. I think is ok that doesn't share anything with `HttpClient` implementation.

Alright, will do 👍

PR #1205, opened 2021-02-08
Prevent http client spans in serilog sink

colin-higgins:
Prevents noise and top level span problems.
Typically prevalent in AAS, but affects anywhere that uses the Datadog Serilog Sink package.

@DataDog/apm-dotnet


zacharycmontoya (approved):

lucaspimentel (approved):

colin-higgins:
Doesn't solve the issue.
Will require a more in depth follow up to exclude certain callers.

PR #1208, opened 2021-02-10
Add calltarget instrumentation for ASP.NET MVC and WebAPI

kevingosse:
Also converted the callsite integration to duck-typing to be able to share most of the logic.


src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AspNet/ApiController_ExecuteAsync_Integration.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// <returns>Calltarget state value</returns>
+        public static CallTargetState OnMethodBegin<TTarget, TController>(TTarget instance, TController controllerContext, CancellationToken cancellationToken)
+            where TController : IHttpControllerContext
+        {
+            var scope = AspNetWebApi2Integration.CreateScope(controllerContext, out _);
+
+            if (scope != null)
+            {
+                return new CallTargetState(scope, controllerContext);
+            }
```
tonyredondo, line 56: I think in this case we lose all the benefits of using the `constraint` trick, the `struct` created by the ducktyping will be boxed in both `CreateScope` and `new `CallTargetState`.

So for this I rather prefer to remove the `constraints` and convert the `IHttpControllerContext` to a struct with the `[DuckCopy]` attribute, and just use the `As<T>` method when we need the access, that will prevent the additional allocation.

Or...

If we accept that single allocation then do the boxing once at the beginning of the method?

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AspNet/ApiController_ExecuteAsync_Integration.cs
```diff
@@ -0,0 +1,124 @@
...
+                }
+                else
+                {
+                    // Looks like we won't be able to get the final status code
+                    scope.Dispose();
+                }
+            }
+            else
+            {
+                var statusCode = responseMessage.GetProperty("StatusCode");
```
tonyredondo, line 108: We can transform this to duck typing, so we don't mix sigil with this implementation.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AspNet/AsyncControllerActionInvoker_BeginInvokeAction_Integration.cs
```diff
@@ -0,0 +1,72 @@
...
+        public static CallTargetState OnMethodBegin<TTarget, TContext>(TTarget instance, TContext controllerContext, string actionName, AsyncCallback callback, object state)
+            where TContext : IControllerContext
+        {
+            Scope scope = null;
+
+            try
+            {
+                if (HttpContext.Current != null)
+                {
+                    scope = AspNetMvcIntegration.CreateScope(controllerContext);
```
tonyredondo, line 54: Same as previous comment related to boxing.

src/Datadog.Trace.ClrProfiler.Managed/CallTarget/Handlers/IntegrationMapper.cs
```diff
@@ -38,7 +38,7 @@ internal static DynamicMethod CreateBeginMethodDelegate(Type integrationType, Ty
             MethodInfo onMethodBeginMethodInfo = integrationType.GetMethod(BeginMethodName, BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static);
             if (onMethodBeginMethodInfo is null)
             {
-                throw new NullReferenceException($"Couldn't find the method: {BeginMethodName} in type: {integrationType.FullName}");
+                return null;
```
tonyredondo, line 41: 🤟

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/HttpHeadersCollection.cs
```diff
@@ -0,0 +1,45 @@
+#if NETFRAMEWORK
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Datadog.Trace.Headers;
+
+namespace Datadog.Trace.ClrProfiler.Integrations.AspNet
+{
+    internal readonly struct HttpHeadersCollection : IHeadersCollection
```
tonyredondo, line 9: IRequestHeaders is a proxy (struct) so I think we can make this type to use the struct + constraint trick to avoid the allocation when boxing the struct to IRequestHeaders

tonyredondo (changes requested):
Overall is looking good, I think we can make some changes to reduce allocations.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AspNet/ApiController_ExecuteAsync_Integration.cs
kevingosse: Good catch, I actually missed it

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AspNet/ApiController_ExecuteAsync_Integration.cs
kevingosse: As discussed on Slack, I'll go for the "box only once" way

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetWebApi2Integration.cs
```diff
@@ -139,21 +140,23 @@ public static class AspNetWebApi2Integration
...
         /// <param name="cancellationToken">The cancellation token</param>
         /// <returns>A task with the result</returns>
         private static async Task<T> ExecuteAsyncInternal<T>(
             Func<object, object, CancellationToken, object> instrumentedMethod,
             object apiController,
-            object controllerContext,
+            object context,
             CancellationToken cancellationToken)
         {
+            var controllerContext = context.As<IHttpControllerContext>();
```
tonyredondo, line 152: We should revisit this when we merge this: https://github.com/DataDog/dd-trace-dotnet/pull/1220

tonyredondo (approved):
LGTM

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetWebApi2Integration.cs
kevingosse: Done in https://github.com/DataDog/dd-trace-dotnet/pull/1208/commits/a13a1df56f76b3e06c54a653676e8c68dcc4fa66

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/HttpHeadersCollection.cs
zacharycmontoya: I'm still new to constrained virt calls so please excuse me if this is a silly question, but does this accomplish the struct + constraint trick? Or does this type need to be changed to have one generic type parameter `T` and then constrain that generic type parameter like `where T : IRequestHeaders`?

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetWebApi2Integration.cs
```diff
@@ -198,7 +201,7 @@ public static class AspNetWebApi2Integration
             }
         }
 
-        private static Scope CreateScope(object controllerContext, out AspNetTags tags)
+        internal static Scope CreateScope(IHttpControllerContext controllerContext, out AspNetTags tags)
```
zacharycmontoya, line 204: Can this proxy DuckType be changed from an interface to a struct?

zacharycmontoya (approved):
Overall this looks good to me. I left a couple questions to help me understand how we're using/optimizing the duck typing proxies, but don't consider it blocking

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/HttpHeadersCollection.cs
andrewlock: And I'm new to both "constrained virt calls" _and_ the "struct + constraint trick" 😄 Is there anywhere I can read more about both? 🙂

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetWebApi2Integration.cs
kevingosse: In theory yes, but the object graph is quite deep: 

```
IHttpControllerContext
 - RouteData -> Route -> RouteTemplate
 - Request -> Method -> Method
```
Without data, I'm a bit afraid that the cost of the massive struct could outweigh the benefits. I'd rather go for the easiest solution for now, and we can come back to it and write some benchmarks when we have more time.

PR #1215, opened 2021-02-16
Remove _W string operator from the native profiler

tonyredondo:
This `PR` removes the `_W` string operator from the profiler and replace it with a new `WStr` macro that redirects to a `L` prefix on windows and an `u` prefix on non windows. Simplifying the wide string declaration and reducing the `_W` overhead.

@DataDog/apm-dotnet


macrogreg:
Minor: Instead of calling it `_LU(..)` we could consider `WideStr(..)` or `WStr(..)`?

tonyredondo:
> Minor: Instead of calling it `_LU(..)` we could consider `WideStr(..)` or `WStr(..)`?

Sure, I’ll change it to WStr()

tonyredondo:
> Minor: Instead of calling it `_LU(..)` we could consider `WideStr(..)` or `WStr(..)`?

done in 5bd15f8

zacharycmontoya (approved):
LGTM! Nice clean up!

lucaspimentel (approved):

PR #1217, opened 2021-02-17
Remove heap allocations from COR_SIGNATURE definitions in the native profiler

tonyredondo:
This PR replaces the COR_SIGNATURE heap allocation to a stack allocation.

Related to this optimization: https://publicwork.wordpress.com/2019/06/27/stack-allocation-vs-heap-allocation-performance-benchmark/

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace.ClrProfiler.Native/calltarget_tokens.cpp
+++ b/src/Datadog.Trace.ClrProfiler.Native/calltarget_tokens.cpp
@@ -7,6 +7,8 @@
 
 namespace trace {
 
+const int signatureBufferSize = 50;
+
 /**
  * PRIVATE
  **/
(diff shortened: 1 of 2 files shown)


zacharycmontoya (approved):
Great change! Thank you for the continuous improvements to the codebase!

src/Datadog.Trace.ClrProfiler.Native/calltarget_tokens.cpp
```diff
@@ -79,7 +81,7 @@ HRESULT CallTargetTokens::EnsureCorLibTokens() {
     unsigned type_buffer;
     auto type_size = CorSigCompressToken(typeRef, &type_buffer);
 
-    auto* signature = new COR_SIGNATURE[runtimeTypeHandle_size + type_size + 4];
+    COR_SIGNATURE signature[signatureBufferSize];
```
lucaspimentel, line 84: [Non-blocking] I know we were not doing it before, but should we worry about zeroing out the buffer?
```suggestion
    COR_SIGNATURE signature[signatureBufferSize] {};
```

lucaspimentel (approved):
LGTM, just left one question.

src/Datadog.Trace.ClrProfiler.Native/calltarget_tokens.cpp
tonyredondo: Zeroing has a cost and I was hopping to avoid that.  When we use it we also pass the actual length so that shouldn’t be a problem.

PR #1218, opened 2021-02-17
Remove Sync-over-async in DatadogHttpClient

andrewlock:
This PR removes the [sync-over-async we were previously doing](https://github.com/DataDog/dd-trace-dotnet/blob/c23974c1cc38d96da81cc61c64ee6540eec8cd9f/src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs#L68) for simplicity.

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace/Agent/Transports/HttpStreamRequest.cs
+++ b/src/Datadog.Trace/Agent/Transports/HttpStreamRequest.cs
@@ -43,27 +43,18 @@ namespace Datadog.Trace.Agent.Transports
                 // send request, get response
                 var response = await _client.SendAsync(request, bidirectionalStream, bidirectionalStream).ConfigureAwait(false);
 
-                // buffer the entire contents for now
-                MemoryStream responseContentStream;
-                if (response.Content.Length.HasValue)
+                // Content-Length is required as we don't support chunked transfer
+                var contentLength = response.Content.Length;
+                if (!contentLength.HasValue)
                 {
-                    var buffer = new byte[response.Content.Length.Value];
-                    responseContentStream = new MemoryStream(buffer);
-                    await response.Content.CopyToAsync(buffer).ConfigureAwait(false);
-                    responseContentStream.Position = 0;
-                }
-                else
-                {
-                    responseContentStream = new MemoryStream();
-                    await response.Content.CopyToAsync(responseContentStream, ResponseReadBufferSize).ConfigureAwait(false);
-                    responseContentStream.Position = 0;
+                    throw new Exception("Content-Length is required but was not provided");
                 }
 
-                var contentLength = response.ContentLength;
-                if (contentLength != null && contentLength != responseContentStream.Length)
-                {
-                    throw new Exception("Content length from http headers does not match content's actual length.");
-                }
+                // buffer the entire contents for now
+                var buffer = new byte[contentLength.Value];
+                var responseContentStream = new MemoryStream(buffer);
+                await response.Content.CopyToAsync(buffer).ConfigureAwait(false);
+                responseContentStream.Position = 0;
 
                 return new HttpStreamResponse(response.StatusCode, responseContentStream.Length, response.GetContentEncoding(), responseContentStream);
             }
--- a/src/Datadog.Trace/Agent/Transports/HttpStreamRequest.cs
+++ b/src/Datadog.Trace/Agent/Transports/HttpStreamRequest.cs
@@ -44,9 +44,20 @@ namespace Datadog.Trace.Agent.Transports
                 var response = await _client.SendAsync(request, bidirectionalStream, bidirectionalStream).ConfigureAwait(false);
 
                 // buffer the entire contents for now
-                var responseContentStream = new MemoryStream();
-                await response.Content.CopyToAsync(responseContentStream, ResponseReadBufferSize).ConfigureAwait(false);
-                responseContentStream.Position = 0;
+                MemoryStream responseContentStream;
+                if (response.Content.Length.HasValue)
+                {
+                    var buffer = new byte[response.Content.Length.Value];
+                    responseContentStream = new MemoryStream(buffer);
+                    await response.Content.CopyToAsync(buffer).ConfigureAwait(false);
+                    responseContentStream.Position = 0;
+                }
+                else
+                {
+                    responseContentStream = new MemoryStream();
+                    await response.Content.CopyToAsync(responseContentStream, ResponseReadBufferSize).ConfigureAwait(false);
+                    responseContentStream.Position = 0;
+                }
 
                 var contentLength = response.ContentLength;
                 if (contentLength != null && contentLength != responseContentStream.Length)
--- a/src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs
+++ b/src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs
@@ -1,3 +1,4 @@
+using System;
 using System.IO;
 using System.Text;
 using System.Threading.Tasks;
@@ -21,7 +22,7 @@ namespace Datadog.Trace.HttpOverStreams
         public async Task<HttpResponse> SendAsync(HttpRequest request, Stream requestStream, Stream responseStream)
         {
             await SendRequestAsync(request, requestStream).ConfigureAwait(false);
-            return ReadResponse(responseStream);
+            return await ReadResponseAsync(responseStream);
         }
 
         private async Task SendRequestAsync(HttpRequest request, Stream requestStream)
@@ -44,10 +45,10 @@ namespace Datadog.Trace.HttpOverStreams
             await requestStream.FlushAsync().ConfigureAwait(false);
         }
 
-        private HttpResponse ReadResponse(Stream responseStream)
+        private async Task<HttpResponse> ReadResponseAsync(Stream responseStream)
         {
             var headers = new HttpHeaders();
-            char currentChar;
+            char currentChar = char.MinValue;
             int streamPosition = 0;
 
             // https://tools.ietf.org/html/rfc2616#section-4.2
@@ -62,14 +63,15 @@ namespace Datadog.Trace.HttpOverStreams
             var stringBuilder = new StringBuilder();
 
             var chArray = new byte[1];
-            void GoNextChar()
+
+            async Task GoNextChar()
             {
-                streamPosition++;
-                chArray[0] = (byte)responseStream.ReadByte();
+                await responseStream.ReadAsync(chArray, offset: 0, count: 1);
                 currentChar = Encoding.ASCII.GetChars(chArray)[0];
+                streamPosition++;
             }
 
-            bool IsNewLine()
+            async Task<bool> IsNewLine()
             {
                 if (currentChar.Equals(DatadogHttpValues.CarriageReturn))
                 {
@@ -77,7 +79,7 @@ namespace Datadog.Trace.HttpOverStreams
                     if (DatadogHttpValues.CrLfLength > 1)
                     {
                         // Skip the newline indicator
-                        GoNextChar();
+                        await GoNextChar();
                     }
 
                     return true;
@@ -89,13 +91,13 @@ namespace Datadog.Trace.HttpOverStreams
             // Skip to status code
             while (streamPosition < statusCodeStart)
             {
-                GoNextChar();
+                await GoNextChar();
             }
 
             // Read status code
             while (streamPosition < statusCodeEnd)
             {
-                GoNextChar();
+                await GoNextChar();
                 stringBuilder.Append(currentChar);
             }
 
@@ -110,14 +112,14 @@ namespace Datadog.Trace.HttpOverStreams
             // Skip to reason
             while (streamPosition < startOfReasonPhrase)
             {
-                GoNextChar();
+                await GoNextChar();
             }
 
             // Read reason
             do
             {
-                GoNextChar();
-                if (IsNewLine())
+                await GoNextChar();
+                if (await IsNewLine())
                 {
                     break;
                 }
@@ -132,10 +134,10 @@ namespace Datadog.Trace.HttpOverStreams
             // Read headers
             do
             {
-                GoNextChar();
+                await GoNextChar();
 
                 // Check for end of headers
-                if (IsNewLine())
+                if (await IsNewLine())
                 {
                     // Empty line, content starts next
                     break;
@@ -151,7 +153,7 @@ namespace Datadog.Trace.HttpOverStreams
                     }
 
                     stringBuilder.Append(currentChar);
-                    GoNextChar();
+                    await GoNextChar();
                 }
                 while (true);
 
@@ -161,9 +163,9 @@ namespace Datadog.Trace.HttpOverStreams
                 // Read value
                 do
                 {
-                    GoNextChar();
+                    await GoNextChar();
 
-                    if (IsNewLine())
+                    if (await IsNewLine())
                     {
                         // Next header pair starts
                         break;
--- a/src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs
+++ b/src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs
@@ -22,7 +22,7 @@ namespace Datadog.Trace.HttpOverStreams
         public async Task<HttpResponse> SendAsync(HttpRequest request, Stream requestStream, Stream responseStream)
         {
             await SendRequestAsync(request, requestStream).ConfigureAwait(false);
-            return await ReadResponseAsync(responseStream);
+            return await ReadResponseAsync(responseStream).ConfigureAwait(false);
         }
 
         private async Task SendRequestAsync(HttpRequest request, Stream requestStream)
@@ -66,7 +66,7 @@ namespace Datadog.Trace.HttpOverStreams
 
             async Task GoNextChar()
             {
-                await responseStream.ReadAsync(chArray, offset: 0, count: 1);
+                await responseStream.ReadAsync(chArray, offset: 0, count: 1).ConfigureAwait(false);
                 currentChar = Encoding.ASCII.GetChars(chArray)[0];
                 streamPosition++;
             }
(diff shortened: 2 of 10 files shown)


src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs
```diff
@@ -21,7 +22,7 @@ internal class DatadogHttpClient
         public async Task<HttpResponse> SendAsync(HttpRequest request, Stream requestStream, Stream responseStream)
         {
             await SendRequestAsync(request, requestStream).ConfigureAwait(false);
-            return ReadResponse(responseStream);
+            return await ReadResponseAsync(responseStream);
```
lucaspimentel, line 25: Shouldn't this `await` have `ConfigureAwait(false)`? Alternatively, can we return the task from `ReadResponseAsync()` without awaiting it?

src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs
```diff
@@ -62,23 +63,22 @@ private HttpResponse ReadResponse(Stream responseStream)
             var stringBuilder = new StringBuilder();
 
             var chArray = new byte[1];
-            void GoNextChar()
+
+            async Task GoNextChar()
             {
-                streamPosition++;
-                chArray[0] = (byte)responseStream.ReadByte();
+                await responseStream.ReadAsync(chArray, offset: 0, count: 1);
```
lucaspimentel, line 69: `ConfigureAwait(false)`? Here and every other new `await`.

src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs
```diff
@@ -62,23 +63,22 @@ private HttpResponse ReadResponse(Stream responseStream)
...
             {
                 if (currentChar.Equals(DatadogHttpValues.CarriageReturn))
                 {
                     // end of headers
-                    if (DatadogHttpValues.CrLfLength > 1)
-                    {
-                        // Skip the newline indicator
-                        GoNextChar();
-                    }
+                    // Next character should be a LineFeed, regardless of Linux/Windows
```
lucaspimentel, line 79: > regardless of Linux/Windows

Good catch

src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs
```diff
@@ -62,23 +63,22 @@ private HttpResponse ReadResponse(Stream responseStream)
...
+                    // Skip the newline indicator
```
lucaspimentel, line 80: Should we confirm that the next byte is actually a line feed? Is there a risk that a carriage returns appears in the stream without a line feed following it?

src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs
andrewlock: Yeah, you're right, no need to await

src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs
andrewlock: I'm too used to writing aspnetcore code where that doesn't matter any more 😳

src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs
andrewlock: This is the same behaviour as we were using on Windows previously, A CR in the header without a LF would be incorrect according to the HTTP spec AFACIT. I guess we should throw if we encounter that?

src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs
andrewlock: fixed in https://github.com/DataDog/dd-trace-dotnet/pull/1218/commits/1db429214309b593581ad47b1bffbe8b67d599ce

src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs
andrewlock: fixed in https://github.com/DataDog/dd-trace-dotnet/pull/1218/commits/1db429214309b593581ad47b1bffbe8b67d599ce

src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs
andrewlock: Updated in https://github.com/DataDog/dd-trace-dotnet/pull/1218/commits/bce14fb4452c93bc6866947d68dd25c5402e22b5 to throw if we don't get a LF

zacharycmontoya (approved):

src/Datadog.Trace/Agent/Transports/HttpStreamRequest.cs
```diff
@@ -43,17 +43,19 @@ public async Task<IApiResponse> PostAsync(ArraySegment<byte> traces)
...
+                if (!contentLength.HasValue)
                 {
-                    throw new Exception("Content length from http headers does not match content's actual length.");
+                    throw new Exception("Content-Length is required but was not provided");
                 }
 
+                // buffer the entire contents for now
+                var buffer = new byte[contentLength.Value];
+                var responseContentStream = new MemoryStream(buffer);
+                await response.Content.CopyToAsync(buffer).ConfigureAwait(false);
```
lucaspimentel, line 56: It seems like we could have handled both cases (copying from request content and from response content) with the `IHttpContent.CopyToAsync(Stream, ...)` overload. Here we could have passed the `MemoryStream` instead of a byte array as the destination. What do you think?

src/Datadog.Trace/Agent/Transports/HttpStreamRequest.cs
```diff
@@ -43,17 +43,19 @@ public async Task<IApiResponse> PostAsync(ArraySegment<byte> traces)
...
+                responseContentStream.Position = 0;
```
lucaspimentel, line 57: It seems like we could have handled both cases (copying from request content and from response content) with the `IHttpContent.CopyToAsync(Stream, ...)` overload. Here we could have passed the `MemoryStream` instead of a byte array as the destination. What do you think?

lucaspimentel: Something like this:

```csharp
                var responseContentStream = new MemoryStream(contentLength.Value);
                await response.Content.CopyToAsync(contentLength.Value, null).ConfigureAwait(false);
                responseContentStream.Position = 0;
```

... and then in `StreamContent`, move the logic to copy a specific number of bytes from `CopyToAsync(byte[])` to `CopyToAsync(Stream, ...)` (as it used to be originally).

andrewlock: >and then in StreamContent, we could move the logic that copies a specific number of bytes from `CopyToAsync(byte[])` to `CopyToAsync(Stream, ...)`.

Unfortunately, there's no API for copying a fixed number of bytes from one stream to another. We _could_ create another buffer in `CopyToAsync()` , so we copy bytes from response stream -> buffer -> parameter stream, but that seems like an extra copy for no real good reason?

After the changes in https://github.com/DataDog/dd-trace-dotnet/pull/1218/commits/eb26be51549c55856ae87a3d420116f1dfe696bc, we no longer ever use the `bufferSize` parameter so I've removed that in https://github.com/DataDog/dd-trace-dotnet/pull/1218/commits/8d8d1429ad5cc242f1b19352513ffb45da091876, but I think the `byte[]` approach is still the simplest here @lucaspimentel?

src/Datadog.Trace/HttpOverStreams/HttpContent/BufferContent.cs
```diff
@@ -15,9 +15,26 @@ public BufferContent(ArraySegment<byte> buffer)
...
+        public Task CopyToAsync(byte[] buffer)
+        {
+            if (_buffer.Count > buffer.Length)
+            {
+                throw new ArgumentOutOfRangeException(
+                    nameof(buffer),
+                    $"Buffer of size {buffer.Length} is not large enough to hold content of size {_buffer.Count}");
+            }
+
+            _buffer.Array?.CopyTo(buffer, _buffer.Offset);
```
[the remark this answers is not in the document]
andrewlock: Fixed the implementation in https://github.com/DataDog/dd-trace-dotnet/pull/1218/commits/454f3fe8c8747e7689fd697c1bc5e2f1f3a38b5e

src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs
```diff
@@ -89,13 +94,13 @@ bool IsNewLine()
...
+                await GoNextChar().ConfigureAwait(false);
             }
 
             // Read status code
             while (streamPosition < statusCodeEnd)
             {
-                GoNextChar();
+                await GoNextChar().ConfigureAwait(false);
                 stringBuilder.Append(currentChar);
             }
```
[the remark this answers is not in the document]
andrewlock: Optimised this as suggested in https://github.com/DataDog/dd-trace-dotnet/pull/1218/commits/03d5edfb87a1d97aa83358a29d2278033b626e68.

Also added `ReadUntil(StringBuilder, Char)` and `ReadUntilNewLine(StringBuilder)` in https://github.com/DataDog/dd-trace-dotnet/pull/1218/commits/10455f8de8564e60a8586b89655735f9e70610c6 to further reduce async allocations.

Added a unit test for response parsing to ensure I hadn't broken anything in https://github.com/DataDog/dd-trace-dotnet/pull/1218/commits/5040416e575bdbe5d11e211712303eaba359b40f

src/Datadog.Trace/Agent/Transports/HttpStreamRequest.cs
lucaspimentel: >but that seems like an extra copy for no real good reason

Yeah, good point. Thanks for looking into it, though.

src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs
```diff
@@ -57,27 +58,67 @@ private HttpResponse ReadResponse(Stream responseStream)
...
+            }
+
+            async Task SkipUntil(int requiredStreamPosition)
+            {
+                var advanceCount = requiredStreamPosition - streamPosition;
+                // Not required in release mode, as should only arise from programming error
+                System.Diagnostics.Debug.Assert(advanceCount > 0, "RequiredStreamPosition should be greater than 0");
+                System.Diagnostics.Debug.Assert(advanceCount <= bufferSize, "RequiredStreamPosition should be less than buffer size");
+
+                await responseStream.ReadAsync(chArray, offset: 0, count: advanceCount).ConfigureAwait(false);
```
lucaspimentel, line 82: `ReadAsync()` will return the number of bytes read, which can be less than the count requested. I think we may need to loop here until all the bytes we want are read.

src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs
```diff
@@ -57,27 +58,67 @@ private HttpResponse ReadResponse(Stream responseStream)
...
 
-            var chArray = new byte[1];
-            void GoNextChar()
+            var chArray = new byte[bufferSize];
+
+            async Task GoNextChar()
             {
-                streamPosition++;
-                chArray[0] = (byte)responseStream.ReadByte();
+                await responseStream.ReadAsync(chArray, offset: 0, count: 1).ConfigureAwait(false);
```
lucaspimentel, line 70: If we are at the end of the stream, `ReadAsync()` will return `0` and we will probably set `currentChar` to whatever previous value was in `chArray[0]`.

src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs
```diff
@@ -57,27 +58,67 @@ private HttpResponse ReadResponse(Stream responseStream)
...
+                System.Diagnostics.Debug.Assert(advanceCount <= bufferSize, "RequiredStreamPosition should be less than buffer size");
+
+                await responseStream.ReadAsync(chArray, offset: 0, count: advanceCount).ConfigureAwait(false);
+                currentChar = Encoding.ASCII.GetChars(chArray)[advanceCount - 1];
+                streamPosition += advanceCount;
+            }
+
+            async Task ReadUntil(StringBuilder builder, char stopChar)
+            {
+                while (!currentChar.Equals(stopChar))
```
lucaspimentel, line 89: What happens if we never find a `stopChar`?

src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs
```diff
@@ -87,15 +128,12 @@ bool IsNewLine()
...
-            {
-                GoNextChar();
-            }
+            await SkipUntil(statusCodeStart).ConfigureAwait(false);
 
             // Read status code
             while (streamPosition < statusCodeEnd)
             {
-                GoNextChar();
+                await GoNextChar().ConfigureAwait(false);
```
lucaspimentel, line 136: If we know that the status code is always 3 chars/bytes, should we read read 3 bytes from the stream instead of one at a time?

lucaspimentel (commented):
Left some concerns around reading bytes from a stream. I'm probably being paranoid.

src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs
andrewlock: Good catch 👍

src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs
andrewlock: Yeah, that's the same as the previous behaviour I think - I guess we can probably throw if we expect to get bytes but don't get any?

src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs
andrewlock: For a network stream, we either hang, or the connection is closed when we read over the end, same as the previous implementation 😉

src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs
andrewlock: I don't think it's worth optimising this specific case now. We should have a proper buffered reader at some point, which would handle the general case

lucaspimentel: Agreed. Resolving this.

src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs
andrewlock: Fixed in https://github.com/DataDog/dd-trace-dotnet/pull/1218/commits/656a470756c5926b3d68df33e3bfa4313e44babc. This throws is we unexpectedly reach the end of the stream. This won't actually occur with a NetworkStream, as either it will spin waiting for new data, or the connection will be closed and throw, but makes more sense logically.

src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs
andrewlock: Fixed in https://github.com/DataDog/dd-trace-dotnet/pull/1218/commits/656a470756c5926b3d68df33e3bfa4313e44babc. Throws if we unexpectedly reach the end of the stream (won't happen with NetworkStream, as that will either hang or throw)

lucaspimentel (approved):

colin-higgins (approved):

src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs
```diff
@@ -57,27 +58,83 @@ private HttpResponse ReadResponse(Stream responseStream)
...
+                while (totalBytesRead < advanceCount)
+                {
+                    var bytesRead = await responseStream.ReadAsync(chArray, offset: 0, count: advanceCount).ConfigureAwait(false);
+                    if (bytesRead == 0)
+                    {
+                        throw new InvalidOperationException($"Unexpected end of stream at position {streamPosition}");
+                    }
+
+                    totalBytesRead += bytesRead;
+                }
```
kevingosse, line 97: I think the `advanceCount <= bufferSize` check should also be there in release (especially since we run the tests in release mode now )
Though you shouldn't even need it. If you make sure that you read `Math.Min(advanceCount, bufferSize)`, your code should work as-is with any value of `advanceCount`
(it would just require an extra bit of logic to get the position of the `currentChar` after the reading loop)

src/Datadog.Trace/HttpOverStreams/HttpContent/StreamContent.cs
```diff
@@ -15,14 +16,38 @@ public StreamContent(Stream stream, long? length)
...
+
+            if (Length > buffer.Length)
             {
-                return Stream.CopyToAsync(destination);
+                throw new ArgumentException($"Provided buffer was smaller {buffer.Length} than the content length {Length}");
             }
 
-            return Stream.CopyToAsync(destination, bufferSize.Value);
+            var length = 0;
+            var remaining = Length.Value;
```
kevingosse, line 37: Given that part of the logic comes from `Length.Value` being a long, I think you should use explicit typing for this line

```suggestion
            long remaining = Length.Value;
```

src/Datadog.Trace/HttpOverStreams/DatadogHttpClient.cs
andrewlock: fixed in https://github.com/DataDog/dd-trace-dotnet/pull/1218/commits/c8b9639fdd7acf8ab8c0c774736c8cc661a1bbee

> (it would just require an extra bit of logic to get the position of the currentChar after the reading loop)

This should have been in there anyway, to handle the case (where fewer than `advanceCount` were returned in the stream)
so you inadvertently fixed a bug too :smile:

PR #1220, opened 2021-02-17
Replacing DuckType `As<T>` with DuckCast, TryDuckCast, DuckAs and DuckIs

tonyredondo:
Replacing DuckType `As<T>` with DuckCast, TryDuckCast, DuckAs and DuckIs

https://github.com/DataDog/dd-trace-dotnet/blob/tony/ducktyping-duckcast/src/Datadog.Trace/DuckTyping/DuckTypeExtensions.cs#L14-L133

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace/DuckTyping/DuckTypeExtensions.cs
+++ b/src/Datadog.Trace/DuckTyping/DuckTypeExtensions.cs
@@ -38,7 +38,7 @@ namespace Datadog.Trace.DuckTyping
         /// <param name="instance">Object instance</param>
         /// <param name="value">Ducktype instance</param>
         /// <returns>true if the object instance was ducktyped; otherwise, false.</returns>
-        public static bool DuckIs<T>(this object instance, out T value)
+        public static bool TryDuckCast<T>(this object instance, out T value)
         {
             try
             {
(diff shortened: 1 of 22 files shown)


src/Datadog.Trace/DuckTyping/DuckTypeExtensions.cs
```diff
@@ -22,7 +27,109 @@ public static T As<T>(this object instance)
...
+        /// <param name="instance">Object instance</param>
+        /// <param name="value">Ducktype instance</param>
+        /// <returns>true if the object instance was ducktyped; otherwise, false.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool TryDuckCast<T>(this object instance, out T value)
+        {
+            try
+            {
+                value = DuckType.Create<T>(instance);
+                return true;
```
lucaspimentel, line 47: Should we call `DuckType.CanCreate<T>(instance)` first to avoid throwing if we know it won't work? I'm not sure what the performance implications are for the common case (when we can create). Even if we do this, we probably want to keep the try/catch because `TryFoo()` methods should never throw.
```suggestion
                if(DuckType.CanCreate<T>(instance))
                {
                    value = DuckType.Create<T>(instance);
                    return true;
                }

                value = default;
                return false;
```

tonyredondo: I'd prefer to avoid calling `CanCreate` for performance reasons, that's why in this case is like a best effort approach. Also as you said, we must keep the try/catch so I don't think it worth it.

lucaspimentel: I see, thanks for clarifying!

lucaspimentel (approved):
Thanks!

zacharycmontoya (approved):

PR #1221, opened 2021-02-18
CIApp: Traits as a Dictionary<string, string[]>

tonyredondo:
Change `test.traits` format following the CiApp specs.

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/NUnit/NUnitIntegration.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/NUnit/NUnitIntegration.cs
@@ -82,6 +82,7 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Testing.NUnit
             // Get traits
             if (testMethodProperties != null)
             {
+                Dictionary<string, List<string>> traits = new Dictionary<string, List<string>>();
                 skipReason = (string)testMethodProperties.Get("_SKIPREASON");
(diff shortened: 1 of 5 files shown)


src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/NUnit/NUnitIntegration.cs
```diff
@@ -75,6 +75,7 @@ internal static Scope CreateScope<TContext>(TContext executionContext, Type targ
             // Get traits
             if (testMethodProperties != null)
             {
+                Dictionary<string, List<string>> traits = new Dictionary<string, List<string>>();
```
kevingosse, line 78: We really should use var in those cases 😄

kevingosse (approved):

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/NUnit/NUnitIntegration.cs
tonyredondo: oh, not again... the `var` discussion...

andrewlock: If you feel like a different argument, you could use a [new expression](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/new-operator) instead :trollface:

lucaspimentel (approved):

PR #1222, opened 2021-02-18
Use release artifacts in package.sh

kevingosse:


andrewlock (approved):
Who knows what it does 🤷‍♂️ But it _looks_ right 😂

tonyredondo (approved):

PR #1184, opened 2021-02-02
CallTarget Integration: ADO.NET MySqlData client version 6

tonyredondo:
Add MySqlData client version 6 support and execute the tests in multiple client versions.

@DataDog/apm-dotnet

PR #1225, opened 2021-02-19
Strengthen type check in the method resolution for CallSite instrumentation and improve fallback mechanism

zacharycmontoya:
### Issue
It is possible to load the same assembly multiple times in an application by loading it into different assembly load contexts. One example is in .NET Core where an assembly may be loaded into the Default load context and then again into its own Assembly Load Context. When this happens, the Types from one assembly are not considered equal to the corresponding Type in the other assembly, even though the information that we can publicly observe (e.g. MVID and metadata tokens) appear to be the same. This causes a crash condition in our automatic instrumentation when the following happens:

1. Assembly1 (e.g. `System.Data.SqlClient.dll`) is loaded through the default load context
2. Our `ModuleLookup.PopulateModules` method caches a key-value pair of Assembly1.MVID and Assembly1.Module
3. The same Assembly1, let's call it Assembly1~New, is loaded into a new ALC
4. Automatic instrumentation is triggered for Assembly1\~New. We retrieve the MethodInfo by getting the cached Assembly1.Module (NOT equivalent to Assembly1\~New.Module), get a metadata token from it, and construct a dynamic method that calls into the original method
5. Invoking the original method results in a type cast exception on the instance object

### Fix
One potential solution that was considered but rejected was improving the module caching to consider multiple versions of the same assembly. The only way to differentiate between the assemblies would be to compare Assembly objects or the IntPtr to the Assembly, which seemed too complex to me.

The solution pursued is to detect this scenario by adding an additional type check after we get the target MethodInfo and check that the instance object can be assigned to the Type containing the MethodInfo. Failing this check forces the method resolution to use the reflection fallback to find the right method.

This fix resolves the issue, but required a cascading set of changes to perform stronger name checks on return types when method overloads exist. Once this was done, further testing demonstrated the fallback mechanism worked for all of our integrations (minus RabbitMQ).

### Changes
- When resolving the original MethodInfo via metadata token, add a check to ensure the object instance can correctly be assigned to the MethodInfo type
- Enforce a stricter type check on `Task<>` return types to remove ambiguity in fallback method resolution. Modify all CallSite integrations that return `Task<>` to be explicit about the generic type
- Add failure mode to all ADO.NET test-application programs, which will run in integration-tests CI pipeline

### Remaining Work
After testing all affected CallSite integrations except `AspNetWebApi2` and `XUnit`, the only integration whose fallback method resolution fails is `RabbitMQ`. That may require the integration point to be modified, so the work should be deferred to a separate work item.

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace.ClrProfiler.Managed/Emit/MethodBuilder.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/Emit/MethodBuilder.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using System.Reflection.Emit;
+using System.Threading.Tasks;
 using Datadog.Trace.ClrProfiler.Helpers;
 using Datadog.Trace.Configuration;
 using Datadog.Trace.DogStatsd;
@@ -421,6 +422,12 @@ namespace Datadog.Trace.ClrProfiler.Emit
                 return null;
             }
 
+            if (!methodInfo.IsStatic && !methodInfo.ReflectedType.IsAssignableFrom(_concreteType))
+            {
+                Log.Warning($"_concreteType cannot be assigned to the type containing the MethodInfo representing the instance method: {detailMessage}");
+                return null;
+            }
+
             return methodInfo;
         }
 
@@ -511,7 +518,27 @@ namespace Datadog.Trace.ClrProfiler.Emit
                             continue;
                         }
 
-                        if ($"{typesToCheck[i].Namespace}.{typesToCheck[i].Name}" != _namespaceAndNameFilter[i])
+                        // If return type is Task<>, perform a stronger check to ensure we find the correct generic type
+                        if (i == 0 && m.ReturnType.IsGenericType && typeof(Task).IsAssignableFrom(m.ReturnType))
+                        {
+                            if (_namespaceAndNameFilter[i] == ClrNames.IgnoreGenericTask)
+                            {
+                                // Allow for not specifying
+                                continue;
+                            }
+
+                            // If the type argument of Task is T (some generic), do this
+                            var typeArgument = m.ReturnType.GenericTypeArguments[0];
+                            if (typeArgument.IsGenericParameter && ClrNames.GenericParameterTask != _namespaceAndNameFilter[i])
+                            {
+                                return false;
+                            }
+                            else if (!typeArgument.IsGenericParameter && $"{ClrNames.GenericTask}<{typeArgument.Namespace}.{typeArgument.Name}>" != _namespaceAndNameFilter[i])
+                            {
+                                return false;
+                            }
+                        }
+                        else if ($"{typesToCheck[i].Namespace}.{typesToCheck[i].Name}" != _namespaceAndNameFilter[i])
                         {
                             return false;
                         }
--- a/src/Datadog.Trace.ClrProfiler.Managed/Emit/MethodBuilder.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/Emit/MethodBuilder.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using System.Reflection.Emit;
+using System.Threading.Tasks;
 using Datadog.Trace.ClrProfiler.Helpers;
 using Datadog.Trace.Configuration;
 using Datadog.Trace.DogStatsd;
@@ -424,8 +425,7 @@ namespace Datadog.Trace.ClrProfiler.Emit
             if (!methodInfo.IsStatic && !methodInfo.ReflectedType.IsAssignableFrom(_concreteType))
             {
                 Log.Warning($"_concreteType cannot be assigned to the type containing the MethodInfo representing the instance method: {detailMessage}");
-                throw new Exception($"_concreteType cannot be assigned to the type containing the MethodInfo representing the instance method: {detailMessage}");
-                // return null;
+                return null;
             }
 
             return methodInfo;
@@ -478,9 +478,6 @@ namespace Datadog.Trace.ClrProfiler.Emit
         {
             var logDetail = $"mdToken {_mdToken} on {_concreteTypeName}.{_methodName} in {_resolutionModule?.FullyQualifiedName ?? "NULL"}, {_resolutionModule?.ModuleVersionId ?? _moduleVersionId}";
             Log.Warning($"Using fallback method matching ({logDetail})");
-            throw new Exception($"Using fallback method matching ({logDetail})");
-
-            /*
 
             var statsd = Tracer.Instance.Statsd;
 
@@ -575,7 +572,6 @@ namespace Datadog.Trace.ClrProfiler.Emit
             }
 
             return methodInfo;
-            */
         }
 
         private bool ParametersAreViable(MethodInfo mi)
--- a/src/Datadog.Trace.ClrProfiler.Managed/Emit/MethodBuilder.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/Emit/MethodBuilder.cs
@@ -478,6 +478,9 @@ namespace Datadog.Trace.ClrProfiler.Emit
         {
             var logDetail = $"mdToken {_mdToken} on {_concreteTypeName}.{_methodName} in {_resolutionModule?.FullyQualifiedName ?? "NULL"}, {_resolutionModule?.ModuleVersionId ?? _moduleVersionId}";
             Log.Warning($"Using fallback method matching ({logDetail})");
+            throw new Exception($"Using fallback method matching ({logDetail})");
+
+            /*
 
             var statsd = Tracer.Instance.Statsd;
 
@@ -572,6 +575,7 @@ namespace Datadog.Trace.ClrProfiler.Emit
             }
 
             return methodInfo;
+            */
         }
 
         private bool ParametersAreViable(MethodInfo mi)
--- a/src/Datadog.Trace.ClrProfiler.Managed/Emit/MethodBuilder.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/Emit/MethodBuilder.cs
@@ -421,10 +421,10 @@ namespace Datadog.Trace.ClrProfiler.Emit
                 return null;
             }
 
-            if (!methodInfo.IsStatic && methodInfo.ReflectedType != _concreteType)
+            if (!methodInfo.IsStatic && !methodInfo.ReflectedType.IsAssignableFrom(_concreteType))
             {
-                Log.Warning($"MethodInfo for instance method defined on type that does not match the input instance: {detailMessage}");
-                throw new Exception($"MethodInfo for instance method defined on type that does not match the input instance: {detailMessage}");
+                Log.Warning($"_concreteType cannot be assigned to the type containing the MethodInfo representing the instance method: {detailMessage}");
+                throw new Exception($"_concreteType cannot be assigned to the type containing the MethodInfo representing the instance method: {detailMessage}");
                 // return null;
             }
 
--- a/src/Datadog.Trace.ClrProfiler.Managed/Emit/MethodBuilder.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/Emit/MethodBuilder.cs
@@ -421,6 +421,13 @@ namespace Datadog.Trace.ClrProfiler.Emit
                 return null;
             }
 
+            if (!methodInfo.IsStatic && methodInfo.ReflectedType != _concreteType)
+            {
+                Log.Warning($"MethodInfo for instance method defined on type that does not match the input instance: {detailMessage}");
+                throw new Exception($"MethodInfo for instance method defined on type that does not match the input instance: {detailMessage}");
+                // return null;
+            }
+
             return methodInfo;
         }
 
--- a/src/Datadog.Trace.ClrProfiler.Managed/Emit/MethodBuilder.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/Emit/MethodBuilder.cs
@@ -4,7 +4,6 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using System.Reflection.Emit;
-using System.Threading.Tasks;
 using Datadog.Trace.ClrProfiler.Helpers;
 using Datadog.Trace.Configuration;
 using Datadog.Trace.DogStatsd;
@@ -419,12 +418,6 @@ namespace Datadog.Trace.ClrProfiler.Emit
             if (!ParametersAreViable(methodInfo))
             {
                 Log.Warning($"Parameters not viable: {detailMessage}");
-                return null;
-            }
-
-            if (!methodInfo.IsStatic && !methodInfo.ReflectedType.IsAssignableFrom(_concreteType))
-            {
-                Log.Warning($"_concreteType cannot be assigned to the type containing the MethodInfo representing the instance method: {detailMessage}");
                 return null;
             }
 
--- a/src/Datadog.Trace.ClrProfiler.Managed/Integrations/AdoNet/SqlCommandIntegration.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/Integrations/AdoNet/SqlCommandIntegration.cs
@@ -343,7 +343,7 @@ namespace Datadog.Trace.ClrProfiler.Integrations.AdoNet
                        .Start(moduleVersionPtr, mdToken, opCode, methodName)
                        .WithConcreteType(sqlCommandType)
                        .WithParameters(commandBehavior, cancellationToken)
-                       .WithNamespaceAndNameFilters(ClrNames.GenericTask, AdoNetConstants.TypeNames.CommandBehavior, ClrNames.CancellationToken)
+                       .WithNamespaceAndNameFilters($"{ClrNames.GenericTask}<{sqlDataReaderType.FullName}>", AdoNetConstants.TypeNames.CommandBehavior, ClrNames.CancellationToken)
                        .Build();
             }
             catch (Exception ex)
@@ -550,7 +550,7 @@ namespace Datadog.Trace.ClrProfiler.Integrations.AdoNet
             TargetAssemblies = new[] { MicrosoftSqlClientAssemblyName },
             TargetType = MicrosoftSqlCommandTypeName,
             TargetMethod = AdoNetConstants.MethodNames.ExecuteNonQueryAsync,
-            TargetSignatureTypes = new[] { "System.Threading.Tasks.Task`1<System.Int32>", ClrNames.CancellationToken },
+            TargetSignatureTypes = new[] { ClrNames.Int32Task, ClrNames.CancellationToken },
             TargetMinimumVersion = Major1,
             TargetMaximumVersion = Major2)]
         public static object MicrosoftSqlClientExecuteNonQueryAsync(
@@ -589,7 +589,7 @@ namespace Datadog.Trace.ClrProfiler.Integrations.AdoNet
                        .Start(moduleVersionPtr, mdToken, opCode, methodName)
                        .WithConcreteType(targetType)
                        .WithParameters(cancellationToken)
-                       .WithNamespaceAndNameFilters(ClrNames.GenericTask, ClrNames.CancellationToken)
+                       .WithNamespaceAndNameFilters(ClrNames.Int32Task, ClrNames.CancellationToken)
                        .Build();
             }
             catch (Exception ex)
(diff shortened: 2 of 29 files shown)


src/Datadog.Trace.ClrProfiler.Managed/Emit/MethodBuilder.cs
```diff
@@ -421,6 +422,12 @@ private MethodInfo VerifyMethodFromToken(MethodInfo methodInfo)
                 return null;
             }
 
+            if (!methodInfo.IsStatic && !methodInfo.ReflectedType.IsAssignableFrom(_concreteType))
+            {
+                Log.Warning($"_concreteType cannot be assigned to the type containing the MethodInfo representing the instance method: {detailMessage}");
```
andrewlock, line 427: nit: we should probably use structure logging here instead of interpolated strings. It can reduce GC pressure generally and Serilog is optimised for it. There's other log messages in this file that could do with the same treatment 🙂

test/Datadog.Trace.ClrProfiler.IntegrationTests/AdoNet/MicrosoftDataSqlClientTests.cs
```diff
@@ -42,14 +42,15 @@ public void SubmitsTracesWithNetStandard(string packageVersion, bool enableCallT
             // Once this is fully supported, this will add another 2 complete groups for all frameworks instead
             // of 4 extra spans on net461 and netcoreapp2.0+
 #if NET461
-            var expectedSpanCount = 77; // 7 queries * 11 groups
+            var expectedSpanCount = 133; // 7 queries * 11 groups
```
andrewlock, line 45: If this is the new number, we should prob update the comment

test/Datadog.Trace.ClrProfiler.IntegrationTests/AdoNet/MicrosoftDataSqlClientTests.cs
```diff
@@ -42,14 +42,15 @@ public void SubmitsTracesWithNetStandard(string packageVersion, bool enableCallT
...
 #else
-            var expectedSpanCount = 81; // 7 queries * 11 groups + 4 spans from generic wrapper on .NET Core
+            // Update to final value once tests pass
+            var expectedSpanCount = 137; // 7 queries * 11 groups + 4 spans from generic wrapper on .NET Core
```
andrewlock, line 48: I guess this was the right number? Should probably update the comments

test/Datadog.Trace.ClrProfiler.IntegrationTests/AdoNet/MySqlCommandTests.cs
```diff
@@ -100,21 +100,26 @@ private void SubmitsTracesWithNetStandard(string packageVersion, bool enableCall
...
 #if NET452
-            var expectedSpanCount = 50; // 7 queries * 7 groups + 1 internal query
-#else
             var expectedSpanCount = 78; // 7 queries * 11 groups + 1 internal query
+#elif NET461
+            var expectedSpanCount = 134;
+#else
+            var expectedSpanCount = 137; // 7 queries * 19 groups + 1 internal query
+#endif
+
```
andrewlock, line 109: Update the comments to reflect the numbers?

test/Datadog.Trace.ClrProfiler.IntegrationTests/AdoNet/NpgsqlCommandTests.cs
```diff
@@ -39,17 +39,22 @@ public void SubmitsTracesWithNetStandard(string packageVersion, bool enableCallT
...
             // Once this is fully supported, this will add another 2 complete groups instead.
 #if NET452
-            var expectedSpanCount = 50; // 7 queries * 7 groups + 1 internal query
-#else
             var expectedSpanCount = 78; // 7 queries * 11 groups + 1 internal query
+#elif NET461
+            var expectedSpanCount = 134; // 7 queries * 11 groups + 1 internal query
+#else
+            // Update to final value once tests pass
+            var expectedSpanCount = 135; // 7 queries * 11 groups + 1 internal query
```
andrewlock, line 47: As above RE comments

test/Datadog.Trace.ClrProfiler.IntegrationTests/AdoNet/SystemDataSqlClientTests.cs
```diff
@@ -42,19 +42,20 @@ public void SubmitsTracesWithNetStandard(string packageVersion, bool enableCallT
...
 #if NET452
-            var expectedSpanCount = 70; // 7 queries * 10 groups
-#elif NET461
             var expectedSpanCount = 98; // 7 queries * 14 groups
+#elif NET461
+            var expectedSpanCount = 154; // 7 queries * 14 groups
 #else
-            var expectedSpanCount = 102; // 7 queries * 14 groups + 4 spans from generic wrapper on .NET Core
+            // Update to final value once tests pass
+            var expectedSpanCount = 158; // 7 queries * 14 groups + 4 spans from generic wrapper on .NET Core
```
andrewlock, line 50: As above RE comments

andrewlock (approved):
LGTM, just v. minor comments

One question - what's the expected performance impact of not caching the result when we have to fallback? How much is that code path hit? (I don't understand the nitty-gritty, so that may not even be a valid question!)

kevingosse:
> One potential solution that was considered but rejected was improving the module caching to consider multiple versions of the same assembly. The only way to differentiate between the assemblies would be to compare Assembly objects or the IntPtr to the Assembly, which seemed too complex to me.

I don't think it's that complex? As far as I know, assemblies are singletons, so it's really just a matter of adding an assembly reference check in the key of the MethodBuilder.

Tested on .NET Framework:

```csharp
            void SomeMethod(object instance)
            {
                var class1 = (Class1)instance;
            }

            var ass1 = typeof(Class1).Assembly;
            var ass2 = AppDomain.CurrentDomain.GetAssemblies().First(a => a.FullName.Contains("Library1"));
            var ass3 = Assembly.LoadFrom("Library1.dll");
            var ass4 = Assembly.Load(File.ReadAllBytes("Library1.dll"));

            Console.WriteLine(ass1.Modules.First().ModuleVersionId); // eadc2f50-c25c-44cd-9095-a720340a9d87
            Console.WriteLine(ass2.Modules.First().ModuleVersionId); // eadc2f50-c25c-44cd-9095-a720340a9d87
            Console.WriteLine(ass3.Modules.First().ModuleVersionId); // eadc2f50-c25c-44cd-9095-a720340a9d87
            Console.WriteLine(ass4.Modules.First().ModuleVersionId); // eadc2f50-c25c-44cd-9095-a720340a9d87

            Console.WriteLine(object.ReferenceEquals(ass1, ass2)); // True
            Console.WriteLine(object.ReferenceEquals(ass1, ass3)); // True
            Console.WriteLine(object.ReferenceEquals(ass1, ass4)); // False

            var obj1 = ass1.CreateInstance("Library1.Class1");
            var obj2 = ass2.CreateInstance("Library1.Class1");
            var obj3 = ass3.CreateInstance("Library1.Class1");
            var obj4 = ass4.CreateInstance("Library1.Class1");

            SomeMethod(obj1); // OK
            SomeMethod(obj2); // OK
            SomeMethod(obj3); // OK
            SomeMethod(obj4); // InvalidCastException
```

kevingosse:
For completeness, on .NET Core (with AssemblyLoadContext) : 

```csharp
            void SomeMethod(object instance)
            {
                try
                {
                    var class1 = (Class1) instance;
                    Console.WriteLine("OK");
                }
                catch (Exception ex)
                {
                    Console.WriteLine(ex.GetType().Name);
                }
            }

            var loadContext1 = new AssemblyLoadContext1();
            var loadContext2 = new AssemblyLoadContext2();

            var ass1 = typeof(Class1).Assembly;
            var ass2 = AppDomain.CurrentDomain.GetAssemblies().First(a => a.FullName.Contains("ClassLibrary1"));
            var ass3 = Assembly.LoadFrom("ClassLibrary1.dll");
            var ass4 = Assembly.Load(File.ReadAllBytes("ClassLibrary1.dll"));

            var ass5 = loadContext1.LoadFromAssemblyPath(Path.Combine(Environment.CurrentDirectory, "ClassLibrary1.dll"));
            var ass6 = loadContext2.LoadFromAssemblyPath(Path.Combine(Environment.CurrentDirectory, "ClassLibrary1.dll"));

            Console.WriteLine(ass1.Modules.First().ModuleVersionId); // b622394a-556f-4de5-a4d8-11397d978585
            Console.WriteLine(ass2.Modules.First().ModuleVersionId); // b622394a-556f-4de5-a4d8-11397d978585
            Console.WriteLine(ass3.Modules.First().ModuleVersionId); // b622394a-556f-4de5-a4d8-11397d978585
            Console.WriteLine(ass4.Modules.First().ModuleVersionId); // b622394a-556f-4de5-a4d8-11397d978585
            Console.WriteLine(ass5.Modules.First().ModuleVersionId); // b622394a-556f-4de5-a4d8-11397d978585
            Console.WriteLine(ass6.Modules.First().ModuleVersionId); // b622394a-556f-4de5-a4d8-11397d978585

            Console.WriteLine(object.ReferenceEquals(ass1, ass2)); // True
            Console.WriteLine(object.ReferenceEquals(ass1, ass3)); // True
            Console.WriteLine(object.ReferenceEquals(ass1, ass4)); // False
            Console.WriteLine(object.ReferenceEquals(ass1, ass5)); // False
            Console.WriteLine(object.ReferenceEquals(ass1, ass6)); // False
            Console.WriteLine(object.ReferenceEquals(ass5, ass6)); // False

            var obj1 = ass1.CreateInstance("ClassLibrary1.Class1");
            var obj2 = ass2.CreateInstance("ClassLibrary1.Class1");
            var obj3 = ass3.CreateInstance("ClassLibrary1.Class1");
            var obj4 = ass4.CreateInstance("ClassLibrary1.Class1");
            var obj5 = ass5.CreateInstance("ClassLibrary1.Class1");
            var obj6 = ass6.CreateInstance("ClassLibrary1.Class1");

            SomeMethod(obj1); // OK
            SomeMethod(obj2); // OK
            SomeMethod(obj3); // OK
            SomeMethod(obj4); // InvalidCastException
            SomeMethod(obj5); // InvalidCastException
            SomeMethod(obj6); // InvalidCastException
```

zacharycmontoya:
> I don't think it's that complex? As far as I know, assemblies are singletons, so it's really just a matter of adding an assembly reference check to the key of the MethodBuilder.
You're right, a multi-part key would do the trick fairly easily. I will go ahead and implement this. For some reason I thought the only way to do this would be to pass some more info from the profiler's JITCompilationStarted side which is why I had abandoned that thought.

zacharycmontoya:
> > I don't think it's that complex? As far as I know, assemblies are singletons, so it's really just a matter of adding an assembly reference check to the key of the MethodBuilder.
> 
> You're right, a multi-part key would do the trick fairly easily. I will go ahead and implement this. For some reason I thought the only way to do this would be to pass some more info from the profiler's JITCompilationStarted side which is why I had abandoned that thought.

@kevingosse After some further research and testing, I would need to get the Assembly of the caller method to make this work. We don't have this information today. In our current instrumentation, we only pass the MVID of the caller method to the integration method via the IL rewriting. If we figured out a safe way to also pass a pointer to the calling method's Assembly or some sort of AssemblyID, this could work. Otherwise, we'd have to create a StackFrame inside our integration method to determine what method just called our CallSite instrumentation, which might be too expensive since this work happens once per integration method, per callsite, per underlying instance type. Do you think I should work that out in this PR or can we keep track of this for future improvements?

zacharycmontoya:
> LGTM, just v. minor comments
> 
> One question - what's the expected performance impact of not caching the result when we have to fallback? How much is that code path hit? (I don't understand the nitty-gritty, so that may not even be a valid question!)

No worries, this is definitely a valid question! This fallback method resolution (that occurs when we rewrite the call sites to do automatic instrumentation) doesn't affect any caching of results. Since we currently do CallSite instrumentation by jumping in the middle of the caller and library method, we have to determine which method was originally intended to be invoked. Our preferred path is to get the module that was originally making the library call and just do a metadata token lookup. Since we originally created a 1:1 mapping between MVID (unique per assembly file, but not assembly instance) and module, this fails in this scenario. As a fallback, we use reflection to look up the methods on the object instance and find the correct method that way. So the new behavior is that, when this scenario arises with a problematic object instance we will invoke the additional bit of reflection logic, but this doesn't change how many times we enter this method resolution logic in general or invalidate any other cache lookups. Does that answer your question?

test/Datadog.Trace.ClrProfiler.IntegrationTests/AdoNet/MicrosoftDataSqlClientTests.cs
zacharycmontoya: I put off updating these 😅 Done in f9829db87629aa5c357cc4ad5a69a51fcd1da1ad

test/Datadog.Trace.ClrProfiler.IntegrationTests/AdoNet/MicrosoftDataSqlClientTests.cs
zacharycmontoya: I put off updating these 😅 Done in f9829db87629aa5c357cc4ad5a69a51fcd1da1ad

test/Datadog.Trace.ClrProfiler.IntegrationTests/AdoNet/MySqlCommandTests.cs
zacharycmontoya: I put off updating these 😅 Done in f9829db87629aa5c357cc4ad5a69a51fcd1da1ad

test/Datadog.Trace.ClrProfiler.IntegrationTests/AdoNet/NpgsqlCommandTests.cs
zacharycmontoya: I put off updating these 😅 Done in f9829db87629aa5c357cc4ad5a69a51fcd1da1ad

test/Datadog.Trace.ClrProfiler.IntegrationTests/AdoNet/SystemDataSqlClientTests.cs
zacharycmontoya: I put off updating these 😅 Done in f9829db87629aa5c357cc4ad5a69a51fcd1da1ad

src/Datadog.Trace.ClrProfiler.Managed/Emit/MethodBuilder.cs
zacharycmontoya: Do you mind if I make the logging changes in a separate PR? 😄

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AdoNet/SqlCommandIntegration.cs
```diff
@@ -343,7 +343,7 @@ public static class SqlCommandIntegration
                        .Start(moduleVersionPtr, mdToken, opCode, methodName)
                        .WithConcreteType(sqlCommandType)
                        .WithParameters(commandBehavior, cancellationToken)
-                       .WithNamespaceAndNameFilters(ClrNames.GenericTask, AdoNetConstants.TypeNames.CommandBehavior, ClrNames.CancellationToken)
+                       .WithNamespaceAndNameFilters($"{ClrNames.GenericTask}<{sqlDataReaderType.FullName}>", AdoNetConstants.TypeNames.CommandBehavior, ClrNames.CancellationToken)
```
lucaspimentel, line 346: From looking at its source code, `Type.FullName` is not trivial. Instead of using `sqlDataReaderType.FullName`, consider reusing the result of `$"{sqlClientNamespace}.{SqlDataReaderTypeName}"` from line 327 above.

src/Datadog.Trace.ClrProfiler.Managed/Integrations/HttpMessageHandlerIntegration.cs
```diff
@@ -187,7 +187,7 @@ public static class HttpMessageHandlerIntegration
                        .Start(moduleVersionPtr, mdToken, opCode, Send)
                        .WithConcreteType(httpMessageHandler)
                        .WithParameters(request, cancellationToken)
-                       .WithNamespaceAndNameFilters(NamespaceAndNameFilters)
+                       .WithNamespaceAndNameFilters(ClrNames.HttpResponseMessage, ClrNames.HttpRequestMessage, ClrNames.CancellationToken)
```
lucaspimentel, line 190: Doesn't this return a `Task<HttpResponseMessage>`?

test/test-applications/integrations/Samples.MySql/Program.cs
```diff
@@ -42,7 +53,7 @@ private static MySqlConnection OpenConnection()
                 if (oldMySqlServer)
                 {
                     var host = Environment.GetEnvironmentVariable("MYSQL57_HOST") ?? "localhost";
-                    var port = Environment.GetEnvironmentVariable("MYSQL57_PORT") ?? "3307";
+                    var port = Environment.GetEnvironmentVariable("MYSQL57_PORT") ?? "3407";
```
lucaspimentel, line 56: Was this the wrong port?

zacharycmontoya: Yes, by default in our `docker-compose.yml`, we expose the `mysql57` service on localhost port 3407

src/Datadog.Trace.ClrProfiler.Managed/Integrations/HttpMessageHandlerIntegration.cs
zacharycmontoya: No, this is the new synchronous Send API in .NET 5 😄 I specifically had to change this because `NamespaceAndNameFilters` was incorrectly setting a return type of `Task<HttpResponseMessage>`

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AdoNet/SqlCommandIntegration.cs
zacharycmontoya: Good idea, done in 70644ebb551fd0b8a466e760dd73af3df4a8ff10

test/test-applications/integrations/Samples.MySql/Program.cs
lucaspimentel: Was this working before then? Maybe we don't need the two MySQL instances?

zacharycmontoya: This was relatively new so I don't think the fallback was working before. I'm not sure why we have two different MySql instances, @tonyredondo would know

src/Datadog.Trace.ClrProfiler.Managed/Integrations/HttpMessageHandlerIntegration.cs
lucaspimentel: My bad. Something in the github UI maybe we thing this was inside `HttpMessageHandler_SendAsync()`. Thanks!

test/test-applications/integrations/Samples.MySql/Program.cs
zacharycmontoya: There was this, but I haven't looked into it any further: https://github.com/DataDog/dd-trace-dotnet/pull/1184#discussion_r577043869

lucaspimentel: Ok, thanks for clarifying. I'll let it go (_for now!_)

tonyredondo: This is because MySql 8 is a totally different database from MySql 5.7, they changed a lot (and skipped some numbers in between).

https://severalnines.com/database-blog/moving-mysql-57-mysql-80-what-you-should-know

One of the changes is the authorization and connection mechanism, so client from different versions are not compatibles by default (maybe with a lot of configuration changes can work), so it was simpler to expose two instances from the server side.

tonyredondo: The port change must be a left over from the local test, integration tests doesn't fail because uses the [env-vars](https://github.com/DataDog/dd-trace-dotnet/blob/zach/fix/master-adonet-fix-main-dev/docker-compose.yml#L458-L459)

src/Datadog.Trace.ClrProfiler.Managed/Emit/MethodBuilder.cs
andrewlock: Haha, of course!

kevingosse (approved):

kevingosse:
> can we keep track of this for future improvements?

👍

PR #1230, opened 2021-02-23
CallTarget Integration: ServiceStack.Redis and StackExchange.Redis 

tonyredondo:
This PR adds support for both `ServiceStack.Redis` and `StackExchange.Redis` using CallTarget.

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Redis/StackExchange/RedisExecuteAsyncIntegration.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Redis/StackExchange/RedisExecuteAsyncIntegration.cs
@@ -7,30 +7,14 @@ using Datadog.Trace.Configuration;
 namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Redis.StackExchange
 {
     /// <summary>
-    /// StackExchange.Redis.RedisBase.ExecuteAsync[T] calltarget instrumentation
+    /// StackExchange.Redis.[RedisBase/RedisBatch/RedisTransaction].ExecuteAsync[T] calltarget instrumentation
     /// </summary>
-    [InstrumentMethod(
-        AssemblyName = "StackExchange.Redis",
-        TypeName = "StackExchange.Redis.RedisBase",
-        MethodName = "ExecuteAsync",
-        ReturnTypeName = "System.Threading.Tasks.Task`1<T>",
-        ParameterTypeNames = new[] { "StackExchange.Redis.Message", "StackExchange.Redis.ResultProcessor`1[!!0]", "StackExchange.Redis.ServerEndPoint" },
-        MinimumVersion = "1.0.0",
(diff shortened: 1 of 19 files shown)


src/Datadog.Trace.ClrProfiler.Native/clr_helpers.cpp
```diff
@@ -1121,10 +1121,10 @@ mdToken FunctionMethodArgument::GetTypeTok(ComPtr<IMetaDataEmit2>& pEmit, mdAsse
       pEmit->DefineTypeRefByName(corLibRef, SystemChar, &token);
       break;
     case ELEMENT_TYPE_I1:
-      pEmit->DefineTypeRefByName(corLibRef, SystemByte, &token);
+      pEmit->DefineTypeRefByName(corLibRef, SystemSByte, &token);
```
zacharycmontoya, line 1124: Oops, good catch!

zacharycmontoya (approved):
LGTM!

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Redis/StackExchange/RedisExecuteAsyncIntegration.cs
```diff
@@ -0,0 +1,63 @@
...
+using Datadog.Trace.Configuration;
+
+namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Redis.StackExchange
+{
+    /// <summary>
+    /// StackExchange.Redis.[RedisBase/RedisBatch/RedisTransaction].ExecuteAsync[T] calltarget instrumentation
+    /// </summary>
+    [RedisExecuteAsyncInstrumentMethod(TypeName = "StackExchange.Redis.RedisBase")]
+    [RedisExecuteAsyncInstrumentMethod(TypeName = "StackExchange.Redis.RedisBatch")]
+    [RedisExecuteAsyncInstrumentMethod(TypeName = "StackExchange.Redis.RedisTransaction")]
```
kevingosse, line 14: Looks like we only instrument RedisBatch: https://github.com/DataDog/dd-trace-dotnet/blob/master/src/Datadog.Trace.ClrProfiler.Managed/Integrations/StackExchange.Redis/RedisBatch.cs#L127

tonyredondo: `RedisBase` must be instrumented because there is code used by `RedisBatch` that is implemented in the parent class. I added `RedisTransaction` because also inherits `RedisBase`, but I can remove it if we need to.

In the case `RedisTransaction` we are missing the `ExecuteAsync` doesn't call the base class: https://github.com/StackExchange/StackExchange.Redis/blob/59ef0c4eb4bed27014521e0bf0c24bd18fcf596e/src/StackExchange.Redis/RedisTransaction.cs#L62-L109

kevingosse (approved):

PR #1237, opened 2021-02-24
Update .NET SDK versions in build

andrewlock:
* Update .NET SDK/runtime versions (we don't need the SDK for versions < 5.0)
* Update README with latest SDK requirements

Based on changes in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/53

@DataDog/apm-dotnet

### DIFF
--- a/.azure-pipelines/benchmarks.yml
+++ b/.azure-pipelines/benchmarks.yml
@@ -6,7 +6,7 @@ trigger:
 
 variables:
   buildConfiguration: release
-  dotnetCoreSdkVersion: 5.0.100
+  dotnetCoreSdkVersion: 5.0.103
   ddApiKey: $(DD_API_KEY)
   publishOutput: $(Build.SourcesDirectory)/src/bin/managed-publish
 
--- a/.azure-pipelines/integration-tests.yml
+++ b/.azure-pipelines/integration-tests.yml
@@ -22,8 +22,7 @@ pool:
 variables:
   buildConfiguration: Release
   publishOutput: $(System.DefaultWorkingDirectory)/src/bin/managed-publish
-  dotnetCoreSdkVersion: 3.1.107
-  dotnetCoreSdk5Version: 5.0.100
+  dotnetCoreSdk5Version: 5.0.103
 
 jobs:
 
@@ -31,9 +30,10 @@ jobs:
 
   steps:
   - task: UseDotNet@2
-    displayName: install dotnet core sdk 3.1
+    displayName: install dotnet core runtime 3.1
     inputs:
-      version: $(dotnetCoreSdkVersion)
+      packageType: runtime
+      version: 3.1.x
 
   - task: UseDotNet@2
     displayName: install dotnet core sdk 5
@@ -131,9 +131,10 @@ jobs:
 
   steps:
   - task: UseDotNet@2
-    displayName: install dotnet core sdk 3.1
+    displayName: install dotnet core runtime 3.1
     inputs:
-      version: $(dotnetCoreSdkVersion)
+      packageType: runtime
+      version: 3.1.x
 
   - task: UseDotNet@2
     displayName: install dotnet core sdk 5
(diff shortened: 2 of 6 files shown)


.azure-pipelines/integration-tests.yml
```diff
@@ -22,18 +22,18 @@ pool:
...
+  dotnetCoreSdk5Version: 5.0.103
 
 jobs:
 
 - job: build_linux_profiler
 
   steps:
   - task: UseDotNet@2
-    displayName: install dotnet core sdk 3.1
+    displayName: install dotnet core runtime 3.1
```
lucaspimentel, line 33: I always thought we only needed the latest SDK and only the runtimes for the rest, and it used to be like that, but at some point there was an issue (I can't remember) and we switched to installing all the SDK version instead. I hope this works as it will make the containers a bit smaller and faster to build, right?

lucaspimentel: Oh, looks it was because the base runtime does not include aspnetcore runtime. See #868.

lucaspimentel: Maybe not: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/dotnet-core-tool-installer?view=azure-devops#task-inputs

andrewlock: Soo... yeah... you're right. However, it's _only_ [the Windows x64 job that fails currently](https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=48304&view=logs&j=3c51c6d7-040c-5535-ae2f-42d5f8bb473b)... could be indicative that we're not running all the tests we think we are on other platforms?

As an aside, the Windows agents already have a bazillion versions installed, so we could probably get away without any of these anyway 🤷‍♂️

andrewlock: I don't really understand why the OTel fork was able to make these changes in https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/53 but we can't 🤔

kevingosse: I think it works on Linux because we install the aspnetcore runtime separately in the dockerfile: https://github.com/DataDog/dd-trace-dotnet/blob/master/build/docker/dotnet.dockerfile

.azure-pipelines/benchmarks.yml
```diff
@@ -6,7 +6,7 @@ trigger:
 
 variables:
   buildConfiguration: release
-  dotnetCoreSdkVersion: 5.0.100
+  dotnetCoreSdkVersion: 5.0.103
```
kevingosse, line 9: Maybe we should use `5.0.x` to get all the minor updates

kevingosse (approved):

.azure-pipelines/integration-tests.yml
andrewlock: argh, of course we do, I forget that this is only _part_ of the build definition 😛

.azure-pipelines/benchmarks.yml
andrewlock: I did that initially, and that's what the OTel version does, but in [Gitlab CI we're explicitly using 5.0.103](https://github.com/DataDog/datadog-agent-buildimages/pull/116/files#diff-dbacf0921694e46b4b06a826d1c98092f046739508ef3635251cb2b88cd57134R8-R9). Though it's probably better to keep those in sync?

lucaspimentel: We have had issues before with using latest. A new SDJ is released that requires an update to VS, but the AzDO images are not updated, so builds fail.

lucaspimentel (approved):

docs/README.md
```diff
@@ -38,9 +38,10 @@ Integration tests | [![Build Status](https://dev.azure.com/datadoghq/dd-trace-do
     - Optional: ASP.NET and web development (to build samples)
   - Individual components
     - .NET Framework 4.7 targeting pack
-- [.NET Core 3.1 SDK](https://dotnet.microsoft.com/download/dotnet-core/3.1)
+- [.NET 5.0 SDK](https://dotnet.microsoft.com/download/dotnet/5.0)
```
lucaspimentel, line 41: A few lines above, we should probably specify `Visual Studio 2019 (16.8)` (required for .NET 5.0)

andrewlock: done in https://github.com/DataDog/dd-trace-dotnet/pull/1237/commits/8165b719b204f2681a7816737755960a4499e85b

lucaspimentel: Thanks!

PR #1242, opened 2021-02-26
Adding Git information to test spans

tonyredondo:
This PR adds a git folder parser to include the git information to the test spans, as required in the ciapp specification.

Example: 
image

image

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace.Ci.Shared/CIEnvironmentValues.cs
+++ b/src/Datadog.Trace.Ci.Shared/CIEnvironmentValues.cs
@@ -48,7 +48,7 @@ namespace Datadog.Trace.Ci
 
         public static void DecorateSpan(Span span)
         {
-            if (span == null || !IsCI)
+            if (span == null)
             {
                 return;
             }
@@ -89,6 +89,8 @@ namespace Datadog.Trace.Ci
             Branch = null;
             Tag = null;
             SourceRoot = null;
+
+            GitInfo gitInfo = GitInfo.GetCurrent();
 
             if (EnvironmentHelpers.GetEnvironmentVariable("TRAVIS") != null)
             {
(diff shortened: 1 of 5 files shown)


src/Datadog.Trace.Ci.Shared/CIEnvironmentValues.cs
```diff
@@ -66,6 +81,21 @@ public static void DecorateSpan(Span span)
             span.SetTagIfNotNullOrEmpty(CommonTags.GitCommit, Commit);
             span.SetTagIfNotNullOrEmpty(CommonTags.GitBranch, Branch);
             span.SetTagIfNotNullOrEmpty(CommonTags.GitTag, Tag);
+            span.SetTagIfNotNullOrEmpty(CommonTags.GitCommitAuthorName, AuthorName);
+            span.SetTagIfNotNullOrEmpty(CommonTags.GitCommitAuthorEmail, AuthorEmail);
+            if (AuthorDate != DateTimeOffset.MinValue)
```
lucaspimentel, line 86: Should we use a nullable `DateTimeOffset` for `AuthorDate` and `CommitterDate`?

lucaspimentel (commented):
There's a ton of logic here to parse the git files that has no tests. Were you planning on adding them in a separate PR?

src/Datadog.Trace.Ci.Shared/CIEnvironmentValues.cs
tonyredondo: You're right, done in b69d87d

tonyredondo:
> There's a ton of logic here to parse the git files that has no tests. Were you planning on adding them in a separate PR?

Yes I plan to collect a couple of `git like` folders to run the tests, and submit another PR for that.

zacharycmontoya (approved):
I glossed over the implementations of `TryGetPackageOffset` and `TryGetFromPackageOffset`. But everything besides that, including the git objects and the config parsing, look good. It sounds like you are planning to add tests in a separate PR so I think this PR is good to go

lucaspimentel (approved):
Cool feature. When can we use it? 😆

PR #1244, opened 2021-02-26
Fix CiApp Tests ServiceName

tonyredondo:
This PR fixes the problem with the tests ServiceName using the test suite inside CiApp.

This is done by:
1. Storing the tracer instance as the TestTracer in the integration static constructor.
2. Storing the service name in the integration static constructor.
3. Modify azure pipelines to include the `DD_SERVICE_NAME`.

This protects both variables of any modification inside the tracer tests.

Previous service list:

image

@DataDog/apm-dotnet

### DIFF
--- a/.azure-pipelines/benchmarks.yml
+++ b/.azure-pipelines/benchmarks.yml
@@ -73,6 +73,7 @@ stages:
         projects: '$(System.DefaultWorkingDirectory)/test/benchmarks/Benchmarks.Trace/Benchmarks.Trace.csproj'
         arguments: '-c Release -f netcoreapp3.1 -- -r net472 netcoreapp3.1 -m -f * --iterationTime 2000'
       env:
+        DD_ENV: CI
         DD_SERVICE: dd-trace-dotnet
 
     - task: PowerShell@2
--- a/.azure-pipelines/benchmarks.yml
+++ b/.azure-pipelines/benchmarks.yml
@@ -73,7 +73,7 @@ stages:
         projects: '$(System.DefaultWorkingDirectory)/test/benchmarks/Benchmarks.Trace/Benchmarks.Trace.csproj'
         arguments: '-c Release -f netcoreapp3.1 -- -r net472 netcoreapp3.1 -m -f * --iterationTime 2000'
       env:
-        DD_SERVICE: dd-tracer-dotnet
+        DD_SERVICE: dd-trace-dotnet
 
     - task: PowerShell@2
       inputs:
--- a/.azure-pipelines/benchmarks.yml
+++ b/.azure-pipelines/benchmarks.yml
@@ -73,7 +73,7 @@ stages:
         projects: '$(System.DefaultWorkingDirectory)/test/benchmarks/Benchmarks.Trace/Benchmarks.Trace.csproj'
         arguments: '-c Release -f netcoreapp3.1 -- -r net472 netcoreapp3.1 -m -f * --iterationTime 2000'
       env:
-        DD_SERVICE_NAME: dd-tracer-dotnet
+        DD_SERVICE: dd-tracer-dotnet
 
     - task: PowerShell@2
       inputs:
--- a/.azure-pipelines/benchmarks.yml
+++ b/.azure-pipelines/benchmarks.yml
@@ -72,6 +72,9 @@ stages:
         command: 'run'
         projects: '$(System.DefaultWorkingDirectory)/test/benchmarks/Benchmarks.Trace/Benchmarks.Trace.csproj'
         arguments: '-c Release -f netcoreapp3.1 -- -r net472 netcoreapp3.1 -m -f * --iterationTime 2000'
+      env:
+        DD_SERVICE_NAME: dd-tracer-dotnet
+
     - task: PowerShell@2
       inputs:
         targetType: 'inline'
--- a/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/Common.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/Common.cs
@@ -1,0 +1,20 @@
+using Datadog.Trace.Ci;
+
+namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Testing
+{
+    internal static class Common
+    {
+        static Common()
+        {
+            TestTracer = Tracer.Instance;
+            ServiceName = TestTracer.DefaultServiceName;
+
+            // Preload environment variables.
+            CIEnvironmentValues.DecorateSpan(null);
+        }
+
+        internal static Tracer TestTracer { get; private set; }
+
+        internal static string ServiceName { get; private set; }
+    }
+}
--- a/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/MsTestV2/TestMethodRunnerExecuteIntegration.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/MsTestV2/TestMethodRunnerExecuteIntegration.cs
@@ -25,9 +25,14 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Testing.MsTestV2
     {
         private const string IntegrationName = nameof(IntegrationIds.MsTestV2);
         private static readonly IntegrationInfo IntegrationId = IntegrationRegistry.GetIntegrationInfo(IntegrationName);
+        private static readonly Tracer TestTracer;
+        private static readonly string ServiceName;
 
         static TestMethodRunnerExecuteIntegration()
         {
+            TestTracer = Tracer.Instance;
+            ServiceName = TestTracer.DefaultServiceName;
+
             // Preload environment variables.
             CIEnvironmentValues.DecorateSpan(null);
         }
@@ -54,8 +59,7 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Testing.MsTestV2
             string testSuite = testMethodInfo.TestClassName;
             string testName = testMethodInfo.TestMethodName;
 
-            Tracer tracer = Tracer.Instance;
-            Scope scope = tracer.StartActive("mstest.test");
+            Scope scope = TestTracer.StartActive("mstest.test", serviceName: ServiceName);
             Span span = scope.Span;
 
             span.Type = SpanTypes.Test;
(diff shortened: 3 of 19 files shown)


src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/MsTestV2/TestMethodRunnerExecuteIntegration.cs
```diff
@@ -25,9 +25,14 @@ public class TestMethodRunnerExecuteIntegration
     {
         private const string IntegrationName = nameof(IntegrationIds.MsTestV2);
         private static readonly IntegrationInfo IntegrationId = IntegrationRegistry.GetIntegrationInfo(IntegrationName);
+        private static readonly Tracer TestTracer;
+        private static readonly string ServiceName;
 
         static TestMethodRunnerExecuteIntegration()
         {
+            TestTracer = Tracer.Instance;
```
lucaspimentel, line 33: This assume that users never replace `Tracer.Instance` with a new tracer instance (possibly with a different service name). Is that something we should worry about?

.azure-pipelines/benchmarks.yml
```diff
@@ -72,6 +72,9 @@ stages:
         command: 'run'
         projects: '$(System.DefaultWorkingDirectory)/test/benchmarks/Benchmarks.Trace/Benchmarks.Trace.csproj'
         arguments: '-c Release -f netcoreapp3.1 -- -r net472 netcoreapp3.1 -m -f * --iterationTime 2000'
+      env:
+        DD_SERVICE_NAME: dd-tracer-dotnet
```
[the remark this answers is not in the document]
tonyredondo: oh, thanks!, I will change it.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/MsTestV2/TestMethodRunnerExecuteIntegration.cs
tonyredondo: in a testing scenario I cannot think why we should allow to replace the tracer instance. I did this mostly to avoid our test suite messing with the tracer being used to report the tests.

.azure-pipelines/benchmarks.yml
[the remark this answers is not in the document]
tonyredondo: done in 87d5835

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/MsTestV2/TestMethodRunnerExecuteIntegration.cs
lucaspimentel: It's probably fine. I just get nervous about saving the tracer instance. We're been bitten by bugs like these before.

> I cannot think why we should allow to replace the tracer instance.

We can't really stop users from doing that, can we? 

> to avoid our test suite messing with the tracer being used to report the tests

Out of scope of this PR, but we should make sure that we never use `Tracer.Instance` for testing purposes (only use it for for tracing). Tests should create their own `Tracer` instances if needed.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/MsTestV2/TestMethodRunnerExecuteIntegration.cs
```diff
@@ -54,8 +59,7 @@ public static CallTargetState OnMethodBegin<TTarget>(TTarget instance)
             string testSuite = testMethodInfo.TestClassName;
             string testName = testMethodInfo.TestMethodName;
 
-            Tracer tracer = Tracer.Instance;
-            Scope scope = tracer.StartActive("mstest.test");
+            Scope scope = TestTracer.StartActive("mstest.test", serviceName: ServiceName);
```
lucaspimentel, line 62: To use the same instance consistently, we should probably replace all uses of `Tracer.Instance` with `TestTracer`, like line 49 above.

.azure-pipelines/benchmarks.yml
```diff
@@ -72,6 +72,9 @@ stages:
         command: 'run'
         projects: '$(System.DefaultWorkingDirectory)/test/benchmarks/Benchmarks.Trace/Benchmarks.Trace.csproj'
         arguments: '-c Release -f netcoreapp3.1 -- -r net472 netcoreapp3.1 -m -f * --iterationTime 2000'
+      env:
+        DD_SERVICE: dd-tracer-dotnet
```
zacharycmontoya, line 76:
```suggestion
        DD_SERVICE: dd-trace-dotnet
```

tonyredondo: you're right, done in aac0b03

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/MsTestV2/TestMethodRunnerExecuteIntegration.cs
tonyredondo: thanks for the catch, I did a refactor to use the same instance in multiple classes in aac0b03

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/Common.cs
```diff
@@ -0,0 +1,20 @@
+using Datadog.Trace.Ci;
+
+namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Testing
+{
+    internal static class Common
+    {
+        static Common()
+        {
+            TestTracer = Tracer.Instance;
```
zacharycmontoya, line 9: What if we replace the Tracer entirely with a new one like this:

```suggestion
            TestTracer = new Tracer(Tracer.Instance.Settings);
```

tonyredondo: My only concern is that by doing this will cause disconnected traces If the test calls something that we have an integration for, because the integration will be using the `Tracer.Instance` isn't?  And one of the goals for the CIApp team is to have distributed traces for the tests for integration tests.

zacharycmontoya: Oh I see, yes the traces would be disconnected then. I didn't realize distributed tracing inside integration tests was a goal, thanks for clarifying 👍🏼

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/MsTestV2/TestMethodRunnerExecuteIntegration.cs
tonyredondo: > It's probably fine. I just get nervous about saving the tracer instance. We're been bitten by bugs like these before.

Do you know more about these bugs? To make sure I'm not missing something. Also I can do something like:
```csharp
public Tracer TestTracer => _testTracer ?? Tracer.Instance;
```
To ensure the TestTracer is not null.

> 
> > I cannot think why we should allow to replace the tracer instance.
> 
> We can't really stop users from doing that, can we?

If the `TestTracer` is fixed, if the user changes the `Tracer.Instance` later doesn't affect the test integration. I meant I cannot think a scenario where the user would change the TestTracer instance on purpose.

> 
> > to avoid our test suite messing with the tracer being used to report the tests
> 
> Out of scope of this PR, but we should make sure that we never use `Tracer.Instance` for testing purposes (and only use it for tracing). Tests should create their own `Tracer` instances if needed.

I agree, we currently lack of this isolation, tests can replace the `Tracer.Instance` and also there are some tests that relies on static properties (running in parallel is a source of race conditions).

test/Datadog.Trace.Tests/Configuration/ConfigurationSourceTests.cs
```diff
@@ -161,9 +161,27 @@ public void DefaultSetting(Func<TracerSettings, object> settingGetter, object ex
             // save original value so we can restore later
             var originalValue = Environment.GetEnvironmentVariable(key);
 
-            Environment.SetEnvironmentVariable(key, value, EnvironmentVariableTarget.Process);
-            IConfigurationSource source = new EnvironmentConfigurationSource();
-            var settings = new TracerSettings(source);
+            TracerSettings settings;
+
+            if (key == "DD_SERVICE_NAME")
```
zacharycmontoya, line 166:
```suggestion
            if (key == "DD_SERVICE_NAME" || key == "DD_SERVICE")
```
We accept `DD_SERVICE` and `DD_SERVICE_NAME`. Can you also modify this branch to use `key` instead of `ConfigurationKeys.ServiceName`?

zacharycmontoya: Oh I see, you're trying to test that, even when `DD_SERVICE` is set in CI, that this `DD_SERVICE_NAME` test config takes effect. Nevermind, this should work as-is, ignore my comment

zacharycmontoya (approved):

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/MsTestV2/TestMethodRunnerExecuteIntegration.cs
lucaspimentel: >Do you know more about these bugs?

Users can set configuration in several ways: environment variables, `web.config` files, or in code. For the latter, users need to do something like this (normally during application startup):
```csharp
var settings = new TracerSettings
               {
                   // ... set some properties ...
               };

Tracer.Instance = new Tracer(settings);
```

We had a bug in the ASP.NET Core integration where we cached the value of `Tracer.Instance`, but users who were doing configuration in code would replace `Tracer.Instance` _after_ that, so we did not see any changes to the settings. Until we fixed the bug, the workaround was to always set configuration using environment variables.

tonyredondo: I think for now we should store the `TestTracer` instance and revisit this issue when we change the tests that uses the `Tracer.Instance` in another PR. Because not being able to change the default configuration of the TestTracer is not big issue right now.

lucaspimentel: Sounds good to me. Resolving this thread.

lucaspimentel (approved):

PR #1246, opened 2021-03-01
Fix and extend ASP.NET integration tests

andrewlock:
Fixes ASP.NET MVC 4/5 integration tests being skipped (I think) since https://github.com/DataDog/dd-trace-dotnet/commit/fadfcb4135cbb63d3305808b1cfb2a7ec966d97e due to missing `IXunitSerializable` properties (could be run locally, but seemed to erroneously pass in CI)

Adds additional test cases for route parameter handling - optional parameters, bad values etc. Using conventional and attribute routing. 

@DataDog/apm-dotnet


test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNet/AspNetMvc5Tests.cs
```diff
@@ -2,15 +2,14 @@
...
-using System.Collections.Generic;
 using System.Net;
 using System.Threading.Tasks;
 using Xunit;
 using Xunit.Abstractions;
 
 namespace Datadog.Trace.ClrProfiler.IntegrationTests
 {
-    [Collection("IisTests")]
+    [Collection(nameof(IisFixture))]
```
lucaspimentel, line 12: The MVC 4 tests use `[Collection("IisTests")]` so these will have a different collection name. Was that intentional?

test/Datadog.Trace.ClrProfiler.IntegrationTests/Helpers/SerializableDictionary.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using Datadog.Trace.Vendors.Newtonsoft.Json;
+using Xunit.Abstractions;
+
+namespace Datadog.Trace.ClrProfiler.IntegrationTests
+{
+    public class SerializableDictionary : IXunitSerializable, IEnumerable<KeyValuePair<string, string>>
```
lucaspimentel, line 8: I'm surprised that xUnit does not support serializing ` Dictionary<string, string>` out of the box. 🤔

test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNet/AspNetMvc5Tests.cs
andrewlock: Oops, no, that was something else

test/Datadog.Trace.ClrProfiler.IntegrationTests/Helpers/SerializableDictionary.cs
andrewlock: Yeah, I need to double check, as I initially assumed the dictionary was the problem, but later found that `HttpStatusCode` was an issue. But given even `Guid` is not serializable by default, I strongly suspect the dictionary is not either 😉

test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNet/AspNetMvc5Tests.cs
andrewlock: Fixed in https://github.com/DataDog/dd-trace-dotnet/pull/1246/commits/7d3b76f5a67a58d6c14e758097cb28eae2f4b333

test/Datadog.Trace.ClrProfiler.IntegrationTests/Helpers/SerializableDictionary.cs
andrewlock: Yep, confirmed, you can't use `Dictionary<string, string>` 🙄

lucaspimentel: Thanks for checking!

lucaspimentel (approved):

zacharycmontoya (approved):

PR #1253, opened 2021-03-03
Improve DD_TRACE_HEADER_TAGS to tag web server spans with response headers

zacharycmontoya:
Previously, the `DD_TRACE_HEADER_TAGS` setting would add additional tags to web server spans using the provided header-to-tag mapping, but it would only consider incoming HTTP requests. This PR adds functionality to also consider headers from the HTTP response the web server issues. This helps developers to automatically learn more about the response sent to their users.

Affected integrations:
- AspNet
- AspNetMvc
- AspNetWebApi2 (though headers would never be populated during local tests?)
- AspNetCore

Testing:
- Local testing for all integrations
- AspNetCore integration test added

@DataDog/apm-dotnet


src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
```diff
@@ -67,6 +71,26 @@ public static void AddTagsFromDbCommand(this Span span, IDbCommand command)
             }
         }
 
+        internal static void ApplyHeaderTags(this Span span, IHeadersCollection headers, IDictionary<string, string> headerTags)
```
lucaspimentel, line 74: We usually talk about "setting" tags, not "applying" them. Consider renaming to `SetHeaderTags()`. Level of care low.

src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
```diff
@@ -67,6 +71,26 @@ public static void AddTagsFromDbCommand(this Span span, IDbCommand command)
...
+        {
+            if (headerTags is not null && !headerTags.IsEmpty())
+            {
+                try
+                {
+                    // extract propagation details from http headers
+                    var tagsFromHeaders = SpanContextPropagator.Instance.ExtractHeaderTags(headers, headerTags);
+                    foreach (KeyValuePair<string, string> kvp in tagsFromHeaders)
+                    {
+                        span.SetTag(kvp.Key, kvp.Value);
```
lucaspimentel, line 84: According to the docs (which we failed to follow when we originally added this feature), tag names should be:
```
http.<request_or_response>.headers.<normalized_header_name>
```
Also, we already use `http.request.headers.host`. Should we change this to follow the RFC? And if we do, should we keep using the current tag name for requests in addition to the new name (for backwards compat)?

PR #1254, opened 2021-03-03
Fix CallTargetNativeTest build

tonyredondo:
@DataDog/apm-dotnet


kevingosse (approved):
Shouldn't that be built in the CI as well?

tonyredondo:
> Shouldn't that be built in the CI as well?

Right now is not being built because the test to run this console app is missing, I'll be adding that in another PR. This just fixes the compilation of the solution in local.

andrewlock (approved):

PR #1256, opened 2021-03-03
Fix solution load deadlock for Rider in non windows OS

tonyredondo:
@DataDog/apm-dotnet

### DIFF
--- a/test/Directory.Build.props
+++ b/test/Directory.Build.props
@@ -2,7 +2,8 @@
   <Import Project="..\Directory.Build.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>net452;net461;netcoreapp2.1;netcoreapp3.0;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">net452;net461;netcoreapp2.1;netcoreapp3.0;netcoreapp3.1</TargetFrameworks>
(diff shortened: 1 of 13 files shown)


lucaspimentel (approved):

test/Directory.Build.props
```diff
@@ -21,4 +21,10 @@
   <ItemGroup>
     <Compile Include="$(MSBuildThisFileDirectory)GlobalSuppressions.cs" Link="GlobalSuppressions.test.cs" />
   </ItemGroup>
+  
+  <ItemGroup>
+    <!-- reference assemblies let us target .NET Framework without the SDK (for example, on non-Windows) -->
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0" PrivateAssets="all" />
```
lucaspimentel, line 27: Is there a good way to add this _everywhere_?

tonyredondo: I really just added on projects that were failing to built, that was my intent at least. I don't know if we want this as a global thing.

lucaspimentel: I was just wondering if we should try. Definitely not blocking your PR.

We already it include it in some top-level `props` files, but not all. Was wondering if we should make it more global since it doesn't hurt to have it even in projects that don't need it.

tonyredondo: Yeah, makes sense, I will try that in another PR.

tonyredondo:
I will merge this PR regardless 1 approval, is not critical and unblocks the development :)

PR #1257, opened 2021-03-03
DuckIgnoreAttribute to ignore members to be used when ducktyping

tonyredondo:
This PR Adds the DuckIgnoreAttribute to ignore members when the ducktype proxy is created.

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace/DuckTyping/DuckType.Methods.cs
+++ b/src/Datadog.Trace/DuckTyping/DuckType.Methods.cs
@@ -22,18 +22,33 @@ namespace Datadog.Trace.DuckTyping
                     continue;
                 }
 
-                List<MethodInfo> newMethods = imInterface.GetMethods()
-                    .Where(iMethod =>
+                foreach (MethodInfo intMethod in imInterface.GetMethods())
+                {
+                    if (intMethod.IsSpecialName)
                     {
-                        if (iMethod.IsSpecialName)
-                        {
-                            return false;
-                        }
+                        continue;
+                    }
 
-                        string iMethodString = iMethod.ToString();
-                        return selectedMethods.All(i => i.ToString() != iMethodString);
-                    }).ToList();
-                selectedMethods.AddRange(newMethods);
+                    string intMethodName = intMethod.ToString();
+                    bool found = false;
+                    foreach (MethodInfo currentMethod in selectedMethods)
+                    {
+                        if (currentMethod.ToString() == intMethodName)
+                        {
+                            found = true;
+                            break;
+                        }
+                    }
+
+                    if (!found)
+                    {
+                        MethodInfo prevMethod = baseType.GetMethod(intMethod.Name, DuckAttribute.DefaultFlags, null, intMethod.GetParameters().Select(p => p.ParameterType).ToArray(), null);
+                        if (prevMethod == null || prevMethod.GetCustomAttribute<DuckIgnoreAttribute>() is null)
+                        {
+                            selectedMethods.Add(intMethod);
+                        }
+                    }
+                }
             }
 
             return selectedMethods;
@@ -68,6 +83,12 @@ namespace Datadog.Trace.DuckTyping
             List<MethodInfo> proxyMethodsDefinitions = GetMethods(proxyType);
             foreach (MethodInfo proxyMethodDefinition in proxyMethodsDefinitions)
             {
+                // Ignore the method marked with `DuckIgnore` attribute
+                if (proxyMethodDefinition.GetCustomAttribute<DuckIgnoreAttribute>(true) is not null)
+                {
+                    continue;
+                }
+
                 // Extract the method parameters types
                 ParameterInfo[] proxyMethodDefinitionParameters = proxyMethodDefinition.GetParameters();
                 Type[] proxyMethodDefinitionParametersTypes = proxyMethodDefinitionParameters.Select(p => p.ParameterType).ToArray();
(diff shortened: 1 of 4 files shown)


src/Datadog.Trace/DuckTyping/DuckType.Methods.cs
```diff
@@ -22,18 +22,33 @@ private static List<MethodInfo> GetMethods(Type baseType)
...
+                foreach (MethodInfo intMethod in imInterface.GetMethods())
+                {
+                    if (intMethod.IsSpecialName)
+                    {
+                        continue;
+                    }
+
+                    string intMethodName = intMethod.ToString();
+                    bool found = false;
+                    foreach (MethodInfo currentMethod in selectedMethods)
```
zacharycmontoya, line 34: Suggestion: I think doing a LINQ `Any` might be a little more readable

```
if (selectedMethods.Any(method => method.ToString() == intMethodName))
{
    continue;
}
```

test/Datadog.Trace.DuckTyping.Tests/DuckIgnoreTests.cs
```diff
@@ -0,0 +1,137 @@
...
+        public struct CopyStruct : IGetValue
+        {
+            public ValuesDuckType Value;
+
+            string IGetValue.GetValueProp => Value.ToString();
+
+            public string GetValue() => Value.ToString();
+        }
+
+#if NETCOREAPP3_0_OR_GREATER
```
zacharycmontoya, line 54: How come this definition only works for netcoreapp3.0 and higher?

test/Datadog.Trace.DuckTyping.Tests/DuckIgnoreTests.cs
```diff
@@ -0,0 +1,137 @@
...
+            [DuckIgnore]
+            public string GetValue() => Value.ToString();
+        }
+
+        public class VirtualPrivateProxy : IGetValue
+        {
+            public virtual ValuesDuckType Value { get; }
+
+            [DuckIgnore]
+            public string GetValueProp => Value.ToString();
```
zacharycmontoya, line 83: Could you also test that calling the `[DuckIgnore]` methods works and they return the expected result?

test/Datadog.Trace.DuckTyping.Tests/DuckIgnoreTests.cs
tonyredondo: Interfaces with a default implementation only compiles in that range.

zacharycmontoya: Ah I didn't make the connection that this was an interface and there default interface implementations. Thanks

zacharycmontoya (approved):
LGTM. I also left some non-blocking comments

src/Datadog.Trace/DuckTyping/DuckType.Methods.cs
tonyredondo: What if I tell you I removed the LINQ on purpose? 😂 I wanted to remove the closure and also for a simple comparison. But no strong opinion about it.

test/Datadog.Trace.DuckTyping.Tests/DuckIgnoreTests.cs
tonyredondo: Sure

tonyredondo: Done in b3ca50a

test/Datadog.Trace.DuckTyping.Tests/DuckIgnoreTests.cs
tonyredondo: I added a comment on the type definition :)

src/Datadog.Trace/DuckTyping/DuckType.Methods.cs
zacharycmontoya: Haha that makes sense! I don't have a strong opinion either, it just took me a minute to realize what this loop was doing. Maybe just adding a small comment here or renaming the `bool found` to be a little more explanatory could help

test/Datadog.Trace.DuckTyping.Tests/DuckIgnoreTests.cs
```diff
@@ -0,0 +1,146 @@
...
+namespace Datadog.Trace.DuckTyping.Tests
+{
+    public class DuckIgnoreTests
+    {
+        [Fact]
+        public void NonPublicStructCopyTest()
+        {
+            PrivateStruct instance = default;
+            CopyStruct copy = instance.DuckCast<CopyStruct>();
+            Assert.Equal((int)instance.Value, (int)copy.Value);
```
kevingosse, line 15: Shouldn't you assign a value first? Here the assertion will work even if Value isn't actually copied to the proxy, since you're keeping the default value

kevingosse: Never mind, I didn't realize it was a property

kevingosse (approved):

src/Datadog.Trace/DuckTyping/DuckType.Methods.cs
tonyredondo: I did the renaming in 2719b3c

PR #1263, opened 2021-03-05
Propose fix to Datadog.Trace.Tests.RuntimeMetrics.RuntimeMetricsWriterTests.ShouldCaptureFirstChanceExceptions

zacharycmontoya:
We were previously asserting that no first exceptions were being recorded, however other threads in the same AppDomain could be throwing exceptions which our runtime metrics would catch. Replace this catch-all check with a more targeted one.

@DataDog/apm-dotnet

### DIFF
--- a/test/Datadog.Trace.Tests/RuntimeMetrics/RuntimeMetricsWriterTests.cs
+++ b/test/Datadog.Trace.Tests/RuntimeMetrics/RuntimeMetricsWriterTests.cs
@@ -88,7 +88,11 @@ namespace Datadog.Trace.Tests.RuntimeMetrics
                 writer.PushEvents();
 
                 statsd.Verify(
-                    s => s.Increment(MetricsNames.ExceptionsCount, It.IsAny<int>(), It.IsAny<double>(), It.IsAny<string[]>()),
+                    s => s.Increment(MetricsNames.ExceptionsCount, It.IsAny<int>(), It.IsAny<double>(), new[] { "exception_type:CustomException1" }),
+                    Times.Never);
+
+                statsd.Verify(
+                    s => s.Increment(MetricsNames.ExceptionsCount, It.IsAny<int>(), It.IsAny<double>(), new[] { "exception_type:CustomException2" }),
                     Times.Never);
             }
         }
--- a/test/Datadog.Trace.Tests/RuntimeMetrics/RuntimeMetricsWriterTests.cs
+++ b/test/Datadog.Trace.Tests/RuntimeMetrics/RuntimeMetricsWriterTests.cs
@@ -69,7 +69,11 @@ namespace Datadog.Trace.Tests.RuntimeMetrics
                 }
 
                 statsd.Verify(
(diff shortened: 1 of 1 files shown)


test/Datadog.Trace.Tests/RuntimeMetrics/RuntimeMetricsWriterTests.cs
```diff
@@ -88,7 +88,11 @@ public void ShouldCaptureFirstChanceExceptions()
                 writer.PushEvents();
 
                 statsd.Verify(
-                    s => s.Increment(MetricsNames.ExceptionsCount, It.IsAny<int>(), It.IsAny<double>(), It.IsAny<string[]>()),
+                    s => s.Increment(MetricsNames.ExceptionsCount, It.IsAny<int>(), It.IsAny<double>(), new[] { "exception_type:CustomException1" }),
```
lucaspimentel, line 91: Should we add this to the other "never" verification in lines 71-73 above, too?

zacharycmontoya: Yeah good point, will do

zacharycmontoya: Done in 480242c4869d94210048d348978ff35011f08ce1

zacharycmontoya: Actually, in retrospect, that first one is not needed, because we don't push ANY exceptions until the PushEvents method is called. I'm actually going to revert that last commit

lucaspimentel: In the PR description you said "other threads in the same AppDomain could be throwing exceptions which our runtime metrics would catch." That doesn't apply to this check?

zacharycmontoya: The very first one, it does not, because we don't submit any metrics to StatsD instance until the first `writer.PushEvents()`. That's why it's good to keep the broad `It.IsAny<string[]>()` check as it is before the first `writer.PushEvents()` call

zacharycmontoya: My _theory_ is that there is a race condition in between the `statsd.Invocations.Clear()` and the final `writer.PushEvents()` where exceptions thrown from other test methods could be collected by this runtime metrics writer. We're really just trying to check that the local exceptions we threw got collected, written to StatsD, and will not show up again. For that reason, I'm proposing to only check the custom exceptions

lucaspimentel: Sounds good. Thanks for clarifying. I will resolve this comment.

lucaspimentel (approved):

kevingosse (approved):
Thanks!

Issue #447, opened 2019-07-16
StackOverflowException when iterating through Redis keys


rpmrmartin:
**Describe the bug**
When our ASP.NET application iterates through all the keys on a Redis server, it causes the Datadog .NET Tracer to throw a StackOverflowException.

**To Reproduce**
Run the following code inside an application that is being profiled by the Datadog .NET Tracer:
```csharp
using StackExchange.Redis;
ConnectionMultiplexer redis = RedisCaching.GetRedisMultiplexer();
IServer server = redis.GetServer("MyServer");
IEnumerator ide = server.Keys().GetEnumerator();
while (ide.MoveNext())
{
	// Do something
}
```
The exception is thrown when calling `ide.MoveNext()` for the first time

**Expected behavior**
The application should iterate over the keys without issue

**Runtime environment (please complete the following information):**
- Instrumentation mode: Automatic with MSI installer
- Tracer version: 1.5.0
- OS: Found on Windows Server 2016, reproduced on Windows 10 Pro 1809
- CLR: .NET Framework 4.7.2

**Additional context**
Stack trace:
```
<Repeat the last four calls ad infinitum>  

StackExchange.Redis.dll!StackExchange.Redis.RedisServer.ExecuteAsync<StackExchange.Redis.RedisBase.CursorEnumerable<StackExchange.Redis.RedisKey>.ScanResult>(StackExchange.Redis.Message message, StackExchange.Redis.ResultProcessor<StackExchange.Redis.RedisBase.CursorEnumerable<StackExchange.Redis.RedisKey>.ScanResult> processor, StackExchange.Redis.ServerEndPoint server) Line 577	C#  

Datadog.Trace.ClrProfiler.Managed.dll!Datadog.Trace.ClrProfiler.Integrations.StackExchange.Redis.RedisBatch.ExecuteAsyncInternal<StackExchange.Redis.RedisBase.CursorEnumerable<StackExchange.Redis.RedisKey>.ScanResult>(object redisBase, object message, object processor, object server)	Unknown  

[...]

PR #1264, opened 2021-03-05
Propose fix to Datadog.Trace.ClrProfiler.IntegrationTests.SmokeTests.StackExchangeRedisStackOverflowExceptionSmokeTest.NoExceptions

zacharycmontoya:
The purpose of this small regression test iterates over all keys in the redis database to test the stack overflow failure from https://github.com/DataDog/dd-trace-dotnet/issues/447.

However, when a GET is issued for each key, some values are not strings. This is likely happening because another test application using the same redis instance (`stackexchangeredis` container in `docker-compose.yml`) is simultaneously adding other key/value pairs during its execution. While we could just patch the smoke test, the real fix is to make sure the tests that use the same instance for the operations do not run in parallel. This proposal should not significantly increase the test execution time because the regression tests only run one test each, whereas the `StackExchangeRedisTests` is already running a larger matrix of tests based on CallSite/CallTarget + PackageVersion combinations.

The fix I'm proposing is to put all StackExchange tests into a new `StackExchangeRedisTestCollection`.

@DataDog/apm-dotnet

### DIFF
--- a/test/Datadog.Trace.ClrProfiler.IntegrationTests/TestCollections/StackExchangeRedisTestCollection.cs
+++ b/test/Datadog.Trace.ClrProfiler.IntegrationTests/TestCollections/StackExchangeRedisTestCollection.cs
@@ -1,0 +1,9 @@
+using Xunit;
+
+namespace Datadog.Trace.ClrProfiler.IntegrationTests.TestCollections
+{
+    [CollectionDefinition(nameof(StackExchangeRedisTestCollection), DisableParallelization = true)]
+    public class StackExchangeRedisTestCollection
+    {
+    }
+}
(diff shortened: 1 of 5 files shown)


andrewlock (approved):

kevingosse (approved):

colin-higgins (approved):

test/Datadog.Trace.ClrProfiler.IntegrationTests/TestCollections/StackExchangeRedisTestCollection.cs
```diff
@@ -0,0 +1,9 @@
+using Xunit;
+
+namespace Datadog.Trace.ClrProfiler.IntegrationTests.TestCollections
+{
+    [CollectionDefinition(nameof(StackExchangeRedisTestCollection), DisableParallelization = true)]
```
pellared, line 5: @zacharycmontoya
side note: probablyDisableParallelization = true` was what I was missing in my PoC 😉

pellared (approved):

PR #1269, opened 2021-03-08
Add additional ASP.NET Core tests + fix response code bug

andrewlock:
This PR adds a whole load of additional tests around the ASP.NET Core diagnostic listener. It documents the resource names associated with spans, expected status codes, as well as tags. This work is primarily a precursor to upcoming changes to resource names, to ensure that we can correctly preserve old behaviour behind a feature flag as required.

This PR increases the matrix of ASP.NET Core applications tested:

* ASP.NET Core 2.1 MVC
* ASP.NET Core 3.0/3.1/5.0 MVC with endpoint-routing disabled
* ASP.NET Core 3.0/3.1/5.0 MVC with endpoint-routing enabled
* ASP.NET Core 3.0/3.1/5.0 endpoints
* ASP.NET Core 2.1/3.0/3.1/5.0 endpoints

It also adds tests for various exception handling middleware for each of the above:
* No exception handling, i.e. caught by the web server (Kestrel)
* Basic exception handling (developer exception page)
* Custom exception handling (lets some exceptions bubble up)
* Pipeline-reexecution error handling (e.g. `UseExceptionHandler("/Error")`)

These highlighted two bugs:
* Where there is no exception handling, the status code recorded for the trace will be incorrect (`200` instead of `500`). This is fixed in this PR.
* Where the pipeline is re-executed, the resource name is overwritten with the error-path resource name, instead of preserving the original. This requires more work to fix, so is not addressed in this PR.

To test Razor Pages with `TetsServer` it was necessary to create a separate web app and reference it from the test app. That requires ~~strong naming the sample applications~~.  This got super viral, so I switched to just strong naming the single app.

~~Also, as the diagnostic sources tests require the aspnetcore runtime, I reverted back to installing the SDKs for 2.0/3.0/3.1. If that's undesirable, another option is to move the tests to the integration tests project, but I personally consider these to be closer to unit tests, as they all run in-memory.~~ This caused all sorts of deadlocks etc in the CI, likely due to the way the build scripts don't build the whole project (including samples) etc. Moved the tests to the integration test project and the problem resolved itself

@DataDog/apm-dotnet


test/test-applications/integrations/Samples.AspNetCoreRazorPages/Startup.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            Configuration = configuration;
+        }
+
+        public IConfiguration Configuration { get; }
+
+        // This method gets called by the runtime. Use this method to add services to the container.
+        public void ConfigureServices(IServiceCollection services)
+        {
+#if NETCOREAPP2_1
```
zacharycmontoya, line 20: Is this app still useful to run and test on netcoreapp2.1?

test/Datadog.Trace.IntegrationTests/DiagnosticListeners/AspNetCoreTestApp.cs
```diff
@@ -0,0 +1,198 @@
...
+    public class MyTestController : Controller
+    {
+        public async Task<string> Index()
+        {
+            await Task.Yield();
+            return "Hello world";
+        }
+
+        [HttpGet("/statuscode/{value=200}")]
+        public string SetStatusCode(int value) => value.ToString();
```
zacharycmontoya, line 167: Can you return a response with the specified status code here?

zacharycmontoya (approved):
LGTM. Thanks for lending your knowledge of ASP.NET Core to test all these code paths 😄

test/test-applications/integrations/Samples.AspNetCoreRazorPages/Startup.cs
andrewlock: Yes, `services.AddMvc()` includes Razor Pages support in .NET Core 2.1. (`AddMvc()` is also available in later versions, but the standard approach is the dedicated `AddRazorPages()` call, which avoids loading controller-based dependencies)

test/Datadog.Trace.IntegrationTests/Program.cs
```diff
@@ -0,0 +1,11 @@
+﻿#if NETFRAMEWORK
+namespace Datadog.Trace.IntegrationTests
+{
+    public static class Program
+    {
+        public static void Main()
```
lucaspimentel, line 6: Do we need this?

test/Datadog.Trace.IntegrationTests/Datadog.Trace.IntegrationTests.csproj
```diff
@@ -11,5 +11,23 @@
...
+  </ItemGroup>
+
+  <ItemGroup Condition=" $(TargetFramework.StartsWith('netcoreapp3')) ">
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="3.0.0" />
+    <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="3.0.0" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" $(TargetFramework.StartsWith('net5')) ">
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="5.0.0" />
+    <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="5.0.0" />
```
lucaspimentel, line 31: I'm not a fan of adding all these dependencies to `Datadog.Trace.IntegrationTests`, including `Microsoft.NET.Sdk.Web` and the `Samples.AspNetCoreRazorPages` project. Makes me want to move all these tests into a separate project. Maybe something we can come back and clean up later, no need to change now.

lucaspimentel (approved):

test/Datadog.Trace.IntegrationTests/Datadog.Trace.IntegrationTests.csproj
andrewlock: > I'm not a fan of adding all these dependencies

Any particular reason? This project has _very_ few tests currently (8 tests. 6 of which are skipped). The project reference isn't ideal, but it's basically the only way to test Razor Pages properly, short of spinning up things in IIS etc (which obviously doesn't work on Linux unless we switch to raw kestrel instead). Plus these tests are an order of magnitude faster than using IIS...

That said, we've already got 106 projects... what's 1 more 😬

test/Datadog.Trace.IntegrationTests/Program.cs
andrewlock: Unfortunately, yes. The build fails without it. I'm not entirely sure why, but likely todo with the `Microsoft.NET.Sdk.Web` SDK... 🤔

test/Datadog.Trace.IntegrationTests/Datadog.Trace.IntegrationTests.csproj
lucaspimentel: > Any particular reason?

I'm used to having one test assembly for every assembly (`{assembly}.Tests`) to keep inter-dependencies to a minimum:
- `Datadog.Trace` ➡ `Datadog.Trace.Tests`
- `Datadog.Trace.OpenTracing` ➡ `Datadog.Trace.OpenTracing.Tests`
- `Datadog.Trace.ClrProfiler.Managed` ➡️ `Datadog.Trace.ClrProfiler.Managed.Tests`

But I thought this was `Datadog.Trace.ClrProfiler.IntegrationTests`, so never mind. And maybe having so many individual test assemblies doesn't scale so well (106 projects!).

test/Datadog.Trace.IntegrationTests/Program.cs
lucaspimentel: Looks the web sdk has a different default output type. Adding `<OutputType>Library</OutputType>` to the project file should fix that.

andrewlock: Ah yes, cheers. Fixed in https://github.com/DataDog/dd-trace-dotnet/pull/1269/commits/70859ec8a053b969f710706f43e7691a86c82037

test/Datadog.Trace.IntegrationTests/Datadog.Trace.IntegrationTests.csproj
andrewlock: > I'm used to having one test assembly for every assembly

Me too, + optional integration test project, which is what we have 😄 The real problem with the number of projects is all the sample projects 🤷

PR #443, opened 2019-07-11
Conditional Sigil Nugets to fix conflict with Orleans

colin-higgins:
Changes proposed in this pull request:
Smoke test checking Orleans for crashing.
Conditionally use Sigil for net45 and Sigil-vNext for net61 and netstandard2.0

@DataDog/apm-dotnet

PR #1271, opened 2021-03-08
Delete OrleansCrash and remove it from our integration test suite

zacharycmontoya:
This sample was originally added in https://github.com/DataDog/dd-trace-dotnet/pull/443 to fix an issue with the Sigil library on net461/netstandard2.0. Now that Sigil has been removed, we no longer need this test case, especially because it is demonstrating some unreliable behavior.

@DataDog/apm-dotnet


lucaspimentel (approved):

andrewlock (approved):
Makes sense to remove this project if it's no longer testing anything useful, but I wonder if Orleans is something we should still be generally testing, given it uses runtime generated code, which could be an edge case?

tonyredondo (approved):
image
🥳

PR #684, opened 2020-03-26
Install ASP.NET HttpModule from MSI

zacharycmontoya:
Changes proposed in this pull request:
Edit the MSI to run an `install.cmd`/`uninstall.cmd` on installation and uninstallation, respectively.
- AppCmd.exe is used to add/remove the entry `<add name="DatadogTracingModule" type="Datadog.Trace.AspNet.TracingHttpModule,Datadog.Trace.AspNet,Version=1.0.0.0,Culture=neutral,PublicKeyToken=def86d061d0d2eeb" />` in the `system.webServer/modules` section of the global `applicationHost.config`.
- According to [the documentation](https://docs.microsoft.com/en-us/iis/get-started/getting-started-with-iis/getting-started-with-appcmdexe), AppCmd.exe can be used to manage IIS 7 and higher, which I think should be fine.

The new installation was tested in the following conditions to validate the HttpModule install/uninstall:
- Clean install
- Upgraded install from `1.11.0`
- Upgraded install from `1.12.0`
- Install the latest `1.15.0` official release as an upgrade from a test `1.14.1`-versioned MSI. This correctly triggers the uninstall script.
- Uninstall. This correctly triggers the uninstall script
- Installing the HttpModule before install or removing the HttpModule before uninstall causes an error with the AppCmd.exe command but we finish the installation successfully.
- Installing on a machine without IIS finishes causes an error with the AppCmd.exe command but we finish the installation successfully.

@DataDog/apm-dotnet

PR #1280, opened 2021-03-11
Modify the AspNet integration to be added automatic instrumentation instead of IIS config

zacharycmontoya:
## Background
This PR effectively reverts https://github.com/DataDog/dd-trace-dotnet/pull/684 so that the AspNet module is loaded into the instrumented application via automatic instrumentation instead of through the machine-wide IIS configuration. Unfortunately, I do not know the underlying motivation for that PR so I cannot predict potential issues with this change, however it fixes the following issues from the IIS configuration:

1. The IIS configuration only allows us to restrict the HttpModule to run on .NET Framework 4.0+ and does not allow us to restrict the usage to .NET Framework 4.5+, which is the actual requirement of the assembly. This currently causes a crash when a user attempts to install the .NET Tracer on an IIS host running .NET Framework 4.0 because the AspNet integration will still be loaded (the runtime met the condition of v4.0+) but it depends on libraries and types provided by the .NET Framework 4.5 runtime.
2. The fresh install process of the AAS extension currently requires a full stop because not stopping the application will cause a crash immediately after the installation finishes and the app recycle occurs. This occurs because the IIS configuration takes effect immediately, so the application to attempts to load the AspNet integration but it cannot find it. It's only after a full stop that automatic instrumentation takes effect and provides an AssemblyResolve event that the application will be able to find the AspNet integration.
3. When a prospective customer has another APM vendor installed on the machine, the customer will report our automatic instrumentation as partially working because they will still receive `aspnet.request` spans. The real issue is that the other APM vendor may still be registered as the CLR Profiler so ours cannot run. By moving the AspNet integration to also be triggered via automatic instrumentation, there will be no spans in this scenario and it will be much clearer to diagnose the issue.

## Changes
- Removes `install.cmd` and `uninstall.cmd` from being included and invoked in the MSI
- Adds CallSite and CallTarget integrations to register the HttpModule dynamically when we detect application startup via IIS or IIS Express

## Testing
- Upgrading scenarios tested:
  - Uninstall the previous MSI version and then install the new MSI version: Success
  - Have the previous MSI version installed and do an MSI upgrade: Success
  - Manually add the HttpModule in addition to the new MSI installation: Success
- End-to-end IIS scenarios validated, for both CallSite (actually InsertFirst) and CallTarget instrumentation:
  - .NET Framework
    - .NET Framework 4.5+: Creates  `aspnet.request` spans, 
    - .NET Framework 4.0: No longer crashes and automatic instrumentation is a no-op
  - .NET Core
    - In-Process: Does not create `aspnet.request` spans
    - Out-of-Process: Does not create `aspnet.request` spans
- Azure App Services scenarios validated, for both CallSite (actually InsertFirst) and CallTarget instrumentation:
  - .NET Framework
    -  Creates  `aspnet.request` spans
  - .NET Core
    - Does not create `aspnet.request` spans

@DataDog/apm-dotnet


src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AspNet/HttpModule_Integration.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            if (Interlocked.Exchange(ref _firstInitialization, 0) != 1)
+            {
+                // The HttpModule was already registered
+                return CallTargetState.GetDefault();
+            }
+
+            try
+            {
+                HttpApplication.RegisterModule(typeof(TracingHttpModule));
```
andrewlock, line 51: What would be the impact if tracer versions are out of sync (i.e. an older version of the msi with the module registered in config). Do we need to check if the module is already registered first?

If we try and register the module again, would we see two `aspnet.request` spans for every request, or it would it crash?

andrewlock (commented):
Looks like there's an issue with the CallSite implementation (constructor is missing `opCode`, `mdToken`, `moduleVersionPtr` parameters (see unit tests).

Just wondering, have you tested installing the MSI locally too? 

I assume we don't still need to try to _remove_ the module registration when the tracer MSI is upgraded? Presumably it is the previous version of the installer that runs during uninstall, which will remove the module correctly, so won't be a problem. Is there a scenario where it could still be registered (e.g. the manually added it, or didn't uninstall previous version of the tracer?) 🤔 

I'm also wondering - shouldn't this affect integration tests? Currently we are running samples using IIS express, without the module loaded, and asserting that we receive a single span. I would expect this integration to be loaded now as well and to return two spans per requests?

zacharycmontoya:
> Looks like there's an issue with the CallSite implementation (constructor is missing `opCode`, `mdToken`, `moduleVersionPtr` parameters (see unit tests).
> 
> Just wondering, have you tested installing the MSI locally too?

Yeah, this one uses a slightly different `InsertFirst` mechanism that we previously used for the AspNet loading. Essentially, instead of replacing the method call, we just add a new `call` instruction to the target integration method with no arguments and then it will return. This is why we don't need the other metadata for resolving target methods. I think that this method is just as resilient as our regular CallSite instrumentation, but maybe @tonyredondo could take a look and approve that.

> I assume we don't still need to try to _remove_ the module registration when the tracer MSI is upgraded? Presumably it is the previous version of the installer that runs during uninstall, which will remove the module correctly, so won't be a problem. Is there a scenario where it could still be registered (e.g. the manually added it, or didn't uninstall previous version of the tracer?) 🤔

Correct, I don't think we need to do anything in the new MSI steps because the previous version that ran the `install.cmd` script to add the HttpModule to the IIS config will also run an `uninstall.cmd` script when it's uninstalled. I should verify this though and I will document in the main PR summary the results.

> I'm also wondering - shouldn't this affect integration tests? Currently we are running samples using IIS express, without the module loaded, and asserting that we receive a single span. I would expect this integration to be loaded now as well and to return two spans per requests?

I completely forgot to look at the IIS Express tests we have and validate that (though the `integration-tests` pipeline is successful right now). I will investigate that.

And thank you for your thorough comments, I really appreciate it 😄

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AspNet/HttpModule_Integration.cs
zacharycmontoya: The HttpModule is permanently set (for 1.x versions of the Tracer) to 1.0.0. Perhaps it would more resilient to check if it's already been registered, but our module already has logic to only create one event handler (therefore one span) no matter how many times it's registered: https://github.com/DataDog/dd-trace-dotnet/blob/4541013aa89cc5fc2c8947fb03f28a5854409726/src/Datadog.Trace.AspNet/TracingHttpModule.cs#L62

zacharycmontoya: And yes I was able to verify that running the MSI and then also adding the HttpModule to the configuration results in only one `aspnet.request` span

andrewlock:
> Yeah, this one uses a slightly different InsertFirst mechanism that we previously used for the AspNet loading. Essentially, instead of replacing the method call via IL Rewriting, we just add a new call instruction as the first instruction to the instrumented method and have a contract where the integration takes no arguments and has a void return type. This is not really general purpose, but it works for this particular call and is only used when CallTarget is not enabled. I think that this method is just as resilient as our regular CallSite instrumentation, but maybe @tonyredondo could take a look and approve that.

I guess we need to update the unit tests to ignore this implementation difference then 🙂 

> Correct, I don't think we need to do anything in the new MSI steps. I should verify this though and I will document in the main PR summary the results.

I'm sure you're right, but that would be great!

> I completely forgot to look at the IIS Express tests we have and validate that (though the integration-tests pipeline is successful right now).

Yeah, this seems like a case of "why are my tests passing?!" 😆 If I understand correctly, this instrumentation _should_ be added automatically to IIS Express as well as IIS, so we should be seeing those spans 🤔

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetIntegration.cs
```diff
@@ -0,0 +1,53 @@
...
+            CallerAssembly = "System.Web",
+            CallerType = "System.Web.Compilation.BuildManager",
+            CallerMethod = "InvokePreStartInitMethodsCore",
+            TargetAssembly = "System.Web",
+            TargetType = "System.Web.Compilation.BuildManager",
+            TargetMethod = "InvokePreStartInitMethodsCore",
+            TargetMinimumVersion = "4",
+            TargetMaximumVersion = "4",
+            MethodReplacementAction = MethodReplacementActionType.InsertFirst)]
+        public static void TryLoadHttpModule()
```
tonyredondo, line 33: Looks good!

https://github.com/DataDog/dd-trace-dotnet/blob/master/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp#L1404-L1405

tonyredondo (approved):
LGTM

src/Datadog.Trace.Tools.Runner/Datadog.Trace.Tools.Runner.Tool.csproj
```diff
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
```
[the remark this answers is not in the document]
zacharycmontoya: Yeah something removed it. Anyways, for consistency, I removed the Unicode BOM on `src/Datadog.Trace.Tools.Runner/Datadog.Trace.Tools.Runner.proj` as well. Is that cool with you all? Do you have a preference with or without?

lucaspimentel (approved):

zacharycmontoya:
Update: I have done extensive local testing, but I am waiting until the AspNet integration tests are running in the `integration-tests` pipeline before merging this change. I'm making good progress on that front so that should be done soon.

zacharycmontoya:
> I'm also wondering - shouldn't this affect integration tests? Currently we are running samples using IIS express, without the module loaded, and asserting that we receive a single span. I would expect this integration to be loaded now as well and to return two spans per requests?

Update here, now that the AspNet tests are running in our integration tests, I modified the tests to expect two spans 👍🏼

PR #1288, opened 2021-03-17
Update ASP.NET / MVC / WebApi2 Resource Names

andrewlock:
Currently the resource names for our ASP.NET/WebAPI spans are inconsistently named, and do not make best use of routing information available. This PR makes the naming consistent across MVC/WebAPI2, and moves to using the route information where available.. These changes are currently opt-in, behind a feature flag.

## Current Resource Names

Conventional routed MVC ([e.g. `/Home/Get/0`](https://github.com/DataDog/dd-trace-dotnet/blob/master/test/test-applications/aspnet/Samples.AspNetMvc5/Controllers/HomeController.cs#L28) ) use URL obfuscation for both the "root IIS span" and the "child MVC span", and start with a `/`:

image

Attribute routed MVC ([e.g. `/delay/10`](https://github.com/DataDog/dd-trace-dotnet/blob/master/test/test-applications/aspnet/Samples.AspNetMvc5/Controllers/HomeController.cs#L33-L34)) uses URL obfuscation for the "root IIS span" and the _route_  for the "child MVC span", prefixing both with a `/`:

image

Conventional routed WebAPI2 ([e.g. `/api2/delay/3`](https://github.com/DataDog/dd-trace-dotnet/blob/master/test/test-applications/aspnet/Samples.AspNetMvc5/Controllers/ConventionsController.cs#L12)) uses URL obfuscation for the "root IIS span" and the _route_ for the "child MVC span". Only the root span is prefixed with a `/`:

image

Attribute routed WebApi2 ([e.g. `/api/delay/3`](https://github.com/DataDog/dd-trace-dotnet/blob/master/test/test-applications/aspnet/Samples.AspNetMvc5/Controllers/ApiController.cs#L13-L14)) is the same as the convential routed WebAPI 2: URL obfuscation for the "root IIS span" and the _route_ for the "child MVC span". Only the root span is prefixed with a `/`:

image

## New Unified Resource Names

If you enable the  feature flag, using `DD_TRACE_ASPNET_ROUTE_TEMPLATE_RESOURCE_NAMES_ENABLED=true`, then all these approaches give the same span pattern.

* Instead of URL obfuscation, the route template is used where available (i.e. everywhere except 404s)
* All resource names start with `/`
* Both spans have the same resource name
* Where a route is not found (i.e. it's a 404) we fallback to simple URL obfuscation to remove Ids and `Guid`s
* As with the existing WebApi integration, we replace the common tokens `area`, `controller` and `action` in route names. 
* For conventional routing, if a route has unused route parameters, these are removed from the resource name. This means that paths matching the common (default) route `{controller}/{action}/{id?}` (where `id` is optional), e.g. `/home/action` would have resource name `GET /home/action` instead of `GET /home/action/{id}`. 
* The `aspnet.route` tag is set both on the root span and the mvc span

Conventional routed MVC ([e.g. `/Home/Get/0`](https://github.com/DataDog/dd-trace-dotnet/blob/master/test/test-applications/aspnet/Samples.AspNetMvc5/Controllers/HomeController.cs#L28) )

image

Attribute routed MVC ([e.g. `/delay/10`](https://github.com/DataDog/dd-trace-dotnet/blob/master/test/test-applications/aspnet/Samples.AspNetMvc5/Controllers/HomeController.cs#L33-L34)) 

image

Conventional routed WebAPI2 ([e.g. `/api2/delay/3`](https://github.com/DataDog/dd-trace-dotnet/blob/master/test/test-applications/aspnet/Samples.AspNetMvc5/Controllers/ConventionsController.cs#L12))

image

Attribute routed WebApi2 ([e.g. `/api/delay/3`](https://github.com/DataDog/dd-trace-dotnet/blob/master/test/test-applications/aspnet/Samples.AspNetMvc5/Controllers/ApiController.cs#L13-L14))

image

Unknown routes with IDs are still obfuscated as before:

image

## Testing

As these changes could break customer metrics relying on the old behaviour, the new behaviour is behind an opt-in feature flag. Integration tests confirm both the new behaviour, and the old behaviour when the feature flag is disabled.

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace.AspNet/TracingHttpModule.cs
+++ b/src/Datadog.Trace.AspNet/TracingHttpModule.cs
@@ -123,12 +123,11 @@ namespace Datadog.Trace.AspNet
                 string host = httpRequest.Headers.Get("Host");
                 string httpMethod = httpRequest.HttpMethod.ToUpperInvariant();
                 string url = httpRequest.RawUrl.ToLowerInvariant();
-                string path = UriHelpers.GetRelativeUrl(httpRequest.Url, tryRemoveIds: true);
-                string resourceName = $"{httpMethod} {path.ToLowerInvariant()}";
 
                 var tags = new WebTags();
                 scope = tracer.StartActiveWithTags(_requestOperationName, propagatedContext, tags: tags);
-                scope.Span.DecorateWebServerSpan(resourceName, httpMethod, host, url, tags, tagsFromHeaders);
+                // Leave resourceName blank for now - we'll update it in OnEndRequest
+                scope.Span.DecorateWebServerSpan(resourceName: null, httpMethod, host, url, tags, tagsFromHeaders);
 
                 tags.SetAnalyticsSampleRate(IntegrationId, tracer.Settings, enabledWithGlobalSetting: true);
 
@@ -166,6 +165,11 @@ namespace Datadog.Trace.AspNet
                         && !string.IsNullOrEmpty(resourceName))
                     {
                         scope.Span.ResourceName = resourceName;
+                    }
+                    else
+                    {
+                        string path = UriHelpers.GetRelativeUrl(app.Request.Url, tryRemoveIds: true);
+                        scope.Span.ResourceName = $"{app.Request.HttpMethod.ToUpperInvariant()} {path.ToLowerInvariant()}";
                     }
 
                     scope.Dispose();
--- a/src/Datadog.Trace.AspNet/TracingHttpModule.cs
+++ b/src/Datadog.Trace.AspNet/TracingHttpModule.cs
@@ -161,6 +161,13 @@ namespace Datadog.Trace.AspNet
                     app.Context.Items[_httpContextScopeKey] is Scope scope)
                 {
                     scope.Span.SetHttpStatusCode(app.Context.Response.StatusCode, isServer: true);
+
+                    if (app.Context.Items["__Datadog.Trace.ClrProfiler.Managed.AspNetMvcIntegration-aspnet.resourcename"] is string resourceName
+                        && !string.IsNullOrEmpty(resourceName))
+                    {
+                        scope.Span.ResourceName = resourceName;
+                    }
+
                     scope.Dispose();
                 }
             }
--- a/src/Datadog.Trace.AspNet/TracingHttpModule.cs
+++ b/src/Datadog.Trace.AspNet/TracingHttpModule.cs
@@ -168,7 +168,7 @@ namespace Datadog.Trace.AspNet
                     }
                     else
                     {
-                        string path = UriHelpers.GetRelativeUrl(app.Request.Url, tryRemoveIds: true);
+                        string path = UriHelpers.GetCleanUriPath(app.Request.Url);
                         scope.Span.ResourceName = $"{app.Request.HttpMethod.ToUpperInvariant()} {path.ToLowerInvariant()}";
                     }
 
--- a/src/Datadog.Trace.AspNet/TracingHttpModule.cs
+++ b/src/Datadog.Trace.AspNet/TracingHttpModule.cs
@@ -161,7 +161,7 @@ namespace Datadog.Trace.AspNet
                 {
                     scope.Span.SetHttpStatusCode(app.Context.Response.StatusCode, isServer: true);
 
-                    if (app.Context.Items["__Datadog.Trace.ClrProfiler.Managed.AspNetMvcIntegration-aspnet.resourcename"] is string resourceName
+                    if (app.Context.Items[SharedConstants.HttpContextPropagatedResourceNameKey] is string resourceName
                         && !string.IsNullOrEmpty(resourceName))
                     {
                         scope.Span.ResourceName = resourceName;
--- a/src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetMvcIntegration.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetMvcIntegration.cs
@@ -90,21 +90,34 @@ namespace Datadog.Trace.ClrProfiler.Integrations
                     }
                 }
 
+                string routeUrl = route?.Url;
+                string areaName = (routeValues?.GetValueOrDefault("area") as string)?.ToLowerInvariant();
+                string controllerName = (routeValues?.GetValueOrDefault("controller") as string)?.ToLowerInvariant();
+                string actionName = (routeValues?.GetValueOrDefault("action") as string)?.ToLowerInvariant();
+
+                if (string.IsNullOrEmpty(resourceName) && !string.IsNullOrEmpty(routeUrl))
+                {
+                    resourceName = $"{httpMethod} /{routeUrl.ToLowerInvariant()}";
+                }
+
                 if (string.IsNullOrEmpty(resourceName) && httpContext.Request.Url != null)
                 {
                     var cleanUri = UriHelpers.GetRelativeUrl(httpContext.Request.Url, tryRemoveIds: true);
                     resourceName = $"{httpMethod} {cleanUri.ToLowerInvariant()}";
                 }
 
-                string areaName = (routeValues?.GetValueOrDefault("area") as string)?.ToLowerInvariant();
-                string controllerName = (routeValues?.GetValueOrDefault("controller") as string)?.ToLowerInvariant();
-                string actionName = (routeValues?.GetValueOrDefault("action") as string)?.ToLowerInvariant();
-
                 if (string.IsNullOrEmpty(resourceName))
                 {
                     // Keep the legacy resource name, just to have something
                     resourceName = $"{httpMethod} {controllerName}.{actionName}";
                 }
+
+                // Fail safe to catch templates in routing values
+                resourceName =
+                    resourceName
+                       .Replace("{area}", areaName)
+                       .Replace("{controller}", controllerName)
+                       .Replace("{action}", actionName);
 
                 SpanContext propagatedContext = null;
                 var tracer = Tracer.Instance;
@@ -129,13 +142,6 @@ namespace Datadog.Trace.ClrProfiler.Integrations
                 scope = Tracer.Instance.StartActiveWithTags(OperationName, propagatedContext, tags: tags);
                 Span span = scope.Span;
 
-                // Fail safe to catch templates in routing values
-                resourceName =
-                    resourceName
-                       .Replace("{area}", areaName)
-                       .Replace("{controller}", controllerName)
-                       .Replace("{action}", actionName);
-
                 span.DecorateWebServerSpan(
                     resourceName: resourceName,
                     method: httpMethod,
@@ -144,7 +150,7 @@ namespace Datadog.Trace.ClrProfiler.Integrations
                     tags,
                     tagsFromHeaders);
 
-                tags.AspNetRoute = route?.Url;
+                tags.AspNetRoute = routeUrl;
                 tags.AspNetArea = areaName;
                 tags.AspNetController = controllerName;
                 tags.AspNetAction = actionName;
--- a/src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetMvcIntegration.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetMvcIntegration.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Web;
 using System.Web.Routing;
+using Datadog.Trace.AspNet;
 using Datadog.Trace.ClrProfiler.Emit;
 using Datadog.Trace.ClrProfiler.Integrations.AspNet;
 using Datadog.Trace.Configuration;
@@ -180,7 +181,7 @@ namespace Datadog.Trace.ClrProfiler.Integrations
                 if (newResourceNamesEnabled)
                 {
                     // set the resource name in the HttpContext so TracingHttpModule can update root span
-                    httpContext.Items["__Datadog.Trace.ClrProfiler.Managed.AspNetMvcIntegration-aspnet.resourcename"] = resourceName;
+                    httpContext.Items[SharedConstants.HttpContextPropagatedResourceNameKey] = resourceName;
                 }
             }
             catch (Exception ex)
--- a/src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetMvcIntegration.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetMvcIntegration.cs
@@ -65,6 +65,7 @@ namespace Datadog.Trace.ClrProfiler.Integrations
                 RouteData routeData = controllerContext.RouteData;
                 Route route = routeData?.Route as Route;
                 RouteValueDictionary routeValues = routeData?.Values;
+                bool wasAttributeRouted = false;
 
                 if (route == null && routeData?.Route.GetType().FullName == RouteCollectionRouteTypeName)
                 {
@@ -74,6 +75,7 @@ namespace Datadog.Trace.ClrProfiler.Integrations
                     {
                         // route was defined using attribute routing i.e. [Route("/path/{id}")]
                         // get route and routeValues from the RouteData in routeMatches
+                        wasAttributeRouted = true;
                         route = routeMatches[0].Route as Route;
                         routeValues = routeMatches[0].Values;
 
@@ -118,6 +120,23 @@ namespace Datadog.Trace.ClrProfiler.Integrations
                        .Replace("{area}", areaName)
                        .Replace("{controller}", controllerName)
                        .Replace("{action}", actionName);
+
+                if (!wasAttributeRouted && routeValues is not null && route is not null)
+                {
+                    // Remove unused parameters from conventional route templates
+                    // Don't bother with routes defined using attribute routing
+                    foreach (var parameter in route.Defaults)
+                    {
+                        var parameterName = parameter.Key;
+                        if (parameterName != "area"
+                            && parameterName != "controller"
+                            && parameterName != "action"
+                            && !routeValues.ContainsKey(parameterName))
+                        {
+                            resourceName = resourceName.Replace($"/{{{parameterName}}}", string.Empty);
+                        }
+                    }
+                }
 
                 SpanContext propagatedContext = null;
                 var tracer = Tracer.Instance;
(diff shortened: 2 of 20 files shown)


lucaspimentel (commented):
I'm not done yet. Gotta come back and finish later.

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetMvcIntegration.cs
```diff
@@ -90,22 +93,52 @@ public static Scope CreateScope(ControllerContextStruct controllerContext)
...
+                resourceName =
+                    resourceName
+                       .Replace("{area}", areaName)
+                       .Replace("{controller}", controllerName)
+                       .Replace("{action}", actionName);
+
+                if (newResourceNamesEnabled && !wasAttributeRouted && routeValues is not null && route is not null)
+                {
+                    // Remove unused parameters from conventional route templates
+                    // Don't bother with routes defined using attribute routing
```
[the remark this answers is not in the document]
andrewlock: IMO, it's a difference in philosophy between the two routing types (and will have a slight performance benefit to not doing the extra work). 

Conventional routes (in particular the _default_ route) often have a placeholder `{id?}` (or similar) parameter, as they're trying to match a single route to multiple actions. So displaying `/home/index/{id}` wouldn't make any sense when the URL is `/home/index` and the  action is `Index()`.

In contrast, attribute routes are 1-1 tied to an action. So stripping off the (in my experience, rare) optional parameter would give the appearance of two different endpoints when they aren't,  so doesn't make sense to me I think. For example, given the action:

```csharp
[HttpGet, Route("api/optional-delay/{value}")]
public IActionResult OptionalDelay(int value = 5) { /* */}
```

I think it's more beneficial to have `/api/optional-delay` and `/api/optional-delay/5` both use the route `api/optional-delay/{value}`, instead of two separate routes for the same endpoint. As I say, the nature of the 1-1 mapping means optional parameters are pretty rare in these cases anyway.

Now, you could make an argument that for the conventional routing, if the user calls `/home/index/123`, we will use the route `/home/index/{id}` and they will still be routed to the same `Index()` action. In that case we have two resources with different resource name routes (one with `{id}` and one without), but in my experience this is _very_ rarely used in conventional routing, and more often will point to a bug. And the alternative is never stripping the optional routes, will just look weird most of the time. 

`</essay>`

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetMvcIntegration.cs
```diff
@@ -90,22 +93,52 @@ public static Scope CreateScope(ControllerContextStruct controllerContext)
...
-                string controllerName = (routeValues?.GetValueOrDefault("controller") as string)?.ToLowerInvariant();
-                string actionName = (routeValues?.GetValueOrDefault("action") as string)?.ToLowerInvariant();
-
                 if (string.IsNullOrEmpty(resourceName))
                 {
                     // Keep the legacy resource name, just to have something
                     resourceName = $"{httpMethod} {controllerName}.{actionName}";
                 }
 
+                // Fail safe to catch templates in routing values
```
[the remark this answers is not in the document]
andrewlock: Done in https://github.com/DataDog/dd-trace-dotnet/pull/1288/commits/04fcf807cd12bc7370c81710a8629c5071936d44

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetWebApi2Integration.cs
```diff
@@ -270,17 +271,21 @@ internal static void UpdateSpan(IHttpControllerContext controllerContext, Span s
...
+                    resourceName = $"{method} {(newResourceNamesEnabled ? "/" : string.Empty)}{route.ToLowerInvariant()}";
                 }
                 else if (requestUri != null)
                 {
                     var cleanUri = UriHelpers.GetRelativeUrl(requestUri, tryRemoveIds: true);
                     resourceName = $"{method} {cleanUri.ToLowerInvariant()}";
                 }
+                else
+                {
+                    resourceName = $"{method} {absoluteUri.ToLowerInvariant()}";
```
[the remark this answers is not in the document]
andrewlock: Haha, well it was a refactoring of the previous code (moving initialisation of `resourceName` to the `else` branch), but you're right, looks like that will never be called 😬 I've removed the `absoluteUri` entirely in https://github.com/DataDog/dd-trace-dotnet/pull/1288/commits/6e94a7a9f656ec887efc8dd985d031660ad2b76c, but we could probably just exit early if `RequestUri` is `null`... though this is a bit safer if there's weirdness

src/Datadog.Trace.AspNet/TracingHttpModule.cs
```diff
@@ -161,6 +160,18 @@ private void OnEndRequest(object sender, EventArgs eventArgs)
...
                     scope.Span.SetHttpStatusCode(app.Context.Response.StatusCode, isServer: true);
+
+                    if (app.Context.Items["__Datadog.Trace.ClrProfiler.Managed.AspNetMvcIntegration-aspnet.resourcename"] is string resourceName
+                        && !string.IsNullOrEmpty(resourceName))
+                    {
+                        scope.Span.ResourceName = resourceName;
+                    }
+                    else
+                    {
+                        string path = UriHelpers.GetRelativeUrl(app.Request.Url, tryRemoveIds: true);
```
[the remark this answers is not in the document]
andrewlock: Although...... this is an `internal` class... will that cause breaking change issues? 😬

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetMvcIntegration.cs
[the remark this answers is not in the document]
lucaspimentel: Your logic makes sense to me. Thanks for the essay 😉

src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -180,6 +180,8 @@ public TracerSettings(IConfigurationSource source)
 
             TraceBatchInterval = source?.GetInt32(ConfigurationKeys.SerializationBatchInterval)
                         ?? 100;
+
+            AspnetRouteTemplateResourceNamesEnabled = IsFeatureFlagEnabled(ConfigurationKeys.FeatureFlags.AspnetRouteTemplateResourceNamesEnabled);
```
lucaspimentel, line 184: Most tracer settings can be set in several ways: environment variables, `app.config`/`web.config` file, or in a json file (`~/datadog.json` by default). Most exceptions to this rule are settings used in native code (env vars only).

Currently, `DD_TRACE_ASPNET_ROUTE_TEMPLATE_RESOURCE_NAMES_ENABLED` will only work as an env var. Was this limitation intentional?

src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -488,7 +496,12 @@ internal bool IsIntegrationEnabled(string integrationName)
 
         internal bool IsNetStandardFeatureFlagEnabled()
         {
-            var value = EnvironmentHelpers.GetEnvironmentVariable("DD_TRACE_NETSTANDARD_ENABLED", string.Empty);
+            return IsFeatureFlagEnabled(ConfigurationKeys.FeatureFlags.NetStandardEnabled);
+        }
+
+        internal bool IsFeatureFlagEnabled(string featureFlag)
+        {
+            var value = EnvironmentHelpers.GetEnvironmentVariable(featureFlag, string.Empty);
```
lucaspimentel, line 504: For all boolean settings, we should use the `string.ToBoolean()` extension method:
https://github.com/DataDog/dd-trace-dotnet/blob/d9f44e4ff727dcd6693b887fcc05a7fbd2eb70fc/src/Datadog.Trace/ExtensionMethods/StringExtensions.cs#L30

`DD_TRACE_NETSTANDARD_ENABLED` is a special case. The setting is only used in native code. `IsNetStandardFeatureFlagEnabled()` only checks for `"1"` and `"true"` to match the logic in C++ and it is only used for logging purposes:
https://github.com/DataDog/dd-trace-dotnet/blob/d9f44e4ff727dcd6693b887fcc05a7fbd2eb70fc/src/Datadog.Trace/Tracer.cs#L538

src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -338,5 +338,19 @@ internal static class Debug
...
+        internal static class FeatureFlags
+        {
+            /// <summary>
+            /// Feature Flag for whether to enable the updated ASP.NET resource names
+            /// </summary>
+            /// <seealso cref="TracerSettings.AspnetRouteTemplateResourceNamesEnabled"/>
+            public const string AspnetRouteTemplateResourceNamesEnabled = "DD_TRACE_ASPNET_ROUTE_TEMPLATE_RESOURCE_NAMES_ENABLED";
+
+            /// <summary>
+            /// Feature Flag for whether .NET Standard mode is enabled
```
lucaspimentel, line 351: "What is .NET Standard mode?"
```suggestion
            /// Feature Flag: enables instrumenting calls to netstandard.dll (only applies to CallSite instrumentation)
```

lucaspimentel (commented):
Finished my review. Mostly just questions about the new feature flag.

src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -338,5 +338,19 @@ internal static class Debug
             /// </summary>
             public const string ForceFallbackLookup = "DD_TRACE_DEBUG_LOOKUP_FALLBACK";
         }
+
+        internal static class FeatureFlags
+        {
+            /// <summary>
+            /// Feature Flag for whether to enable the updated ASP.NET resource names
```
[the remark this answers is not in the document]
andrewlock: Well, this _does_ only cover legacy ASP.NET (and Web API 2). I was going to introduce a separate (similar) flag for aspnet core, as that will have even more extensive changes. 

Do you think we should only have a single flag for both aspnet and aspnetcore?

src/Datadog.Trace/Util/UriHelpers.cs
```diff
@@ -21,15 +21,9 @@ public static string CleanUri(Uri uri, bool removeScheme, bool tryRemoveIds)
...
+        public static string GetCleanUriPath(Uri uri)
         {
-            return GetRelativeUrl(uri.AbsolutePath, tryRemoveIds);
-        }
-
-        public static string GetRelativeUrl(string uri, bool tryRemoveIds)
-        {
-            // try to remove segments that look like ids
-            return tryRemoveIds ? CleanUriSegment(uri) : uri;
+            return CleanUriSegment(uri.AbsolutePath);
```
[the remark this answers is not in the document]
andrewlock: I'm on board 👍 Again, assuming changing these `internal` APIs won't break anything 😬 Fixed in https://github.com/DataDog/dd-trace-dotnet/pull/1288/commits/7a4d0ebc3f010ce8940737cdff751a7c6e8498f3

src/Datadog.Trace/Configuration/TracerSettings.cs
andrewlock: > `DD_TRACE_NETSTANDARD_ENABLED` is a special case. The setting is only used in native code.

Ah, good to know, thanks! Fixed in https://github.com/DataDog/dd-trace-dotnet/pull/1288/commits/4b0e7b1ac49093fe57bdd001c90d2bb783dca545

src/Datadog.Trace/Configuration/TracerSettings.cs
andrewlock: Oops, no, I thought I'd reverted that change (turns out that was my other PR https://github.com/DataDog/dd-trace-dotnet/pull/1289). Fixed in https://github.com/DataDog/dd-trace-dotnet/pull/1288/commits/4b0e7b1ac49093fe57bdd001c90d2bb783dca545

src/Datadog.Trace/Util/UriHelpers.cs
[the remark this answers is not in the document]
lucaspimentel: > changing these internal APIs won't break anything 

Good catch! I didn't think of this before. If these methods are called from `Datadog.Trace.ClrProfiler.Managed`, this and previous renames are all breaking changes. When combining different versions of the msi and nuget packages, users can end up with mismatching `Datadog.Trace` and `Datadog.Trace.ClrProfiler.Managed`. We should either revert the renames (sorry!) or keep the old ones around as backwards compatibility shims.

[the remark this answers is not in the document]
andrewlock: > We should either revert the renames (sorry!) or keep the old ones around as backwards compatibility shims.

I'll go with option b. At least that makes it obvious that we can remove them once we hit the mystical v2.0

src/Datadog.Trace.AspNet/TracingHttpModule.cs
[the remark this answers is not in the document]
lucaspimentel: _sigh_ yes, my bad. See my comment here: https://github.com/DataDog/dd-trace-dotnet/pull/1288#discussion_r597689433

src/Datadog.Trace/Util/UriHelpers.cs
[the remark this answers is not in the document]
andrewlock: Reverted in https://github.com/DataDog/dd-trace-dotnet/pull/1288/commits/4b33edb1098b25248368247baa001a9135a63c85

src/Datadog.Trace.AspNet/TracingHttpModule.cs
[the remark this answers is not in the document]
andrewlock: reverted in https://github.com/DataDog/dd-trace-dotnet/pull/1288/commits/4b33edb1098b25248368247baa001a9135a63c85

src/Datadog.Trace/Configuration/ConfigurationKeys.cs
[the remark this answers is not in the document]
lucaspimentel: My bad. I saw `AspNetCoreDiagnosticObserver.cs` listed in the changed files and assumed there were real changes there. In that case, can we list the affected integrations or spans? When I read "ASP.NET spans" I assume it means only the http module and not MVC or Web API 😅 

```suggestion
            /// Feature Flag: enables updated resource names on `aspnet.request`, `aspnet-mvc.request`, and `aspnet-webapi.request` spans
```

src/Datadog.Trace/Util/UriHelpers.cs
```diff
@@ -21,18 +21,27 @@ public static string CleanUri(Uri uri, bool removeScheme, bool tryRemoveIds)
             return $"{uri.Scheme}{Uri.SchemeDelimiter}{uri.Authority}{path}";
         }
 
+        [Obsolete("This method is deprecated and will be removed. Use GetCleanUriPath() instead. " +
+                  "Kept for backwards compatability where there is a version mismatch between manual and automatic instrumentation")]
         public static string GetRelativeUrl(Uri uri, bool tryRemoveIds)
-        {
-            return GetRelativeUrl(uri.AbsolutePath, tryRemoveIds);
-        }
+            => GetRelativeUrl(uri.AbsolutePath, tryRemoveIds);
```
lucaspimentel, line 27: Should we call `GetCleanUriPath()` to avoid the warning about calling an obsolete method?
```suggestion
            => tryRemoveIds ? GetCleanUriPath(uri.AbsolutePath) : uri.AbsolutePath;
```

andrewlock: There is no warning because the method itself is `[Obsolete]` right?

lucaspimentel: TIL. Never mind!

lucaspimentel (approved):

src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -382,6 +385,12 @@ public bool DiagnosticSourceEnabled
         /// </summary>
         internal int TraceBatchInterval { get; }
 
+        /// <summary>
+        /// Gets a value indicating whether the feature flag to enable the updated ASP.NET resource names is enabled
+        /// </summary>
+        /// <seealso cref="ConfigurationKeys.HttpClientErrorStatusCodes"/>
```
zacharycmontoya, line 391:
```suggestion
        /// <seealso cref="ConfigurationKeys.AspnetRouteTemplateResourceNamesEnabled"/>
```

zacharycmontoya (approved):

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetMvcIntegration.cs
```diff
@@ -90,22 +93,52 @@ public static Scope CreateScope(ControllerContextStruct controllerContext)
                     }
                 }
 
+                string routeUrl = route?.Url;
+                string areaName = (routeValues?.GetValueOrDefault("area") as string)?.ToLowerInvariant();
+                string controllerName = (routeValues?.GetValueOrDefault("controller") as string)?.ToLowerInvariant();
+                string actionName = (routeValues?.GetValueOrDefault("action") as string)?.ToLowerInvariant();
```
tonyredondo, line 99: Is the `GetValueOrDefault` a method of `routeValues` or is using an extension method with Sigil?

src/Datadog.Trace.AspNet/TracingHttpModule.cs
```diff
@@ -161,6 +160,18 @@ private void OnEndRequest(object sender, EventArgs eventArgs)
                     app.Context.Items[_httpContextScopeKey] is Scope scope)
                 {
                     scope.Span.SetHttpStatusCode(app.Context.Response.StatusCode, isServer: true);
+
+                    if (app.Context.Items["__Datadog.Trace.ClrProfiler.Managed.AspNetMvcIntegration-aspnet.resourcename"] is string resourceName
```
tonyredondo, line 164: nit: can we declare the key as a constant somewhere else?

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetMvcIntegration.cs
andrewlock: `GetValueOrDefault()` is defined in the library here: https://github.com/DataDog/dd-trace-dotnet/blob/e51c8d2cfd396df40d363d14e61fe93cee3b0f14/src/Datadog.Trace/ExtensionMethods/DictionaryExtensions.cs#L10

Is that a problem? The method will only be used in newer versions of the tracer so I assume it's fine, or is this another case where mismatched auto/manual versions will cause issues?

src/Datadog.Trace.AspNet/TracingHttpModule.cs
andrewlock: done in https://github.com/DataDog/dd-trace-dotnet/pull/1288/commits/461b9a1ddf7fe83dd31bcd399a4721a42bf83391

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetMvcIntegration.cs
tonyredondo: Is a normal dictionary, no problem at all, I haven't use that extension method that's why I asked 🙂 Thanks!

PR #1289, opened 2021-03-17
Update ASP.NET Core Resource Names

andrewlock:
Currently the resource names for our ASP.NET Core spans are inconsistently named, and differ significantly from the ASP.NET/WebAPI resource names ([as discussed in this PR](https://github.com/DataDog/dd-trace-dotnet/pull/1288)). This PR makes the naming consistent across conventional/attribute routing, adds an additional span specifically for the MVC part of the request, and moves to using the route information where available. It also adds new tags for ASP.NET Core spans These changes are currently opt-in, behind a feature flag.

## Current Resource Names

Razor Pages and WebAPI/attribute routes use the route name as the resource name, without a `/` prefix:

image

image

For conventional routes, the pattern `{controller}/{action}` is used, without a `/` prefix:

image

For attribute routes, the route is used, without a `/` prefix:

image

For Unknown URLs, the URL is obfuscated

image

## New Unified Resource Names

If you enable the feature flag, using `DD_TRACE_ASPNETCORE_ROUTE_TEMPLATE_RESOURCE_NAMES_ENABLED=true`, then all these approaches give the same span pattern

* We use two spans, a "root" span `aspnet_core.request` for the overall request pipeline, and a "child" span `aspnet_core.mvc` for the MVC part only. If there is no MVC execution, the second span will be absent
* The root span uses the route information (where available, i.e. not a 404) to set the resource name, including a `/` prefix
* As with the existing WebApi integration, we replace the common tokens `area`, `controller` and `action` in route names.
* MVC spans may have the optional tags `aspnet_core.action`, `area`, `controller`, `route`, `page`. 
* The root span has the `aspnet_core.route` tag and for .NET Core 3+ (when endpoint routing is enabled) will have the `endpoint` tag too. 
* For conventional routing, optional route parameters are removed when not provided.
* We can now handle multiple executions of the pipeline. This is common with exception handlers that re-execute the pipeline to generate a response on error, or for 404s for example. In this case, you may have up to 3 spans - 1 root for the "overall" request, and one for each of the MVC executions. 
* We explicitly test and handle non-MVC endpoints in aspnetcore 3+. We can't instrument the endpoints themselves, but we can extract the routes and attach those to the root (only) span

Razor Pages:

image

image

For conventional routes:

image

For attribute routes:

image

For Unknown URLs, the URL is obfuscated

image

Error hander re-executing the pipeline may have 3 spans (depending on pipeline configuration)

image

## Testing

As these changes could break customer metrics relying on the old behaviour, the new behaviour is behind an opt-in feature flag. Integration tests confirm both the new behaviour, and the old behaviour when the feature flag is disabled.


test/Datadog.Trace.IntegrationTests/DiagnosticListeners/AspNetCoreMvcTestData.cs
```diff
@@ -35,82 +35,140 @@ public static class AspNetCoreMvcTestData
...
+            { $"{ExceptionPagePrefix}/throws", 500, true, $"GET {ExceptionPagePrefix}/throws", EmptyTags(), 1, null, null, null, null },
+            { $"{ReExecuteHandlerPrefix}/Home/Error", 500, true, $"GET {ReExecuteHandlerPrefix}/home/error", ConventionalParentTags(), 3, null, ConventionalChildTags(action: "error"), $"GET {ReExecuteHandlerPrefix}/home/index", ConventionalChildTags() },
+            { $"{ReExecuteHandlerPrefix}/Home/BadHttpRequest", 500, true, $"GET {ReExecuteHandlerPrefix}/home/badhttprequest", ConventionalParentTags(), 3, null, ConventionalChildTags(action: "badhttprequest"), $"GET {ReExecuteHandlerPrefix}/home/index", ConventionalChildTags() },
+            { $"{ReExecuteHandlerPrefix}/throws", 500, true, $"GET {ReExecuteHandlerPrefix}/throws", EmptyTags(), 2, $"GET {ReExecuteHandlerPrefix}/home/index", ConventionalChildTags(), null, null },
+            { $"{StatusCodeReExecutePrefix}/I/dont/123/exist/", 404, false, $"GET {StatusCodeReExecutePrefix}/i/dont/?/exist/", EmptyTags(), 2, $"GET {StatusCodeReExecutePrefix}/home/index", ConventionalChildTags(), null, null },
```
andrewlock, line 77: I think this could really benefit from some "snapshot" style testing for the spans, but I'll tidy that up in a subsequent PR

lucaspimentel:
`DD_TRACE_ASPNET_ROUTE_TEMPLATE_RESOURCE_NAMES_ENABLED` and `DD_TRACE_ASPNETCORE_ROUTE_TEMPLATE_RESOURCE_NAMES_ENABLED` (from #1288) are quite the mouthful and seem redundant to me. What do you think of consolidating both into a shorter name? The setting from the other PR already spans multiple integration, so why not group them all? Maybe something like `DD_TRACE_ROUTE_RESOURCE_ENABLED`? No, that's a terrible name. `DD_TRACE_ROUTE_TEMPLATE_RESOURCE_NAMES_ENABLED`?

andrewlock:
> `DD_TRACE_ASPNET_ROUTE_TEMPLATE_RESOURCE_NAMES_ENABLED` and `DD_TRACE_ASPNETCORE_ROUTE_TEMPLATE_RESOURCE_NAMES_ENABLED` (from #1288) are quite the mouthful and seem redundant to me. What do you think of consolidating both into a shorter name? The setting from the other PR already spans multiple integration, so why not group them all? Maybe something like `DD_TRACE_ROUTE_RESOURCE_ENABLED`? No, that's a terrible name. `DD_TRACE_ROUTE_TEMPLATE_RESOURCE_NAMES_ENABLED`?

I know, they're ridiculously long😬 My reason for adding both was that I wasn't sure whether anyone would want to enable one without the other. For example, we have a customer that needs the aspnet route names, but they haven't mentioned the aspnetcore resource names, even though they're using it. If we group them, it's all or nothing.

`DD_TRACE_ROUTE_TEMPLATE_RESOURCE_NAMES_ENABLED` is fine by me, though it's not much shorter than the existing ones 😉

zacharycmontoya:
> My reason for adding both was that I wasn't sure whether anyone would want to enable one without the other. For example, we have a customer that needs the aspnet route names, but they haven't mentioned the aspnetcore resource names, even though they're using it. If we group them, it's all or nothing.

FWIW, my vote would be to unify them down to one feature, especially as the intent is to introduce consistency everywhere. If a customer would like one set of changes versus the other, we can see if they can set this on a per-application basis (web.config or datadog.json)

src/Datadog.Trace/DiagnosticListeners/IEndpointFeature.cs
```diff
@@ -0,0 +1,16 @@
...
+    /// <summary>
+    /// IEndpointFeature for ducktyping HttpContext
+    /// </summary>
+    public interface IEndpointFeature
+    {
+        /// <summary>
+        /// Delegates to IEndpointFeature.Endpoint;
+        /// </summary>
+        [Duck(Name = "Microsoft.AspNetCore.Http.Features.IEndpointFeature.get_Endpoint")]
+        RouteEndpoint GetEndpoint();
```
zacharycmontoya, line 14: Is there a reason you're proxying the underlying method instead of proxying the property itself?

test/Datadog.Trace.IntegrationTests/Datadog.Trace.IntegrationTests.csproj
```diff
@@ -34,4 +34,8 @@
     <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="5.0.0" />
     <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="5.0.0" />
   </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="FluentAssertions" Version="5.10.3" />
```
zacharycmontoya, line 39: Thanks for adding this 👍🏼 This should make our lives easier

src/Datadog.Trace/DiagnosticListeners/IEndpointFeature.cs
andrewlock: Yes! 🙂 

The default feature collection implementation used by Kestrel [explicitly implements the interface here](https://github.com/dotnet/aspnetcore/blob/28a6f89eb5ca73c631bda39f8b8d690cc2c7eeb4/src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs#L21) . Unfortunately ducktyping can't find this method, even with updated binding flags, the only way I could see to make it work was to get the method, as in this case.

Note that this is implementation specific; the `TestServer` implementation for example _can_ proxy the property directly, hence the additional `EndpointFeatureStruct` definition: https://github.com/DataDog/dd-trace-dotnet/blob/ba9e30544e280200fc61091f3f577819950d0c45/src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs#L882-L886

test/Datadog.Trace.IntegrationTests/Datadog.Trace.IntegrationTests.csproj
andrewlock: I actually didn't mean to include this in the final PR 😂 But @kevingosse has added it to all projects in #1303, so will remove it from here if he merges first 🙂

andrewlock:
> > My reason for adding both was that I wasn't sure whether anyone would want to enable one without the other. For example, we have a customer that needs the aspnet route names, but they haven't mentioned the aspnetcore resource names, even though they're using it. If we group them, it's all or nothing.
> 
> FWIW, my vote would be to unify them down to one feature, especially as the intent is to introduce consistency everywhere. If a customer would like one set of changes versus the other, we can see if they can set this on a per-application basis (web.config or datadog.json)

That makes sense. I've unified the feature flag here and in #1288 to the suggested `DD_TRACE_ROUTE_TEMPLATE_RESOURCE_NAMES_ENABLED` value

lucaspimentel (commented):
Just noticed I have pending comments. Partial review.

src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
```diff
@@ -255,6 +316,169 @@ private static SpanContext ExtractPropagatedContext(HttpRequest request)
...
+
+                            sb.Append('}');
+                        }
+                    }
+                }
+            }
+
+            var simplifiedRoute = StringBuilderCache.GetStringAndRelease(sb);
+
+            return string.IsNullOrEmpty(simplifiedRoute) ? "/" : simplifiedRoute;
```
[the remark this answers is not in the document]
andrewlock: Yeah, I think it probably makes sense to just do it at the end instead, instead of the individual parts.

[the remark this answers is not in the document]
andrewlock: done in https://github.com/DataDog/dd-trace-dotnet/pull/1289/commits/d058e2c5e2205523566a07d3f5a24a1bab1a0b28

src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
```diff
@@ -255,6 +316,169 @@ private static SpanContext ExtractPropagatedContext(HttpRequest request)
...
+            foreach (var pathSegment in routePattern.Segments)
+            {
+                foreach (var part in pathSegment.Parts)
+                {
+                    var partName = part.Name;
+
+                    if (!part.IsParameter)
+                    {
+                        sb.Append('/');
+                        sb.Append(part.Text?.ToLowerInvariant());
```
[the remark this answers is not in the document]
andrewlock: Done in https://github.com/DataDog/dd-trace-dotnet/pull/1289/commits/d058e2c5e2205523566a07d3f5a24a1bab1a0b28

test/Datadog.Trace.IntegrationTests/DiagnosticListeners/AspNetCoreMvcTestData.cs
lucaspimentel: You may want to speak with @Kyle-Verhoog. He was prototyping a tracer-agnostic tool to record and compare trace snapshots. He showed it during one of our team demos. Cool stuff.

src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
```diff
@@ -29,18 +33,25 @@ internal sealed class AspNetCoreDiagnosticObserver : DiagnosticObserver
 
         private const string DiagnosticListenerName = "Microsoft.AspNetCore";
         private const string HttpRequestInOperationName = "aspnet_core.request";
+        private const string MvcOperationName = "aspnet_core.mvc";
```
lucaspimentel, line 36: I think we want something like `aspnet_core_mvc.request` here. I think the `.request` suffix is a convention on all incoming http requests.

lucaspimentel: Separate issue maybe for a separate PR: some operation names have hyphens (`aspnet-mvc.request`) and others underscores (`aspnet_core.request`), but I think they all normalized into one or the other, so it's probably safe to fix them in code without breaking changes.

test/Datadog.Trace.IntegrationTests/Datadog.Trace.IntegrationTests.csproj
lucaspimentel: We should add `FluentAssertions` to `LICENSE-3rdparty.csv`.

lucaspimentel: Actually, we don't distribute it, so I'm not sure about that.

test/Datadog.Trace.IntegrationTests/DiagnosticListeners/AspNetCoreMvcTestData.cs
Kyle-Verhoog: :+1: we use them in Python and they've helped catch a couple of regressions already! I'm happy to help get the test agent setup in your CI if you're interested in trying it out 🙂

andrewlock: Thanks @Kyle-Verhoog, we're going through some major-ish updates to the CI system at the moment, so once those have settled down I'll definitely reach out and look at getting that set up, thanks 🙂

lucaspimentel (approved):

test/Datadog.Trace.IntegrationTests/Datadog.Trace.IntegrationTests.csproj
andrewlock: We don't currently list 3rd party testing libraries as far as I can see? e.g. xUnit, Moq etc. Should we add those too, or are they not listed because they're not in the end product components?

src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
andrewlock: Renamed `aspnet_core.mvc` -> `aspnet_core_mvc.request` in https://github.com/DataDog/dd-trace-dotnet/pull/1289/commits/bd6744cc75b177a49d182462d087d2b7aff05604

test/Datadog.Trace.IntegrationTests/Datadog.Trace.IntegrationTests.csproj
lucaspimentel: I think you're right. Never mind!

src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
```diff
@@ -452,6 +948,42 @@ public void Remove(string name)
...
+
+            /// <summary>
+            /// Gets or sets the HTTP method, as it requires normalization, so avoids repeatedly calculations
+            /// </summary>
+            public string HttpMethod { get; set; }
+
+            /// <summary>
+            /// Gets or Sets the original URL received by the pipeline
+            /// </summary>
+            public string Url { get; set; }
```
zacharycmontoya, line 985: What do you think about renaming this to `OriginalUrl`?

src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
```diff
@@ -304,29 +534,253 @@ private void OnMvcBeforeAction(object arg)
...
+                        if (routeData is not null)
+                        {
+                            var route = routeData.Routers.OfType<RouteBase>().FirstOrDefault();
+                            routeTemplate = route?.ParsedTemplate;
+                        }
+                    }
+
+                    if (routeTemplate is not null)
+                    {
+                        // If we don't have a route, don't overwrite the existing resource name
```
zacharycmontoya, line 618: I found this comment a little confusing trying to reason through a double-negative, how about this?
```suggestion
                        // If we have a route, use the route to determine the resource name
```

src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
```diff
@@ -304,29 +534,253 @@ private void OnMvcBeforeAction(object arg)
...
+                    {
+                        parentTags.AspNetCoreRoute = aspNetRoute;
+                    }
+
+                    parentSpan.ResourceName = span.ResourceName;
+                }
+            }
+        }
+
+        private void OnRoutingEndpointMatched(object arg)
```
zacharycmontoya, line 654: As an ASP.NET Core newbie, assuming each of these private methods are hit in one execution, are the order of these private methods executed in the same order as they are defined in source? It would help readability (for me) if those two orders matched

src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
```diff
@@ -304,29 +534,253 @@ private void OnMvcBeforeAction(object arg)
...
+                        // URL has changed from original, so treat this execution as a "subsequent" request
+                        // Typically occurs for 404s for example
+                        isFirstExecution = false;
+                    }
+                }
+
+                // NOTE: This event is when the routing middleware selects an endpoint. Additional middleware (e.g
+                //       Authorization/CORS) may still run, and the endpoint itself has not started executing.
+
+                var rawEndpointFeature = httpContext.Features[EndpointFeatureType];
```
zacharycmontoya, line 695: Is there a scenario in which `EndpointFeatureType` is null?

src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
```diff
@@ -304,29 +534,253 @@ private void OnMvcBeforeAction(object arg)
...
+                                        ? raw as string
+                                        : null;
+                string actionName = routeValues.TryGetValue("action", out raw)
+                                        ? raw as string
+                                        : null;
+                string areaName = routeValues.TryGetValue("area", out raw)
+                                      ? raw as string
+                                      : null;
+
+                // If we don't have a route, don't overwrite the existing resource name
```
zacharycmontoya, line 746: Can we remove this comment? It seems unnecessary but I could be wrong

src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
```diff
@@ -418,6 +872,48 @@ public struct BadHttpRequestExceptionStruct
...
+        }
+
+        [DuckCopy]
+        public struct RoutePatternPathSegmentStruct
+        {
+            public IEnumerable Parts;
+        }
+
+        [DuckCopy]
+        public struct RoutePatternContentPartStruct
```
zacharycmontoya, line 903: What type is this proxying? I had trouble finding out the type from ASP.NET Core docs

zacharycmontoya (approved):
LGTM! Looking forward to seeing this in action! I left some small comments but nothing blocking

src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
andrewlock: Ah yes, it's actually being used to proxy two different types `RoutePatternLiteralPart` and `RoutePatternSeparatorPart`, as they both expose the `Content` property we need. I'll add doc comments to the types to make it clear.

However, I wonder, is proxying two types like this sensible, or should we have a separate type for each? I assumed this was preferable, as it reduces the number of casts we have to do, and the number of duck types in general 🤔

src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
andrewlock: Done in https://github.com/DataDog/dd-trace-dotnet/pull/1289/commits/f1245d9aabb243613ed22a52a8b708d86d3c9cd2

src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
andrewlock: Yeah, great point, now you mention it, this was messing with my head too 😆 done in https://github.com/DataDog/dd-trace-dotnet/pull/1289/commits/2740121e762e09cadef34d593d28b0a8467790b0

src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
andrewlock: Good point - it _shouldn't_ be if we're in here, but better to be safe 👍 done in https://github.com/DataDog/dd-trace-dotnet/pull/1289/commits/75a6e197db80f988656d972bab82c992e318aac5

src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
zacharycmontoya: Actually this is completely fine, thanks for documenting the two types. When we attempt the duck type for each underlying instance, it will succeed because the `Content` property is present, and we'll just have two `Type`->`DuckTypeProxy` entries in our cache 👍🏼

PR #1290, opened 2021-03-17
Add support for ducktyping in multiple assembly load contexts

tonyredondo:
This PR adds support for ducktyping in multiple assembly load contexts.

On netcoreapp and netstandard checks if the target type for ducktyping is in a different AssemblyLoadContext and creates the ducktype proxy in the same context.

The PR doesn't modify the behavior in .NET Framework

@DataDog/apm-dotnet


zacharycmontoya:
Closing, implemented by #1291

PR #1293, opened 2021-03-18
Add TryConvertToNormalizedTagName method and apply to DD_TRACE_HEADER_TAGS

zacharycmontoya:
Add new extension method string.TryConvertToNormalizedTagName. This follows the tag requirements specified by https://docs.datadoghq.com/getting_started/tagging/#defining-tags

Additionally, use the new method in the DD_TRACE_HEADER_TAGS feature to convert user-specified tag names to valid tag names. Invalid tag names will be ignored by the feature instead of being allowed through.

Note: This is a breaking behavior difference if the user was submitting invalid tags.

@DataDog/apm-dotnet


zacharycmontoya:
Changes are included in #1301

PR #1296, opened 2021-03-19
Add log rotation for native logger

andrewlock:
- Enable the rotating file sink for the native logger
- Include the pid and process name in the rotating logger to avoid multiple processes locking file, as described in https://github.com/DataDog/dd-trace-dotnet/pull/1044

So instead of a single filename:

```
dotnet-tracer-native.log
```

We will have multiple, one for each process (+ additional for rolled files):

```
dotnet-tracer-native-dotnet-30432.log
dotnet-tracer-native-Samples.AspNetCoreMvc31-31872.log
dotnet-tracer-native-Samples.AspNetCoreRazorPages-31872.log
```

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace.ClrProfiler.Native/logging.cpp
+++ b/src/Datadog.Trace.ClrProfiler.Native/logging.cpp
@@ -3,7 +3,7 @@
 #include "pal.h"
 
 #include "spdlog/sinks/null_sink.h"
-#include "spdlog/sinks/basic_file_sink.h"
+#include "spdlog/sinks/rotating_file_sink.h"
 
 #ifndef _WIN32
 typedef struct stat Stat;
@@ -64,7 +64,8 @@ Logger::Logger() {
   spdlog::flush_every(std::chrono::seconds(3));
 
   try {
-    m_fileout = spdlog::basic_logger_mt("Logger", GetLogPath());
+    m_fileout = spdlog::rotating_logger_mt(
+        "Logger", GetLogPath(), 1048576 * 5, 10);
   }
   catch (...) {
     std::cerr << "Logger Handler: Error creating native log file." << std::endl;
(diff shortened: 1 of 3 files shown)


lucaspimentel:
Can you provide examples of what the file names will look like after this change?

zacharycmontoya:
> I'm wondering if we should strip the .exe out, to be closer to the managed logs style

Yeah that would be a nice touch 👍🏼

andrewlock:
> > I'm wondering if we should strip the .exe out, to be closer to the managed logs style
> 
> Yeah that would be a nice touch 👍🏼

Done in https://github.com/DataDog/dd-trace-dotnet/pull/1296/commits/eaeab20245fc41e59cd3ac536c27551a1c9e47af, so logs now look like:

```
dotnet-tracer-native-dotnet-30432.log
dotnet-tracer-native-Samples.AspNetCoreMvc31-31872.log
dotnet-tracer-native-Samples.AspNetCoreRazorPages-31872.log
```

tonyredondo (approved):
Nice!

src/Datadog.Trace.ClrProfiler.Native/logging.cpp
```diff
@@ -63,8 +63,17 @@ Logger::Logger() {
 
   spdlog::flush_every(std::chrono::seconds(3));
 
+  static auto current_process_name = ToString(GetCurrentProcessName());
+  static auto current_process_id= GetPID();
+  static auto current_process_without_extension =
+      current_process_name.substr(0, current_process_name.find_last_of("."));
```
andrewlock, line 69: Wondering if this should be the following instead?

```cpp
static auto current_process_without_extension =
    std::filesystem::path(current_process_name).stem();
```

That handles edge cases like the first letter being a `.`, but that's not realistically going to be a problem for us right? WDYT @tonyredondo

tonyredondo: Well this is defined in C++17 and the linux profiler is compiled using C++11 + some 17 extensions. We can check if that compiles and behave correctly in non windows platforms. If that is the case you have my vote to use it.

andrewlock: ah, oops, good point. Didn't notice the usages of `std::filesystem` were in an `#ifdef`. This should be fine then I guess

zacharycmontoya (approved):

PR #1297, opened 2021-03-22
Increase buffer size to avoid edge cases of span dropping

tonyredondo:
Currently the `AgentWriter` uses a maximum of 10mb as default value in both front and back buffer. This PR increases both maximum value to 45mb in both buffers only when running in the test runner to avoid edge cases when both buffers are full without flushing and a test span can be dropped.

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace.ClrProfiler.Managed/Integrations/Testing/Common.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/Integrations/Testing/Common.cs
@@ -1,4 +1,5 @@
 using Datadog.Trace.Ci;
+using Datadog.Trace.Configuration;
 
 namespace Datadog.Trace.ClrProfiler.Integrations.Testing
 {
@@ -6,7 +7,10 @@ namespace Datadog.Trace.ClrProfiler.Integrations.Testing
     {
         static Common()
         {
-            TestTracer = Tracer.Instance;
+            var settings = TracerSettings.FromDefaultSources();
+            settings.TraceBufferSize = 1024 * 1024 * 45; // slightly lower than the 50mb payload agent limit.
+
+            TestTracer = new Tracer(settings);
             ServiceName = TestTracer.DefaultServiceName;
 
             // Preload environment variables.
--- a/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/Common.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/Common.cs
@@ -1,4 +1,5 @@
 using Datadog.Trace.Ci;
+using Datadog.Trace.Configuration;
 
 namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Testing
 {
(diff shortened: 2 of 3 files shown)


lucaspimentel (approved):

src/Datadog.Trace.ClrProfiler.Managed/Integrations/Testing/Common.cs
```diff
@@ -1,12 +1,16 @@
...
 
 namespace Datadog.Trace.ClrProfiler.Integrations.Testing
 {
     internal static class Common
     {
         static Common()
         {
-            TestTracer = Tracer.Instance;
+            var settings = TracerSettings.FromDefaultSources();
+            settings.TraceBufferSize = 1024 * 1024 * 45; // slightly lower than the 50mb payload agent limit.
```
andrewlock, line 11: I think you can achieve this without having to change the `TestSettings`  api. e.g

```suggestion
            var sources = GlobalSettings.CreateDefaultConfigurationSource();
            var overrideSource = new NameValueConfigurationSource(new NameValueCollection
            {
                { ConfigurationKeys.BufferSize, (1024 * 1024 * 45).ToString() } // slightly lower than the 50mb payload agent limit.
            });
            sources.Add(overrideSource);
            var settings = new TracerSettings(sources);
```

(This requires additional namespace args, so won't compile as-is)

andrewlock (commented):
I think we can achieve this without changing the `TestSettings` API?

src/Datadog.Trace.ClrProfiler.Managed/Integrations/Testing/Common.cs
tonyredondo: Oh didn't know we can do this... great! thank you!

lucaspimentel: You still want `TracerSettings.FromDefaultSources()` instead of `GlobalSettings.CreateDefaultConfigurationSource()`, though. And the different "sources" are meant to be things like env vars, `web.config`, or json files. If you will change settings in code, `TracerSettings` is where you're meant to do it. Or at least this is what we tell users [in our docs](https://docs.datadoghq.com/tracing/setup_overview/setup/dotnet-framework/?tab=code#configuration).

tonyredondo: Ohh, so I cannot do that @andrewlock :(

lucaspimentel: You can, technically, but why would you do it?

andrewlock: > You can, technically, but why would you do it?

Because we probably don't want BufferSize to be editable at any point - it's currently readonly, and should probably remain like that?

> You still want TracerSettings.FromDefaultSources() instead of GlobalSettings.CreateDefaultConfigurationSource()

`TracerSettings.FromDefaultSources()` just calls `GlobalSettings.CreateDefaultConfigurationSource()`, so doesn't seem any harm in doing this?
https://github.com/DataDog/dd-trace-dotnet/blob/1b718dc6ae05dda38a9595bf32846ae3950ce4ab/src/Datadog.Trace/Configuration/TracerSettings.cs#L390-L404

lucaspimentel: I would be fine with adding the property setter to be consistent with other Agent-related settings (host, port, etc are all read only once and later changes are ignored). It also keeps the testing code in this PR smaller (we don't use "fake" sources anywhere else), and the property is internal anyway. It's not a great design, but it what we have for now.

zacharycmontoya (approved):

PR #1300, opened 2021-03-22
CallTarget: Benchmark tests

tonyredondo:
Benchmark tests migration to CallTarget.

@DataDog/apm-dotnet


test/benchmarks/Benchmarks.Trace/HttpClientBenchmark.cs
```diff
@@ -121,17 +70,18 @@ public async Task<string> SendAsync()
                 MdToken,
                 (long)GuidPtr);
 
-            await task;
-
+            task.GetAwaiter().GetResult();
```
gleocadie, line 73: just to understand: Why replacing `await task` by `task.GetAwaiter().GetResult()` ?
Is it because it not well-supported in benchmark .NET ?

tonyredondo: In this case is because `CallTargetSendAsync` cannot use async/await because I'm using `unsafe` in the method. So because both `CallTargetSendAsync` and `SendAsync` will be compared against each other, I want to replicate the same behavior.

tonyredondo: another way to solve it is using a different method for `unsafe` and calling from the benchmark method, but I didn't want to add the extra method call 😅

gleocadie: Sure no pb, it was out of curiosity.
Thanks

test/benchmarks/Benchmarks.Trace/ElasticsearchBenchmark.cs
```diff
@@ -70,6 +75,24 @@ public int CallElasticsearchAsync()
...
+        public unsafe object CallTargetCallElasticsearch()
+        {
+            return CallTarget.Run<RequestPipeline_CallElasticsearch_Integration, RequestPipeline, RequestData, object>(Pipeline, Data, &GetData);
+
+            static object GetData(RequestData data) => null;
+        }
+
+
+        [Benchmark]
+        public unsafe object CallTargetCallElasticsearchAsync()
```
kevingosse, line 89: You should make sure to have the same return type as the callsite counterpart (here, `CallElasticsearchAsync` has a `int` return type) otherwise the comparison is unfair

kevingosse (changes requested):

test/benchmarks/Benchmarks.Trace/ElasticsearchBenchmark.cs
tonyredondo: Right!, done in e06c511

kevingosse (approved):

test/benchmarks/Benchmarks.Trace/AspNetCoreBenchmark.cs
```diff
@@ -132,6 +160,12 @@ public string SendRequest()
         {
             return null;
         }
+
+        [Benchmark]
+        public string CallTargetSendRequest()
```
zacharycmontoya, line 165: Quick question, can we remove the benchmarks in this conditional compilation or no?

tonyredondo: Sadly no, I've just forgot to add it and the CI was complaining while looking for the test. @kevingosse know more about the underlying reason but the way we execute the benchmark tests is based in compiling for `netcoreapp3.1` targeting the `net472` framework and that mix forces us to declare the same benchmarks for both runtimes.

test/benchmarks/Benchmarks.Trace/CallTarget.cs
```diff
@@ -0,0 +1,733 @@
...
+    /// Helper class to simulate the calltarget rewriting for benchmark tests
+    /// </summary>
+    internal unsafe static class CallTarget
+    {
+        // ***************************************************************************************************************
+        //  Run
+        // ***************************************************************************************************************
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static TResult Run<TIntegration, TTarget, TResult>(TTarget targetInstance, delegate*<TResult> bodyCallback)
```
zacharycmontoya, line 19: According to MS docs: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/function-pointers

> These types are represented using the function pointer type as outlined in ECMA-335. This means invocation of a delegate* will use calli where invocation of a delegate will use callvirt on the Invoke method. Syntactically though invocation is identical for both constructs.

Will the resulting `calli` instruction make our benchmarks appear faster than they normally would be? Could we instead use a normal delegate here? I'm not very familiar with the `calli` IL instruction, but I wonder if there's a way we could do `callvirt` and `call` instructions so it better matches our runtime instrumentation

tonyredondo: Well the calli here is on purpose because is the fastest one. The reason is because in our instrumentation there is no call at all because we run the actual method body. In this case the method body is outside the calltarget try/catch construct so I need the fastest call I can make. That means that the current calltarget is faster than the actual calli.

zacharycmontoya: Oh I see, that makes sense, I forgot that we're already inside the method. Thank you for clarifying!! 😄

zacharycmontoya (approved):

PR #1301, opened 2021-03-23
Improve DD_TRACE_HEADER_TAGS to decorate web server spans based on response headers

zacharycmontoya:
Replaces #1253 and #1293

### Changes
Previously, the `DD_TRACE_HEADER_TAGS` setting would accept a list of `<http_header_name>:<tag_name>` pairs and add tags to web server spans with the specified tag name if the header was present on an incoming HTTP request. This PR adds functionality to also consider headers from the HTTP response the web server issues. Additional changes include:

1. In accordance with the HTTP Headers Tracing RFC, `DD_TRACE_HEADER_TAGS` now accepts entries without a specified tag name. In this scenario, the resulting tag name will be `http.<request|response>.headers.<normalized_header_name>`.
2. All tag names are now being normalized

Affected integrations:
- AspNet
- AspNetMvc
- AspNetWebApi2
- AspNetCore

Testing:
- Local testing for all integrations
- AspNetCore integration test added

@DataDog/apm-dotnet


src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetMvcIntegration.cs
```diff
@@ -110,19 +110,16 @@ public static Scope CreateScope(ControllerContextStruct controllerContext)
                 var tracer = Tracer.Instance;
                 var tagsFromHeaders = Enumerable.Empty<KeyValuePair<string, string>>();
 
-                if (tracer.ActiveScope == null)
```
andrewlock, line 113: So if I understand this change correctly, previously we were only extracting the tags from headers when there was a propagated distributed context, and this change ensures we always extract the headers, even when there's no distributed tracing?

src/Datadog.Trace/Configuration/StringConfigurationSource.cs
```diff
@@ -15,8 +15,9 @@ public abstract class StringConfigurationSource : IConfigurationSource
         /// <paramref name="data"/>.
         /// </summary>
         /// <param name="data">A string containing key-value pairs which are comma-separated, and for which the key and value are colon-separated.</param>
+        /// <param name="allowOptionalMappings">Determines whether to create dictionary entries when the input has no value mapping</param>
         /// <returns><see cref="IDictionary{TKey, TValue}"/> of key value pairs.</returns>
-        public static IDictionary<string, string> ParseCustomKeyValues(string data)
+        public static IDictionary<string, string> ParseCustomKeyValues(string data, bool allowOptionalMappings = false)
```
andrewlock, line 20: Not sure if it's an issue, but this is a breaking change in a public API. Maybe we should add as an overload without optional parameter instead and delegate the existing call to this one?

src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -132,13 +133,12 @@ public TracerSettings(IConfigurationSource source)
             GlobalTags = GlobalTags.Where(kvp => !string.IsNullOrWhiteSpace(kvp.Key) && !string.IsNullOrWhiteSpace(kvp.Value))
                                    .ToDictionary(kvp => kvp.Key.Trim(), kvp => kvp.Value.Trim());
 
-            HeaderTags = source?.GetDictionary(ConfigurationKeys.HeaderTags) ??
+            var inputHeaderTags = source?.GetDictionary(ConfigurationKeys.HeaderTags, allowOptionalMappings: true) ??
                          // default value (empty)
                          new ConcurrentDictionary<string, string>();
```
andrewlock, line 138: nit: not much need for this to be a `ConcurrentDictionary<>`, any more seeing as we're immediately enumerating it below and loading into a standard `Dictionary<>`

src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -493,6 +493,24 @@ internal bool IsNetStandardFeatureFlagEnabled()
             return value == "1" || value == "true";
         }
 
+        internal IDictionary<string, string> InitializeHeaderTags(IDictionary<string, string> configurationDictionary)
+        {
+            IDictionary<string, string> headerTags = new Dictionary<string, string>();
```
andrewlock, line 498: Just wondering why explicit `IDictionary<>` interface here rather than var? I realise the return type is `IDictionary<>` just wondering if there's some performance aspect I'm not aware of 🙂

src/Datadog.Trace/ExtensionMethods/StringExtensions.cs
```diff
@@ -69,5 +70,65 @@ public static string TrimEnd(this string value, string suffix, StringComparison
...
+        /// <returns>Returns whether the conversion was successful</returns>
+        public static bool TryConvertToNormalizedTagName(this string value, out string normalizedTagName, bool convertPeriodsToUnderscores = false)
+        {
+            if (string.IsNullOrWhiteSpace(value))
+            {
+                normalizedTagName = null;
+                return false;
+            }
+
+            string trimmedValue = value?.Trim();
```
andrewlock, line 100: nit: I guess we know it's definitely not null here 🙂

src/Datadog.Trace/SpanContextPropagator.cs
```diff
@@ -166,6 +183,47 @@ public SpanContext Extract<T>(T carrier, Func<T, string, IEnumerable<string>> ge
...
+                        {
+                            return key.TagPrefix + "." + normalizedHeaderTagName;
+                        }
+                        else
+                        {
+                            return null;
+                        }
+                    });
+
+                    if (!string.IsNullOrWhiteSpace(tagNameResult))
```
andrewlock, line 219: nit: I guess `tagNameResult` is always either `null` or non-empty, so could probably replace with a null check here?

andrewlock (approved):
LGTM, mostly just questions and nits 🙂 We might have some minor merge conflicts to deal with https://github.com/DataDog/dd-trace-dotnet/pull/1288

src/Datadog.Trace.AspNet/TracingHttpModule.cs
```diff
@@ -181,9 +182,17 @@ private void OnError(object sender, EventArgs eventArgs)
                 var httpException = exception as HttpException;
                 var is404 = httpException?.GetHttpCode() == 404;
 
-                if (exception != null && !is404 && httpContext.Items[_httpContextScopeKey] is Scope scope)
+                if (httpContext.Items[_httpContextScopeKey] is Scope scope)
                 {
-                    scope.Span.SetException(exception);
+                    if (httpContext != null)
```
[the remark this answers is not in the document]
zacharycmontoya: The unnecessary check must have been a copy/paste error, good catch!

src/Datadog.Trace/SpanContextPropagator.cs
zacharycmontoya: Yeah good catch, I'll update that

src/Datadog.Trace/ExtensionMethods/StringExtensions.cs
zacharycmontoya: 🤦🏼 I think I switched the order of checks at some point so this check became unnecessary. Thank you for spotting 😄

src/Datadog.Trace/Configuration/TracerSettings.cs
zacharycmontoya: I really don't know why I did that, I'll update this to use `var` so there's no confusion

src/Datadog.Trace/Configuration/TracerSettings.cs
zacharycmontoya: Good point, will change 😄

src/Datadog.Trace/Configuration/StringConfigurationSource.cs
zacharycmontoya: Yes you're absolutely right, I'll make that change 👍🏼

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetMvcIntegration.cs
zacharycmontoya: It's slightly different than what you stated. Before, we would only parse the distributed tracing headers + header tags when this was a local root span (first web server span in the trace). I modified this so that we extract the header tags for each web server integration, regardless if it's a root span. This adds additional processing when we have a `aspnet.request` span followed by a `aspnet_mvc.request` span. This change isn't strictly necessary to achieve the overall feature since we only need the root span to get the header tags. I'd be okay with removing this and reverting back to the previous implementation. Do you think I should go ahead and remove it?

src/Datadog.Trace/Configuration/StringConfigurationSource.cs
zacharycmontoya: Done in b3f9c8c9975f0bbf1ada23a363a69edf4e77ac8d

src/Datadog.Trace/Configuration/TracerSettings.cs
zacharycmontoya: Done in b3f9c8c9975f0bbf1ada23a363a69edf4e77ac8d

src/Datadog.Trace/ExtensionMethods/StringExtensions.cs
zacharycmontoya: Done in b3f9c8c9975f0bbf1ada23a363a69edf4e77ac8d

src/Datadog.Trace/Configuration/TracerSettings.cs
zacharycmontoya: Done in b3f9c8c9975f0bbf1ada23a363a69edf4e77ac8d

src/Datadog.Trace/SpanContextPropagator.cs
zacharycmontoya: Done in b3f9c8c9975f0bbf1ada23a363a69edf4e77ac8d

src/Datadog.Trace.AspNet/TracingHttpModule.cs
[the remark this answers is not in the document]
zacharycmontoya: Done in b3f9c8c9975f0bbf1ada23a363a69edf4e77ac8d

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetMvcIntegration.cs
zacharycmontoya: I went ahead and reverted the behavior so we only do this once on the root span, just as before. Let me know if you feel strongly one way or the other, but it's probably best to do as few operations as possible. Commit: 7e13cfd3c8bfce55ccc70edc83c1abe3889ee5ba

src/Datadog.Trace/Configuration/JsonConfigurationSource.cs
```diff
@@ -118,8 +118,34 @@ public T GetValue<T>(string key)
         /// </remarks>
         /// <param name="key">The key that identifies the setting.</param>
         /// <returns><see cref="IDictionary{TKey, TValue}"/> containing all of the key-value pairs.</returns>
-        /// <exception cref="JsonReaderException">Thrown if the configuration value is not a valid JSON string.</exception>"
+        /// <exception cref="JsonReaderException">Thrown if the configuration value is not a valid JSON string.</exception>
         public IDictionary<string, string> GetDictionary(string key)
+        {
+            return GetDictionaryInternal(key, false);
```
lucaspimentel, line 124:
```suggestion
            return GetDictionaryInternal(key, allowOptionalMappings: false);
```

src/Datadog.Trace/Configuration/StringConfigurationSource.cs
```diff
@@ -17,6 +17,18 @@ public abstract class StringConfigurationSource : IConfigurationSource
...
+        }
+
+        /// <summary>
+        /// Returns a <see cref="IDictionary{TKey, TValue}"/> from parsing
+        /// <paramref name="data"/>.
+        /// </summary>
+        /// <param name="data">A string containing key-value pairs which are comma-separated, and for which the key and value are colon-separated.</param>
+        /// <param name="allowOptionalMappings">Determines whether to create dictionary entries when the input has no value mapping</param>
+        /// <returns><see cref="IDictionary{TKey, TValue}"/> of key value pairs.</returns>
+        public static IDictionary<string, string> ParseCustomKeyValues(string data, bool allowOptionalMappings = false)
```
lucaspimentel, line 31: I looks like we can make `allowOptionalMappings` non optional since we have the overload without it. (I thought this generated a compiler warning, but maybe it's just R#)

src/Datadog.Trace/Configuration/StringConfigurationSource.cs
```diff
@@ -88,7 +108,18 @@ public abstract class StringConfigurationSource : IConfigurationSource
         /// <returns><see cref="ConcurrentDictionary{TKey, TValue}"/> containing all of the key-value pairs.</returns>
         public IDictionary<string, string> GetDictionary(string key)
         {
-            return ParseCustomKeyValues(GetString(key));
+            return ParseCustomKeyValues(GetString(key), false);
```
lucaspimentel, line 111:
```suggestion
            return ParseCustomKeyValues(GetString(key), allowOptionalMappings: false);
```

src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
```diff
@@ -67,6 +71,26 @@ public static void AddTagsFromDbCommand(this Span span, IDbCommand command)
             }
         }
 
+        internal static void SetHeaderTags(this Span span, IHeadersCollection headers, IDictionary<string, string> headerTags, string defaultTagPrefix)
+        {
+            if (headerTags is not null && !headerTags.IsEmpty())
+            {
+                try
+                {
+                    // extract propagation details from http headers
```
lucaspimentel, line 80: Looks like this comment was copied from somewhere that used `SpanContextPropagator` for something else.

Which made realize that `SpanContextPropagator` is not the right place to copy http header to span tags (we're not propagating any span context), but that's a separate issue 😅

src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
```diff
@@ -67,6 +71,26 @@ public static void AddTagsFromDbCommand(this Span span, IDbCommand command)
...
+                    var tagsFromHeaders = SpanContextPropagator.Instance.ExtractHeaderTags(headers, headerTags, defaultTagPrefix);
+                    foreach (KeyValuePair<string, string> kvp in tagsFromHeaders)
+                    {
+                        span.SetTag(kvp.Key, kvp.Value);
```
lucaspimentel, line 84: Possible optimization for later: add `SetTags(tagsFromHeaders)` which uses `Dictionary<TKey,TValue>.EnsureCapacity(Int32)` (added in .NET Core 2.1) to increase the dictionary size only once.

src/Datadog.Trace/ExtensionMethods/StringExtensions.cs
```diff
@@ -69,5 +70,65 @@ public static string TrimEnd(this string value, string suffix, StringComparison
...
+        ///    - Periods
+        ///    - Slashes
+        ///
+        /// Note: This method will trim leading/trailing whitespace before checking the requirements.
+        /// </summary>
+        /// <param name="value">Input string to convert into tag name</param>
+        /// <param name="normalizedTagName">If the method returns true, the normalized tag name</param>
+        /// <param name="convertPeriodsToUnderscores">A flag determining whether periods should also be converted to underscores</param>
+        /// <returns>Returns whether the conversion was successful</returns>
+        public static bool TryConvertToNormalizedTagName(this string value, out string normalizedTagName, bool convertPeriodsToUnderscores = false)
```
lucaspimentel, line 92: Since this method is internal, I would make `convertPeriodsToUnderscores` non-optional to force us to be explicit about it in every call. To me the optional parameter is a convenience for users in public APIs. (level of care: low)

src/Datadog.Trace/SpanContextPropagator.cs
```diff
@@ -152,11 +159,21 @@ public SpanContext Extract<T>(T carrier, Func<T, string, IEnumerable<string>> ge
             return new SpanContext(traceId, parentId, samplingPriority, null, origin);
         }
 
+        [Obsolete("This method is deprecated and will be removed. Use ExtractHeaderTags<T>(T, IEnumerable<KeyValuePair<string, string>>, string) instead. " +
```
lucaspimentel, line 162: @andrewlock: this `[Obsolete]` is what I meant in your PR

src/Datadog.Trace/ExtensionMethods/StringExtensions.cs
zacharycmontoya: That's a good suggestion, will do

src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
zacharycmontoya: There's a good OTel PR to improve context propagation which removes the header tags from the propagation helpers. Hopefully I can port that over soon...Anyways this comment has been removed in c1ac5ce1868b7c359e741bd04d05e742647a9a33

src/Datadog.Trace/Configuration/StringConfigurationSource.cs
zacharycmontoya: Done in c1ac5ce1868b7c359e741bd04d05e742647a9a33

src/Datadog.Trace/Configuration/StringConfigurationSource.cs
zacharycmontoya: Good idea, done in c1ac5ce1868b7c359e741bd04d05e742647a9a33

src/Datadog.Trace/Configuration/JsonConfigurationSource.cs
zacharycmontoya: Done in c1ac5ce1868b7c359e741bd04d05e742647a9a33

src/Datadog.Trace/ExtensionMethods/StringExtensions.cs
zacharycmontoya: Done in c1ac5ce1868b7c359e741bd04d05e742647a9a33

src/Datadog.Trace/ExtensionMethods/StringExtensions.cs
```diff
@@ -69,5 +70,65 @@ public static string TrimEnd(this string value, string suffix, StringComparison
...
+            string trimmedValue = value.Trim();
+            if (!char.IsLetter(trimmedValue[0]) || trimmedValue.Length > 200)
+            {
+                normalizedTagName = null;
+                return false;
+            }
+
+            if (convertPeriodsToUnderscores)
+            {
+                normalizedTagName = Regex.Replace(input: trimmedValue, pattern: "[^a-zA-Z0-9_:/-]", replacement: "_").ToLowerInvariant();
```
lucaspimentel, line 109: I saw this and my first though was that if we called `ToLowerInvariant()` first, then the regex would have less work to do. (We're going to `ToLowerInvariant()` either way and I'm assuming its performance depends on the length of the string and not with how many upper-case or lower-case letters here are.)

My second that was that regex may be overkill for this relatively simple pattern, so I ran some quick benchmarks:

```
|               Method |           Job |       Runtime |        Mean |     Error |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------------- |-------------- |-------------- |------------:|----------:|----------:|-------:|------:|------:|----------:|
|         RegexToLower |      .NET 4.8 |      .NET 4.8 | 1,180.66 ns | 23.537 ns | 29.766 ns | 0.1869 |     - |     - |    1188 B |
|         ToLowerRegex |      .NET 4.8 |      .NET 4.8 | 1,183.25 ns | 23.517 ns | 27.082 ns | 0.1869 |     - |     - |    1188 B |
| ToLowerStringBuilder |      .NET 4.8 |      .NET 4.8 |   141.35 ns |  2.546 ns |  3.031 ns | 0.0343 |     - |     - |     217 B |
|--------------------- |-------------- |-------------- |------------:|----------:|----------:|-------:|------:|------:|----------:|
|         RegexToLower | .NET Core 3.1 | .NET Core 3.1 |   828.58 ns |  9.275 ns |  7.745 ns | 0.1335 |     - |     - |     840 B |
|         ToLowerRegex | .NET Core 3.1 | .NET Core 3.1 |   807.36 ns |  9.227 ns |  8.180 ns | 0.1335 |     - |     - |     840 B |
| ToLowerStringBuilder | .NET Core 3.1 | .NET Core 3.1 |    99.86 ns |  1.510 ns |  1.339 ns | 0.0343 |     - |     - |     216 B |
```

Looks like changing the order to `ToLowerInvariant()` only makes a tiny difference in .NET Core 3.1 and is insignificant in .NET Framework. Using a `for` look with a `StringBuilder` was definitely much master and memory-efficient, though.

```csharp
var sb = new StringBuilder(trimmedValue.ToLowerInvariant());

for (int x = 0; x < sb.Length; x++)
{
    switch (sb[x])
    {
        case >= 'a' and <= 'z' or >= '0' and <= '9' or '_' or ':' or '/' or '-':
            continue;
        default:
            sb[x] = '_';
            break;
    }
}

normalizedTagName = sb.ToString();
```

src/Datadog.Trace/ExtensionMethods/StringExtensions.cs
lucaspimentel: On second thought (sorry!), do we need the overload without `convertPeriodsToUnderscores` at all?

zacharycmontoya: We don't technically need it in this set of changes...but we kind of should be normalizing the tag names that users give us in `SetTag` 😄

src/Datadog.Trace/ExtensionMethods/StringExtensions.cs
zacharycmontoya: Thanks for suggesting this new approach 😄 Running some benchmarks validates the results you got as well, so I'll make the update

zacharycmontoya: Done in b02a4fe24973e9e6be8aa48b689b30e9818d3d24

src/Datadog.Trace/ExtensionMethods/StringExtensions.cs
lucaspimentel: My mistake! I thought they were overloads (same name), but it's `TryConvertToNormalizedTagName` vs `TryConvertToNormalizedHeaderTagName`. Resolving this comment.

lucaspimentel (approved):

PR #1303, opened 2021-03-23
Compute top-level spans on the tracer side

kevingosse:
This is a requirement to implement flushing of partial traces.

### DIFF
--- a/src/Datadog.Trace/AgentHttpHeaderNames.cs
+++ b/src/Datadog.Trace/AgentHttpHeaderNames.cs
@@ -39,6 +39,11 @@ namespace Datadog.Trace
         public const string ContainerId = "Datadog-Container-ID";
 
         /// <summary>
+        /// Tells the agent whether top-level spans are computed by the tracer
+        /// </summary>
+        public const string ComputedTopLevel = "Datadog-Client-Computed-Top-Level";
+
+        /// <summary>
         /// Gets the default constant header that should be added to any request to the agent
         /// </summary>
         internal static KeyValuePair<string, string>[] DefaultHeaders { get; } =
@@ -47,7 +52,8 @@ namespace Datadog.Trace
             new(TracerVersion, TracerConstants.AssemblyVersion),
             new(HttpHeaderNames.TracingEnabled, "false"), // don't add automatic instrumentation to requests directed to the agent
             new(LanguageInterpreter, FrameworkDescription.Instance.Name),
-            new(LanguageVersion, FrameworkDescription.Instance.ProductVersion)
+            new(LanguageVersion, FrameworkDescription.Instance.ProductVersion),
+            new(ComputedTopLevel, "1")
         };
     }
 }
--- a/src/Datadog.Trace/AgentHttpHeaderNames.cs
+++ b/src/Datadog.Trace/AgentHttpHeaderNames.cs
@@ -41,7 +41,7 @@ namespace Datadog.Trace
         /// <summary>
         /// Tells the agent whether top-level spans are computed by the tracer
         /// </summary>
-        public const string ComputedTopLevel = "Datadog-Client-Computed-Top-Level";
+        public const string ComputedTopLevelSpan = "Datadog-Client-Computed-Top-Level";
 
         /// <summary>
         /// Gets the default constant header that should be added to any request to the agent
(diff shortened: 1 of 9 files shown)


src/Datadog.Trace/Metrics.cs
```diff
@@ -21,5 +21,11 @@ internal static class Metrics
         /// Read: Rate Limiter Priority Sample Rate
         /// </summary>
         public const string SamplingLimitDecision = "_dd.limit_psr";
+
+        /// <summary>
+        /// To be set on top-level spans
+        /// Top-level spans are spans with a different service name from their immediate parent
```
lucaspimentel, line 27:
```suggestion
        /// Identifies top-level spans.
        /// Top-level spans have a different service name from their immediate parent or have no parent.
```

src/Datadog.Trace/AgentHttpHeaderNames.cs
```diff
@@ -38,6 +38,11 @@ internal static class AgentHttpHeaderNames
         /// </summary>
         public const string ContainerId = "Datadog-Container-ID";
 
+        /// <summary>
+        /// Tells the agent whether top-level spans are computed by the tracer
+        /// </summary>
+        public const string ComputedTopLevel = "Datadog-Client-Computed-Top-Level";
```
lucaspimentel, line 44: We can't change the header name, but would you be ok with renaming the property for clarity?
```suggestion
        public const string ComputedTopLevelSpan = "Datadog-Client-Computed-Top-Level";
```

lucaspimentel (approved):
LGTM. Just tweaking some comments and a property name.

lucaspimentel:
Support for `_dd.top_level` was added in Agent 7.25 (https://github.com/DataDog/datadog-agent/pull/6941). Older versions will just ignore the new http header and the new metric, right?https://github.com/DataDog/datadog-agent/pull/6941

test/Datadog.Trace.Tests/SpanTests.cs
```diff
@@ -135,5 +138,28 @@ public void Accurate_Duration(int minimumSleepMilliseconds)
...
+            var spans = new List<(Scope Scope, bool IsTopLevel)>();
+
+            spans.Add((_tracer.StartActive("Root", serviceName: "root"), true));
+            spans.Add((_tracer.StartActive("Child1", serviceName: "root"), false));
+            spans.Add((_tracer.StartActive("Child2", serviceName: "child"), true));
+            spans.Add((_tracer.StartActive("Child3", serviceName: "child"), false));
+            spans.Add((_tracer.StartActive("Child4", serviceName: "root"), true));
+
+            // Dispose the scope to populate the top-level property
+            foreach (var (scope, _) in spans.AsEnumerable().Reverse())
```
zacharycmontoya, line 154: You don't need to dispose these anymore to set the top-level property, do you?

zacharycmontoya (approved):

andrewlock (approved):

test/Datadog.Trace.Tests/SpanTests.cs
lucaspimentel: I was going to ask the same question, but then thought it doesn't hurt (although the comment may be wrong). Thanks for asking 😅

src/Datadog.Trace/Metrics.cs
kevingosse: Done in https://github.com/DataDog/dd-trace-dotnet/pull/1303/commits/87ab80ddc9107b9b6cff3028a319b5824358d6e7

src/Datadog.Trace/AgentHttpHeaderNames.cs
kevingosse: Done in https://github.com/DataDog/dd-trace-dotnet/pull/1303/commits/87ab80ddc9107b9b6cff3028a319b5824358d6e7

test/Datadog.Trace.Tests/SpanTests.cs
kevingosse: You're right. Removed in https://github.com/DataDog/dd-trace-dotnet/pull/1303/commits/87ab80ddc9107b9b6cff3028a319b5824358d6e7

PR #1305, opened 2021-03-24
CallTarget ADO.NET performance improvements.

tonyredondo:
- Improve performance of CallTarget ADO.NET instrumentation in both CPU and Memory.

image

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace.ClrProfiler.Managed/ScopeDBFactory.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/ScopeDBFactory.cs
@@ -26,7 +26,7 @@ namespace Datadog.Trace.ClrProfiler
         {
             if (_dbTypeName == null)
             {
-                // don't create a scope, skip this trace
+                // don't create a scope, skip this span
                 return null;
             }
 
--- a/src/Datadog.Trace.ClrProfiler.Managed/ScopeDBFactory.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/ScopeDBFactory.cs
@@ -1,0 +1,80 @@
+using System;
+using System.Data;
+using Datadog.Trace.ClrProfiler.Integrations.AdoNet;
+using Datadog.Trace.ExtensionMethods;
+using Datadog.Trace.Logging;
+
+namespace Datadog.Trace.ClrProfiler
+{
+    internal static class ScopeDBFactory<T>
+    {
+        private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(ScopeDBFactory<T>));
+
+        private static readonly string _fullName;
+        private static readonly string _dbTypeName;
+        private static readonly string _operationName;
+
+        static ScopeDBFactory()
+        {
+            var type = typeof(T);
+            _fullName = type.FullName;
+            _dbTypeName = ScopeFactory.GetDbType(type.Name);
+            _operationName = $"{_dbTypeName}.query";
+        }
+
+        public static Scope CreateDbCommandScope(Tracer tracer, IDbCommand command)
+        {
+            if (_dbTypeName == null)
+            {
+                // don't create a scope, skip this trace
+                return null;
+            }
+
+            if (!tracer.Settings.IsIntegrationEnabled(AdoNetConstants.IntegrationId))
+            {
+                // integration disabled, don't create a scope, skip this trace
+                return null;
+            }
+
+            if (tracer.Settings.AdoNetExcludedTypes.Count > 0 && tracer.Settings.AdoNetExcludedTypes.Contains(_fullName))
+            {
+                // AdoNet type disabled, don't create a scope, skip this trace
+                return null;
+            }
+
+            Scope scope = null;
+
+            try
+            {
+                Span parent = tracer.ActiveScope?.Span;
+
+                if (parent != null &&
+                    parent.Type == SpanTypes.Sql &&
+                    parent.GetTag(Tags.DbType) == _dbTypeName &&
+                    parent.ResourceName == command.CommandText)
+                {
+                    // we are already instrumenting this,
+                    // don't instrument nested methods that belong to the same stacktrace
+                    // e.g. ExecuteReader() -> ExecuteReader(commandBehavior)
+                    return null;
+                }
+
+                string serviceName = tracer.Settings.GetServiceName(tracer, _dbTypeName);
+
+                var tags = new SqlTags();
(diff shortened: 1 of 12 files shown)


kevingosse (approved):

src/Datadog.Trace.ClrProfiler.Managed/ScopeDBFactory.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        static ScopeDBFactory()
+        {
+            var type = typeof(T);
+            _fullName = type.FullName;
+            _dbTypeName = ScopeFactory.GetDbType(type.Name);
+            _operationName = $"{_dbTypeName}.query";
+        }
+
+        public static Scope CreateDbCommandScope(Tracer tracer, IDbCommand command)
```
lucaspimentel, line 25: Should we add unit tests for `CreateDbCommandScope()`?

src/Datadog.Trace.ClrProfiler.Managed/ScopeDBFactory.cs
```diff
@@ -0,0 +1,80 @@
...
+        {
+            if (_dbTypeName == null)
+            {
+                // don't create a scope, skip this trace
```
lucaspimentel, line 29: I'm going to start fixing these `skip this trace` comments everywhere 😅 
```suggestion
                // don't create a scope, skip this span
```

src/Datadog.Trace.ClrProfiler.Managed/ScopeDBFactory.cs
```diff
@@ -0,0 +1,80 @@
...
+                return null;
+            }
+
+            if (!tracer.Settings.IsIntegrationEnabled(AdoNetConstants.IntegrationId))
+            {
+                // integration disabled, don't create a scope, skip this trace
```
lucaspimentel, line 35:
```suggestion
                // integration disabled, don't create a scope, skip this span
```

src/Datadog.Trace.ClrProfiler.Managed/ScopeDBFactory.cs
```diff
@@ -0,0 +1,80 @@
...
+                return null;
+            }
+
+            if (tracer.Settings.AdoNetExcludedTypes.Count > 0 && tracer.Settings.AdoNetExcludedTypes.Contains(_fullName))
+            {
+                // AdoNet type disabled, don't create a scope, skip this trace
```
lucaspimentel, line 41:
```suggestion
                // AdoNet type disabled, don't create a scope, skip this span
```

src/Datadog.Trace.ClrProfiler.Managed/ScopeDBFactory.cs
tonyredondo: You’re right! I’ll add it.

src/Datadog.Trace.ClrProfiler.Managed/ScopeDBFactory.cs
tonyredondo: Haha

src/Datadog.Trace.ClrProfiler.Managed/ScopeDBFactory.cs
```diff
@@ -0,0 +1,80 @@
...
+                return null;
+            }
+
+            Scope scope = null;
+
+            try
+            {
+                Span parent = tracer.ActiveScope?.Span;
```
lucaspimentel, line 49: I know we haven't done this before, but since users can set `Tracer.Instance = null` using the nuget, we should probably be more defensive about `tracer == null`. What do you think?

lucaspimentel: Could be a separate PR where we do something in ALL integrations. Probably disable auto-instrumentation completely if `Tracer.Instance == null`.

tonyredondo: yeah, I agree we should have that check. I'll prepare a separated PR for that.

andrewlock (approved):

src/Datadog.Trace.ClrProfiler.Managed/ScopeDBFactory.cs
```diff
@@ -0,0 +1,80 @@
...
+        private static readonly string _fullName;
+        private static readonly string _dbTypeName;
+        private static readonly string _operationName;
+
+        static ScopeDBFactory()
+        {
+            var type = typeof(T);
+            _fullName = type.FullName;
+            _dbTypeName = ScopeFactory.GetDbType(type.Name);
+            _operationName = $"{_dbTypeName}.query";
```
andrewlock, line 22: Just to confirm (and for prosperity), this is the source of the speedup, correct? i.e. we're not calling `GetType()` and allocating these strings for every call to `CreateDbCommandScope()`?

tonyredondo: that's right! optimisation using generics static to pre-calculate const values for the same type.

src/Datadog.Trace.ClrProfiler.Managed/ScopeDBFactory.cs
tonyredondo: done in ab7a67e

zacharycmontoya (approved):

test/Datadog.Trace.ClrProfiler.Managed.Tests/ScopeDBFactoryTests.cs
```diff
@@ -0,0 +1,148 @@
...
+        [MemberData(nameof(GetDbCommandScopeData))]
+        public void CreateDbCommandScope_ReturnsNullForExcludedAdoNetTypes(IDbCommand command)
+        {
+            // Set up tracer
+            var collection = new NameValueCollection
+            {
+                { ConfigurationKeys.AdoNetExcludedTypes, command.GetType().FullName }
+            };
+            IConfigurationSource source = new NameValueConfigurationSource(collection);
+            var tracerSettings = new TracerSettings(source);
```
lucaspimentel, line 35: FYI, `TracerSettings.AdoNetExcludedTypes` is mutable so in this case you could have used:
```csharp
var tracerSettings = new TracerSettings();
settings.AdoNetExcludedTypes.Add(command.GetType().FullName);
```

Last time we talked about this it was a get-only property.

PR #1306, opened 2021-03-24
Report Trace Drop/Keep Rate to the Agent

andrewlock:
This adds the `_dd.trace_kr` tag to root spans in a trace. The keep rate is calculated as:

```csharp
var keepRate = 1 - droppedTraceCount/createdTraceCount
```

This excludes sampling, so traces excluded due to sampling do not count towards the drop or keep rate. 

The rate is calculated as a moving average over `n` time buckets of length `t`. So we count the number of dropped and created traces in time `t` (by default, `t` is 1s long), and then average those bucket values across the previous `n` buckets (by default, `n` is 10)

I believe we only drop traces in two places: where the API rejects a trace, or where the buffer is full and so we have to drop a trace. 

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace/Agent/AgentWriter.cs
+++ b/src/Datadog.Trace/Agent/AgentWriter.cs
@@ -36,6 +36,7 @@ namespace Datadog.Trace.Agent
         private readonly ManualResetEventSlim _serializationMutex = new ManualResetEventSlim(initialState: false, spinCount: 0);
 
         private readonly int _batchInterval;
+        private readonly IKeepRateCalculator _keepRateCalculator;
 
         /// <summary>
         /// The currently active buffer.
@@ -54,10 +55,16 @@ namespace Datadog.Trace.Agent
         }
 
         public AgentWriter(IApi api, IDogStatsd statsd, bool automaticFlush = true, int maxBufferSize = 1024 * 1024 * 10, int batchInterval = 100)
+        : this(api, statsd, MovingAverageKeepRateCalculator.CreateDefaultKeepRateCalculator(), automaticFlush, maxBufferSize, batchInterval)
+        {
+        }
+
+        internal AgentWriter(IApi api, IDogStatsd statsd, IKeepRateCalculator keepRateCalculator, bool automaticFlush, int maxBufferSize, int batchInterval)
         {
             _api = api;
             _statsd = statsd;
             _batchInterval = batchInterval;
+            _keepRateCalculator = keepRateCalculator;
 
             var formatterResolver = SpanFormatterResolver.Instance;
 
@@ -125,6 +132,7 @@ namespace Datadog.Trace.Agent
             var completedTask = await Task.WhenAny(_serializationTask, delay)
                 .ConfigureAwait(false);
 
+            _keepRateCalculator.CancelUpdates();
             if (completedTask != delay)
             {
                 await Task.WhenAny(_flushTask, Task.Delay(TimeSpan.FromSeconds(20)))
@@ -264,12 +272,22 @@ namespace Datadog.Trace.Agent
 
                 if (buffer.TraceCount > 0)
                 {
-                    await _api.SendTracesAsync(buffer.Data, buffer.TraceCount).ConfigureAwait(false);
+                    var success = await _api.SendTracesAsync(buffer.Data, buffer.TraceCount).ConfigureAwait(false);
+
+                    if (success)
+                    {
+                        _keepRateCalculator.IncrementKeeps(buffer.TraceCount);
+                    }
+                    else
+                    {
+                        _keepRateCalculator.IncrementDrops(buffer.TraceCount);
+                    }
                 }
             }
             catch (Exception ex)
             {
                 Log.Error(ex, "An unhandled error occurred while flushing a buffer");
+                _keepRateCalculator.IncrementDrops(buffer.TraceCount);
             }
             finally
             {
@@ -303,6 +321,14 @@ namespace Datadog.Trace.Agent
                 return null;
             }
 
+            // Add the current keep rate to the root span
+            var rootSpan = trace[0].Context.TraceContext?.RootSpan;
+            if (rootSpan is not null)
+            {
+                var currentKeepRate = _keepRateCalculator.GetKeepRate();
+                rootSpan.Tags.SetMetric(Metrics.TracesKeepRate, currentKeepRate);
+            }
+
             // We use a double-buffering mechanism
             // This allows the serialization thread to keep doing its job while a buffer is being flushed
             var buffer = _activeBuffer;
@@ -330,6 +356,7 @@ namespace Datadog.Trace.Agent
 
             // All the buffers are full :( drop the trace
             Log.Warning("Trace buffer is full. Dropping a trace.");
+            _keepRateCalculator.IncrementDrops(1);
 
             if (_statsd != null)
             {
--- a/src/Datadog.Trace/Agent/AgentWriter.cs
+++ b/src/Datadog.Trace/Agent/AgentWriter.cs
@@ -5,6 +5,7 @@ using System.Threading.Tasks;
 using Datadog.Trace.Agent.MessagePack;
 using Datadog.Trace.DogStatsd;
 using Datadog.Trace.Logging;
+using Datadog.Trace.Tagging;
 using Datadog.Trace.Util;
 using Datadog.Trace.Vendors.StatsdClient;
 
@@ -326,7 +327,14 @@ namespace Datadog.Trace.Agent
             if (rootSpan is not null)
             {
                 var currentKeepRate = _traceKeepRateCalculator.GetKeepRate();
-                rootSpan.Tags.SetMetric(Metrics.TracesKeepRate, currentKeepRate);
+                if (rootSpan.Tags is CommonTags commonTags)
+                {
+                    commonTags.TracesKeepRate = currentKeepRate;
+                }
+                else
+                {
+                    rootSpan.Tags.SetMetric(Metrics.TracesKeepRate, currentKeepRate);
+                }
             }
 
             // We use a double-buffering mechanism
--- a/src/Datadog.Trace/Agent/AgentWriter.cs
+++ b/src/Datadog.Trace/Agent/AgentWriter.cs
@@ -36,7 +36,7 @@ namespace Datadog.Trace.Agent
         private readonly ManualResetEventSlim _serializationMutex = new ManualResetEventSlim(initialState: false, spinCount: 0);
 
         private readonly int _batchInterval;
-        private readonly IKeepRateCalculator _keepRateCalculator;
+        private readonly IKeepRateCalculator _traceKeepRateCalculator;
 
         /// <summary>
         /// The currently active buffer.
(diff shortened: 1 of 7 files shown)


lucaspimentel (commented):
First pass. Looks good so far, just a few nits about naming

src/Datadog.Trace/Agent/MovingAverageKeepRateCalculator.cs
```diff
@@ -0,0 +1,164 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Datadog.Trace.Logging;
+
+namespace Datadog.Trace.Agent
+{
+    /// <summary>
+    /// Used to calculate the Trace Keep Rate, tracking the number of
+    /// traces kept and dropped that should have been sent to the backend.
+    /// Traces ignored due to sampling are not included in these rates.
```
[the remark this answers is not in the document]
andrewlock: Yep, that was my misunderstanding tbh, I thought we did sampling in the tracer but never checked. Clarified in https://github.com/DataDog/dd-trace-dotnet/pull/1306/commits/304f6f2944efc274a20e541019532d02112d406b

src/Datadog.Trace/Agent/MovingAverageKeepRateCalculator.cs
```diff
@@ -0,0 +1,164 @@
...
+    /// </summary>
+    internal class MovingAverageKeepRateCalculator : IKeepRateCalculator
+    {
+        private const int DefaultKeepRate = 10;
```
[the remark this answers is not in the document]
andrewlock: Done in https://github.com/DataDog/dd-trace-dotnet/pull/1306/commits/85d0233b43fc4cb5c16a6edca3415fbde960bccc

src/Datadog.Trace/Agent/AgentWriter.cs
```diff
@@ -303,6 +321,14 @@ SpanBuffer SwapBuffers()
                 return null;
             }
 
+            // Add the current keep rate to the root span
+            var rootSpan = trace[0].Context.TraceContext?.RootSpan;
+            if (rootSpan is not null)
+            {
+                var currentKeepRate = _keepRateCalculator.GetKeepRate();
+                rootSpan.Tags.SetMetric(Metrics.TracesKeepRate, currentKeepRate);
```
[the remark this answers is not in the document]
andrewlock: Good call. Done in https://github.com/DataDog/dd-trace-dotnet/pull/1306/commits/9c56262128bf1bde83fdff77f70d5455020e6626

src/Datadog.Trace/Agent/MovingAverageKeepRateCalculator.cs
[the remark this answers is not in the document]
lucaspimentel: No worries. Thanks for updating the comment!

src/Datadog.Trace/Agent/MovingAverageKeepRateCalculator.cs
```diff
@@ -0,0 +1,165 @@
...
+
+            _windowSize = windowSize;
+            _bucketDuration = bucketDuration;
+            _dropped = new uint[windowSize];
+            _created = new uint[windowSize];
+
+            if (bucketDuration != Timeout.InfiniteTimeSpan)
+            {
+                Task.Run(UpdateBucketTaskLoopAsync)
+                    .ContinueWith(t => Log.Error(t.Exception, "Error in "), TaskContinuationOptions.OnlyOnFaulted);
```
lucaspimentel, line 50: Error in .... ?

src/Datadog.Trace/Agent/MovingAverageKeepRateCalculator.cs
```diff
@@ -0,0 +1,165 @@
...
+            => new MovingAverageKeepRateCalculator(DefaultWindowSize, DefaultBucketDuration);
+
+        /// <summary>
+        /// Increment the number of kept traces
+        /// </summary>
+        public void IncrementKeeps(int count)
+        {
+            if (count < 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(count));
```
tonyredondo, line 64: Can we move this to a ThrowHelper static class/method, I know there is a lot of code like this, but this method is a good candidate for inlining due the size. But for now the JIT is no able to inline methods that throws or that have try/catch blocks.

src/Datadog.Trace/Agent/MovingAverageKeepRateCalculator.cs
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        /// <summary>
+        /// Increment the number of dropped traces
+        /// </summary>
+        public void IncrementDrops(int count)
+        {
+            if (count < 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(count));
```
tonyredondo, line 77: same as previous comment.

src/Datadog.Trace/Agent/MovingAverageKeepRateCalculator.cs
```diff
@@ -0,0 +1,165 @@
...
+            result = result << 32;
+            return result | (long)total;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private void UnpackBits(long bits, out uint hits, out uint total)
+        {
+            hits = (uint)(bits >> 32);
+            total = (uint)(bits & 0xffffffffL);
+        }
```
tonyredondo, line 163: clever, I like it!

src/Datadog.Trace/Agent/MovingAverageKeepRateCalculator.cs
```diff
@@ -0,0 +1,165 @@
...
+                {
+                    return;
+                }
+
+                UpdateBucket();
+
+                await Task.WhenAny(
+                               Task.Delay(_bucketDuration),
+                               _processExit.Task)
+                          .ConfigureAwait(false);
```
tonyredondo, line 146: I see this a lot in the repo too, In this case for old frameworks `Task.WhenAny` uses a `params` so create a new array on the call, and also it does a defensive copy internally, so 2 arrays per call, and because we are in a loop we allocate that in each pass, we could create a Task[2] array before the while and reuse for each loop. so we reduce to 1 array per call.

This problem doesn't happen in newer runtime versions. 

It just an allocation comment, feel free to ignore it if you want.

tonyredondo (approved):
LGTM, I left a few nits.

src/Datadog.Trace/Agent/MovingAverageKeepRateCalculator.cs
tonyredondo: Thinking about that, I will make a new PR to catch all the throws with a global `ThrowHelper` static class.

src/Datadog.Trace/Agent/MovingAverageKeepRateCalculator.cs
tonyredondo: Thinking about that, I will make a new PR to catch all the throws with a global `ThrowHelper` static class.

src/Datadog.Trace/Agent/MovingAverageKeepRateCalculator.cs
andrewlock: Good to know, thanks! 

Given we know there's a single thread here, can we pre-allocate the array as a field, avoid the allocation entirely?

```csharp
private readonly Task[] _updateLoopTasks = new Task[2];

 private async Task UpdateBucketTaskLoopAsync()
{
    _updateLoopTasks[0] = _processExit.Task;

    while (true)
    {
        if (_processExit.Task.IsCompleted)
        {
            return;
        }

        UpdateBucket();

       _updateLoopTasks[1] =  Task.Delay(_bucketDuration)

        await Task.WhenAny(_updateLoopTasks).ConfigureAwait(false);
    }
}
```

tonyredondo: I think we can but that will add an allocation to newer framework versions 😄 (in new versions there is an overload for 2 Task parameters [here](https://source.dot.net/#System.Private.CoreLib/Task.cs,6054))

And the target for the optimization is this one: https://referencesource.microsoft.com/#mscorlib/system/threading/Tasks/Task.cs,6469

Maybe we can leave this kind of optimization in another PR...

andrewlock: Oh wait, I didn't read your comment properly, I suggested an uglier version of what you suggested 🤦‍♂️.  I dunno, I'm inclined to fix it now as it won't get fixed otherwise 😉 Also, AFAICT, the [newer 2 Task Parameter version is only available in .NET 5](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.whenany?view=net-5.0#System_Threading_Tasks_Task_WhenAny_System_Threading_Tasks_Task_System_Threading_Tasks_Task_), so most versions could do with the optimisation?

andrewlock: + we don't explicitly compile for .NET 5.0 (maybe we should?), so we'll not get that optimisation currently anyway?

src/Datadog.Trace/Agent/MovingAverageKeepRateCalculator.cs
andrewlock: Fixed in https://github.com/DataDog/dd-trace-dotnet/pull/1306/commits/810c31c4743992d52a4f8ddc2c35e9dadbe9c27b

src/Datadog.Trace/Agent/MovingAverageKeepRateCalculator.cs
tonyredondo: oh you're right!... ha ha I see you already implement it.

tonyredondo: although we must be careful with `NET5_0_OR_GREATER` and the CI sdk version :D

andrewlock: > although we must be careful with NET5_0_OR_GREATER and the CI sdk version :D

Yep, but as we're not compiling for .NET 5 at all at the moment, I feel like we can cross that bridge when we come to it 😉

src/Datadog.Trace/Agent/MovingAverageKeepRateCalculator.cs
```diff
@@ -0,0 +1,175 @@
...
+        private const int DefaultWindowSize = 10;
+        private static readonly TimeSpan DefaultBucketDuration = TimeSpan.FromSeconds(1);
+        private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor<MovingAverageKeepRateCalculator>();
+
+        private readonly int _windowSize;
+        private readonly TimeSpan _bucketDuration;
+        private readonly uint[] _dropped;
+        private readonly uint[] _created;
+
+        private readonly CancellationTokenSource _cts = new();
```
zacharycmontoya, line 26: I don't see this being used anywhere, can it be removed?

zacharycmontoya (approved):
LGTM

src/Datadog.Trace/Agent/MovingAverageKeepRateCalculator.cs
andrewlock: Oops, yes, usage was replaced in https://github.com/DataDog/dd-trace-dotnet/pull/1306/commits/8302920160132346da769025b19ba66e3be9ec45. Removed in https://github.com/DataDog/dd-trace-dotnet/pull/1306/commits/7abb3b19a8c5637ef73a9775bffa7d5aed7b83c8

PR #1307, opened 2021-03-24
Disable log rate limiting by default

andrewlock:
@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace/Logging/DatadogLogging.cs
+++ b/src/Datadog.Trace/Logging/DatadogLogging.cs
@@ -156,8 +156,8 @@ namespace Datadog.Trace.Logging
                 return rate;
             }
 
-            // We don't want to rate limit messages by default when in debug mode
-            return GlobalSettings.Source.DebugEnabled ? 0 : DefaultLogMessageRateLimit;
+            // By default, we don't rate limit log messages;
+            return 0;
         }
 
         private static string GetLogDirectory()
--- a/src/Datadog.Trace/Logging/DatadogLogging.cs
+++ b/src/Datadog.Trace/Logging/DatadogLogging.cs
@@ -14,7 +14,8 @@ namespace Datadog.Trace.Logging
     internal static class DatadogLogging
     {
         internal static readonly LoggingLevelSwitch LoggingLevelSwitch = new LoggingLevelSwitch(DefaultLogLevel);
-        private const int DefaultLogMessageRateLimit = 60;
+        // By default, we don't rate limit log messages;
+        private const int DefaultLogMessageRateLimit = 0;
         private const LogEventLevel DefaultLogLevel = LogEventLevel.Information;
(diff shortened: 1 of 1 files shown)


src/Datadog.Trace/Logging/DatadogLogging.cs
```diff
@@ -156,8 +156,8 @@ private static int GetRateLimit()
                 return rate;
             }
 
-            // We don't want to rate limit messages by default when in debug mode
-            return GlobalSettings.Source.DebugEnabled ? 0 : DefaultLogMessageRateLimit;
```
lucaspimentel, line 160: Was `DefaultLogMessageRateLimit` not configurable anywhere?

andrewlock: That's a good point, I should remove it. The _default_ Log Message rate limit was a constant, 60s. You could override that (and you still can). Alternatively, I can change this to `return DefaultLogMessageRateLimit` and set `DefaultLogMessageRateLimit = 0`.

lucaspimentel: Yeah, I was thinking we should just default `DD_TRACE_LOGGING_RATE` to `0`. What do you think? I also just noticed [it only works through env vars](https://github.com/DataDog/dd-trace-dotnet/blob/e51c8d2cfd396df40d363d14e61fe93cee3b0f14/src/Datadog.Trace/Logging/DatadogLogging.cs#L151) and not other config sources.

andrewlock: > I also just noticed it only works through env vars and not other config sources.

😬 

> Yeah, I was thinking we should just default DD_TRACE_LOGGING_RATE to 0.

Sure, I don't mind either way. Though doesn't make sense to have the `DebugEnabled` conditional any more I don't think

lucaspimentel: Agreed. Thanks!

andrewlock: > I also just noticed it only works through env vars and not other config sources.

Hmmm, there's _lots_ of things in there that only load from EnvVariables 🤔 `MaxLogFileSize`, `LogDirectory`, `ProfilerLogPath`... Should they be loading from other sources too?

andrewlock: > Hmmm, there's lots of things in there that only load from EnvVariables

I think it's because we need the logger _before_ we create the configuration sources? We could add them to the `GlobalSettings` collection instead, but that doesn't look very "standard"?

lucaspimentel: Let's just disable log rate limiting by default first (this PR as it stands). Can we add the rest to the backlog for later?

lucaspimentel (approved):

zacharycmontoya (approved):

PR #1310, opened 2021-03-24
Add abstraction for Outbound HTTP span names and attributes

zacharycmontoya:
Port changes from OTEL PR: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/63

This PR adds an abstraction layer `IOutboundHttpConvention` to allow the Tracer to configure tags based on the runtime configuration.

This change currently keeps the OTEL-semantics for decorating HTTP client spans.

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace/Configuration/ConventionType.cs
+++ b/src/Datadog.Trace/Configuration/ConventionType.cs
@@ -1,0 +1,23 @@
+namespace Datadog.Trace.Configuration
+{
+    /// <summary>
+    /// Semantic convention used to when defining operation names, span tags, statuses etc.
+    /// </summary>
+    public enum ConventionType
+    {
+        /// <summary>
+        /// The default convention. Currently it is Datadog (sic!).
+        /// </summary>
+        Default,
+
+        /// <summary>
+        /// The OpenTelemetry convention.
+        /// </summary>
+        OpenTelemetry,
+
+        /// <summary>
+        /// The Datadog convention. Current default.
+        /// </summary>
+        Datadog,
+    }
+}
--- a/src/Datadog.Trace/Configuration/TracerSettings.cs
+++ b/src/Datadog.Trace/Configuration/TracerSettings.cs
@@ -170,6 +170,9 @@ namespace Datadog.Trace.Configuration
                                            "500-599";
             HttpServerErrorStatusCodes = ParseHttpCodesToArray(httpServerErrorStatusCodes);
 
+            Enum.TryParse(source?.GetString(ConfigurationKeys.Convention) ?? "default", ignoreCase: true, out ConventionType conventionType);
+            Convention = conventionType;
+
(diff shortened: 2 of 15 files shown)


tonyredondo:
We need to check both benchmarks and throughput tests for this PR, it could lead to performance regressions.

src/Datadog.Trace/Configuration/ConventionType.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        /// <summary>
+        /// The OpenTelemetry convention.
+        /// </summary>
+        OpenTelemetry,
+
+        /// <summary>
+        /// The Datadog convention. Current default.
+        /// </summary>
+        Datadog,
```
lucaspimentel, line 21: Shouldn't we use something like `Datadog = Default` so they have the same value?
```suggestion
        /// <summary>
        /// The Datadog convention. Same as <see cref="Default"/>.
        /// </summary>
        Datadog = Default
```

src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -170,6 +170,9 @@ public TracerSettings(IConfigurationSource source)
                                            "500-599";
             HttpServerErrorStatusCodes = ParseHttpCodesToArray(httpServerErrorStatusCodes);
 
+            Enum.TryParse(source?.GetString(ConfigurationKeys.Convention) ?? "default", ignoreCase: true, out ConventionType conventionType);
```
lucaspimentel, line 173: As it is written, a setting of `"OpenTelemetry, Datadog"` will result in `(ConventionType)3`, which is invalid.

(`OpenTelemetry` is 1, `Datadog` is 2, and `TryParse()` will use bitwise `OR` even if the enum type doesn't have `[Flags]`)

src/Datadog.Trace/Conventions/OtelHttpTags.cs
```diff
@@ -0,0 +1,18 @@
+using Datadog.Trace.ExtensionMethods;
+using Datadog.Trace.Tagging;
+
+namespace Datadog.Trace.Conventions
+{
+    internal class OtelHttpTags : HttpTags
```
lucaspimentel, line 6: Do we want this in our repo?

src/Datadog.Trace/Conventions/OutboundHttpArgs.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// </summary>
+        public readonly string HttpMethod;
+
+        /// <summary>
+        /// Request's URI.
+        /// </summary>
+        public readonly Uri RequestUri;
+
+        /// <summary>
+        /// Request's URI.
```
lucaspimentel, line 28: Wrong comment for `IntegrationInfo`

zacharycmontoya:
Closing along with https://github.com/DataDog/dd-trace-dotnet/pull/1315 . If there's strong motivation to share code paths with OTEL, we can reopen this.

PR #1311, opened 2021-03-25
Partial OTEL port to improve TagsList

zacharycmontoya:
I looked to port https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/71

However, I don't think we necessarily need to take the changes from there, which can be summed up as adding a Metrics and Tags property on the ITags interface, to allow their Zipkin exporter to write the tags/metrics.

In the end, I just ported one unit test from the PR and modified it to work with our structure.

@DataDog/apm-dotnet

### DIFF
--- a/test/Datadog.Trace.Tests/Tagging/TagsListTests.cs
+++ b/test/Datadog.Trace.Tests/Tagging/TagsListTests.cs
@@ -14,29 +14,36 @@ namespace Datadog.Trace.Tests.Tagging
     public class TagsListTests
     {
         [Fact]
-        public void SetTag_WillNotCauseDuplicates()
+        public void GetTag_GetMetric_ReturnUpdatedValues()
         {
-            // Initialize common tags
-            var tags = new DerivedCommonTags()
+            var tags = new CommonTags();
+            var span = new Span(new SpanContext(42, 41), DateTimeOffset.UtcNow, tags);
+
+            tags.Environment = "Test";
+            tags.SamplingLimitDecision = 0.5;
+
+            // Override the properties
+            span.SetTag(Tags.Env, "Overridden Environment");
+            span.SetMetric(Metrics.SamplingLimitDecision, 0.75);
+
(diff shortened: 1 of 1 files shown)


andrewlock (approved):

test/Datadog.Trace.Tests/Tagging/TagsListTests.cs
```diff
@@ -5,13 +5,40 @@
...
+
+            // Initialize custom tags
+            tags.SetTag("sample.1", "Temp 1");
+            tags.SetTag("sample.2", "Temp 2");
+
+            // Try set existing tag
+            tags.SetTag(Tags.Version, "v2.0");
+            tags.SetTag("sample.2", "Temp 3");
+
+            var all = tags.GetAllTags();
```
kevingosse, line 34: Rather than adding this method, wouldn't it be better to:
 - call `GetTag` and make sure the new value of the tag is returned
 - call `SerializeTo` and check that the tag is serialized only once

In the end, we don't _really_ care if the value is stored internally once or n times (though it's still nice to confirm it), what matters the most is the value returned by the public APIs

zacharycmontoya: Yes very good point, I'll modify the test to do that

zacharycmontoya: Updated in be93545517654d92f5ccb123509b17996a9b578f . I made a slight pivot and focused this new unit test on GetTag and GetMetric. I'm leaving the serialization test to the Serialize unit test. What do you think @kevingosse ?

kevingosse (approved):

test/Datadog.Trace.Tests/Tagging/TagsListTests.cs
kevingosse: LGTM 👍

PR #1313, opened 2021-03-25
Add support for flushing partial traces

kevingosse:
Add a way to partially flush large traces. The partial flush is opt-in, and triggered when the number of closed spans in a trace gets above a given threshold.

New configuration keys:
 - `DD_TRACER_PARTIAL_FLUSH_ENABLED`: enables the partial flush
 - `DD_TRACER_PARTIAL_FLUSH_MIN_SPANS`: changes the threshold (default is 500 closed spans in a trace)

### DIFF
--- a/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
+++ b/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
@@ -305,12 +305,14 @@ namespace Datadog.Trace.Configuration
         /// <summary>
         /// Configuration key to enable sending partial traces to the agent
         /// </summary>
-        public const string PartialFlushEnabled = "DD_TRACER_PARTIAL_FLUSH_ENABLED";
+        /// <seealso cref="TracerSettings.PartialFlushEnabled"/>
+        public const string PartialFlushEnabled = "DD_TRACE_PARTIAL_FLUSH_ENABLED";
 
         /// <summary>
         /// Configuration key to set the minimum number of closed spans in a trace before it's partially flushed
         /// </summary>
-        public const string PartialFlushMinSpans = "DD_TRACER_PARTIAL_FLUSH_MIN_SPANS";
+        /// <seealso cref="TracerSettings.PartialFlushMinSpans"/>
+        public const string PartialFlushMinSpans = "DD_TRACE_PARTIAL_FLUSH_MIN_SPANS";
 
         /// <summary>
         /// String format patterns used to match integration-specific configuration keys.
--- a/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
+++ b/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
@@ -303,6 +303,16 @@ namespace Datadog.Trace.Configuration
         public const string HttpClientErrorStatusCodes = "DD_HTTP_CLIENT_ERROR_STATUSES";
 
         /// <summary>
+        /// Configuration key to enable sending partial traces to the agent
+        /// </summary>
+        public const string PartialFlushEnabled = "DD_TRACER_PARTIAL_FLUSH_ENABLED";
+
+        /// <summary>
+        /// Configuration key to set the minimum number of closed spans in a trace before it's partially flushed
+        /// </summary>
+        public const string PartialFlushMinSpans = "DD_TRACER_PARTIAL_FLUSH_MIN_SPANS";
+
+        /// <summary>
         /// String format patterns used to match integration-specific configuration keys.
         /// </summary>
         public static class Integrations
--- a/src/Datadog.Trace/Configuration/TracerSettings.cs
+++ b/src/Datadog.Trace/Configuration/TracerSettings.cs
@@ -190,9 +190,12 @@ namespace Datadog.Trace.Configuration
                 // default value
                 ?? false;
 
-            PartialFlushMinSpans = source?.GetInt32(ConfigurationKeys.PartialFlushMinSpans)
-                // default value
-                ?? 500;
+            var partialFlushMinSpans = source?.GetInt32(ConfigurationKeys.PartialFlushMinSpans);
+
+            if ((partialFlushMinSpans ?? 0) <= 0)
+            {
+                PartialFlushMinSpans = 500;
+            }
         }
 
         /// <summary>
--- a/src/Datadog.Trace/Configuration/TracerSettings.cs
+++ b/src/Datadog.Trace/Configuration/TracerSettings.cs
@@ -24,6 +24,8 @@ namespace Datadog.Trace.Configuration
         /// </summary>
         public const int DefaultAgentPort = 8126;
 
+        private int _partialFlushMinSpans;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="TracerSettings"/> class with default values.
         /// </summary>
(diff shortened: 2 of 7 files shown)


src/Datadog.Trace/TraceContext.cs
```diff
@@ -104,9 +105,10 @@ public void CloseSpan(Span span)
 
             lock (_spans)
             {
+                _spans.Add(span);
```
zacharycmontoya, line 108: @lucaspimentel do the spans need to be ordered in any way? My thinking is no but I want to double check. This would basically add the root span as the last element instead of the first element in the previous implementation

kevingosse: That's a good point, gonna check how the agent processes them

src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -302,6 +302,16 @@ public static class ConfigurationKeys
...
 
+        /// <summary>
+        /// Configuration key to enable sending partial traces to the agent
+        /// </summary>
+        public const string PartialFlushEnabled = "DD_TRACER_PARTIAL_FLUSH_ENABLED";
+
+        /// <summary>
+        /// Configuration key to set the minimum number of closed spans in a trace before it's partially flushed
+        /// </summary>
+        public const string PartialFlushMinSpans = "DD_TRACER_PARTIAL_FLUSH_MIN_SPANS";
```
zacharycmontoya, line 313:
```suggestion
        public const string PartialFlushMinSpans = "DD_TRACE_PARTIAL_FLUSH_MIN_SPANS";
```

src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -302,6 +302,16 @@ public static class ConfigurationKeys
         /// <seealso cref="TracerSettings.HttpClientErrorStatusCodes"/>
         public const string HttpClientErrorStatusCodes = "DD_HTTP_CLIENT_ERROR_STATUSES";
 
+        /// <summary>
+        /// Configuration key to enable sending partial traces to the agent
+        /// </summary>
+        public const string PartialFlushEnabled = "DD_TRACER_PARTIAL_FLUSH_ENABLED";
```
zacharycmontoya, line 308:
```suggestion
        public const string PartialFlushEnabled = "DD_TRACE_PARTIAL_FLUSH_ENABLED";
```

src/Datadog.Trace/TraceContext.cs
kevingosse: `GetRoot` looks for the root in the whole trace, so that should be OK: https://github.com/DataDog/datadog-agent/blob/8f141e218ee5bd74d6a7742e73844220c5cda603/pkg/trace/traceutil/trace.go#L27

As a matter of fact, it starts looking by the end, so it looks like we are actually expected to push them in that order

src/Datadog.Trace/TraceContext.cs
```diff
@@ -104,9 +105,10 @@ public void CloseSpan(Span span)
...
                 _openSpans--;
 
-                if (_openSpans == 0)
+                if (_openSpans == 0 || ShouldTriggerPartialFlush())
```
zacharycmontoya, line 111: @kevingosse just to make sure I understand, by setting the "top level" metric on the correct span, the agent will accept these partial traces and there's nothing else we need to do? And the behavior exists on the 0.4 endpoint? Also, it's worth documenting on what version of the Agent this functionality was enabled, perhaps in the ConfigurationKey's

src/Datadog.Trace/TraceContext.cs
lucaspimentel: > do the spans need to be ordered in any way? 

Not that I know of. 

> it starts looking by the end

Interesting. I wonder how much this affects Trace Agent performance (having tracers add the root span at the start vs end of the array).

src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -302,6 +302,16 @@ public static class ConfigurationKeys
         /// <seealso cref="TracerSettings.HttpClientErrorStatusCodes"/>
         public const string HttpClientErrorStatusCodes = "DD_HTTP_CLIENT_ERROR_STATUSES";
 
+        /// <summary>
+        /// Configuration key to enable sending partial traces to the agent
+        /// </summary>
```
andrewlock, line 307:
```suggestion
        /// </summary>
        /// <seealso cref="TracerSettings.PartialFlushEnabled"/>
```

src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -302,6 +302,16 @@ public static class ConfigurationKeys
...
+        public const string PartialFlushEnabled = "DD_TRACER_PARTIAL_FLUSH_ENABLED";
+
+        /// <summary>
+        /// Configuration key to set the minimum number of closed spans in a trace before it's partially flushed
+        /// </summary>
```
andrewlock, line 312:
```suggestion
        /// </summary>
        /// <seealso cref="TracerSettings.PartialFlushMinSpans"/>
```

src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -350,6 +360,28 @@ public bool DiagnosticSourceEnabled
...
+        /// Gets or sets the minimum number of closed spans in a trace before it's partially flushed
+        /// </summary>
+        public int PartialFlushMinSpans
+        {
+            get => _partialFlushMinSpans;
+            set
+            {
+                if (value <= 0)
+                {
+                    throw new ArgumentException("The value must be strictly greater than 0", nameof(PartialFlushMinSpans));
```
andrewlock, line 378: Not just greater, **strictly** greater 😄

andrewlock (approved):
Nice 🙂

src/Datadog.Trace/TraceContext.cs
kevingosse: That's correct.

And yes, I will add an agent version check before merging

kevingosse: Actually, if that's OK with you, I'm going to merge the PR as-is and add the agent check in a separate PR.
The agent check requires a version of the agent that hasn't been released yet, so I'd like to publish a pre-release before implementing it.

src/Datadog.Trace/Configuration/TracerSettings.cs
lucaspimentel: Hmm, this is new. I don't think we throw on invalid settings anywhere else. Part of me things that we should throw when users are setting this from code, but if a user sets env var `DD_TRACE_PARTIAL_FLUSH_MIN_SPANS=-1`, their app may crash during startup (even if `DD_TRACE_PARTIAL_FLUSH_ENABLED=false`).

Should we consider logging this and setting the default value instead? Or changing the logic so it only throw when setting from code but not from other sources?

src/Datadog.Trace/TraceContext.cs
```diff
@@ -13,7 +13,7 @@ internal class TraceContext : ITraceContext
 
         private readonly DateTimeOffset _utcStart = DateTimeOffset.UtcNow;
         private readonly long _timestamp = Stopwatch.GetTimestamp();
-        private readonly List<Span> _spans = new List<Span>();
+        private List<Span> _spans = new List<Span>();
```
lucaspimentel, line 16: Doesn't look like we needed to remove the `readonly`.

kevingosse: Right, that because I started by doing something much more complicated 😅  gonna restore it

src/Datadog.Trace/Configuration/TracerSettings.cs
kevingosse: In https://github.com/DataDog/dd-trace-dotnet/pull/1313/commits/a02af7372630a04734e42ac8718265848fa3b8b2, I changed that to not throw if the value is set from the configuration

lucaspimentel: Thanks.

lucaspimentel (approved):

zacharycmontoya (approved):
🎉

tonyredondo (approved):
LGTM!

colin-higgins (approved):

PR #865, opened 2020-08-18
Skip DNS resolution when agent is bound to localhost

kevingosse:
Attempting to re-merge https://github.com/DataDog/dd-trace-dotnet/pull/841 with fixes for failing tests

PR #1314, opened 2021-03-25
Listen on localhost instead of 127.0.0.1

andrewlock:
Previously I was seeing `HttpListenerException : The network location cannot be reached` when running the tests locally (on Windows). This seems to fix the issue, though I haven't tested it on Linux/mac

@DataDog/apm-dotnet

### DIFF
--- a/test/Datadog.Trace.TestHelpers/MockTracerAgent.cs
+++ b/test/Datadog.Trace.TestHelpers/MockTracerAgent.cs
@@ -60,7 +60,6 @@ namespace Datadog.Trace.TestHelpers
                 // seems like we can't reuse a listener if it fails to start,
                 // so create a new listener each time we retry
                 var listener = new HttpListener();
-                listener.Prefixes.Add($"http://127.0.0.1:{port}/");
                 listener.Prefixes.Add($"http://localhost:{port}/");
 
                 try
--- a/test/Datadog.Trace.Tests/DogStatsDTests.cs
+++ b/test/Datadog.Trace.Tests/DogStatsDTests.cs
@@ -110,7 +110,7 @@ namespace Datadog.Trace.Tests
             {
                 var settings = new TracerSettings
                 {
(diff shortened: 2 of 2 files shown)


lucaspimentel:
I had this same error for a while, then I figured out it was some bad config on my machine. Can't remember what it was, though 😞

lucaspimentel:
I also seem to remember something about `localhost` being several seconds slower (if IPv6 was enabled?), which is why we switched to `127.0.0.1`. IIRC, @kevingosse figured that one out, so he may remember better than me.

kevingosse:
> I also seem to remember something about `localhost` being several seconds slower (if IPv6 was enabled?), which is why we switched to `127.0.0.1`. IIRC, @kevingosse figured that one out, so he may remember better than me.
> 
> edit: see #841 and #865

That's correct for the tracer (agent is bound to ipv4 only, and "localhost" first resolves to ipv6, which adds a 2 seconds delay at the connection). And then I had to do the same change to the MockAgent because somehow it broke the tests (as you can see in the integration tests of this PR 😛 ). I'm don't know why though. In theory it should be fine to have the MockAgent listen to localhost.

andrewlock:
For the record, I ran 

```
netsh http show iplisten
```

and found this:

```
IP addresses present in the IP listen list:
-------------------------------------------

    ::
```

Which I remember adding to try and work around an IIS express issues a while back. I removed the entry using 

```
netsh http delete iplisten ::
```

and bingo, the tests pass again.

Thanks all!

> And then I had to do the same change to the MockAgent because somehow it broke the tests
That's a whole lot of weird 🤷

lucaspimentel:
> I removed the entry [...] and bingo, the tests pass again.

Interesting! I have:
```
IP addresses present in the IP listen list:
-------------------------------------------

    ::
    127.0.0.1
```
Looks like I was in the same state as you (only `::`) and then I added `127.0.0.1`. I'll try removing both. Seems cleaner.

PR #1315, opened 2021-03-25
Port OTEL refactor

zacharycmontoya:
Port changes from OTEL PR's:
- open-telemetry/opentelemetry-dotnet-instrumentation#74
- open-telemetry/opentelemetry-dotnet-instrumentation#78

PR 74 introduces the following refactorings, with 78 adding small changes. The refactorings include:
1. Renaming `IAgentWriter` to `ITraceWriter`
2. Refactoring `AgentWriter` to use an `IMetrics` object with a default `NullMetric` implementation, which helps remove null checks everywhere it's used

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace/Abstractions/IMetrics.cs
+++ b/src/Datadog.Trace/Abstractions/IMetrics.cs
@@ -1,0 +1,14 @@
+namespace Datadog.Trace.Abstractions
+{
+    internal interface IMetrics
+    {
+        /// <summary>
+        /// Increments the specified counter.
+        /// </summary>
+        /// <param name="statName">The name of the metric.</param>
+        /// <param name="value">The amount of increment.</param>
+        /// <param name="sampleRate">Percentage of metric to be sent.</param>
+        /// <param name="tags">Array of tags to be added to the data.</param>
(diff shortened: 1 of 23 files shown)


andrewlock (commented):
These changes all make sense, but unfortunately they're breaking changes for us I think, as they change "public" `internal` methods 🙁 Probably need to add this to the 2.0 list...

zacharycmontoya:
> These changes all make sense, but unfortunately they're breaking changes for us I think, as they change "public" `internal` methods 🙁 Probably need to add this to the 2.0 list...

I initially thought that these changes were fine because at this point in time of the repo, `Datadog.Trace.ClrProfiler.Managed` doesn't call any of the internal methods I've modified here, but now I realize that it's possible that *older* versions of `Datadog.Trace.ClrProfiler.Managed` might have these calls. I'm thinking I can address this during the game day to identify all the various callsites that expose these mismatching issues, so I'll follow up on this PR when I have more information.

zacharycmontoya:
Closing because I think we're at the point now where the OTEL instrumentation repo and the Datadog repo are diverging in the managed code paths. Will re-open if there's strong motivation

PR #1316, opened 2021-03-25
Add Thread.Sleep for CallTargetNativeTest to avoid seg faults

zacharycmontoya:
This works around an issue we've seen before in netcoreapp2.1 tests.

@DataDog/apm-dotnet

### DIFF
--- a/test/test-applications/instrumentation/CallTargetNativeTest/Program.cs
+++ b/test/test-applications/instrumentation/CallTargetNativeTest/Program.cs
@@ -85,6 +85,12 @@ namespace CallTargetNativeTest
                     Console.WriteLine("Run with the profiler and use a number from 0-9/all as an argument.");
                     break;
             }
+
+#if NETCOREAPP2_1
+            // Add a delay to avoid a race condition on shutdown: https://github.com/dotnet/coreclr/pull/22712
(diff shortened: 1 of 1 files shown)


lucaspimentel (approved):

andrewlock (approved):

zacharycmontoya:
I rebased just to catch up to latest master so I cancelled the `Crank` and `benchmarks` runs

tonyredondo:
I think we need to collect the mem dump for this one. 

The segfault seems different from this fix: https://github.com/DataDog/dd-trace-dotnet/blob/4541013aa89cc5fc2c8947fb03f28a5854409726/src/Datadog.Trace.ClrProfiler.Native/cor_profiler_base.h#L219

tonyredondo (changes requested):
Can we get a memory dump before merging this?

zacharycmontoya:
We've started collecting dumps in CI on April 7, 2021 via https://github.com/DataDog/dd-trace-dotnet/pull/1329, but the test has not failed since then, according to Azure DevOps. I will keep this PR open until I can get a dump for us to look at.

zacharycmontoya:
Closing because I haven't seen this test fail in the past 14 days, so the fix is unneeded

PR #1317, opened 2021-03-25
CIApp: Testing frameworks integrations test suite and fixes

tonyredondo:
This PR includes:

- Fixes on scenarios where we were loosing spans.
- Adds integration tests for each supported testing framework and version.
- Refactor due the fixes to share similar code between testing framework integrations.

@DataDog/apm-dotnet


Datadog.Trace.proj
```diff
@@ -80,8 +80,7 @@
     </MSBuild>
   </Target>
 
-  <Target Name="RestoreAndBuildSamplesForPackageVersions" Condition="'$(TestAllPackageVersions)'=='true'">
-
+  <Target Name="RestoreAndBuildCsharpProjects">
```
andrewlock, line 83: Urgh, git sucks for tracking this stuff - it looks like you've split the  `RestoreAndBuildSamplesForPackageVersions` target into two, but then you're calling them both twice anyway in the build scripts? Which AFAICT, will result in `RestoreAndBuildCsharpProjects`, `RestoreAndBuildCsharpProjects`, and `RestoreSamplesForPackageVersions` all being called twice? That seems unnecessary? 🤔

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/Common.cs
```diff
@@ -13,12 +19,58 @@ static Common()
             TestTracer = new Tracer(settings);
             ServiceName = TestTracer.DefaultServiceName;
 
+            Tracer.Instance = TestTracer;
```
andrewlock, line 22: Should we really be "resetting" the `Tracer.Instance` like this? What if this module is initialized (for some reason) in non-testing code? We'd overwrite the static tracer instance then, or alternatively overwrite `Tracer.Instance` 🤔

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/Common.cs
```diff
@@ -13,12 +19,58 @@ static Common()
...
+                    // We have to ensure the flush of the buffer after we finish the tests of an assembly.
+                    // For some reason, sometimes when all test are finished none of the callbacks to handling the tracer disposal is triggered.
+                    // So the last spans in buffer aren't send to the agent.
+                    Log.Debug("Integration flushing spans.");
+                    await TestTracer.FlushAsync().ConfigureAwait(false);
+                    // The current agent writer FlushAsync method can return inmediately if a payload is being sent (there is buffer lock)
+                    // There is not api in the agent writer that guarantees the send has been sucessfully completed.
+                    // Until we change the behavior of the agentwriter we should at least wait 2 seconds before returning.
+                    Log.Debug("Waiting 2 seconds to flush.");
+                    await Task.Delay(2000).ConfigureAwait(false);
```
andrewlock, line 66: This is only flushing on _assembly_ unload right? Not for individual tests/theorys? 2s seems like a reasonable trade off at the assembly level, but not at the test level!

Also though, 2s is pretty arbitrary, is it worth us adding an API to the agent writer to e.g. only return after all buffers are flushed? Or alternatively, if we're not actually closing the Tracer, do we _have_ to make sure everything's flushed at this point?

test/Datadog.Trace.TestHelpers/EnvironmentHelper.cs
```diff
@@ -150,7 +150,8 @@ public static void ClearProfilerEnvironmentVariables()
             int aspNetCorePort,
             int? statsdPort,
             string processPath,
-            StringDictionary environmentVariables)
+            StringDictionary environmentVariables,
+            bool ignoringProfilerProcesses = false)
```
andrewlock, line 154: nit: should this be a command, ie. `ignoreProfilerProcesses`?

test/test-applications/integrations/Samples.XUnitTests/xunit.runner.json
```diff
@@ -0,0 +1,5 @@
+﻿{
+  "shadowCopy": false,
+  "diagnosticMessages": true,
+  "internalDiagnosticMessages": true
```
andrewlock, line 4: Is it worth using one of the xunit method rename options in here, to make sure we handle that correctly? https://xunit.net/docs/configuration-files#methodDisplay

andrewlock (approved):
LGTM AFAICT, added a few comments/questions :)

Datadog.Trace.proj
tonyredondo: Well I thought the same... but @zacharycmontoya helped me with this one...

His commit message: 

>The issue with the XUnit test was that it relied on importing the file xunit.runner.visualstudio/<version>/build/<TFM>/xunit.runner.visualstudio.props (I'm guessing to place xunit.runner.visualstudio.dotnetcore.testadapter.dll in the output directory), but it was not imported because the Import evaluation for the Publish MSBuild Task occurred before the Restore MSBuild Task finished. The unoptimized solution for now is to invoke MSBuild twice. This requires two separate MSBuild processes to start up and might cause accidental rebuilds, but it works for now.

I'm open to a better approach for this, but the problem is publishing the XUnit test suite doesn't publish correctly the test adapter, so trying to run the tests will fail.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/Common.cs
tonyredondo: Mmm I have to overwrite the tracer instance for integration tests, so new spans are created as children of the test spans. This is required also for billing so those children spans will not be billed by APM. 

But you're right, having this code in a `.cctor` is dangerous, I'll move this to a method and call it before the test span generation, so we ensure we are in a testing scope.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/Common.cs
tonyredondo: > This is only flushing on assembly unload right? Not for individual tests/theorys? 2s seems like a reasonable trade off at the assembly level, but not at the test level!

Yes this for assembly unload only.

> Also though, 2s is pretty arbitrary, is it worth us adding an API to the agent writer to e.g. only return after all buffers are flushed? Or alternatively, if we're not actually closing the Tracer, do we have to make sure everything's flushed at this point?

I talked with @kevingosse a little bit about supporting this return after flush on the agent writer, this also will fix some scenarios where `ProcessExit` and `CancelKeyPress` events are involved. In some cases we don't close the tracer and only call for flushing, but we don't have guarantees that buffers are actually flushed after that call, and the process will close after.

We need to make sure the buffer is flushed at this point because I found that the testhost can finish without rising neither the `ProcessExit` and `CancelKeyPress` events.

test/test-applications/integrations/Samples.XUnitTests/xunit.runner.json
tonyredondo: good point, I'll check.

Datadog.Trace.proj
andrewlock: Yikes, what a mess. This may well explain weird behaviours I've seen in other places. That's what we get for trying to use MSBuild to do fancy stuff 🙁 I'll have to work out how this is going to play with the Nuke branch at some point...

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/Common.cs
tonyredondo: done in 61f4f30

test/test-applications/integrations/Samples.XUnitTests/xunit.runner.json
tonyredondo: Trying `method` in f91a299. I think if it everything passes I will revert to `classAndMethod` as the default one.

test/Datadog.Trace.TestHelpers/EnvironmentHelper.cs
tonyredondo: I didn't do a method for this just because the default behavior is to set `DD_PROFILER_PROCESSES` env var and I want to skip that in this case. If I want to do that in a method I need to store that state and check it in this method as well. It was simpler adding this default argument.

andrewlock: Sorry, I wasn't clear here, I actually just meant renaming the var 🙂  e.g.
```suggestion
            bool ignoreProfilerProcesses = false)
```

tonyredondo: oh sorry I misunderstood that completely 😅. accepting your suggestion. Thank you!

andrewlock: Don't forget to update the usages in the method too 😉

test/test-applications/integrations/Samples.XUnitTests/xunit.runner.json
tonyredondo: fixed in 698a77a

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/XUnit/XUnitIntegration.cs
```diff
@@ -17,7 +23,7 @@ internal static Scope CreateScope(ref TestRunnerStruct runnerInstance, Type targ
 
             string testFramework = "xUnit " + testInvokerAssemblyName.Version.ToString();
 
-            Scope scope = Common.TestTracer.StartActive("xunit.test", serviceName: Common.ServiceName);
+            Scope scope = Common.TestTracer.StartActive("xunit.test", serviceName: Common.TestTracer.DefaultServiceName);
```
zacharycmontoya, line 26: Are there any remaining uses of `Common.ServiceName`? Maybe we can remove that? Or redefine the property to always return `TestTracer.DefaultServiceName`?

zacharycmontoya: My apologies, the GitHub UI didn't refresh so I didn't see the latest code. Looks like you already did what I was asking! Resolving.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/MsTestV2/MsTestIntegration.cs
```diff
@@ -0,0 +1,158 @@
...
+            string testSuite = testMethodInfo.TestClassName;
+            string testName = testMethodInfo.TestMethodName;
+
+            Scope scope = Common.TestTracer.StartActive("mstest.test", serviceName: Common.TestTracer.DefaultServiceName);
+            Span span = scope.Span;
+
+            span.Type = SpanTypes.Test;
+            span.SetTraceSamplingPriority(SamplingPriority.AutoKeep);
+            span.ResourceName = $"{testSuite}.{testName}";
+            span.SetTag(TestTags.Suite, testSuite);
```
zacharycmontoya, line 36: This seems like an appropriate time to create a new `TestTags` type that implements `ITags`, but that doesn't need to happen in this PR

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/MsTestV2/TestMethodAttributeExecuteIntegration.cs
```diff
@@ -47,7 +51,7 @@ public class TestMethodAttributeExecuteIntegration
         /// <returns>A response value, in an async scenario will be T of Task of T</returns>
         public static CallTargetReturn<TReturn> OnMethodEnd<TTarget, TReturn>(TTarget instance, TReturn returnValue, Exception exception, CallTargetState state)
         {
-            if (Common.TestTracer.Settings.IsIntegrationEnabled(IntegrationId))
+            if (MsTestIntegration.IsEnabled)
             {
                 Scope scope = Common.TestTracer.ActiveScope;
```
zacharycmontoya, line 56: Should this be using `state.Scope`, since you are setting it in `OnMethodBegin`?

src/Datadog.Trace.ClrProfiler.Managed/Integrations/Testing/NUnitIntegration.cs
```diff
@@ -342,20 +183,148 @@ private static void FinishScope(Scope scope, object testExecutionContext, Except
             }
 
             execute(workShift);
-            SynchronizationContext context = SynchronizationContext.Current;
+            AutoInstrumentation.Testing.Common.FlushSpans();
```
zacharycmontoya, line 186: Should this use the FlushSpans overload that takes an integration?

src/Datadog.Trace.ClrProfiler.Managed/Integrations/Testing/NUnitIntegration.cs
```diff
@@ -342,20 +183,148 @@ private static void FinishScope(Scope scope, object testExecutionContext, Except
...
+                    mdToken: mdToken,
+                    opCode: opCode,
+                    instrumentedType: NUnitTestAssemblyRunnerType,
+                    methodName: WaitForCompletionMethod,
+                    instanceType: testAssemblyRunnerType.AssemblyQualifiedName);
+                throw;
+            }
+
+            object result = execute(testAssemblyRunner, timeout);
+            AutoInstrumentation.Testing.Common.FlushSpans();
```
zacharycmontoya, line 240: Should this use the FlushSpans overload that takes an integration?

test/Datadog.Trace.IntegrationTests/Properties/launchSettings.json
```diff
@@ -0,0 +1,27 @@
+{
```
zacharycmontoya, line 1: I think we can delete this file

zacharycmontoya (approved):
Overall looks good! I left some small comments but nothing blocking

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/MsTestV2/MsTestIntegration.cs
tonyredondo: Yeah all the `ITags` work came after the test frameworks integrations, I'll update the implementation in another PR.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/MsTestV2/TestMethodAttributeExecuteIntegration.cs
tonyredondo: Good catch, fixed in 45506f2

src/Datadog.Trace.ClrProfiler.Managed/Integrations/Testing/NUnitIntegration.cs
tonyredondo: Yes, fixed in 45506f2

src/Datadog.Trace.ClrProfiler.Managed/Integrations/Testing/NUnitIntegration.cs
tonyredondo: Yes, fixed in 45506f2

test/Datadog.Trace.IntegrationTests/Properties/launchSettings.json
tonyredondo: done in 19331fa

PR #1318, opened 2021-03-26
Enable JetBrains Tools in the Benchmarks projects

tonyredondo:
This PR enables the JetBrains dotTrace and dotMemory tools for the benchmark tests.
@DataDog/apm-dotnet

### DIFF
--- a/test/benchmarks/Benchmarks.Trace/Benchmarks.Trace.csproj
+++ b/test/benchmarks/Benchmarks.Trace/Benchmarks.Trace.csproj
@@ -28,6 +28,7 @@
 
   <ItemGroup>
     <PackageReference Include="BenchmarkDotNet" Version="0.12.1" />
+    <PackageReference Include="JetBrains.Profiler.Api" Version="1.1.7" />
     <PackageReference Include="System.Net.Http" Version="4.3.4" />
   </ItemGroup>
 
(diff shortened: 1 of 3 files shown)


test/benchmarks/Benchmarks.Trace/Benchmarks.Trace.csproj
```diff
@@ -28,6 +28,7 @@
 
   <ItemGroup>
     <PackageReference Include="BenchmarkDotNet" Version="0.12.1" />
+    <PackageReference Include="JetBrains.Profiler.Api" Version="1.1.7" />
```
lucaspimentel, line 31: I've never used it before, but it looks like `JetBrains.Profiler.SelfApi` doesn't require other software (like dotTrace and dotMemory) to be installed on the machine. What do you think of using that instead?

lucaspimentel, line 31: We should add this to `LICENSE-3rdparty.csv`.

tonyredondo: I didn't use that because I don't know if could cause a license problem.

tonyredondo: done in 5e27760

lucaspimentel (approved):

test/benchmarks/Benchmarks.Trace/Benchmarks.Trace.csproj
zacharycmontoya: I re-opened this because I'm surprised we need to include this in our 3rd party license file, as we're not redistributing their code. Do we really need to add that?

zacharycmontoya (approved):

test/benchmarks/Benchmarks.Trace/Benchmarks.Trace.csproj
lucaspimentel: No, I think Zach is right. I just had this same conversation with @andrewlock in another PR. Just me being too aggressive about `LICENSE-3rdparty.csv`.

PR #1324, opened 2021-03-30
Reduce closures and allocations

tonyredondo:
This PR contains the following optimizations:
1. Reduce allocations by:
1.1. Converting some class to readonly struct.
1.2 Removing array creation on the caller caused by the `params` keyword.
1.3 Reducing the number of closures.
1.4 Check if IEnumerables are arrays to avoid the GetEnumerator call
1.5 Use the constraint trick on SpanContextPropagator.ParseString to avoid boxing on IHeaderCollections structs

2. Adds a new `ThrowHelper` class to remove the exception throw from possible hotpaths, and increase the chance of inlining. (try/catch blocks and throws disables the inlining of a method)

Benchmarks:

image
image

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace/Agent/AgentWriter.cs
+++ b/src/Datadog.Trace/Agent/AgentWriter.cs
@@ -122,6 +122,29 @@ namespace Datadog.Trace.Agent
 
             var delay = Task.Delay(TimeSpan.FromSeconds(20));
 
+#if NET5_0_OR_GREATER
+            var completedTask = await Task.WhenAny(_serializationTask, delay).ConfigureAwait(false);
+
+            if (completedTask != delay)
+            {
+                await Task.WhenAny(_flushTask, Task.Delay(TimeSpan.FromSeconds(20))).ConfigureAwait(false);
+
+                if (_frontBuffer.TraceCount == 0 && _backBuffer.TraceCount == 0)
+                {
+                    // All good
+                    return;
+                }
+
+                // In some situations, the flush thread can exit before flushing all the threads
+                // Force a flush for the leftover traces
+                completedTask = await Task.WhenAny(Task.Run(() => FlushBuffers(flushAllBuffers: true)), delay).ConfigureAwait(false);
+
+                if (completedTask != delay)
+                {
+                    return;
+                }
+            }
+#else
             // Internally WhenAny will make a defensive copy of the array
             // so we reuse the outer one.
             // https://source.dot.net/#System.Private.CoreLib/Task.cs,6028
@@ -146,7 +169,7 @@ namespace Datadog.Trace.Agent
 
                 // In some situations, the flush thread can exit before flushing all the threads
                 // Force a flush for the leftover traces
-                whenAnyArray[0] = Task.Factory.StartNew(state => ((AgentWriter)state).FlushBuffers(flushAllBuffers: true), this).Unwrap();
+                whenAnyArray[0] = Task.Run(() => FlushBuffers(flushAllBuffers: true));
                 whenAnyArray[1] = delay;
                 completedTask = await Task.WhenAny(whenAnyArray).ConfigureAwait(false);
 
@@ -155,6 +178,7 @@ namespace Datadog.Trace.Agent
                     return;
                 }
             }
+#endif
 
             Log.Warning("Could not flush all traces before process exit");
         }
@@ -167,7 +191,7 @@ namespace Datadog.Trace.Agent
                 // Enqueue a watermark to know when it's done serializing all currently enqueued traces
                 var tcs = new TaskCompletionSource<bool>(TaskOptions);
 
-                WriteWatermark(state => CompleteTaskCompletionSource((TaskCompletionSource<bool>)state), tcs);
+                WriteWatermark(() => CompleteTaskCompletionSource(tcs));
 
                 await tcs.Task.ConfigureAwait(false);
             }
@@ -175,9 +199,9 @@ namespace Datadog.Trace.Agent
             await FlushBuffers().ConfigureAwait(false);
         }
 
-        internal void WriteWatermark(Action<object> watermark, object state, bool wakeUpThread = true)
+        internal void WriteWatermark(Action watermark, bool wakeUpThread = true)
         {
-            _pendingTraces.Enqueue(new WorkItem(watermark, state));
+            _pendingTraces.Enqueue(new WorkItem(watermark));
 
             if (wakeUpThread)
             {
@@ -372,7 +396,7 @@ namespace Datadog.Trace.Agent
                         if (item.Trace == null)
                         {
                             // Found a watermark
-                            item.ExecuteCallback();
+                            item.Callback();
                             continue;
                         }
 
@@ -406,26 +430,18 @@ namespace Datadog.Trace.Agent
         private readonly struct WorkItem
         {
             public readonly Span[] Trace;
-            public readonly Action<object> Callback;
-            public readonly object State;
+            public readonly Action Callback;
 
             public WorkItem(Span[] trace)
             {
                 Trace = trace;
                 Callback = null;
-                State = null;
             }
 
-            public WorkItem(Action<object> callback, object state = null)
+            public WorkItem(Action callback)
             {
                 Trace = null;
                 Callback = callback;
-                State = state;
-            }
-
-            public void ExecuteCallback()
-            {
-                Callback?.Invoke(State);
             }
         }
     }
--- a/src/Datadog.Trace/Agent/AgentWriter.cs
+++ b/src/Datadog.Trace/Agent/AgentWriter.cs
@@ -146,7 +146,7 @@ namespace Datadog.Trace.Agent
 
                 // In some situations, the flush thread can exit before flushing all the threads
                 // Force a flush for the leftover traces
-                whenAnyArray[0] = Task.Factory.StartNew(state => ((AgentWriter)state).FlushBuffers(flushAllBuffers: true), this);
+                whenAnyArray[0] = Task.Factory.StartNew(state => ((AgentWriter)state).FlushBuffers(flushAllBuffers: true), this).Unwrap();
                 whenAnyArray[1] = delay;
                 completedTask = await Task.WhenAny(whenAnyArray).ConfigureAwait(false);
 
--- a/src/Datadog.Trace/Agent/AgentWriter.cs
+++ b/src/Datadog.Trace/Agent/AgentWriter.cs
@@ -122,13 +122,17 @@ namespace Datadog.Trace.Agent
 
             var delay = Task.Delay(TimeSpan.FromSeconds(20));
 
-            var completedTask = await Task.WhenAny(_serializationTask, delay)
-                .ConfigureAwait(false);
+            Task[] whenAnyArray = new Task[2];
+
+            whenAnyArray[0] = _serializationTask;
+            whenAnyArray[1] = delay;
+            var completedTask = await Task.WhenAny(whenAnyArray).ConfigureAwait(false);
 
             if (completedTask != delay)
             {
-                await Task.WhenAny(_flushTask, Task.Delay(TimeSpan.FromSeconds(20)))
-                    .ConfigureAwait(false);
+                whenAnyArray[0] = _flushTask;
+                whenAnyArray[1] = Task.Delay(TimeSpan.FromSeconds(20));
+                await Task.WhenAny(whenAnyArray).ConfigureAwait(false);
 
                 if (_frontBuffer.TraceCount == 0 && _backBuffer.TraceCount == 0)
                 {
@@ -138,8 +142,9 @@ namespace Datadog.Trace.Agent
 
                 // In some situations, the flush thread can exit before flushing all the threads
                 // Force a flush for the leftover traces
-                completedTask = await Task.WhenAny(Task.Run(() => FlushBuffers(flushAllBuffers: true)), delay)
-                    .ConfigureAwait(false);
+                whenAnyArray[0] = Task.Factory.StartNew(state => ((AgentWriter)state).FlushBuffers(flushAllBuffers: true), this);
+                whenAnyArray[1] = delay;
+                completedTask = await Task.WhenAny(whenAnyArray).ConfigureAwait(false);
 
                 if (completedTask != delay)
                 {
@@ -158,7 +163,7 @@ namespace Datadog.Trace.Agent
                 // Enqueue a watermark to know when it's done serializing all currently enqueued traces
                 var tcs = new TaskCompletionSource<bool>(TaskOptions);
 
-                WriteWatermark(() => CompleteTaskCompletionSource(tcs));
+                WriteWatermark(state => CompleteTaskCompletionSource((TaskCompletionSource<bool>)state), tcs);
 
                 await tcs.Task.ConfigureAwait(false);
             }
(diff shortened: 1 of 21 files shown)


src/Datadog.Trace/TracingProcessManager.cs
```diff
@@ -114,141 +114,145 @@ private static Task StartProcessWithKeepAlive(ProcessMetadata metadata)
             var path = metadata.ProcessPath;
             Log.Debug("Starting keep alive for {Process}.", path);
 
-            return Task.Run(
-                async () =>
+            return Task.Factory.StartNew(state => InternalStartProcessWithKeepAlive(state), metadata);
```
lucaspimentel, line 117: Should this use `TaskCreationOptions.LongRunning`? Similar to https://github.com/DataDog/dogstatsd-csharp-client/pull/151.

src/Datadog.Trace/Headers/NameValueHeadersCollection.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace Datadog.Trace.Headers
 {
-    internal class NameValueHeadersCollection : IHeadersCollection
+    internal readonly struct NameValueHeadersCollection : IHeadersCollection
```
lucaspimentel, line 8: Are these changes from `class` to `struct` breaking if called from an older version of `Datadog.Trace.ClrProfiler.Managed`?

tonyredondo: Definitely, returning an interface reference (boxing if steuct) vs returning a value. There’s a couple breaking changes related to this, but reduces heap allocations.

src/Datadog.Trace/TracingProcessManager.cs
tonyredondo: I guess it should. I was removing the closure, but you’re right.

kevingosse: The method is async, so there's no benefit in using LongRunning (the execution will switch back to the threadpool at the first await)

src/Datadog.Trace/Agent/AgentWriter.cs
```diff
@@ -122,13 +122,21 @@ public async Task FlushAndCloseAsync()
 
             var delay = Task.Delay(TimeSpan.FromSeconds(20));
 
-            var completedTask = await Task.WhenAny(_serializationTask, delay)
-                .ConfigureAwait(false);
+            // Internally WhenAny will make a defensive copy of the array
+            // so we reuse the outer one.
+            // https://source.dot.net/#System.Private.CoreLib/Task.cs,6028
+            // https://referencesource.microsoft.com/#mscorlib/system/threading/Tasks/Task.cs,6478
+            Task[] whenAnyArray = new Task[2];
```
kevingosse, line 129: Note that this is counter-productive on .NET 5, where they added a `Task.WhenAny` overload that takes two parameters: https://source.dot.net/#System.Private.CoreLib/Task.cs,6054

But we don't target .NET 5 yet, so good for now.

src/Datadog.Trace/Agent/AgentWriter.cs
```diff
@@ -138,8 +146,9 @@ await Task.WhenAny(_flushTask, Task.Delay(TimeSpan.FromSeconds(20)))
 
                 // In some situations, the flush thread can exit before flushing all the threads
                 // Force a flush for the leftover traces
-                completedTask = await Task.WhenAny(Task.Run(() => FlushBuffers(flushAllBuffers: true)), delay)
-                    .ConfigureAwait(false);
+                whenAnyArray[0] = Task.Factory.StartNew(state => ((AgentWriter)state).FlushBuffers(flushAllBuffers: true), this).Unwrap();
```
kevingosse, line 149: I would switch back to `Task.Run`. `FlushAndCloseAsync` is called only once in the app lifetime, and the closure is negligible versus all the allocations from starting a new task. This is a case where, I believe, the added complexity and verbosity outweighs the benefits

src/Datadog.Trace/Agent/AgentWriter.cs
```diff
@@ -194,13 +203,13 @@ private void RequestFlush()
 
         private async Task FlushBuffersTaskLoopAsync()
         {
+            Task[] whenAnyArray = new Task[3];
```
kevingosse, line 206: I don't understand the benefit of this one, since you don't reuse the array

src/Datadog.Trace/Agent/AgentWriter.cs
```diff
@@ -397,18 +406,26 @@ private void SerializeTracesLoop()
         private readonly struct WorkItem
         {
             public readonly Span[] Trace;
-            public readonly Action Callback;
+            public readonly Action<object> Callback;
+            public readonly object State;
```
kevingosse, line 410: You're making the struct bigger, with a parameter that is almost never used. You should make sure it's not slowing down the 99% case (= workitems with a trace but no callback)

src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
```diff
@@ -67,6 +71,7 @@ public static void AddTagsFromDbCommand(this Span span, IDbCommand command)
             }
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
kevingosse, line 74: Aren't those a bit big for inlining? 🤔

src/Datadog.Trace/FrameworkDescription.cs
```diff
@@ -14,19 +15,19 @@ internal partial class FrameworkDescription
...
+            new KeyValuePair<int, string>(528040, "4.8"),
+            new KeyValuePair<int, string>(461808, "4.7.2"),
+            new KeyValuePair<int, string>(461308, "4.7.1"),
+            new KeyValuePair<int, string>(460798, "4.7"),
+            new KeyValuePair<int, string>(394802, "4.6.2"),
+            new KeyValuePair<int, string>(394254, "4.6.1"),
+            new KeyValuePair<int, string>(393295, "4.6"),
+            new KeyValuePair<int, string>(379893, "4.5.2"),
+            new KeyValuePair<int, string>(378675, "4.5.1"),
+            new KeyValuePair<int, string>(378389, "4.5"),
```
kevingosse, line 30: C# 9, you can rewrite into `new(528040, "4.8")`, in this case I think it improves readability

src/Datadog.Trace/RuntimeMetrics/PerformanceCountersListener.cs
```diff
@@ -132,50 +133,100 @@ private void TryUpdateCounter(string path, PerformanceCounterWrapper counter, re
...
         {
-            var instanceNames = _memoryCategory.GetInstanceNames().Where(n => n.StartsWith(_processName)).ToArray();
-
-            if (instanceNames.Length == 1)
-            {
-                return instanceNames[0];
-            }
-
-            foreach (var name in instanceNames)
+            var allInstanceNames = _memoryCategory.GetInstanceNames();
```
kevingosse, line 171: I have mixed feelings about making this method so complex even though it's called once every 10 seconds, but your call

src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
```diff
@@ -63,7 +63,7 @@ internal RuntimeMetricsWriter(IDogStatsd statsd, TimeSpan delay, Func<IDogStatsd
 
             try
             {
-                _listener = initializeListener(statsd, delay);
+                _listener = initializeListener?.Invoke(statsd, delay) ?? InitializeListener(statsd, delay);
```
kevingosse, line 66: You're changing the behavior here. It means you will call `InitializeListener` if `initializeListener` returns a null value, which wasn't the case before

src/Datadog.Trace/ScopeManagerBase.cs
```diff
@@ -31,23 +31,22 @@ public Scope Activate(Span span, bool finishOnClose)
         {
             var newParent = Active;
             var scope = new Scope(newParent, span, this, finishOnClose);
-            var scopeOpenedArgs = new SpanEventArgs(span);
 
-            if (newParent == null)
+            if (newParent == null && TraceStarted != null)
             {
-                TraceStarted?.Invoke(this, scopeOpenedArgs);
+                TraceStarted(this, new SpanEventArgs(span));
```
kevingosse, line 37: That's a bad practice. You're introducing a subtle race condition if TraceStarted becomes null between line 35 and line 37

src/Datadog.Trace/ScopeManagerBase.cs
```diff
@@ -31,23 +31,22 @@ public Scope Activate(Span span, bool finishOnClose)
...
             }
 
-            SpanOpened?.Invoke(this, scopeOpenedArgs);
+            SpanOpened?.Invoke(this, new SpanEventArgs(span));
 
             Active = scope;
 
-            if (newParent != null)
+            if (newParent != null && SpanDeactivated != null)
```
kevingosse, line 44: Same as above

src/Datadog.Trace/TracingProcessManager.cs
```diff
@@ -114,141 +114,145 @@ private static Task StartProcessWithKeepAlive(ProcessMetadata metadata)
             var path = metadata.ProcessPath;
             Log.Debug("Starting keep alive for {Process}.", path);
 
-            return Task.Run(
-                async () =>
+            return Task.Factory.StartNew(state => InternalStartProcessWithKeepAlive(state), metadata).Unwrap();
```
kevingosse, line 117: As for the previous Task.Run, for something that is going to be called only once I'd rather pay the price of the closure rather than the added complexity of the Unwrap 😃

kevingosse (changes requested):

src/Datadog.Trace/Agent/AgentWriter.cs
andrewlock: See the conversation/change in https://github.com/DataDog/dd-trace-dotnet/pull/1306#discussion_r604472591 😄

src/Datadog.Trace/Agent/AgentWriter.cs
tonyredondo: well is inside a while(true) and every `Task.WhenAny()` will create 2 Arrays, the one from the `params` and the  internal defensive copy. The benefit is removing 1 allocation in each loop (that's why is declared outside the loop).

src/Datadog.Trace/Agent/AgentWriter.cs
tonyredondo: 🤔 makes sense I'm passing the struct as value so the bigger, the slower. This is probably a case of CPU vs Memory, and we prefer performance... I'll revert this change.

src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
tonyredondo: Yeah, the attribute will not do anything here. Will remove it.

src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
tonyredondo: oh yeah, that's right, reverting that.

src/Datadog.Trace/ScopeManagerBase.cs
tonyredondo: Makes sense, I really don't remember why I changed, I think I was testing because the original only adds a `dup` opcode to use the same value... reverting it...

src/Datadog.Trace/TracingProcessManager.cs
tonyredondo: I don't have a strong opinion on this, you're right is only called once. In the other side `Unwrap` is pretty simple nowdays (https://source.dot.net/#System.Private.CoreLib/TaskExtensions.cs,e4f7cc9af0c22301), Is not automatic as Task.Run (https://source.dot.net/#System.Private.CoreLib/Task.cs,5291).

src/Datadog.Trace/Agent/AgentWriter.cs
tonyredondo: Yeah, we already had this conversation :D I'll add the `#if`.

src/Datadog.Trace/Agent/AgentWriter.cs
tonyredondo: I hate DisplayClasses 🤣, will revert for readability.

src/Datadog.Trace/Headers/NameValueHeadersCollection.cs
lucaspimentel: I think we should prioritize stability over performance / resource usage, at least until we resolve the version mismatch issues.

src/Datadog.Trace/TracingProcessManager.cs
lucaspimentel: Should `InternalStartProcessWithKeepAlive()` be `async` at all? All the `await` are on `Task.Delay()` only.

src/Datadog.Trace/Headers/NameValueHeadersCollection.cs
tonyredondo: then maybe we should push back these improvements until we resolve the compatibilities issues we have. I'll resolve and handle all the requested changes but leave the PR with the `no-merge` flag.

lucaspimentel: thanks!

src/Datadog.Trace/Agent/AgentWriter.cs
tonyredondo: done in 40744c8

src/Datadog.Trace/Agent/AgentWriter.cs
tonyredondo: done in 40744c8

src/Datadog.Trace/Agent/AgentWriter.cs
tonyredondo: done in 40744c8

src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
tonyredondo: done in 40744c8

src/Datadog.Trace/FrameworkDescription.cs
tonyredondo: done in 40744c8

src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
tonyredondo: done in b3feead

src/Datadog.Trace/ScopeManagerBase.cs
tonyredondo: done in b3feead

src/Datadog.Trace/ScopeManagerBase.cs
tonyredondo: done in 40744c8

src/Datadog.Trace/TracingProcessManager.cs
tonyredondo: reverted on 2027883

src/Datadog.Trace/TracingProcessManager.cs
tonyredondo: I reverted the change in 2027883

src/Datadog.Trace/Agent/AgentWriter.cs
kevingosse: Right, I forgot about the loop

src/Datadog.Trace/Agent/AgentWriter.cs
andrewlock: Careful, you switched to WhenAll at one point there 😉

kevingosse: Right, sorry 😄  Fixed.

tonyredondo:
Most of the listed improvements were already implemented for v2.0, closing this PR.

PR #1325, opened 2021-03-31
Add analyzer project + ThreadAbortAnlayzer to detect inifinte loops on ThreadAbortException

andrewlock:
Added an analyzer project + unit test project (based on the default analyzer template in VS) for detecting the following construct:

## Why do we need this analyzer?

Due to [a bug in the .NET Framework runtime](https://github.com/dotnet/runtime/issues/9633), tight try-catch blocks inside  while loops are vulnerable to ending up in an infinite loop when a `ThreadAbortException` occurs. Normally, a `ThreadAbortException` is rethrown automatically from a catch block, but that does not happen on .NET 4.6.1 (for example) when built in Release mode for `x64`. 

## What does the analyzer do? 

The analyzer detects the following construct in code:

```csharp
while(...)
{
    try {
    }
    catch(ex) {
    }
}
```

and flags the catch block as an error. The codefix encourages you to add a `throw;` statement to make it "safe" e.g.:

```csharp
while(...)
{
    try {
    }
    catch(ex) {
       throw;
    }
}
```

In reality, you will probably want to add an extra catch block for ThreadAbortException [as in this example](https://github.com/DataDog/dogstatsd-csharp-client/pull/137/files#diff-9690ede94b874fd062de66c80f4dba8fcef1bb17bb55027aa04de33aa8f79842R115-R123), the important point is we detect the construct itself.

Running the analyzer on our current project shows there are no cases of this pattern in our current code base (given the limitations described below)

## Analyzer features

* Detects a try-catch block tightly nested inside a `while` loop
* Finds the first catch block that could catch the `ThreadAbortException`
* Checks that the exception is rethrown.
* Flags the pattern as an error, offers a codefix to add the `throw;` statement:

Flagging the pattern in Visual Studio:
analyzer_04

Offering a CodeFix:
analyzer_05

The fixed code is now "safe"
analyzer_06

Works in Rider/VS Code  etc too, no need for VSIX or anything:
analyzer_07

## Limitations

The analyzer is not particularly sophisticated, and there are various limitations

* **Not supported on `net45`**. The analyzers target .NET Standard2.0, so we can't used them in `net45` projects. This isn't a massive problem, as the issue doesn't manifest in `net45`, and the `net461` target we have _will_ flag them. 
* **Assumes infinite while loop**. For simplicity, the analyzer doesn't check the expression in the while loop, and assumes it will loop infinitely. That's a conservative approach
* **Exception filters are not considered**. For simplicity, I've ignored exception filters. That means we may assume an exception is caught when it is not, though this would _generally_ be more conservative. Unlikely to be an issue in practice I think
* **Analyzers do not run in Autogenerated code**. This unfortunately means we _wouldn't_ have caught the original bug that spawned this work, a bug in [the Vendored StatsD library](https://github.com/DataDog/dd-trace-dotnet/blob/cee651f3a473d353b795779d1539fb696afeb019/src/Datadog.Trace/Vendors/StatsdClient/Worker/AsynchronousWorker.cs#L140-L148) 🙁 I can't see an easy workaround for that yet

@DataDog/apm-dotnet


Datadog.Trace.sln
```diff
@@ -433,6 +433,10 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "DuplicateTypeProxy", "test\
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Samples.AspNetCoreRazorPages", "test\test-applications\integrations\Samples.AspNetCoreRazorPages\Samples.AspNetCoreRazorPages.csproj", "{1B9E6BF4-9D48-4988-9945-248096119E46}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Datadog.Trace.Tools.Analyzers", "src\Datadog.Trace.Tools.Analyzers\Datadog.Trace.Tools.Analyzers.csproj", "{5450EA0B-56D3-4E29-932E-094AD037B345}"
```
lucaspimentel, line 436: This project should probably live in the top-level `tools` folder.

NuGet.config
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <!--<clear />-->
+    <add key="nuget" value="https://api.nuget.org/v3/index.json" />
+    <add key="dotnet-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" />
```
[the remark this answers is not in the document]
andrewlock: This is for the preview testing packages. They're included in the default VS analyzer template - it's kinda crazy they're not on nuget though yeah. Frankly I currently have 0 idea how to get the CI to pass, but once I do, yeah, I'll look into this 😉 

```xml
<PackageReference Include="Microsoft.CodeAnalysis.CSharp.Analyzer.Testing.MSTest" Version="1.0.1-beta1.*" />
    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.CodeFix.Testing.MSTest" Version="1.0.1-beta1.*" />
    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.CodeRefactoring.Testing.MSTest" Version="1.0.1-beta1.*" />
    <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic.Analyzer.Testing.MSTest" Version="1.0.1-beta1.*" />
    <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic.CodeFix.Testing.MSTest" Version="1.0.1-beta1.*" />
    <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic.CodeRefactoring.Testing.MSTest" Version="1.0.1-beta1.*" />
```

Datadog.Trace.sln
andrewlock: I specifically didn't put it in there because it needs to be built with all the _src_ projects, which means fewer updates to the CI. But as I can't get the `x86` build to pass-as-is this is probably going to be shelved until I've rehashed the CI anyway, so I'll look into it then

NuGet.config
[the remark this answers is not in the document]
andrewlock: FWIW, that feed is relied on by the dotnet sdk too. I guess it's their "dev" feed: https://github.com/dotnet/sdk/blob/main/NuGet.config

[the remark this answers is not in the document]
andrewlock: It took me sufficient time to get this working in CI, that they've since released the packages to nuget.org, so this is no longer necessary. Removed in https://github.com/DataDog/dd-trace-dotnet/pull/1325/commits/a44bf18da36dc17bed5646816c135c49b518f9f5

test/Datadog.Trace.Tools.Analyzers.Tests/ThreadAbortAnalyzerTests.cs
```diff
@@ -0,0 +1,203 @@
+﻿using System.Threading.Tasks;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using VerifyCS = Datadog.Trace.Tools.Analyzers.Tests.CSharpCodeFixVerifier<
+    Datadog.Trace.Tools.Analyzers.ThreadAbortAnalyzer.ThreadAbortAnalyzer,
+    Datadog.Trace.Tools.Analyzers.ThreadAbortAnalyzer.ThreadAbortCodeFixProvider>;
+
+namespace Datadog.Trace.Tools.Analyzers.Tests
+{
+    [TestClass]
```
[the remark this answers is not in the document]
andrewlock: They've since released XUnit packages for this, so switched to those instead in https://github.com/DataDog/dd-trace-dotnet/pull/1325/commits/a44bf18da36dc17bed5646816c135c49b518f9f5

src/Datadog.Trace.Tools.Analyzers/ThreadAbortAnalyzer/ThreadAbortSyntaxHelper.cs
```diff
@@ -0,0 +1,82 @@
...
+                {
+                    willCatchThreadAbort = true;
+
+                    // We're in the catch block that will catch the ThreadAbort
+                    // Make sure that we re-throw the exception
+                    // This is a very basic check, in that it doesn't check control flow etc
+                    // It requires that you have a throw; in the catch block
+
+                    // We are only checking the direct ancestors (nesting breaks this analysis)
+                    // and if you have an expression, it must be the exception declared in the
```
lucaspimentel, line 57: "it must be the exception declared in the..." ?

lucaspimentel (approved):

test/Datadog.Trace.Tools.Analyzers.Tests/ThreadAbortAnalyzerTests.cs
```diff
@@ -0,0 +1,208 @@
...
+    public class ThreadAbortAnalyzerTests
+    {
+        private const string DiagnosticId = ThreadAbortAnalyzer.ThreadAbortAnalyzer.DiagnosticId;
+
+        // No diagnostics expected to show up
+        [TestMethod]
```
zacharycmontoya, line 20: Would you mind updating these attributes to `[Fact]`, `[Theory]`, etc.?

test/Datadog.Trace.Tools.Analyzers.Tests/ThreadAbortAnalyzerTests.cs
```diff
@@ -0,0 +1,208 @@
+﻿// <copyright file="ThreadAbortAnalyzerTests.cs" company="Datadog">
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2 License.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
+// </copyright>
+
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
```
zacharycmontoya, line 7: Are these test cases being invoked by our dotnet test commands?

zacharycmontoya (approved):
LGTM. I left some small Xunit comments

src/Datadog.Trace.Tools.Analyzers/ThreadAbortAnalyzer/ThreadAbortSyntaxHelper.cs
andrewlock: 🤷‍♂️😆 This was all boilerplate code in the analyzer template. Looks like that's all been incorporated into the NuGet packages now, so I can delete all those extra files 🎉

test/Datadog.Trace.Tools.Analyzers.Tests/ThreadAbortAnalyzerTests.cs
andrewlock: Yes, they're run as part of the various UnitTest stages. The fact they were (accidentally) MSTest vs XUnit doesn't affect that, but I've updated to XUnit now anyway 🙂

test/Datadog.Trace.Tools.Analyzers.Tests/ThreadAbortAnalyzerTests.cs
andrewlock: Oops, my bad! Fixed in https://github.com/DataDog/dd-trace-dotnet/pull/1325/commits/c6bade60b87a4c4cc4066238b4adaf5d07eb5875

PR #1326, opened 2021-03-31
Replace Task.Delay with Thread.Sleep in dogstatsd

kevingosse:
Replace Task.Delay with Thread.Sleep in dogstatsd.

Also going to push the same change upstream.

The worker is running on a dedicated thread, so it can't await.


andrewlock (approved):

lucaspimentel (approved):

lucaspimentel:
for future reference: https://github.com/DataDog/dogstatsd-csharp-client/pull/159

PR #1327, opened 2021-03-31
Update Uri cleaning algorithm to be more agressive

andrewlock:
Update the algorithm used for URL cleaning (based on [this branch]( https://github.com/DataDog/dd-trace-dotnet/compare/lpimentel/url-quantization) by @lucaspimentel) from needing to be strictly a long/guid to:

* Replaces any hex-ish (number/a-f/`-`) string with `?` (instead of needing to be )
* If it's numbers and dashes only, always replaces.
* If it contains alpha, must be at least 16 chars long and contain a number

Added in a back-compat way, that avoids changing internal (public) APIs.

Conditionally enabled the new format in ASP.NET integration when the new route names flag is enabled.

Note that we're currently using the URI cleaning method to clean _outgoing_ request URLs too. We probably want to update that to use the newer algorithm, but I feel like that might need to be behind a feature flag too? Should we use the same one? Seeing as _some_ resource names are changing, seems like that might be reasonable?

> Note: Once/if this is approved, I'll merge and then rebase https://github.com/DataDog/dd-trace-dotnet/pull/1289 on top, and enable the new algo in that branch too

@DataDog/apm-dotnet


src/Datadog.Trace/Util/UriHelpers.cs
```diff
@@ -135,6 +172,84 @@ public static string GetCleanUriPath(string absolutePath)
...
+                return false;
+            }
+
+            var containsNumber = false;
+            foreach (var c in segment)
+            {
+                switch (c)
+                {
+                    case >= '0' and <= '9':
+                        containsNumber = true;
```
lucaspimentel, line 189: There's a possible edge-case regression here: if something would have parsed as `Guid` before but had no numbers, we are no longer replacing it.
```
aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa
bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
```
Setting `containsNumber = true` on hyphens (but only if length is long enough?) takes care of the first case, but what about the second one?

My quick take is that is a string only has A-F and is longer than 16 chars, we should assume it's a hex string and replace it. On the other hand, my math (`6^32 / 16^32`) says only a very small fraction (2.3e-14) of all possible GUIDs contain letters only (ignoring other characters like hyphens). Maybe I'm overthinking this.

src/Datadog.Trace/Util/UriHelpers.cs
```diff
@@ -135,6 +172,84 @@ public static string GetCleanUriPath(string absolutePath)
 #endif
         }
 
+#if NETCOREAPP
```
lucaspimentel, line 175: I'd like to see some benchmark numbers show that the extra complexity of having two copies of the same method are worth it.

lucaspimentel (commented):
Thanks for taking over, @andrewlock! LGTM me overall. I would like to see some benchmarks to make sure we're not regressing performance in general and also to justify having two separate `IsIdentifierSegment()` implementations.

src/Datadog.Trace/Util/UriHelpers.cs
```diff
@@ -7,7 +7,12 @@ internal static class UriHelpers
     {
         public static string CleanUri(Uri uri, bool removeScheme, bool tryRemoveIds)
         {
-            var path = tryRemoveIds ? GetCleanUriPath(uri.AbsolutePath) : uri.AbsolutePath;
+            return CleanUri(uri, removeScheme, tryRemoveIds, useLegacyIdCleaning: true);
```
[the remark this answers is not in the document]
andrewlock: I'm in two minds - on the one hand we don't want those to be obsolete, we just want them to have the opposite behaviour 😅 But by adding the new APIs, we can never remove either 🤔  Not really a good way round that I guess, unless we change the behaviour behind the scenes, which is obviously not good either...

src/Datadog.Trace/Util/UriHelpers.cs
andrewlock: Yeah, I considered that one, but as you say it's a very small fraction of all the GUIDs, but more importantly, _those aren't valid GUIDs_ 😉 Version 1, 2, 3, and 4 GUIDs all include the version as the first char in the middle group, i.e. a valid version 1 GUID would be `aaaaaaaa-aaaa-1aaa-aaaa-aaaaaaaaaaaa`, which would obviously be fine 🙂 . 

Another thing I considered was SHA hashes though, which seem like another good candidate. No those _could_ legitimately be all chars, but then those are looking at the v. small numbers you're talking about here too... hyhens obviously don't help with the hash thing though...

Not sure, I'm in two minds 🙂

lucaspimentel: Yes, it's game of chance, I guess. How long must a string of `[A-F]` be until we're confident that it's meant to be hexadecimal? The current implementation is probably fine.

src/Datadog.Trace/Util/UriHelpers.cs
andrewlock: 😆 yep, good call, turns out totally not worth it! 

```bash
BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19042
Intel Core i7-8750H CPU 2.20GHz (Coffee Lake), 1 CPU, 12 logical and 6 physical cores
.NET Core SDK=5.0.201
  [Host]     : .NET Core 3.1.13 (CoreCLR 4.700.21.11102, CoreFX 4.700.21.11602), X64 RyuJIT
  DefaultJob : .NET Core 3.1.13 (CoreCLR 4.700.21.11102, CoreFX 4.700.21.11602), X64 RyuJIT


|                             Method |     Mean |   Error |  StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------------------------- |---------:|--------:|--------:|-------:|------:|------:|----------:|
|     StringBased_LongUrlWithChanges | 320.7 ns | 4.63 ns | 4.10 ns | 0.0339 |     - |     - |     160 B |
|       SpanBased_LongUrlWithChanges | 268.1 ns | 3.70 ns | 3.63 ns | 0.0339 |     - |     - |     160 B |
|  StringBased_LongUrlWithoutChanges | 241.2 ns | 1.18 ns | 1.04 ns | 0.0353 |     - |     - |     168 B |
|    SpanBased_LongUrlWithoutChanges | 207.5 ns | 1.83 ns | 1.62 ns | 0.0355 |     - |     - |     168 B |
|    StringBased_ShortUrlWithChanges | 137.8 ns | 1.90 ns | 1.68 ns | 0.0153 |     - |     - |      72 B |
|      SpanBased_ShortUrlWithChanges | 114.7 ns | 2.19 ns | 1.83 ns | 0.0153 |     - |     - |      72 B |
| StringBased_ShortUrlWithoutChanges | 134.7 ns | 2.40 ns | 2.00 ns | 0.0169 |     - |     - |      80 B |
|   SpanBased_ShortUrlWithoutChanges | 118.9 ns | 0.83 ns | 0.65 ns | 0.0169 |     - |     - |      80 B |
```

I'll remove the span-based implementation in favour of the string-based one

andrewlock: @lucaspimentel I simplified the implementation significantly in https://github.com/DataDog/dd-trace-dotnet/pull/1327/commits/1a1704d2b2ee4534465b65354525d6470130913a. As you suspected, there was no need for two copies of the same method, and in fact we could significantly simplify the "main" method too. Added some benchmarks to the description

src/Datadog.Trace/Util/UriHelpers.cs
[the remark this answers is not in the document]
lucaspimentel: I was thinking we change all our code to use the new overload with `useLegacyIdCleaning` and leave the old overload only for backwards compat (e.g. an older `Datadog.Trace.ClrProfiler.Managed` calling into newer `Datadog.Trace`). But now that I think about it, we renamed these methods from something like `GetRelativeUri()` so the compat ship has sailed.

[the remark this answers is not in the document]
andrewlock: We reverted that naming change I believe? 🤔 https://github.com/DataDog/dd-trace-dotnet/pull/1288/commits/3be6c517d37674f94935478c149605aadcba8427

I guess what we _could_ do, is rework the new methods to assume the new behaviour, and conditionally call the old methods for the "obsolete"/legacy behaviour? Getting a bit confused tbh 😵 Have to be aware that these current methods are in the prerelease build, so if we're going to change them, should do it asap!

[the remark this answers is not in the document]
lucaspimentel: >on the one hand we don't want those to be obsolete

Why not? I think we should not call the old overloads anymore. We should always pass the `useLegacyIdCleaning` argument.

src/Datadog.Trace/Util/UriHelpers.cs
lucaspimentel: Awesome, thanks!

src/Datadog.Trace/Util/UriHelpers.cs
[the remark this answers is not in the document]
andrewlock: > Why not?

Because the `useLegacyIdCleaning` arg is ugly, I'd rather we could not expose it in the public API at all 🙂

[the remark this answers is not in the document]
andrewlock: I think I was overthinking this. I've updated the methods to be obsolete in https://github.com/DataDog/dd-trace-dotnet/pull/1327/commits/4dc20e81fb3edffa5b8aafc51ede86d295b24dc8, and also explicitly documented the "main" `CleanUri` method.

Also updated the aspnetcore integration to use the new cleaning method when the feature flag is enabled in https://github.com/DataDog/dd-trace-dotnet/pull/1327/commits/b9e59a3f6f8dbbba6f4f15f3f33477ea2ebbf6b7

zacharycmontoya (approved):

lucaspimentel (approved):

PR #1329, opened 2021-04-01
Fix references to dotnet-tracer-native.log

andrewlock:
In #1296 we added rotation for the native log file, which means the "bare" dotnet-tracer-native.log file isn't used.

This PR updates the build scripts to account for that, so that we still publish the log results for viewing in the ci pipelines

@DataDog/apm-dotnet

### DIFF
--- a/build/docker/Datadog.Trace.ClrProfiler.IntegrationTests.arm64.sh
+++ b/build/docker/Datadog.Trace.ClrProfiler.IntegrationTests.arm64.sh
@@ -18,4 +18,5 @@ wait-for-it mongo:27017 -- \
 wait-for-it postgres:5432 -- \
 dotnet vstest test/Datadog.Trace.ClrProfiler.IntegrationTests/bin/$buildConfiguration/$publishTargetFramework/publish/Datadog.Trace.ClrProfiler.IntegrationTests.dll --logger:trx --ResultsDirectory:test/Datadog.Trace.ClrProfiler.IntegrationTests/results --TestCaseFilter:Category!=ArmUnsupported
 
-cp /var/log/datadog/dotnet/dotnet-tracer-native*.log /project/
+mkdir /project/logs
+cp /var/log/datadog/dotnet/dotnet-tracer-native*.log /project/logs
--- a/build/docker/Datadog.Trace.ClrProfiler.IntegrationTests.arm64.sh
+++ b/build/docker/Datadog.Trace.ClrProfiler.IntegrationTests.arm64.sh
@@ -4,7 +4,6 @@ set -euxo pipefail
 cd "$( dirname "${BASH_SOURCE[0]}" )"/../../
 
 mkdir -p /var/log/datadog/dotnet
-touch /var/log/datadog/dotnet/dotnet-tracer-native.log
 
 dotnet vstest test/Datadog.Trace.IntegrationTests/bin/$buildConfiguration/$publishTargetFramework/publish/Datadog.Trace.IntegrationTests.dll --logger:trx --ResultsDirectory:test/Datadog.Trace.IntegrationTests/results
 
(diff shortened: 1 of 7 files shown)


build/docker/Datadog.Trace.ClrProfiler.IntegrationTests.arm64.sh
```diff
@@ -1,14 +1,20 @@
 #!/bin/bash
-set -euxo pipefail
+set -uxo pipefail
```
pellared, line 2: why removing `-e`? (it looks especially strange when `-o pipefail` is still here while there is no `|`)
I guess instead of removing `-e` you can simply add `|| true` to a place where you want to ignore the error

the same comment applies to other places

build/docker/Datadog.Trace.ClrProfiler.IntegrationTests.arm64.sh
```diff
@@ -17,5 +23,14 @@ wait-for-it sqledge:1433 -- \
...
+# Collect run data
+mkdir /project/build_data
+
+cp /var/log/datadog/dotnet/* /project/build_data/
+cp /tmp/coredump* /project/build_data/ 2>/dev/null || :
 
-cp /var/log/datadog/dotnet/dotnet-tracer-native.log /project/
+if [ $st01 -eq 1 ] || [ $st02 -eq 1 ] || [ $st03 -eq 1 ]; then
+    exit 1
+fi
```
pellared, line 36: nit: missing EOL

(applies to other places)

pellared (commented):
👍

build/docker/Datadog.Trace.ClrProfiler.IntegrationTests.arm64.sh
andrewlock: We want to grab the logs even if the "main" step fails, but then fail the job _afterwards_. I don't think we can do that with the `|| true` approach 🤔

pellared: try using `trap`

e.g.
```
cleanup() {
    # ...
}

trap cleanup EXIT
```

andrewlock: That has slightly different behaviour to the changes in this though - that will still exit on the first error, as opposed to running all three commands and _then_ exiting. I assume we want the latter approach right @tonyredondo ?

tonyredondo: Well yeah the intention on this one is to run all 3 dotnet commands no matter what and collect all logs and crashdumps that is generated.

that’s why we were avoiding the early exit.

pellared: I think it would be better to unset the `-e` before executing the tests: https://superuser.com/a/940504
And then do something like this to check if any of the test failed:  https://stackoverflow.com/a/1000395/859275

andrewlock: For simplicity I switched to reenabling `-e` and using `trap` instead. So we'll fail on first error, but I don't think that's a big deal. We copy the logs and the crashdumps in any case

tonyredondo (approved):
👍🏻

pellared (approved):

samples/ServiceFabricRemoting/DatadogInstall.ps1
```diff
@@ -37,7 +37,7 @@ function Set-MachineEnvironmentVariable {
 
 Set-MachineEnvironmentVariable 'DD_DOTNET_TRACER_HOME' $DD_DOTNET_TRACER_HOME
 Set-MachineEnvironmentVariable 'DD_INTEGRATIONS' "$DD_DOTNET_TRACER_HOME\integrations.json"
-Set-MachineEnvironmentVariable 'DD_TRACE_LOG_PATH' "$LOGS_PATH\dotnet-tracer-native.log"
+Set-MachineEnvironmentVariable 'DD_TRACE_LOG_DIRECTORY ' "$LOGS_PATH"
```
zacharycmontoya, line 40:
```suggestion
Set-MachineEnvironmentVariable 'DD_TRACE_LOG_DIRECTORY' "$LOGS_PATH"
```

zacharycmontoya: Not sure if the trailing whitespace would cause errors but...better to be safe than sorry 😄

zacharycmontoya (approved):

PR #1331, opened 2021-04-01
Increase diagnostic observer test timeout

andrewlock:
Try and reduce test flakiness. Would be nice not to have this time based approach but this is the same approach `MockTracerAgent` uses so...

@DataDog/apm-dotnet

### DIFF
--- a/test/Datadog.Trace.IntegrationTests/DiagnosticListeners/AspNetCoreDiagnosticObserverTests.cs
+++ b/test/Datadog.Trace.IntegrationTests/DiagnosticListeners/AspNetCoreDiagnosticObserverTests.cs
@@ -74,17 +74,19 @@ namespace Datadog.Trace.IntegrationTests.DiagnosticListeners
 
                 // The diagnostic observer runs on a separate thread
                 // This gives time for the Stop event to run and to be flushed to the writer
-                var iterations = 10;
-                while (iterations > 0)
+                const int timeoutInMilliseconds = 10000;
+
+                var deadline = DateTime.Now.AddMilliseconds(timeoutInMilliseconds);
+                while (DateTime.Now < deadline)
                 {
                     if (writer.Traces.Count > 0)
                     {
                         break;
                     }
 
-                    Thread.Sleep(10);
-                    iterations--;
+                    Thread.Sleep(200);
                 }
+                
             }
 
             var trace = Assert.Single(writer.Traces);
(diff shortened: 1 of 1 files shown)


lucaspimentel:
StyleCop!
https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=55601&view=logs&j=4fbced83-508e-5fe0-c978-5c71ec0fc506&t=baa549dc-8c15-51ff-9616-5cf4cfc294db&l=122

andrewlock:
Probably shouldn't have YOLOed that one without compiling it 😳

kevingosse:
Is there a way to rethink the test without a delay? From experience, increasing the timeout may reduce flakiness, but that won't fix it.

andrewlock:
> Is there a way to rethink the test without a delay? From experience, increasing the timeout may reduce flakiness, but that won't fix it.

Not that I'm aware of🤷‍♂️ Every integration test that uses the `MockTracerAgent` (i.e. most of them) uses this approach, but I'm open to suggestions!

zacharycmontoya:
The diagnostic listener events that we subscribe to should happen synchronously with the request, right? For some reason I had it in my head that the callbacks are interleaved within the execution of the request. If this is the case, when we receive a response from the http client, we can just force the trace writer to flush and then we can immediately start asserting information about the trace, and we wouldn't need to do any waiting.

andrewlock:
> The diagnostic listener events that we subscribe to should happen synchronously with the request, right?

Yes, that's correct, and we test that functionality in "unit" tests

> If this is the case, when we receive a response from the http client, we can just force the trace writer to flush and then we can immediately start asserting information about the trace, and we wouldn't need to do any waiting.

I don't entirely follow on this statement - I can certainly call `Tracer.FlushTracesAsync()` after the http call, which should improve things. But we still need this "waiting" code to wait for the traces to appear at the agent side?

There's nothing particularly special about these diagnostic observer tests compared to the other integration tests we use elsewhere - in all cases we (currently) have a timeout waiting for the traces to appear at a mock agent. The only difference with these tests is I (for no apparent reason) used a much shorter timeout, so this increases it to be of the same magnitude as the other mock tracer.

kevingosse:
> IMO, we should merge this for now, as currently this test is very flaky. Even if it doesn't outright fix it, the test will be less flaky after this!

I agree it can't hurt. :shipit:

kevingosse (approved):

colin-higgins (approved):

zacharycmontoya (approved):

PR #1332, opened 2021-04-01
ProcessExit event handler improvements

tonyredondo:
This PR improves the way we handle the ProcessExit event by ensuring that the tracer handler is the last one to be executed from the event.

#### The background:
The process signals received on the CLR are handled this way:

1. `SIGTERM` is handled by `AppDomain.CurrentDomain.ProcessExit`
2. `SIGINT` and `SIGQUIT` is handled by `Console.CancelKeyPress`
3. `SIGKILL` can't be caught or ignored.

https://github.com/DataDog/dd-trace-dotnet/pull/1323#issuecomment-810232442

In scenarios where `IHostApplicationLifetime` is being used we can face the case where the tracer handles the `ProcessExit` event where the traces are flushed and the writer is closed but the actual application can still be creating new spans. As @andrewlock describes in the discussion here: https://github.com/DataDog/dd-trace-dotnet/pull/1323#issuecomment-810379331

This PR solves the issue by ensuring the tracer handler for ProcessExit will be the last one of the event, to send all the spans generated on previous handlers.

Also on `SIGINT` and `SIGQUIT` signals changes the actual behavior to only flush and not close due we can receive a `SIGTERM` later.

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace/Tracer.cs
+++ b/src/Datadog.Trace/Tracer.cs
@@ -671,7 +671,39 @@ namespace Datadog.Trace
 
         private void CurrentDomain_ProcessExit(object sender, EventArgs e)
         {
-            RunShutdownTasks();
+            // This handles a grateful shutdown or a SIGTERM
+            // Multiple delegates can be registered to this event. (eg. IHostApplicationLifetime instances)
+            // So we must be sure we are the last handler of the event to ensure we are flushing all the
+            // possible traces we are generating.
+
+            // For this, we try to get the internal ProcessExit delegate from the event.
+            var processExitDelegate = DelegatesHelper.GetInternalProcessExitDelegate();
+            if (processExitDelegate != null)
+            {
+                // With the internal MulticastDelegate we first create the delegate instance
+                // we want to run at last of the event handler.
+
+                // Sets the final delegate to be executed by the ProcessExit event.
+                var lastDelegate = new EventHandler((s, earg) =>
+                {
+                    RunShutdownTasks();
+                });
+
+                // Try to modify the MulticastDelegate invocation list and set the last delegate.
+                if (!DelegatesHelper.TrySetLastDelegate(processExitDelegate, lastDelegate))
+                {
+                    // This means we are in the last delegate already
+                    // or we were unable to set the last delegate, in
+                    // any case we fallback by running the delegate now.
+                    lastDelegate(sender, e);
+                }
+            }
+            else
+            {
+                // If we were unable to extract the internal delegate of the event then
+                // we just fallback and run the code here.
+                RunShutdownTasks();
+            }
         }
 
         private void CurrentDomain_UnhandledException(object sender, UnhandledExceptionEventArgs e)
@@ -682,7 +714,17 @@ namespace Datadog.Trace
 
         private void Console_CancelKeyPress(object sender, ConsoleCancelEventArgs e)
         {
-            RunShutdownTasks();
+            // Console Cancel KeyPress is raised on SIGINT and SIGQUIT signal,
+            // here we ensure we flush and not close the writer.
+            // There's a possibility of a graceful shutdown handled by ProcessExit.
+            try
+            {
+                _agentWriter.FlushTracesAsync().GetAwaiter().GetResult();
+            }
+            catch (Exception ex)
+            {
+                Log.Error(ex, "Error flushing traces on shutdown.");
+            }
         }
 
         private void CurrentDomain_DomainUnload(object sender, EventArgs e)
--- a/src/Datadog.Trace/Tracer.cs
+++ b/src/Datadog.Trace/Tracer.cs
@@ -695,7 +695,7 @@ namespace Datadog.Trace
                     // This means we are in the last delegate already
                     // or we were unable to set the last delegate, in
                     // any case we fallback by running the delegate now.
-                    lastDelegate(sender, e);
+                    RunShutdownTasks();
                 }
             }
             else
--- a/src/Datadog.Trace/Tracer.cs
+++ b/src/Datadog.Trace/Tracer.cs
@@ -671,7 +671,7 @@ namespace Datadog.Trace
 
         private void CurrentDomain_ProcessExit(object sender, EventArgs e)
         {
-            // This handles a grateful shutdown or a SIGTERM
+            // This handles a graceful shutdown or a SIGTERM
             // Multiple delegates can be registered to this event. (eg. IHostApplicationLifetime instances)
             // So we must be sure we are the last handler of the event to ensure we are flushing all the
             // possible traces we are generating.
--- a/src/Datadog.Trace/Tracer.cs
+++ b/src/Datadog.Trace/Tracer.cs
@@ -714,16 +714,24 @@ namespace Datadog.Trace
 
         private void Console_CancelKeyPress(object sender, ConsoleCancelEventArgs e)
         {
-            // Console Cancel KeyPress is raised on SIGINT and SIGQUIT signal,
-            // here we ensure we flush and not close the writer.
+            // Console Cancel KeyPress is raised when CTRL+C or CTRL+BREAK are pressed in a console windows
+            // On Unix: is registered using SIGINT and SIGQUIT signals.
+            // On Windows: is registered with Kernel32.SetConsoleCtrlHandler.
+            // Here we ensure we flush and not close the writer.
             // There's a possibility of a graceful shutdown handled by ProcessExit.
+            SynchronizationContext context = SynchronizationContext.Current;
             try
             {
+                SynchronizationContext.SetSynchronizationContext(null);
                 _agentWriter.FlushTracesAsync().GetAwaiter().GetResult();
             }
             catch (Exception ex)
             {
                 Log.Error(ex, "Error flushing traces on shutdown.");
+            }
+            finally
+            {
+                SynchronizationContext.SetSynchronizationContext(context);
             }
         }
 
@@ -734,13 +742,19 @@ namespace Datadog.Trace
 
         private void RunShutdownTasks()
         {
+            SynchronizationContext context = SynchronizationContext.Current;
             try
             {
-                _agentWriter.FlushAndCloseAsync().Wait();
+                SynchronizationContext.SetSynchronizationContext(null);
+                _agentWriter.FlushAndCloseAsync().GetAwaiter().GetResult();
             }
             catch (Exception ex)
             {
                 Log.Error(ex, "Error flushing traces on shutdown.");
+            }
+            finally
+            {
+                SynchronizationContext.SetSynchronizationContext(context);
             }
         }
 
--- a/src/Datadog.Trace/Tracer.cs
+++ b/src/Datadog.Trace/Tracer.cs
@@ -717,21 +717,65 @@ namespace Datadog.Trace
             // Console Cancel KeyPress is raised when CTRL+C or CTRL+BREAK are pressed in a console windows
             // On Unix: is registered using SIGINT and SIGQUIT signals.
             // On Windows: is registered with Kernel32.SetConsoleCtrlHandler.
-            // Here we ensure we flush and not close the writer.
-            // There's a possibility of a graceful shutdown handled by ProcessExit.
-            SynchronizationContext context = SynchronizationContext.Current;
-            try
+            // ...
+            // If ConsoleCancelEventArgs.Cancel is false then the Process will be exit without calling ProcessExit event. So we shuld flush and close.
+            // If ConsoleCancelEventArgs.Cancel is true the process will not exit so we will only flush.
+            // Because the Cancel property can be modified by the application handlers we should listen at the end of the queue.
+
+            // For this, we try to get the internal CancelKeyPress delegate from the event.
+            var cancelKeyPressDelegate = DelegatesHelper.GetInternalCancelKeyPressDelegate();
+            if (cancelKeyPressDelegate != null)
             {
-                SynchronizationContext.SetSynchronizationContext(null);
-                _agentWriter.FlushTracesAsync().GetAwaiter().GetResult();
+                // With the internal MulticastDelegate we first create the delegate instance
+                // we want to run at last of the event handler.
+
+                // Sets the final delegate to be executed by the CancelKeyPress event.
+                var lastDelegate = new ConsoleCancelEventHandler(HandleCancelKeyPress);
+
+                // Try to modify the MulticastDelegate invocation list and set the last delegate.
+                if (!DelegatesHelper.TrySetLastDelegate(cancelKeyPressDelegate, lastDelegate))
+                {
+                    // This means we are in the last delegate already
+                    // or we were unable to set the last delegate, in
+                    // any case we fallback by running the delegate now.
+                    HandleCancelKeyPress(sender, e);
+                }
             }
-            catch (Exception ex)
+            else
             {
-                Log.Error(ex, "Error flushing traces on shutdown.");
+                // If we were unable to extract the internal delegate of the event then
+                // we just fallback and run the code here.
+                HandleCancelKeyPress(sender, e);
             }
-            finally
+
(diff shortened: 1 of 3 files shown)


src/Datadog.Trace/Tracer.cs
```diff
@@ -671,7 +671,39 @@ private void InitializeLibLogScopeEventSubscriber(IScopeManager scopeManager, st
...
+                    RunShutdownTasks();
+                });
+
+                // Try to modify the MulticastDelegate invocation list and set the last delegate.
+                if (!DelegatesHelper.TrySetLastDelegate(processExitDelegate, lastDelegate))
+                {
+                    // This means we are in the last delegate already
+                    // or we were unable to set the last delegate, in
+                    // any case we fallback by running the delegate now.
+                    lastDelegate(sender, e);
```
andrewlock, line 698: Do we need to call `lastDelegate(sender, e)`? Can we just run `RunShutdownTasks()` like in the `else` fallback case below?

andrewlock (approved):
Nice! 😄 Witchcraft! 🧙

test/Datadog.Trace.Tests/DelegatesHelperTests.cs
```diff
@@ -0,0 +1,137 @@
...
+            Assert.Equal(2, lstEvents[2]);
+            Assert.Equal(3, lstEvents[3]);
+
+            void TargetObject_MyEvent(object sender, EventArgs e)
+            {
+                lstEvents.Add(1);
+
+                var targetDelegate = TargetObject.GetDelegate();
+
+                var lastDelegate = new EventHandler((s, eArgs) =>
```
zacharycmontoya, line 111: In a separate unit test, can you set `var lastDelegate` to `TargetObject_MyEvent` to make sure it doesn't add itself? Based on my understanding, that's the role of the small stackframe capture check in `TrySetLastDelegate`

zacharycmontoya (approved):
Seriously, some pure wizardry right here!

test/Datadog.Trace.Tests/DelegatesHelperTests.cs
tonyredondo: done in f861559

src/Datadog.Trace/Tracer.cs
tonyredondo: done in f861559

src/Datadog.Trace/Tracer.cs
```diff
@@ -682,7 +714,17 @@ private void CurrentDomain_UnhandledException(object sender, UnhandledExceptionE
 
         private void Console_CancelKeyPress(object sender, ConsoleCancelEventArgs e)
         {
-            RunShutdownTasks();
+            // Console Cancel KeyPress is raised on SIGINT and SIGQUIT signal,
+            // here we ensure we flush and not close the writer.
+            // There's a possibility of a graceful shutdown handled by ProcessExit.
+            try
+            {
+                _agentWriter.FlushTracesAsync().GetAwaiter().GetResult();
```
lucaspimentel, line 722: Is there any danger of a deadlock here? For example, if there's a custom `SynchronizationContext`?

src/Datadog.Trace/Util/DelegatesHelper.cs
```diff
@@ -0,0 +1,109 @@
...
+    {
+        /// <summary>
+        /// Tries to get the internal ProcessExit delegate from the event using reflection.
+        /// </summary>
+        /// <returns>MulticastDelegate instance or null</returns>
+        public static MulticastDelegate GetInternalProcessExitDelegate()
+        {
+            // This methods tries to get the internal delegte for the process exit event.
+
+#if NETSTANDARD2_0 || NETCOREAPP || NETCOREAPP3_1_OR_GREATER
```
lucaspimentel, line 20: Do we need `NETCOREAPP3_1_OR_GREATER`? Isn't `NETCOREAPP` enough?

lucaspimentel:
I see a lot of comments about `SIGTERM`, `SIGINT`, `SIGQUIT`, and `SIGKILL`. Was any of this tested on Windows (.NET Core _and_ .NET Framework)?

src/Datadog.Trace/Tracer.cs
```diff
@@ -671,7 +671,39 @@ private void InitializeLibLogScopeEventSubscriber(IScopeManager scopeManager, st
 
         private void CurrentDomain_ProcessExit(object sender, EventArgs e)
         {
-            RunShutdownTasks();
+            // This handles a grateful shutdown or a SIGTERM
```
lucaspimentel, line 674:
```suggestion
            // This handles a graceful shutdown or a SIGTERM
```

src/Datadog.Trace/Tracer.cs
```diff
@@ -682,7 +714,17 @@ private void CurrentDomain_UnhandledException(object sender, UnhandledExceptionE
 
         private void Console_CancelKeyPress(object sender, ConsoleCancelEventArgs e)
         {
-            RunShutdownTasks();
+            // Console Cancel KeyPress is raised on SIGINT and SIGQUIT signal,
```
lucaspimentel, line 717: This is also raised when user presses CTRL+C in a Windows console app :) 
Did we test that this and `ProcessExit` still interact correctly?

src/Datadog.Trace/Tracer.cs
tonyredondo: Adding the synchronization context check causes no harm. In the other hand I’ll be suprised if this method could be called with a synchronization context at all. Being raised by an OS notification.

src/Datadog.Trace/Util/DelegatesHelper.cs
tonyredondo: Do you know if NETCOREAPP includes NET5? That was the reason to add it.

lucaspimentel: Good point. I always think of .NET Core as including .NET 5. We should check.

tonyredondo:
> I see a lot of comments about `SIGTERM`, `SIGINT`, `SIGQUIT`, and `SIGKILL`. Was any of this tested on Windows (.NET Core _and_ .NET Framework)?

The principal delegates functionality is being tested with the unit tests in all platforms and framework versions, including arm64.

About both events CTRL+C and ProcessExit. The behavior is pretty similar, the only difference is now we don’t close the agent writer on CTRL+C. Because we were assuming that key combination always finish the process and that’s not true as seen in the example with the dd-trace tool and the .Cancel property in the event args.

However I’ll run a simple console app test with a tracer instance before merge to be 100% sure.

src/Datadog.Trace/Tracer.cs
tonyredondo: I’ll add the comment about CTRL+C. About the test, see my comment below.

src/Datadog.Trace/Util/DelegatesHelper.cs
andrewlock: Yep, .NET 5 defines `NETCOREAPP`: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives (tested and confirmed)

tonyredondo: Great!!! Then I will remove the “greater” one

src/Datadog.Trace/Tracer.cs
tonyredondo: Yeah, it doesn't seems to use any syncronization context, the signals comes directly from the OS and in .NET Framework is handled by queueing to the [ThreadPool](https://referencesource.microsoft.com/#mscorlib/system/console.cs,1758), in .NET 5 is handled from a new [background thread (because the actual one is called from unmanaged code)](https://source.dot.net/#System.Console/System/ConsolePal.Unix.cs,1469)

In .NET Core 2.1: [Unix](https://github.com/dotnet/corefx/blob/release/2.1/src/System.Console/src/System/ConsolePal.Unix.cs#L1253) / Windows the delegate is registered by [this interop](https://github.com/dotnet/corefx/blob/release/2.1/src/Common/src/Interop/Windows/kernel32/Interop.SetConsoleCtrlHandler.cs#L19)

I don't think adding the `SyncronizationContext` check will change anything here.

tonyredondo: Because causes no harm, done in d39e660

src/Datadog.Trace/Util/DelegatesHelper.cs
tonyredondo: done in d39e660

src/Datadog.Trace/Tracer.cs
tonyredondo: I've improved the comment in d39e660

tonyredondo:
I ended up using the same logic for `CancelKeyPress` event as well... to handle all the cases...

#### Scenario 1 and 2: 
Code:
```csharp
        static async Task Main(string[] args)
        {
            _ = Tracer.Instance;
            Console.WriteLine("Hello World!");
            await Task.Delay(5000).ConfigureAwait(false);
        }
```

1) Wait for normal exit (ProcessExit handler):
```bash
Hello World!
  FlushAndCloseAsync

```
2) `Ctrl+C` before finishing the `Task.Delay` (CancelKeyPress handler / ProcessExit didn't trigger):
```bash
Hello World!
  FlushAndCloseAsync
^C
```

#### Scenario 3 and 4:
Code:
```csharp
        static async Task Main(string[] args)
        {
            _ = Tracer.Instance;
            AppDomain.CurrentDomain.ProcessExit += CurrentDomain_ProcessExit;
            Console.WriteLine("Hello World!");
            await Task.Delay(5000).ConfigureAwait(false);
        }

        private static void CurrentDomain_ProcessExit(object sender, EventArgs e)
        {
            Console.WriteLine("Finishing the application...");
        }
```

1) Wait for normal exit (ProcessExit handler):
```bash
Hello World!
Finishing the application...
  FlushAndCloseAsync
```
2) `Ctrl+C` before finishing the `Task.Delay` (CancelKeyPress handler / ProcessExit didn't trigger):
```bash
Hello World!
  FlushAndCloseAsync
^C
```

#### Scenario 5 and 6:
Code:
```csharp
        static async Task Main(string[] args)
        {
            _ = Tracer.Instance;
            Console.CancelKeyPress += Console_CancelKeyPress;
            Console.WriteLine("Hello World!");
            await Task.Delay(5000).ConfigureAwait(false);
        }

        private static void Console_CancelKeyPress(object sender, ConsoleCancelEventArgs e)
        {
            Console.WriteLine($"Cancel Key Pressed: {e.Cancel}");
            e.Cancel = true; // Don't exit the process
        }
```

1) Wait for normal exit (ProcessExit handler):
```bash
Hello World!
  FlushAndCloseAsync
```
2) `Ctrl+C` before finishing the `Task.Delay` [CancelKeyPress handler (FlushTracesAsync) and ProcessExit handler (FlushAndCloseAsync)]:
```bash
Hello World!
Cancel Key Pressed: False
  FlushTracesAsync
  FlushAndCloseAsync
```

#### Scenario 7 and 8:
Code:
```csharp
        static async Task Main(string[] args)
        {
            _ = Tracer.Instance;
            Console.CancelKeyPress += Console_CancelKeyPress;
            Console.WriteLine("Hello World!");
            await Task.Delay(5000).ConfigureAwait(false);
        }

        private static void Console_CancelKeyPress(object sender, ConsoleCancelEventArgs e)
        {
            Console.WriteLine($"Cancel Key Pressed: {e.Cancel}");
            e.Cancel = false; // Exit the process
        }
```

1) Wait for normal exit (ProcessExit handler):
```bash
Hello World!
  FlushAndCloseAsync
```
2) `Ctrl+C` before finishing the `Task.Delay` [CancelKeyPress handler (FlushAndCloseAsync)]:
```bash
Hello World!
Cancel Key Pressed: False
  FlushAndCloseAsync
^C
```

#### Scenario 9 and 10:
Code:
```csharp
    class Program
    {
        static async Task Main(string[] args)
        {
            _ = Tracer.Instance;
            Console.CancelKeyPress += Console_CancelKeyPress;
            AppDomain.CurrentDomain.ProcessExit += CurrentDomain_ProcessExit;
            Console.WriteLine("Hello World!");
            await Task.Delay(5000).ConfigureAwait(false);
        }

        private static void CurrentDomain_ProcessExit(object sender, EventArgs e)
        {
            Console.WriteLine("Finishing the application...");
        }

        private static void Console_CancelKeyPress(object sender, ConsoleCancelEventArgs e)
        {
            Console.WriteLine($"Cancel Key Pressed: {e.Cancel}");
            e.Cancel = true; // Don't exit the process
        }
    }
```

1) Wait for normal exit (ProcessExit handler):
```bash
Hello World!
Finishing the application...
  FlushAndCloseAsync
```
2) `Ctrl+C` before finishing the `Task.Delay` [CancelKeyPress handler (FlushTracesAsync) and ProcessExit handler (FlushAndCloseAsync)]:
```bash
Hello World!
Cancel Key Pressed: False
  FlushTracesAsync
Finishing the application...
  FlushAndCloseAsync
```

#### Scenario 11 and 12:
Code:
```csharp
    class Program
    {
        static async Task Main(string[] args)
        {
            _ = Tracer.Instance;
            Console.CancelKeyPress += Console_CancelKeyPress;
            AppDomain.CurrentDomain.ProcessExit += CurrentDomain_ProcessExit;
            Console.WriteLine("Hello World!");
            await Task.Delay(5000).ConfigureAwait(false);
        }

        private static void CurrentDomain_ProcessExit(object sender, EventArgs e)
        {
            Console.WriteLine("Finishing the application...");
        }

        private static void Console_CancelKeyPress(object sender, ConsoleCancelEventArgs e)
        {
            Console.WriteLine($"Cancel Key Pressed: {e.Cancel}");
            e.Cancel = false; // Exit the process
        }
    }
```

1) Wait for normal exit (ProcessExit handler):
```bash
Hello World!
Finishing the application...
  FlushAndCloseAsync
```
2) `Ctrl+C` before finishing the `Task.Delay` [CancelKeyPress handler (FlushTracesAsync) and ProcessExit handler (FlushAndCloseAsync)]:
```bash
Hello World!
Cancel Key Pressed: False
  FlushAndCloseAsync
^C
```

tonyredondo:
For `IHostApplicationLifetime` I've tested it with a simple Worker app:

```csharp
    public class Worker : BackgroundService
    {
        private readonly ILogger<Worker> _logger;

        public Worker(ILogger<Worker> logger)
        {
            _logger = logger;
        }

        protected override async Task ExecuteAsync(CancellationToken stoppingToken)
        {
            _ = Tracer.Instance;
            while (!stoppingToken.IsCancellationRequested)
            {
                _logger.LogInformation("Worker running at: {time}", DateTimeOffset.Now);
                await Task.Delay(1000, stoppingToken);
            }
        }
    }
```

Output after pressing CTRL+C:

```bash
info: WorkerService1.Worker[0]
      Worker running at: 04/05/2021 14:22:27 +02:00
info: Microsoft.Hosting.Lifetime[0]
      Application started. Press Ctrl+C to shut down.
info: Microsoft.Hosting.Lifetime[0]
      Hosting environment: Development
info: Microsoft.Hosting.Lifetime[0]
      Content root path: C:\Users\danielredondo\source\repos\WorkerService1\WorkerService1
info: WorkerService1.Worker[0]
      Worker running at: 04/05/2021 14:22:28 +02:00
info: WorkerService1.Worker[0]
      Worker running at: 04/05/2021 14:22:29 +02:00
info: WorkerService1.Worker[0]
      Worker running at: 04/05/2021 14:22:30 +02:00
info: Microsoft.Hosting.Lifetime[0]
      Application is shutting down...
  FlushTracesAsync
  FlushAndCloseAsync

C:\Users\danielredondo\source\repos\WorkerService1\WorkerService1\bin\Debug\netcoreapp3.1\WorkerService1.exe (process 4908) exited with code 0.
```

lucaspimentel (approved):

PR #1336, opened 2021-04-02
Generate fake proj for dependabot

colin-higgins:
Allows us to use dependabot for vendored dependencies.

🍯 🐝 🐝 🐝

@DataDog/apm-dotnet


.dependabot/config.yaml
```diff
@@ -0,0 +1,7 @@
+version: 1
```
zacharycmontoya, line 1: Silly question, is it okay that it's not `config.yml`? I don't know how sensitive the tooling is to the extension

colin-higgins: Not a silly question, just an oversight on my part, thanks.
I don't know if it would work or not actually.

zacharycmontoya (approved):
LGTM

honeypot/Datadog.Dependabot.Honeypot.csproj
```diff
@@ -0,0 +1,26 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
+    <RootNamespace>Datadog.Dependabot.Honeypot</RootNamespace>
```
lucaspimentel, line 5: Why `Honeypot`?

tools/UpdateVendors/Datadog.Dependabot.Honeypot.txt
```diff
@@ -0,0 +1,11 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
```
lucaspimentel, line 1: Can we add the word `template` to this file? Maybe instead of the `txt` extension?

honeypot/Datadog.Dependabot.Honeypot.csproj
colin-higgins: It seemed like an apt name.
https://en.wikipedia.org/wiki/Honeypot_(computing)
It's not an attacker, but it isn't a REAL project: It's a fake project designed to attract the dependabot.

lucaspimentel: >designed to attract the dependabot.

Ah, I didn't think of it that way. Thanks.

tools/UpdateVendors/Datadog.Dependabot.Honeypot.txt
colin-higgins: Done

lucaspimentel (approved):

PR #1338, opened 2021-04-03
Implement Dependabot Config

colin-higgins:
Previous attempt (#1336) used the documentation for the dependabot preview.

https://docs.github.com/en/code-security/supply-chain-security/configuration-options-for-dependency-updates

We can also use this to keep samples up to date with the latest libraries.

@DataDog/apm-dotnet

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,6 +1,6 @@
-version: 1
+version: 2
 update_configs:
-  - package_manager: "dotnet:nuget"
+  - package_manager: "nuget"
     directory: "/honeypot"
     update_schedule: "daily"
     default_labels:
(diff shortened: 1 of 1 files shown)


lucaspimentel:
Does this keep the default settings? Dependabot already looks at all our `csproj` and `packages.config` files and we don't want to remove those.

lucaspimentel:
I was wrong. Dependabot is checking for _security_ updates in all projects. This adds checks for all updates (not only security-related). Still, it would be helpful to include all our projects here and not only vendored code.

lucaspimentel (approved):
Thanks!

tonyredondo (approved):

PR #1345, opened 2021-04-05
Upgrade Serilog & Serilog.Sinks.File Vendors

colin-higgins:
Handles PRS: 
https://github.com/DataDog/dd-trace-dotnet/pull/1342
https://github.com/DataDog/dd-trace-dotnet/pull/1340

Re-applies:
https://github.com/DataDog/dd-trace-dotnet/commit/5183d979377c01008437dd0894fb6889e4f981ae#diff-eac63a30dea0728e970ab1a177a57414c065c6cf26af5f4ea0b9db257d3a5492R38

@DataDog/apm-dotnet


src/Datadog.Trace/Vendors/Serilog.Sinks.File/Sinks/File/PathRoller.cs
```diff
@@ -56,7 +56,8 @@ public PathRoller(string path, RollingInterval interval)
                 "(?<" + PeriodMatchGroup + ">\\d{" + _periodFormat.Length + "})" +
                 "(?<" + SequenceNumberMatchGroup + ">_[0-9]{3,}){0,1}" +
                 Regex.Escape(_filenameSuffix) +
-                "$");
+                "$",
+                RegexOptions.Compiled);
```
colin-higgins, line 60: Slower construction, faster matching. I wonder if this will be noticeable in the relenv?

src/Datadog.Trace/Vendors/Serilog.Sinks.File/Sinks/File/PeriodicFlushToDiskSink.cs
```diff
@@ -13,6 +27,7 @@ namespace Datadog.Trace.Vendors.Serilog.Sinks.File
     /// <summary>
     /// A sink wrapper that periodically flushes the wrapped sink to disk.
     /// </summary>
+    [Obsolete("This type will be removed from the public API in a future version; use `WriteTo.File(flushToDiskInterval:)` instead.")]
```
colin-higgins, line 30: Should we remove classes like this to make our package smaller or is that too micro?

andrewlock: I would have said it's not worth the hassle personally, especially if we're going to be doing this vendoring more often, but I'm not passionate about it either way

andrewlock (approved):

src/Datadog.Trace/Vendors/Serilog/Core/LevelOverrideMap.cs
```diff
@@ -66,11 +66,19 @@ public LevelOverride(string context, LoggingLevelSwitch levelSwitch)
                 .ToArray();
         }
 
-        public void GetEffectiveLevel(string context, out LogEventLevel minimumLevel, out LoggingLevelSwitch levelSwitch)
+        public void GetEffectiveLevel(
+#if FEATURE_SPAN
```
colin-higgins, line 70: We can enable this for netcore if we change up the flags with a transform. Thoughts?

colin-higgins: Done.

src/Datadog.Trace/Datadog.Trace.csproj
```diff
@@ -46,11 +46,15 @@
...
   </PropertyGroup>
 
-  <PropertyGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' OR '$(TargetFramework)' == 'netcoreapp3.1' ">
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
     <DefineConstants>$(DefineConstants);ASYNCLOCAL;HASHTABLE</DefineConstants>
   </PropertyGroup>
 
-  <!-- Remove System.Collections.NonGeneric. This was defined in Serilog 2.8.0 but it is unnecessary -->
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'netcoreapp3.1' ">
+    <DefineConstants>$(DefineConstants);ASYNCLOCAL;HASHTABLE;FEATURE_DEFAULT_INTERFACE;FEATURE_SPAN</DefineConstants>
```
colin-higgins, line 54: I had to split the definition of netstandard2.0 and netcoreapp3.1.
The original has netstandard2.1, which makes me wonder if we should start to compile against that and start taking advantage of spans throughout the main code base.

andrewlock: As we discussed, netstandard2.1 doesn't give us anything over .NET Core 3.0, as we don't currently intend to support Mono or Xamarin, so I think this is the best option for now

andrewlock (approved):
I wonder if we should vendor the unit tests in too, so that we can be sure we haven't broken anything in the vendoring process?

src/Datadog.Trace/Vendors/Serilog.Sinks.File/Sinks/File/PeriodicFlushToDiskSink.cs
zacharycmontoya: I agree that we should minimize the manual changes we make

src/Datadog.Trace/Vendors/Serilog/Parsing/PropertyToken.cs
```diff
@@ -60,7 +60,8 @@ public PropertyToken(string propertyName, string rawText, string formatObsolete,
         /// <param name="alignment">The alignment applied to the property, if any.</param>
```
zacharycmontoya, line 60: How did you spot the above `readonly int? _position;` change? Did you have to manually revert that?

src/Datadog.Trace/Vendors/Serilog/Core/LevelOverrideMap.cs
zacharycmontoya: You didn't manually add this to the file, right? You just found the ifdef and added it to our csproj?

zacharycmontoya (approved):
Looks good! Before merging, let's also make sure that our transitive dependencies haven't increased past the framework. The fact that no NuGet packages were added in the `Datadog.Trace.csproj` is a really good sign, but I would feel even more comfortable knowing that running the `PrepareRelease` tool results in the same files being published in the MSI.

src/Datadog.Trace/Vendors/Serilog/Parsing/PropertyToken.cs
colin-higgins: I checked the commit history on the vendor directly and combed through them.
I manually changed it back.

src/Datadog.Trace/Vendors/Serilog/Core/LevelOverrideMap.cs
colin-higgins: Correct.

lucaspimentel (approved):

colin-higgins:
I compared the output of a local msi build to the latest release.
image

PR #1347, opened 2021-04-06
Log a warning if partial flush is enabled with an old agent

kevingosse:
Partial flush requires agent version 7.26.0+.

Note that 7.26.0 does not report the version (this feature is added in 7.27.0), so there is a window of time during which customers may get the warning even though they run with a supported agent.

### DIFF
--- a/src/Datadog.Trace/Agent/Api.cs
+++ b/src/Datadog.Trace/Agent/Api.cs
@@ -29,6 +29,7 @@ namespace Datadog.Trace.Agent
         {
         }
 
+        // Internal constructor used for tests
         internal Api(Uri baseEndpoint, IApiRequestFactory apiRequestFactory, IDogStatsd statsd, IDatadogTracer tracer)
         {
             Log.Debug("Creating new Api");
--- a/src/Datadog.Trace/Agent/Api.cs
+++ b/src/Datadog.Trace/Agent/Api.cs
@@ -2,7 +2,6 @@ using System;
 using System.Collections.Generic;
 using System.Net.Sockets;
 using System.Threading.Tasks;
-using Datadog.Trace.Agent.MessagePack;
 using Datadog.Trace.Agent.Transports;
 using Datadog.Trace.DogStatsd;
 using Datadog.Trace.Logging;
@@ -20,14 +19,20 @@ namespace Datadog.Trace.Agent
 
         private readonly IApiRequestFactory _apiRequestFactory;
         private readonly IDogStatsd _statsd;
-        private readonly FormatterResolverWrapper _formatterResolver = new FormatterResolverWrapper(SpanFormatterResolver.Instance);
         private readonly string _containerId;
         private readonly Uri _tracesEndpoint;
+        private readonly IDatadogTracer _tracer;
         private string _cachedResponse;
 
         public Api(Uri baseEndpoint, IApiRequestFactory apiRequestFactory, IDogStatsd statsd)
+            : this(baseEndpoint, apiRequestFactory, statsd, tracer: null)
+        {
+        }
+
+        internal Api(Uri baseEndpoint, IApiRequestFactory apiRequestFactory, IDogStatsd statsd, IDatadogTracer tracer)
         {
             Log.Debug("Creating new Api");
+            _tracer = tracer;
             _tracesEndpoint = new Uri(baseEndpoint, TracesPath);
             _statsd = statsd;
             _containerId = ContainerMetadata.GetContainerId();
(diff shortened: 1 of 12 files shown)


src/Datadog.Trace/Tracer.cs
```diff
@@ -353,6 +355,18 @@ void IDatadogTracer.Write(Span[] trace)
...
 
+        /// <summary>
+        /// Reports the detected version of the agent
+        /// </summary>
+        /// <param name="version">Version of the agent</param>
+        void IDatadogTracer.ReportAgentVersion(string version)
+        {
+            if (ShouldLogPartialFlushWarning(version))
+            {
+                Log.Warning("DATADOG TRACER DIAGNOSTICS - Partial flush should only be enabled with agent 7.26.0+ (detected version: {version})", version);
```
colin-higgins, line 366: If this is detected, do we fallback to non-partial flush?

kevingosse: I can't reliably detect the agent, because 7.26.0 has no version check (even though it's fine to use with partial flush), or simply because the agent may not be responsive at the moment the app is started. Because of that, I think it's better to just log a warning. I'm afraid false negatives could cause much pain.

andrewlock (approved):

colin-higgins (approved):

src/Datadog.Trace/Agent/Api.cs
```diff
@@ -20,14 +19,20 @@ internal class Api : IApi
...
         private readonly Uri _tracesEndpoint;
+        private readonly IDatadogTracer _tracer;
         private string _cachedResponse;
 
         public Api(Uri baseEndpoint, IApiRequestFactory apiRequestFactory, IDogStatsd statsd)
+            : this(baseEndpoint, apiRequestFactory, statsd, tracer: null)
+        {
+        }
+
+        internal Api(Uri baseEndpoint, IApiRequestFactory apiRequestFactory, IDogStatsd statsd, IDatadogTracer tracer)
```
lucaspimentel, line 32: This is already an `internal` class. We usually make members `public` in cases like this.
```suggestion
        public Api(Uri baseEndpoint, IApiRequestFactory apiRequestFactory, IDogStatsd statsd, IDatadogTracer tracer)
```

src/Datadog.Trace/Tracer.cs
lucaspimentel: Isn't this going to say `(detected version: )` with older Agent versions?

lucaspimentel, line 366: Isn't this going to say `(detected version: )` with older Agent versions?

src/Datadog.Trace/Agent/Api.cs
lucaspimentel: I looked into this some more and I think this overload is only used for testing. Can you add a comment to that effect? Thanks.

lucaspimentel: I find the circular dependency weird (`Tracer` -> `IAgentWriter` -> `IApi` and now `Api` -> `IDatadogTracer`), which made think some more.

I'm wondering if `ReportAgentVersion()` should be on `Api` instead of `Tracer`, since communicating to the Agent falls within `Api`'s responsibilities and not `Tracer`'s. In theory, `Tracer` shouldn't know how or where traces are going.

lucaspimentel (changes requested):
>You need to leave a comment indicating the requested changes.

You at my other comments 😅

src/Datadog.Trace/Tracer.cs
kevingosse: Yeah, but we know what it means. Do you prefer that I put a placeholder value?

lucaspimentel: Should we remove `(detected version: {version})` if it's never going to show a version?
```suggestion
                Log.Warning("DATADOG TRACER DIAGNOSTICS - Partial flush should only be enabled with Agent 7.26.0+", version);
```

src/Datadog.Trace/Agent/Api.cs
kevingosse: > I looked into this some more and I think this overload is only used for testing. Can you add a comment to that effect? Thanks.

Sure

> I find the circular dependency weird (Tracer -> IAgentWriter -> IApi and now Api -> IDatadogTracer), which made think some more.

The dependency is not new, we're already calling `Tracer.Instance.Sampler.SetDefaultSampleRates(apiResponse?.RateByService);` in Api.

src/Datadog.Trace/Tracer.cs
kevingosse: Fixed in https://github.com/DataDog/dd-trace-dotnet/pull/1347/commits/b43b7e20436d6690d745020454ad8422655ae030

src/Datadog.Trace/Agent/Api.cs
lucaspimentel: Oh, well... this is fine for now, then. Maybe someday we can come back and clean this all up.

lucaspimentel (approved):

zacharycmontoya (approved):

PR #1349, opened 2021-04-07
OTel Port: Fix double parsing

zacharycmontoya:
Source: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/96

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace/Configuration/StringConfigurationSource.cs
+++ b/src/Datadog.Trace/Configuration/StringConfigurationSource.cs
@@ -1,5 +1,6 @@
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Globalization;
 using Datadog.Trace.ExtensionMethods;
 
 namespace Datadog.Trace.Configuration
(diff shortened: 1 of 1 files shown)


kevingosse (approved):
Nice catch. It's a breaking change but that makes it consistent with the documentation.

andrewlock (approved):

src/Datadog.Trace/Configuration/StringConfigurationSource.cs
```diff
@@ -89,7 +90,7 @@ public abstract class StringConfigurationSource : IConfigurationSource
         {
             string value = GetString(key);
 
-            return double.TryParse(value, out double result)
+            return double.TryParse(value, NumberStyles.Any, CultureInfo.InvariantCulture, out double result)
```
lucaspimentel, line 93: `NumberStyles.Any` allows syntax like `1e2` or `(1)`. Are we ok with those? I would think we only want to allow leading and trailing whitespace, decimal point, and maybe leading sign if we even accept negative values.

lucaspimentel: Should we add this to `int.TryParse()` above for consistency? Not sure that it will have any effect, though, since that one should be a string of digits only.

PR #1350, opened 2021-04-07
Simplify moving average calculator

andrewlock:
Was previously going a bit overboard with keeping things atomic, when in reality we only need the current buckets and keep rate to be atomic.

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace/Agent/MovingAverageKeepRateCalculator.cs
+++ b/src/Datadog.Trace/Agent/MovingAverageKeepRateCalculator.cs
@@ -26,7 +26,9 @@ namespace Datadog.Trace.Agent
         private readonly TaskCompletionSource<bool> _processExit = new TaskCompletionSource<bool>();
 
         private int _index = 0;
-        private long _totals = 0;
+        private ulong _sumDrops = 0;
+        private ulong _sumCreated = 0;
+        private double _keepRate = 0;
 
         private long _latestDrops = 0;
         private long _latestKeeps = 0;
@@ -84,14 +86,8 @@ namespace Datadog.Trace.Agent
         /// </summary>
         public double GetKeepRate()
         {
-            var totals = Interlocked.Read(ref _totals);
-            UnpackBits(totals, out var totalDropped, out var totalCreated);
-            if (totalCreated == 0)
-            {
-                return 0;
-            }
-
-            return 1 - ((double)totalDropped / totalCreated);
+            // Essentially Interlock.Read(double)
+            return Interlocked.CompareExchange(ref _keepRate, 0, 0);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


colin-higgins (approved):

zacharycmontoya (approved):

src/Datadog.Trace/Agent/MovingAverageKeepRateCalculator.cs
```diff
@@ -111,21 +107,23 @@ internal void UpdateBucket()
             var previousDropped = _dropped[index];
             var previousCreated = _created[index];
 
-            var latestDropped = Math.Min(Interlocked.Exchange(ref _latestDrops, 0), uint.MaxValue);
-            var latestCreated = Math.Min(Interlocked.Exchange(ref _latestKeeps, 0) + latestDropped, uint.MaxValue);
+            // Cap at unit.MaxValue events. Very unlikely to reach this value!
```
lucaspimentel, line 110:
```suggestion
            // Cap at uint.MaxValue events. Very unlikely to reach this value!
```

src/Datadog.Trace/Agent/MovingAverageKeepRateCalculator.cs
```diff
@@ -84,14 +86,8 @@ public void IncrementDrops(int count)
...
-            var totals = Interlocked.Read(ref _totals);
-            UnpackBits(totals, out var totalDropped, out var totalCreated);
-            if (totalCreated == 0)
-            {
-                return 0;
-            }
-
-            return 1 - ((double)totalDropped / totalCreated);
+            // Essentially Interlock.Read(double)
+            return Interlocked.CompareExchange(ref _keepRate, 0, 0);
```
kevingosse, line 90:
```suggestion
            return Volatile.Read(ref _keepRate);
```

This will emit the same cmpxchg on x86, but more optimized code on x64

src/Datadog.Trace/Agent/MovingAverageKeepRateCalculator.cs
```diff
@@ -111,21 +107,23 @@ internal void UpdateBucket()
...
-            var newTotalDropped = (uint)Math.Min(((long)previousTotalDropped) - previousDropped + latestDropped, uint.MaxValue);
-            var newTotalCreated = (uint)Math.Min(((long)previousTotalCreated) - previousCreated + latestCreated, uint.MaxValue);
+            var keepRate = newSumCreated == 0
+                               ? 0
+                               : 1 - ((double)newSumDropped / newSumCreated);
 
-            // Pack the totals into a single value we can read and write atomically
-            var packedTotal = PackBits(newTotalDropped, newTotalCreated);
-            Interlocked.Exchange(ref _totals, packedTotal);
+            Interlocked.Exchange(ref _keepRate, keepRate);
```
kevingosse, line 121:
```suggestion
            Volatile.Write(ref _keepRate, keepRate);
```
This will produce the same Interlocked.Exchange on x86, but more optimized code on x64 (unless you care about the results being published immediately, which I doubt)

src/Datadog.Trace/Agent/MovingAverageKeepRateCalculator.cs
andrewlock: Really?! Everything I read suggested that `double` read/write wasn't atomic with Volatile 🤔 e.g. [StackOverflow is lying to me](https://stackoverflow.com/questions/12435325/does-volatile-read-volatile-write-for-double-atomic)?

kevingosse: Stackoverflow is talking about the `volatile` keyword, which indeed doesn't guarantee atomicity. The `Volatile` class is different. It's implemented as an Interlocked operation and replaced by the JIT on architectures that allow it: https://referencesource.microsoft.com/#mscorlib/system/threading/volatile.cs,189
Atomicity is guaranteed.

kevingosse: `Volatile.Read` is implemented as an atomic operation (on properly aligned data, so all bets are off if you start using `StructLayout`). It uses Interlocked on 32 bit platforms, and native read/write on 64 bit platforms: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Threading/Volatile.cs#L91

That said, it's true that this atomicity is not documented on MSDN.

andrewlock: You're right of course - I hadn't twigged they were talking about the volatile keyword rather than the class. I can hardly find anything about the class!

Don't suppose you have a blog post about it do you? 😉

lucaspimentel: Do we need to worry about atomicity? i.e. is this code ever called from multiple threads concurrently?

andrewlock: > Do we need to worry about atomicity? i.e. is the read/write code ever called from multiple threads concurrently?

Yes 🙂 the update loop runs on a separate thread to the agent which is returning the value

kevingosse (approved):

lucaspimentel (approved):

PR #1351, opened 2021-04-07
Add benchmark for Serilog log injection

kevingosse:


tonyredondo (approved):

colin-higgins (approved):

test/benchmarks/Benchmarks.Trace/SerilogBenchmark.cs
```diff
@@ -0,0 +1,79 @@
...
+            {
+                StartupDiagnosticLogEnabled = false,
+                LogsInjectionEnabled = false,
+                Environment = "env",
+                ServiceVersion = "version"
+            };
+
+            BaselineTracer = new Tracer(baselineSettings, new DummyAgentWriter(), null, null, null);
+
+            EnrichedLogger = new LoggerConfiguration()
```
zacharycmontoya, line 42: So `.Enrich.FromLogContext()` is currently a requirement for logs injection to work, but should we also measure a third case where `.Enrich.FromLogContext()` is enabled but our logs injection isn't added? That would help isolate the impact of logs injection

andrewlock (approved):

test/benchmarks/Benchmarks.Trace/SerilogBenchmark.cs
kevingosse: That's fair, added in https://github.com/DataDog/dd-trace-dotnet/pull/1351/commits/7e040c75fe1694851165147d0a60030c68214b89

Interestingly it doesn't seem to make a difference in the results. Looks like most of the stuff is lazy, so the first property added to the context makes the biggest impact.

PR #1353, opened 2021-04-08
Add explicit nuget.config to workaround AzDo failures

andrewlock:
Seems to be related to [this issue](https://github.com/actions/virtual-environments/issues/3038) which relates to GitHub actions, but as AzDo runs on the same underlying infrastructure, probably the same problem.

Apparently this is a commonly-reported intermittent issue related to "ambient state" on the builder causing nuget.org to be omitted from the NuGet "Feeds used" (actions/virtual-environments#1090 (comment)). Add the nuget.org packageSource explicitly using nuget.config to avoid the issue, as recommended by:

actions/setup-dotnet#155 (comment)
actions/virtual-environments#3038 (comment)
NuGet/Home#10586 (comment)

@DataDog/apm-dotnet

### DIFF
--- a/nuget.config
+++ b/nuget.config
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <!--To inherit the global NuGet package sources remove the <clear/> line below -->
+    <clear />
+    <add key="nuget" value="https://api.nuget.org/v3/index.json" />
+  </packageSources>
+</configuration>


lucaspimentel (approved):

colin-higgins (approved):

nuget.config
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <!--To inherit the global NuGet package sources remove the <clear/> line below -->
+    <clear />
```
lucaspimentel, line 5: Do we need this `<clear>`? Would this obliterate any feeds in my user-scoped `nuget.config` file. Some feeds currently in my file:
```
<!-- local nupkg files used for testing purposes -->
<add key="local" value="C:\Users\lucas.pimentel\Downloads\nuget-packages" />

<!-- added by VS -->
<add key="Microsoft Visual Studio Offline Packages" value="C:\Program Files (x86)\Microsoft SDKs\NuGetPackages\" />

<-- our feed in AzDO, used ocasionally -->
<add key="Azure DevOps Public Feed" value="https://pkgs.dev.azure.com/datadoghq/dd-trace-dotnet/_packaging/Public_Feed/nuget/v3/index.json" />

<-- added by Service Fabric SDK -->
<add key="Microsoft Azure Service Fabric SDK" value="C:\Program Files\Microsoft SDKs\Service Fabric\packages" />
```

andrewlock: Yes and no. This _will_ blat user-scoped nuget.config, but that was partly intentional to make things reproducible. It also avoids any naming clashes, but I can work around that by providing a random name, and seeing what happens when a feed is added twice 🤔

IMO, if new feeds are required (e.g. service fabric) they should be included here, so it's reproducible. The temp local packages one is something I do too, but normally just include it temporarily. 

Nevertheless, I'll try removing the `<clear />` and see how it goes

lucaspimentel: Don't worry too much about it. I can always remove the `<clear>` locally when needed.

zacharycmontoya:
This was implemented in https://github.com/DataDog/dd-trace-dotnet/pull/1363 so I'll close this PR

PR #1354, opened 2021-04-08
Don't run unit test or runner pipelines on all branch pushes

andrewlock:
There's a lot of duplicated effort caused by these branch-only pushes, and with limited runners available (especially macos), this causes a backlog of runs for PRs.

We can always manually trigger a pipeline for a branch if we want a pre-PR check, so I think this is probably a good option?

@DataDog/apm-dotnet

### DIFF
--- a/.azure-pipelines/runner.yml
+++ b/.azure-pipelines/runner.yml
@@ -1,7 +1,8 @@
 trigger:
   branches:
     include:
-      - '*'
+      - master
+      - refs/tags/*
     exclude:
       - refs/pull/*/head
   paths:
--- a/.azure-pipelines/unit-tests.yml
+++ b/.azure-pipelines/unit-tests.yml
@@ -1,7 +1,8 @@
 trigger:
   branches:
     include:
-      - '*'
+      - master
+      - refs/tags/*
     exclude:
       - refs/pull/*/head
   paths:


kevingosse:
I agree we can refrain from executing the runner pipeline, but the unit-tests seem important?

andrewlock:
> I agree we can refrain from executing the runner pipeline, but the unit-tests seem important?

I don't know @kevingosse , I disagree, but I guess it depends how they're used. Currently,

* I push branches periodically, even when I know they're not "ready"
* That causes a (pointless) run of the unit-test pipeline, that is only visible if you specifically go the unit-tests pipeline page in AzDo. That uses up 4 of the hosted agents which are then not available for running PRs etc (1 for each platform + arm)
* When I do push and make a PR, I then get 2 instances of the unit-test pipeline running on the same commit (one for the branch push, one for the PR)

Given that if I want to see the unit-test pipeline results on a branch-only (non-PR) push I have to go to AzDo _anyway_ and so could easily start a new run, it doesn't seem worth also wasting runners for all the times it's _not_ necessary?

kevingosse:
> > I agree we can refrain from executing the runner pipeline, but the unit-tests seem important?
> 
> I don't know @kevingosse , I disagree, but I guess it depends how they're used. Currently,
> 
> * I push branches periodically, even when I know they're not "ready"
> * That causes a (pointless) run of the unit-test pipeline, that is only visible if you specifically go the unit-tests pipeline page in AzDo. That uses up 4 of the hosted agents which are then not available for running PRs etc (1 for each platform + arm)
> * When I do push and make a PR, I then get 2 instances of the unit-test pipeline running on the same commit (one for the branch push, one for the PR)
> 
> Given that if I want to see the unit-test pipeline results on a branch-only (non-PR) push I have to go to AzDo _anyway_ and so could easily start a new run, it doesn't seem worth also wasting runners for all the times it's _not_ necessary?

Maybe I misunderstood what this PR is doing. 

Basically, I believe unit tests should be run:
 - At every PR
 - At every merge to master

Does that still hold true with your changes?

andrewlock:
> > > I agree we can refrain from executing the runner pipeline, but the unit-tests seem important?
> > 
> > 
> > I don't know @kevingosse , I disagree, but I guess it depends how they're used. Currently,
> > 
> > * I push branches periodically, even when I know they're not "ready"
> > * That causes a (pointless) run of the unit-test pipeline, that is only visible if you specifically go the unit-tests pipeline page in AzDo. That uses up 4 of the hosted agents which are then not available for running PRs etc (1 for each platform + arm)
> > * When I do push and make a PR, I then get 2 instances of the unit-test pipeline running on the same commit (one for the branch push, one for the PR)
> > 
> > Given that if I want to see the unit-test pipeline results on a branch-only (non-PR) push I have to go to AzDo _anyway_ and so could easily start a new run, it doesn't seem worth also wasting runners for all the times it's _not_ necessary?
> 
> Maybe I misunderstood what this PR is doing.
> 
> Basically, I believe unit tests should be run:
> 
> * At every commit to a PR
> * At every merge to master
> 
> Does that still hold true with your changes?

Yes, that still holds true 🙂

This stops us running the pipelines when I push to `andrew/foo-bar` _unless_ there's a PR for that branch

kevingosse:
Sounds good to me then :)

kevingosse (approved):

colin-higgins (approved):

zacharycmontoya (approved):

PR #1360, opened 2021-04-12
Rewrite Accurate_Duration test to reduce flakiness 

kevingosse:
Rewrite the test to prevent failures due to unrelated timing issues. 

The new test can give false positives, but shouldn't give false negatives.

The reasoning behind the test is: let's imagine that Span uses DateTime internally, with a 15 ms precision. Depending on how it's implemented, for a time of 5 ms it will report either 0 ms or 15 ms. If the former, then the first assertion will fail. If the latter, then the second assertion will fail.

### DIFF
--- a/test/Datadog.Trace.Tests/SpanTests.cs
+++ b/test/Datadog.Trace.Tests/SpanTests.cs
@@ -91,51 +91,35 @@ namespace Datadog.Trace.Tests
             Assert.Equal(TimeSpan.Zero, span.Duration);
         }
 
-        [Theory]
-        [InlineData(3)]
-        [InlineData(10)]
-        [InlineData(100)]
-        public void Accurate_Duration(int minimumSleepMilliseconds)
+        [Fact]
+        public void Accurate_Duration()
         {
-            // TODO: refactor how we measure time so we can lower this threshold
+            // Check that span has the same precision as stopwatch
+
             const int iterations = 10;
-            const int maxDifference = 15;
-            TimeSpan totalStopwatchTime = TimeSpan.Zero;
-            TimeSpan totalSpanTime = TimeSpan.Zero;
-            Span span;
             var stopwatch = new Stopwatch();
 
-            // execute once to ensure JIT compilation
-            using (span = _tracer.StartSpan("Operation"))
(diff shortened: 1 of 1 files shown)


andrewlock (approved):
Makes sense to me 👍 Might be worth including the rationale in the PR description as a comment in the test?

test/Datadog.Trace.Tests/SpanTests.cs
```diff
@@ -91,51 +91,35 @@ public void Finish_EndTimeInThePast_DurationIs0()
...
-            // The difference should be less than the threshold
-            _output.WriteLine($"Average elapsed time: {avgStopwatch:0.0} ms");
-            _output.WriteLine($"Average span time: {avgSpan:0.0} ms");
-            Assert.True(
-                stopwatchSpanDiff < maxDifference,
-                $"Span duration difference ({stopwatchSpanDiff}ms) outside of allowed threshold {maxDifference}ms. Expected average: {avgStopwatch}ms, actual average: {avgSpan}ms");
+                stopwatch.Restart();
+                using (span = _tracer.StartSpan("Operation"))
+                {
+                }
```
lucaspimentel, line 117: Any concerns with the span duration here being as short as possible? Should we create fake work so there's at least _some_ duration?

kevingosse: It seems unlikely to happen, but it's true that the test is useless if the work is too short to be measured with `Stopwatch`. Gonna add a Thread.Sleep just to be sure

kevingosse: Done in https://github.com/DataDog/dd-trace-dotnet/pull/1360/commits/22f40459215c1c19c33e5469a4e92bc3c1d60820

kevingosse:
> Makes sense to me 👍 Might be worth including the rationale in the PR description as a comment in the test?

Done in https://github.com/DataDog/dd-trace-dotnet/pull/1360/commits/22f40459215c1c19c33e5469a4e92bc3c1d60820

zacharycmontoya (approved):

lucaspimentel (approved):

PR #1364, opened 2021-04-13
Add symbols to MSI

kevingosse:
Include the PDB symbols to the installer

### DIFF
--- a/src/WindowsInstaller/Files.Managed.Net45.GAC.wxs
+++ b/src/WindowsInstaller/Files.Managed.Net45.GAC.wxs
@@ -13,14 +13,32 @@
       </Component>
       <Component Win64="$(var.Win64)">
         <Condition>NOT WIX_IS_NETFRAMEWORK_461_OR_LATER_INSTALLED</Condition>
+        <File Id="net45_GAC_Datadog.Trace.AspNet.pdb"
+              Source="$(var.TracerHomeDirectory)\net45\Datadog.Trace.AspNet.pdb"
+              KeyPath="yes" Checksum="yes" Assembly=".net"/>
+      </Component>
+      <Component Win64="$(var.Win64)">
+        <Condition>NOT WIX_IS_NETFRAMEWORK_461_OR_LATER_INSTALLED</Condition>
         <File Id="net45_GAC_Datadog.Trace.ClrProfiler.Managed.Core.dll"
               Source="$(var.TracerHomeDirectory)\net45\Datadog.Trace.ClrProfiler.Managed.Core.dll"
               KeyPath="yes" Checksum="yes" Assembly=".net"/>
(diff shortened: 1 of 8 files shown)


colin-higgins:
Those wix files are actually kept in sync with the PrepareRelease tool.
We should probably add a comment mentioning they are auto-generated.
image
https://github.com/DataDog/dd-trace-dotnet/blob/master/build/tools/PrepareRelease/Program.cs#L45
https://github.com/DataDog/dd-trace-dotnet/blob/master/build/tools/PrepareRelease/SyncMsiContent.cs

lucaspimentel:
For the dlls we install in the GAC, we should copy the PDBs into `%WINDIR%\System32\Symbols\dll` (in addition to `Program Files` like you did here).

kevingosse:
> For the dlls we install in the GAC, we should copy the PDBs into `%WINDIR%\System32\Symbols\dll` (in addition to `Program Files` like you did here).

I've considered it, but I'm not sure we should. I don't really like putting additional files into System32, and it doesn't bring much (the symbols are still available in the Datadog folder, the developer just has to load them manually).

Not to mention that the GAC is versioned and `%WINDIR%\System32\Symbols\dll` wouldn't be.

lucaspimentel (changes requested):
See @colin-higgins's comment about the `PrepareRelease` tool: https://github.com/DataDog/dd-trace-dotnet/pull/1364#issuecomment-818742470

kevingosse:
> Those wix files are actually kept in sync with the PrepareRelease tool.

I was wondering why nobody ever reformatted those files 😄

src/WindowsInstaller/Files.Managed.Net45.GAC.wxs
```diff
@@ -11,18 +11,36 @@
               Source="$(var.TracerHomeDirectory)\net45\Datadog.Trace.AspNet.dll"
               KeyPath="yes" Checksum="yes" Assembly=".net"/>
       </Component>
+      <Component Win64="$(var.Win64)">
```
zacharycmontoya, line 14: Looks like we need additional logic...we don't want the pdb files to be added to this "Net45.GAC" file or the "Net461.GAC" file

kevingosse: ugh, good catch

kevingosse: Fixed in https://github.com/DataDog/dd-trace-dotnet/pull/1364/commits/fe760dd3272ceddde06de35493b7882e0bdd6dca

zacharycmontoya (approved):
LGTM good stuff!

lucaspimentel (approved):

PR #1365, opened 2021-04-13
Include symbols in Linux packages

kevingosse:


lucaspimentel:
Is the Linux build already generating pdb files?

tonyredondo:
> Is the Linux build already generating pdb files?

Do you mean for native? Linux native libs doesn’t have pdb as windows. But we need to enable the debugging info within the release type using RelWithDebInfo in CMAKE.

https://cmake.org/cmake/help/v3.0/variable/CMAKE_BUILD_TYPE.html

kevingosse:
I'm only focusing on the managed PDBs for now.

I confirm that with this change they end up in the artifacts as expected: https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=57971&view=artifacts&pathAsName=false&type=publishedArtifacts

lucaspimentel:
I meant for managed code, since this PR is only packaging the PDBs for `netstandard2.0` and `netcoreapp3.1` (it assumes we already generate them?), but you also make good point wrt native. Could be a separate PR, though. One step at a time? 😅

zacharycmontoya (approved):

lucaspimentel (approved):

PR #1400, opened 2021-04-14
Stop calling Environment.Exit in tests

kevingosse:
I've seen a case where traces are not correctly flushed at the end of the test, and I wonder if it could somehow be related to `Environment.Exit`. Marking the HttpListener thread as background should achieve the same result, so we'll see if the failure happens again.


tonyredondo (approved):

lucaspimentel (approved):
Nice catch.

PR #1401, opened 2021-04-14
Generate nuget symbols package

kevingosse:
This will generate an additional snupkg package that contains the symbols for the library. The package can then be uploaded to nuget symbol server: https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg#nugetorg-symbol-server

### DIFF
--- a/.azure-pipelines/packages.yml
+++ b/.azure-pipelines/packages.yml
@@ -71,7 +71,6 @@ stages:
         packagesToPack: src/Datadog.Trace/Datadog.Trace.csproj;src/Datadog.Trace.OpenTracing/Datadog.Trace.OpenTracing.csproj
         packDirectory: $(System.DefaultWorkingDirectory)/nuget-output
         configuration: $(buildConfiguration)
-        arguments: -p:IncludeSymbols=true -p:SymbolPackageFormat=snupkg
 
     - publish: $(System.DefaultWorkingDirectory)/nuget-output
       artifact: nuget-packages
--- a/.azure-pipelines/packages.yml
+++ b/.azure-pipelines/packages.yml
@@ -71,6 +71,7 @@ stages:
         packagesToPack: src/Datadog.Trace/Datadog.Trace.csproj;src/Datadog.Trace.OpenTracing/Datadog.Trace.OpenTracing.csproj
         packDirectory: $(System.DefaultWorkingDirectory)/nuget-output
         configuration: $(buildConfiguration)
+        arguments: -p:IncludeSymbols=true -p:SymbolPackageFormat=snupkg
 
     - publish: $(System.DefaultWorkingDirectory)/nuget-output
       artifact: nuget-packages
--- a/src/Datadog.Trace.OpenTracing/Datadog.Trace.OpenTracing.csproj
+++ b/src/Datadog.Trace.OpenTracing/Datadog.Trace.OpenTracing.csproj
@@ -6,12 +6,17 @@
     <Title>Datadog APM - OpenTracing</Title>
     <Description>Provides OpenTracing support for Datadog APM</Description>
     <PackageTags>$(PackageTags);OpenTracing</PackageTags>
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
+    <IncludeSymbols>true</IncludeSymbols>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
   </PropertyGroup>
 
   <ItemGroup>
     <ProjectReference Include="..\Datadog.Trace\Datadog.Trace.csproj" />
 
     <PackageReference Include="OpenTracing" Version="0.12.1" />
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All"/>
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' OR '$(TargetFramework)' == 'netcoreapp3.1' ">
--- a/src/Datadog.Trace.OpenTracing/Datadog.Trace.OpenTracing.csproj
+++ b/src/Datadog.Trace.OpenTracing/Datadog.Trace.OpenTracing.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <!-- NuGet -->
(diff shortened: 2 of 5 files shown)


.azure-pipelines/packages.yml
```diff
@@ -71,6 +71,7 @@ stages:
         packagesToPack: src/Datadog.Trace/Datadog.Trace.csproj;src/Datadog.Trace.OpenTracing/Datadog.Trace.OpenTracing.csproj
         packDirectory: $(System.DefaultWorkingDirectory)/nuget-output
         configuration: $(buildConfiguration)
+        arguments: -p:IncludeSymbols=true -p:SymbolPackageFormat=snupkg
```
lucaspimentel, line 74: We should add this in the gitlab script (`tasks/dotnettracer.py`) as well, since we want to upload signed nuget packages soon.

Or would it be easier to add this properties to the two `csproj` files instead?

kevingosse: Good point, gonna move that to the csproj

kevingosse: Done in https://github.com/DataDog/dd-trace-dotnet/pull/1401/commits/15d0a5c24075eac04ecc86a1db1fd18cbd4e7570

lucaspimentel:
One more thing (sorry 😅). Looks like we should also enable deterministic builds:

https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/sourcelink
links to
https://github.com/clairernovotny/DeterministicBuilds

kevingosse:
> One more thing, since you're adding SourceLink (sorry 😅). Looks like we should also enable deterministic builds in CI:
> 
> https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/sourcelink
> links to
> https://github.com/clairernovotny/DeterministicBuilds

Well I don't really know what it brings, but I don't mind.

src/Datadog.Trace.OpenTracing/Datadog.Trace.OpenTracing.csproj
```diff
@@ -1,17 +1,26 @@
...
     <Description>Provides OpenTracing support for Datadog APM</Description>
     <PackageTags>$(PackageTags);OpenTracing</PackageTags>
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
+    <IncludeSymbols>true</IncludeSymbols>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(TF_BUILD)' == 'true'">
+    <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
```
lucaspimentel, line 16: We want to add `ContinuousIntegrationBuild` in gitlab builds, too. Add as property here? https://github.com/DataDog/dd-trace-dotnet/blob/eed37a0999a38804105b5c4ed2a7b0f1fcd5bd54/tasks/dotnettracer.py#L41

lucaspimentel: Unless there's a way to detect gitlab builds like you did with `TF_BUILD`.

src/Datadog.Trace/Datadog.Trace.csproj
```diff
@@ -23,8 +27,13 @@
 
   <ItemGroup>
     <PackageReference Include="LibLog" Version="5.0.6" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All"/>
   </ItemGroup>
 
+  <PropertyGroup Condition="'$(TF_BUILD)' == 'true'">
+    <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
```
lucaspimentel, line 34: Same as previous comment.

lucaspimentel:
> Well I don't really know what it brings

I'm not an expert, but the docs I linked above say:
> Deterministic builds enable verification that the resulting binary was built from the specified source and provide traceability.

Also, [from Wikipedia](https://en.wikipedia.org/wiki/Reproducible_builds) (emphasis mine):
> Reproducible builds, also known as deterministic compilation, is a process of compiling software which ensures the resulting binary code can be reproduced. Source code compiled using deterministic compilation will always output the same binary.
> 
> Reproducible builds can act as **part of a chain of trust**; the source code can be signed, and deterministic compilation **can prove that the binary was compiled from trusted source code**.

kevingosse:
Sure, but unless we set an autonomous system that downloads the artifacts, tries building from source, and checks that the binaries match, it doesn't bring any additional safety by itself 🙂

src/Datadog.Trace.OpenTracing/Datadog.Trace.OpenTracing.csproj
kevingosse: Done in https://github.com/DataDog/dd-trace-dotnet/pull/1401/commits/1e7fcf739a4acd23493549956a6c6aba69ac5b37
Is there any way to test it's properly working? 🤔

lucaspimentel: We can trigger a gitlab build and check the output.

lucaspimentel (approved):

src/Datadog.Trace.OpenTracing/Datadog.Trace.OpenTracing.csproj
zacharycmontoya: To minimize changes in the GitLab steps, we should check `'$(GITLAB_CI)' == 'true'` in this csproj (source: https://docs.gitlab.com/ee/ci/variables/predefined_variables.html)

src/Datadog.Trace.OpenTracing/Datadog.Trace.OpenTracing.csproj
```diff
@@ -1,17 +1,26 @@
...
   </PropertyGroup>
```
zacharycmontoya, line 17: Can we also set `Deterministic=true` for CI builds?

zacharycmontoya: Nevermind, apparently it defaults to true: https://devblogs.microsoft.com/dotnet/producing-packages-with-source-link/#deterministic-builds

src/Datadog.Trace.OpenTracing/Datadog.Trace.OpenTracing.csproj
zacharycmontoya: What do you think of moving the PropertyGroup that sets the ContinuousIntegrationBuild variable into the parent `Directory.Build.props` so all of the projects receive the same treatment?

zacharycmontoya (approved):

andrewlock (approved):

PR #990, opened 2020-10-19
Revert usage of IgnoresAccessChecksToAttribute in DuckTyping due incompatibility to partially trusted scenarios.

tonyredondo:
Revert usage of IgnoresAccessChecksToAttribute in DuckTyping due incompatibility to partially trusted scenarios.

The custom `IgnoresAccessChecksToAttribute` doesn't work in partially trusted scenarios, so we remove it and handle it with the normal fallback to the DynamicMethod approach used in .NET Framework 4.5.

This fixes the error for redis integration tests:
```
2020-10-19T11:04:54.3572110Z  StandardOutput:
2020-10-19T11:04:54.3572319Z  Running test with RedisClient
2020-10-19T11:04:54.3572530Z  Running sync operations
2020-10-19T11:04:54.3572739Z  Setting string...
2020-10-19T11:04:54.3572920Z  Getting string...
2020-10-19T11:04:54.3573141Z  Values match. value1 == value1
2020-10-19T11:04:54.3573320Z  
2020-10-19T11:04:54.3573486Z  Running async operations
2020-10-19T11:04:54.3573698Z  Setting string...
2020-10-19T11:04:54.3573878Z  Getting string...
2020-10-19T11:04:54.3574092Z  Values match. value1 == value1
2020-10-19T11:04:54.3574268Z  
2020-10-19T11:04:54.3574471Z  Running test with RedisStrongNameClient
2020-10-19T11:04:54.3574700Z  Running sync operations
2020-10-19T11:04:54.3574891Z  Setting string...
2020-10-19T11:04:54.3575058Z  
2020-10-19T11:04:54.3575228Z  Profile attached: True
2020-10-19T11:04:54.3575662Z  Datadog.Trace.dll path: /project/reproductions/StackExchange.Redis.AssemblyConflict.SdkProject/bin/Debug/netcoreapp3.1/publish/Datadog.Trace.dll
2020-10-19T11:04:54.3576366Z  Datadog.Trace.ClrProfiler.Managed.dll path: /project/reproductions/StackExchange.Redis.AssemblyConflict.SdkProject/bin/Debug/netcoreapp3.1/publish/Datadog.Trace.ClrProfiler.Managed.dll
2020-10-19T11:04:54.3577051Z  Sigil.dll path: /project/reproductions/StackExchange.Redis.AssemblyConflict.SdkProject/bin/Debug/netcoreapp3.1/publish/Sigil.dll
2020-10-19T11:04:54.3577396Z  
[...]

PR #1402, opened 2021-04-14
Ducktype reverse proxy

tonyredondo:
This `PR` adds the support for reverse proxy in the DuckTyping mechanism. 

This means instead of defining an interface to create a proxy over an obscure class, in this scenario is the opposite, the interface is obscure and we create a class that implement it.

For example: This will help on scenarios like implementing a serilog sink (`ILogEventSink`) without a compile time dependency to serilog.

The PR also includes the `IgnoresAccessChecksToAttribute` that allows to create proxy types to non public types (Although this feature is not supported for .NET Framework 4.5), overpassing the limitations on #990 caused not by the partial trust problem but by adding the attribute to an AssemblyBuilder after we emit a type from the same assembly builder. 

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace/DuckTyping/DuckType.Methods.cs
+++ b/src/Datadog.Trace/DuckTyping/DuckType.Methods.cs
@@ -324,14 +324,14 @@ namespace Datadog.Trace.DuckTyping
                         {
                             if (NeedsDuckChaining(proxyParamType, targetParamType))
                             {
-                                // Load the argument and cast it as Duck type
+                                // Load the argument (our proxy type) and cast it as Duck type (the original type)
                                 il.WriteLoadArgument(idx, false);
                                 if (UseDirectAccessTo(proxyTypeBuilder, proxyParamType) && proxyParamType.IsValueType)
                                 {
                                     il.Emit(OpCodes.Box, proxyParamType);
                                 }
 
-                                // We call DuckType.CreateCache<>.Create()
+                                // We call DuckType.CreateCache<>.Create(object instance)
                                 MethodInfo getProxyMethodInfo = typeof(CreateCache<>)
                                     .MakeGenericType(targetParamType).GetMethod("Create");
 
--- a/src/Datadog.Trace/DuckTyping/DuckType.Methods.cs
+++ b/src/Datadog.Trace/DuckTyping/DuckType.Methods.cs
@@ -14,38 +14,42 @@ namespace Datadog.Trace.DuckTyping
         private static List<MethodInfo> GetMethods(Type baseType)
         {
             List<MethodInfo> selectedMethods = new List<MethodInfo>(GetBaseMethods(baseType));
-            Type[] implementedInterfaces = baseType.GetInterfaces();
-            foreach (Type imInterface in implementedInterfaces)
+            // If the base type is an interface we must make sure we implement all methods, including from other interfaces
+            if (baseType.IsInterface)
             {
-                if (imInterface == typeof(IDuckType))
+                Type[] implementedInterfaces = baseType.GetInterfaces();
+                foreach (Type imInterface in implementedInterfaces)
                 {
-                    continue;
-                }
-
-                foreach (MethodInfo interfaceMethod in imInterface.GetMethods())
-                {
-                    if (interfaceMethod.IsSpecialName)
+                    if (imInterface == typeof(IDuckType))
                     {
                         continue;
                     }
 
-                    string interfaceMethodName = interfaceMethod.ToString();
-                    bool methodAlreadySelected = false;
-                    foreach (MethodInfo currentMethod in selectedMethods)
+                    foreach (MethodInfo interfaceMethod in imInterface.GetMethods())
                     {
-                        if (currentMethod.ToString() == interfaceMethodName)
+                        if (interfaceMethod.IsSpecialName)
                         {
-                            methodAlreadySelected = true;
-                            break;
+                            continue;
                         }
-                    }
 
-                    if (!methodAlreadySelected)
-                    {
-                        MethodInfo prevMethod = baseType.GetMethod(interfaceMethod.Name, DuckAttribute.DefaultFlags, null, interfaceMethod.GetParameters().Select(p => p.ParameterType).ToArray(), null);
-                        if (prevMethod == null || prevMethod.GetCustomAttribute<DuckIgnoreAttribute>() is null)
+                        string interfaceMethodName = interfaceMethod.ToString();
+                        bool methodAlreadySelected = false;
+                        foreach (MethodInfo currentMethod in selectedMethods)
                         {
-                            selectedMethods.Add(interfaceMethod);
+                            if (currentMethod.ToString() == interfaceMethodName)
+                            {
+                                methodAlreadySelected = true;
+                                break;
+                            }
+                        }
+
+                        if (!methodAlreadySelected)
+                        {
+                            MethodInfo prevMethod = baseType.GetMethod(interfaceMethod.Name, DuckAttribute.DefaultFlags, null, interfaceMethod.GetParameters().Select(p => p.ParameterType).ToArray(), null);
+                            if (prevMethod == null || prevMethod.GetCustomAttribute<DuckIgnoreAttribute>() is null)
+                            {
+                                selectedMethods.Add(interfaceMethod);
+                            }
                         }
                     }
                 }
@@ -101,6 +105,9 @@ namespace Datadog.Trace.DuckTyping
                 {
                     DuckTypeTargetMethodNotFoundException.Throw(proxyMethodDefinition);
                 }
+
+                // Check if target method is a reverse method
+                bool isReverse = targetMethod.GetCustomAttribute<DuckReverseMethodAttribute>(true) is not null;
 
                 // Gets the proxy method definition generic arguments
                 Type[] proxyMethodDefinitionGenericArguments = proxyMethodDefinition.GetGenericArguments();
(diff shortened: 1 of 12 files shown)


test/Datadog.Trace.DuckTyping.Tests/ReverseProxyTests.cs
```diff
@@ -0,0 +1,162 @@
+#if !NET452
```
zacharycmontoya, line 1: Is there a test case we can run on `net452` just to make sure the code is exercised?

tonyredondo: well no in non public methods, but yeah I can create a test case with only public types.

src/Datadog.Trace/DuckTyping/DuckType.Utilities.cs
```diff
@@ -50,14 +76,46 @@ private static bool NeedsDuckChaining(Type targetType, Type proxyType)
...
+
+        /// <summary>
+        /// Gets if the direct access method should be used or the inderect method (dynamic method)
+        /// </summary>
+        /// <param name="builder">Type builder</param>
+        /// <param name="targetType">Target type</param>
+        /// <returns>true for direct method; otherwise, false.</returns>
+        private static bool UseDirectAccessTo(TypeBuilder builder, Type targetType)
+        {
+            return UseDirectAccessTo(builder is null ? null : (ModuleBuilder)builder.Module, targetType);
```
zacharycmontoya, line 108:
```suggestion
            return UseDirectAccessTo((ModuleBuilder)builder?.Module, targetType);
```

test/Datadog.Trace.DuckTyping.Tests/ReverseProxyTests.cs
zacharycmontoya: Could we also assert known failure modes for `net452` as well?

tonyredondo: done in 0e8e287

src/Datadog.Trace/DuckTyping/DuckType.Utilities.cs
tonyredondo: done in 0e8e287

test/Datadog.Trace.DuckTyping.Tests/ReverseProxyTests.cs
zacharycmontoya: Thanks!

src/Datadog.Trace/DuckTyping/DuckType.Methods.cs
```diff
@@ -504,6 +539,32 @@ private static MethodInfo SelectTargetMethod(Type targetType, MethodInfo proxyMe
                     continue;
                 }
 
+                // Check if the candidate method is a reverse mapped method
+                DuckReverseMethodAttribute reverseMethodAttribute = candidateMethod.GetCustomAttribute<DuckReverseMethodAttribute>(true);
+                if (reverseMethodAttribute?.Arguments is not null)
```
zacharycmontoya, line 544: `reverseMethodAttribute?.Arguments is not null` will also evaluate to `null` for a reverse proxy method that takes no arguments. Is that intentional? Would we ever have a need to define a reverse proxy method like that?

tonyredondo: Yeah, the `Arguments` attribute intended use is for 2 cases:
1) Multiple overloads with same number of attributes with arguments types that we don't have at compile time (ex: custom type).
2) To enable ducktyping over the arguments.

So is optional, if you don't define the arguments the normal `SelectTargetMethod` algorithm will try to select the best match for the one you are trying to ducktype.

tonyredondo: For the second case here you have an example: https://github.com/DataDog/dd-trace-dotnet/pull/1402/files#diff-fb46ebabc7e0a000432c1d76a6dc06a298ffeef5795e6ff2edbae76c25c11ff8R54 . ILogEvent will be the ducktype interface for `Datadog.Trace.Vendors.Serilog.Events.LogEvent`

src/Datadog.Trace/DuckTyping/DuckType.Methods.cs
```diff
@@ -308,7 +315,35 @@ private static void CreateMethods(TypeBuilder proxyTypeBuilder, Type proxyType,
...
+                            if (NeedsDuckChaining(proxyParamType, targetParamType))
+                            {
+                                // Load the argument and cast it as Duck type
+                                il.WriteLoadArgument(idx, false);
+                                if (UseDirectAccessTo(proxyTypeBuilder, proxyParamType) && proxyParamType.IsValueType)
+                                {
+                                    il.Emit(OpCodes.Box, proxyParamType);
+                                }
+
+                                // We call DuckType.CreateCache<>.Create()
```
zacharycmontoya, line 334: Really small nit, but I think it may help other reviewers:

```suggestion
                                // We call DuckType.CreateCache<>.Create(object instance)
```

src/Datadog.Trace/DuckTyping/DuckType.Methods.cs
```diff
@@ -308,7 +315,35 @@ private static void CreateMethods(TypeBuilder proxyTypeBuilder, Type proxyType,
...
+                            targetParamType = UseDirectAccessTo(proxyTypeBuilder, targetParamType) ? targetParamType : typeof(object);
+                            il.WriteSafeTypeConversion(proxyParamType, targetParamType);
+
+                            targetMethodParametersTypes[idx] = targetParamType;
+                        }
+                        else
+                        {
+                            if (NeedsDuckChaining(proxyParamType, targetParamType))
+                            {
+                                // Load the argument and cast it as Duck type
```
zacharycmontoya, line 327:
```suggestion
                                // Load the argument (our proxy type) and cast it as Duck type (the original type)
```

src/Datadog.Trace/DuckTyping/DuckType.Methods.cs
```diff
@@ -308,7 +315,35 @@ private static void CreateMethods(TypeBuilder proxyTypeBuilder, Type proxyType,
...
                             // If the target parameter type is public or if it's by ref we have to actually use the original target type.
-                            targetParamType = UseDirectAccessTo(targetParamType) ? targetParamType : typeof(object);
+                            targetParamType = UseDirectAccessTo(proxyTypeBuilder, targetParamType) ? targetParamType : typeof(object);
+                            il.WriteSafeTypeConversion(proxyParamType, targetParamType);
+
+                            targetMethodParametersTypes[idx] = targetParamType;
+                        }
+                        else
+                        {
+                            if (NeedsDuckChaining(proxyParamType, targetParamType))
```
zacharycmontoya, line 325: Now that we're creating Duck types both ways, I think it would help readability if we switched the order of the arguments so that the first argument is the "From" type and the second argument is the "To" type and if we also used the parameter names in this invocation specifically to make clear that this is going the opposite way. I think this feedback could be applied in a separate refactoring PR so it doesn't confuse reviewers on the real changes needed for reverse proxying

zacharycmontoya (approved):
Looks great!

src/Datadog.Trace/DuckTyping/DuckType.Methods.cs
tonyredondo: Totally agree. I'll do that in another PR once this one gets merged.

test/Datadog.Trace.DuckTyping.Tests/ReverseProxyTests.cs
```diff
@@ -0,0 +1,195 @@
...
+
+#pragma warning disable SA1201 // Elements should appear in the correct order
+#pragma warning disable SA1602 // Enumeration items should be documented
+
+namespace Datadog.Trace.DuckTyping.Tests
+{
+    public class ReverseProxyTests
+    {
+        [Fact]
+        public void InterfaceReverseProxyTest()
```
kevingosse, line 14:
```suggestion
        public void PrivateInterfaceReverseProxyTest()
```

test/Datadog.Trace.DuckTyping.Tests/ReverseProxyTests.cs
```diff
@@ -0,0 +1,195 @@
...
+        {
+            public string Name { get; }
+
+            public object Value { get; }
+        }
+
+        // ************************************************************************************
+
+        [Fact]
+        public void AbstractClassReverseProxyTest()
```
kevingosse, line 110:
```suggestion
        public void PrivateAbstractClassReverseProxyTest()
```

test/Datadog.Trace.DuckTyping.Tests/ReverseProxyTests.cs
```diff
@@ -0,0 +1,195 @@
...
+                public string ToString();
+
+                public string ToString(string format, IFormatProvider formatProvider);
+            }
+        }
+
+        // ************************************************************************************
+
+        [Fact]
+        public void PublicInterfaceReverProxyTest()
```
kevingosse, line 173:
```suggestion
        public void PublicInterfaceReverseProxyTest()
```

test/Datadog.Trace.DuckTyping.Tests/ReverseProxyTests.cs
```diff
@@ -0,0 +1,195 @@
...
+        {
+            var instance = new PublicBaseInterface();
```
kevingosse, line 175: Could you do the same thing as the private interface test, but with the "real" serilog instead of the vendored version? This way we'll be completely sure that it works in the intended scenario

kevingosse (changes requested):
Great work, just a few improvements for the tests

test/Datadog.Trace.DuckTyping.Tests/ReverseProxyTests.cs
tonyredondo: done in 2e0237f

kevingosse (approved):

PR #1409, opened 2021-04-16
Allow CallTarget instrumentation of nested classes

andrewlock:
Allows specifying the nested type to instrument using the pattern "Fully.Qualified.Parent+Nested"
Only supports one level of nesting

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace.ClrProfiler.Native/clr_helpers.cpp
+++ b/src/Datadog.Trace.ClrProfiler.Native/clr_helpers.cpp
@@ -1586,7 +1586,7 @@ bool FindTypeDefByName(
     }
     instrumentedMethodTypeName = nameParts[1];
 
-  } else if (nameParts.size() > 1) {
+  } else if (nameParts.size() > 2) {
     Warn("Invalid TypeDef-only one layer of nested classes are supported: ",
          instrumentationTargetMethodTypeName, ", Module: ", assemblyName);
     return false;
--- a/src/Datadog.Trace.ClrProfiler.Native/clr_helpers.cpp
+++ b/src/Datadog.Trace.ClrProfiler.Native/clr_helpers.cpp
@@ -1603,5 +1603,7 @@ bool FindTypeDefByName(
           ", Module: ", assemblyName);
     return false;
   }
+
+  return true;
 }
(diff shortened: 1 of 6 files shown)


src/Datadog.Trace.ClrProfiler.Native/clr_helpers.cpp
```diff
@@ -1602,4 +1602,46 @@ HRESULT FunctionMethodSignature::TryParse() {
...
+      // available in both. Eg: WinHttpHandler, CurlHandler, and some methods in
+      // System.Data
+      Debug("Can't load the parent TypeDef: ", nameParts[0],
+            " for nested class: ", instrumentationTargetMethodTypeName,
+            ", Module: ", assemblyName);
+      return false;
+    }
+    instrumentedMethodTypeName = nameParts[1];
+
+  } else if (nameParts.size() > 1) {
```
tonyredondo, line 1629: I know this works, but I was why >1? is >2.

tonyredondo (approved):
LGTM

src/Datadog.Trace.ClrProfiler.Native/clr_helpers.cpp
lucaspimentel: I agree. The case where `== 2` is already handled above so this is confusing.
```suggestion
  } else if (nameParts.size() > 2) {
```

tonyredondo (approved):

src/Datadog.Trace.ClrProfiler.Native/clr_helpers.cpp
andrewlock: For the back story, it's because I incorrectly had `==1` as the first case initially and forgot to change this one when I fixed it 😬 

I'll leave my badge and gun/keyboard at reception.

lucaspimentel (approved):

src/Datadog.Trace.ClrProfiler.Native/clr_helpers.cpp
tonyredondo: 😂

zacharycmontoya (approved):

PR #1413, opened 2021-04-19
Fix git parser on really big pack files (>2GB)

tonyredondo:
This PR fixes an issue detected in very large pack files where the git metadata couldn't be loaded.

The problem was before we were loading the object size using only 2 bytes, the PR changes to multiple bytes.

The PR has been tested loading data from the linux kernel git repo.

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace.Ci.Shared/GitInfo.cs
+++ b/src/Datadog.Trace.Ci.Shared/GitInfo.cs
@@ -439,57 +439,61 @@ namespace Datadog.Trace.Ci
             public static bool TryGetFromPackageOffset(GitPackageOffset packageOffset, out GitCommitObject commitObject)
             {
                 commitObject = default;
-
-                string packFile = Path.ChangeExtension(packageOffset.FilePath, ".pack");
-                if (File.Exists(packFile))
+                try
                 {
-                    // packfile format explanation:
-                    // https://codewords.recurse.com/issues/three/unpacking-git-packfiles#:~:text=idx%20file%20contains%20the%20index,pack%20file.&text=Objects%20in%20a%20packfile%20can,of%20storing%20the%20whole%20object.
-
-                    using (var fs = new FileStream(packFile, FileMode.Open, FileAccess.Read, FileShare.Read))
-                    using (var br = new BigEndianBinaryReader(fs))
+                    string packFile = Path.ChangeExtension(packageOffset.FilePath, ".pack");
+                    if (File.Exists(packFile))
                     {
-                        // Move to the offset of the object
-                        fs.Seek(packageOffset.Offset, SeekOrigin.Begin);
-                        byte[] packData = br.ReadBytes(2);
+                        // packfile format explanation:
+                        // https://codewords.recurse.com/issues/three/unpacking-git-packfiles#:~:text=idx%20file%20contains%20the%20index,pack%20file.&text=Objects%20in%20a%20packfile%20can,of%20storing%20the%20whole%20object.
 
-                        // We build the size combining the bits from sizeParts
-                        // using bitwise operations (BigEndian).
-                        // Example:
-                        // - sizeParts = [-100, 53] => [10011100, 00110101]
-                        // - size = 00000000 00000000 00000000 00000000
-                        int size = 0;
-
-                        // Clean first bit and add bits to size using OR.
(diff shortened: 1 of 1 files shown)


src/Datadog.Trace.Ci.Shared/GitInfo.cs
```diff
@@ -439,44 +439,62 @@ public static bool TryGetFromObjectFile(string filePath, out GitCommitObject com
...
+                                    defStream.Close();
+                                    string strContent = Encoding.UTF8.GetString(buffer, 0, readBytes);
+                                    commitObject = new GitCommitObject(strContent);
+                                    return true;
+                                }
+                            }
                         }
                     }
                 }
+                catch
```
lucaspimentel, line 495: Are we ok throwing away the exception here? No logging or anything?

tonyredondo: Yeah, I didn't want to add a dependency to Datadog.Trace assembly for the logger, but I think I will.
The method is a `Try...` so I should keep the try/catch.

src/Datadog.Trace.Ci.Shared/GitInfo.cs
```diff
@@ -439,44 +439,62 @@ public static bool TryGetFromObjectFile(string filePath, out GitCommitObject com
...
-                            byte[] buffer = new byte[8192];
-                            int readBytes = defStream.Read(buffer, 0, buffer.Length);
-                            defStream.Close();
-                            string strContent = Encoding.UTF8.GetString(buffer, 0, readBytes);
-                            commitObject = new GitCommitObject(strContent);
-                            return true;
+                            // Extract the object size
+                            int objectSize = (int)(packData[0] & 0x0F);
+
+                            if (packData[0] > 128)
```
zacharycmontoya, line 460: In the previous implementation, this block was executed in the else clause for `packData[0] < 128`, which basically meant `packData[0] >= 128`. Is the new implementation where it's `packData[0] > 128` intended?

tonyredondo: Oh you're right, good catch!
The comparison is because in the specification from 8 bits, they use 7 for the value and 1 to indicate you should read another byte to continue loading the value. And yes: `128 = 0b1000_0000` so the value is 0, but we should read the next byte.

tonyredondo: done in afe9ff7

src/Datadog.Trace.Ci.Shared/GitInfo.cs
tonyredondo: done in afe9ff7

zacharycmontoya (approved):
I'm no expert on the parsing but LGTM

lucaspimentel (approved):

src/Datadog.Trace.Ci.Shared/GitInfo.cs
```diff
@@ -413,70 +417,107 @@ private GitCommitObject(string content)
...
+                            else
+                            {
+                                Log.Warning<int>("The object size is outside of an acceptable range: {objectSize}", objectSize);
+                            }
                         }
                     }
                 }
+                catch (Exception ex)
+                {
+                    Log.Error(ex, ex.Message);
```
andrewlock, line 519: nit: error messages
```suggestion
                    Log.Error(ex, "Error loading commit information from package offset");
```

src/Datadog.Trace.Ci.Shared/GitInfo.cs
```diff
@@ -504,67 +545,87 @@ public static bool TryGetPackageOffset(string idxFilePath, string commitSha, out
...
+                            }
+
+                            packageOffset = new GitPackageOffset(idxFilePath, (long)packOffset);
+                            return true;
+                        }
                     }
                 }
+                catch (Exception ex)
+                {
+                    Log.Error(ex, ex.Message);
```
andrewlock, line 627: nit: error messages
```suggestion
                    Log.Error(ex, "Error getting package offset");
```

andrewlock (approved):
I'll take your word the parsing is correct 😄 Just minor nits around error messages

PR #1414, opened 2021-04-19
Re-enable AspNet integration tests in CI

zacharycmontoya:
### Summary
Runs the AspNet tests added by https://github.com/DataDog/dd-trace-dotnet/pull/1288 . Currently, this only tests the `AspNetMvc` and `AspNetWebApi2` integrations because the `AspNet` integration is not yet enabled via automatic instrumentation. That will be completed in https://github.com/DataDog/dd-trace-dotnet/pull/1280

The tests are run in the Windows IIS runs for the following reasons:
- Restoring the projects and publishing can be grouped into the existing `samples-iis.sln` solution
- Instrumenting IIS and IIS Express requires the Datadog.Trace assemblies to be in the GAC

### Changes
- Adding the AspNet CI applications to `samples-iis.sln`, which we are already properly restoring and publishing in the Windows IIS jobs
- Removing some compile-time references to `Datadog.Trace.dll` and `Datadog.Trace.ClrProfiler.Managed.dll` with lookups via reflection. This is used so the site can give us a clear indicator whether automatic instrumentation is running
- Miscellaneous change: Stop Visual Studio from auto-generating a `launchSetings.json` for the Datadog.Trace.IntegrationTests project

@DataDog/apm-dotnet

### DIFF
--- a/test/Datadog.Trace.IntegrationTests/Datadog.Trace.IntegrationTests.csproj
+++ b/test/Datadog.Trace.IntegrationTests/Datadog.Trace.IntegrationTests.csproj
@@ -2,6 +2,9 @@
 
   <PropertyGroup>
     <OutputType>Library</OutputType>
+
+    <!-- Tell Visual Studio to not create a new launchSettings.json file, even though we have AspNetCore assets -->
+    <NoDefaultLaunchSettingsFile>true</NoDefaultLaunchSettingsFile>
   </PropertyGroup>
   
   <ItemGroup>
--- a/test/test-applications/aspnet/Samples.AspNetMvc5/Samples.AspNetMvc5.csproj
+++ b/test/test-applications/aspnet/Samples.AspNetMvc5/Samples.AspNetMvc5.csproj
@@ -143,17 +143,13 @@
     <Folder Include="Models\" />
   </ItemGroup>
   <ItemGroup>
+    <ProjectReference Include="..\..\..\..\src\Datadog.Trace\Datadog.Trace.csproj">
+      <Project>{8e1baa6a-47cc-47f0-a7d6-74741118eb7c}</Project>
+      <Name>Datadog.Trace</Name>
+    </ProjectReference>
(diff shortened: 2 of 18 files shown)


test/test-applications/aspnet/Samples.AspNetMvc5/Samples.AspNetMvc5.csproj
```diff
@@ -143,18 +143,14 @@
     <Folder Include="Models\" />
   </ItemGroup>
   <ItemGroup>
+    <ProjectReference Include="..\..\..\..\src\Datadog.Trace\Datadog.Trace.csproj">
```
lucaspimentel, line 146: Do we still need this reference? You were able to remove it in `Samples.AspNetMvc4`.

lucaspimentel (approved):

test/Datadog.Trace.IntegrationTests/Datadog.Trace.IntegrationTests.csproj
```diff
@@ -2,6 +2,9 @@
 
   <PropertyGroup>
     <OutputType>Library</OutputType>
+
+    <!-- Tell Visual Studio to not create a new launchSettings.json file, even though we have AspNetCore assets -->
+    <NoDefaultLaunchSettingsFile>true</NoDefaultLaunchSettingsFile>
```
andrewlock, line 7: 🎉 TIL

andrewlock (approved):

test/test-applications/aspnet/Samples.AspNetMvc5/Samples.AspNetMvc5.csproj
zacharycmontoya: Yes we still need it here because we have a `DistributedTracingApiController` that tests distributed tracing and needs to access `Span` properties at compile-time. In the future we can make this reflection-based if we need to trim it out, but for now it's okay

lucaspimentel: Ok, thanks!

PR #996, opened 2020-10-22
Fix tags serialization

kevingosse:
Rewriting the map header was too optimistic: I didn't consider that MessagePack maps actually had variable-size headers.

Instead, I rewrote the logic to get the exact number of tags/metrics before writing the header. It means iterating once more on the list of additional tags/metrics, but I don't expect the overhead to be significant.

PR #1416, opened 2021-04-20
Serialize tags/metrics in a single pass

kevingosse:
Tags and metrics were initially serialized in a single pass, and the count was back-patched at the end. This caused issues because the MessagePack map header has a variable size, so in some cases I ended up not allocating enough space. This was fixed by switching to a two-pass algorithm, with a performance hit: https://github.com/DataDog/dd-trace-dotnet/pull/996

I just realized that there's actually an API to write a fixed-size header, so I'm switching back to a single-pass algorithm. The existing unit test `TagsListTests.Serialization` verifies that it does not introduce any regression (I confirmed beforehand that the test still fails if I use `MessagePackBinary.WriteMapHeader` instead of `MessagePackBinary.WriteMapHeaderForceMap32Block`).

### DIFF
--- a/src/Datadog.Trace/Tagging/TagsList.cs
+++ b/src/Datadog.Trace/Tagging/TagsList.cs
@@ -238,6 +238,13 @@ namespace Datadog.Trace.Tagging
         protected virtual IProperty<double?>[] GetAdditionalMetrics() => ArrayHelper.Empty<IProperty<double?>>();
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static void WriteTag(ref byte[] bytes, ref int offset, string key, string value)
+        {
+            offset += MessagePackBinary.WriteString(ref bytes, offset, key);
+            offset += MessagePackBinary.WriteString(ref bytes, offset, value);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private static void WriteMetric(ref byte[] bytes, ref int offset, string key, double value)
         {
             offset += MessagePackBinary.WriteString(ref bytes, offset, key);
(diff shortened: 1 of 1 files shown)


src/Datadog.Trace/Tagging/TagsList.cs
```diff
@@ -252,48 +259,42 @@ private int WriteTags(ref byte[] bytes, int offset)
...
                 {
                     count += tags.Count;
 
-                    offset += MessagePackBinary.WriteMapHeader(ref bytes, offset, count);
-
                     foreach (var pair in tags)
                     {
-                        offset += MessagePackBinary.WriteString(ref bytes, offset, pair.Key);
-                        offset += MessagePackBinary.WriteString(ref bytes, offset, pair.Value);
+                        WriteTag(ref bytes, ref offset, pair.Key, pair.Value);
```
lucaspimentel, line 276: Shouldn't we check these tag values for null like we do for the additional tags below?
We never want to send null tag values to the Agent because it breaks deserialization. (Maybe this is fixed in newer versions, but we should avoid it if possible.)

Same comment for metrics below.

andrewlock: I think the `SetTag` implementation ensures it's not null as far as I can see? 🤔

andrewlock (approved):

src/Datadog.Trace/Tagging/TagsList.cs
lucaspimentel: Yes, but these well-known tags are set via properties and not stored in a dictionary, like the "additional tags" set using `SetTag`. For example you can set `WebTags.HttpMethod = null`
https://github.com/DataDog/dd-trace-dotnet/blob/5bfda09d1670c84846371fba35eb4d546949fc54/src/Datadog.Trace/Tagging/WebTags.cs#L17

lucaspimentel: You're right. I misread `SetTag()`. For some reason I thought `GetAdditionalTags()` returned the tags from the list, not the properties. I had them reversed.

lucaspimentel (approved):

src/Datadog.Trace/Tagging/TagsList.cs
kevingosse: As mentioned by Andrew, we null-check the tags bound to properties because we can't prevent them from being null, but for the others we catch that in `SetTag`

zacharycmontoya (approved):

PR #1418, opened 2021-04-21
Enable endpoint routing in aspnetcore benchmark

kevingosse:
The impact of changes introduced in https://github.com/DataDog/dd-trace-dotnet/pull/1289 can only be measured with endpoint routing enabled.

It also means upgrading the benchmark from aspnetcore 2.x to 3.x, but it's probably more representative of what customers use.

### DIFF
--- a/test/benchmarks/Benchmarks.Trace/AspNetCoreBenchmark.cs
+++ b/test/benchmarks/Benchmarks.Trace/AspNetCoreBenchmark.cs
@@ -69,7 +69,13 @@ namespace Benchmarks.Trace
 
             public void Configure(IApplicationBuilder builder)
             {
-                builder.UseMvcWithDefaultRoute();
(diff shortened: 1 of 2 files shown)


andrewlock (commented):
Is it worth benchmarking both with and without endpoint routing? You could create a separate `Startup` class for that and continue to use `UseMvc()`?

kevingosse:
> Is it worth benchmarking both with and without endpoint routing? You could create a separate `Startup` class for that and continue to use `UseMvc()`?

I considered it but I don't think we have any code path that is triggered only when endpoint routing is disabled. And the benchmark pipeline is already taking a while so I'm cautious about adding new benchmarks 😃

andrewlock (approved):

lucaspimentel (approved):

PR #1420, opened 2021-04-22
Optimize aspnetcore when enabling route template resource names

kevingosse:
- Remove boxing of headers
- Don't register the aspnetcore feature if RouteTemplateResourceNamesEnabled is false
- Use specialized tags object to reduce allocations when the feature is disabled

### DIFF
--- a/src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
+++ b/src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
@@ -71,6 +71,12 @@ namespace Datadog.Trace.ExtensionMethods
             }
         }
 
+        // Added to avoid a breaking change
+        internal static void SetHeaderTags(this Span span, IHeadersCollection headers, IDictionary<string, string> headerTags, string defaultTagPrefix)
+        {
+            SetHeaderTags<IHeadersCollection>(span, headers, headerTags, defaultTagPrefix);
+        }
+
         internal static void SetHeaderTags<T>(this Span span, T headers, IDictionary<string, string> headerTags, string defaultTagPrefix)
             where T : IHeadersCollection
         {
--- a/src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
+++ b/src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
@@ -71,7 +71,8 @@ namespace Datadog.Trace.ExtensionMethods
             }
         }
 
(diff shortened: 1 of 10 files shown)


andrewlock (approved):

src/Datadog.Trace/ExtensionMethods/SpanExtensions.cs
```diff
@@ -71,7 +71,8 @@ public static void AddTagsFromDbCommand(this Span span, IDbCommand command)
             }
         }
 
-        internal static void SetHeaderTags(this Span span, IHeadersCollection headers, IDictionary<string, string> headerTags, string defaultTagPrefix)
+        internal static void SetHeaderTags<T>(this Span span, T headers, IDictionary<string, string> headerTags, string defaultTagPrefix)
```
zacharycmontoya, line 74: This might be a breaking change because there are callsites in Datadog.Trace.ClrProfiler.Managed that call this and expect the second argument to have the type `IHeadersCollection`. Without testing this, I fear it may require changing the name so the original call from earlier versions of Datadog.Trace.ClrProfiler.Managed while find the method

kevingosse: I wonder if I can add it as an overload, and let overload resolution do its magic. Gonna give it a try

andrewlock: Good catch. If it does work, whack the old `[Obsolete]` on the original 🙂 ... ready for that version 2.0 cleanup 😉

kevingosse: Done in https://github.com/DataDog/dd-trace-dotnet/pull/1420/commits/411c8f49cbf0367c9deff5a90710bbffbda92012

zacharycmontoya: FYI, I tried to create a MissingMethodException with this branch at the d235f81281926dddae568d5e6b1e19df73566608 commit, but I wasn't able to because its only used on .NET Core is inside the same assembly with the AspNetCoreDiagnosticObserver. If the new overload isn't a burden, then let's keep it just to be safe 👍🏼

kevingosse: @zacharycmontoya  it required changing a few calls to make sure it's directed to the new method, but I don't think it's a burden 😃

zacharycmontoya (approved):

PR #1425, opened 2021-04-23
DuckType extensions optimization.

tonyredondo:
Optimizations for `TryDuckCast` and `DuckAs` extension methods.

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace/DuckTyping/DuckTypeExtensions.cs
+++ b/src/Datadog.Trace/DuckTyping/DuckTypeExtensions.cs
@@ -73,7 +73,7 @@ namespace Datadog.Trace.DuckTyping
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static bool TryDuckCast(this object instance, Type targetType, out object value)
         {
-            if (instance is not null)
+            if (instance is not null && targetType is not null)
             {
 #if NET45
                 if (!targetType.IsVisible)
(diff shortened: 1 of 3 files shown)


src/Datadog.Trace/DuckTyping/DuckTypeExtensions.cs
```diff
@@ -41,10 +41,14 @@ public static object DuckCast(this object instance, Type targetType)
...
+            if (instance is not null)
             {
-                value = DuckType.Create<T>(instance);
-                return true;
+                DuckType.CreateTypeResult proxyResult = DuckType.CreateCache<T>.GetProxy(instance.GetType());
+                if (proxyResult.Success)
+                {
+                    value = proxyResult.CreateInstance<T>(instance);
+                    return true;
+                }
```
andrewlock: I understand it may be for optimisation reasons, but could we not reduce the duplication in these extensions? for example:

```csharp
public static bool TryDuckCast<T>(this object instance, out T value)
    => instance.TryDuckCast(typeof(T), value);

public static bool DuckAs<T>(this object instance)
    => instance.TryDuckCast<T>(out var value) ? value : null;

public static bool DuckAs<T>(this object instance, Type targetType)
    => instance.TryDuckCast(targetType, out var value) ? value : null;
```

andrewlock (approved):

src/Datadog.Trace/DuckTyping/DuckTypeExtensions.cs
tonyredondo: Yeah If I do that, we remove the `CreateCache<T>` optimization, so it will perform a bit worse. 

static field vs dictionary lookup.

andrewlock: oh yeah, I missed that.  Ignore me 🙂

lucaspimentel (approved):

zacharycmontoya (approved):

PR #1426, opened 2021-04-23
Adds native callback stats.

tonyredondo:
This PR adds lightweight native callback stats to the process exit message.

#### Example:

```
...
04/23/21 04:29:16.018 PM [10736|13008] [info] Exiting ReJIT request thread.
04/23/21 04:29:16.018 PM [10736|14028] [warning] Exiting. Stats: [Initialize=30ms, ModuleLoadFinished=607ms/343, CallTargetRequestRejit=1ms/14, CallTargetRewriter=2ms/4, AssemblyLoadFinished=21ms/343, ModuleUnloadStarted=0ms/0, JitCompilationStarted=523ms/14997, JitInlining=3ms/18398]
```

```
...
04/23/21 04:29:16.539 PM [3860|14456] [info] Exiting ReJIT request thread.
04/23/21 04:29:16.539 PM [3860|15184] [warning] Exiting. Stats: [Initialize=26ms, ModuleLoadFinished=153ms/86, CallTargetRequestRejit=1ms/5, CallTargetRewriter=1ms/2, AssemblyLoadFinished=10ms/86, ModuleUnloadStarted=0ms/0, JitCompilationStarted=711ms/11193, JitInlining=2ms/21933]
```

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace.ClrProfiler.Native/stats.h
+++ b/src/Datadog.Trace.ClrProfiler.Native/stats.h
@@ -8,22 +8,17 @@
 namespace trace {
 
 class SWStat {
-  std::chrono::nanoseconds* _value;
-  std::mutex* _mutex;
+  std::atomic_ullong* _value;
   std::chrono::steady_clock::time_point _startTime;
 
  public:
-  SWStat(std::chrono::nanoseconds* value, std::mutex* mutex) {
+  SWStat(std::atomic_ullong* value) {
     _value = value;
-    _mutex = mutex;
     _startTime = std::chrono::steady_clock::now();
   }
   ~SWStat() {
-    auto increment = std::chrono::steady_clock::now() - _startTime;
-    {
-      std::lock_guard<std::mutex> guard(*_mutex);
-      *_value += increment;
-    }
+    auto increment = (std::chrono::steady_clock::now() - _startTime).count();
+    _value->fetch_add(increment);
   }
 };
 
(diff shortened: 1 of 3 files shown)


src/Datadog.Trace.ClrProfiler.Native/stats.h
```diff
@@ -0,0 +1,118 @@
...
+    callTargetRequestRejitCount = 0;
+    jitInliningCount = 0;
+    jitCompilationStartedCount = 0;
+    moduleUnloadStartedCount = 0;
+    moduleLoadFinishedCount = 0;
+    assemblyLoadFinishedCount = 0;
+  }
+  SWStat CallTargetRequestRejitMeasure() {
+    callTargetRequestRejitCount++;
+    return SWStat(&callTargetRequestRejit);
```
andrewlock, line 64: Something I realised I've never checked - is the native profiler single threaded? I assume it must be (or we enforce that through queueing etc), but if not, these counters aren't thread safe right?

andrewlock (approved):

src/Datadog.Trace.ClrProfiler.Native/stats.h
tonyredondo: Is not single threaded, callbacks can run from multiple threads, but we enforce syncronization with a lock guards in the callbacks, ex:
https://github.com/DataDog/dd-trace-dotnet/blob/32973349544145eb43dddefb6976ea5d0d4b206b/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp#L293-L295

But you're right the guard is after creating this SWStat instance, I will add synchronisation here too.

src/Datadog.Trace.ClrProfiler.Native/stats.h
```diff
@@ -1,22 +1,29 @@
...
+    _mutex = mutex;
+    _startTime = std::chrono::steady_clock::now();
   }
   ~SWStat() {
-    *_value += std::chrono::steady_clock::now() - _startTime;
+    auto increment = std::chrono::steady_clock::now() - _startTime;
+    {
+      std::lock_guard<std::mutex> guard(*_mutex);
+      *_value += increment;
+    }
```
andrewlock, line 26: Would it be possible to replace all these mutexes with `std::atomic<T>::fetch_add`? That's in C++11 - can't remember what version we target?

tonyredondo: Sadly no, at least not directly. I tried but we can't use `std::chrono::time_point` as a template instance, due compilation error, I guess one way to do it is to convert `std::chrono::time_point` to `long long` every time.

tonyredondo: trying this approach in db0520d

zacharycmontoya (approved):

PR #1427, opened 2021-04-23
Prevent dependabot from opening PR's against the Microsoft.Build.Framework NuGet package

zacharycmontoya:
Fixes #1393

@DataDog/apm-dotnet

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -53,6 +53,9 @@ updates:
       - dependency-name: "System.Reflection.Emit"
       - dependency-name: "System.Reflection.Emit.Lightweight"
       ### End Datadog.Trace.csproj ignored dependencies
+
+      # Lock Microsoft.Build.Framework for widest compatibility when instrumenting builds
+      - dependency-name: "Microsoft.Build.Framework"
   - package-ecosystem: "nuget"
     directory: "/src/Datadog.Trace.OpenTracing"
     schedule:


.github/dependabot.yml
```diff
@@ -53,6 +53,9 @@ updates:
       - dependency-name: "System.Reflection.Emit"
       - dependency-name: "System.Reflection.Emit.Lightweight"
       ### End Datadog.Trace.csproj ignored dependencies
+
+      # Lock Microsoft.Build.Framework for widest compatibility when instrumenting builds
+      - dependency-name: "Microsoft.Build.Framework"
```
andrewlock, line 58: nit: I think this should be _before_ the "ignored dependencies" comment?
```suggestion

      # Lock Microsoft.Build.Framework for widest compatibility when instrumenting builds
      - dependency-name: "Microsoft.Build.Framework"
      ### End Datadog.Trace.csproj ignored dependencies    
```

andrewlock (approved):

.github/dependabot.yml
zacharycmontoya: Oh, I had the all the Datadog.Trace.csproj dependencies grouped together since those needed to be copy/pasted to Datadog.Trace.ClrProfiler.Managed.csproj as well, but this one is unique to the latter project

andrewlock: Ah, I see, my bad!

lucaspimentel (approved):

PR #1429, opened 2021-04-23
Run Postgres tests in Windows integration-test pipelines

zacharycmontoya:
The `Samples.Npgsql` test application was not previously run because there's no postgres container that is a Windows container. After looking through the test code in the `npgsql/npgsql` repo, it turns out there are Windows downloads and it can be reliably installed and run on Windows.

This PR adds the setup code to start up postgres in the Windows VM (this could later be separated into its own Windows container) and runs the integration test.

@DataDog/apm-dotnet

### DIFF
--- a/.azure-pipelines/steps/install-windows-services.yml
+++ b/.azure-pipelines/steps/install-windows-services.yml
@@ -1,0 +1,77 @@
+steps:
+- bash: |
+    # Sourced from https://github.com/npgsql/npgsql/blob/v5.0.4/.github/workflows/build.yml
+    # Some configuration is tweaked for tests in the npgsql/npgsql repo
+
+    # Find EnterpriseDB version number
+    EDB_VERSION=$(pwsh -c "
+        \$global:progressPreference='silentlyContinue';
+        Invoke-WebRequest -URI https://www.postgresql.org/applications-v2.xml |
+            Select-Object -ExpandProperty Content |
+            Select-Xml -XPath '/applications/application[id=\"postgresql_13\" and platform=\"windows-x64\"]/version/text()' |
+            Select-Object -First 1 -ExpandProperty Node |
+            Select-Object -ExpandProperty Value")
+
+    # Install PostgreSQL
+    echo "Installing PostgreSQL (version: ${EDB_VERSION})"
+    curl -o pgsql.zip -L https://get.enterprisedb.com/postgresql/postgresql-${EDB_VERSION}-windows-x64-binaries.zip
+    unzip pgsql.zip -x 'pgsql/include/**' 'pgsql/doc/**' 'pgsql/pgAdmin 4/**' 'pgsql/StackBuilder/**'
+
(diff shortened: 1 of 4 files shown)


.azure-pipelines/steps/install-windows-services.yml
```diff
@@ -0,0 +1,77 @@
...
+      curl -Ls https://download.osgeo.org/postgis/windows/pg13 |
+      sed -n 's/.*>postgis-bundle-pg13-\(3.[0-9]*.[0-9]*\)x64.zip<.*/\1/p' |
+      tail -n 1)
+
+    # Install PostGIS
+    echo "Installing PostGIS (version: ${OSGEO_VERSION})"
+    POSTGIS_FILE="postgis-bundle-pg13-${OSGEO_VERSION}x64"
+    curl -o postgis.zip -L https://download.osgeo.org/postgis/windows/pg13/${POSTGIS_FILE}.zip
+    unzip postgis.zip -d postgis
+    cp -a postgis/$POSTGIS_FILE/. pgsql/
```
andrewlock, line 35: Do we actually need PostGIS? I'm assuming our tests don't use it?

If not, then don't need this download or the `CREATE EXTENSION postgis` below

andrewlock (approved):

.azure-pipelines/steps/install-windows-services.yml
zacharycmontoya: No our integration tests don't use it, I can remove it

andrewlock: Just had a separate thought - is there any way we can just run these services in Linux via wsl2 + docker-compose? I'm guessing not, but it would prob make things easier from a local testing point of view 🤔

zacharycmontoya: We could make these steps into a docker image, and then add that to docker-compose, if you would like. The only thing is this code will have to run in a Windows environment (container or scripts) so we can run in the Windows CI

zacharycmontoya: Or I could translate this to a powershell script?

zacharycmontoya: We can always run `docker-compose up -d postgres` and run the Linux container locally, so maybe we don't need to address this at the moment. What do you think?

andrewlock: > The only thing is this code will have to run in a Windows environment (container or scripts) so we can run in the Windows CI

This is what I was questioning really, could we run the _Linux_ containers on the Windows Hosted Agents, so that would have simplified things a lot. But [it's not possible](https://developercommunity.visualstudio.com/t/support-for-docker-with-linux-containers-on-window/366304), so never mind 🙂 

I don't think it's worth trying to convert this to Docker or powershell, as we'll only use it in CI, as you say

zacharycmontoya:
I'm going to close this PR but I'll leave the branch here in case we have a need for it.

PR #1431, opened 2021-04-26
Run the npgsql/npgsql tests in our pipelines with the profiler attached

zacharycmontoya:
- Adds a new Azure DevOps pipeline to run integration tests from libraries, starting with Npgsql
  - Runs Npgsql tests at tags v4.1.9 and v5.0.7, on both Ubuntu and Windows, with automatic instrumentation enabled
  - It re-uses the build steps from the consolidated pipeline to build the .NET Tracer artifacts
  - Scheduled to run daily at 3am UTC
  - Sets up Postgres with several extensions like PostGIS. These setup instructions were copied directly from the source repo

Results
- Several tests fail on the v5.0.7 tag. The root cause appears to be the runtime types of exception objects being different when Datadog automatic instrumentation is attached AND CallTarget is enabled
  - Npgsql.Tests.CommandTests.CancelAsyncHard
  - Npgsql.Tests.SecurityTests.Bug1718
  - Npgsql.Tests.CommandTests.CancelAsyncSoft
  - Npgsql.Tests.CommandTests.CancelAsyncImmediately

@DataDog/apm-dotnet

### DIFF
--- a/.azure-pipelines/third-party-test-suites.yml
+++ b/.azure-pipelines/third-party-test-suites.yml
@@ -9,174 +9,410 @@ schedules:
     - master
   always: true
 
-jobs:
-- job: Postgres
-  strategy:
-    matrix:
-      Ubuntu_Major5:
-        imageName: ubuntu-20.04
-        os: ubuntu
-        tagName: v5.0.4
-      Windows_Major5:
-        imageName: windows-2019
-        os: windows
-        tagName: v5.0.4
-        serverDirectory: npgsql/.build
-      Ubuntu_Major4:
-        imageName: ubuntu-20.04
-        os: ubuntu
-        tagName: v4.1.9
-      Windows_Major4:
-        imageName: windows-2019
-        os: windows
-        tagName: v4.1.9
-        serverDirectory: npgsql/.build
+stages:
+# The job definitions for this stage are copy/pasted from the runner.yml file
+- stage: build_dotnet_tool
   variables:
-    pg_major: 13
-    postgis_version: 3
-  pool:
-    vmImage: $(imageName)
-  steps:
+    publishOutput: $(System.DefaultWorkingDirectory)/src/bin/managed-publish
+    dotnetCoreSdk5Version: 5.0.103
+  jobs:
 
-  - checkout: none
+  - job: build_linux_profiler
 
-  - script: |
-      git clone https://github.com/npgsql/npgsql.git
-      cd $(System.DefaultWorkingDirectory)/npgsql
-      git checkout $(tagName)
-    displayName: Checkout npgsql at tag $(tagName)
+    steps:
+    - task: UseDotNet@2
+      displayName: install dotnet core sdk 5.0
+      inputs:
+        packageType: sdk
+        version: $(dotnetCoreSdk5Version)
+        includePreviewVersions: true
 
-  - bash: |
-      # Sourced from https://github.com/npgsql/npgsql/blob/v5.0.4/.github/workflows/build.yml
+    - task: DotNetCoreCLI@2
+      displayName: dotnet build Datadog.Trace.ClrProfiler.Managed.Loader
+      inputs:
+        command: build
+        projects: src/Datadog.Trace.ClrProfiler.Managed.Loader/Datadog.Trace.ClrProfiler.Managed.Loader.csproj
+        arguments: --configuration Release
 
-      # First uninstall any PostgreSQL installed on the image
-      dpkg-query -W --showformat='${Package}\n' 'postgresql-*' | xargs sudo dpkg -P postgresql
+    - task: DotNetCoreCLI@2
+      displayName: dotnet publish Datadog.Trace.ClrProfiler.Managed 2.0
+      inputs:
+        command: publish
+        publishWebProjects: false
+        modifyOutputPath: false
+        zipAfterPublish: false
+        projects: src/Datadog.Trace.ClrProfiler.Managed/Datadog.Trace.ClrProfiler.Managed.csproj
+        arguments: --configuration Release --framework netstandard2.0 --output $(publishOutput)/netstandard2.0
 
-      wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | sudo apt-key add -
-      sudo sh -c 'echo "deb http://apt.postgresql.org/pub/repos/apt/ focal-pgdg main 13" >> /etc/apt/sources.list.d/pgdg.list'
-      sudo apt-get update -qq
-      sudo apt-get install -qq postgresql-${{ variables.pg_major }}
-      sudo -u postgres psql -c "CREATE USER npgsql_tests SUPERUSER PASSWORD 'npgsql_tests'"
-      sudo -u postgres psql -c "CREATE DATABASE npgsql_tests OWNER npgsql_tests"
-      sudo -u postgres psql -c "CREATE EXTENSION citext" npgsql_tests
-      sudo -u postgres psql -c "CREATE EXTENSION hstore" npgsql_tests
-      sudo -u postgres psql -c "CREATE EXTENSION ltree" npgsql_tests
+    - task: DotNetCoreCLI@2
+      displayName: dotnet publish Datadog.Trace.ClrProfiler.Managed 3.1
+      inputs:
+        command: publish
+        publishWebProjects: false
+        modifyOutputPath: false
+        zipAfterPublish: false
+        projects: src/Datadog.Trace.ClrProfiler.Managed/Datadog.Trace.ClrProfiler.Managed.csproj
+        arguments: --configuration Release --framework netcoreapp3.1 --output $(publishOutput)/netcoreapp3.1
+        
+    - task: DockerCompose@0
+      displayName: docker-compose run Profiler
+      inputs:
+        containerregistrytype: Container Registry
+        dockerComposeCommand: run -e buildConfiguration=Release Profiler
 
-      # To disable PostGIS for prereleases (because it usually isn't available until late), surround with the following:
-      # if [ -z "${{ variables.pg_prerelease }}" ]; then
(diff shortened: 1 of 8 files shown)


.azure-pipelines/third-party-test-suites.yml
```diff
@@ -0,0 +1,369 @@
...
+        dd-trace --dd-env CI --dd-service npgsql.tests --dd-version v4.1.9 dotnet "test npgsql/test/Npgsql.Tests --logger trx;LogFileName=npgsql.tests.$(tagName).trx --results-directory npgsql/test/results"
+      env:
+        DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
+      displayName: Test
+
+    - script: |
+        dd-trace --dd-env CI --dd-service npgsql.plugintests --dd-version v4.1.9 dotnet "test npgsql/test/Npgsql.PluginTests --logger trx;LogFileName=npgsql.plugintests.$(os).$(tagName).trx --results-directory npgsql/test/results"
+      env:
+        DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
+      displayName: Test Plugins
```
andrewlock, line 361: Could probably make these a Nuke Target, but given they're not doing much, probably easier not to bother!

.azure-pipelines/third-party-test-suites.yml
```diff
@@ -0,0 +1,369 @@
...
+  dependsOn: dotnet_tool
+  jobs:
+  - job: Postgres
+    strategy:
+      matrix:
+        Ubuntu_Major5:
+          imageName: ubuntu-20.04
+          os: ubuntu
+          tagName: v5.0.7
+        Windows_Major5:
```
andrewlock, line 193: Should we add Alpine + Arm64 to this matrix?

.azure-pipelines/third-party-test-suites.yml
```diff
@@ -0,0 +1,369 @@
...
+        else
+              echo "host all npgsql_tests_scram all scram-sha-256" > pgsql/PGDATA/pg_hba.conf
+        fi
+        echo "host all all all md5" >> pgsql/PGDATA/pg_hba.conf
+        echo "host replication all all md5" >> pgsql/PGDATA/pg_hba.conf
+      displayName: Start PostgreSQL ${{ variables.pg_major }} (Windows)
+      condition: startsWith(variables['imageName'], 'windows')
+
+    - script: |
+        dd-trace --dd-env CI --dd-service npgsql.tests --dd-version v4.1.9 dotnet "test npgsql/test/Npgsql.Tests --logger trx;LogFileName=npgsql.tests.$(tagName).trx --results-directory npgsql/test/results"
```
andrewlock, line 352: I think `4.1.9` should be `$(tagName)`? And maybe we want to split by os?
```suggestion
        dd-trace --dd-env CI --dd-service npgsql.tests.$(os) --dd-version $(tagName) dotnet "test npgsql/test/Npgsql.Tests --logger trx;LogFileName=npgsql.tests.$(tagName).trx --results-directory npgsql/test/results"
```

andrewlock (approved):
LGTM, I don't see the test failures though, are we explicitly ignoring them?

Also, there's a lot (200 lines) of duplication with the ultimate-pipeline. I wonder if we should either:

* Move the final stage to the ultimate-pipeline, but only run it on the scheduled build
* Extract all the build stages to be "steps" that can be imported in both places (doesn't remove the duplication in variables etc)
* Update this pipeline to pull the dd-trace directly from the latest master build. This would let you remove all the build and tool stages, and just pull the latest dd-tool from the latest master branch build.

The final option seems like quite an elegant solution to me.

.azure-pipelines/third-party-test-suites.yml
```diff
@@ -0,0 +1,369 @@
...
+      env:
+        DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
+      displayName: Test
+
+    - script: |
+        dd-trace --dd-env CI --dd-service npgsql.plugintests --dd-version v4.1.9 dotnet "test npgsql/test/Npgsql.PluginTests --logger trx;LogFileName=npgsql.plugintests.$(os).$(tagName).trx --results-directory npgsql/test/results"
```
tonyredondo, line 358: Non blocking, just a personal preference...

I think you can avoid the `"` by using a `--` before the final command to wrap.
Also I prefer to use `=` to make all option values more clear.

```suggestion
        dd-trace --dd-env=CI --dd-service=npgsql.plugintests --dd-version=v4.1.9 -- dotnet test npgsql/test/Npgsql.PluginTests --logger trx;LogFileName=npgsql.plugintests.$(os).$(tagName).trx --results-directory npgsql/test/results
```

zacharycmontoya:
> Update this pipeline to pull the dd-trace directly from the latest master build. This would let you remove all the build and tool stages, and just pull the latest dd-tool from the latest master branch build.

This is a great idea. I was able to get implement it in 8187315 . Since some of our master builds don't build the dotnet tool, I have to add a build tag to the conslidated-pipeline to specify which of our builds produce it, and then the new pipeline will search builds with the tag to find the latest artifact to download.

.azure-pipelines/third-party-test-suites.yml
zacharycmontoya: Yeah that's what I was thinking 😄 Maybe if some other repo testing gets to be more complicated, we could reconsider

.azure-pipelines/third-party-test-suites.yml
zacharycmontoya: Thanks for the suggestion, I applied the `option=value`.
For the `"` style, I tried to remove them but the command line was interpreting `-- results-directory` as a separate command after running the `dd-trace` command, so I put them back

.azure-pipelines/third-party-test-suites.yml
zacharycmontoya: I wasn't considering adding those because I was focused on running only the platforms that the repo was already testing on. This way I don't spend a lot of time trying to figure out how to do the setup on new platforms. Do you think the current Windows/Ubuntu testing is sufficient?

.azure-pipelines/third-party-test-suites.yml
zacharycmontoya: Good catch on the tag name. For the operating system, our CI already adds a tag to the tests by the OS Platform so I'm thinking of not adding the OS to the service name. Any objections @andrewlock @tonyredondo ?

zacharycmontoya:
> LGTM, I don't see the test failures though, are we explicitly ignoring them?
Since I had to publish the test results separately via the `PublishTestResults@2` task, the script to run the tests exited successfully and the test results upload didn't automatically trigger a failure. I kept those defaults just to make sure that I could tell things were working while I continued to tweak the run, but now I'm turning on the setting `failTaskOnFailedTests: true` so that failed tests will cause the build to fail.

zacharycmontoya:
At this time, I think I've addressed all of the feedback and you can see from the testing of commit 88e4dea8cf47da57004f733db3ffcb60ad4f01af, the Third Party Test Suite is failing now with those four tests I listed, which is fine for now. Once I get approvals (so I don't need to run the test suite pipeline again) I'll modify the test run to grab the latest dotnet-tool build from master instead of this branch.

.azure-pipelines/third-party-test-suites.yml
andrewlock: Ah, good point. I was thinking about what _we_ run on, but yeah agreed, we definitely don't want to take on that complexity

.azure-pipelines/third-party-test-suites.yml
```diff
@@ -0,0 +1,239 @@
...
+
+# Global variables
+variables:
+  buildConfiguration: Release
+  dotnetCoreSdk5Version: 5.0.103
+  relativeTracerHome: /src/bin/windows-tracer-home
+  relativeArtifacts: /src/bin/artifacts
+  ddTracerHome: $(System.DefaultWorkingDirectory)/src/bin/dd-tracer-home
+  tracerHome: $(System.DefaultWorkingDirectory)/src/bin/windows-tracer-home
+  artifacts: $(System.DefaultWorkingDirectory)/src/bin/artifacts
```
lucaspimentel, line 20: Aren't these moving to `$(System.DefaultWorkingDirectory)/bin/...` (dropping the `src`)

.azure-pipelines/third-party-test-suites.yml
```diff
@@ -0,0 +1,239 @@
...
+
+    # Download the latest runner tool from the 'consolidated-pipeline' pipeline, from the master branch
+    - task: DownloadPipelineArtifact@2
+      inputs:
+        source: 'specific'
+        project: 'dd-trace-dotnet'
+        pipeline: 54
+        runVersion: 'latestFromBranch'
+        # runBranch: 'refs/heads/master' # Restore once this is tested
+        runBranch: 'refs/heads/zach/test/third-party-npgsql'
```
lucaspimentel, line 88: Is this a TODO for later?

zacharycmontoya: Yes, before merging to master I'll swap out the lines 😄 Until I merge this, I can't pull build artifacts from master, only this branch which has the required change.

lucaspimentel (approved):

.azure-pipelines/third-party-test-suites.yml
zacharycmontoya: I copied and pasted the values from our consolidated pipeline yml. I think you're right that the location has changed, but maybe we're no longer using some of these variables

zacharycmontoya:
Getting ready to merge so I updated the `runBranch` to grab Tracer artifacts from master in the latest commit

.azure-pipelines/third-party-test-suites.yml
andrewlock: We should probably update the consolidated pipeline, but it's not a big issue. These variables override the "default" values, which, as you say, has dropped the src.

PR #1433, opened 2021-04-27
add DSC configuration to deploy Agent and Tracer

lucaspimentel:
Add a PowerShell DSC Configuration that can be used to deploy the Datadog Agent and the .NET Tracer on a Windows machine. For now, this requires updating the msi product ids the Configuration after every release manually, but this PR includes a script to extract those from an msi file.


tools/PowerShell-DSC/DatadogApmDotnet.ps1
```diff
@@ -0,0 +1,60 @@
...
+    )
+
+    Import-DscResource -ModuleName PSDscResources -Name MsiPackage
+    Import-DscResource -ModuleName PSDscResources -Name Environment
+
+    # Version of the Agent package to be installed
+    $AgentVersion = '7.27.0'
+
+    # Version of the Tracer package to be installed
+    $TracerVersion = '1.26.1'
```
zacharycmontoya, line 22: Can you add this to `SetAllVersions.cs`?

tools/PowerShell-DSC/DatadogApmDotnet.ps1
```diff
@@ -0,0 +1,60 @@
...
+
+        # .NET Tracer msi installer
+        if ($InstallTracer) {
+            MsiPackage 'dd-trace-dotnet' {
+                Path      = "https://github.com/DataDog/dd-trace-dotnet/releases/download/v$TracerVersion/datadog-dotnet-apm-$TracerVersion-x64.msi"
+                ProductId = '00B19BDB-EC40-4ADF-A73F-789A7721247A'
+                Ensure    = 'Present'
+            }
+
+            Environment 'COR_PROFILER' {
```
zacharycmontoya, line 43: Why do we want to set the environment variable here instead of sticking with the .NET Tracer MSI defaults?

lucaspimentel: You are right. I was experimenting with the idea of doing as much as possible during installation so that the app deployment only needs to set one env var (`COR_ENABLE_PROFILING`) instead of two. My thinking was that the app itself only needs to know if it wants to enable tracing or not, and shouldn't worry about getting the profiler GUID right.

This idea could extend to the msi itself, too. We could set `COR_PROFILER` globally so there's one less env vars for users to set when hosting outside of IIS.

But it was just an experiment, so we can remove this.

tools/PowerShell-DSC/DatadogApmDotnet.ps1
lucaspimentel: I was still figuring things out and I wasn't sure if we wanted to version this file like that. Another option could be to have a DSC Configuration for each version:
```
DatadogApmDotnetTracer-1.26.1.ps1
DatadogApmDotnetTracer-1.27.0.ps1
```
(but then we probably want to split the Agent into a separate Configuration with it's own version? rabbit hole...)

The other issue is that updating this and tagging the `master` branch is not correct, since we still need to update the product id in this file _after_ building the msi. Not sure what to do about that either.

lucaspimentel: I'm not sure what happens if a Configuration requires one msi at one point in time, and later the same Configuration requires a different msi. Does one Config replace the other, or does the machine stay in a unstable state trying to apply _both_ Configs, but applying one undoes the other?

lucaspimentel: If we could update the product id and tag the commit with the correct version, then users could access the versioned files by tag?
https://github.com/DataDog/dd-trace-dotnet/blob/v1.26.1/tools/PowerShell-DSC/DatadogApmDotnet.ps1
(404, don't try it)

PR #1434, opened 2021-04-29
Native callbacks performance improvements

tonyredondo:
This PR contains a couple of performance optimizations to the ClrProfiler native callbacks when using CallTarget.

This includes:
1) A common and generic environment variables parser using macros.
2) Improve the `AssemblyLoadFinished` callbacks ( this applies for both CallTarget and Callsite)
3) Improvements in `ModuleLoadFinished` and `JITCompilationStarted` callback when using CallTarget

Also changes the test helpers to write the `ProcessId` on error to help find the native log filename into the artifacts folder.

### Example running the `Samples.HttpMessageHandler` app in CallTarget:

image

@DataDog/apm-dotnet


src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -282,14 +260,6 @@ HRESULT STDMETHODCALLTYPE CorProfiler::AssemblyLoadFinished(AssemblyID assembly_
     return S_OK;
   }
 
-  if (!is_attached_) {
-    return S_OK;
-  }
```
kevingosse, line 287: Is it safe to remove this check?

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -301,57 +271,68 @@ HRESULT STDMETHODCALLTYPE CorProfiler::AssemblyLoadFinished(AssemblyID assembly_
...
   if (!assembly_info.IsValid()) {
+    Debug("AssemblyLoadFinished: ", assembly_id, " ", hr_status);
     return S_OK;
   }
 
-  ComPtr<IUnknown> metadata_interfaces;
-  auto hr = this->info_->GetModuleMetaData(assembly_info.manifest_module_id, ofRead | ofWrite,
-                                           IID_IMetaDataImport2,
-                                           metadata_interfaces.GetAddressOf());
+  const auto is_managed_profiler = assembly_info.name == WStr("Datadog.Trace.ClrProfiler.Managed");
```
kevingosse, line 278: We should remember to update that when merging the projects

cc @lucaspimentel

src/Datadog.Trace.ClrProfiler.Native/environment_variables_util.h
```diff
@@ -0,0 +1,79 @@
+#ifndef DD_CLR_PROFILER_ENVIRONMENT_VARIABLES_UTIL_H_
```
kevingosse, line 1: @gleocadie Would you mind checking this file? I don't expect any error but I'm not familiar with macros

kevingosse (approved):
LGTM, I would just like somebody more fluent than me in C++ to check the macros

src/Datadog.Trace.ClrProfiler.Native/clr_helpers.cpp
```diff
@@ -375,18 +375,17 @@ std::vector<IntegrationMethod> FilterIntegrationsByTarget(
...
       found = true;
-    }
-    for (auto& assembly_ref : EnumAssemblyRefs(assembly_import)) {
-      const auto metadata_ref =
-          GetReferencedAssemblyMetadata(assembly_import, assembly_ref);
-      // Info(L"-- assembly ref: " , assembly_name , " to " , ref_name);
-      if (AssemblyMeetsIntegrationRequirements(metadata_ref, i.replacement)) {
-        found = true;
+    } else {
+      for (auto& assembly_ref : EnumAssemblyRefs(assembly_import)) {
```
gleocadie, line 381: I know that the same behavior as before, but once AssemblyMeetsIntegrationRequirements returns true, we can break the loop ?
I don't know how time we are in this loop.

src/Datadog.Trace.ClrProfiler.Native/environment_variables_util.h
gleocadie: LGTM

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
tonyredondo: yeah, the check is still on line 298, after the lock guard, this was an optimisation to avoid the lock in case we already know that the profiler is not attached, but the common execution is the opposite.

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -301,57 +271,68 @@ HRESULT STDMETHODCALLTYPE CorProfiler::AssemblyLoadFinished(AssemblyID assembly_
...
+    auto hr = this->info_->GetModuleMetaData(assembly_info.manifest_module_id, ofRead | ofWrite, IID_IMetaDataImport2, metadata_interfaces.GetAddressOf());
 
-  if (assembly_info.name == WStr("Datadog.Trace.ClrProfiler.Managed")) {
-    // Configure a version string to compare with the profiler version
-    std::stringstream ss;
-    ss << assembly_metadata.version.major << '.'
-       << assembly_metadata.version.minor << '.'
-       << assembly_metadata.version.build;
+    if (FAILED(hr)) {
+      Warn("AssemblyLoadFinished failed to get metadata interface for module "
```
lucaspimentel, line 289: Weird how your auto-formatting removed the line breaks in line 286 above, but split `"id "` into a separate string here.

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
lucaspimentel, line 278: We should change this name to avoid conflicts with Continuous Profiler. Maybe something like `is_instrumentation_assembly`?

lucaspimentel: Even without Continuous Profiler, "managed profiler" was not correct even though we use it in several places 😞  The _clr profiler_ has two components: a native library and a managed library.

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
tonyredondo: I think is the visual studio plugin to remove the blank spaces messing with the google format for c++ or something.

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
tonyredondo: sounds good to me, will do.

gleocadie (approved):

src/Datadog.Trace.ClrProfiler.Native/clr_helpers.cpp
tonyredondo: done in 461a6c5

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
tonyredondo: done in 461a6c5

PR #1435, opened 2021-04-30
Optimize log injection with Serilog

kevingosse:
- Use a single enricher to add properties to logs (LibLog uses `LogContext.PushProperty` which creates an enricher for _every_ property)
 - Cache the properties which value doesn't change over time (version, environment, service)
 - Convert `SpanEventArgs` to struct
 - Change the benchmark scenario to be more realistic: use nested spans and use a message template
 - Use AsyncLocal to flow the Serilog context instead of the shared stack. This fixes issues when span operations are interleaved between multiple threads. It also means we register the enricher only once per trace, instead of for every span.
 - Add a unit test for the previous point

Benchmark:
|           Method |     Mean |     Error |    StdDev | Ratio | RatioSD |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------- |---------:|----------:|----------:|------:|--------:|-------:|------:|------:|----------:|
|              Log | 3.306 μs | 0.0440 μs | 0.0672 μs |  1.00 |    0.00 | 0.0153 |     - |     - |   1.48 KB |
|  EnrichedLog old | 6.934 μs | 0.0547 μs | 0.0457 μs |  2.07 |    0.05 | 0.0610 |     - |     - |   5.66 KB |
|  EnrichedLog new | 4.960 μs | 0.0585 μs | 0.0519 μs |  1.48 |    0.04 | 0.0305 |     - |     - |   2.67 KB |

There's still a significant overhead, but I'm afraid we can't do much more with the Serilog API. We'll see how it behaves on the reliability environment, but if we want to go further we should start thinking about using automatic instrumentation for log injection.


lucaspimentel (approved):

test/benchmarks/Benchmarks.Trace/Properties/launchSettings.json
```diff
@@ -0,0 +1,19 @@
+{
```
zacharycmontoya, line 1: I think you can delete this file. I'll bet VS is adding it automatically but you can add this to the csproj to stop it from doing that:
```xml
    <!-- Tell Visual Studio to not create a new launchSettings.json file, even though we have AspNetCore assets -->
    <NoDefaultLaunchSettingsFile>true</NoDefaultLaunchSettingsFile>
```

test/Datadog.Trace.Tests/Logging/SerilogLogProviderTests.cs
```diff
@@ -103,6 +104,56 @@ public void DisabledLibLogSubscriberDoesNotAddCorrelationIdentifiers()
...
+            var barrier1 = new Barrier(2);
+            var barrier2 = new Barrier(2);
+
+            var task1 = Task.Run(() => Thread1(barrier1, tracer));
+            barrier1.SignalAndWait(); // Wait until thread1 has created both scopes
+
+            var task2 = Task.Run(() => Thread2(barrier2, tracer));
+            barrier2.SignalAndWait(); // Wait until thread2 has created both scopes
+
+            // Stack is: Empty -> outer1 -> Empty -> outer2
```
zacharycmontoya, line 125: How did the second Empty get here?

zacharycmontoya: Also, instead of committing comments that document the problematic state, can we document the expected state?

zacharycmontoya (approved):
LGTM with some small comments on the unit test

test/benchmarks/Benchmarks.Trace/Properties/launchSettings.json
kevingosse: TIL

test/Datadog.Trace.Tests/Logging/SerilogLogProviderTests.cs
kevingosse: Serilog context is asynclocal, so when the first span is pushed for the second thread, the context is still empty.

PR #1438, opened 2021-04-30
Stop using Contoso in Samples.MultiDomainHost tests

kevingosse:


test/test-applications/integrations/dependency-libs/Samples.MultiDomainHost.App.NuGetJsonWithRedirects/Program.cs
```diff
@@ -20,34 +24,123 @@ static void InnerMethodToAllowProfilerInjection()
...
             }
 
-            // Add dependency on System.Net.HttpMessageHandler which lives in the System.Net.Http assembly
-            // System.Net.Http can be loaded in the named AppDomain if there's a bindingRedirect enforced
-            Console.WriteLine($"[HttpClient] sending request to {url}");
-            try
+            Console.WriteLine("All done!");
+        }
+
+        private static HttpListener StartHttpListenerWithPortResilience(string uriTemplate, out string uri)
```
lucaspimentel, line 59: Should we move `StartHttpListenerWithPortResilience()` into one of the shared test projects instead of copy/pasting it several times?

kevingosse: That's what I would have normally done, but those tests seem to be all about what dependencies are loaded in what order, so I'm afraid having a shared project could break something (like, if somebody in the future adds to the shared project a dependency to an assembly that is not supposed to be loaded in those tests)

lucaspimentel: Good point! Resolving this comment.

lucaspimentel (approved):

zacharycmontoya (approved):

PR #1440, opened 2021-05-01
Add an OWIN test application to the integration-tests pipeline

zacharycmontoya:
This addresses a gap in our AspNet testing. Even though we do not currently instrument the entrypoint of the OWIN pipeline, we still instrument WebApi2 in this context so we can test against this.

@DataDog/apm-dotnet


test/test-applications/integrations/Samples.Owin.WebApi/Controllers/HomeController.cs
```diff
@@ -0,0 +1,66 @@
...
+{
+    [RoutePrefix("")]
+    public class HomeController : ApiController
+    {
+        [Route("")]
+        [HttpGet]
+        public IHttpActionResult Index()
+        {
+            return Ok();
+        }
```
andrewlock: For simplicity, do we want to just share sources with the existing ASP.NET WebApi2 Sample? Just a thought, not passionate about it, but would ensure we test all the same code paths :)

andrewlock (approved):

zacharycmontoya:
Closing. The work was superseded by #1525

PR #1442, opened 2021-05-03
clean up "PrepareRelease.exe msi"

lucaspimentel:
Small refactoring to clean up dependencies in `PrepareRelease.exe msi` tool discovered while experimenting with merging multiple solution projects into one. Other tools in `PrepareRelease.exe` are not affected. By consolidation build steps, no changes are needed in `PrepareRelease` to support consolidating assemblies in #1443.

- **Use the MSBuild target `PublishManagedProfilerOnDisk` from the `PrepareRelease` tool instead of calling `dotnet publish` multiple times. This better aligns with other build steps that also publish artifacts to a directory.**
- Refactor the MSBuild target above to use a list of TFMs and reduce code duplication.
- Replace reference to `Datadog.Trace.TestHelpers` with a reference to `Datadog.Core.Tools`. The code from the latter was used through a transitive dependency that is no longer needed (less total dependencies).
- Pass the solution path down the method call chain instead of computing it multiple times.
- Pass the output path to the build script and use the same default as the MSBuild target (instead of hardcoding it to a different path).

Testing done:
- build locally with `msbuild Datadog.Trace.proj /t:PublishManagedProfilerOnDisk` (from a _Developer Command Prompt for VS_)
- build locally with `dotnet msbuild Datadog.Trace.proj /t:PublishManagedProfilerOnDisk` (doesn't require a _Developer Command Prompt for VS_)
- run `packages` pipeline in CI: https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=61548&view=results
- run `PrepareRelease.exe msi` and ensure it generates the same output (no `wxs` file diffs)


build/tools/PrepareRelease/publish-all.bat
```diff
@@ -1,13 +1,2 @@
-
-REM SET SOLUTION_DIR=C:\Github\dd-trace-dotnet
-
-SET TOOL_BUILD_CONFIG=Release
-SET INTEGRATIONS_PROJ=%SOLUTION_DIR%\src\Datadog.Trace.ClrProfiler.Managed\Datadog.Trace.ClrProfiler.Managed.csproj
-SET OUTPUT_DIR=%SOLUTION_DIR%\build\tools\PrepareRelease\bin\tracer-home
-
 RMDIR "%OUTPUT_DIR%" /S /Q
-
-dotnet publish %INTEGRATIONS_PROJ% -c %TOOL_BUILD_CONFIG% -f net45 -o "%OUTPUT_DIR%\net45"
```
lucaspimentel, line 10: Note: removing these `dotnet publish` calls is what initially motivated this PR.

zacharycmontoya (approved):
LGTM

lucaspimentel:
Note to self: `src/Datadog.Trace.Tools.Runner/Datadog.Trace.Tools.Runner.proj` can use similar clean up as `build/tools/PrepareRelease/PrepareRelease.csproj`.

andrewlock (approved):

PR #1443, opened 2021-05-03
Merge auto-instrumentation code into Datadog.Trace.dll

lucaspimentel:
_I recommend looking at this PR one commit at a time. They are batched into big changes: move files, delete files, upgrade projects and pipelines, fix regressions, add tests, etc)_

### Goals
- prevent future changes to `internal` APIs from becoming breaking changes when applications mix different versions of the installers and nuget packages
  - allows safer changes to `internal` APIs
  - unblocks several pending PRs which would otherwise incur breaking changes
  - enables further improvements to version mismatch scenarios (e.g. shared storage of active scope)
- small bonus: method calls that previously crossed assembly boundaries are now candidates for method inlining

### Changes
Merge the following projects into the `Datadog.Trace` project:
- `Datadog.Trace.ClrProfiler.Managed`
- `Datadog.Trace.ClrProfiler.Managed.Core`
- `Datadog.Trace.AspNet` (note: we are leaving an empty `Datadog.Trace.AspNet.dll` with type-forwarding for backwards compatibility)
- `Datadog.Trace.Ci.Shared`

This PR moves all code files below, while keeping the same namespaces to avoid changing type names.
Source directory | Destination directory | Namespace (unchanged)
--|--|--
`src/Datadog.Trace.ClrProfiler.Managed` | `src/Datadog.Trace/ClrProfiler` | `Datadog.Trace.ClrProfiler`
`src/Datadog.Trace.ClrProfiler.Managed.Core` | `src/Datadog.Trace/ClrProfiler` | `Datadog.Trace.ClrProfiler`
`src/Datadog.Trace.AspNet` | `src/Datadog.Trace/AspNet` | `Datadog.Trace.AspNet`
`src/Datadog.Trace.Ci.Shared` | `src/Datadog.Trace/Ci` | `Datadog.Trace.Ci`

Other changes to fix regressions caused by changes in this PR:
- `GenerateIntegrationDefinitions`: ignore types that cannot be loaded. In particular, `TracingHttpModule` inherits from `IHttpModule`, which is not available to the nuke builds because they run on `net5.0`. (8ff2bd909096f39b61e94373dd937bbf9f6d6403)
- In .NET Core, allow newer versions of `Datadog.Trace.dll` from native C++ code (6aa89f9f914a42fa27f134a0859b80fb0c79d5e6)

### Test results
I added some test apps and script to `/test/test-applications/regression/MismatchedTracerVersions` ([link](https://github.com/DataDog/dd-trace-dotnet/tree/lpimentel/merge-projects/test/test-applications/regression/MismatchedTracerVersions)). The scripts can build multiple tracer versions with a single command, and run one of the apps by loading the specific tracer home version and nuget version.

For test results, see [testing_results.md](https://github.com/DataDog/dd-trace-dotnet/blob/lpimentel/merge-projects/test/test-applications/regression/MismatchedTracerVersions/testing_results.md).


src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetCore.cs
```diff
@@ -53,12 +53,14 @@ private static Assembly AssemblyResolve_ManagedProfilerDependencies(object sende
...
-            // 1) The AssemblyVersion is greater than or equal to the version used by Datadog.Trace.ClrProfiler.Managed, the assembly
+            // 1) The AssemblyVersion is greater than or equal to the version used by Datadog.Trace, the assembly
             //    will load successfully and will not invoke this resolve event.
-            // 2) The AssemblyVersion is lower than the version used by Datadog.Trace.ClrProfiler.Managed, the assembly will fail to load
+            // 2) The AssemblyVersion is lower than the version used by Datadog.Trace, the assembly will fail to load
             //    and invoke this resolve event. It must be loaded in a separate AssemblyLoadContext since the application will only
             //    load the originally referenced version
-            if (assemblyName.Name.StartsWith("Datadog.Trace.ClrProfiler.Managed", StringComparison.OrdinalIgnoreCase)
+
+            // TODO: confirm that this check is now correct since StartsWith("Datadog.Trace") will match more assembly names
```
lucaspimentel, line 62: TODO: confirm that this check is now correct since StartsWith("Datadog.Trace") will match more assembly names

Datadog.Trace.proj
```diff
@@ -129,38 +129,15 @@
...
+      <!-- Publish Datadog.Trace.MSBuild since we want both Datadog.Trace.MSBuild and its dependency, Datadog.Trace -->
+      <ManagedProfilerPublishProject Include="src\Datadog.Trace.MSBuild\Datadog.Trace.MSBuild.csproj" />
     </ItemGroup>
 
     <MSBuild Targets="Publish" Projects="@(ManagedProfilerPublishProject)" BuildInParallel="false" Properties="Platform=AnyCPU;TargetFramework=%(TargetFramework.Identity);PublishDir=$(TracerHomeDirectory)\%(TargetFramework.Identity)">
       <Output TaskParameter="TargetOutputs" ItemName="CollectedBuildOutput"/>
     </MSBuild>
   </Target>
 
-  <Target Name="PublishBuildLoggerOnDisk">
```
lucaspimentel, line 140: TODO: confirm that this is correct (I don't remember making this change)

src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetCore.cs
lucaspimentel: @zacharycmontoya: I think to match previous behavior with regards to `Datadog.Trace.dll`, we should remove the first `if` clause entirely and only leave:
```csharp
if (File.Exists(path))
{
	StartupLogger.Debug("Loading {0} with DependencyLoadContext.LoadFromAssemblyPath", path);
	return DependencyLoadContext.LoadFromAssemblyPath(path); // Load unresolved framework and third-party dependencies into a custom Assembly Load Context
}
```
If previous versions, if an app included an older nuget package (for example, 1.28.0 home directory and 1.27.1 nuget), we would load both version of `Datadog.Trace.dll`, and all spans would show up, although automatic and manual spans would be split into two traces.

With `assemblyName.Name.StartsWith("Datadog.Trace")`, we fail to load the new dll, so only manual spans show up.

WDYT?

lucaspimentel: (I made the change and waiting for the build to test it.)

lucaspimentel: Confirmed it now behaves like before. Regression fixed.

zacharycmontoya: Yeah sounds good to me, if the user provides their own `Datadog.Trace.dll`, then we must load our own into a separate ALC 👍🏼

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -354,18 +354,31 @@ HRESULT STDMETHODCALLTYPE CorProfiler::AssemblyLoadFinished(AssemblyID assembly_
...
             std::stringstream ss;
             ss << assembly_metadata.version.major << '.' << assembly_metadata.version.minor << '.'
                << assembly_metadata.version.build;
 
             auto assembly_version = ToWSTRING(ss.str());
 
-            // Check that Major.Minor.Build match the profiler version
-            if (assembly_version == ToWSTRING(PROFILER_VERSION))
+            // Check that Major.Minor.Build matches the profiler version.
+            // On .NET Core, allow managed library to be a higher version than the native library.
```
andrewlock, line 377: This is something we don't allow currently right? Do we definitely want to allow that?

While it seems like it would simplify a number of things, it will also significantly reduce the amount of flexibility we have, as it fixes the surface area between managed and native as what we have currently.

e.g. for Telemetry, I'm going to PInvoke into the native lib to get some stats. With this change, I may end up PInvoking into a native lib that doesn't have the method etc. Similarly, the other way, we can't really change any of the managed methods that the native lib calls into, as an "old" native lib could be calling a newer lib (that doesn't have the same signature)

test/test-applications/regression/MismatchedTracerVersions/Build-Tracer.ps1
```diff
@@ -0,0 +1,44 @@
+Param (
+    [String[]]
+    $TracerVersions = ('1.30.0', '1.31.0', '2.0.0'),
+
+    [ValidateScript( { Test-Path $_ -PathType 'Container' })]
+    [String]
+    $SolutionDirectory = (Join-Path $PSScriptRoot '../../../..'),
+
+    [String]
+    $DestinationDirectory = (Join-Path $PSScriptRoot 'tracer-homes')
```
andrewlock, line 10: You adding more script files faster than I can replace them with nuke 😛 
source

andrewlock (approved):
LGTM 🎉 Just a few comments and questions

src/WindowsInstaller/Files.Managed.Net45.GAC.wxs
```diff
@@ -5,24 +5,6 @@
   <?include $(sys.CURRENTDIR)\Config.wxi?>
   <Fragment>
     <ComponentGroup Id="Files.Managed.Net45.GAC" Directory="net45.GAC">
-      <Component Win64="$(var.Win64)">
-        <Condition>NOT WIX_IS_NETFRAMEWORK_461_OR_LATER_INSTALLED</Condition>
-        <File Id="net45_GAC_Datadog.Trace.AspNet.dll"
-              Source="$(var.TracerHomeDirectory)\net45\Datadog.Trace.AspNet.dll"
-              KeyPath="yes" Checksum="yes" Assembly=".net"/>
-      </Component>
```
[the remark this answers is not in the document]
zacharycmontoya: This is outdated now that we're adding the assembly back with a type-forward, but the MSI that installs the binaries to the GAC will manage the reference counting and remove them from the GAC on uninstall, so we can feel comfortable removing entries for files we no longer will be installing

Datadog.Trace.Minimal.sln
```diff
@@ -48,29 +46,17 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Datadog.Trace.ClrProfiler.M
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Datadog.Core.Tools", "tools\Datadog.Core.Tools\Datadog.Core.Tools.csproj", "{3BEACB10-89FE-4F74-8022-1A52F223CE82}"
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Datadog.Trace.AspNet", "src\Datadog.Trace.AspNet\Datadog.Trace.AspNet.csproj", "{B34EDBC7-C5FB-409D-8472-BC7469D6F2BD}"
```
zacharycmontoya, line 51: Could you make sure to add this back in to the `Datadog.Trace.Minimal.sln` file?

LICENSE-3rdparty.csv
```diff
@@ -4,7 +4,7 @@ Datadog.Trace,https://github.com/damianh/LibLog,MIT,Copyright (C) 2011-2017 Dami
 Datadog.Trace,https://github.com/serilog/serilog,Apache-2.0,Copyright (c) 2013-2018 Serilog Contributors
 Datadog.Trace,https://github.com/serilog/serilog-sinks-file,Apache-2.0,Copyright (c) 2016 Serilog Contributors
 Datadog.Trace,https://github.com/JamesNK/Newtonsoft.Json,MIT,Copyright (c) 2007 James Newton-King
-Datadog.Trace.ClrProfiler.Managed,https://github.com/kevin-montrose/Sigil,MS-PL,2013-2016 Kevin Montrose
+Datadog.Trace,https://github.com/kevin-montrose/Sigil,MS-PL,2013-2016 Kevin Montrose
```
zacharycmontoya, line 7: We don't use Sigil anymore so we can actually remove this

build/_build/PrepareRelease/SetAllVersions.cs
```diff
@@ -140,15 +132,6 @@ public void Run()
                 "src/Datadog.Trace/TracerConstants.cs",
                 FourPartVersionReplace);
 
-            // Locked AssemblyVersion #.0.0.0 updates
```
zacharycmontoya, line 143: For the long-term, do you think we should keep shipping a shim for Datadog.Trace.AspNet when Datadog.Trace goes 2.0? Or just stop producing it once we hit 2.0? If we want to keep producing it, then we'll need to restore this

src/Datadog.Trace.Tools.Runner/Datadog.Trace.Tools.Runner.Tool.csproj
```diff
@@ -53,6 +53,4 @@
     <ProjectReference Include="..\Datadog.Trace\Datadog.Trace.csproj" />
   </ItemGroup>
 
-  <Import Project="..\Datadog.Trace.Ci.Shared\Datadog.Trace.Ci.Shared.projitems" Label="Shared" />
```
zacharycmontoya, line 56: Does this still build? The `Crank/Importer.cs` files uses the `Datadog.Trace.Ci` namespace so it may depend on some types from there

src/Datadog.Trace.Tools.Runner/Datadog.Trace.Tools.Runner.Standalone.csproj
```diff
@@ -44,6 +44,4 @@
     <ProjectReference Include="..\Datadog.Trace\Datadog.Trace.csproj" />
   </ItemGroup>
 
-  <Import Project="..\Datadog.Trace.Ci.Shared\Datadog.Trace.Ci.Shared.projitems" Label="Shared" />
```
zacharycmontoya, line 47: Does this still build? The `Crank/Importer.cs` files uses the `Datadog.Trace.Ci` namespace so it may depend on some types from there

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
lucaspimentel: Today we require an exact version match between `Datadog.Trace.ClrProfiler.Native.dll` and `Datadog.Trace.ClrProfiler.Managed.dll`, but on Core CLR we could have a newer `Datadog.Trace.dll` if the app is combining an older msi installer with a newer nuget package. This "just works" (if there are no breaking changes in the APIs) because we don't check the version of `Datadog.Trace.dll` anywhere.

With `Datadog.Trace.ClrProfiler.Managed.dll` going away, the native code must now check if `Datadog.Trace.dll` is loaded, but _a newer version of that assembly was already allowed before_, so making the version check more flexible avoids a regression. Example: an app has a newer nuget package and an older msi installed. Auto-instrumentation today loads the newer `Datadog.Trace.dll` and it works. If we leave the strict version check, auto-instrumentation will stop working after this PR because the native dll won't find an assembly with the exact version it is looking for.

Breaking API changes can still happen between versions, we're just moving the boundary where those breaks can happen. 

Before:
```
Datadog.Trace.ClrProfiler.Native.dll ➡️ Datadog.Trace.ClrProfiler.Managed.dll ➡️ (can break here) ➡️ Datadog.Trace.dll
         1.x                     1.x                                 1.y
```
After:
```
Datadog.Trace.ClrProfiler.Native.dll ➡️ (can break here) ➡️ Datadog.Trace.dll
         1.x                                       1.y
```

> it fixes the surface area between managed and native

I think this surface area is smaller and easier to manage the the previous surface area. Also, this tight coupling should go away soon when we stop using `integrations.json` file.

> for Telemetry, I'm going to PInvoke into the native lib to get some stats. With this change, I may end up PInvoking into a native lib that doesn't have the method

That may be something you need to handle anyway, since we can't always guarantee that the version of managed and native libraries will be the same, or that the native library will even be present.

zacharycmontoya: I don't think we have a choice but to support it, because in one specific scenario there will only be one version of Datadog.Trace loaded:

- NuGet package: Datadog.Trace v1.n
- Platform package: Datadog.Trace v1.n-1
- Result: All loads return Datadog.Trace v.1.n

We're basically removing possible MissingMethodException's in managed code and replacing it with missing PInvoke calls.

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -863,7 +863,7 @@ HRESULT STDMETHODCALLTYPE CorProfiler::JITCompilationStarted(FunctionID function
 
     // The first time a method is JIT compiled in an AppDomain, insert our startup
     // hook which, at a minimum, must add an AssemblyResolve event so we can find
-    // Datadog.Trace.ClrProfiler.Managed.dll and its dependencies on-disk since it
+    // Datadog.Trace.dll and its dependencies on-disk since it
     // is no longer provided in a NuGet package
```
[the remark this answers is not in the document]
zacharycmontoya: For context, at a previous point in time the user had to do two things to get automatic instrumentation working:
1. Install the platform package
2. Add the Datadog.Trace.ClrProfiler.Managed NuGet package

At a certain point we added the new loader mechanisms so that only the platform package was required. So the code path here is conveying that we must ensure the loader was injected, which will trigger the Datadog.Trace.ClrProfiler.Managed assembly load and tell us if we can proceed with instrumentation or not

zacharycmontoya (approved):
LGTM! Left a couple of comments

src/GlobalSuppressions.cs
```diff
@@ -4,3 +4,8 @@
 // </copyright>
 
 using System.Diagnostics.CodeAnalysis;
+
+[assembly: System.Diagnostics.CodeAnalysis.SuppressMessage(
+    "StyleCop.CSharp.OrderingRules",
+    "SA1202:Elements must be ordered by access",
+    Justification = "Allow custom ordering in integrations.")]
```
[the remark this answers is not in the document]
lucaspimentel: I always hated this rule (and a few others) because it never lets me order things the way I want to, so this gave me some joy 😂

test/Datadog.Trace.ClrProfiler.Managed.Tests/Datadog.Trace.ClrProfiler.Managed.Tests.csproj
```diff
@@ -23,6 +23,9 @@
   <ItemGroup Condition="  !$(TargetFramework.StartsWith('net4')) ">
     <PackageReference Include="System.Net.Http" Version="4.3.4" />
     <PackageReference Include="System.Data.SqlClient" Version="4.7.0" />
+    <PackageReference Include="Microsoft.AspNetCore" Version="2.2.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc" Version="2.2.0" />
+    <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="2.2.0" />
```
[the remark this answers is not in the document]
lucaspimentel: I honestly don't remember. I'll try removing them.

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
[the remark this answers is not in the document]
lucaspimentel: Thanks. I'll fix the comment.

test/Datadog.Trace.ClrProfiler.Managed.Tests/Datadog.Trace.ClrProfiler.Managed.Tests.csproj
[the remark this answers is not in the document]
lucaspimentel: Removed in a98bfcc57ca296174dea9549f52a14bab36ecdcc.

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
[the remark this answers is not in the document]
lucaspimentel: Removed the "no longer provided in a NuGet package" in 2a4b753ead784999d7735111ff3d9be7eb63b409 since that was referring to `Datadog.Trace.ClrProfiler.Managed.dll`

Datadog.Trace.Minimal.sln
lucaspimentel: Thanks. This is why I want to get rid of this file 😅 Added back in f34a441eb9fc391588eca4c58fb58e138f018566.

LICENSE-3rdparty.csv
lucaspimentel: Thanks. Removed in a3780dd5877d74e3cd3550bdfc741dc6ab543806.

build/_build/PrepareRelease/GenerateIntegrationDefinitions.cs
```diff
@@ -326,6 +326,18 @@ private static string GetMethodSignature(MethodInfo method, object attribute, st
...
 
+        public static IEnumerable<Type> GetLoadableTypes(this Assembly assembly)
+        {
+            try
+            {
+                return assembly.GetTypes();
+            }
+            catch (ReflectionTypeLoadException e)
+            {
+                return e.Types.Where(t => t != null);
```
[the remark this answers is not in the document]
lucaspimentel: Suggestion accepted in d6614c8e5d98e9042105633bed287ccc7958ffeb.

build/_build/PrepareRelease/SetAllVersions.cs
lucaspimentel: Restored in 3911e194524affdf86edb87aeaa507e8030feb4f for now. We can decide later if we want to drop `Datadog.Trace.AspNet` in 2.0.

lucaspimentel: Actually, if the reason to keep this is backwards-compat, should it stay at `1.0.0.0`? Otherwise users will need to update their `web.config` to specify `2.0.0.0` after the bump.

zacharycmontoya: Yeah you're absolutely right, we need to keep it at `1.0.0.0` always

src/Datadog.Trace.Tools.Runner/Datadog.Trace.Tools.Runner.Tool.csproj
lucaspimentel: The types in namespace `Datadog.Trace.Ci` now live in `Datadog.Trace.dll`, which is referenced above via `<ProjectReference>`. See https://github.com/DataDog/dd-trace-dotnet/tree/lpimentel/merge-projects/src/Datadog.Trace/Ci

src/Datadog.Trace.Tools.Runner/Datadog.Trace.Tools.Runner.Standalone.csproj
lucaspimentel: The types in namespace `Datadog.Trace.Ci` now live in `Datadog.Trace.dll`, which is referenced above via `<ProjectReference>`. See https://github.com/DataDog/dd-trace-dotnet/tree/lpimentel/merge-projects/src/Datadog.Trace/Ci

test/Datadog.Trace.ClrProfiler.Managed.Tests/Datadog.Trace.ClrProfiler.Managed.Tests.csproj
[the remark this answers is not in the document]
lucaspimentel: Now I remember. Some tests fail with:
```
System.Reflection.ReflectionTypeLoadException : Unable to load one or more of the requested types.
Could not load file or assembly 'Microsoft.AspNetCore.Http.Abstractions, Version=2.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'. The system cannot find the file specified.
```
See https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=71459&view=ms.vss-test-web.build-test-results-tab&runId=3132482&resultId=100017&paneView=debug

src/GlobalSuppressions.cs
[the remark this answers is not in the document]
robertpi: I find this rule very annoying too, glad to see the back of it.

PR #1444, opened 2021-05-04
Add kafka instrumentation

andrewlock:
Adds support for instrumenting [Confluent.Kafka](https://github.com/confluentinc/confluent-kafka-dotnet) v1.4.0+

Producers inject the span context into Kafka record headers. Consumers generate a span when a message is received using the extracted propagated context, and leave it open until the next call to `Consumer.Consume()` (or `Consumer.Close()`/`Dispose()`/`Unsubscribe()`). 

This approach  is required due to the design of the [Confluent.Kafka Consumer API](https://github.com/confluentinc/confluent-kafka-dotnet/blob/1.5.x/src/Confluent.Kafka/IConsumer.cs#L54), and assumes that you use the "standard" consumer setup of a `while(true)` loop, [as shown in examples](https://github.com/confluentinc/confluent-kafka-dotnet/blob/e2b0155654c48282faffcbf1cc195c3366e669ba/examples/Consumer/Program.cs#L92), e.g.:

```
using var consumer = new ConsumerBuilder<Ignore, string>(config).Build();
consumer.Subscribe(topics);
while (true)
{
    // blocks until a message is available, closes existing consume span if open
    var consumeResult = consumer.Consume(cancellationToken);
    // on completion, context propagated from producer and new span created
    // ...use consumerResult in handler
}

// Close final span
consumer.Close();
``` 

If customers are using a different setup, then they can disable the consumer behaviour using `DD_TRACE_KAFKA_CREATE_CONSUMER_SCOPE_ENABLED=false`. This will still create the span for the `ConsumeResult`, but will immediately close it, rather than leaving it open. Customers can then manually extract the headers from the `ConsumeResult` to reconstruct the context for manual instrumentation as they see fit

## Metadata 
- Service = `{service}-kafka`
- Span kind = `producer` | `consumer`
- Span type = `queue`
- Span name = `kafka.produce` | `kafka.consume`
- Resource Name = `{Produce|Consume} Topic {TopicName}`
- Tags
    - `component=kafka` 
    - `kafka.partition`
    - `kafka.offset`
    - `kafka.tombstone`
- Metrics
    - `message.queue_time_ms`

@DataDog/apm-dotnet


build/docker/Datadog.Trace.ClrProfiler.IntegrationTests.sh
```diff
@@ -34,4 +34,6 @@ wait-for-it elasticsearch5:9200 -- \
 wait-for-it sqlserver:1433 -- \
 wait-for-it mongo:27017 -- \
 wait-for-it postgres:5432 -- \
+wait-for-it -t 60 kafka-zookeeper:2181 -- \
+wait-for-it -t 60 kafka-broker:9092 -- \
```
colin-higgins, line 38: I like that you added a timeout, we should probably consider adding a more reasonable one for each

src/Datadog.Trace/Configuration/IntegrationIds.cs
```diff
@@ -13,6 +13,7 @@ internal enum IntegrationIds
         AspNetMvc,
         AspNetWebApi2,
         GraphQL,
+        Kafka,
```
zacharycmontoya, line 16: Similar to this comment, can you add this to the end of the enums just in case: https://github.com/DataDog/dd-trace-dotnet/pull/1457#discussion_r630895274

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaConsumerConsumeIntegration.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// </summary>
+        /// <typeparam name="TTarget">Type of the target</typeparam>
+        /// <param name="instance">Instance value, aka `this` of the instrumented method.</param>
+        /// <param name="millisecondsTimeout">The maximum period of time the call may block.</param>
+        /// <returns>Calltarget state value</returns>
+        public static CallTargetState OnMethodBegin<TTarget>(TTarget instance, int millisecondsTimeout)
+        {
+            // If we are already in a consumer scope, close it, and start a new one on method exit.
+            KafkaHelper.CloseConsumerScope(Tracer.Instance);
+            return CallTargetState.GetDefault();
```
zacharycmontoya, line 33: If `Tracer.Instance.Settings.KafkaCreateConsumerScopeEnabled` is set and the span only represents the consume action, I'm thinking you could get the current DateTimeOffset here so the duration of the span represents when the method started and returned. You can see one example in the RabbitMQ integration for reference below. Do you think this would be useful?

https://github.com/DataDog/dd-trace-dotnet/blob/33ca4afc9c935b9800a12d934821430af8a2d0e2/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/RabbitMQ/BasicGetIntegration.cs#L36-L39

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaConsumerConsumeIntegration.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// <param name="response">Response instance</param>
+        /// <param name="exception">Exception instance in case the original code threw an exception.</param>
+        /// <param name="state">Calltarget state value</param>
+        /// <returns>A response value, in an async scenario will be T of Task of T</returns>
+        public static CallTargetReturn<TResponse> OnMethodEnd<TTarget, TResponse>(TTarget instance, TResponse response, Exception exception, CallTargetState state)
+        {
+            IConsumeResult consumeResult = null;
+            if (response is not null)
+            {
+                // Have to do the cast here, not in the method signature, as otherwise response is never null
```
zacharycmontoya, line 51: If you add the generic constraint `TResponse : IConsumeResult, IDucktype`, then you could do your "null check" by checking if `response.Instance is not null`. I'm not sure if one way or the other is faster, but at least that way you wouldn't have to go out of your way to do the cast

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -0,0 +1,220 @@
...
+
+                var activeScope = tracer.ActiveScope;
+                var currentSpan = activeScope?.Span;
+                if (currentSpan?.OperationName != KafkaConstants.ConsumeOperationName)
+                {
+                    // Not currently in a consumer operation
+                    return;
+                }
+
+                // TODO: record end-to-end time?
```
zacharycmontoya, line 179: The end-to-end time is already recorded as the Span duration. Did you have something else in mind?

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/ITimestamp.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// <summary>
+    /// Timestamp struct for duck-typing
+    /// Requires boxing, but necessary as we need to duck-type <see cref="Type"/> too
+    /// </summary>
+    public interface ITimestamp
+    {
+        /// <summary>
+        /// Gets the timestamp type
+        /// </summary>
+        public TimestampType Type { get; }
```
zacharycmontoya, line 14: If we declared this to return an `int`, would that allow us to do any additional optimizations?

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaConstants.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        internal const string IntegrationName = nameof(IntegrationIds.Kafka);
+        internal const string ConsumeOperationName = "kafka.consume";
+        internal const string ProduceOperationName = "kafka.produce";
+        internal const string TopicPartitionTypeName = "Confluent.Kafka.TopicPartition";
+        internal const string MessageTypeName = "Confluent.Kafka.Message`2[!0,!1]";
+        internal const string ConsumeResultTypeName = "Confluent.Kafka.ConsumeResult`2[!0,!1]";
+        internal const string ActionOfDeliveryReportTypeName = "System.Action`1[Confluent.Kafka.DeliveryReport`2[!0,!1]]";
+        internal const string TaskDeliveryReportTypeName = "System.Threading.Tasks.Task`1[Confluent.Kafka.DeliveryReport`2[!0,!1]]";
+        internal const string DeliveryReportOfNullTypeName = "Confluent.Kafka.DeliveryReport`2[Confluent.Kafka.Null,Confluent.Kafka.Null]";
```
zacharycmontoya, line 15: I didn't see this being used anywhere, can it be removed?

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaProduceSyncIntegration.cs
```diff
@@ -0,0 +1,67 @@
...
+        public static CallTargetState OnMethodBegin<TTarget, TTopicPartition, TMessage, TDeliveryHandler>(TTarget instance, TTopicPartition topicPartition, TMessage message, TDeliveryHandler deliveryHandler)
+            where TMessage : IMessage
+        {
+            // manually doing duck cast here so we have access to the _original_ TopicPartition type
+            // as a generic parameter, for injecting headers
+            Scope scope = KafkaHelper.CreateProducerScope(
+                Tracer.Instance,
+                topicPartition.DuckCast<ITopicPartition>(),
+                isTombstone: message.Value is null,
+                finishOnClose: deliveryHandler is null);
```
zacharycmontoya, line 42: If the span doesn't get closed when the Scope is disposed, who ensures it gets closed?

zacharycmontoya: Resolving, I hadn't finished reading to the ProduceSync DeliveryHandler integration

src/Datadog.Trace/Tags.cs
```diff
@@ -214,6 +214,26 @@ public static class Tags
         /// </summary>
         public const string MessageSize = "message.size";
 
+        /// <summary>
+        /// The partition associated with a record
+        /// </summary>
+        public const string Partition = "partition";
```
zacharycmontoya, line 220: Can we prefix these with `kafka.`? Do you know what other tracers have done for tag names?

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaProduceSyncDeliveryHandlerIntegration.cs
```diff
@@ -0,0 +1,182 @@
...
+        /// <param name="instance">Instance value, aka `this` of the instrumented method.</param>
+        /// <param name="exception">Exception instance in case the original code threw an exception.</param>
+        /// <param name="state">Calltarget state value</param>
+        /// <returns>A response value, in an async scenario will be T of Task of T</returns>
+        public static CallTargetReturn OnMethodEnd<TTarget>(TTarget instance, Exception exception, CallTargetState state)
+        {
+            if (state.State is Action<ITypedDeliveryHandlerShimAction> updateHandlerAction
+             && instance.TryDuckCast<ITypedDeliveryHandlerShimAction>(out var shim))
+            {
+                updateHandlerAction.Invoke(shim);
```
zacharycmontoya, line 97: Can you create a try/catch block to wrap this `Invoke` call instead of inside of the delegate? That way, we can remove the the use of the `Span` argument inside the delegate and we can get rid of that closure

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaProduceSyncDeliveryHandlerIntegration.cs
```diff
@@ -0,0 +1,182 @@
...
+                    return CallTargetState.GetDefault();
+                }
+
+                var newAction = CachedWrapperDelegate<TActionOfDeliveryReport>.CreateWrapper(handler, span);
+
+                Action<ITypedDeliveryHandlerShimAction> updateHandlerAction = inst =>
+                {
+                    try
+                    {
+                        inst.Handler = newAction;
```
zacharycmontoya, line 62: I don't understand why we need to re-set the Handler field inside of a new Action instead of just executing the code. Is there a specific reason for that? If we can avoid that, then that's one fewer allocation

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaConsumerConsumeIntegration.cs
andrewlock: If I'm understanding correctly, I'm not sure if it's very useful in the Kafka case, as it would often be a "meaningless" number IMO, given `Consumer.Consume()` can block for an arbitrary amount of time. 

For example, imagine my message handler takes 100ms to run, but a message isn't available for 3s. Using this approach would give the impression that the message-handling loop took 3.1s, instead of the 0.1s it would with the current implementation?

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaConsumerConsumeIntegration.cs
andrewlock: Ahhh, I was trying this before without the `IDucktype` interface, and was running into issues because `response` is never null! Hadn't noticed `IDucktype` before, I'll update that, thanks 👍

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaHelper.cs
andrewlock: The Java implementation records the _trace_ end-to-end time. That requires passing the trace start time around as baggage (as the start will be a distributed trace!) Seems like a lot of work, and to my mind would require sending that data in _all_ distributed spans, which touches a lot of places. I'm not convinced of the utility, what do you think?

https://github.com/DataDog/dd-trace-java/blob/e5f2514a90b04f4f7620d1497207b5a0224d0a3e/dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/KafkaDecorator.java#L93-L105

src/Datadog.Trace/Tags.cs
andrewlock: I copied these from Java 🙂 https://github.com/DataDog/dd-trace-java/blob/e5f2514a90b04f4f7620d1497207b5a0224d0a3e/internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/InstrumentationTags.java#L11-L15

I agree `kafka.` prefixes would be better, but I thought we'd probably need to stay consistent with other tracers?

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaProduceSyncDeliveryHandlerIntegration.cs
andrewlock: I _think_ we have to do it this way. 

We need to set the handler in `MethodEnd`, not in `MethodBegin`. But we have to generate the `newAction` variable in `MethodBegin` as that's where we have access to the `TActionOfDeliveryReport` type. 

With the approach you're suggesting, we would need to pass the `newAction` as state in `CallTargetState`. But then we wouldn't be able to `cast` it to the correct type in `MethodEnd`, as we don't have the `TActionOfDeliveryReport` generic parameter there. 

So I _think_ this approach is the only one that is really doable - I couldn't find a way to set it directly in `MethodEnd` without either losing the benefits of `CachedWrapperDelegate`, or having to set the `Handler` field through reflection?

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/ITimestamp.cs
andrewlock: I thought I'd tried that but I couldn't get the duck typing to work 🤔  I'll give it another try

src/Datadog.Trace/Configuration/IntegrationIds.cs
andrewlock: Done in https://github.com/DataDog/dd-trace-dotnet/pull/1444/commits/dff1497fe9b3af8a43b35d72a32c5a91fa8b1fdc

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaConsumerConsumeIntegration.cs
andrewlock: Done in https://github.com/DataDog/dd-trace-dotnet/pull/1444/commits/277945283c59bd5569a1d00d9904bd322675bc8c

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/ITimestamp.cs
andrewlock: done in https://github.com/DataDog/dd-trace-dotnet/pull/1444/commits/b1419d91627e7511ed11be120367c23809606ff1

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaConstants.cs
andrewlock: Done in https://github.com/DataDog/dd-trace-dotnet/pull/1444/commits/119930f7ef7af6994f9d59c0debc1cea8d7847b6

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaProduceSyncDeliveryHandlerIntegration.cs
andrewlock: Done in https://github.com/DataDog/dd-trace-dotnet/pull/1444/commits/67ead42e5d0b9db676a513c47530817c3281cc4a

src/Datadog.Trace/Tags.cs
lucaspimentel: I would try to find an RFC or worst case scenario look at the source code for other tracers as well. Often Java implements features first (before there is a cross-language RFC), so copying their implementation is not always the best idea.

Datadog.Trace.sln
```diff
@@ -430,6 +430,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "DuplicateTypeProxy", "test\
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Samples.AspNetCoreRazorPages", "test\test-applications\integrations\Samples.AspNetCoreRazorPages\Samples.AspNetCoreRazorPages.csproj", "{1B9E6BF4-9D48-4988-9945-248096119E46}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Samples.Kafka", "test\test-applications\integrations\Samples.Kafka\Samples.Kafka.csproj", "{760B1586-23B4-418A-B7D9-ABC0AC023FAE}"
```
lucaspimentel, line 433: I don't know how the `FAE04EC0` keep sneaking in, but I'm pretty sure this should be `9A19103F`. If you remove the project and re-add it with Visual Studio, it will change the GUID. (The GUIDs for `Samples.AspNetCoreRazorPages` above and `FeatureTracking` below are already getting fixed in another PR: #1463)

integrations.json
```diff
@@ -612,6 +612,176 @@
...
+          ],
+          "minimum_major": 1,
+          "minimum_minor": 4,
+          "minimum_patch": 0,
+          "maximum_major": 1,
+          "maximum_minor": 65535,
+          "maximum_patch": 65535
+        },
+        "wrapper": {
+          "assembly": "Datadog.Trace.ClrProfiler.Managed, Version=1.26.2.0, Culture=neutral, PublicKeyToken=def86d061d0d2eeb",
```
lucaspimentel, line 635: 1.26.3 is already released. Can you merge/rebase `master` and re-generate this file?

lucaspimentel: If we ran the integration tests on .NET Framework, they would fail since the assembly with exact version 1.26.2 would not be found.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -0,0 +1,220 @@
...
+        {
+            // TODO: Allow disabling globally
+            if (!_headersInjectionEnabled)
+            {
+                return;
+            }
+
+            try
+            {
+                var headers = message.Headers ?? CachedMessageHeadersHelper<TTopicPartitionMarker>.CreateAndSetHeaders(message);
```
lucaspimentel, line 206: I think there's more coupling here than needed by passing `message`. How about something like this so that `CachedMessageHeadersHelper.CreateHeaders()` is only concerned with creating the headers and not setting them?
```csharp
if(message.Headers == null)
{
    message.Headers = CachedMessageHeadersHelper<TTopicPartitionMarker>.CreateHeaders();
}

var adapter = new KafkaHeadersCollectionAdapter(message.Headers);
```

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -0,0 +1,220 @@
...
+        /// Try to inject the prop
+        /// </summary>
+        /// <param name="context">The Span context to propagate</param>
+        /// <param name="message">The duck-typed Kafka Message object</param>
+        /// <typeparam name="TTopicPartitionMarker">The TopicPartition type (used  optimisation purposes)</typeparam>
+        /// <typeparam name="TMessage">The type of the duck-type proxy</typeparam>
+        internal static void TryInjectHeaders<TTopicPartitionMarker, TMessage>(SpanContext context, TMessage message)
+            where TMessage : IMessage
+        {
+            // TODO: Allow disabling globally
```
lucaspimentel, line 198: Does this TODO need to be resolved in this PR?

test/test-applications/integrations/Samples.Kafka/Properties/launchSettings.json
```diff
@@ -0,0 +1,21 @@
...
+      "environmentVariables": {
+        "DD_DOTNET_TRACER_HOME": "C:\\repos\\dd-trace-dotnet\\src\\bin\\tracer-home",
+        "COR_ENABLE_PROFILING": "1",
+        "CORECLR_ENABLE_PROFILING": "1",
+        "COR_PROFILER_PATH": "C:\\repos\\dd-trace-dotnet\\src\\bin\\tracer-home\\win-x64\\Datadog.Trace.ClrProfiler.Native.dll",
+        "DD_DUMP_ILREWRITE_ENABLED": "1",
+        "CORECLR_PROFILER": "{846F5F1C-F9AE-4B07-969E-05C26BC060D8}",
+        "DD_TRACE_CALLTARGET_ENABLED": "true",
+        "DD_INTEGRATIONS": "C:\\repos\\dd-trace-dotnet\\src\\bin\\tracer-home\\integrations.json",
+        "DD_ENV": "andrew",
```
lucaspimentel, line 14: Hello, `andrew`

test/test-applications/integrations/Samples.Kafka/Properties/launchSettings.json
```diff
@@ -0,0 +1,21 @@
+{
+  "profiles": {
+    "Samples.Kafka": {
+      "commandName": "Project",
+      "environmentVariables": {
+        "DD_DOTNET_TRACER_HOME": "C:\\repos\\dd-trace-dotnet\\src\\bin\\tracer-home",
```
lucaspimentel, line 6: please remove the absolute paths

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaHelper.cs
andrewlock: > the .NET Tracer doesn't have baggage support yet

Well that solves that then 😄

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaHelper.cs
andrewlock: I think that can be removed actually - I was initially questioning whether we needed to do that or not, based on the Java tracer https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/KafkaProducerInstrumentation.java#L87

However, I tested with an older broker (pre <0.11.0) with the 1.4.x library and it just silently didn't propagate them, so I think we're safe. I'll remove this comment

integrations.json
andrewlock: Can't keep up with the pace of releases atm 😂

Datadog.Trace.sln
andrewlock: I wonder if it's due to Rider. I created the razor pages one and this one. Tony created Feature Tracking (I'm guessing with Rider). 

If only _.sln_ files were actually understandable. Doesn't seem to make any difference which GUID I use!

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaHelper.cs
andrewlock: Sure, this method was originally doing a lot more, I'll fix as suggested

src/Datadog.Trace/Tags.cs
andrewlock: Looks like ruby uses `kafka.partition` etc

https://github.com/DataDog/dd-trace-rb/blob/e63e65f19887d011cd957f7e58361f4af0df2050/lib/ddtrace/contrib/racecar/ext.rb#L19-L21

as does js:

https://github.com/DataDog/dd-trace-js/blob/7b27059fe3b9de6788c0517a2b499ff6b78da5ff/packages/datadog-plugin-kafkajs/src/index.js#L67-L69

So I'll add the `kafka.` prefixes 👍

Datadog.Trace.sln
andrewlock: Also, with 100+ projects...

image

test/test-applications/integrations/Samples.Kafka/Properties/launchSettings.json
andrewlock: done in https://github.com/DataDog/dd-trace-dotnet/pull/1444/commits/13c2a27450c0accdcb58f61a77ea84f5f859df08

test/test-applications/integrations/Samples.Kafka/Properties/launchSettings.json
andrewlock: removed in https://github.com/DataDog/dd-trace-dotnet/pull/1444/commits/13c2a27450c0accdcb58f61a77ea84f5f859df08

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaHeadersCollectionAdapter.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public KafkaHeadersCollectionAdapter(IHeaders headers)
+        {
+            _headers = headers;
+        }
+
+        public IEnumerable<string> GetValues(string name)
+        {
+            // This only returns the _last_ bytes. Accessing other values is more expensive and should generally be unneccessary
+            if (_headers.TryGetLastBytes(name, out var bytes))
```
lucaspimentel, line 23: How many bytes does this read? Are we sure we never read a partial multi-byte character?

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaHelper.cs
andrewlock: done in https://github.com/DataDog/dd-trace-dotnet/pull/1444/commits/4a4325cd76b7494fbfd6d1a04afdcab3dd06516c

integrations.json
andrewlock: updated in https://github.com/DataDog/dd-trace-dotnet/pull/1444/commits/57a4a98b60245400ef1254a75d45c632e732611b

Datadog.Trace.sln
andrewlock: Added and removed in https://github.com/DataDog/dd-trace-dotnet/pull/1444/commits/a9f0a2c6d9c7b5d7d536031f6bf87dd93c1f6d98... VS did...stuff...

src/Datadog.Trace/Tags.cs
andrewlock: added `kafka.` in https://github.com/DataDog/dd-trace-dotnet/pull/1444/commits/0786ef38b54096f694424e973a1474dade8f5aac

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaHeadersCollectionAdapter.cs
andrewlock: It reads all the bytes for the last instance of the header with name `name`. Headers are stored as raw bytes in Kafka, but you can have multiple headers with the same name, hence "last" in the name.

lucaspimentel: Oh, so it's "get the bytes for the last header," not "get the last {count} bytes". Thanks for clarifying!

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaConsumerConsumeIntegration.cs
zacharycmontoya: Got it, I forgot that the `Consumer.Consume()` was blocking. The current implementation sounds good then!

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaProduceSyncDeliveryHandlerIntegration.cs
```diff
@@ -0,0 +1,179 @@
...
+        }
+
+        /// <summary>
+        /// Helper method used by <see cref="CachedWrapperDelegate{TActionDelegate}"/> to create a delegate
+        /// </summary>
+        /// <param name="originalHandler">The original delivery report handler </param>
+        /// <param name="span">A <see cref="Span"/> that can be manipulated when the action is invoked</param>
+        /// <typeparam name="TDeliveryReport">Type of the delivery report</typeparam>
+        /// <returns>The wrapped action</returns>
+        public static Action<TDeliveryReport> WrapAction<TDeliveryReport>(Action<TDeliveryReport> originalHandler, Span span)
```
zacharycmontoya, line 107: Do you think there would be issues if we did not pass a `Span` argument and had the returned delegate just call `Tracer.Instance.ActiveScope`?

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaProduceSyncDeliveryHandlerIntegration.cs
zacharycmontoya: You're right, given all the constraints this does seem like the best way to do it. However, what if instead of instrumenting the constructor of `Confluent.Kafka.Producer`2+TypedDeliveryHandlerShim_Action` we instead just instrument the `HandleDeliveryReport` method itself? Looking at the decompilation for `Confluent.Kafka 1.4.3.0`, all the fields we are reading come straight from the `TopicPartitionOffsetError` property of the `DeliveryReport` argument and these values look to be the same when they're eventually passed to the handler. That could remove a lot of this complexity and possibly some allocations. Have you already tried this approach and, if so, were there some gotcha's there?

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaProduceSyncDeliveryHandlerIntegration.cs
andrewlock: Yeah, I think that could be problematic... The callback could be executed at any point, and I don't think the Thread is even guaranteed. so `Tracer.Instance.ActiveScope.Span` is very likely to _not_ be the `Span` we pass in here I think...

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaProduceSyncDeliveryHandlerIntegration.cs
andrewlock: Absolutely agree this would be an easier approach, and we did consider it. The main problems with this approach are:

1. You lose the correct "start time" for the produce span
2. More critically, we'd lose the correct _parent_ span, given that the callback runs once the `Produce()` method has finished, and potentially a different span is active..
3. We don't have access to the message Value in the callback, so wouldn't be able to tag Tombstones any more

No. 2 is the big problem, and the main reason we went for the current, more complex, approach. Does that make sense to you?

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaProduceSyncDeliveryHandlerIntegration.cs
zacharycmontoya: Understood, thanks for clarifying

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaProduceSyncDeliveryHandlerIntegration.cs
zacharycmontoya: Got it, I didn't realize how crazy this instrumentation scenario would be. Thanks for taking the time to respond to everything 😄

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaProduceSyncDeliveryHandlerIntegration.cs
```diff
@@ -0,0 +1,179 @@
...
+                    return CallTargetState.GetDefault();
+                }
+
+                var newAction = CachedWrapperDelegate<TActionOfDeliveryReport>.CreateWrapper(handler, span);
+
+                Action<ITypedDeliveryHandlerShimAction> updateHandlerAction = inst => inst.Handler = newAction;
+
+                // store the call to update the handler variable as state
+                // so we update it at the _end_ of the constructor
+                return new CallTargetState(scope: null, state: updateHandlerAction);
```
zacharycmontoya, line 62:
```suggestion
                return new CallTargetState(scope: Tracer.Instance?.ActiveScope, state: updateHandlerAction);
```

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/CachedMessageHeadersHelper.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            var headersType = typeof(TMarkerType).Assembly.GetType("Confluent.Kafka.Headers");
+
+            ConstructorInfo ctor = headersType.GetConstructor(System.Type.EmptyTypes);
+
+            DynamicMethod createHeadersMethod = new DynamicMethod(
+                $"KafkaCachedMessageHeadersHelpers",
+                headersType,
+                null,
+                typeof(DuckType).Module,
```
zacharycmontoya, line 22: With our recent robustness to assembly context's, I'm worried that creating this in the `typeof(DuckType).Module` module may not work in some edge cases. It's hard to make concrete without a repro, but the `Samples.ServiceStackRedis` program shows one example of loading the same assembly into multiple contexts and then instrumenting program runs that use each assembly. If this is something you feel should be addressed in a separate PR, we can punt this.

To resolve this, I'm thinking we can re-use some of the logic from our DuckType library found below to generate a new ModuleBuilder (which would then be supplied as the module argument). And instead of getting the generic Type `TMarkerType`, we would pass in the `topicPartition` object itself and get the assembly from the runtime type. What do you think?

https://github.com/DataDog/dd-trace-dotnet/blob/a5878ce09f443eea6165a839f09afc674ea0b357/src/Datadog.Trace/DuckTyping/DuckType.Statics.cs#L82-L96

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaProduceSyncDeliveryHandlerIntegration.cs
andrewlock: Is this necessary? You use the previous approach in the BasicGetIntegration! Don't mind making the change, but I assume they're equivalent as we don't use the scope in `MethodEnd`? https://github.com/DataDog/dd-trace-dotnet/blob/a5878ce09f443eea6165a839f09afc674ea0b357/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/RabbitMQ/BasicGetIntegration.cs#L38

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/CachedMessageHeadersHelper.cs
andrewlock: Just so I understand, the issue you're describing is:

* Multiple instances of the Confluent.Kafka library are loaded
* The `TMarkerType` would have a different identity for each assembly, so the static constructor would be called once for each assembly
* As we're currently creating the type in the `typeof(DuckType).Module`, we'd get an exception trying to create the second one (in the same module)

If so, then yes, I think your suggestion is worth doing. Will have to add caching using a different approach than the static constructor too I guess.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaProduceSyncDeliveryHandlerIntegration.cs
zacharycmontoya: Sorry, I should have explained my suggestion. I think you need this change because in the `OnMethodEnd` below, you try to access `state.Scope?.Span.Finish()` but currently `state.Scope` will always be null. So in this case I think it's needed

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/CachedMessageHeadersHelper.cs
zacharycmontoya: Yes that about sums it up with one minor clarification:

> As we're currently creating the type in the typeof(DuckType).Module, we'd get an exception trying to create the second one (in the same module)
What we observed is that we wouldn't get an exception when we create the dynamic method, but when we go to execute it. The type from new instance of Confluent.Kafka would be passed into the new DynamicMethod, but because the DynamicMethod is still in the same module as the previously created one, the DynamicMethod would resolve type references to the previous (different) assembly/type and a type cast exception would occur.

I don't have a repro right now but since I've been babbling about theoretical stuff I can spend some time on Monday to come up with a repro to validate the changes.

zacharycmontoya: I think the issue is that you're trying to assign the result all back to the same `Message` object. For each assembly test, can you dynamically create a new `Message` object via reflection (in addition to the headers object) instead of re-using the original one?

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/KafkaProduceSyncDeliveryHandlerIntegration.cs
andrewlock: Oh! haha, that's because I moved the try catch to avoid the closure as in your other suggestion. I think actually this flags that by removing the closure I introduced a bug: we can't rely on just fetching the ambient scope here, because we need to ensure we close the span passed to the handler which is _not_ active.

Instead, I think I should remove the `try-catch` from `OnMethodEnd`, and instead move it into the delegate itself (kind of). I think that should work... 🤔

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/CachedMessageHeadersHelper.cs
andrewlock: Unfortunately, I tried that in https://github.com/DataDog/dd-trace-dotnet/pull/1444/commits/c1d613706f8b578b986b1eac5d8a1fe4cc033c97 but still no failures! 🙁

zacharycmontoya: After spending several hours on this I couldn't get it to reproduce. On close inspection, I do think this scenario is different from previously failing scenarios and is safe to use with only one common Module. It's hard to remember the exact details of the previous issue but I believe it centered on the fact that we were using metadata tokens as inputs to be resolved instead of actual MethodInfo / Types that were already instantiated by the runtime. In this scenario, the DynamicMethod does not take any input arguments, it has a return type that we specify via a Type object, and the constructor we invoke has already been created as a RuntimeConstructorInfo object. So long story short, I'm like 70% sure we don't have to do anything extra here

zacharycmontoya: And I can also confirm that the message objects and header objects created in your unit test look 100% correct. Each test results in a header object that comes from the same runtime assembly as the message object.

tonyredondo: Just to clarify, the problem we had on the Ducktyping related to ALC wasn't caused by the `DynamicMethod`, the `DynamicMethod` always returned the right thing. 

The problem was related to the `ModuleBuilder` where we dynamically create and emit a [`delegate`](https://github.com/DataDog/dd-trace-dotnet/blob/master/src/Datadog.Trace/DuckTyping/ILHelpersExtensions.cs#L24-L51). This delegate was used later to [call the `CreateDelegate` method from the `DynamicMethod`](https://github.com/DataDog/dd-trace-dotnet/blob/master/src/Datadog.Trace/DuckTyping/DuckType.Statics.cs#L121-L125). Hence the problem if we emit the delegate for a type in one module builder we cannot reuse the same delegate if we use a dynamic method with a type in a different context. In our case was failing the casting. Same type, different context. The fix? create a new module builder so we can emit the delegate again.

zacharycmontoya (approved):
LGTM. Thanks for all of the follow-up 😄

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Kafka/CachedMessageHeadersHelper.cs
andrewlock: Thanks for the follow up guys, that all makes sense, happy we can explain the problem now! 🙂 

There's just one final thing I don't quite understand. In the test, the DynamicMethod is created 3 times: one for each context. But as I understand it, we're always creating the dynamic module, with the same name (`"KafkaCachedMessageHeadersHelpers"`) in the _same_ `typeof(DuckType).Module`. Why does that not cause name conflicts (or similar) in the module? Is it because the `typeof(DuckType).Module`s end up being in different contexts too (even though we didn't explicitly use `Assembly.LoadFile`/`AssemblyLoadContext` for that assembly)?

zacharycmontoya: I think this is okay because it's not a global name, it's just a name used for stack traces. In other parts of the duck type library we have to be careful and use unique names for proxy types that are created

Datadog.Trace.sln
```diff
@@ -1615,6 +1617,16 @@ Global
...
+		{94B50277-FB50-4B42-BA79-770ADB24CB80}.Debug|Any CPU.ActiveCfg = Debug|x86
+		{94B50277-FB50-4B42-BA79-770ADB24CB80}.Debug|x64.ActiveCfg = Debug|Any CPU
+		{94B50277-FB50-4B42-BA79-770ADB24CB80}.Debug|x64.Build.0 = Debug|Any CPU
+		{94B50277-FB50-4B42-BA79-770ADB24CB80}.Debug|x86.ActiveCfg = Debug|Any CPU
+		{94B50277-FB50-4B42-BA79-770ADB24CB80}.Debug|x86.Build.0 = Debug|Any CPU
+		{94B50277-FB50-4B42-BA79-770ADB24CB80}.Release|Any CPU.ActiveCfg = Release|x86
+		{94B50277-FB50-4B42-BA79-770ADB24CB80}.Release|x64.ActiveCfg = Release|Any CPU
+		{94B50277-FB50-4B42-BA79-770ADB24CB80}.Release|x64.Build.0 = Release|Any CPU
+		{94B50277-FB50-4B42-BA79-770ADB24CB80}.Release|x86.ActiveCfg = Release|Any CPU
+		{94B50277-FB50-4B42-BA79-770ADB24CB80}.Release|x86.Build.0 = Release|Any CPU
```
lucaspimentel, line 1629: If you want to be able to test `Samples.Kafka` in both x86 and x64 from Visual Studio, these configurations need to be fixed. I don't know if that's something we care about or we just leave it for CI to test (not using the solution file).

Right now, the solution configuration is mapping to project configuration like this:
Solution (VS) | Project
--|--
Any CPU | x86
x64 | Any CPU
x86 | Any CPU

src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -388,6 +391,13 @@ public int PartialFlushMinSpans
             }
         }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether a span context should be created on exiting a successful Kafka
+        /// Consumer.Consume() call, and closed on entering Consumer.Consume().
+        /// </summary>
+        /// <seealso cref="ConfigurationKeys.KafkaCreateConsumerScopeEnabled"/>
+        public bool KafkaCreateConsumerScopeEnabled { get; set; }
```
lucaspimentel, line 399: Are we sure we want to make this a public API?

test/test-applications/integrations/Samples.Kafka/Properties/launchSettings.json
```diff
@@ -0,0 +1,20 @@
...
+        "DD_DOTNET_TRACER_HOME": "..\\..\\..\\..\\src\\bin\\tracer-home",
+        "COR_ENABLE_PROFILING": "1",
+        "CORECLR_ENABLE_PROFILING": "1",
+        "COR_PROFILER_PATH": "..\\..\\..\\..\\src\\bin\\tracer-home\\win-x64\\Datadog.Trace.ClrProfiler.Native.dll",
+        "DD_DUMP_ILREWRITE_ENABLED": "1",
+        "CORECLR_PROFILER": "{846F5F1C-F9AE-4B07-969E-05C26BC060D8}",
+        "DD_TRACE_CALLTARGET_ENABLED": "true",
+        "DD_INTEGRATIONS": "..\\..\\..\\..\\src\\bin\\tracer-home\\integrations.json",
+        "CORECLR_PROFILER_PATH": "..\\..\\..\\..\\src\\bin\\tracer-home\\win-x64\\Datadog.Trace.ClrProfiler.Native.dll",
+        "COR_PROFILER": "{846F5F1C-F9AE-4B07-969E-05C26BC060D8}"
```
lucaspimentel, line 15: You can use variables here instead of relative paths. This will be helpful when we move directories around (again). Here's an example: https://github.com/DataDog/dd-trace-dotnet/blob/91196b609f3f39beabb70c72c1de1f8fc2072400/test/test-applications/integrations/Samples.AspNetCoreMvc21/Properties/launchSettings.json#L17-L26

lucaspimentel: Also using `src/bin/tracer-home` is not entirely correct. I don't think we build the tracer home directory in CI before running tests (although we probably should do that instead of copying the files into each sample's output path).

Datadog.Trace.sln
andrewlock: Weird, I can test in Rider, just fine (I have been!) Should be fixed though I guess. I'll look into it (forgot to update these when I added and removed the project to fix the guids)

src/Datadog.Trace/Configuration/TracerSettings.cs
andrewlock: I guess that's a more general question - _should_ we expose these toggles as public APIs for customers who are configuring in code?

src/Datadog.Trace/Tags.cs
```diff
@@ -214,6 +214,26 @@ public static class Tags
...
+
+        /// <summary>
+        /// Whether the record was a "tombstone" record
+        /// </summary>
+        public const string KafkaTombstone = "kafka.tombstone";
+
+        /// <summary>
+        /// The length of time a record has been on the queue
+        /// </summary>
+        public const string MessageQueueTimeMs = "message.queue_time_ms";
```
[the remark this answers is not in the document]
andrewlock: Yeah, probably should move it

test/test-applications/integrations/Samples.Kafka/Properties/launchSettings.json
andrewlock: Yeah, I'm not convinced that works 🤔 Those are MSBuild variables, but launchsettings.json doesn't do that interpolation., so the env variables become exactly that E.g. if I run that Samples.AspNetCoreMvc21 project, I see this: 
image

Datadog.Trace.sln
lucaspimentel: They will work in VS as they are now, but you will not be testing on the architecture you expect. The fact that your `launchSettings.json` only has the path to `win-x64` suggests you haven't tested on x86 (even if you thought you did because you selected x86 in the VS solution configuration drop-down).

test/test-applications/integrations/Samples.Kafka/Properties/launchSettings.json
andrewlock: > Also using ...win-x64\\Datadog.Trace.ClrProfiler.Native.dll won't work when testing x86

I'll give you that one 😛 I'm inclined to just remove these anyway - they're only for local testing, and I'm inclined to having a Nuke target that sets all this up for us...

Datadog.Trace.sln
andrewlock: > The fact that your launchSettings.json only has the path to win-x64 suggests you haven't tested on x86

You got me 😉  Given changing configuration is a VS-hanging event, I relied on CI to do that for me 😉

test/test-applications/integrations/Samples.Kafka/Properties/launchSettings.json
lucaspimentel: >Those are MSBuild variables, but launchsettings.json doesn't do that interpolation

Oh, right. I think they only work when running from inside VS. It's been a long time, but I guess we decided this was better than hard-coded or relative paths.

> I'm inclined to just remove these anyway

I agree, but being able to run them from VS makes it easier to debug the attached tracer.

src/Datadog.Trace/Configuration/TracerSettings.cs
lucaspimentel: Generally speaking, if we want the setting to be discoverable (Intellisense) and part of the permanent public API, make them public. But if it's meant to be a temporary/experimental feature flag, we keep them internal and tell users to use the env var instead.

andrewlock: So this _isn't_ meant to be experimental. Users may need to set it to `false` if they're not using the Kafka library in the "usual" way, i.e. `Consumer.Consume()` in a `while(true)` loop:

https://github.com/DataDog/dd-trace-dotnet/blob/934d97c9fcbbfd750d8dc06343057a72680aa486/test/test-applications/integrations/Samples.Kafka/Consumer.cs#L57-L80

Not really sure _why_ they would need to, but Java have run into situations where it would be required, so good to have the escape hatch

lucaspimentel: Sounds to me like this will stay, then. Thanks!

Datadog.Trace.sln
andrewlock: Fixed in https://github.com/DataDog/dd-trace-dotnet/pull/1444/commits/6506c6db87d0099956fe993310a8ee419c5ad625. I'm thinking I should add a Nuke target to create a new sample app with the correct configuration etc. There's currently a lot of faff required to get this right

src/Datadog.Trace/Tags.cs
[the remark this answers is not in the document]
andrewlock: Moved in https://github.com/DataDog/dd-trace-dotnet/pull/1444/commits/f83a0818062baa9b33ef9b108281d297afcd4006

test/test-applications/integrations/Samples.Kafka/Properties/launchSettings.json
andrewlock: Updated in https://github.com/DataDog/dd-trace-dotnet/pull/1444/commits/39873364a3ccc860177f66f81f5699069e0fe690

lucaspimentel (approved):

PR #1445, opened 2021-05-04
Add and enforce copyright headers for C# files under /build, /src, and /tools

zacharycmontoya:
- Modifies stylecop rules to enforce copyright headers on C# files
- Runs the automated rule to add the missing copyright headers
- Affects all subdirectories EXCEPT /test, which currently doesn't have stylecop applied

@DataDog/apm-dotnet


test/benchmarks/Benchmarks.Trace/Properties/launchSettings.json
```diff
@@ -0,0 +1,27 @@
+{
```
lucaspimentel, line 1: VS keeps trying to add this file on my machine and I keep deleting it 😅 
@kevingosse are we ok adding this?

stylecop.json
```diff
@@ -6,6 +6,11 @@
     },
     "documentationRules": {
       "companyName": "Datadog",
+      "copyrightText": "Unless explicitly stated otherwise all files in this repository are licensed under the {licenseName} License.\nThis product includes software developed at Datadog (https://www.datadoghq.com/). Copyright {year} Datadog, Inc.",
```
lucaspimentel, line 9: Is this the only config change and everything else is just adding the headers?

test/benchmarks/Benchmarks.Trace/Properties/launchSettings.json
zacharycmontoya: I had @kevingosse remove this and tell VS to stop generating it through the csproj. We should no longer have these `launchSettings.json` popping up now after rebasing off `master`

lucaspimentel: thanks!

lucaspimentel (approved):

PR #1446, opened 2021-05-04
Make sure Tracer version in DatadogInstall.ps1 also gets updated by the PrepareRelease tool

zacharycmontoya:
@DataDog/apm-dotnet

### DIFF
--- a/build/tools/PrepareRelease/SetAllVersions.cs
+++ b/build/tools/PrepareRelease/SetAllVersions.cs
@@ -28,6 +28,9 @@ namespace PrepareRelease
             SynchronizeVersion(
                 "samples/AutomaticTraceIdInjection/SerilogExample/SerilogExample.csproj",
                 DatadogTraceNugetDependencyVersionReplace);
+            SynchronizeVersion(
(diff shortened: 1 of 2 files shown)


lucaspimentel (approved):

lucaspimentel:
Thanks, but I believe we're moving away from recommending users install the Agent and Tracer when deploying Service Fabric services (as these scripts do) and to instead install them on the host/VM/node beforehand. If so, should we remove those parts of the scripts here and only show how to add the env vars to enable tracing? Or remove the whole sample app?

zacharycmontoya:
> Thanks, but I believe we're moving away from recommending users install the Agent and Tracer when deploying Service Fabric services (as these scripts do) and to instead install them on the host/VM/node beforehand. If so, should we remove those parts of the scripts here and only show how to add the env vars to enable tracing? Or remove the whole sample app?

I see I see. I think it would be valuable to keep a sample app in the repo, that way if someone has a cluster they can easily try it out with a functioning configuration. So maybe we just remove the installation bits

lucaspimentel:
Meanwhile, this change doesn't hurt so I'll approve.

lucaspimentel (approved):

PR #1454, opened 2021-05-07
Convert ASP.NET tests to use Snapshot Testing with VerifyTests/Verify

andrewlock:
> Note: The +/- line count is deceiving. If you exclude the snapshot files, there is a massive drop in lines of code, and an even bigger drop in number of characters 😄  e.g. [this file is gone!](https://github.com/DataDog/dd-trace-dotnet/pull/1454/files?file-filters%5B%5D=.cs#diff-e3af8f677c8d42a4c82f17f577da5c02d16e449334550592b6dba3798c8a5fb0) 

This initial PoC creates some helpers for verifying `MockTracerAgent.Span`s and converts our existing ASP.NET integration tests to use snapshot testing. 

Challenges: 
- ~Needed to write a custom scrubber for token-replacement of IDs. This is pretty crude atm but it works. May be enabled OOTB in future Verify versions~ Done in https://github.com/VerifyTests/Verify/pull/351 and https://github.com/VerifyTests/Verify/pull/352!
- Verify doesn't support < .NET 4.6.1. That may mean it's not a viable option for every case given we need to support .NET 4.5.2 🙁 
- Ran into some weird errors due to trying to create filenames with `/` in them. [Solved by stripping the `/`](https://github.com/DataDog/dd-trace-dotnet/blob/de6965cfcf92de8e70452a054ea47cf5b3ccf4c0/test/Datadog.Trace.ClrProfiler.IntegrationTests/Helpers/VerifyHelper.cs#L22-L25). 
- Multiple test classes in a file requires[ a bit of extra verbosity](https://github.com/DataDog/dd-trace-dotnet/blob/de6965cfcf92de8e70452a054ea47cf5b3ccf4c0/test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNet/AspNetMvc4Tests.cs#L68-L70) that won't be required in most cases
- Running integration tests locally is still a royal pain ATM, which makes generating the initial verification text a bit of a pain. Obviously we _should_ be able to do this though, and Nuke will help 😉
- [Had to do a bit of csproj fiddling](https://github.com/DataDog/dd-trace-dotnet/blob/de6965cfcf92de8e70452a054ea47cf5b3ccf4c0/test/Datadog.Trace.ClrProfiler.IntegrationTests/Datadog.Trace.ClrProfiler.IntegrationTests.csproj#L27-L28) to get the nesting to work, again because we have multiple test classes in a file in this case. Overall it works well though: 
- Different architectures apparently have different stack traces, so need to use different snapshots per architecture. We _could_ just strip out the error entirely, but if this is stable, would be a better option I think. [Native support should be coming soon to Verify](https://github.com/VerifyTests/Verify/pull/367).

image

Wins:
- All the benefits of snapshot testing - easier to understand, tests more things, easier to understand failures
- Easily plugs into existing tests by just replacing all the asserts with a single Verify method
- Personally think this is a great option for bread-and-butter integration testing. Makes visualising traces a lot easier, and gives more complete coverage

Remaining questions:
- ~How stable are these spans across platforms?~ well integration tests work, so there we go!
- ~Need to make the environments deterministic - just noticed my system level `DD_ENV=andrew` has made its way into the span!~ Added explicit setting of `DD_ENV` for sample projects in https://github.com/DataDog/dd-trace-dotnet/pull/1454/commits/5c93c356234cc4a14676c770be5e49971758417b. Can be overwritten, but should ensure consistency in other environments.
- ~There appears to be an issue with the solution path scrubbing, which is causing the current failures in the integration tests pipeline. The attributes don't appear to be embedded in the integration test project for some reason. This is probably something to do with our convoluted CI build process (as it works in VS) but I'm not sure why...~ 🤞 Fixed in https://github.com/DataDog/dd-trace-dotnet/pull/1454/commits/de6965cfcf92de8e70452a054ea47cf5b3ccf4c0, it's because `$(SolutionDir)` isn't set when calling `dotnet build` on a _project_ as opposed to the _solution_.
- Need to check how Verify handles reordering in lists/dictionaries


test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNet/AspNetMvc4Tests.cs
```diff
@@ -47,56 +66,90 @@ public AspNetMvc4TestsCallTargetClassic(IisFixture iisFixture, ITestOutputHelper
...
-                tags);
+            var spans = await GetWebServerSpans(path, _iisFixture.Agent, _iisFixture.HttpPort, statusCode);
+
+            var sanitisedPath = VerifyHelper.SanitisePathsForVerify(path);
+
+            var settings = VerifyHelper.GetSpanVerifierSettings(sanitisedPath, statusCode);
+
+            // Overriding the type name here as we have multiple test classes in the file
+            // Ensures that we get nice file nesting in Solution Explorer
+            await Verifier.Verify(spans, settings)
```
zacharycmontoya, line 151: Can we also set up Verify to place all snapshots in its own directory? I'm thinking something like `/test/snapshots/<ProjectName>/**/verified.txt`

test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNet/AspNetWebApi2Tests.cs
```diff
@@ -46,58 +66,94 @@ public AspNetWebApi2TestsCallTargetClassic(IisFixture iisFixture, ITestOutputHel
...
             _iisFixture = iisFixture;
             _iisFixture.TryStartIis(this, classicMode);
+            _testName = nameof(AspNetWebApi2Tests)
+                      + (enableCallTarget ? ".CallSite" : ".CallTarget")
+                      + (classicMode ? ".Classic" : ".Integrated")
+                      + (enableFeatureFlag ? ".NoFF" : ".WithFF")
+                      + (RuntimeInformation.ProcessArchitecture == Architecture.X64 ? ".X64" : ".X86"); // assume that arm is the same
         }
 
+        public static TheoryData<string, int> Data() => new()
```
zacharycmontoya, line 117: This is so nice, we just need to specify the inputs and ensure the output is expected 😍

zacharycmontoya (approved):
LGTM. Let's verify all the things!

tonyredondo (approved):
LGTM nice!

test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNet/AspNetWebApi2Tests.cs
```diff
@@ -46,58 +66,94 @@ public AspNetWebApi2TestsCallTargetClassic(IisFixture iisFixture, ITestOutputHel
...
+    }
+
+    [UsesVerify]
     public abstract class AspNetWebApi2Tests : TestHelper, IClassFixture<IisFixture>
     {
         private readonly IisFixture _iisFixture;
+        private readonly string _testName;
 
-        public AspNetWebApi2Tests(IisFixture iisFixture, ITestOutputHelper output, bool enableCallTarget, bool classicMode)
+        public AspNetWebApi2Tests(IisFixture iisFixture, ITestOutputHelper output, bool enableCallTarget, bool classicMode, bool enableFeatureFlag)
```
lucaspimentel, line 98: Consider renaming parameter `enableFeatureFlag` to something more obvious (_which_ feature flag?). Something like `enableRouteTemplateResourceNames` (but shorter?). Thanks!

test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNet/AspNetWebApi2Tests.cs
```diff
@@ -46,58 +66,94 @@ public AspNetWebApi2TestsCallTargetClassic(IisFixture iisFixture, ITestOutputHel
...
+                SetEnvironmentVariable(ConfigurationKeys.FeatureFlags.RouteTemplateResourceNamesEnabled, "true");
+            }
 
             _iisFixture = iisFixture;
             _iisFixture.TryStartIis(this, classicMode);
+            _testName = nameof(AspNetWebApi2Tests)
+                      + (enableCallTarget ? ".CallSite" : ".CallTarget")
+                      + (classicMode ? ".Classic" : ".Integrated")
+                      + (enableFeatureFlag ? ".NoFF" : ".WithFF")
+                      + (RuntimeInformation.ProcessArchitecture == Architecture.X64 ? ".X64" : ".X86"); // assume that arm is the same
```
lucaspimentel, line 114: Should we include ARM support now so we don't have to come back and fix later?
https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.architecture

lucaspimentel (approved):
Big step forward for our testing suite, thanks! Left a couple of non-blocking comments.

test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNet/AspNetWebApi2Tests.cs
andrewlock: I'm in two minds. Will the call stack be different on Arm? Maybe not? Might not be necessary? I'm inclined to punt until we do run this on ARM (Windows Arm, I guess it's just a matter of time?)

test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNet/AspNetWebApi2Tests.cs
andrewlock: done in https://github.com/DataDog/dd-trace-dotnet/pull/1454/commits/20bec17c173d003e9d861571b0461ff5746bc898

PR #1455, opened 2021-05-07
Fixes dynamically emitted methods signatures

### LINKED ISSUE #1232: Some dynamically emitted methods (such as IsAlreadyLoaded) have incorrect signatures
Examples:

https://github.com/DataDog/dd-trace-dotnet/blob/5178fc60c24f6b1886f69df0963c3b69fa4647cf/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp#L1759-L1772

https://github.com/DataDog/dd-trace-dotnet/blob/5178fc60c24f6b1886f69df0963c3b69fa4647cf/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp#L1799-L1814

Note that I'm not aware of any runtime impact from this for the CLR. However, when attempting to use datadog + https://github.com/microsoft/clrinstrumentationengine, it creates warnings regarding the signature mismatch.

tonyredondo:
Fixes #1232 

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
+++ b/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
@@ -1738,7 +1738,6 @@ HRESULT CorProfiler::GenerateVoidILStartupMethod(const ModuleID module_id,
     IMAGE_CEE_CS_CALLCONV_DEFAULT, // Calling convention
     0,                             // Number of parameters
     ELEMENT_TYPE_VOID,             // Return type
-    ELEMENT_TYPE_OBJECT            // List of parameter types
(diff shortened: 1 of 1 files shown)


andrewlock (approved):
Have we checked if there are any other incorrectly emitted method signatures, other than the ones flagged in the issue?

tonyredondo:
> Have we checked if there are any other incorrectly emitted method signatures, other than the ones flagged in the issue?

Yes I checked all signatures in `cor_profiler.cpp`

lucaspimentel (approved):

colin-higgins (approved):

PR #1457, opened 2021-05-10
Msmq integration

anna-git:
Fixes #

- 3 intercepted methods 
- integration test
- sample program

@DataDog/apm-dotnet


src/Datadog.Trace/Tagging/MsmqTags.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        public string QueueLastModifiedTime { get; set; }
+
+        public string UniqueQueueName { get; set; }
+
+        public string TransactionType { get; set; }
+
+        public string IsTransactionalQueue { get; set; }
+
+        public string InstrumentationName { get; set; }
```
andrewlock, line 44: This is a lot of tags, do we definitely want them all? I've learnt (the hard way!) that each tag has a significant performance impact! IMO, we could probably lose:

- `QueueLabel`
-  `QueueLastModifiedTime`
- `IsTransactionalQueue`
- `UniqueQueueName` (i.e. either keep `formatName` or `queueName`, depending on which is more useful!)
- `TransactionType`

Additionally `InstrumentationName` is normally created as a `readonly` field in the Tags, rather than specified at creation 🙂

src/Datadog.Trace/Tagging/MsmqTags.cs
```diff
@@ -0,0 +1,48 @@
...
+               new Property<MsmqTags, string>(Trace.Tags.MsmqQueueLastModifiedTime, t => t.QueueLastModifiedTime, (t, v) => t.QueueLastModifiedTime = v),
+               new Property<MsmqTags, string>(Trace.Tags.MsmqQueueUniqueName, t => t.UniqueQueueName, (t, v) => t.UniqueQueueName = v),
+               new Property<MsmqTags, string>(Trace.Tags.MsmqIsTransactionalQueue, t => t.IsTransactionalQueue, (t, v) => t.IsTransactionalQueue = v),
+               new Property<MsmqTags, string>(Trace.Tags.MsmqTransactionType, t => t.TransactionType, (t, v) => t.TransactionType = v));
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MsmqTags"/> class.
+        /// </summary>
+        /// <param name="spanKind">kind of span</param>
+        public MsmqTags(string spanKind) => SpanKind = spanKind;
```
andrewlock, line 25: You'll need to create a default constructor version of this too, as there's a unit tests that checks for its existence

test/Datadog.Trace.ClrProfiler.IntegrationTests/MsmqTests.cs
```diff
@@ -0,0 +1,108 @@
...
+            using var agent = new MockTracerAgent(agentPort);
+            using var processResult = RunSampleAndWaitForExit(agent.Port, arguments: $"5 5");
+            Assert.True(processResult.ExitCode >= 0, $"Process exited with code {processResult.ExitCode} and exception: {processResult.StandardError}");
+
+            var spans = agent.WaitForSpans(totalTransactions);
+            Assert.True(spans.Count >= totalTransactions, $"Expecting at least {totalTransactions} spans, only received {spans.Count}");
+            var msmqSpans = spans.Where(span => string.Equals(span.Service, ExpectedServiceName, StringComparison.OrdinalIgnoreCase));
+            foreach (var span in msmqSpans)
+            {
+                Assert.Equal(SpanTypes.Queue, span.Type);
```
andrewlock, line 52: Just a minor one, but we've shifted towards using FluentAssertions to make this easier to read. e.g. 
```suggestion
                span.Type.Should().Be(SpanTypes.Queue);
```

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Msmq/MsmqCommon.cs
```diff
@@ -0,0 +1,58 @@
...
+namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Msmq
+{
+    internal class MsmqCommon
+    {
+        internal const string IntegrationName = nameof(IntegrationIds.Msmq);
+        internal static readonly IntegrationInfo IntegrationId = IntegrationRegistry.GetIntegrationInfo(IntegrationName);
+
+        private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(MsmqCommon));
+
+        internal static Scope CreateScope(Tracer tracer, string command, string spanKind, string queueName, string formatName, string queueLabel, DateTime queueLastModifiedTime, bool withinTransaction, string transactionType, bool transactionalQueue, out MsmqTags tags)
```
andrewlock, line 18: This is a matter of preference, but could we pass an `IMessageQueue` in here instead of the strings? Also I don't think the out parameter is used anywhere?

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Msmq/IMessageQueue.cs
```diff
@@ -0,0 +1,146 @@
...
+    /// </summary>
+    public interface IMessageQueue
+    {
+        /// <summary>
+        ///     gets a value indicating whether a cache of connections will be maintained
+        ///     by the application.
+        /// Returns:
+        ///   true to create and use a connection cache; otherwise, false.
+        /// </summary>
+        bool EnableConnectionCache { get; }
```
andrewlock, line 16: I don't think all of these properties are used in the integration? We have to generate code for each extra property, so AFAIK it's best to only specify the properties you actually need

andrewlock (commented):
Looking good! 🙂 I left a few comments mostly around whether we want to include all of the queue details as tags for performance reasons

src/Datadog.Trace/Tagging/MsmqTags.cs
anna-git: Oh ok good to know! will remove those and keep format which gives the host info.

In msmq, if we send a message within a transaction to a non transactional queue, or the opposite, sending a message outside of a transaction context to a transactional queue, the message is never sent but the methods are called and nothing crashes, which is kind of an odd behavior I got stuck on. So I thought it might be useful to a customer to see, in the operation msmq.send, for example: 

is_transactional_queue > False
transaction_type> Single (there is Auto, None, Single)

but it's just a remark, I'm not sure of their requirements here :)

andrewlock: If unique name includes Path (as it seems to from that example), then that will probably be sufficient IMO. (Spoken as someone that has never used MSMQ!)

integrations.json
```diff
@@ -612,6 +612,86 @@
       }
     ]
   },
+  {
```
kevingosse, line 615: Please remember to re-generate this file before merging, as we are in version 1.26.3 now

src/Datadog.Trace.ClrProfiler.Managed.Core/ClrNames.cs
```diff
@@ -38,5 +40,11 @@ internal static class ClrNames
         public const string GenericParameterTask = "System.Threading.Tasks.Task`1<T>";
         public const string ObjectTask = "System.Threading.Tasks.Task`1<System.Object>";
         public const string Int32Task = "System.Threading.Tasks.Task`1<System.Int32>";
+
+        public const string MsmqCursorHandle = "System.Messaging.Interop.CursorHandle";
+        public const string MsmqMessageQueueTransaction = "System.Messaging.MessageQueueTransaction";
+        public const string MsmqMessageQueueTransactionType = "System.Messaging.MessageQueueTransactionType";
+        public const string MsmqMessagePropertyFilter = "System.Messaging.MessagePropertyFilter";
+        public const string MsmqMessage = "System.Messaging.Message";
```
kevingosse, line 48: As much as possible, we should put in ClrNames only base types that may be useful for other integrations. For the MSMQ ones, you should declare them in the same folder as the integration (for instance in `MsmqConstants`)

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Msmq/MessageQueue_ReceiveCurrent_Integration.cs
```diff
@@ -0,0 +1,67 @@
...
+    ReturnTypeName = ClrNames.MsmqMessage,
+    ParameterTypeNames = new[] { ClrNames.TimeSpan, ClrNames.Int32, ClrNames.MsmqCursorHandle, ClrNames.MsmqMessagePropertyFilter, ClrNames.MsmqMessageQueueTransaction, ClrNames.MsmqMessageQueueTransactionType },
+    MinimumVersion = "4.0.0",
+    MaximumVersion = "4.*.*",
+    IntegrationName = IntegrationName)]
+    public class MessageQueue_ReceiveCurrent_Integration
+    {
+        private const string CommandPeek = "msmq.peek";
+        private const string CommandConsume = "msmq.consume";
+        private const string IntegrationName = nameof(IntegrationIds.Msmq);
```
kevingosse, line 30: NIT: you could move them to `MsmqConstants` (same for similar constants in other integration files)

src/Datadog.Trace/Configuration/IntegrationIds.cs
```diff
@@ -13,6 +13,7 @@ internal enum IntegrationIds
         AspNetMvc,
         AspNetWebApi2,
         GraphQL,
+        Msmq,
```
kevingosse, line 16: I don't think it would cause any version conflict, but I'd feel safer if you declared it last, so that the numerical id of the existing integrations don't change

src/Datadog.Trace/Tagging/MsmqTags.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// <inheritdoc/>
+        public override string SpanKind { get; }
+
+        public string UniqueQueueName { get; set; }
+
+        public string MessageWithTransaction { get; set; }
+
+        public string InstrumentationName => "msmq";
+
+        public string IsTransactionalQueue { get; set; }
```
kevingosse, line 37: I wonder if metrics would be a better storage than string for boolean value... I don't know if we have booleans elsewhere. @lucaspimentel thoughts?

test/test-applications/integrations/Samples.Msmq/Program.cs
```diff
@@ -0,0 +1,76 @@
+using System;
+using System.Linq;
+using System.Messaging;
+
+namespace Samples.Msmq
+{
+    public class Program
+    {
+        const string PrivateTransactionalQueuePath = ".\\Private$\\private-transactional-queue";
```
kevingosse, line 9: I'm curious to see if that will work in the CI agent. That would be a pleasant surprise if the MSMQ service is activated.

kevingosse (changes requested):
Great stuff 👍  Only minor changes needed

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Msmq/MsmqCommon.cs
```diff
@@ -0,0 +1,57 @@
...
+namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Msmq
+{
+    internal class MsmqCommon
+    {
+        internal const string IntegrationName = nameof(IntegrationIds.Msmq);
+        internal static readonly IntegrationInfo IntegrationId = IntegrationRegistry.GetIntegrationInfo(IntegrationName);
+
+        private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(MsmqCommon));
+
+        internal static Scope CreateScope(Tracer tracer, string command, string spanKind, IMessageQueue messageQueue, bool? messagePartofTransaction = null)
```
tonyredondo, line 18: If we use `IMessageQueue` as parameter then the proxy struct will be boxed. You can use the generic constraint as in the integration.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Msmq/MsmqCommon.cs
```diff
@@ -0,0 +1,57 @@
...
+            }
+
+            Scope scope = null;
+
+            try
+            {
+                var tags = new MsmqTags(spanKind)
+                {
+                    Command = command,
+                    IsTransactionalQueue = messageQueue.Transactional.ToString(),
```
tonyredondo, line 33: I don't know if the tags support `boolean` types directly but maybe we can avoid the string allocation here. cc. @kevingosse

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Msmq/MsmqCommon.cs
```diff
@@ -0,0 +1,57 @@
...
+                    UniqueQueueName = messageQueue.FormatName,
+                };
+                if (messagePartofTransaction.HasValue)
+                {
+                    tags.MessageWithTransaction = messagePartofTransaction.ToString();
```
tonyredondo, line 38: same as my previous comment.

tonyredondo (changes requested):
Looking good so far!, I've just left a couple of comments performance related.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Msmq/MsmqCommon.cs
kevingosse: Yeah I added another comment wondering if we should store them as metrics instead. That said, I don't think `Boolean.ToString()` allocates, I need to check but I think the string is cached internally

kevingosse: https://referencesource.microsoft.com/#mscorlib/system/boolean.cs,90

src/Datadog.Trace/Tagging/MsmqTags.cs
lucaspimentel: What's the purpose of this value? If we want to show it in the APM UI, it needs to be a tag. We don't show span "metrics" to users.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Msmq/MsmqCommon.cs
lucaspimentel: Tags only support strings. Metrics support doubles, but they are not displayed in the UI. Anything we want to show to users as span metadata needs to be a string tag.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Msmq/MsmqCommon.cs
```diff
@@ -0,0 +1,57 @@
...
+using Datadog.Trace.Configuration;
+using Datadog.Trace.DuckTyping;
+using Datadog.Trace.Logging;
+using Datadog.Trace.Tagging;
+
+namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Msmq
+{
+    internal class MsmqCommon
+    {
+        internal const string IntegrationName = nameof(IntegrationIds.Msmq);
```
lucaspimentel, line 13: Can you reuse the value in `MsmqConstants`?

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Msmq/MessageQueue_Purge_Integration.cs
```diff
@@ -0,0 +1,52 @@
...
+    TypeName = "System.Messaging.MessageQueue",
+    MethodName = "Purge",
+    ReturnTypeName = ClrNames.Void,
+    MinimumVersion = "4.0.0",
+    MaximumVersion = "4.*.*",
+    IntegrationName = IntegrationName)]
+    public class MessageQueue_Purge_Integration
+    {
+        private const string Command = "msmq.purge";
+        private const string IntegrationName = nameof(IntegrationIds.Msmq);
```
lucaspimentel, line 23: Can you reuse the value in `MsmqConstants.IntegrationName`? Consider also moving `"msmq.purge"` there.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Msmq/MessageQueue_SendInternal_Integration.cs
```diff
@@ -0,0 +1,55 @@
...
+    TypeName = "System.Messaging.MessageQueue",
+    MethodName = "SendInternal",
+    ReturnTypeName = ClrNames.Void,
+    ParameterTypeNames = new[] { ClrNames.Object, ClrNames.MsmqMessageQueueTransaction, ClrNames.MsmqMessageQueueTransactionType },
+    MinimumVersion = "4.0.0",
+    MaximumVersion = "4.*.*",
+    IntegrationName = IntegrationName)]
+    public class MessageQueue_SendInternal_Integration
+    {
+        private const string IntegrationName = nameof(IntegrationIds.Msmq);
```
lucaspimentel, line 22: Can you reuse the value in `MsmqConstants`?

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Msmq/MsmqCommon.cs
anna-git: thank you for your comments! so in the end is it ok to use bool.ToString, it doesnt seem like it's allocating memory but using constants right?

kevingosse: Sounds about right. Thanks for the clarification Lucas.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Msmq/MsmqCommon.cs
```diff
@@ -0,0 +1,57 @@
...
+                    tags.MessageWithTransaction = messagePartofTransaction.ToString();
+                }
+
+                var serviceName = tracer.Settings.GetServiceName(tracer, MsmqConstants.ServiceName);
+
+                scope = tracer.StartActiveWithTags(MsmqConstants.OperationName, serviceName: serviceName, tags: tags);
+
+                var span = scope.Span;
+                span.Type = SpanTypes.Queue;
+                span.ResourceName = command;
```
lucaspimentel, line 47: What will the resource name look like? I tried looking in the tests and we're not asserting that there (we probably should).

lucaspimentel: If users are sending message to different queues, we want those grouped by queue name in the UI. Should we appent the Path to the resource name, too. Something like `{command} {path}`. Similar to http requests which use `{verb} {path}` (e.g. `GET /users/edit/{id}`).

lucaspimentel:
@anna-git, can we close this PR in favor of #1463?

anna-git:
> @anna-git, can we close this PR in favor of #1463?

yes please thanks :)

PR #1462, opened 2021-05-12
Enable classic mode in IIS tests

kevingosse:
Extend the test matrix to run the IIS test in classic and integrated modes. We can't configure that from the IIS Express command line, so I switched to a templated applicationHost.config file.

I tested first on 1.26.2 to confirm the tests were crashing: https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=64259&view=results

### DIFF
--- a/test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNet/AspNetWebFormsTests.cs
+++ b/test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNet/AspNetWebFormsTests.cs
@@ -1,6 +1,4 @@
 #if NET461 || NET452
-#pragma warning disable SA1402 // File may only contain a single class
-#pragma warning disable SA1649 // File name must match first type name
 
 using System;
 using System.Linq;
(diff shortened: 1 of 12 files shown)


test/Datadog.Trace.ClrProfiler.IntegrationTests/Helpers/IisFixture.cs
```diff
@@ -1,30 +1,31 @@
...
 
         public int HttpPort { get; private set; }
 
-        public void TryStartIis(TestHelper helper)
+        public void TryStartIis(TestHelper helper, bool classicMode)
         {
             lock (this)
             {
-                if (_iisExpress == null)
+                if (_iisExpress.Process == null)
```
zacharycmontoya, line 21: Does this allow the two sets of tests to run at the same time? I'd expect only one invocation of this to happen so only either an app pool in Integrated Mode is started or an app pool in Classic Mode is started

kevingosse: The configuration file is copied to a temporary name, so it shouldn't cause any issue with multiple concurrent executions (at least, no more issues than it did before)

zacharycmontoya: I'm just surprised that we can execute this multiple times. It looks like if I call `TryStartIis(helper, false)` followed by `TryStartIis(helper, true)` then we'd only start IIS Express once in Integrated Mode so we wouldn't be testing Classic Mode

test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNet/AspNetWebFormsTests.cs
```diff
@@ -23,7 +23,7 @@ public AspNetWebFormsTests(IisFixture iisFixture, ITestOutputHelper output)
             SetServiceVersion("1.0.0");
 
             _iisFixture = iisFixture;
-            _iisFixture.TryStartIis(this);
+            _iisFixture.TryStartIis(this, classicMode: false);
```
lucaspimentel, line 26: We don't want to run this test in classic mode as well?

kevingosse: This test uses OWIN, so it crashes in classic mode :(

test/Datadog.Trace.ClrProfiler.IntegrationTests/Helpers/IisFixture.cs
kevingosse: Yeah but we don't 😛 You have for instance `public class AspNetWebApi2WithFeatureFlagCallTargetIntegrated` and `public class AspNetWebApi2WithFeatureFlagCallTargetClassic`. They're separate classes, so they don't share the instance of `IisFixture`

andrewlock: The aspnet tests use an `IClassFixture<>`. That means the `IISFixture` is only reused for [within-class tests (i.e. theories)](https://xunit.net/docs/shared-context#class-fixture). As we change mode in different classes, this should work as far as I can tell.

That said, if possible (I'm not sure it is), it might be nice to add a check in the web application itself to assert it's running in the correct mode?

andrewlock (approved):

test/Datadog.Trace.ClrProfiler.IntegrationTests/Helpers/IisFixture.cs
zacharycmontoya: Ah thanks, my understanding of the class fixtures wasn't accurate. Thanks for clarifying!

zacharycmontoya (approved):

PR #1463, opened 2021-05-12
Msmq  integration, taking into account all comments from PR #1457.

anna-git:
Fixes #

Taken into accounts all comments from #1457  because I couldn't rebase successfully despite all attempts

@DataDog/apm-dotnet

### DIFF
--- a/Datadog.Trace.sln
+++ b/Datadog.Trace.sln
@@ -438,6 +438,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Samples.AspNetCoreRazorPage
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "FeatureTracking", "tools\FeatureTracking\FeatureTracking.csproj", "{2829C811-66EE-4B81-96CA-2ACB00B8E813}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Samples.Msmq", "test\test-applications\integrations\Samples.Msmq\Samples.Msmq.csproj", "{662B587F-97B5-4CEF-ABF9-6C76A6DBD29E}"
+EndProject
 Global
 	GlobalSection(SharedMSBuildProjectFiles) = preSolution
 		src\Datadog.Trace.Ci.Shared\Datadog.Trace.Ci.Shared.projitems*{1f146d40-8b21-4630-8049-14fa4bbbb217}*SharedItemsImports = 5
@@ -1603,6 +1605,16 @@ Global
 		{2829C811-66EE-4B81-96CA-2ACB00B8E813}.Release|x64.Build.0 = Release|Any CPU
 		{2829C811-66EE-4B81-96CA-2ACB00B8E813}.Release|x86.ActiveCfg = Release|Any CPU
 		{2829C811-66EE-4B81-96CA-2ACB00B8E813}.Release|x86.Build.0 = Release|Any CPU
+		{662B587F-97B5-4CEF-ABF9-6C76A6DBD29E}.Debug|Any CPU.ActiveCfg = Debug|x86
+		{662B587F-97B5-4CEF-ABF9-6C76A6DBD29E}.Debug|x64.ActiveCfg = Debug|x64
+		{662B587F-97B5-4CEF-ABF9-6C76A6DBD29E}.Debug|x64.Build.0 = Debug|x64
+		{662B587F-97B5-4CEF-ABF9-6C76A6DBD29E}.Debug|x86.ActiveCfg = Debug|x86
+		{662B587F-97B5-4CEF-ABF9-6C76A6DBD29E}.Debug|x86.Build.0 = Debug|x86
+		{662B587F-97B5-4CEF-ABF9-6C76A6DBD29E}.Release|Any CPU.ActiveCfg = Release|x86
+		{662B587F-97B5-4CEF-ABF9-6C76A6DBD29E}.Release|x64.ActiveCfg = Release|x64
+		{662B587F-97B5-4CEF-ABF9-6C76A6DBD29E}.Release|x64.Build.0 = Release|x64
+		{662B587F-97B5-4CEF-ABF9-6C76A6DBD29E}.Release|x86.ActiveCfg = Release|x86
+		{662B587F-97B5-4CEF-ABF9-6C76A6DBD29E}.Release|x86.Build.0 = Release|x86
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
@@ -1727,6 +1739,7 @@ Global
 		{34B67004-7249-4EF1-8E12-6E6DA37EA6BE} = {498A300E-D036-49B7-A43D-821D1CAF11A5}
 		{1B9E6BF4-9D48-4988-9945-248096119E46} = {BAF8F246-3645-42AD-B1D0-0F7EAFBAB34A}
 		{2829C811-66EE-4B81-96CA-2ACB00B8E813} = {5D8E1F81-B820-4736-B797-271B0FE787EE}
+		{662B587F-97B5-4CEF-ABF9-6C76A6DBD29E} = {BAF8F246-3645-42AD-B1D0-0F7EAFBAB34A}
 	EndGlobalSection
 	GlobalSection(ExtensibilityGlobals) = postSolution
 		SolutionGuid = {160A1D00-1F5B-40F8-A155-621B4459D78F}
--- a/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Msmq/MessageQueue_Purge_Integration.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Msmq/MessageQueue_Purge_Integration.cs
@@ -1,0 +1,51 @@
+using System;
+using Datadog.Trace.ClrProfiler.CallTarget;
+using Datadog.Trace.Configuration;
+using Datadog.Trace.Logging;
+using Datadog.Trace.Tagging;
+
+namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Msmq
+{
+    /// <summary>
+    /// Msmq calltarget instrumentation
+    /// </summary>
+    [InstrumentMethod(
+    AssemblyName = "System.Messaging",
+    TypeName = "System.Messaging.MessageQueue",
+    MethodName = "Purge",
+    ReturnTypeName = ClrNames.Void,
+    MinimumVersion = "4.0.0",
+    MaximumVersion = "4.*.*",
+    IntegrationName = MsmqConstants.IntegrationName)]
+    public class MessageQueue_Purge_Integration
+    {
+        private const string Command = MsmqConstants.MsmqPurgeCommand;
+
+        /// <summary>
+        /// OnMethodBegin callback
+        /// </summary>
+        /// <typeparam name="TMessageQueue">Message queue</typeparam>
+        /// <param name="instance">Instance value, aka `this` of the instrumented method.</param>
+        /// <returns>Calltarget state value</returns>
+        public static CallTargetState OnMethodBegin<TMessageQueue>(TMessageQueue instance)
+            where TMessageQueue : IMessageQueue
+        {
+            var scope = MsmqCommon.CreateScope(Tracer.Instance, Command, SpanKinds.Client, instance);
+            return new CallTargetState(scope);
+        }
+
+        /// <summary>
+        /// OnMethodEnd callback
+        /// </summary>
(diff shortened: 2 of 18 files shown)


Datadog.Trace.sln
```diff
@@ -428,15 +428,17 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "crank", "crank", "{E3FB283A
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "DuplicateTypeProxy", "test\test-applications\regression\DuplicateTypeProxy\DuplicateTypeProxy.csproj", "{34B67004-7249-4EF1-8E12-6E6DA37EA6BE}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Samples.AspNetCoreRazorPages", "test\test-applications\integrations\Samples.AspNetCoreRazorPages\Samples.AspNetCoreRazorPages.csproj", "{1B9E6BF4-9D48-4988-9945-248096119E46}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Samples.AspNetCoreRazorPages", "test\test-applications\integrations\Samples.AspNetCoreRazorPages\Samples.AspNetCoreRazorPages.csproj", "{1B9E6BF4-9D48-4988-9945-248096119E46}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "FeatureTracking", "tools\FeatureTracking\FeatureTracking.csproj", "{2829C811-66EE-4B81-96CA-2ACB00B8E813}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "FeatureTracking", "tools\FeatureTracking\FeatureTracking.csproj", "{2829C811-66EE-4B81-96CA-2ACB00B8E813}"
```
lucaspimentel, line 433: Yay, I was about to push a PR to fix these two project GUIDs.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Msmq/MsmqCommon.cs
```diff
@@ -0,0 +1,54 @@
...
+                    tags.MessageWithTransaction = messagePartofTransaction.ToString();
+                }
+
+                var serviceName = tracer.Settings.GetServiceName(tracer, MsmqConstants.ServiceName);
+
+                scope = tracer.StartActiveWithTags(MsmqConstants.OperationName, serviceName: serviceName, tags: tags);
+
+                var span = scope.Span;
+                span.Type = SpanTypes.Queue;
+                span.ResourceName = command;
```
lucaspimentel, line 44: Copying my comments from [here](https://github.com/DataDog/dd-trace-dotnet/pull/1457/files#r631160043):
> What will the resource name look like? I tried looking in the tests and we're not asserting that there (we probably should).
>
>If users are sending messages to different queues, we want those grouped by queue name in the UI. Should we append the Path to the current resource name?. Something like `{command} {path}`? Similar to http requests which use `{verb} {path}` (e.g. `GET /users/edit/{id}`).

anna-git: oh I see now, yes 👍

test/test-applications/integrations/Samples.Msmq/Samples.Msmq.csproj
```diff
@@ -0,0 +1,16 @@
...
+    <!-- override to remove core -->
+    <TargetFrameworks>net452;net461</TargetFrameworks>
+    <OutputType>Exe</OutputType>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|net452|x64'">
+    <WarningLevel>4</WarningLevel>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|net452|x86'">
+    <WarningLevel>4</WarningLevel>
+  </PropertyGroup>
```
lucaspimentel, line 12: I'm guessing VS added this. We should probably remove these lines since they only apply to `Debug` and `net452`. If we want `<WarningLevel>4</WarningLevel>` for all configurations, we can add it to the `<PropertyGroup>` above.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Msmq/MsmqCommon.cs
lucaspimentel: I saw you check the resource name in the tests, thanks! What do you think of the second part of the comment (adding the path to the resource name)?

anna-git: yes so the we would have a resource name like that then:

resource_name:"msmq.purge DIRECT=OS:desktop-duevvft\\Private$\\private-nontransactional-queue"
resource_name:"msmq.send DIRECT=OS:desktop-duevvft\\Private$\\private-transactional-queue"

I could bring back the Path property (and not the whole name with host) to get something shorter without the host name

resource_name:"msmq.purge \\Private$\\private-nontransactional-queue"
resource_name:"msmq.send \\Private$\\private-transactional-queue"

image

lucaspimentel: I'm not familiar with MSMQ and what users would want to see in the resource name, but I'm leaning towards Path only instead of the whole name.

test/Datadog.Trace.ClrProfiler.IntegrationTests/MsmqTests.cs
```diff
@@ -0,0 +1,111 @@
...
+    {
+        private const string ExpectedServiceName = "Samples.Msmq-msmq";
+
+        public MsmqTests(ITestOutputHelper output)
+            : base("Msmq", output)
+        {
+            SetServiceVersion("1.0.0");
+        }
+
+        [Trait("Category", "EndToEnd")]
```
zacharycmontoya, line 22:
```suggestion
        [Trait("Category", "EndToEnd")]
        [Trait("RunOnWindows", "True")]
```

Sorry this isn't intuitive, but the `RunOnWindows=true` is a filter we use when deciding which integration tests to run on Windows

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Msmq/MessageQueue_Purge_Integration.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// <summary>
+        /// OnMethodBegin callback
+        /// </summary>
+        /// <typeparam name="TMessageQueue">Message queue</typeparam>
+        /// <param name="instance">Instance value, aka `this` of the instrumented method.</param>
+        /// <returns>Calltarget state value</returns>
+        public static CallTargetState OnMethodBegin<TMessageQueue>(TMessageQueue instance)
+            where TMessageQueue : IMessageQueue
+        {
+            var scope = MsmqCommon.CreateScope(Tracer.Instance, Command, SpanKinds.Producer, instance);
```
zacharycmontoya, line 33: Since the purge action isn't directly producing or consuming a message, maybe we can use `SpanKinds.Client` instead of `SpanKinds.Producer` here

src/Datadog.Trace/Tags.cs
```diff
@@ -318,6 +318,26 @@ public static class Tags
...
+
+        /// <summary>
+        /// The name of the Msmq queue the message was published to, containing host name and path.
+        /// </summary>
+        public const string MsmqQueueUniqueName = "msmq.queue_unique_name";
+
+        /// <summary>
+        /// A boolean indicating if it's part of a transaction.
+        /// </summary>
+        public const string MsmqMessageWithTransaction = "msmq.message_with_transaction";
```
zacharycmontoya, line 339: What do you think of the following tag names?
- `msmq.queue.name`
- `msmq.queue.transactional`
- `msmq.message.transactional`

test/Datadog.Trace.ClrProfiler.IntegrationTests/PackageVersionsComprehensive.g.cs
```diff
@@ -73,6 +73,7 @@ public class PackageVersionsComprehensive
                 new object[] { "2.12.0" },
                 new object[] { "2.12.1" },
                 new object[] { "2.12.2" },
+                new object[] { "2.12.3" },
```
lucaspimentel, line 76: The changes in this file are unrelated to this new instrumentation. Can you leave them out of this PR? (feel free to push a separate PR to update these)

test/Datadog.Trace.ClrProfiler.IntegrationTests/PackageVersionsLatestMinors.g.cs
```diff
@@ -40,7 +40,7 @@ public class PackageVersionsLatestMinors
                 new object[] { "2.9.3" },
                 new object[] { "2.10.4" },
                 new object[] { "2.11.6" },
-                new object[] { "2.12.2" },
+                new object[] { "2.12.3" },
```
lucaspimentel, line 43: The changes in this file are unrelated to this new instrumentation. Can you leave them out of this PR?

test/Datadog.Trace.ClrProfiler.IntegrationTests/PackageVersionsComprehensive.g.cs
anna-git: reverted them:)

anna-git:
did a rebase to get the fix for the build

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Msmq/MsmqCommon.cs
anna-git: In the end used command - path 
without name of the machine

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Msmq/MsmqCommon.cs
```diff
@@ -0,0 +1,54 @@
...
+
+            Scope scope = null;
+
+            try
+            {
+                var tags = new MsmqTags(spanKind)
+                {
+                    Command = command,
+                    IsTransactionalQueue = messageQueue.Transactional.ToString(),
+                    Path = messageQueue.Path,
```
zacharycmontoya, line 31: For the Path tag (not the resource name), I think it would be desirable to include the machine name, that way users can filter further on that. Is that reasonable @lucaspimentel or should we not include the machine name on this tag?

zacharycmontoya (approved):
LGTM with one small question about a tag's value 👍🏼

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Msmq/MsmqCommon.cs
```diff
@@ -0,0 +1,54 @@
+using System;
+using System.Text;
+using Datadog.Trace.Configuration;
+using Datadog.Trace.Logging;
+using Datadog.Trace.Tagging;
+
+namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Msmq
+{
+    internal class MsmqCommon
```
tonyredondo, line 9: nit: `internal static`

tonyredondo (approved):
Nice job you did here, just one nitpicking comment.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Msmq/MsmqCommon.cs
anna-git: The machine name property doesn't give me the machine name but just a "."  so does seem useless..
The only way to get it is to use the format name property of the queue. So path and format name look almost the same.. 
Path: Private$\myQueue3
FormatName: DIRECT=OS:desktop-duevvft\Private$\myQueue3
we need the path for grouping by resource name: 
             `  span.ResourceName = $"{command} {messageQueue.Path}";`

I can try to isolate the machine name by removing the "Path" part from the format name, it should cover all sorts of cases?

kevingosse (approved):
LGTM 👍

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Msmq/MsmqCommon.cs
zacharycmontoya: Got it. With that info, then I would recommend keeping what you have now so we don't spend extra cycles splitting and allocating new strings 😄

PR #1468, opened 2021-05-13
Clean up project files (reduce conditionals and reference assemblies)

lucaspimentel:
#### Remove most `<TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">`

Most checks for `'$(OS)' == 'Windows_NT'` have been redundant since we added references to `Microsoft.NETFramework.ReferenceAssemblies`. These references assemblies allow the .NET SDK to target .NET Framework (`net45`, `net461`, etc) even if the .NET Framework SDK is not installed (for example, on Linux and macOS).

#### Remove `<PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies">`

.NET SDK 5.0 automatically includes the .NET Framework reference assemblies if needed, so we no longer need the explicit reference to `Microsoft.NETFramework.ReferenceAssemblies`.

#### Remove `<PublicSign Condition="...">true</...>`

.NET SDK 3.0 added support for signing assemblies on non-Windows OS. This [public signing](https://github.com/dotnet/runtime/blob/main/docs/project/public-signing.md) workaround to sign on Linux is a holdover from the 2.x era.

#### Remove `<TargetFrameworks Condition="...">$(TargetFrameworks);net5.0</...>`

We now support .NET 5 everywhere and require it to build the solution. No need to check the MSBuild version anymore.

### DIFF
--- a/test/Directory.Build.props
+++ b/test/Directory.Build.props
@@ -2,9 +2,9 @@
   <Import Project="..\Directory.Build.props" />
 
   <PropertyGroup>
-    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">net452;net461;netcoreapp2.1;netcoreapp3.0;netcoreapp3.1</TargetFrameworks>
-    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netcoreapp2.1;netcoreapp3.0;netcoreapp3.1</TargetFrameworks>
-    <TargetFrameworks Condition="$(MSBuildVersion) >= 16.8.0">$(TargetFrameworks);net5.0</TargetFrameworks>
+    <!-- only run .NET Framework tests on Windows -->
+    <TargetFrameworks Condition="'$(TargetFrameworks)' == '' and '$(OS)' == 'Windows_NT'">net452;net461;netcoreapp2.1;netcoreapp3.0;netcoreapp3.1;net5.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(TargetFrameworks)' == ''">netcoreapp2.1;netcoreapp3.0;netcoreapp3.1;net5.0</TargetFrameworks>
 
     <!-- Hide warnings for EOL .NET Core targets (e.g. netcoreapp3.0) -->
     <CheckEolTargetFramework>false</CheckEolTargetFramework>
--- a/test/Directory.Build.props
+++ b/test/Directory.Build.props
@@ -22,10 +22,5 @@
   <ItemGroup>
     <Compile Include="$(MSBuildThisFileDirectory)GlobalSuppressions.cs" Link="GlobalSuppressions.test.cs" />
   </ItemGroup>
-  
-  <ItemGroup>
-    <!-- reference assemblies let us target .NET Framework without the SDK (for example, on non-Windows) -->
-    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0" PrivateAssets="all" />
-  </ItemGroup>
 
 </Project>
--- a/test/test-applications/integrations/Directory.Build.props
+++ b/test/test-applications/integrations/Directory.Build.props
@@ -1,3 +1,0 @@
-<Project>
-  <Import Project="..\Directory.Build.props" />
-</Project>
--- a/test/test-applications/integrations/Samples.Dapper/Samples.Dapper.csproj
+++ b/test/test-applications/integrations/Samples.Dapper/Samples.Dapper.csproj
@@ -3,8 +3,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
      <!-- override to remove net452 -->
-    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">net461;netcoreapp2.1;netcoreapp3.0;netcoreapp3.1</TargetFrameworks>
-    <TargetFrameworks Condition="$(MSBuildVersion) >= 16.8.0 AND !$(TargetFrameworks.Contains('net5.0'))">$(TargetFrameworks);net5.0</TargetFrameworks>
+    <TargetFrameworks>net461;netcoreapp2.1;netcoreapp3.0;netcoreapp3.1;net5.0</TargetFrameworks>
   </PropertyGroup>
 
   <ItemGroup>
--- a/test/test-applications/integrations/Samples.RabbitMQ/Samples.RabbitMQ.csproj
+++ b/test/test-applications/integrations/Samples.RabbitMQ/Samples.RabbitMQ.csproj
@@ -5,10 +5,7 @@
     <DefineConstants Condition="$(ApiVersion) >= 6.0.0">$(DefineConstants);RABBITMQ_6_0</DefineConstants>
 
     <!-- override to remove net452 -->
-    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">net461;netcoreapp2.1;netcoreapp3.0;netcoreapp3.1</TargetFrameworks>
-    <!-- <TargetFrameworks Condition="'$(OS)' == 'Windows_NT' AND $(ApiVersion) &lt; 6.0.0">$(TargetFrameworks);net452</TargetFrameworks> -->
-    <TargetFrameworks Condition="$(MSBuildVersion) >= 16.8.0 AND !$(TargetFrameworks.Contains('net5.0'))">$(TargetFrameworks);net5.0</TargetFrameworks>
-    <DefineConstants Condition="$(ApiVersion) >= 6.0.0">$(DefineConstants);RABBITMQ_6_0</DefineConstants>
+    <TargetFrameworks>net461;netcoreapp2.1;netcoreapp3.0;netcoreapp3.1;net5.0</TargetFrameworks>
 
     <!-- Required to build multiple projects with the same Configuration|Platform -->
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
@@ -18,7 +15,7 @@
   <ItemGroup>
     <ProjectReference Include="..\..\..\..\src\Datadog.Trace\Datadog.Trace.csproj" />
   </ItemGroup>
-  
+
   <ItemGroup>
     <PackageReference Include="RabbitMQ.Client" Version="$(ApiVersion)" />
   </ItemGroup>
--- a/test/test-applications/integrations/dependency-libs/Samples.NoMultiLoader.Deps/Samples.NoMultiLoader.Deps.csproj
+++ b/test/test-applications/integrations/dependency-libs/Samples.NoMultiLoader.Deps/Samples.NoMultiLoader.Deps.csproj
@@ -1,11 +1,10 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">net45;net461;netstandard2.0;netcoreapp3.1</TargetFrameworks>
-    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netstandard2.0;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>net45;net461;netstandard2.0;netcoreapp3.1</TargetFrameworks>
     <LangVersion>latest</LangVersion>
   </PropertyGroup>
-  
+
   <ItemGroup Condition="$(TargetFramework.StartsWith('net4'))">
     <Reference Include="System.Data" />
   </ItemGroup>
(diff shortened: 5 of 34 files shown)


build/docker/build.sh
```diff
@@ -42,7 +42,7 @@ fi
 
 dotnet publish -f $publishTargetFramework -c $buildConfiguration test/test-applications/instrumentation/CallTargetNativeTest/CallTargetNativeTest.csproj -p:Configuration=$buildConfiguration -p:ManagedProfilerOutputDirectory="$PUBLISH_OUTPUT"
 
-for sample in Samples.Elasticsearch Samples.Elasticsearch.V5 Samples.ServiceStack.Redis Samples.StackExchange.Redis Samples.SqlServer Samples.Microsoft.Data.SqlClient Samples.MongoDB Samples.HttpMessageHandler Samples.WebRequest Samples.Npgsql Samples.MySql Samples.GraphQL Samples.Dapper Samples.NoMultiLoader Samples.RabbitMQ Samples.RuntimeMetrics Samples.FakeDbCommand Samples.Microsoft.Data.Sqlite Samples.SQLite.Core Samples.OracleMDA Samples.OracleMDA.Core ; do
```
lucaspimentel, line 45: Note: Skipping `Samples.OracleMDA` because it targets .NET Framework only. `Samples.OracleMDA.Core` target .NET Core/5.

test/Directory.Build.props
```diff
@@ -2,9 +2,9 @@
   <Import Project="..\Directory.Build.props" />
 
   <PropertyGroup>
-    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">net452;net461;netcoreapp2.1;netcoreapp3.0;netcoreapp3.1</TargetFrameworks>
-    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netcoreapp2.1;netcoreapp3.0;netcoreapp3.1</TargetFrameworks>
-    <TargetFrameworks Condition="$(MSBuildVersion) >= 16.8.0">$(TargetFrameworks);net5.0</TargetFrameworks>
+    <!-- only run .NET Framework tests on Windows -->
```
lucaspimentel, line 5: Note: keeping the `'$(OS)' == 'Windows_NT'` here to avoid .NET Framework on Linux/macOS. Otherwise `dotnet test` tried to find mono to run the tests.

test/test-applications/integrations/Directory.Build.props
```diff
@@ -1,3 +0,0 @@
-<Project>
-  <Import Project="..\Directory.Build.props" />
```
lucaspimentel, line 2: Note: looking for `Directory.Build.props` up the folder hierarchy is already the default behavior.

test/test-applications/integrations/Samples.RabbitMQ/Samples.RabbitMQ.csproj
```diff
@@ -5,10 +5,7 @@
     <DefineConstants Condition="$(ApiVersion) >= 6.0.0">$(DefineConstants);RABBITMQ_6_0</DefineConstants>
 
     <!-- override to remove net452 -->
-    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">net461;netcoreapp2.1;netcoreapp3.0;netcoreapp3.1</TargetFrameworks>
-    <!-- <TargetFrameworks Condition="'$(OS)' == 'Windows_NT' AND $(ApiVersion) &lt; 6.0.0">$(TargetFrameworks);net452</TargetFrameworks> -->
-    <TargetFrameworks Condition="$(MSBuildVersion) >= 16.8.0 AND !$(TargetFrameworks.Contains('net5.0'))">$(TargetFrameworks);net5.0</TargetFrameworks>
-    <DefineConstants Condition="$(ApiVersion) >= 6.0.0">$(DefineConstants);RABBITMQ_6_0</DefineConstants>
```
lucaspimentel, line 11: Note: this `<DefineConstants>` was a duplicate of line 5 above.

test/test-applications/regression/NLog10LogsInjection.NullReferenceException/NLog10LogsInjection.NullReferenceException.csproj
```diff
@@ -1,8 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">net452;net461</TargetFrameworks>
```
lucaspimentel, line 4: Was there any reason to not run this on .NET Core on Windows?

test/Directory.Build.props
```diff
@@ -2,9 +2,9 @@
...
+    <TargetFrameworks Condition="'$(TargetFrameworks)' == '' and '$(OS)' == 'Windows_NT'">net452;net461;netcoreapp2.1;netcoreapp3.0;netcoreapp3.1;net5.0</TargetFrameworks>
```
andrewlock, line 6: `'$(TargetFrameworks)' == ''` will this ever be the case? can't we just have the OS check here?

test/test-applications/integrations/Samples.Dapper/Samples.Dapper.csproj
```diff
@@ -3,8 +3,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
      <!-- override to remove net452 -->
-    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">net461;netcoreapp2.1;netcoreapp3.0;netcoreapp3.1</TargetFrameworks>
-    <TargetFrameworks Condition="$(MSBuildVersion) >= 16.8.0 AND !$(TargetFrameworks.Contains('net5.0'))">$(TargetFrameworks);net5.0</TargetFrameworks>
+    <TargetFrameworks>net461;netcoreapp2.1;netcoreapp3.0;netcoreapp3.1;net5.0</TargetFrameworks>
```
andrewlock, line 6: https://github.com/DataDog/dd-trace-dotnet/pull/1477 is about _not_ building samples unless we're going to use them in CI (to avoid waste). This change (and other samples changes) make these build in CI on Linux where previously they weren't. Is that a change we def want to make? 

FWIW, I'd rather have the simpler project files, as we can do optimisations in the CI build itself for speed separately.

test/test-applications/integrations/dependency-libs/Samples.NoMultiLoader.Deps/Samples.NoMultiLoader.Deps.csproj
```diff
@@ -1,11 +1,10 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">net45;net461;netstandard2.0;netcoreapp3.1</TargetFrameworks>
-    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netstandard2.0;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>net45;net461;netstandard2.0;netcoreapp3.1</TargetFrameworks>
```
andrewlock, line 4: Any reason we're not including .NET 5.0 here? 

Also, I know it's not from your changes, but why net45 and not net452 like everything else?

test/test-applications/integrations/dependency-libs/Samples.SqlServer.Vb/Samples.SqlServer.Vb.vbproj
```diff
@@ -1,9 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">net45;net461;netstandard2.0;netcoreapp3.1</TargetFrameworks>
-    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netstandard2.0;netcoreapp3.1</TargetFrameworks>
-
+    <TargetFrameworks>net45;net461;netstandard2.0;netcoreapp3.1;net5.0</TargetFrameworks>
```
andrewlock, line 4: Isn't this the "default" set of frameworks?

test/test-applications/regression/NLog10LogsInjection.NullReferenceException/NLog10LogsInjection.NullReferenceException.csproj
andrewlock: I assume [because NLog 1.0.0.5005](https://www.nuget.org/packages/NLog/1.0.0.505) doesn't support .NET Core

andrewlock (approved):
LGTM, just some minor comments

test/test-applications/regression/NLog10LogsInjection.NullReferenceException/NLog10LogsInjection.NullReferenceException.csproj
zacharycmontoya: Yeah, I expect it to crash at runtime, if not sooner, because it doesn't support .NET Core. Can you keep this as-is without the OS condition?

test/test-applications/integrations/dependency-libs/Samples.NoMultiLoader.Deps/Samples.NoMultiLoader.Deps.csproj
zacharycmontoya: I think this can use the default frameworks, since it doesn't use framework-specific code

test/test-applications/integrations/dependency-libs/Samples.SqlServer.Vb/Samples.SqlServer.Vb.vbproj
zacharycmontoya: This is using `net45` instead of `net452`. But I think it can use `net452` just like the C# project, so we could use the default frameworks here

test/Directory.Build.props
zacharycmontoya: No, we need to keep the `'$(TargetFrameworks)' == ''` check so that we can specify a smaller subset of target frameworks in project .csproj's, like when we can't build against `net452`

zacharycmontoya (approved):
LGTM with some small comments

test/Directory.Build.props
lucaspimentel: I assumed this was here in case we specified `TargetFrameworks` externally, like a msbuild property in the command line. This file is included before the project file, and the project file overwrites any previous value of `<TargetFrameworks>`.

For example, `/src/Datadog.Trace.AspNet/Datadog.Trace.AspNet.csproj` overwrites the value from `/src/Directory.Build.props`.

https://github.com/DataDog/dd-trace-dotnet/blob/13ee2e4ed79dfa9332b2aa8a916899d2d5b43a01/src/Datadog.Trace.AspNet/Datadog.Trace.AspNet.csproj#L4

https://github.com/DataDog/dd-trace-dotnet/blob/13ee2e4ed79dfa9332b2aa8a916899d2d5b43a01/src/Directory.Build.props#L5

test/test-applications/integrations/Samples.Dapper/Samples.Dapper.csproj
lucaspimentel: TBH, #1477 was more about getting around build errors in one of those projects (i.e. kicking the can down the road 😅). Adding an extra target here on Linux/macOS only adds a few seconds to the build, and I'd rather have the simpler project files as well. It may even help Linux/macOS users find compiler errors that only happen when targeting .NET Framework.

test/test-applications/integrations/dependency-libs/Samples.NoMultiLoader.Deps/Samples.NoMultiLoader.Deps.csproj
lucaspimentel: Done in bc24eaf9.

test/test-applications/regression/NLog10LogsInjection.NullReferenceException/NLog10LogsInjection.NullReferenceException.csproj
lucaspimentel: Done in 1acaf20b.

test/test-applications/integrations/dependency-libs/Samples.SqlServer.Vb/Samples.SqlServer.Vb.vbproj
lucaspimentel: Done in c3aee2b4242d603e045ce26dcdb943f9b5efd47b.

test/test-applications/integrations/dependency-libs/Samples.NoMultiLoader.Deps/Samples.NoMultiLoader.Deps.csproj
lucaspimentel: Actually, these have an empty props so they need to specify target. Fixed in 94de9af8.

test/test-applications/integrations/dependency-libs/Samples.SqlServer.Vb/Samples.SqlServer.Vb.vbproj
lucaspimentel: These have an empty props so they need to specify target. Fixed in 94de9af8.

PR #1469, opened 2021-05-16
Fix Call opcode in DuckTyping on struct targets.

tonyredondo:
This PR fixes methods and properties calls on DuckTyping over struct targets by changing the `CallVirt` to a `Call` opcode.

A strange behavior was observed on net461 in Release mode when using `CallVirt` opcode. With this fix the DuckType test suite passes in both `Debug` and `Release`.

This PR has been tested in all 8 pipelines available.

@DataDog/apm-dotnet


test/Datadog.Trace.DuckTyping.Tests/DuckIgnoreTests.cs
```diff
@@ -82,6 +84,8 @@ public abstract class AbstractPrivateProxy : IGetValue
 
             [DuckIgnore]
             public string GetValue() => Value.ToString();
+
+            public abstract int GetAnswerToMeaningOfLife();
```
andrewlock, line 88: 😂

andrewlock (approved):
Thanks for figuring this one out!

zacharycmontoya (approved):
Good find!

lucaspimentel (approved):

PR #1471, opened 2021-05-17
Add automatic instrumentation for AWS SQS

zacharycmontoya:
Adds support for AWSSDK.SQS v3.0.0+

## Design
- Integration Name = `AwsSqs` for SQS integrations, `AwsSdk` for underlying integration shared in `AWSSDK.Core` library
  - Note: The current design is to skip SQS integrations if either integration is disabled through configuration
- Instrumented Operations
  - CreateQueue
  - DeleteQueue
  - PurgeQueue
  - ReceiveMessage
  - SendMessage
  - SendMessageBatch

Metadata 
- Service = `aws-sqs`
- Span name = `sqs.request`
- Resource Name = `SQS.{OperationName}`
- Tags
  - HTTP tags (the client calls create underlying HTTP requests and client responses contain HTTP status codes)
    - `http.method`
    - `http.status_code`
  - General AWS tags
    - `aws.agent=dotnet-aws-sdk`
    - `aws.operation`
    - `aws.region`
    - `aws.service=SQS`
    - `aws.requestId`
  - Queue tags
    - `aws.queue.name`
    - `aws.queue.url`

Context Propagation
- Automatic instrumentation adds distributed tracing headers for each SendMessage and SendMessageBatch operation
  -  Message attribute: key=`_datadog`, value=`{"x-datadog-trace-id":"<uint64>", "x-datadog-parent-id":"<uint64>", "x-datadog-sampling-priority":"<int>", "x-datadog-origin":"<string>"}`
  - The headers are consolidated into one message attribute because SQS only allows a maximum of 10 message attributes on a single message. This design complies with other DD tracers
- Automatic instrumentation for `ReceiveMessage` creates a span but does not attempt to link a span to its associated `SendMessage`/`SendMessageBatch` span
  - This is because messages can be polled multiple times before they get timed out by the queue or deleted from an SQS client request

## Testing
- Add `Samples.AWS.SQS` program to CI to test instrumentation against a local `elasticmq` container that supports the SQS interface
- Tested and verified `aws.region` and `aws.requestId` tags when sending SQS requests to an AWS account

@DataDog/apm-dotnet


lucaspimentel:
I noticed the service name is `aws-sdk`. When we add instrumentation for other features in the AWS SDK (e.g. S3, Aurora, DynamoDB), do we really want them all grouped under the same service?

zacharycmontoya:
> I noticed the service name is `aws-sdk`. When we add instrumentation for other features in the AWS SDK (e.g. S3, Aurora, DynamoDB), do we really want them all grouped under the same service?

I wasn't 100% sure when I was working on it, but now that I'm looking at it, the Python and JS tracer both seem to do the second with `sqs` and `s3`, though Java does lump them all together as a general `aws-sdk` like I did here.

I think moving to individual service names like `sqs` and `s3` makes the most sense, unless there's pushback.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AWS/SQS/CachedMessageHeadersHelper.cs
```diff
@@ -0,0 +1,74 @@
...
+                constructedDictType,
+                null,
+                typeof(DuckType).Module,
+                true);
+
+            ILGenerator dictIL = createDictMethod.GetILGenerator();
+            dictIL.Emit(OpCodes.Newobj, dictionaryCtor);
+            dictIL.Emit(OpCodes.Ret);
+
+            _createDict = (Func<object>)createDictMethod.CreateDelegate(typeof(Func<object>));
```
andrewlock, line 61: I'm guessing it doesn't make any difference, but could we make this delegate `Func<IDictionary>`, to save the cast in the method below? Just a little more explicit :)

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AWS/SQS/ContextPropagation.cs
```diff
@@ -0,0 +1,79 @@
...
+        public static void Inject<TMessageRequest>(SpanContext context, IDictionary messageAttributes)
+        {
+            /* TODO: Either use the optimized StringBuilder or decide the optimization is not worth it
+            StringBuilder sb = new();
+            sb.Append("{");
+            SpanContextPropagator.Instance.Inject(context, sb, ((sb, key, value) => sb.Append($"\"{key}\":\"{value}\",")));
+            sb.Remove(startIndex: sb.Length - 1, length: 1); // Remove trailing comma
+            sb.Append("}");
+            var resultString = sb.ToString();
+            */
```
andrewlock, line 38: Is this a TODO for completion in this PR? Are there any benchmarks for this? Given it will run for every message, it's probably worth optimising if we can

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AWS/SQS/CreateQueueAsyncIntegration.cs
```diff
@@ -0,0 +1,72 @@
...
+            where TCreateQueueRequest : ICreateQueueRequest, IDuckType
+        {
+            if (request.Instance is null)
+            {
+                return CallTargetState.GetDefault();
+            }
+
+            var scope = AwsSqsCommon.CreateScope(Tracer.Instance, $"{AwsConstants.AwsService}.{Operation}", out AwsSqsTags tags);
+            tags.Operation = Operation;
+            tags.Service = AwsConstants.AwsService;
```
andrewlock, line 44: If service name is going to become `sqs` [as mentioned in this comment](https://github.com/DataDog/dd-trace-dotnet/pull/1471#issuecomment-844462806), is it still worth having this tag? I'm guessing so, given service name can be customised based on `DD_SERVICE`, just checking.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AWS/SQS/CreateQueueAsyncIntegration.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// <returns>A response value, in an async scenario will be T of Task of T</returns>
+        public static TResponse OnAsyncMethodEnd<TTarget, TResponse>(TTarget instance, TResponse response, Exception exception, CallTargetState state)
+            where TResponse : ICreateQueueResponse
+        {
+            if (state.Scope?.Span.Tags is AwsSqsTags tags)
+            {
+                tags.QueueUrl = response.QueueUrl;
+            }
+
+            state.Scope.DisposeWithException(exception);
```
andrewlock, line 68: I think we're missing a null propagator? Though not _all_ our integrations do this, so I'm not 100%. But if not, then the null propagator on the previous `if()` doesn't make sense 😉 

```suggestion
            state.Scope?.DisposeWithException(exception);
```

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AWS/SQS/CreateQueueIntegration.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// <returns>A response value, in an async scenario will be T of Task of T</returns>
+        public static CallTargetReturn<TResponse> OnMethodEnd<TTarget, TResponse>(TTarget instance, TResponse response, Exception exception, CallTargetState state)
+            where TResponse : ICreateQueueResponse
+        {
+            if (state.Scope?.Span.Tags is AwsSqsTags tags)
+            {
+                tags.QueueUrl = response.QueueUrl;
+            }
+
+            state.Scope.DisposeWithException(exception);
```
andrewlock, line 66:
```suggestion
            state.Scope?.DisposeWithException(exception);
```

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AWS/SQS/DeleteMessageAsyncIntegration.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// <typeparam name="TTarget">Type of the target</typeparam>
+        /// <typeparam name="TResponse">Type of the response, in an async scenario will be T of Task of T</typeparam>
+        /// <param name="instance">Instance value, aka `this` of the instrumented method.</param>
+        /// <param name="response">Response instance</param>
+        /// <param name="exception">Exception instance in case the original code threw an exception.</param>
+        /// <param name="state">Calltarget state value</param>
+        /// <returns>A response value, in an async scenario will be T of Task of T</returns>
+        public static TResponse OnAsyncMethodEnd<TTarget, TResponse>(TTarget instance, TResponse response, Exception exception, CallTargetState state)
+        {
+            state.Scope.DisposeWithException(exception);
```
andrewlock, line 62:
```suggestion
            state.Scope?.DisposeWithException(exception);
```

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AWS/SQS/DeleteMessageBatchAsyncIntegration.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// <typeparam name="TTarget">Type of the target</typeparam>
+        /// <typeparam name="TResponse">Type of the response, in an async scenario will be T of Task of T</typeparam>
+        /// <param name="instance">Instance value, aka `this` of the instrumented method.</param>
+        /// <param name="response">Response instance</param>
+        /// <param name="exception">Exception instance in case the original code threw an exception.</param>
+        /// <param name="state">Calltarget state value</param>
+        /// <returns>A response value, in an async scenario will be T of Task of T</returns>
+        public static TResponse OnAsyncMethodEnd<TTarget, TResponse>(TTarget instance, TResponse response, Exception exception, CallTargetState state)
+        {
+            state.Scope.DisposeWithException(exception);
```
andrewlock, line 62:
```suggestion
            state.Scope?.DisposeWithException(exception);
```

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AWS/SQS/ReceiveMessageAsyncIntegration.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// <typeparam name="TTarget">Type of the target</typeparam>
+        /// <typeparam name="TResponse">Type of the response, in an async scenario will be T of Task of T</typeparam>
+        /// <param name="instance">Instance value, aka `this` of the instrumented method.</param>
+        /// <param name="response">Response instance</param>
+        /// <param name="exception">Exception instance in case the original code threw an exception.</param>
+        /// <param name="state">Calltarget state value</param>
+        /// <returns>A response value, in an async scenario will be T of Task of T</returns>
+        public static TResponse OnAsyncMethodEnd<TTarget, TResponse>(TTarget instance, TResponse response, Exception exception, CallTargetState state)
+        {
+            state.Scope.DisposeWithException(exception);
```
andrewlock, line 62:
```suggestion
            state.Scope?.DisposeWithException(exception);
```

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AWS/SQS/ReceiveMessageIntegration.cs
```diff
@@ -0,0 +1,64 @@
...
+        /// <typeparam name="TTarget">Type of the target</typeparam>
+        /// <typeparam name="TResponse">Type of the response</typeparam>
+        /// <param name="instance">Instance value, aka `this` of the instrumented method.</param>
+        /// <param name="response">Response instance</param>
+        /// <param name="exception">Exception instance in case the original code threw an exception.</param>
+        /// <param name="state">Calltarget state value</param>
+        /// <returns>A response value, in an async scenario will be T of Task of T</returns>
+        public static CallTargetReturn<TResponse> OnMethodEnd<TTarget, TResponse>(TTarget instance, TResponse response, Exception exception, CallTargetState state)
+        {
+            state.Scope.DisposeWithException(exception);
```
andrewlock, line 60:
```suggestion
            state.Scope?.DisposeWithException(exception);
```

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AWS/SQS/SendMessageAsyncIntegration.cs
```diff
@@ -0,0 +1,72 @@
...
+            }
+
+            var requestProxy = request.DuckCast<ISendMessageRequest>();
+
+            var scope = AwsSqsCommon.CreateScope(Tracer.Instance, $"{AwsConstants.AwsService}.{Operation}", out AwsSqsTags tags);
+            tags.Operation = Operation;
+            tags.Service = AwsConstants.AwsService;
+            tags.QueueUrl = requestProxy.QueueUrl;
+
+            if (scope?.Span?.Context != null)
```
andrewlock, line 48: Is there a case here where `scope!=null` but `scope.Span==null` 🤔

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AWS/SQS/SendMessageAsyncIntegration.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// <typeparam name="TTarget">Type of the target</typeparam>
+        /// <typeparam name="TResponse">Type of the response, in an async scenario will be T of Task of T</typeparam>
+        /// <param name="instance">Instance value, aka `this` of the instrumented method.</param>
+        /// <param name="response">Response instance</param>
+        /// <param name="exception">Exception instance in case the original code threw an exception.</param>
+        /// <param name="state">Calltarget state value</param>
+        /// <returns>A response value, in an async scenario will be T of Task of T</returns>
+        public static TResponse OnAsyncMethodEnd<TTarget, TResponse>(TTarget instance, TResponse response, Exception exception, CallTargetState state)
+        {
+            state.Scope.DisposeWithException(exception);
```
andrewlock, line 68:
```suggestion
            state.Scope?.DisposeWithException(exception);
```

test/Datadog.Trace.ClrProfiler.IntegrationTests/AWS/AwsSqsTests.cs
```diff
@@ -0,0 +1,169 @@
...
+            : base("Aws.Sqs", output)
+        {
+        }
+
+        [Theory]
+        [MemberData(nameof(PackageVersions.AwsSqs), MemberType = typeof(PackageVersions))]
+        [Trait("Category", "EndToEnd")]
+        public void SubmitsTraces(string packageVersion)
+        {
+            SetCallTargetSettings(enableCallTarget: true, enableMethodInlining: true);
```
andrewlock, line 121: This method doesn't existing any more in master! Looks like you need a rebase :)
```suggestion
            SetCallTargetSettings(enableCallTarget: true);
```

test/Datadog.Trace.ClrProfiler.IntegrationTests/AWS/AwsSqsTests.cs
```diff
@@ -0,0 +1,169 @@
...
+
+                var spans = agent.WaitForSpans(_expectedSpans.Count, operationName: AwsExpectation.IntegrationOperationName);
+                spans.Should().HaveCountGreaterOrEqualTo(_expectedSpans.Count);
+
+                spans.OrderBy(s => s.Start).Should().BeEquivalentTo(_expectedSpans, options => options
+                    .WithStrictOrdering()
+                    .ExcludingMissingMembers()
+                    .ExcludingDefaultSpanProperties()
+                    .AssertMetricsMatchExcludingKeys("_dd.tracer_kr", "_sampling_priority_v1")
+                    .AssertTagsMatchAndSpecifiedTagsPresent("env", "aws.requestId", "aws.queue.url"));
```
andrewlock, line 137: Very slick! 🎉 Though feels like trying to create our own version of Verify 😄

test/Datadog.Trace.ClrProfiler.IntegrationTests/AWS/AwsSqsTests.cs
```diff
@@ -0,0 +1,169 @@
...
+            // DeleteQueue
+            AwsSqsSpan.GetDefault("DeleteQueue"),
+            AwsSqsSpan.GetDefault("DeleteQueue"),
+
+#endif
+            // CreateSqsQueueAsync
+            AwsSqsSpan.GetDefault("CreateQueue")
+                    .WithTag("aws.queue.name", "MySQSQueue"),
+            AwsSqsSpan.GetDefault("CreateQueue")
+                    .WithTag("aws.queue.name", "MySQSQueue2"),
```
andrewlock, line 68: I'm guessing these should all have the `Async` suffix 🙂

andrewlock (commented):
I see you fought with the same things I did with Kafka! 😄 Mostly minor comments and questions

docker-compose.yml
```diff
@@ -168,6 +173,19 @@ services:
...
+      dockerfile: ./docker/dotnet.dockerfile
+    image: datadog-dotnet
+    command: bash -c "/project/docker/with-profiler-logs.bash wait-for-it elasticmq:9324 -- /project/docker/with-profiler.bash dotnet /project/samples/Samples.Aws.Sqs/bin/${buildConfiguration}/${publishTargetFramework}/publish/Samples.Aws.Sqs.dll"
+    environment:
+    - AWS_SQS_HOST=elasticmq:9324
+    volumes:
+    - ./:/project
+    depends_on:
+    - aws_sqs
+
```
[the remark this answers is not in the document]
zacharycmontoya: Yeah I think it's fine to remove this. I think the one benefit to having this docker-compose service is that it's helpful for running samples on Linux, but that should be taken care of with Nuke

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AWS/SDK/RuntimePipelineInvokeAsyncIntegration.cs
```diff
@@ -0,0 +1,79 @@
...
+namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.AWS.SDK
+{
+    /// <summary>
+    /// AWSSDK.Core InvokeAsync calltarget instrumentation
+    /// </summary>
+    [InstrumentMethod(
+        AssemblyName = "AWSSDK.Core",
+        TypeName = "Amazon.Runtime.Internal.RuntimePipeline",
+        MethodName = "InvokeAsync",
+        ReturnTypeName = "System.Threading.Tasks.Task`1<T>",
```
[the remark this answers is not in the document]
zacharycmontoya: Looks like we already have `ClrNames.GenericParameterTask` for this. I'll make sure to use that, thanks for pointing this out 😄

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AWS/SDK/RuntimePipelineInvokeAsyncIntegration.cs
```diff
@@ -0,0 +1,79 @@
...
+            where TResponse : IAmazonWebServiceResponse
+        {
+            if (state.Scope?.Span.Tags is AwsSdkTags tags)
+            {
+                if (state.State is not null)
+                {
+                    // The request object is populated later by the Marshaller,
+                    // so we wait until the method end callback to read it
+                    tags.HttpMethod = state.State.DuckCast<IExecutionContext>().RequestContext.Request.HttpMethod.ToUpperInvariant();
+                }
```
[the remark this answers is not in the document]
zacharycmontoya: Yes you're absolutely right, good catch! Fixed in b6828c0b9ea4e3484668086081d78b7c396972e1

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AWS/SDK/RuntimePipelineInvokeSyncIntegration.cs
```diff
@@ -0,0 +1,79 @@
...
+            where TResponseContext : IResponseContext
+        {
+            if (state.Scope?.Span.Tags is AwsSdkTags tags)
+            {
+                if (state.State is not null)
+                {
+                    // The request object is populated later by the Marshaller,
+                    // so we wait until the method end callback to read it
+                    tags.HttpMethod = state.State.DuckCast<IExecutionContext>().RequestContext.Request.HttpMethod.ToUpperInvariant();
+                }
```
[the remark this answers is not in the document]
zacharycmontoya: Fixed in b6828c0b9ea4e3484668086081d78b7c396972e1

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AWS/SDK/RuntimePipelineInvokeAsyncIntegration.cs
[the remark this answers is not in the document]
zacharycmontoya: Fixed in 186d1e0a5ef147d21379fbf6a13d012eed63e10c

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AWS/SQS/CreateQueueIntegration.cs
zacharycmontoya: `DisposeWithException` does a null check on the `Scope` argument so I was going to keep this as-is

test/Datadog.Trace.ClrProfiler.IntegrationTests/AWS/AwsSqsTests.cs
zacharycmontoya: Thanks, I finished rebasing and fixed this method invocation in 23bf71d649949821743706ee64a69c287d3fe9b9

test/Datadog.Trace.ClrProfiler.IntegrationTests/AWS/AwsSqsTests.cs
zacharycmontoya: Thanks! I think I started this branch before we had started testing Verify so I thought it would be a great way to express the key tags and metrics. I don't really want to scratch this because it's doing its job and we haven't merged Verify yet..

test/Datadog.Trace.ClrProfiler.IntegrationTests/AWS/AwsSqsTests.cs
zacharycmontoya: Good eye, but this proposal was intentional to not have the `Async` suffix. While the .NET SDK has the `Async` suffix on method names, the SQS API does not have that suffix so I wanted to pair 1:1 with the actual API operations. Do you think that's the correct approach or we should match this to the client SDK method name?

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AWS/SQS/SendMessageAsyncIntegration.cs
zacharycmontoya: Yeah you're right, in this automatic instrumentation I know that if I created a Scope then there is a Span associated with it. I'll remove the additional null-check here. Fixed in 9a9e2a736689c1cf169919587ae856e406d8c188

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AWS/SQS/CachedMessageHeadersHelper.cs
zacharycmontoya: Yeah I can do that. Fixed in aec4cd2f86db5039a545fa176f147f7a609044ab

build/docker/build.sh
```diff
@@ -42,7 +42,7 @@ fi
...
 dotnet publish -f $publishTargetFramework -c $buildConfiguration test/test-applications/instrumentation/CallTargetNativeTest/CallTargetNativeTest.csproj -p:Configuration=$buildConfiguration -p:ManagedProfilerOutputDirectory="$PUBLISH_OUTPUT"
 
-for sample in Samples.Elasticsearch Samples.Elasticsearch.V5 Samples.ServiceStack.Redis Samples.StackExchange.Redis Samples.SqlServer Samples.Microsoft.Data.SqlClient Samples.MongoDB Samples.HttpMessageHandler Samples.WebRequest Samples.Npgsql Samples.MySql Samples.GraphQL Samples.Dapper Samples.NoMultiLoader Samples.RabbitMQ Samples.RuntimeMetrics Samples.FakeDbCommand Samples.Microsoft.Data.Sqlite Samples.SQLite.Core Samples.OracleMDA Samples.OracleMDA.Core ; do
+for sample in Samples.Elasticsearch Samples.Elasticsearch.V5 Samples.ServiceStack.Redis Samples.StackExchange.Redis Samples.SqlServer Samples.Microsoft.Data.SqlClient Samples.MongoDB Samples.HttpMessageHandler Samples.WebRequest Samples.Npgsql Samples.MySql Samples.GraphQL Samples.Dapper Samples.NoMultiLoader Samples.RabbitMQ Samples.RuntimeMetrics Samples.FakeDbCommand Samples.Microsoft.Data.Sqlite Samples.SQLite.Core Samples.OracleMDA Samples.OracleMDA.Core Samples.Aws.Sqs; do
```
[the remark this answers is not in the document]
zacharycmontoya: Yep! Fixed in ef467fbc92ebc08ae7574d942fb206dcc53c2ba1

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AWS/SQS/CreateQueueAsyncIntegration.cs
zacharycmontoya: My gut says yes, and I think it would also help to have `aws.service=sqs` as a tag for doing filtering in the UI as well

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AWS/SQS/ContextPropagation.cs
zacharycmontoya: Now that I have a complete implementation, I will benchmark this. I just wanted to get it feature-complete before addressing this 😄

zacharycmontoya:
> I think moving to individual service names like `sqs` and `s3` makes the most sense, unless there's pushback.

Service name changed to `sqs` in 50545acdcc8c1469990e1d076b5692c1f26073f4

Datadog.Trace.sln
```diff
@@ -1562,10 +1564,10 @@ Global
...
 		{1ACC5FC6-2F80-4A0C-A46C-9A97C0AB181B}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
 		{1ACC5FC6-2F80-4A0C-A46C-9A97C0AB181B}.Debug|Any CPU.Build.0 = Debug|Any CPU
-		{1ACC5FC6-2F80-4A0C-A46C-9A97C0AB181B}.Debug|x64.ActiveCfg = Debug|x64
-		{1ACC5FC6-2F80-4A0C-A46C-9A97C0AB181B}.Debug|x64.Build.0 = Debug|x64
-		{1ACC5FC6-2F80-4A0C-A46C-9A97C0AB181B}.Debug|x86.ActiveCfg = Debug|x86
-		{1ACC5FC6-2F80-4A0C-A46C-9A97C0AB181B}.Debug|x86.Build.0 = Debug|x86
+		{1ACC5FC6-2F80-4A0C-A46C-9A97C0AB181B}.Debug|x64.ActiveCfg = Debug|Any CPU
+		{1ACC5FC6-2F80-4A0C-A46C-9A97C0AB181B}.Debug|x64.Build.0 = Debug|Any CPU
+		{1ACC5FC6-2F80-4A0C-A46C-9A97C0AB181B}.Debug|x86.ActiveCfg = Debug|Any CPU
+		{1ACC5FC6-2F80-4A0C-A46C-9A97C0AB181B}.Debug|x86.Build.0 = Debug|Any CPU
```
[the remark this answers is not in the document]
zacharycmontoya: Fixed in 5917ba5b0166831a6b1942c351504d111fd9a63b

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AWS/SQS/CreateQueueAsyncIntegration.cs
zacharycmontoya: I followed up on another comment, but the `DisposeWithException` extension method does a null check on the `Scope` argument. Would it be more readable to do the null check outside the extension method and then have the extension method expect an non-null value?

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AWS/SQS/ContextPropagation.cs
zacharycmontoya: Changed over to the pure StringBuilder approach in f51fb2b2562313b14ed5d68673100cd360f9f8da which does introduce some savings in execution time!

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AWS/SQS/CreateQueueAsyncIntegration.cs
andrewlock: Ah, good point. Well, IMO the null check at the callsite is better for readability and expectations, but having it in the extension method is safer!  My biggest issue is with inconsistency, so I think we leave it as-is, and do a small cleanup to make all the invocations identical (i.e. drop the null check from callsites)

test/Datadog.Trace.ClrProfiler.IntegrationTests/AWS/AwsSqsTests.cs
andrewlock: > Good Eye

Except I missed that the `OperationName` doesn't have the `*Async` suffix in the instrumentation 😆 

> Do you think that's the correct approach or we should match this to the client SDK method name?

That's a good question. I think your intuition is correct - given this is about resource names, the "resource" is identical, regardless of whether you're using the `CreateQueue` or `CreateQueueAsync` method.

test/Datadog.Trace.ClrProfiler.IntegrationTests/AWS/AwsSqsTests.cs
```diff
@@ -0,0 +1,174 @@
...
+            AwsSqsSpan.GetDefault("DeleteMessageBatch"),
+            AwsSqsSpan.GetDefault("ReceiveMessage"),
+            AwsSqsSpan.GetDefault("DeleteMessageBatch"),
+
+            // DeleteQueue
+            AwsSqsSpan.GetDefault("DeleteQueue"),
+            AwsSqsSpan.GetDefault("DeleteQueue"),
+
+#endif
+            // CreateSqsQueueAsync
```
andrewlock, line 69: May be worth adding a note reflecting [the explanation in your comment](https://github.com/DataDog/dd-trace-dotnet/pull/1471#discussion_r638171112)? Just so it's clearer it's not a typo? 🙂 

```suggestion
            // Span resource names do not include the Async prefix
            // CreateSqsQueueAsync
```

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AWS/SQS/ContextPropagation.cs
```diff
@@ -0,0 +1,47 @@
...
+        private const string SqsKey = "_datadog";
+
+        private static void Inject<TMessageRequest>(SpanContext context, IDictionary messageAttributes)
+        {
+            // Consolidate headers into one JSON object with <header_name>:<value>
+            StringBuilder sb = new();
+            sb.Append("{");
+            SpanContextPropagator.Instance.Inject(context, sb, ((carrier, key, value) => carrier.Append($"\"{key}\":\"{value}\",")));
+            sb.Remove(startIndex: sb.Length - 1, length: 1); // Remove trailing comma
+            sb.Append("}");
```
andrewlock, line 25: It's very minor, but seeing as this change was performance reasons, I believe using `Append(char)` over `Append(string)` has some performance improvements, as well as using `AppendFormat()` instead of interpolation to avoid allocating an extra string.

```suggestion
            StringBuilder sb = new();
            sb.Append('{');
            SpanContextPropagator.Instance.Inject(context, sb, ((carrier, key, value) => carrier.AppendFormat("\"{0}\":\"{1}\",", key, value)));
            sb.Remove(startIndex: sb.Length - 1, length: 1); // Remove trailing comma
            sb.Append('}');
```

docker-compose.yml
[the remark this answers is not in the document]
zacharycmontoya: Resolved in 3a2ef6c08d8ce4d9ad7f8607fd96bfc83b98d05a

test/Datadog.Trace.ClrProfiler.IntegrationTests/AWS/AwsSqsTests.cs
zacharycmontoya: Fixed in d19668f4916196542b7a0f7329193a9f80499c53

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AWS/SQS/ContextPropagation.cs
zacharycmontoya: Great suggestion, a quick benchmark shows even fewer allocations now! Fixed in 4ae4651a277ffb107d72e19425c5e800d1fe92b0

zacharycmontoya: For reference:

```
|                Method |     Mean |    Error |   StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------------------- |---------:|---------:|---------:|-------:|------:|------:|----------:|
|        ExistingMethod | 594.4 ns | 11.79 ns | 24.08 ns | 0.3138 |     - |     - |    1484 B |
| WithAndrewSuggestions | 553.9 ns | 11.01 ns | 26.60 ns | 0.1783 |     - |     - |     842 B |
```

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AWS/SQS/CreateQueueAsyncIntegration.cs
zacharycmontoya: Cleanup done in separate PR: https://github.com/DataDog/dd-trace-dotnet/pull/1493

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AWS/SQS/ContextPropagation.cs
zacharycmontoya: Uhh, actually fixed in 1ff312053519354e4683e32be2a12fdf5578b84d

integrations.json
```diff
@@ -256,6 +256,386 @@
...
+          ],
+          "minimum_major": 3,
+          "minimum_minor": 0,
+          "minimum_patch": 0,
+          "maximum_major": 3,
+          "maximum_minor": 65535,
+          "maximum_patch": 65535
+        },
+        "wrapper": {
+          "assembly": "Datadog.Trace.ClrProfiler.Managed, Version=1.26.3.0, Culture=neutral, PublicKeyToken=def86d061d0d2eeb",
```
lucaspimentel, line 281: We are up to 1.27.0 now. Can you regenerate this file?

lucaspimentel (commented):
Thoughts on changing the service name from `aws-sdk` to `aws-sqs` and the span name from `aws.request` to `aws-sqs.request`? I think the "sdk" in the service name is reduntant (aren't _all_ client libraries also SDKs?), and we would want to separate the different AWS services.

zacharycmontoya:
> Thoughts on changing the service name from `aws-sdk` to `aws-sqs` and the span name from `aws.request` to `aws-sqs.request`? I think the "sdk" in the service name is reduntant (aren't _all_ client libraries also SDKs?), and we would want to separate the different AWS services.

I changed the service name to `sqs` so I've updated it in the PR summary. Is that fine or do you specifically want `aws-sqs`? Whichever name you prefer, we can use the same name for the span name as well, to replace `aws.request`

integrations.json
zacharycmontoya: Updated in the latest commit 860bf31684800742947bd053a691335739097718

lucaspimentel:
> I changed the service name to sqs so I've updated it in the PR summary. Is that fine or do you specifically want aws-sqs? Whichever name you prefer, we can use the same name for the span name as well, to replace "aws" in aws.request

Service name `sqs` is probably fine. I would change the span name to `sqs.request`, too. It's not that the span name needs to match the service name, but `aws.request` is too ambiguous. It could be any one of 100 services in AWS.

lucaspimentel (approved):

zacharycmontoya:
The crank job timed out and the integration-tests failed on an unrelated XUnit test. I'll go ahead and merge this!

PR #1473, opened 2021-05-18
Cosmos Db support

robertpi:
- Call target instrumentation only
- Test only run on windows, since the emulator is not supported on other
  platforms

- Integration Name = `CosmosDb` 
- Instrumented Operations
  - Client.GetDatabaseQueryIterator
  - Client.GetDatabaseQueryStreamIterator
  - Container.GetItemQueryIterator
  - Container.GetItemQueryStreamIterator
  - Database.GetUserQueryIterator
  - Database.GetContainerQueryStreamIterator
  - Database.GetContainerQueryStreamIterator

Metadata 
- Service = `cosmosdb`
- Span name = `cosmosdb.query`
- Resource Name = The SQL being executed
- Tags
  - Database tags
    - `db.type`
    - `db.name`
  - Network tags
    - `out.host`
  - Cosmos specific tags
    - `cosmosdb.container`

This what the traces screen looks like:
2021-06-07 15_51_15-APM Traces _ Datadog

This is an example of the details present in a trace:
2021-06-07 15_52_08-Trace 6930924… (env_robert) _ Datadog

@DataDog/apm-dotnet


Datadog.Trace.sln
```diff
@@ -1562,8 +1564,8 @@ Global
 		{B0F3148C-1251-4AAA-B6E6-357A8E675EA1}.Release|x86.Build.0 = Release|x86
 		{1ACC5FC6-2F80-4A0C-A46C-9A97C0AB181B}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
 		{1ACC5FC6-2F80-4A0C-A46C-9A97C0AB181B}.Debug|Any CPU.Build.0 = Debug|Any CPU
-		{1ACC5FC6-2F80-4A0C-A46C-9A97C0AB181B}.Debug|x64.ActiveCfg = Debug|x64
-		{1ACC5FC6-2F80-4A0C-A46C-9A97C0AB181B}.Debug|x64.Build.0 = Debug|x64
+		{1ACC5FC6-2F80-4A0C-A46C-9A97C0AB181B}.Debug|x64.ActiveCfg = Debug|Any CPU
+		{1ACC5FC6-2F80-4A0C-A46C-9A97C0AB181B}.Debug|x64.Build.0 = Debug|Any CPU
```
andrewlock, line 1568: Which project is this a change for, it's not the cosmos one AFAICT? Was it intentional?

build/docker/Datadog.Trace.ClrProfiler.IntegrationTests.sh
```diff
@@ -34,4 +34,4 @@ wait-for-it elasticsearch5:9200 -- \
 wait-for-it sqlserver:1433 -- \
 wait-for-it mongo:27017 -- \
 wait-for-it postgres:5432 -- \
-dotnet vstest test/Datadog.Trace.ClrProfiler.IntegrationTests/bin/$buildConfiguration/$publishTargetFramework/publish/Datadog.Trace.ClrProfiler.IntegrationTests.dll --logger:trx --ResultsDirectory:test/Datadog.Trace.ClrProfiler.IntegrationTests/results
+dotnet vstest test/Datadog.Trace.ClrProfiler.IntegrationTests/bin/$buildConfiguration/$publishTargetFramework/publish/Datadog.Trace.ClrProfiler.IntegrationTests.dll --logger:trx --ResultsDirectory:test/Datadog.Trace.ClrProfiler.IntegrationTests/results --TestCaseFilter:Category!=LinuxUnsupported
```
andrewlock, line 37: I know you added the `ArmUnsupported` trait to the test to cover this, but I think we should probably add the `LinuxUnsupported` to the arm64 bash script too. e.g. down the line we might support Windows on Arm. Not critical, just a thought

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/CosmosDb/QueryIteratorsIntegrations.cs
```diff
@@ -0,0 +1,308 @@
...
+        IntegrationName = IntegrationName)]
+    [InstrumentMethod(
+        AssemblyName = MicrosoftAzureCosmosClientAssemblyName,
+        TypeName = "Microsoft.Azure.Cosmos.CosmosClient",
+        MethodName = "GetDatabaseQueryStreamIterator",
+        ReturnTypeName = MicrosoftAzureCosmosFeedIteratorTypeName,
+        ParameterTypeNames = new[] { ClrNames.String, ClrNames.String, MicrosoftAzureCosmosQueryRequestOptionsTypeName, },
+        MinimumVersion = Major3Minor0,
+        MaximumVersion = Major3MinorX,
+        IntegrationName = IntegrationName)]
```
andrewlock, line 155: I wonder if it would make more sense to split some of these into separate integrations rather than reusing the same class. With the current design we have to do a lot of `if...else if` and trying duck casts. If they were split to support just the expected parameters we could avoid that complexity + overhead? Most of the logic could move to a Helper class that does the actual span creation, similar to some of the other integrations

src/Datadog.Trace/Configuration/IntegrationIds.cs
```diff
@@ -25,6 +25,7 @@ internal enum IntegrationIds
         StackExchangeRedis,
         ServiceRemoting,
         RabbitMQ,
-        Msmq
+        Msmq,
+        CosmosDb,
```
andrewlock, line 29: I'm super-pro trailing commas 👍 😄

test/test-applications/integrations/Samples.CosmosDb/Samples.CosmosDb.csproj
```diff
@@ -0,0 +1,21 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <ApiVersion Condition="'$(ApiVersion)' == ''">3.18.0</ApiVersion>
+    <OutputType>Exe</OutputType>
+    <IsPackable>false</IsPackable>
+    <LangVersion>latest</LangVersion>
+    <!-- override to remove net452 -->
+    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">net461;netcoreapp2.1;netcoreapp3.0;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks Condition="$(MSBuildVersion) &gt;= 16.8.0 AND !$(TargetFrameworks.Contains('net5.0'))">$(TargetFrameworks);net5.0</TargetFrameworks>
```
andrewlock, line 10: I don't think we need this any more, according to Lucas 🙂

andrewlock (approved):
LGTM, just some relatively minor questions 🙂

test/test-applications/integrations/Samples.CosmosDb/Samples.CosmosDb.csproj
lucaspimentel: As long as we're using SDK 5.0+, we should be able to simplify into:
```xml
<TargetFrameworks>net461;netcoreapp2.1;netcoreapp3.0;netcoreapp3.1;net5.0</TargetFrameworks>
```
I'm doing that across the the board in #1468, but still working through some bumps.

Datadog.Trace.sln
lucaspimentel: Looks like `Samples.AspNetCoreSimpleController.csproj`. Can you revert this change? We want the x64 solution configuration to use the x64 project configuration, not Any CPU.

tonyredondo (changes requested):
Looking good so far, I have a few comments to reduce allocations.

src/Datadog.Trace/SpanTypes.cs
```diff
@@ -61,5 +61,10 @@ public static class SpanTypes
         /// The span type for msbuild integration.
         /// </summary>
         public const string Build = "build";
+
+        /// <summary>
+        /// The span type for msbuild integration.
```
lucaspimentel, line 66:
```suggestion
        /// The span type for CosmosDb integration.
```

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/CosmosDb/QueryIteratorsIntegrations.cs
```diff
@@ -0,0 +1,308 @@
...
+            try
+            {
+                string query;
+                if (queryDefinition is string queryDefinitionString)
+                {
+                    query = queryDefinitionString;
+                }
+                else
+                {
+                    var queryDefinitionObj = DuckType.Create<IQueryDefinition>(queryDefinition);
```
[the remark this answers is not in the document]
tonyredondo: Also consider using the `TryCreate` DuckType extension method: 
https://github.com/DataDog/dd-trace-dotnet/blob/master/src/Datadog.Trace/DuckTyping/DuckTypeExtensions.cs#L71

src/Datadog.Trace/Tagging/CosmosDbTags.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Datadog.Trace.Tagging
+{
+    internal class CosmosDbTags : InstrumentationTags
+    {
+        protected static readonly IProperty<string>[] CosmosDbTagsProperties =
+            InstrumentationTagsProperties.Concat(
+                new ReadOnlyProperty<CosmosDbTags, string>(Trace.Tags.InstrumentationName, t => t.InstrumentationName),
+                new Property<CosmosDbTags, string>(Trace.Tags.CosmosDbContainer, t => t.ContainerId, (t, v) => t.ContainerId = v),
+                new Property<CosmosDbTags, string>(Trace.Tags.CosmosDbDatabase, t => t.DatabaseId, (t, v) => t.DatabaseId = v),
```
lucaspimentel, line 13: Should we reuse `db.name` instead of adding the new tag name `cosmosdb.database`?
(but then what do we do with `cosmosdb.container`?)

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/CosmosDb/QueryIteratorsIntegrations.cs
```diff
@@ -0,0 +1,308 @@
...
+                    Host = endpoint
+                };
+
+                var serviceName = tracer.Settings.GetServiceName(tracer, ServiceName);
+                var scope = tracer.StartActiveWithTags(OperationName, tags: tags, serviceName: serviceName);
+
+                var span = scope.Span;
+
+                span.ResourceName = query;
+                span.Type = SpanTypes.CosmosDb;
```
lucaspimentel, line 280: If resource names look like SQL queries (`SELECT X FROM Y`), we should consider using the `"sql"` span type (and add a tag `db.type = cosmosdb`). The span type will trigger query obfuscating in the Agent and special features in the UI, like syntax highlighting and showing the services under the "db" category.

andrewlock:
Just read about this for running the CosmosDb emuluator on linux @robertpi - might be worth a look? https://gist.github.com/StefArroyo/86a87056300fa58c2042e84cdfaa0e1c#run-the-cosmos-db-linux-emulator-on-linux

test/test-applications/integrations/Samples.CosmosDb/Samples.CosmosDb.csproj
robertpi: So, seems Lucas got his PR in before me, so I have adapted this now.

src/Datadog.Trace/Tagging/CosmosDbTags.cs
robertpi: I've used `db.name`, I guess `cosmosdb.container` will be okay on it's own.

Datadog.Trace.sln
```diff
@@ -1570,8 +1572,8 @@ Global
 		{B0F3148C-1251-4AAA-B6E6-357A8E675EA1}.Release|x86.Build.0 = Release|x86
 		{1ACC5FC6-2F80-4A0C-A46C-9A97C0AB181B}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
 		{1ACC5FC6-2F80-4A0C-A46C-9A97C0AB181B}.Debug|Any CPU.Build.0 = Debug|Any CPU
-		{1ACC5FC6-2F80-4A0C-A46C-9A97C0AB181B}.Debug|x64.ActiveCfg = Debug|Any CPU
-		{1ACC5FC6-2F80-4A0C-A46C-9A97C0AB181B}.Debug|x64.Build.0 = Debug|Any CPU
+		{1ACC5FC6-2F80-4A0C-A46C-9A97C0AB181B}.Debug|x64.ActiveCfg = Debug|x64
+		{1ACC5FC6-2F80-4A0C-A46C-9A97C0AB181B}.Debug|x64.Build.0 = Debug|x64
```
robertpi, line 1576: {1ACC5FC6-2F80-4A0C-A46C-9A97C0AB181B} is Samples.AspNetCoreSimpleController.csproj, seems to makes sense to hqve the x64 settings using x64

Datadog.Trace.sln
```diff
@@ -635,8 +637,8 @@ Global
 		{4B243CF1-4269-45C6-A238-1A9BFA58B8CC}.Release|x64.Build.0 = Release|Any CPU
 		{4B243CF1-4269-45C6-A238-1A9BFA58B8CC}.Release|x86.ActiveCfg = Release|Any CPU
 		{4B243CF1-4269-45C6-A238-1A9BFA58B8CC}.Release|x86.Build.0 = Release|Any CPU
-		{0D546118-B70A-44D0-B675-39EDB99FCEEE}.Debug|Any CPU.ActiveCfg = Debug|x86
-		{0D546118-B70A-44D0-B675-39EDB99FCEEE}.Debug|Any CPU.Build.0 = Debug|x86
+		{0D546118-B70A-44D0-B675-39EDB99FCEEE}.Debug|Any CPU.ActiveCfg = Debug|x64
+		{0D546118-B70A-44D0-B675-39EDB99FCEEE}.Debug|Any CPU.Build.0 = Debug|x64
```
robertpi, line 641: {0D546118-B70A-44D0-B675-39EDB99FCEEE} is Datadog.Trace.ClrProfiler.IntegrationTests.csproj, Any CPU needs to be either x86 or x64, so it knows what version of the test to run, I think x64 makes more sense as a default.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/CosmosDb/QueryIteratorsIntegrations.cs
robertpi: I've split this into a common class which is then called from classes that are specific to the type of integration, this is similar to the way GraphQL and a few other integrations are implemented. 

Using methods specific to each method signature and `TryDuckCast` reduces nesting.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/CosmosDb/QueryIteratorsIntegrations.cs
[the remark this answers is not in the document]
robertpi: I've moved everything over to use structs.

tonyredondo (approved):
LGTM!

lucaspimentel:
@robertpi: can you update the PR description to include the span metadata that this integrations will generate? Fields like service name, ~operation~ span name, resource name, and other tags. See #1471 for an example. Thanks!

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/CosmosDb/QueryIteratorsIntegrations.cs
lucaspimentel: Did we make any change here? If not, can you explain? Thanks. (unresolving)

src/Datadog.Trace/SpanTypes.cs
```diff
@@ -66,5 +66,10 @@ public static class SpanTypes
         /// The span type for msbuild integration.
         /// </summary>
         public const string Build = "build";
+
+        /// <summary>
+        /// The span type for CosmosDb integration.
+        /// </summary>
+        public const string CosmosDb = "cosmosdb";
```
lucaspimentel, line 73: As mentioned in https://github.com/DataDog/dd-trace-dotnet/pull/1473#discussion_r635485072, we should consider using the `sql` span type for this integration.

lucaspimentel (changes requested):
See my comments about possibly using the `sql` span type.

src/Datadog.Trace/SpanTypes.cs
robertpi: I already moved to the sql span type, I forgot to remove this. Fixed in the latest version.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/CosmosDb/CosmosCommon.cs
```diff
@@ -0,0 +1,174 @@
...
+                string containerId = properties.Item1;
+                string databaseId = properties.Item2;
+                string endpoint = properties.Item3;
+
+                if (endpoint == null)
+                {
+                    Console.WriteLine();
+                }
+
+                var tags = new CosmosDbTags
```
lucaspimentel, line 92: We should call `InstrumentationTags.SetAnalyticsSampleRate()` from every integration (with `enabledWithGlobalSetting: false` in this case). See an example here:
https://github.com/DataDog/dd-trace-dotnet/blob/b68cdfef714b378cd61ced1fa2894e0e215d90a9/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/GraphQL/GraphQLCommon.cs#L52

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/CosmosDb/CosmosCommon.cs
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        public static CallTargetState CreateClientCallStateExt<TTarget, TQueryDefinition>(TTarget instance, TQueryDefinition queryDefinition)
+        {
+            return CreateCosmosDbCallState(GetClientPropterties, instance, queryDefinition);
+        }
+
+        private static CallTargetState CreateCosmosDbCallState<TTarget, TQueryDefinition>(Func<object, Tuple<string, string, string>> extractProperties, TTarget instance, TQueryDefinition queryDefinition)
+        {
+            if (!Tracer.Instance.Settings.IsIntegrationEnabled(IntegrationName))
```
lucaspimentel, line 48: We have some optimizations for this where you can use an `IntegrationInfo` instead of a string. IIRC the goal is to reduce heap allocations on every call to the integration.

Example usage:
https://github.com/DataDog/dd-trace-dotnet/blob/b68cdfef714b378cd61ced1fa2894e0e215d90a9/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/GraphQL/GraphQLCommon.cs#L21-L22
https://github.com/DataDog/dd-trace-dotnet/blob/b68cdfef714b378cd61ced1fa2894e0e215d90a9/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/GraphQL/GraphQLCommon.cs#L34

lucaspimentel (commented):
Looks like setting analytics sample rate is missing and there's a small optimization when checking if the integration is enabled.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/CosmosDb/CosmosCommon.cs
robertpi: I have add this, but isn't ensure this is called from every single integration a bit error prone? Could they be parameters in the base constructor tags for example?

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/CosmosDb/CosmosCommon.cs
```diff
@@ -0,0 +1,174 @@
...
+
+                var properties = extractProperties(instance);
+
+                string containerId = properties.Item1;
+                string databaseId = properties.Item2;
+                string endpoint = properties.Item3;
+
+                if (endpoint == null)
+                {
+                    Console.WriteLine();
```
lucaspimentel, line 89: Is this left over from previous testing code?

robertpi: Yes, removed.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/CosmosDb/CosmosCommon.cs
robertpi: Done.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/CosmosDb/CosmosCommon.cs
lucaspimentel: Yes, we should look at improving the dev experience here. It's gone through several iterations over time and we haven't gone back to clean it up 😅

lucaspimentel (approved):

PR #1476, opened 2021-05-19
Add Nuke build project and update consolidated pipeline

andrewlock:
This PR introduces [Nuke](https://nuke.build/) as a C# orchestrator for our build system. There are several longer term goals:

* **Consolidate the build definition**. Currently our build definition is scattered across multiple dockerfiles, bash scripts, in-line yaml definitions, and MS Build targets. This PR aims to move the vast majority of the orchestration to Nuke, with the pipeline only being responsible for direct interactions with Azure Devops e.g. uploading artifacts etc
* **Improve local testing** Currently it is hard to replicate failures in CI locally. By using Nuke targets, it should be easier to reproduce failures locally
* **Consolidate tools** We currently have several "utility" tools, e.g. `GeneratePackageVersions` etc. We can consolidate these into the Nuke build tool, which will make it easier for onboarding (as all tools are listed in one discoverable place).
* **Improve performance** Our builds currently have a lot of duplication in them. Moving to Nuke won't increase performance per-se, but it makes it easier for us to iterate on the build process in future to improve results.
* **Make it easier to port the build to other CI systems** By removing the reliance on Azure-Devops-specific tasks, it should be easier to move the build system to GitHub actions, GitLab etc.

This initial PR doesn't attempt to achieve all these goals. Rather it aims to (mostly) replicate our current build process in C#. It's a first pass attempt 🙂 

Caveats:
* The current build definition is quite verbose, we can almost certainly improve this by pulling out common code, just as we would for normal C#.
* There are still a lot of extra MSBuild Targets fiddling happening. Some of that may well need to stay, as we still want an easy "builds in VS" <kbd>F5</kbd> experience. There are almost certainly some things we can improve though, and being able to run CI builds locally will help with that.

## Getting Stated:

Depending on your platform (Windows/Linux/Mac) there are bootstrapper scripts for running Nuke in the root directory: _build.cmd_, _build.ps1_, _build.sh_. There is also a _docker.sh_ file that can be used to run Linux builds using the same docker files as the CI build. 

To see all the available targets, run `build.cmd --help`.

To clean out any old cruft, build the tracer home directory, and build and run the managed unit tests (for example) you could run `build.cmd Clean BuildTracerHome BuildAndRunManagedUnitTests`.

## Code organisation

* The Main build targets are listed in */build/_build/Build.cs*. Ideally it wouldn't be nested or named like this, but this was the easiest approach to get started. I'll likely move it when we clean up the old pipelines etc
* _Build.Steps.cs_ contains most of the specific build steps that go into the larger targets. This is very verbose at the moment, but can definitely be improved
* _Build.Utilities.cs_ contains a few extra helpful targets, not required by the build itself. E.g. scripts to add and remove libraries from the GAC (for testing), running a sample project with tracing enabled etc.
* Added new dockerfiles for linux rather than trying to change the old ones, to avoid breaking existing pipelines
* Added new docker-compose targets for now, to avoid breaking existing pipelines. The old ones can be cleaned up when we remove the old pipelines

@DataDog/apm-dotnet


docker.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+set -euox pipefail
```
zacharycmontoya, line 2: Do you still need this file? I see that `run-in-docker.yml` looks similar but doesn't end up using this

zacharycmontoya (approved):
This is great, thanks for putting this together! I'm looking forward to future improvements that this enables! 🚀

docker.sh
andrewlock: It's not needed for the build no, it's a convenience method for running the Linux builds locally 🙂 It's essentially the "dockerised" version of `./build.sh`

e.g. To test the Linux builds, I can open up wsl, and run `./docker.sh Clean BuildTracerHome PackageTracerHome` and it runs the same steps as `run-in-docker.yml`

I'm happy to remove it if people don't see the value, but I have been using it frequently! 🙂

zacharycmontoya: I'm happy to keep this, I just wanted to understand how it was used. Thanks for clarifying 😄

build/_build/docker/debian.dockerfile
```diff
@@ -0,0 +1,56 @@
...
+
+# Copy the build project in and build it
+COPY . /build
+RUN dotnet build /build
+WORKDIR /project
+
+FROM base as tester
+
+# Install .NET Core runtimes using install script
+# FYI there is no arm64 runtime available for .NET Core 2.1, so just install the .NET Core runtime in that case
```
lucaspimentel, line 39: I think some of these were meant to say "ASP.NET Core runtime" instead of ".NET Core runtime". (`aspnetcore` vs `dotnet`)
```suggestion
# Install ASP.NET Core runtimes using install script
# There is no arm64 runtime available for .NET Core 2.1, so just install the .NET Core runtime in that case
```

docker-compose.yml
```diff
@@ -579,6 +582,120 @@ services:
     - kafka-broker
     - kafka-zookeeper
 
+  IntegrationTests2:
```
lucaspimentel, line 585: `IntegrationTests2`? Consider renaming or at least adding a comment to explain why this one is just "2".

build/_build/Build.Utilities.cs
```diff
@@ -0,0 +1,167 @@
...
+        .Requires(() => SampleName)
+        .Requires(() => Framework)
+        .Requires(() => IsWin)
+        .Executes(() =>
+        {
+            var envVars = new Dictionary<string,string>()
+            {
+                {"COR_ENABLE_PROFILING", "1"},
+                {"COR_PROFILER", "{846F5F1C-F9AE-4B07-969E-05C26BC060D8}"},
+                {"COR_PROFILER_PATH", TracerHomeDirectory / "win-x64" / "Datadog.Trace.ClrProfiler.Native.dll" },
```
lucaspimentel, line 124:
```suggestion
                {"COR_PROFILER_PATH_32", TracerHomeDirectory / "win-x86" / "Datadog.Trace.ClrProfiler.Native.dll" },
                {"COR_PROFILER_PATH_64", TracerHomeDirectory / "win-x64" / "Datadog.Trace.ClrProfiler.Native.dll" },
```

build/_build/Build.Utilities.cs
```diff
@@ -0,0 +1,167 @@
...
+                {"CORECLR_ENABLE_PROFILING", "1"},
+                {"CORECLR_PROFILER", "{846F5F1C-F9AE-4B07-969E-05C26BC060D8}"},
+                {"CORECLR_PROFILER_PATH", TracerHomeDirectory / "win-x64" / "Datadog.Trace.ClrProfiler.Native.dll" },
```
lucaspimentel, line 127:
```suggestion
                {"CORECLR_PROFILER_PATH_32", TracerHomeDirectory / "win-x86" / "Datadog.Trace.ClrProfiler.Native.dll" },
                {"CORECLR_PROFILER_PATH_64", TracerHomeDirectory / "win-x64" / "Datadog.Trace.ClrProfiler.Native.dll" },
```

.azure-pipelines/steps/install-dotnet-5-sdk.yml
```diff
@@ -0,0 +1,7 @@
+steps:
+- task: UseDotNet@2
+  displayName: install dotnet core sdk 5
+  inputs:
+    packageType: sdk
+    version: $(dotnetCoreSdk5Version)
```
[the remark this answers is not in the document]
andrewlock: It could be, but as it's always the same, we just define it at the top level here https://github.com/DataDog/dd-trace-dotnet/blob/andrew/ultimate-nuke-2/.azure-pipelines/ultimate-pipeline.yml#L19 and it filters through. If it's undefined, this would break, so makes things _slightly_ less verbose

docker-compose.yml
lucaspimentel: Oh, I see in the PR description:
>Added new docker-compose targets for now, to avoid breaking existing pipelines. The old ones can be cleaned up when we remove the old pipelines

.azure-pipelines/ultimate-pipeline.yml
```diff
@@ -17,10 +17,15 @@ pr:
 variables:
   buildConfiguration: Release
   dotnetCoreSdk5Version: 5.0.103
+  relativeTracerHome: /src/bin/windows-tracer-home
```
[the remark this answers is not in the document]
andrewlock: I'm 100% behind these moves, but I tried to do all of that once before and ended up spinning my wheels a long time 😄 I'm inclined to merge it pretty much as-is, and address these extra things in a follow up PR. I also want to:

* Move most of the other tool projects into Nuke so it's a "one stop shop"
* Remove the old code
* Reduce the reliance on directory.build.props where not necessary
* Make speed improvements

The list goes on 😄

build/_build/Build.Utilities.cs
```diff
@@ -0,0 +1,167 @@
...
+        .Description("Runs an IIS sample app, enabling profiling.")
+        .Requires(() => SampleName)
+        .Requires(() => IsWin)
+        .Executes(() =>
+        {
+            var envVars = new Dictionary<string,string>(new ProcessStartInfo().Environment)
+            {
+                {"COR_ENABLE_PROFILING", "1"},
+                {"COR_PROFILER", "{846F5F1C-F9AE-4B07-969E-05C26BC060D8}"},
+                {"COR_PROFILER_PATH", TracerHomeDirectory / "win-x64" / "Datadog.Trace.ClrProfiler.Native.dll" },
```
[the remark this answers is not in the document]
andrewlock: Ah ok. This isn't alctually used in CI, it's just a utility for local dev, but I'll fix them

build/_build/Build.Steps.cs
```diff
@@ -0,0 +1,1002 @@
...
+                .SetProject(Solution.GetProject(Projects.ClrProfilerManaged))
+                .SetConfiguration(BuildConfiguration)
+                .EnableNoBuild()
+                .EnableNoRestore()
+                .CombineWith(TargetFrameworks, (p, framework) => p
+                    .SetFramework(framework)
+                    .SetOutput(TracerHomeDirectory / framework)));
+        });
+
+    Target PublishNativeProfilerWindows => _ => _
```
lucaspimentel, line 270: We want to move away from using the word "profiler" to avoid confusion with the Continuous Profiler product, so we should rename some steps. It doesn't need to be in this PR, but at some point before switching over completely to Nuke and the new pipeline. We haven't agreed on new names yet, but they should distinguish between the native and managed components of the Tracer, instead of "profiler" (native) vs "tracer" (managed).

andrewlock: Yep, no problem. Just need to decide what those names are 🙂

docker.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
andrewlock: Done in https://github.com/DataDog/dd-trace-dotnet/pull/1476/commits/58c21483cbf71eca06ae24eae32df154a0e36618

.azure-pipelines/ultimate-pipeline.yml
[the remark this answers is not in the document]
lucaspimentel: Sounds good, thanks!

docker-compose.yml
lucaspimentel: > I could rename it to `NukeIntegrationTests` for now 

If it's not too much trouble, thanks!

build/_build/Build.Steps.cs
lucaspimentel: Don't let this block this PR, though. I know the whole thing is still work in progress. Just something to keep in mind.

lucaspimentel (approved):

PR #1483, opened 2021-05-24
Add GH Action to auto-create benchmark branch

andrewlock:
When we push a tag, we want to auto-create the benchmarks branch. This should extract the version number for the tag, and create a corresponding benchmarks branch. The benchmarks branches are used by CI to show historical performance. We use a branch instead of keying directly off the tags, as we don't want to benchmark _every_ release, and so we can delete branches if applicable.

What's that? No, I haven't tested this 🙈 But I have checked the `action-create-branch` action looks legit at least!

@DataDog/apm-dotnet

### DIFF
--- a/.github/workflows/create_benchmarks_branch.yml
+++ b/.github/workflows/create_benchmarks_branch.yml
@@ -1,0 +1,22 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


tonyredondo (approved):
LGTM but... I hope it works, We trust you!. 🤣

zacharycmontoya (approved):
This probably works 😄

PR #1487, opened 2021-05-25
Add DuckInclude attribute

kevingosse:
Add a DuckInclude attribute to force the ducktyping library to include overriden base object methods.

This will be used in log4net log injection to override ToString.


tonyredondo (approved):
LGTM

andrewlock (approved):

PR #1488, opened 2021-05-25
Temporal skip to avoid flakiness in xUnit 2.2.0 test

tonyredondo:
This is just a temporal workaround to avoid the flakiness in the pipeline. While the research to the proper fix is ongoing

@DataDog/apm-dotnet


kevingosse (approved):

andrewlock (approved):
Approved, by how Flakey is this test? And how slow is the startup? Is it possible to do a basic retry instead of skipping entirely? Whether that would be "better" or not I don't know 😄

tonyredondo:
> Approved, by how Flakey is this test? And how slow is the startup? Is it possible to do a basic retry instead of skipping entirely? Whether that would be "better" or not I don't know

Well I tought it was fixed already but it just happened to @kevingosse in his branch, I don’t know how flakey it is. Is related to the flushing issue, apparantly the test host finishes before the flushing finishes, so it’s a race condition.

PR #1493, opened 2021-05-25
Refactor DisposeWithException invocations to remove null conditional operator

zacharycmontoya:
The DisposeWithException extension method already provides a null-check so this `?.` is unnecessary. Removing them makes all the method invocations consistent.

@DataDog/apm-dotnet


tonyredondo (approved):

kevingosse (approved):

andrewlock (approved):
Thanks for doing this!

PR #1497, opened 2021-05-27
Remove a few benchmarks to make the pipeline faster

kevingosse:
Benchmarks removed:
 - WriteAndFlushTraces / WriteAndFlushTracesWithOrigin / WriteAndFlushEnrichedTracesWithOrigin (keeping only WriteAndFlushEnrichedTraces)
 - Log for NLog/log4net/Serilog (keeping only EnrichedLog)

### DIFF
--- a/test/benchmarks/Benchmarks.Trace/AgentWriterBenchmark.cs
+++ b/test/benchmarks/Benchmarks.Trace/AgentWriterBenchmark.cs
@@ -17,10 +17,7 @@ namespace Benchmarks.Trace
         private const int SpanCount = 1000;
 
         private static readonly IAgentWriter AgentWriter;
-        private static readonly Span[] Spans;
         private static readonly Span[] EnrichedSpans;
-        private static readonly Span[] SpansWithOrigin;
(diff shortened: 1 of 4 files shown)


tonyredondo (approved):

lucaspimentel (approved):

andrewlock (approved):
Just wondering, is it worth keeping these benchmarks but only running them in the nightly builds, instead of for PRs? 

BenchmarkDotNet [supports categories](https://benchmarkdotnet.org/articles/samples/IntroCategories.html), so we could get the benefits of catching regressions on master by running a larger benchmark suite, but keep the overhead down for PRs?

kevingosse:
> Just wondering, is it worth keeping these benchmarks but only running them in the nightly builds, instead of for PRs?
> 
> BenchmarkDotNet [supports categories](https://benchmarkdotnet.org/articles/samples/IntroCategories.html), so we could get the benefits of catching regressions on master by running a larger benchmark suite, but keep the overhead down for PRs?

Well I don't think those ones bring any value, but for the other benchmarks it could be an option yes

PR #1499, opened 2021-05-28
Cleanup project files (redux)

lucaspimentel:
Follow-up from #1468 to clean up a few things I missed
- Remove more msbuild property conditions
  - and add one to avoid building .NET Framework apps if they won't be used
- Remove more redundant `Directory.Build.props` files
- Remove `<RuntimeIdentifier>`/`<RuntimeIdentifiers>`

For files that need to import `Directory.Build.props` from parent directories, use the snippet recommended in 
https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build
```xml
<Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" />
```


test/test-applications/regression/Directory.Build.props
```diff
@@ -1,3 +0,0 @@
-<Project>
-  <Import Project="..\Directory.Build.props" />
-</Project>
```
andrewlock, line 3: Can you delete these references from the sln files too - there's a few "missing files" in solution explorer since the last cleanup too

andrewlock (approved):

test/test-applications/regression/Directory.Build.props
lucaspimentel: oops, thanks!

lucaspimentel: Fixed in e9a092c7ec1846561ccc80477d8268027a030cf9

tonyredondo (approved):

PR #1501, opened 2021-05-31
AgentWriter FlushTracesAsync changes

tonyredondo:
This PR's changes the behavior of the FlushTracesAsync method to enqueue flush continuations per buffer.

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/Common.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/Common.cs
@@ -76,11 +76,6 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Testing
                     // So the last spans in buffer aren't send to the agent.
                     Log.Debug("Integration flushing spans.");
                     await TestTracer.FlushAsync().ConfigureAwait(false);
-                    // The current agent writer FlushAsync method can return inmediately if a payload is being sent (there is buffer lock)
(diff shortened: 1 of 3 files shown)


kevingosse (approved):
LGTM. Any chance you could write a unit test that reproduces the flush race condition that you're protecting against? (I'm not sure it's possible, so I'm approving this PR either way)

tonyredondo:
> LGTM. Any chance you could write a unit test that reproduces the flush race condition that you're protecting against? (I'm not sure it's possible, so I'm approving this PR either way)

I'll try, but that's hard.

kevingosse:
> > LGTM. Any chance you could write a unit test that reproduces the flush race condition that you're protecting against? (I'm not sure it's possible, so I'm approving this PR either way)
> 
> I'll try, but that's hard.

I think you can mock the `IApi` and use that to block the first flush. Then invoke another flush. I _think_ before your PR the other flush would complete synchronously.

tonyredondo:
> > > LGTM. Any chance you could write a unit test that reproduces the flush race condition that you're protecting against? (I'm not sure it's possible, so I'm approving this PR either way)
> > 
> > 
> > I'll try, but that's hard.
> 
> I think you can mock the `IApi` and use that to block the first flush. Then invoke another flush. I _think_ that before your PR the other flush would complete synchronously without calling `IApi` again.

Done in https://github.com/DataDog/dd-trace-dotnet/pull/1501/commits/1208bf2bae3a71e06d67cb300a03ce46ac4d13d4

zacharycmontoya (approved):

PR #1502, opened 2021-05-31
Avoid the trace copy when flushing spans

kevingosse:
When flushing spans, we make a copy of the span array because `List<T>` gives no mean to fetch the internal buffer. This PR introduces a custom `ArrayBuilder<T>` that uses the same growing mechanics as `List<T>`, but exposes the internal buffer.

Benchmarks:

|               Method |     Mean |    Error |   StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------------- |---------:|---------:|---------:|-------:|------:|------:|----------:|
|  StartFinishSpan old | 770.4 ns | 16.65 ns | 48.82 ns | 0.0048 |     - |     - |     488 B |
|  StartFinishSpan new | 554.8 ns | 10.28 ns |  8.59 ns | 0.0048 |     - |     - |     432 B |
| StartFinishScope old | 885.6 ns | 24.11 ns | 69.95 ns | 0.0076 |     - |     - |     608 B |
| StartFinishScope new | 655.1 ns | 13.12 ns | 15.61 ns | 0.0057 |     - |     - |     552 B |

### DIFF
--- a/test/Datadog.Trace.Tests/Util/ArrayBuilderTests.cs
+++ b/test/Datadog.Trace.Tests/Util/ArrayBuilderTests.cs
@@ -32,7 +32,7 @@ namespace Datadog.Trace.Tests.Util
 
             var result = _builder.GetArray();
 
-            result.Should().BeEquivalentTo(Enumerable.Range(0, 20));
+            result.Should().BeEquivalentTo(Enumerable.Range(0, numberOfElements));
         }
 
         [Fact]
--- a/test/Datadog.Trace.Tests/Util/ArrayBuilderTests.cs
+++ b/test/Datadog.Trace.Tests/Util/ArrayBuilderTests.cs
@@ -47,5 +47,41 @@ namespace Datadog.Trace.Tests.Util
             result.Count.Should().Be(0);
             result.Array.Length.Should().Be(numberOfElements);
         }
+
+        [Fact]
+        public void Empty()
+        {
+            ArrayBuilder<int> builder = default;
+
+            var result = builder.GetArray();
+
+            result.Count.Should().Be(0);
+        }
+
+        [Fact]
+        public void DoubleSizeWhenGrowing()
+        {
+            ArrayBuilder<int> builder = default;
+
+            builder.GetArray().Array.Should().BeEmpty();
+
+            for (int i = 0; i < 4; i++)
+            {
+                builder.Add(i);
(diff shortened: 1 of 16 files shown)


tonyredondo (approved):
LGTM nice improvement.

test/Datadog.Trace.Tests/Util/ArrayBuilderTests.cs
```diff
@@ -0,0 +1,51 @@
...
+
+            _builder.Count.Should().Be(numberOfElements);
+
+            var result = _builder.GetArray();
+
+            result.Should().BeEquivalentTo(Enumerable.Range(0, numberOfElements));
+        }
+
+        [Fact]
+        public void InitialCapacity()
```
zacharycmontoya, line 39: Can you also write a test asserting how the array builder grows?

zacharycmontoya (approved):
Good stuff 👍🏼

lucaspimentel:
Unsolicited throwaway ideas for possible future improvements:
- use telemetry (when we have it) to choose a better default size (is 4 too small?)
- make the array size configurable by advanced users. This could reduce allocations consistently large traces. (for example, why start at 4 if you know every trace will have at least 30 spans?)
- get fancy and determine the best size dynamically based on trace history (for example, if the last 10 traces had ~100 spans, start with an array of size 100. Too much? 😅)

lucaspimentel:
What list size did we use for the benchmarks in the PR description? Should we run multiple tests at different trace sizes?

Currently we reuse the same list after calling `List<T>.Clear()`, which keeps the current capacity. This means that once we're "warmed up" and created larger traces, we rarely need to grow the list again. For each trace, we allocate an additional array for the copy (`List<T>.ToArray()`) before calling `Clear()`.

For users with consistently large traces, I expect the new `ArraySegment` will result in more allocations and more copying arrays around since the array needs to grow many more times.

lucaspimentel (approved):

test/Datadog.Trace.Tests/Util/ArrayBuilderTests.cs
kevingosse: Good call, I actually found a bug while doing this.

Fixed in https://github.com/DataDog/dd-trace-dotnet/pull/1502/commits/8c4919102d2ff6f1ad894c3639fcf28a3c4c8423

PR #1504, opened 2021-06-01
CIApp: Correctness and stability fixes.

tonyredondo:
- Fixes a scenario when a NUnit test suite fails on the OneTimeSetUp method gets reported different than the console output.
- Improves render of parameter values.

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/NUnit/IResultState.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/NUnit/IResultState.cs
@@ -17,12 +17,6 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Testing.NUnit
         TestStatus Status { get; }
 
         /// <summary>
-        /// Gets the label under which this test result is
-        /// categorized, or <see cref="string.Empty"/> if none.
-        /// </summary>
-        string Label { get; }
-
-        /// <summary>
         /// Gets the stage of test execution in which
         /// the failure or other result took place.
         /// </summary>
--- a/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/NUnit/IResultState.cs
+++ b/src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/NUnit/IResultState.cs
@@ -1,0 +1,31 @@
+// <copyright file="IResultState.cs" company="Datadog">
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2 License.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
+// </copyright>
+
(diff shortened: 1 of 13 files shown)


src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/NUnit/IResultState.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// Gets the TestStatus for the test.
+        /// </summary>
+        /// <value>The status.</value>
+        TestStatus Status { get; }
+
+        /// <summary>
+        /// Gets the label under which this test result is
+        /// categorized, or <see cref="string.Empty"/> if none.
+        /// </summary>
+        string Label { get; }
```
andrewlock, line 23: I don't think this property is used, do you need it?

andrewlock (approved):

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/NUnit/IResultState.cs
tonyredondo: you're right, removing it.

tonyredondo: done in fd3279a

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/Testing/NUnit/NUnitCompositeWorkItemSkipChildrenIntegration.cs
```diff
@@ -36,6 +40,60 @@ public class NUnitCompositeWorkItemSkipChildrenIntegration
...
+                        {
+                            if (item.GetType().Name == "CompositeWorkItem")
+                            {
+                                var compositeWorkItem2 = item.DuckCast<ICompositeWorkItem>();
+                                foreach (var item2 in compositeWorkItem2.Children)
+                                {
+                                    var testResult = item2.DuckCast<IWorkItem>().Result;
+                                    Scope scope = NUnitIntegration.CreateScope(testResult.Test, typeof(TTarget));
+                                    scope.Span.Error = true;
+                                    scope.Span.SetTag(Tags.ErrorMsg, compositeWorkItem.Result.Message);
```
zacharycmontoya, line 64: Just to double check, did you intend for these tags to come from `compositeWorkItem` or did you mean to use `compositeWorkItem2`?

tonyredondo: yes, the first `compositeWorkItem` contains the actual exception from the `OneTimeSetup` method, so what I do is to loop over the children and tag with the same exception from the setup method.

zacharycmontoya (approved):

PR #1510, opened 2021-06-02
Fix race condition in runner pipeline

kevingosse:
This should fix the `Artifact linux-tracer-home_arm64 was not found for build xxxxx` error.


lucaspimentel (approved):

andrewlock (approved):

tonyredondo (approved):
ups, my fault on this one. Thanks for the fix!

PR #1511, opened 2021-06-02
Remove callsite benchmarks and set iteration time back to 2 seconds

kevingosse:
We should keep the tests until callsite is completely deprecated, but there's no point in keeping the benchmarks. Removing them should give some relief to the CI, and allows to restore the old iteration time.

### DIFF
--- a/.azure-pipelines/benchmarks.yml
+++ b/.azure-pipelines/benchmarks.yml
@@ -68,7 +68,7 @@ stages:
       inputs:
         command: 'run'
         projects: '$(System.DefaultWorkingDirectory)/test/benchmarks/Benchmarks.Trace/Benchmarks.Trace.csproj'
-        arguments: '-c Release -f netcoreapp3.1 -- -r net472 netcoreapp3.1 -m -f * --iterationTime 1500'
+        arguments: '-c Release -f netcoreapp3.1 -- -r net472 netcoreapp3.1 -m -f * --iterationTime 2000'
       env:
         DD_ENV: CI
         DD_SERVICE: dd-trace-dotnet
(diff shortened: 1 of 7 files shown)


.azure-pipelines/benchmarks.yml
```diff
@@ -80,7 +80,7 @@ stages:
       inputs:
         command: 'run'
         projects: '$(System.DefaultWorkingDirectory)/test/benchmarks/Benchmarks.Trace/Benchmarks.Trace.csproj'
-        arguments: '-c Release -f netcoreapp3.1 -- -r net472 netcoreapp3.1 -m -f * --iterationTime 1500'
+        arguments: '-c Release -f netcoreapp3.1 -- -r net472 netcoreapp3.1 -m -f * --iterationTime 2000'
```
andrewlock, line 83: This needs to be updated in the ultimate-pipeline too

andrewlock (approved):
Good call. Good ratio of green/red there! 

A couple of related thoughts:
* Is the name change going to cause confusion comparing historically? Probably not, given we re-run these every night anyway?
* Should we backport this to the benchmark branches?
* Should we do something similar for throughput tests?
* Do we want to trim down which benchmark branches we run (do they all use call target currently)?

kevingosse:
> * Is the name change going to cause confusion comparing historically? Probably not, given we re-run these every night anyway?

As you pointed out, I don't think it will be an issue because we re-run them.

> * Should we backport this to the benchmark branches?

Yes. 

> * Should we do something similar for throughput tests?

Now that you mention it, probably yes. Let's do that in a separate PR though.

> * Do we want to trim down which benchmark branches we run (do they all use call target currently)?

1.17.0 does not implement calltarget, so it's time to archive it 😃

.azure-pipelines/benchmarks.yml
kevingosse: Looks like I forgot to decrease the iteration time in the ultimate-pipeline, so I don't need to revert it 😄

tonyredondo (approved):
LGTM

PR #1512, opened 2021-06-02
Convert GitLab package signing build to use Nuke targets

andrewlock:
Instead of using python for the package signing, use a Nuke target instead. Makes for easier local testing of the signing process too.

@DataDog/apm-dotnet


.gitlab-ci.yml
```diff
@@ -3,7 +3,7 @@ stages:
   - binary_build
   
 variables:
-  DATADOG_AGENT_WINBUILDIMAGES: v3946939-487746f
+  DATADOG_AGENT_WINBUILDIMAGES: v4603787-d533a1c
```
andrewlock, line 6: Docker image didn't have web targets installed. As I want to re-use the same build steps in both environments, I needed to add them. Created a PR (https://github.com/DataDog/datadog-agent-buildimages/pull/135) to the Windows Build image to add them.

build/_build/gitlab.bat
```diff
@@ -0,0 +1,16 @@
+if not exist c:\mnt\ goto nomntdir
+@echo PARAMS %*
+
+call "%VSTUDIO_ROOT%\vc\auxiliary\build\vcvars64.bat"
```
andrewlock, line 4: The bat file calls _vcvars64.bat_ to set the VS C++ compiler bits, which sets a bunch of environment variables. I couldn't find a nice way of doing that outside of the approach used here (and used previously)

build/_build/DotNetSettingsExtensions.cs
```diff
@@ -88,4 +97,16 @@ public static DotNetTestSettings EnableTrxLogOutput(this DotNetTestSettings sett
...
+    /// <summary>
+    /// GitLab installs MSBuild in a non-standard place that causes issues for Nuke trying to resolve it
+    /// </summary>
+    public static MSBuildSettings SetMSBuildPath(this MSBuildSettings settings)
+    {
+        var vsRoot = Environment.GetEnvironmentVariable("VSTUDIO_ROOT");
+
+        return settings
+            .When(!string.IsNullOrEmpty(vsRoot),
+                c => c.SetProcessToolPath(Path.Combine(vsRoot, "MSBuild", "Current", "Bin", "MSBuild.exe")));
```
andrewlock, line 110: The Windows Build image installs MSBuild in a weird location, so Nuke can't find it. Have to point it to the right place based on the presence of an environment variable. Not the nicest, but it works

build/_build/Build.Gitlab.cs
```diff
@@ -0,0 +1,144 @@
...
+            try
+            {
+                var (certPath, certPassword) = UseTestPfxCertificate
+                    ? (@"test_cert.pfx", "Passw0rd")
+                    : await GetSigningMaterial(tempFileName);
+
+                Logger.Info("Signing material retrieved, signing binaries...");
+                TracerHomeDirectory.GlobFiles("**/Datadog*.dll")
+                    .Concat(ArtifactsDirectory.GlobFiles("**/*.msi"))
+                    .ForEach(binaryPath => SignBinary(certPath, certPassword, binaryPath));
```
lucaspimentel, line 46: It looks like this step signs both dll and msi files, which means the msi was already built with unsigned dlls. Is that correct? What we want to do is:
1. build dlls
2. sign dlls
3. build msi (using the signed dlls from step 2)
4. build nuget (using the signed dlls from step 2)

andrewlock: Good catch. I was tricked by undescriptive msbuild target names 😅 

I thought this built the MSI because it defines MSI variables:
https://github.com/DataDog/dd-trace-dotnet/blob/84d1f1b9d8fff82eb2c4e7aba88c5b357ad35195/tasks/dotnettracer.py#L41

Turns out we don't build it until here (after signing dlls) 
https://github.com/DataDog/dd-trace-dotnet/blob/84d1f1b9d8fff82eb2c4e7aba88c5b357ad35195/tasks/dotnettracer.py#L67-L75

I'll update the PR 👍

andrewlock: Done in https://github.com/DataDog/dd-trace-dotnet/pull/1512/commits/f3b5703dc9731b3b51c55be11c5cd2bfcff90dd9

Sequence looks better now:

```
════════════════════════════════════════════════════
Target                          Status      Duration
────────────────────────────────────────────────────
Restore                         Executed        0:16
CreateRequiredDirectories       Executed      < 1sec
CompileManagedSrc               Executed        0:19
PublishManagedProfiler          Executed        0:06
CompileNativeSrcMacOs           Skipped                // EnvironmentInfo.IsOsx
CompileNativeSrcLinux           Skipped                // EnvironmentInfo.IsLinux
CompileNativeSrcWindows         Executed        0:02
PublishNativeProfilerMacOs      Skipped                // EnvironmentInfo.IsOsx
PublishNativeProfilerLinux      Skipped                // EnvironmentInfo.IsLinux
PublishNativeProfilerWindows    Executed      < 1sec
CopyIntegrationsJson            Executed      < 1sec
SignDlls                        Executed        0:13
ZipTracerHome                   Executed        0:02
BuildMsi                        Executed        0:15
PackNuGet                       Executed        0:02
SignMsiAndNupkg                 Executed        0:03
────────────────────────────────────────────────────
Total                                           1:21
════════════════════════════════════════════════════
```

lucaspimentel (approved):
LGTM. We should confirm we're getting the expected output from gitlab before merging this.

zacharycmontoya (approved):
LGTM 👍🏼  Like Lucas said, would be good to verify the output before merging

andrewlock:
> LGTM. We should confirm we're getting the expected output from gitlab before merging this.

Verified the msi files, and the files _installed_ by the msi using:

```bash
> signtool verify /pa .\datadog-dotnet-apm-1.27.0-x64.msi
File: .\datadog-dotnet-apm-1.27.0-x64.msi
Index  Algorithm  Timestamp
========================================
0      sha1       Authenticode
```

I verified the signature on:

- Both installers
- The 64 bit (dlls) installed for every framework version by the x64 installer 
- The x86 bit native dll installed by the x64 installer

The signing certificate was the same both before and after this change
image


Didn't install the x86 installer, but I'm pretty comfortable it's using the right dlls there

Other notes:

* Neither the previous or current pipelines were signing the dlls included in the NuGet packages - we were signing the dlls that are in the _publish_ folder, but `dotnet nuget publish` uses the binaries in the _bin_ folder. Fixed that in https://github.com/DataDog/dd-trace-dotnet/pull/1512/commits/9665988935b5deaf789cd3a607dd7252b32e99b6
* Confirmed the dlls in the tracer-home.zip are signed, but noticed we were using the wrong filename here (`tracer-home.zip` vs `windows-tracer-home.zip`. Fixed in https://github.com/DataDog/dd-trace-dotnet/pull/1512/commits/de9bfc568ba2d3129640cfdbb52fb519c6fc1996
* I didn't realise that we ship _System.Diagnostics.DiagnosticSource.dll_, _System.Threading.dll_ etc as part of the netstandard msi!

PR #1513, opened 2021-06-02
Delete old unused pipelines and files

andrewlock:
> We'll need to remove the `unit-tests` check requirement when we merge this. Not sure if we can introduce a similar requirement for the "unit test" portion of the consolidated pipeline?

Deletes all the old docker-compose targets, dockerfiles, bash scripts, azure devops pipelines and MSBuild targets that were superseded by the Nuke build in #1476.

~We probably don't want to merge this just yet, until we've verified the Nuke build is functioning as expected. It may be worth manually disabling the other pipelines in AzDo, running with that for a week or so, and then merging this once we're sure it's all working correctly?~ It's been a couple of weeks, and I recently discovered the integrations test pipeline wasn't working as expected, so it may be time...


kevingosse (approved):

Datadog.Trace.sln
```diff
@@ -196,42 +196,6 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Datadog.Trace.ClrProfiler.M
...
-		build\docker\package.alpine.dockerfile = build\docker\package.alpine.dockerfile
-		build\docker\package.dockerfile = build\docker\package.dockerfile
-		build\docker\package.sh = build\docker\package.sh
-		build\docker\with-profiler-logs.bash = build\docker\with-profiler-logs.bash
-		build\docker\with-profiler.bash = build\docker\with-profiler.bash
-	EndProjectSection
-EndProject
-Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = ".azure-pipelines", ".azure-pipelines", "{C52D6695-4E05-4930-88F8-0EFF8056A967}"
-	ProjectSection(SolutionItems) = preProject
-		.azure-pipelines\benchmarks.yml = .azure-pipelines\benchmarks.yml
```
lucaspimentel, line 226: Should we add `ultimate-pipeline.yml` to the solution?

lucaspimentel (approved):

Datadog.Trace.sln
andrewlock: It's already part of the solution as it's automatically included by Nuke in the build project🙂 

image

lucaspimentel: Awesome, thanks!

PR #1514, opened 2021-06-03
Flush both span buffers in a FlushTracesAsync() call

tonyredondo:
`FlushTracesAsync()` is a public api to flush traces in the AgentWriter, this PR ensures that both front and back buffer are flushed when this method is called, to match the flush behavior of `FlushAndCloseAsync()`

@DataDog/apm-dotnet


andrewlock (approved):

kevingosse (approved):
"public", it's still internal 😛

PR #1515, opened 2021-06-03
Enable ultimate pipeline for PRs

andrewlock:
Disable the other pipelines for now (should still be able to run them manually). Left the unit tests and integration test pipelines enabled for now, but we may want to disable those too?

Also configured the ultimate pipeline to run the throughput and benchmark tests nightly.

@DataDog/apm-dotnet

### DIFF
--- a/.azure-pipelines/ultimate-pipeline.yml
+++ b/.azure-pipelines/ultimate-pipeline.yml
@@ -1,17 +1,15 @@
 trigger:
   branches:
     include:
-      - master # Only run on master for now
+      - master
+      - refs/tags/*
     exclude:
       - refs/pull/*/head
   paths:
     exclude:
       - docs/*
       - .github/*
-pr:
-  branches:
-    exclude: # Don't run PR validation for now
-    - '*'
+
 
 # Global variables
 variables:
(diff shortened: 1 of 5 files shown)


.azure-pipelines/ultimate-pipeline.yml
```diff
@@ -119,6 +126,7 @@ stages:
       artifact: build-linux-$(matrixName)-working-directory
 
 - stage: build_macos
+  condition: ne(variables['Build.Reason'], 'Schedule') # skip for scheduled builds
```
kevingosse, line 129: I would be nice to have more generic variables. I imagine that we will want to manually run Crank at some point

kevingosse (approved):

lucaspimentel (approved):

.azure-pipelines/ultimate-pipeline.yml
andrewlock: Added this option in https://github.com/DataDog/dd-trace-dotnet/pull/1515/commits/c9e9cb54e7a3b16234180611280bd728563f1335, so you can set the `run_benchmarks_only` variable when triggering the pipeline manually. Can also set the `skip_benchmarks` variable for the opposite effect
image

PR #1517, opened 2021-06-03
Modify GetReJITParameters callback method to never return E_FAIL HRESULT

tonyredondo:
This PR modifies all `E_FAIL` HRESULT return values from the `GetReJITParameters` execution path to `S_FALSE` due a possible deadlock in applications on old runtime versions, as explained here:

https://github.com/dotnet/runtime/blob/main/src/coreclr/vm/rejit.cpp#L1094-L1096

If we fail during the `GetReJITParameters()` we can end in an scenario like this:

`Try Rejit` -> `Failed` -> `Revert` -> `Try Rejit`-> `Failed`-> `Revert` -> ...

- Also adds a check to ensure the managed profiler is loaded before the rewriting process.

Note: Some `S_OK` were changed to `S_FALSE` to express the original intent.

@DataDog/apm-dotnet


zacharycmontoya (approved):
LGTM

kevingosse (approved):

lucaspimentel (approved):

PR #1520, opened 2021-06-04
Delete project NugetProfilerVersionMismatch

lucaspimentel:
#### Delete project `NugetProfilerVersionMismatch`

This project is not built or run anywhere (CI or locally), and the nuget package `Datadog.Trace.ClrProfiler.Managed` it references has been deprecated for a while.

I am adding new tests for assembly version mismatches as part of the work for #1443.

edit: this project was not in any solution file, either


lucaspimentel:
>Was this project even in the sln file? If so, can we delete it

It wasn't. Thanks for making sure the sln is always updated 👍🏽

kevingosse (approved):

PR #1521, opened 2021-06-07
Fix Nuke pipeline and remove some unused sample projects

andrewlock:


lucaspimentel (approved):
Thanks!

build/_build/Build.Steps.cs
```diff
@@ -363,30 +363,33 @@ partial class Build
...
+                if (DirectoryExists(newDir))
                 {
-                    Logger.Info($"Skipping '{target}' as already exists");
+                    Logger.Info($"Skipping '{newDir}' as already exists");
                 }
                 else
                 {
-                    CopyDirectoryRecursively(source, target, DirectoryExistsPolicy.Fail, FileExistsPolicy.Fail);
+                    EnsureExistingDirectory(newDir.Parent);
+                    Cmd.Value(arguments: $"cmd /c mklink /J \"{newDir}\" \"{existingDir}\"");
```
zacharycmontoya, line 392: Does the cmd command work on Linux?

zacharycmontoya (approved):

build/_build/Build.Steps.cs
andrewlock: It doesn't need to, this one is only called in `BuildWindowsIntegrationTests` 🙂

Yes, it kinda sucks that we have completely different builds on each architecture 😉

PR #1523, opened 2021-06-08
Fix CosomsDb tests and disable PR triggers for old pipelines

andrewlock:
Fix CosmosDb tests in unified pipeline 

Disable scheduled triggers for benchmark/throughput (run by ultimate pipeline).

Previous attempt just disabled for CI pushes, not PR validation

@DataDog/apm-dotnet

### DIFF
--- a/build/_build/Build.Steps.cs
+++ b/build/_build/Build.Steps.cs
@@ -981,6 +981,14 @@ partial class Build
             ParallelIntegrationTests.ForEach(EnsureResultsDirectory);
             ClrProfilerIntegrationTests.ForEach(EnsureResultsDirectory);
 
+
+            var filter = (string.IsNullOrEmpty(Filter), IsArm64) switch
+            {
+                (true, false) => "Category!=LinuxUnsupported",
+                (true, true) => "(Category!=ArmUnsupported)&(Category!=LinuxUnsupported",
+                _ => Filter
+            };
+
             try
             {
                 // Run these ones in parallel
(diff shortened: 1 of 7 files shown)


build/_build/Build.Steps.cs
```diff
@@ -980,6 +981,14 @@ projectPath switch
             ParallelIntegrationTests.ForEach(EnsureResultsDirectory);
             ClrProfilerIntegrationTests.ForEach(EnsureResultsDirectory);
 
+
+            var filter = (string.IsNullOrEmpty(Filter), IsArm64) switch
+            {
+                (true, false) => "Category!=LinuxUnsupported",
+                (true, true) => "(Category!=ArmUnsupported)&(Category!=LinuxUnsupported",
```
lucaspimentel, line 988: I think we should come back later and clean up all the test categories. Wouldn't it be easier if each test declared which platforms it supports instead of which ones is _doesn't_ support? Instead of the current "run this test if the platform is _not not_ supported". Backlog? 😅

lucaspimentel (approved):

build/_build/Build.Steps.cs
andrewlock: Well, that gets noisy quickly if most tests run on most platforms. And if we add a new platform, we have to remember to update every test, instead of just excluding the ones that break, so I'm not sure tbh🙂 Personally I like that we make tests opt-_out_ for a platform

kevingosse (approved):

PR #1524, opened 2021-06-09
Small build improvements

andrewlock:
- xunit/nunit/mstest samples are built using multi api so don't need to be built separately
- Fix occasional flaky build for xunit sample on arm64
- Ignore NuGet package related warnings in CompileMultiApiPackageVersionSamples to reduce noise
- Try and reduce noise in build by setting ARFLAGS for re2 build
- Minor globbing improvement in build file
- Fix AWS SQS tests

@DataDog/apm-dotnet

### DIFF
--- a/build/_build/Build.Steps.cs
+++ b/build/_build/Build.Steps.cs
@@ -918,6 +918,8 @@ partial class Build
             // Annoyingly this rebuilds everything again and again.
             var targets = new [] { "RestoreSamplesForPackageVersionsOnly", "RestoreAndBuildSamplesForPackageVersionsOnly" };
 
+            // /nowarn:NU1608 - Detected package version outside of dependency constraint
+            // /nowarn:NU1701 - Package 'x' was restored using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETCoreApp,Version=v2.1'.
             DotNetMSBuild(x => x
                 .SetTargetPath(MsBuildProject)
                 .SetConfiguration(BuildConfiguration)
@@ -925,6 +927,7 @@ partial class Build
                 .SetProperty("TargetFramework", Framework.ToString())
                 .SetProperty("ManagedProfilerOutputDirectory", TracerHomeDirectory)
                 .SetProperty("BuildInParallel", "true")
+                .SetProcessArgumentConfigurator(arg => arg.Add("/nowarn:NU1701 /nowarn:NU1608"))
                 .AddProcessEnvironmentVariable("TestAllPackageVersions", "true")
                 .When(TestAllPackageVersions, o => o.SetProperty("TestAllPackageVersions", "true"))
                 .CombineWith(targets, (c, target) => c.SetTargets(target))
--- a/build/_build/Build.Steps.cs
+++ b/build/_build/Build.Steps.cs
@@ -848,6 +848,9 @@ partial class Build
                 "Samples.Microsoft.Data.Sqlite",
                 "Samples.OracleMDA",
                 "Samples.OracleMDA.Core",
+                "Samples.XUnitTests",
+                "Samples.NUnitTests",
+                "Samples.MSTestTests",
             };
 
             var projectsToBuild = sampleProjects
--- a/build/_build/Build.Steps.cs
+++ b/build/_build/Build.Steps.cs
@@ -918,7 +918,6 @@ partial class Build
             // Annoyingly this rebuilds everything again and again.
             var targets = new [] { "RestoreSamplesForPackageVersionsOnly", "RestoreAndBuildSamplesForPackageVersionsOnly" };
 
-            // /nowarn:NU1608 - Detected package version outside of dependency constraint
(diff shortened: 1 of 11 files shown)


build/_build/Build.Steps.cs
```diff
@@ -915,13 +918,16 @@ projectPath switch
             // Annoyingly this rebuilds everything again and again.
             var targets = new [] { "RestoreSamplesForPackageVersionsOnly", "RestoreAndBuildSamplesForPackageVersionsOnly" };
 
+            // /nowarn:NU1608 - Detected package version outside of dependency constraint
+            // /nowarn:NU1701 - Package 'x' was restored using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETCoreApp,Version=v2.1'.
```
lucaspimentel, line 922: Should we ignore this warning? Why are we restoring a net461 package into netcore2.1 app?

test/test-applications/integrations/Samples.ServiceStack.Redis/Samples.ServiceStack.Redis.csproj
```diff
@@ -6,6 +6,8 @@
     <!-- Required to build multiple projects with the same Configuration|Platform -->
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <GenerateTargetFrameworkAttribute>false</GenerateTargetFrameworkAttribute>
+    <!-- Ignore Found conflicts between different versions of "System.ValueTuple" (and others) that could not be resolved.   -->
```
lucaspimentel, line 9: Should we try to fix these conflicts instead of ignoring the warnings?

test/test-applications/regression/StackExchange.Redis.StackOverflowException/StackExchange.Redis.StackOverflowException.csproj
```diff
@@ -3,6 +3,8 @@
   <PropertyGroup>
     <TargetFrameworks>net461;netcoreapp2.1;netcoreapp3.0;netcoreapp3.1;net5.0</TargetFrameworks>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
+    <!-- Ignore  Found conflicts between different versions of "System.Diagnostics.PerformanceCounter" that could not be resolved. -->
```
lucaspimentel, line 6: Should we try to fix these conflicts instead of ignoring the warnings?

lucaspimentel (commented):
Left some question about ignoring warnings vs fixing the underlying issue if possible.

build/_build/Build.Steps.cs
andrewlock: > Why are we restoring a net461 package into netcore2.1 app

A bunch of packages don't support .NET Core:

*  MySql.Data 6.9.12
* StackExchange.Redis 1.0.488
* MongoDB.Driver 2.0.2
* RabbitMQ.Client 3.6.9

Thanks to the .NET 461 compat shim, they _can_ still be used with .NET Core however. This isn't something we can "fix", other than not testing those packages, but given this is a "valid" setup, I think we should be testing it. Hence ignoring the warnings.

> nowarn:NU1608 - Detected package version outside of dependency constraint

This is an issue with the ES NEST library, again when we're testing multiple different versions. We could maybe remove the JSON.NET lib entirely and just reference it transitively. I'll give that a try.

test/test-applications/integrations/Samples.ServiceStack.Redis/Samples.ServiceStack.Redis.csproj
andrewlock: That would be great, but I'm not sure exactly why they're there. I _guess_ it's related to the .NET STandard 1.5 target  in StackExchange.Redis 1.2.6, but I'm not sure how to fix that? 🤷

build/_build/Build.Steps.cs
andrewlock: Realised I added the nowarn to Samples.Elasticsearch.V5 directly, so we don't need this "global" nowarn I don't think

lucaspimentel: Thanks for clarifying!

lucaspimentel (approved):

test/test-applications/integrations/Samples.ServiceStack.Redis/Samples.ServiceStack.Redis.csproj
andrewlock: Fixed in https://github.com/DataDog/dd-trace-dotnet/pull/1524/commits/95099cfe6984ee156419b8f2721be3327aa8b498

test/test-applications/regression/StackExchange.Redis.StackOverflowException/StackExchange.Redis.StackOverflowException.csproj
andrewlock: Fixed in https://github.com/DataDog/dd-trace-dotnet/pull/1524/commits/c2a1e0d0120ac8915d6ab42f40fd590cf9c6f42d

PR #1525, opened 2021-06-09
Add an OWIN test application to the integration-tests pipeline

andrewlock:
From #1440 by @zacharycmontoya:

> This addresses a gap in our AspNet testing. Even though we do not currently instrument the entrypoint of the OWIN pipeline, we still instrument WebApi2 in this context so we can test against this.

This rebases that PR which has got stale. I reworked the tests to use the new snapshot testing approach, and switched to theory tests instead of Fact (with class fixture)

@DataDog/apm-dotnet


kevingosse (approved):

test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNetCore/AspNetCoreMvcTestBase.cs
```diff
@@ -39,7 +39,7 @@ protected AspNetCoreMvcTestBase(string sampleAppName, ITestOutputHelper output,
             HttpClient.DefaultRequestHeaders.Add(HeaderName1WithMapping, HeaderValue1);
             HttpClient.DefaultRequestHeaders.Add(HeaderName2, HeaderValue2);
             SetEnvironmentVariable(ConfigurationKeys.HeaderTags, $"{HeaderName1UpperWithMapping}:{HeaderTagName1WithMapping},{HeaderName2},{HeaderName3}");
-            SetEnvironmentVariable(ConfigurationKeys.HttpServerErrorStatusCodes, "400-403, 500-501-234, s342, 500");
+            SetEnvironmentVariable(ConfigurationKeys.HttpServerErrorStatusCodes, "400-403, 500-501-234, s342, 500-503");
```
lucaspimentel, line 42: What do you think of removing the invalid values here? `500-501-234, s342`
We shouldn't be testing the parsing logic here for this setting. That should be a different unit test.
(feel free to punt this)

test/test-applications/integrations/Samples.Owin.WebApi2/Properties/launchSettings.json
```diff
@@ -0,0 +1,22 @@
+{
```
lucaspimentel, line 1: We can use this file in .NET Framework apps? Is it an OWIN thing?

test/test-applications/integrations/Samples.Owin.WebApi2/Program.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Threading;
+using Microsoft.Owin.Hosting;
+
+namespace Samples.Owin.WebApi2
+{
+    public class Program
+    {
+        private static string Host()
+        {
+            return Environment.GetEnvironmentVariable("ASPNETCORE_URLS") ?? "http://localhost:9000/";
```
lucaspimentel, line 12: `ASPNETCORE`?

test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNetCore/AspNetCoreMvcTestBase.cs
andrewlock: Yeah, I agree, we shouldn't be testing that here

test/test-applications/integrations/Samples.Owin.WebApi2/Properties/launchSettings.json
andrewlock: 🤷 😄 I assume it got created by accident, but given it has all the vars, I'm guessing it works? 🤔  Haven't tried it but I guess @zacharycmontoya did!

test/test-applications/integrations/Samples.Owin.WebApi2/Program.cs
andrewlock: Yeah, that's for simplicity because the test uses the `AspNetCoreMvcTestBase` class. It doesn't really matter what variable we use, and this makes it easy!

test/test-applications/integrations/Samples.Owin.WebApi2/Properties/launchSettings.json
zacharycmontoya: I wrote the OWIN application as an SDK app, so I think it works! An easy way to verify this is to delete it and see what happens 😄

lucaspimentel: Confirmed it works. Thanks!

lucaspimentel (approved):

test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNetCore/AspNetCoreMvcTestBase.cs
zacharycmontoya: Good point, feel free to remove the middle two groups

zacharycmontoya (approved):
LGTM. Thanks for reviving this work!

test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNetCore/AspNetCoreMvcTestBase.cs
andrewlock: I remove these invalid values in https://github.com/DataDog/dd-trace-dotnet/pull/1525/commits/7e5f8d3f94180528d461007e093201768485435b

Added them as a test case to `TracerSettingsTests.ParseHttpCodes` instead

lucaspimentel: Thanks!

PR #1529, opened 2021-06-11
Downgrade the version of cmake used for MacOs builds

andrewlock:
Version 3.20.x causes the loader dll and pdb to not be embedded correctly in the _.dylib_ file.
You can confirm this behaviour by running `xcrun size -x -l -m Datadog.Trace.ClrProfiler.Native.dylib` and checking for the `Segment binary` section. In a bad build (with 3.20.x) it will be missing.

@DataDog/apm-dotnet

### DIFF
--- a/.azure-pipelines/runner.yml
+++ b/.azure-pipelines/runner.yml
@@ -114,6 +114,7 @@ jobs:
   pool:
     vmImage: macOS-10.15
   steps:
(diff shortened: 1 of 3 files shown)


tonyredondo (approved):
Thank you!

lucaspimentel (approved):

PR #1547, opened 2021-06-16
Merge hotfix 1.27.1

lucaspimentel:


kevingosse:
Not sure why the integration tests are failing, I think you should try rebasing

zacharycmontoya (approved):

.azure-pipelines/ultimate-pipeline.yml
```diff
@@ -17,6 +19,8 @@ schedules:
     branches:
       include:
         - master
+        - release/*
+        - hotfix/*
```
andrewlock, line 23: I don't know if we want to necessarily run these branches every night? But maybe we do?

andrewlock (approved):

PR #1548, opened 2021-06-17
Use a UUID for runtime-id instead of container id

kevingosse:
Also add the runtime-id tag on top-level spans.

### DIFF
--- a/src/Datadog.Trace/Tagging/TagsList.cs
+++ b/src/Datadog.Trace/Tagging/TagsList.cs
@@ -17,6 +17,8 @@ namespace Datadog.Trace.Tagging
         private static byte[] _metaBytes = StringEncoding.UTF8.GetBytes("meta");
         private static byte[] _metricsBytes = StringEncoding.UTF8.GetBytes("metrics");
         private static byte[] _originBytes = StringEncoding.UTF8.GetBytes(Trace.Tags.Origin);
+        private static byte[] _runtimeIdBytes = StringEncoding.UTF8.GetBytes(Trace.Tags.RuntimeId);
+        private static byte[] _runtimeIdValueBytes = StringEncoding.UTF8.GetBytes(Tracer.RuntimeId);
 
         private List<KeyValuePair<string, double>> _metrics;
         private List<KeyValuePair<string, string>> _tags;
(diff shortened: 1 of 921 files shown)


andrewlock (approved):

src/Datadog.Trace/Tracer.cs
```diff
@@ -664,16 +671,10 @@ private static IDogStatsd CreateDogStatsdClient(TracerSettings settings, string
...
-                                       $"service:{serviceName}"
+                                       $"service:{serviceName}",
+                                       $"{Tags.RuntimeId}:{RuntimeId}"
                                    };
 
-                var containerId = ContainerMetadata.GetContainerId();
-
-                if (containerId != null)
-                {
-                    constantTags.Add($"runtime-id:{containerId}");
```
lucaspimentel, line 674: Should we keep this? (but as `container-id:{containerId}`

src/Datadog.Trace/Tagging/TagsList.cs
```diff
@@ -305,6 +307,13 @@ private int WriteTags(ref byte[] bytes, int offset, Span span)
                 }
             }
 
+            if (span.IsTopLevel)
+            {
+                count++;
+                offset += MessagePackBinary.WriteStringBytes(ref bytes, offset, _runtimeIdBytes);
```
lucaspimentel, line 313: I like this idea of adding tags during serialization instead of adding them to the in-memory object. We should adopt this for other cases, like global tags added to every span (`DD_TAGS`).

lucaspimentel (approved):
Left some non-blocking comments.

src/Datadog.Trace/Tracer.cs
kevingosse: I don't think so. I'm not aware of anything that will use it

colin-higgins (approved):

zacharycmontoya (approved):

src/Datadog.Trace/Tracer.cs
lucaspimentel: You're right. I thought you were removing this: https://github.com/DataDog/dd-trace-dotnet/blob/b68cdfef714b378cd61ced1fa2894e0e215d90a9/src/Datadog.Trace/Agent/Api.cs#L76

PR #1549, opened 2021-06-17
Add ASP.NET Core ILogger instrumentation for logs injection

andrewlock:
The initial implementation was done by @zacharycmontoya - I added tests and extended to cover .NET Core 3+

This instruments [the `HostingApplication.ProcessRequestAsync` method](https://github.com/dotnet/aspnetcore/blob/4fc4748c196de4bae8da1e146638f484e7e4b99c/src/Hosting/Hosting/src/Internal/HostingApplication.cs#L86-L89), which wraps execution of the middleware pipeline, _after the `DiagnosticSource.BeginRequest` has executed (and hence after a `TraceId` has been generated).

Limitations:

* The integration does not add `SpanId` to the logs. Doing so would require a different instrumentation point.
* The logs injection integration only works for ASP.NET Core requests.
* `TraceId` injection is disabled for .NET Framework. The integration is still enabled, but as the diagnostic-source based tracing doesn't work, the `trace_id` would always be `0`, which defeats the point of logs injection. `dd_service`, `dd_env`, and `dd_version` are added though.
 
As part of this integration, I also updated the 3.1 sample to run and test against .NET 5, and the 2.1 to test and run against .NET Framework (to confirm existing no-tracing/no-trace_id behaviour)

@DataDog/apm-dotnet


src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AspNetCore/ILoggerIntegration.cs
```diff
@@ -0,0 +1,86 @@
...
+        /// <param name="instance">Instance value, aka `this` of the instrumented method.</param>
+        /// <param name="context">The context of the request</param>
+        /// <returns>Calltarget state value</returns>
+        public static CallTargetState OnMethodBegin<TTarget, TContext>(TTarget instance, TContext context)
+            where TTarget : IHostingApplication
+        {
+            IDisposable disposable = null;
+            if (Tracer.Instance.Settings.LogsInjectionEnabled)
+            {
+                var logger = (ILogger)DuckType.Create(typeof(ILogger), instance.Diagnostics.Logger, ILoggerType);
```
andrewlock, line 59: Theoretically, this shouldn't be necessary, but if we make instance.Diagnostics an `ILogger` (instead of an `object`) and use chaining, then it fails. I can't work out why...

src/Datadog.Trace/AspNetCore/DatadogLoggingScope.cs
```diff
@@ -0,0 +1,101 @@
...
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Globalization;
+
+namespace Datadog.Trace.AspNetCore
+{
+    // Implementation based on AspNetCore Hosting implementation: https://github.com/dotnet/aspnetcore/blob/1a2b3260c6161ae9b7f639de228a6eb0488a1075/src/Hosting/Hosting/src/Internal/HostingLoggerExtensions.cs#L114
+    internal class DatadogLoggingScope : IReadOnlyList<KeyValuePair<string, object>>
```
andrewlock, line 14: Wonder if we should make this public, so customers can use it with manual instrumentation?

src/Datadog.Trace/AspNetCore/DatadogLoggingScope.cs
```diff
@@ -0,0 +1,101 @@
...
+            }
+        }
+
+        public KeyValuePair<string, object> this[int index]
+        {
+            get
+            {
+                if (index == 0)
+                {
+                    return new KeyValuePair<string, object>("dd_service", _service);
```
andrewlock, line 53: These identifiers are "serilog" style. If users are using ILogger injection _and_ NLog/Log4NET, then they will have two sets of identifiers I think. Not sure what the best way to tackle that is

src/Datadog.Trace/AspNetCore/DatadogLoggingScope.cs
zacharycmontoya: Oh, is this for non-aspnetcore use cases? I was thinking we would start by only exposing an ApplicationBuilder extension method that the user can call, and this could be an implementation detail.

andrewlock: Yeah, I was thinking non-aspnetcore use cases. The `ApplicationBuilder` makes sense, but then we would have to take dependencies on the aspnetcore packages, which I wasn't sure if we'd want to? We can always make this public later if we want to, so probably makes sense to keep it internal for now.

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AspNetCore/ILoggerIntegration.cs
andrewlock: Figured it out - it's due to the implementation in ASP.NET Core, which uses explicit interface implementation: 

https://github.com/dotnet/extensions/blob/v2.1.28/src/Logging/Logging.Abstractions/src/LoggerOfT.cs#L32-L35

Our DuckTyping fails on that currently.

andrewlock: Thanks to #1555, we can now use the simpler dock-type chaining approach (done in https://github.com/DataDog/dd-trace-dotnet/pull/1549/commits/0557c1aa8f5625963cb30905e10f8a2e5154fda6)

tonyredondo: I think you’re missing the wildcard: https://github.com/DataDog/dd-trace-dotnet/blob/master/test/Datadog.Trace.DuckTyping.Tests/DuckExplicitInterfaceTests.cs#L50

lucaspimentel: Can we resolve this thread?

src/Datadog.Trace/AspNetCore/DatadogLoggingScope.cs
lucaspimentel: >but then we would have to take dependencies on the aspnetcore packages

In a new [`Datadog.Trace.AspNetCore`](https://www.nuget.org/packages/Datadog.Trace.AspNetCore/) package! 😅

build/_build/Build.Steps.cs
```diff
@@ -882,7 +882,7 @@ projectPath switch
                     {
                         "Samples.AspNetCoreMvc21" => Framework == TargetFramework.NETCOREAPP2_1,
                         "Samples.AspNetCoreMvc30" => Framework == TargetFramework.NETCOREAPP3_0,
-                        "Samples.AspNetCoreMvc31" => Framework == TargetFramework.NETCOREAPP3_1,
+                        "Samples.AspNetCoreMvc31" => Framework == TargetFramework.NETCOREAPP3_1 || Framework == TargetFramework.NET5_0,
```
lucaspimentel, line 885: No `Samples.AspNetCoreMvc50`? 😝

src/Datadog.Trace/DuckTyping/DuckTypeExceptions.cs
```diff
@@ -56,6 +56,23 @@ internal static void Throw()
         }
     }
 
+    /// <summary>
+    /// DuckType target type instance is null
+    /// </summary>
+    public class DuckTypeTargetTypeDefinitionIsNull : DuckTypeException
+    {
+        private DuckTypeTargetTypeDefinitionIsNull()
+            : base($"The target type definition is null.")
```
lucaspimentel, line 65:
```suggestion
            : base("The target type definition is null.")
```

lucaspimentel (approved):

build/_build/Build.Steps.cs
andrewlock: It's a reasonable question, tbh, and given we have separate projects for 3.0 and 3.1, maybe we should?🤔

lucaspimentel: Not in this PR of course. To the backlog!

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AspNetCore/ILoggerIntegration.cs
andrewlock: Yep, the problem should be resolved by https://github.com/DataDog/dd-trace-dotnet/pull/1561, so going to rebase and check

andrewlock:
I wonder if we should provide a way to disable this log injection integration. It's written as part of the aspnetcore integration, but given it isn't totally complete (i.e. it only works for aspnetcore requests, not all `ILogger`, not manual instrumentation etc), it would add unnecessary  duplication if users are _also_ using serilog/nlog/log4net integration. 

Those other integrations can be "disabled" by just not wiring them up, but we don't have a way of disabling this `ILogger` integration, while keeping aspnetcore and Serilog/nlog/log4net log injection enabled. A simple "fix" would be to use a different `IntegrationName` for this, instead of bundling as part of `AspNetCore`?

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AspNetCore/IHostingApplication.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    /// <summary>
+    /// HostingApplication interface for ducktyping
+    /// </summary>
+    public interface IHostingApplication
+    {
+        /// <summary>
+        /// Gets the diagnostics field
+        /// </summary>
+        [Duck(Name = "_diagnostics", Kind = DuckKind.Field)]
```
zacharycmontoya, line 18: If #1601 is merged soon, then we can refactor this to

```csharp
[DuckField(Name = "_diagnostics")]
```

zacharycmontoya:
> I wonder if we should provide a way to disable this log injection integration. It's written as part of the aspnetcore integration, but given it isn't totally complete (i.e. it only works for aspnetcore requests, not all `ILogger`, not manual instrumentation etc), it would add unnecessary duplication if users are _also_ using serilog/nlog/log4net integration.
> 
> Those other integrations can be "disabled" by just not wiring them up, but we don't have a way of disabling this `ILogger` integration, while keeping aspnetcore and Serilog/nlog/log4net log injection enabled. A simple "fix" would be to use a different `IntegrationName` for this, instead of bundling as part of `AspNetCore`?

How about we change the `IntegrationName` to `ILogger`, since all of our automatic logs injection for ILogger will have to be done by automatic instrumentation?

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AspNetCore/ILoggerIntegration.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// <typeparam name="TTarget">Type of the target</typeparam>
+        /// <typeparam name="TContext">Type of the controller context</typeparam>
+        /// <param name="instance">Instance value, aka `this` of the instrumented method.</param>
+        /// <param name="context">The context of the request</param>
+        /// <returns>Calltarget state value</returns>
+        public static CallTargetState OnMethodBegin<TTarget, TContext>(TTarget instance, TContext context)
+            where TTarget : IHostingApplication
+        {
+            IDisposable disposable = null;
+            if (Tracer.Instance.Settings.LogsInjectionEnabled)
```
zacharycmontoya, line 50: I think we'll need to modify the check to also check the integration:
```csharp
if (Tracer.Instance.Settings.LogsInjectionEnabled && Tracer.Instance.Settings.IsIntegrationEnabled(IntegrationId))
```
paired with
```csharp
```

src/Datadog.Trace/AspNetCore/DatadogLoggingScope.cs
zacharycmontoya: Yeah they'll have two sets of identifiers...perhaps we'll have a better idea of how to tackle it once we see it in use?

zacharycmontoya (approved):
LGTM, and since I hadn't looked at this in a while I think I acquired an objective perspective to review it properly 😆

andrewlock:
Superseded by #1663. I'll pull the additional tests (for .NET Framework etc) out in a separate PR

PR #1556, opened 2021-06-18
Testing unusual Unicode sequences for game day

robertpi:
Fixes #

@DataDog/apm-dotnet


zacharycmontoya:
I think this could make for a great unit test instead of a standalone test app, but this is probably low priority at the moment

PR #1558, opened 2021-06-21
Fix issue with CustomSerilogLogProvider

zacharycmontoya:
Restrict usage of optimized `CustomSerilogLogProvider` to versions of Serilog where the `Serilog.Context.LogContext` type has the `Push` method that takes in exactly one `ILogEventEnricher`. This fixes a crash condition in earlier versions of the library where the method does not exist. In those cases, fall back to the original LibLog implementation where we call `PushProperty` on each field.

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace/Logging/CustomSerilogLogProvider.cs
+++ b/src/Datadog.Trace/Logging/CustomSerilogLogProvider.cs
@@ -5,6 +5,7 @@
 
 using System;
 using System.Linq.Expressions;
+using System.Reflection;
 using Datadog.Trace.Logging.LogProviders;
 
 namespace Datadog.Trace.Logging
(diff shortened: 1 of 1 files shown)


src/Datadog.Trace/Logging/CustomSerilogLogProvider.cs
```diff
@@ -27,15 +28,20 @@ public IDisposable OpenContext(object enricher)
...
             var ndcContextType = Type.GetType("Serilog.Context.LogContext, Serilog");
+            return ndcContextType?.GetMethod("Push", GetLogEnricherType());
+        }
 
-            var logEventEnricherType = GetLogEnricherType();
-
-            var pushPropertyMethod = ndcContextType.GetMethod("Push", logEventEnricherType);
+        private static Func<object, IDisposable> GetPush()
+        {
+            var pushPropertyMethod = GetPushMethodInfo();
```
lucaspimentel, line 42: Should we add more error-checking in case any of these reflection calls fail? e.g. if `pushPropertyMethod` is `null`?

zacharycmontoya: We could if we want to be extra careful...but I don't think it's needed. If `GetPushMethodInfo` returns `null`, it should never have gotten to this point because `IsLoggerAvailable()`, which is the first check before instantiating and using this logger, would have returned `false`

zacharycmontoya:
Superseded by #1560

PR #1560, opened 2021-06-21
Utilize lower-overhead CustomSerilogLogProvider for older LogContext.PushProperties API when LogContext.Push API is not present

zacharycmontoya:
This PR is an optimization of #1558. That previous PR uses the slow LibLog fallback for all Serilog versions where `LogContext.Push` is not present. This PR adds an optimization when `LogContext.PushProperties` is present by using the faster Enricher method, which requires us to wrap the enricher inside of an array.

## Performance Measurements
Newer versions of Serilog **(unchanged)**: `LogContext.Push(ILogEventEnricher)` vs `LogContext.PushProperty(string, object, bool)` -- using Serilog 2.10.0
|       CurrentLogProvider |     Toolchain |     Mean |     Error |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------------- |-------------- |---------:|----------:|----------:|-------:|------:|------:|----------:|
| CustomSerilogLogProvider |        net472 | 6.010 μs | 0.1181 μs | 0.2330 μs | 0.6027 |     - |     - |    2.8 KB |
|       SerilogLogProvider |        net472 | 7.939 μs | 0.1571 μs | 0.2303 μs | 1.1139 |     - |     - |   5.17 KB |
|                          |               |          |           |           |        |       |       |           |
| CustomSerilogLogProvider | netcoreapp3.1 | 5.041 μs | 0.0493 μs | 0.0437 μs | 0.0381 |     - |     - |   2.61 KB |
|       SerilogLogProvider | netcoreapp3.1 | 7.350 μs | 0.1103 μs | 0.0978 μs | 0.0839 |     - |     - |    5.6 KB |

Older versions of Serilog **(changed)**: `LogContext.PushProperties(ILogEventEnricher[])` vs `LogContext.PushProperty(string, object, bool)` -- using Serilog 2.0.0
|       CurrentLogProvider |     Toolchain |      Mean |     Error |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------------- |-------------- |----------:|----------:|----------:|-------:|------:|------:|----------:|
| CustomSerilogLogProvider |        net472 |  7.404 μs | 0.1424 μs | 0.1851 μs | 0.8240 |     - |     - |   3.82 KB |
|       SerilogLogProvider |        net472 | 12.412 μs | 0.1865 μs | 0.1653 μs | 1.4496 |     - |     - |   6.71 KB |
|                          |               |           |           |           |        |       |       |           |
| CustomSerilogLogProvider | netcoreapp3.1 |  6.038 μs | 0.1191 μs | 0.1958 μs | 0.0534 |     - |     - |   3.62 KB |
|       SerilogLogProvider | netcoreapp3.1 |  9.053 μs | 0.1771 μs | 0.3455 μs | 0.1068 |     - |     - |   7.23 KB |

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace/Logging/CustomSerilogLogProvider.cs
+++ b/src/Datadog.Trace/Logging/CustomSerilogLogProvider.cs
@@ -6,30 +6,47 @@
 using System;
 using System.Linq.Expressions;
 using System.Reflection;
+using System.Runtime.CompilerServices;
 using Datadog.Trace.Logging.LogProviders;
 
 namespace Datadog.Trace.Logging
 {
     internal class CustomSerilogLogProvider : SerilogLogProvider, ILogProviderWithEnricher
     {
+        private static ConditionalWeakTable<object, object> _enricherToArrayConditionalWeakTable;
         private static Func<object, IDisposable> _pushMethod;
+        private static Func<object, IDisposable> _openContext;
 
         public CustomSerilogLogProvider()
         {
-            _pushMethod = GetPush();
+            _enricherToArrayConditionalWeakTable = new();
+
+            var logEnricherType = GetLogEnricherType();
+            if (GetPushMethodInfo() != null)
+            {
+                _openContext = _pushMethod = GeneratePushDelegate(GetPushMethodInfo(), logEnricherType);
+            }
+            else
+            {
+                _pushMethod = GeneratePushDelegate(GetPushPropertiesMethodInfo(), logEnricherType.MakeArrayType());
+                _openContext = (enricher) =>
+                {
+                    // Use a conditional weak table to cache a map between the enricher and an array of size 1 containing the enricher,
+                    // since the API requires an argument of type ILogEventEnricher[]
+                    object properties = _enricherToArrayConditionalWeakTable.GetValue(key: enricher, createValueCallback: ConditionalWeakTableCallback);
+                    return _pushMethod(properties);
+                };
+            }
         }
 
-        public IDisposable OpenContext(object enricher)
-        {
-            return _pushMethod(enricher);
-        }
+        public IDisposable OpenContext(object enricher) => _openContext(enricher);
 
         public ILogEnricher CreateEnricher() => new SerilogEnricher(this);
 
         internal static Type GetLogEnricherType() => Type.GetType("Serilog.Core.ILogEventEnricher, Serilog");
 
         internal static new bool IsLoggerAvailable() =>
-            SerilogLogProvider.IsLoggerAvailable() && GetPushMethodInfo() != null;
+            SerilogLogProvider.IsLoggerAvailable() && (GetPushMethodInfo() != null || GetPushPropertiesMethodInfo() != null);
 
         private static MethodInfo GetPushMethodInfo()
         {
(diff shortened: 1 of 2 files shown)


lucaspimentel:
Does this PR supersede #1558? Or it is incremental (merge both)?

lucaspimentel (approved):

src/Datadog.Trace/Logging/CustomSerilogLogProvider.cs
```diff
@@ -5,44 +5,80 @@
...
         {
-            _pushMethod = GetPush();
-        }
+            _enricherToArrayConditionalWeakTable = new();
 
-        public IDisposable OpenContext(object enricher)
-        {
-            return _pushMethod(enricher);
+            var logEnricherType = GetLogEnricherType();
+            if (GetPushMethodInfo() != null)
```
kevingosse, line 25: You should store the return value of `GetPushMethodInfo` in a field to avoid calling it twice

src/Datadog.Trace/Logging/CustomSerilogLogProvider.cs
```diff
@@ -5,44 +5,80 @@
...
+                _openContext = _pushMethod = GeneratePushDelegate(GetPushMethodInfo(), logEnricherType);
+            }
+            else
+            {
+                _pushMethod = GeneratePushDelegate(GetPushPropertiesMethodInfo(), logEnricherType.MakeArrayType());
+                _openContext = (enricher) =>
+                {
+                    // Use a conditional weak table to cache a map between the enricher and an array of size 1 containing the enricher,
+                    // since the API requires an argument of type ILogEventEnricher[]
+                    object properties = _enricherToArrayConditionalWeakTable.GetValue(key: enricher, createValueCallback: ConditionalWeakTableCallback);
```
kevingosse, line 36: I think we can remove this whole ConditionalWeakTable by adding some logic in `SerilogEnricher` that knows whether to wrap the value in an array or not

zacharycmontoya: Done in 35efec447b758d7fd8fd7988af26a1447c2c8756 , let me know what you think

src/Datadog.Trace/Logging/CustomSerilogLogProvider.cs
zacharycmontoya: Done in 35efec447b758d7fd8fd7988af26a1447c2c8756

src/Datadog.Trace/Logging/CustomSerilogLogProvider.cs
```diff
@@ -5,44 +5,83 @@
...
+using System.Reflection;
+using System.Runtime.CompilerServices;
 using Datadog.Trace.Logging.LogProviders;
 
 namespace Datadog.Trace.Logging
 {
     internal class CustomSerilogLogProvider : SerilogLogProvider, ILogProviderWithEnricher
     {
         private static Func<object, IDisposable> _pushMethod;
+        private static NoOpDisposable _cachedDisposable;
```
kevingosse, line 17: You don't actually need to store it since it will be kept alive by the lambda closure

kevingosse (approved):
LGTM

src/Datadog.Trace/Logging/CustomSerilogLogProvider.cs
zacharycmontoya: So you're saying I could use a local and then reference it in my lambda? I just want to make sure the `_pushMethod` is returning the same object over and over, not creating a new one

lucaspimentel (approved):

src/Datadog.Trace/Logging/CustomSerilogLogProvider.cs
zacharycmontoya: I did that in fbc3aa2030f909177e874b4cb813f518fe687901 , let me know with a 👍🏼 if that looks fine to you. We shouldn't ever run into this scenario because of the new `IsLoggerAvailable()` check, but it's there just in case

zacharycmontoya:
> Does this PR supersede #1558? Or it is incremental (merge both)?

This PR *does* supersede the previous PR so I will close that if we merge this. The previous one was an unoptimized, quick patch fix. This one performs the fix *and* does it optimally 😄

src/Datadog.Trace/Logging/CustomSerilogLogProvider.cs
kevingosse: Yep, that's fine :)

PR #1561, opened 2021-06-23
Add support for NestedTypes and GenericParameters to EnsureTypeVisibility

andrewlock:
The `ILogger` instrumentation implementation in #1549 highlighted a case in the duck typing library that we weren't handling. If a duck-typed target is a generic, and has a non-public type as the generic argument the duck typing call would fail.

This PR adds support for non-public nested types and generic arguments.
It's tested with the specific example we encountered for `ILogger`. I've kept the additional exploratory tests @tonyredondo wrote initially too. 

@DataDog/apm-dotnet


tonyredondo (approved):
Nice! Thanks!

lucaspimentel (approved):

PR #1563, opened 2021-06-23
Fix x86 builds in consolidated pipeline

andrewlock:
This PR addresses two issues:
* The consolidated pipeline was not setting the build platform correctly
* The x86 builds have been broken for a long time (all pipelines)

The `x86` builds have been silently broken for a long time - the problem is that you can't test assemblies compiled for `x86` with the 64-bit `dotnet` SDK; you have to use the 32-bit SDK. Unfortunately, xunit just fails to load the assembly so doesn't _explicitly_ fail, it just bails out.

To run the x86 builds locally, you need to install the 32 bit .NET SDK (32 bit runtimes are already installed). When doing so, be careful that your Paths don't get messed up: make sure you are using the 64-bit version by default (check with `dotnet --info`). 

Nuke will automatically locate the 32 bit SDK for the x86 Windows Integration tests.

Note that we currently only run the x86 builds on Windows, so there are no changes required for Linux/mac

> I also looked into compiling the integration test project as AnyCPU, and only running the samples as x86. As the x86 _runtime_ is installed along with the x64 runtime, this means no further installs are required. However, it means we need to a) indicate which platform to run (previously we were relying on the `/platform` switch, b) we need to locate the x86 runtime version of `dotnet.exe` as appropriate. On top of this, the build gets more complex, because dotnet seems to hate it if you try and build _part_ of the solution as x64 (unless you do build all on the .sln file, which we don't)

@DataDog/apm-dotnet


.azure-pipelines/steps/install-dotnet-sdk-32bit.yml
```diff
@@ -0,0 +1,30 @@
+parameters:
+  - name: sdkVersion
+    type: string
+    default: ''
+
+  - name: channel
+    type: string
+    default: ''
```
lucaspimentel, line 8: This won't work as expected if you pass both a version and a channel. It will run two installed, one with each argument, instead of a single install with both.

lucaspimentel: Maybe I'm missing something and I understand the difference 😄

andrewlock: Correct, you need to pick one 😛 

Specifying both doesn't makes sense, so I could add a check that you don't specify both and just error

andrewlock: done in https://github.com/DataDog/dd-trace-dotnet/pull/1563/commits/5de37199d1a2aa77f560cc2de418e38af3b93dc1

lucaspimentel: Thanks!

lucaspimentel (approved):

kevingosse (approved):

PR #1567, opened 2021-06-28
Nuke build: overwrite files when copying trace home directory

lucaspimentel:
Overwrite files when copying the trace directory in the Nuke `CreateDdTracerHome` task

### DIFF
--- a/build/_build/Build.Steps.cs
+++ b/build/_build/Build.Steps.cs
@@ -250,7 +250,7 @@ partial class Build
             var dest = TracerHomeDirectory;
 
             Logger.Info($"Copying '{source}' to '{dest}'");
-            CopyFileToDirectory(source, dest, FileExistsPolicy.OverwriteIfNewer);
+            CopyFileToDirectory(source, dest, FileExistsPolicy.Overwrite);
         });
 
     Target PublishManagedProfiler => _ => _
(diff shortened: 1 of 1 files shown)


build/_build/Build.Steps.cs
```diff
@@ -339,7 +339,7 @@ partial class Build
        .Executes(() =>
        {
            // start by copying everything from the tracer home dir
-           CopyDirectoryRecursively(TracerHomeDirectory, DDTracerHomeDirectory, DirectoryExistsPolicy.Merge);
+           CopyDirectoryRecursively(TracerHomeDirectory, DDTracerHomeDirectory, DirectoryExistsPolicy.Merge, FileExistsPolicy.OverwriteIfNewer);
```
kevingosse, line 342: I think we should use `FileExistsPolicy.Overwrite` instead. Files _should_ always be newer, but in the unlikely case they wouldn't be, I'm afraid some files would be overwritten and other not, leading to inconsistencies that will be hard to figure out.

lucaspimentel: Sounds good. Should I fix the other uses of `FileExistsPolicy.OverwriteIfNewer` as well?

lucaspimentel: Changed them all in 7a8bee1519e525a6cd6eb33184480b1e55a4d2c3

andrewlock (approved):

kevingosse (approved):

PR #1574, opened 2021-07-01
Use the WAF as dynamic library

robertpi:
Remove the code to merge the .lib / C++ coee into the
ClrProfiler.Native.[dll|so|dynlib].

This means we now need to be able to find the plaform specific library,
so I've added some code for this. We use the environment variables to
find where the native library might be installed and load it from this
location.

Ensure that if the native library doesn't load for any reason the client
app can contiune functioning without security.

Fix the sample apps and security integrarion tests so they model a
datadog install in reasonably realistic way and that the integration
tests continue to work.

Change the way the rules file is stored. A resource file is probably
better than a very large constant.

Also: add back missing code from asp.net mvc scope creation

Fixes #

@DataDog/apm-dotnet


andrewlock (commented):
Just some minor comments, I'll defer to others for more practical reviews!

test/test-applications/Directory.Build.props
```diff
@@ -31,7 +31,7 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(ExcludeNativeProfiler)' != 'true'">
-    <None Include="$(ProfilerOutputDirectory)\*.dll;$(ProfilerOutputDirectory)\*.so;$(ProfilerOutputDirectory)\*.pdb"
+    <Content Include="$(ProfilerOutputDirectory)\*.dll;$(ProfilerOutputDirectory)\*.so;$(ProfilerOutputDirectory)\*.pdb"
```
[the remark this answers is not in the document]
robertpi: Changed it back, I think it was something of an experiment that wasn't needed in the end.

src/Datadog.Trace/AppSec/Waf/NativeBindings/Native.cs
```diff
@@ -3,22 +3,82 @@
...
 {
     internal class Native
     {
-#if !NETFRAMEWORK
-        private const string DllName = "Datadog.Trace.ClrProfiler.Native";
+#if NETFRAMEWORK
+        private const string DllName = "Sqreen.dll";
+#else
+        private const string DllName = "Sqreen";
+#endif
```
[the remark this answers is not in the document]
robertpi: The conditional logic repeated in the construct is because tuple deconstruction is not support in .NET framework 4.5. I have come up with a way to remove the conditional compilation in the construct, but I can't see anyway of avoiding it here.

src/Datadog.Trace/AppSec/Waf/NativeBindings/Native.cs
```diff
@@ -3,22 +3,82 @@
...
+            var (runtimeIdPart1, libPrefix, libExt) =
+                Environment.OSVersion.Platform switch
+                {
+                    PlatformID.MacOSX => ("osx", "lib", "dylib"),
+                    PlatformID.Unix => ("linux", "lib", "so"),
+                    PlatformID.Win32NT => ("win", string.Empty, "dll"),
+                    PlatformID.Win32S => ("win", string.Empty, "dll"),
+                    PlatformID.Win32Windows => ("win", string.Empty, "dll"),
+                    PlatformID.WinCE => ("win", string.Empty, "dll"),
+                    PlatformID.Xbox => throw new NotSupportedException(),
```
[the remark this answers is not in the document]
robertpi: It's a bit annoying that expression style switch doesn't allow regroup multiple cases together, so I've move to the old statement based switch which does allow cases to be regrouped.

It also reminded me that throwing isn't really a good idea in a static constructor, as it will mean any member accessed after static constructor failure will also throw, so I've changed that as well.

src/Datadog.Trace/AppSec/Waf/NativeBindings/Native.cs
```diff
@@ -30,7 +90,7 @@ internal class Native
         [DllImport(DllName)]
         internal static extern void pw_clearRuleH(IntPtr wafHandle);
 
-        [DllImport("Datadog.Trace.ClrProfiler.Native")]
+        [DllImport(DllName)]
```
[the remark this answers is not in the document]
robertpi: Yes, dotnet core on Windows supports post fixing the dll, that's documented here: https://docs.microsoft.com/en-us/dotnet/standard/native-interop/cross-platform

What surprises me that .NET Framework consuming a .NET standard library seems able to do this, or an least our integration tests seem to test this scenario and it seems to be working.

src/Datadog.Trace/AppSec/Waf/rule-set.txt
```diff
@@ -0,0 +1,4177 @@
+{
+  "rules": [
+    {
+      "rule_id": "rule_942140",
+      "filters": [
+        {
```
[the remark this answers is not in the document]
robertpi: Yes, no problem.

It's reasonably unlikely we'll ever need to change this file, it's provided by the app sec security team (who also created the native library we're loading)

It's also unlikely that it'll stay as part of the repository long term, the plan is that once the fleet management agent remote control protocol is implemented these rules will be provided by the datadog back end, via this new service.

anna-git (approved):

Datadog.Trace.Security.sln
```diff
@@ -37,8 +37,7 @@ EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "test-applications", "test-applications", "{9518425A-36A5-4B8F-B0B8-6137DB88441D}"
 	ProjectSection(SolutionItems) = preProject
 		test\test-applications\Directory.Build.props = test\test-applications\Directory.Build.props
-	EndProjectSection
-EndProject
+	EndProjectSectionEndProject
```
lucaspimentel, line 40: This looks like a typo. `EndProjectSectionEndProject` => `EndProjectSection` and `EndProject`

src/Datadog.Trace/AppSec/Security.cs
```diff
@@ -37,14 +37,16 @@ public Security()
 
         internal Security(InstrumentationGateway instrumentationGateway = null, IPowerWaf powerWaf = null)
         {
+            var found = bool.TryParse(Environment.GetEnvironmentVariable("DD_ENABLE_SECURITY"), out var enabled);
```
lucaspimentel, line 40: We have a (ridiculous long but optimized) `String.ToBoolean()` extension method which allows for different truthy values (`true`, `T`, `yes`, `Y`, `1`). We try to use it for all boolean configuration.

https://github.com/DataDog/dd-trace-dotnet/blob/b68cdfef714b378cd61ced1fa2894e0e215d90a9/src/Datadog.Trace/ExtensionMethods/StringExtensions.cs#L37

lucaspimentel: Is `DD_ENABLE_SECURITY` the final name for this across languages? In APM most "enabled" settings follow the `DD_{FEATURE}_ENABLED` pattern.

src/Datadog.Trace/AppSec/Waf/NativeBindings/Native.cs
```diff
@@ -39,43 +57,200 @@ internal class Native
...
         internal static extern bool pw_addMap(ref PWArgs map, string entryName, ulong entryNameLength, PWArgs entry);
 
-        [DllImport("Datadog.Trace.ClrProfiler.Native")]
+        [DllImport(DllName)]
         internal static extern void pw_freeArg(ref PWArgs input);
 
 #pragma warning restore SA1300 // Element should begin with upper-case letter
+
+        private static List<string> GetDataDogNativeFolders(string runtimeId)
```
lucaspimentel, line 99: It's `Datadog`, not `DataDog` 😉 

(Yes, it's https://github.com/DataDog on GitHub, but nobody dares change it because it could break case sensitive package managers.)

src/Datadog.Trace/AppSec/Waf/NativeBindings/Native.cs
```diff
@@ -39,43 +57,200 @@ internal class Native
...
+        }
+
+        private static IEnumerable<string> GetProfilerFolders()
+        {
+            var profilerPathsEnvVars = new List<string>()
+            {
+                "CORECLR_PROFILER_PATH",
+                "CORECLR_PROFILER_PATH_32",
+                "CORECLR_PROFILER_PATH_64",
+                "COR_PROFILER_PATH",
```
lucaspimentel, line 118: `COR_PROFILER_PATH_64` and `COR_PROFILER_PATH_32` are also supported on newer versions of .NET Framework. I could never find documentation of the exact version, but it was added around the 4.6 time frame.

src/Datadog.Trace/AppSec/Waf/NativeBindings/Native.cs
```diff
@@ -39,43 +57,200 @@ internal class Native
...
+                    break;
+                case PlatformID.Unix:
+                    runtimeIdPart1 = "linux";
+                    libPrefix = "lib";
+                    libExt = "so";
+                    break;
+                case PlatformID.Win32NT:
+                case PlatformID.Win32S:
+                case PlatformID.Win32Windows:
+                case PlatformID.WinCE:
```
lucaspimentel, line 228: We only support `Win32NT` out of these four.

src/Datadog.Trace/AppSec/Waf/NativeBindings/Native.cs
```diff
@@ -39,43 +57,200 @@ internal class Native
...
+            }
+        }
+
+        private static void GetLibNameAndRuntimeId(out string libName, out string runtimeId)
+        {
+            string runtimeIdPart1, libPrefix, libExt;
+
+            switch (Environment.OSVersion.Platform)
+            {
+                case PlatformID.MacOSX:
```
lucaspimentel, line 215: [The documentation](https://docs.microsoft.com/en-us/dotnet/api/system.platformid) says this value was only returned in Silverlight. .NET Core on macOS returns `Unix`. You want want to use you `FrameworkDescription` helper instead.
https://github.com/DataDog/dd-trace-dotnet/blob/beba073dc8f9dd86b5f9becdb19feb4bdb8b69a0/src/Datadog.Trace/FrameworkDescription.cs#L16

src/Datadog.Trace/AppSec/Security.cs
robertpi: It is not the final version, the final version is DD_APPSEC_ENABLED. I logged this jira to remind me to change it, https://datadoghq.atlassian.net/browse/APPSEC-747, buy I'll do it now, as you mention it.

robertpi: Also use the correct boolean converter now.

src/Datadog.Trace/AppSec/Waf/NativeBindings/Native.cs
robertpi: If only we'd written it in case insensitive VB.NET!

src/Datadog.Trace/AppSec/Waf/NativeBindings/Native.cs
robertpi: Changed.

src/Datadog.Trace/AppSec/Waf/NativeBindings/Native.cs
robertpi: fixed.

Datadog.Trace.Security.sln
robertpi: Thanks, correct.

src/Datadog.Trace/AppSec/Waf/NativeBindings/Native.cs
robertpi: My idea had been simple to take any available environment variable, even if it wasn't were the process loaded the profiler from, we might find the library in the same location. However, this could create problems with loading the wrong version of the library, so I changed the method just to look at the env var that would be correct for the process type.

PR #1586, opened 2021-07-09
Split Windows regression tests and integration tests to save drive space

andrewlock:
The `RunOnWindows` tests and `SmokeTests` tests are actually independent (currently) so we can save drive space by splitting them into two jobs, one which builds the regression/smoke tests, another which builds the samples. 

That saves us ~2GB in both jobs, which should stave off drive space issues for a while

> The integration tests have a direct dependency on the AspNetCoreRazor pages app for diagnostic source testing, so made that explicit in the proj file

@DataDog/apm-dotnet

### DIFF
--- a/build/_build/Build.cs
+++ b/build/_build/Build.cs
@@ -182,7 +182,7 @@ partial class Build : NukeBuild
 
     Target BuildAndRunWindowsRegressionTests => _ => _
         .Requires(() => IsWin)
-        .Description("Builds and runs the Windows (non-IIS) integration tests")
+        .Description("Builds and runs the Windows regression tests")
         .DependsOn(BuildWindowsRegressionIntegrationTests)
         .DependsOn(RunWindowsRegressionTests);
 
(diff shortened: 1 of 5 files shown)


lucaspimentel (approved):

tonyredondo (approved):

build/_build/Build.cs
```diff
@@ -172,6 +180,12 @@ void DeleteReparsePoints(string path)
         .DependsOn(BuildWindowsIntegrationTests)
         .DependsOn(RunWindowsIntegrationTests);
 
+    Target BuildAndRunWindowsRegressionTests => _ => _
+        .Requires(() => IsWin)
+        .Description("Builds and runs the Windows (non-IIS) integration tests")
```
zacharycmontoya, line 185:
```suggestion
        .Description("Builds and runs the Windows regression tests")
```

zacharycmontoya (approved):

PR #1590, opened 2021-07-09
Disable S3 upload step in the consolidated pipeline

kevingosse:
Currently, the consolidated pipeline generates invalid artifacts. Until that's fixed, only the Packages pipeline must be used to upload to S3.


kevingosse:
Changed the default value of `push_artifacts_to_s3` at the pipeline level instead.

PR #1591, opened 2021-07-09
Add new options to dd-trace tool

tonyredondo:
This PR adds 3 new options to `dd-trace` command to set: `agent-host`, `agent-port` and custom `env-vars`.

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace.Tools.Runner/Options.cs
+++ b/src/Datadog.Trace.Tools.Runner/Options.cs
@@ -38,8 +38,17 @@ namespace Datadog.Trace.Tools.Runner
         [Option("agent-url", Required = false, HelpText = "Datadog trace agent url.")]
         public string AgentUrl { get; set; }
 
+        [Option("agent-host", Required = false, HelpText = "Datadog trace agent hostname.")]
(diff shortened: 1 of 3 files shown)


lucaspimentel:
We should consider using `DD_AGENT_URL` instead, since that setting is meant to replace `DD_AGENT_HOST`/`DD_TRACE_AGENT_PORT`. Once we add these parameters to the cli tool, they are public API and we'll need to support them "forever." See team discussion on Slack.

tonyredondo:
> We should consider using `DD_AGENT_URL` instead, since that setting is meant to replace `DD_AGENT_HOST`/`DD_TRACE_AGENT_PORT`. Once we add these parameters to the cli tool, they are public API and we'll need to support them "forever." See team discussion on Slack.

I’m confused… I already have support for something similar but the key contains Tracer: https://github.com/DataDog/dd-trace-dotnet/blob/master/src/Datadog.Trace.Tools.Runner/Utils.cs#L119

tonyredondo:
Both `DD_AGENT_HOST` and `DD_TRACE_AGENT_PORT` will be deprecated, closing this.

PR #1594, opened 2021-07-12
CIApp - Add support to add custom env-vars from dd-trace wrapper

tonyredondo:
@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace.Tools.Runner/Properties/launchSettings.json
+++ b/src/Datadog.Trace.Tools.Runner/Properties/launchSettings.json
@@ -2,7 +2,7 @@
   "profiles": {
     "Datadog.Trace.Tools.Runner": {
       "commandName": "Project",
-      "commandLineArgs": "--crank-import=\"C:\\temp\\crank\\results.json\""
+      "commandLineArgs": "--dd-env=ci --env-vars=myVar1=myValue1,myVar2=myValue2 -- dotnet --info"
     }
   }
 }
--- a/src/Datadog.Trace.Tools.Runner/Properties/launchSettings.json
+++ b/src/Datadog.Trace.Tools.Runner/Properties/launchSettings.json
@@ -2,7 +2,7 @@
   "profiles": {
     "Datadog.Trace.Tools.Runner": {
       "commandName": "Project",
-      "commandLineArgs": "--dd-env=ci --env-vars=myVar1=myValue1,myVar2=myValue2 -- dotnet --info"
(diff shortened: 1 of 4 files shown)


src/Datadog.Trace.Tools.Runner/Properties/launchSettings.json
```diff
@@ -2,7 +2,7 @@
   "profiles": {
     "Datadog.Trace.Tools.Runner": {
       "commandName": "Project",
-      "commandLineArgs": "--crank-import=\"C:\\temp\\crank\\results.json\""
+      "commandLineArgs": "--dd-env=ci --env-vars=myVar1=myValue1,myVar2=myValue2 -- dotnet --info"
```
lucaspimentel, line 5:
```
--env-vars=myVar1=myValue1,myVar2=myValue2
```
This looks a little weird with the `=` serving two different functions. Doesn't `CommandLineParser` support using the same parameter multiple times and mapping that to an collection (instead of us splitting manually using commas)?
```
--env-vars myVar1=myValue1 --env-vars myVar2=myValue2
```

tonyredondo: I tried, but is not supported, also have some conflict with another IEnumerable<string> to collect the command to wrap + args.

andrewlock: Maybe this works? `--dd-env ci --env-vars myVar1=myValue1,myVar2=myValue2 -- dotnet --info`?

src/Datadog.Trace.Tools.Runner/Utils.cs
```diff
@@ -119,6 +119,21 @@ internal class Utils
                 envVars["DD_TRACE_AGENT_URL"] = options.AgentUrl;
             }
 
+            if (!string.IsNullOrWhiteSpace(options.EnvironmentValues))
+            {
+                foreach (var keyValue in options.EnvironmentValues.Split(','))
```
andrewlock, line 124: Do you want to auto-trim white space around entries here?

```suggestion
                foreach (var keyValue in options.EnvironmentValues.Split(',', StringSplitOptions.TrimEntries))
```

andrewlock (approved):

zacharycmontoya (approved):

src/Datadog.Trace.Tools.Runner/Properties/launchSettings.json
tonyredondo: Yeah that works too

lucaspimentel (approved):

PR #1597, opened 2021-07-12
Fixes couple of native memory leaks

tonyredondo:
This PR is a continuation of https://github.com/DataDog/dd-trace-dotnet/pull/1596 handling a couple of possible memory leaks in the native library.

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
+++ b/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
@@ -241,7 +241,7 @@ HRESULT STDMETHODCALLTYPE CorProfiler::Initialize(IUnknown* cor_profiler_info_un
 
 
   // set event mask to subscribe to events and disable NGEN images
-  if (is_net46_or_greater) 
+  if (is_net46_or_greater)
   {
         hr = info6->SetEventMask2(event_mask, COR_PRF_HIGH_ADD_ASSEMBLY_REFERENCES);
 
@@ -634,6 +634,10 @@ HRESULT STDMETHODCALLTYPE CorProfiler::ModuleUnloadStarted(ModuleID module_id)
         }
 
         module_id_to_info_map_.erase(module_id);
+        if (rejit_handler != nullptr)
+        {
+            rejit_handler->ReleaseModule(module_id);
+        }
         delete metadata;
     }
 
@@ -651,6 +655,8 @@ HRESULT STDMETHODCALLTYPE CorProfiler::Shutdown()
     if (rejit_handler != nullptr)
     {
         rejit_handler->Shutdown();
+        delete[] rejit_handler;
+        rejit_handler = nullptr;
     }
     Warn("Exiting. Stats: ", Stats::Instance()->ToString());
     is_attached_.store(false);
@@ -1769,7 +1775,7 @@ std::string CorProfiler::GetILCodes(const std::string& title, ILRewriter* rewrit
             }
             else if (cInstr->m_opcode == CEE_LDSTR)
             {
-                LPWSTR szString = new WCHAR[1024];
+                WCHAR szString[1024];
                 ULONG szStringLength;
                 auto hr = module_metadata->metadata_import->GetUserString((mdString) cInstr->m_Arg32, szString, 1024,
                                                                           &szStringLength);
@@ -3478,6 +3484,8 @@ HRESULT CorProfiler::CallTarget_RewriterCallback(RejitHandlerModule* moduleHandl
     }
 
     hr = rewriter.Export();
+
+    delete[] newEHClauses;
 
     if (FAILED(hr))
     {
--- a/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
+++ b/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
@@ -655,7 +655,7 @@ HRESULT STDMETHODCALLTYPE CorProfiler::Shutdown()
     if (rejit_handler != nullptr)
     {
         rejit_handler->Shutdown();
-        delete[] rejit_handler;
+        delete rejit_handler;
         rejit_handler = nullptr;
     }
     Warn("Exiting. Stats: ", Stats::Instance()->ToString());
@@ -3484,8 +3484,6 @@ HRESULT CorProfiler::CallTarget_RewriterCallback(RejitHandlerModule* moduleHandl
     }
 
     hr = rewriter.Export();
-
-    delete[] newEHClauses;
 
     if (FAILED(hr))
     {
--- a/src/Datadog.Trace.ClrProfiler.Native/module_metadata.h
+++ b/src/Datadog.Trace.ClrProfiler.Native/module_metadata.h
@@ -52,7 +52,7 @@ public:
     {
         if (calltargetTokens != nullptr)
         {
-            delete[] calltargetTokens;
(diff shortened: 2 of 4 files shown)


src/Datadog.Trace.ClrProfiler.Native/rejit_handler.cpp
```diff
@@ -135,6 +134,18 @@ bool RejitHandler::TryGetModule(ModuleID moduleId, RejitHandlerModule** moduleHa
...
 }
 
+void RejitHandler::ReleaseModule(ModuleID moduleId)
+{
+    std::lock_guard<std::mutex> guard(modules_lock);
+    auto find_res = modules.find(moduleId);
+    if (find_res != modules.end())
+    {
+        auto moduleHandler = find_res->second;
+        delete[] moduleHandler;
```
gleocadie, line 144: shoudn't it be `delete moduleHandler` instead ?

src/Datadog.Trace.ClrProfiler.Native/rejit_handler.h
```diff
@@ -37,46 +40,58 @@ class RejitHandlerModuleMethod
...
         this->methodReplacement = nullptr;
     }
-    inline mdMethodDef GetMethodDef()
+    ~RejitHandlerModuleMethod()
+    {
+        this->pFunctionControl = nullptr;
+        this->module = nullptr;
+        if (this->functionInfo != nullptr)
+        {
+            delete[] this->functionInfo;
```
gleocadie, line 60: shouldn't it be `delete this->functionInfo` instead ?

src/Datadog.Trace.ClrProfiler.Native/rejit_handler.h
```diff
@@ -94,28 +109,38 @@ class RejitHandlerModule
...
         this->handler = handler;
     }
-    inline ModuleID GetModuleId()
+    ~RejitHandlerModule()
+    {
+        this->metadata = nullptr;
+        this->handler = nullptr;
+        for (auto moduleMethod : methods)
+        {
+            delete[] moduleMethod.second;
```
gleocadie, line 127: should'nt it be `delete moduleMethod.second` instead?

src/Datadog.Trace.ClrProfiler.Native/rejit_handler.h
```diff
@@ -151,9 +176,23 @@ class RejitHandler
         this->rejit_queue_ = new BlockingQueue<RejitItem>();
         this->rejit_queue_thread_ = new std::thread(enqueue_thread, this);
     }
+    ~RejitHandler()
+    {
+        if (this->rejit_queue_ != nullptr)
+        {
+            delete[] this->rejit_queue_;
```
gleocadie, line 183: shoudn't it be `delete this->rejit_queue_;` instead?

src/Datadog.Trace.ClrProfiler.Native/rejit_handler.h
```diff
@@ -151,9 +176,23 @@ class RejitHandler
...
+            this->rejit_queue_ = nullptr;
+        }
+        if (this->rejit_queue_thread_ != nullptr)
+        {
+            delete[] this->rejit_queue_thread_;
```
gleocadie, line 188: shouldn't it be`delete this->rejit_queue_thread_;` instead ?

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -651,6 +655,8 @@ HRESULT STDMETHODCALLTYPE CorProfiler::Shutdown()
     if (rejit_handler != nullptr)
     {
         rejit_handler->Shutdown();
+        delete[] rejit_handler;
```
gleocadie, line 658: should'nt it be `delete rejit_handler;` instead ?

src/Datadog.Trace.ClrProfiler.Native/module_metadata.h
```diff
@@ -48,6 +48,14 @@ class ModuleMetadata
         corAssemblyProperty(corAssemblyProperty)
     {
     }
+    ~ModuleMetadata()
+    {
+        if (calltargetTokens != nullptr)
+        {
+            delete[] calltargetTokens;
```
gleocadie, line 55: shouldn't it be `delete calltargetTokens` instead ?

gleocadie (changes requested):
most of the deletion should be `delete` and not `delete[]`.
Rule being:
- Allocate using `new`, delete with `delete`
- Allocate using `new[]` delete with `delete[]`

The mixup is considered as undefined behavior

Maybe someday, as a big refactoring moving to `std::unique_ptr`/`std::shared_ptr` (depending on the semantic) could be a better  option.

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -3479,6 +3485,8 @@ HRESULT CorProfiler::CallTarget_RewriterCallback(RejitHandlerModule* moduleHandl
 
     hr = rewriter.Export();
 
+    delete[] newEHClauses;
```
gleocadie, line 3488: That could be risky no? you pass its pointer to the rewriter and ends it lifecyle before the rewriter lifecyle.
Couldn't you use `std::unique_ptr` instead ? (and passing the pointer to the rewriter using the `.get` method )?

src/Datadog.Trace.ClrProfiler.Native/module_metadata.h
gleocadie: I think in this case you can go for a `std::unique_ptr`.

tonyredondo:
> most of the deletion should be `delete` and not `delete[]`.
> Rule being:
> 
> * Allocate using `new`, delete with `delete`
> * Allocate using `new[]` delete with `delete[]`
> 
> The mixup is considered as undefined behavior
> 
> Maybe someday, as a big refactoring moving to `std::unique_ptr`/`std::shared_ptr` (depending on the semantic) could be a better option.

Yes is a WIP, I had to go out before completing this. That’s why the WIP label and the Draft 😊

src/Datadog.Trace.ClrProfiler.Native/rejit_handler.h
gleocadie: It's possible to use `std::unique_ptr` here instead.

src/Datadog.Trace.ClrProfiler.Native/rejit_handler.h
gleocadie: for `methods` it's also possible to use `std::unique_ptr`

src/Datadog.Trace.ClrProfiler.Native/rejit_handler.h
```diff
@@ -151,9 +176,23 @@ class RejitHandler
         this->rejit_queue_ = new BlockingQueue<RejitItem>();
         this->rejit_queue_thread_ = new std::thread(enqueue_thread, this);
     }
+    ~RejitHandler()
```
gleocadie, line 179: for both, it's possible to use std::unique_ptr instead

gleocadie:
I tested a bit and overall, you can replace every raw pointers by a std::unique_ptr ;)

tonyredondo:
Superseeded by #1600

PR #1599, opened 2021-07-13
Add a public ForceFlushAsync API on the Tracer

kevingosse:
To be used when the process or appdomain is about to get killed in a non-graceful manner.

Note: 
 - I did not make the old `Tracer.FlushAsync` public because the name does not properly convey the information that this API shouldn't be used under normal circumstances
 - I did not delete the old `Tracer.FlushAsync` because that would be a breaking change (used in `Datadog.Trace.ClrProfiler.Managed`)

### DIFF
--- a/src/Datadog.Trace/Tracer.cs
+++ b/src/Datadog.Trace/Tracer.cs
@@ -377,7 +377,7 @@ namespace Datadog.Trace
         /// To be called when the appdomain or the process is about to be killed in a non-graceful way.
         /// </summary>
         /// <returns>Task used to track the async flush operation</returns>
-        public Task ForceFlush() => FlushAsync();
+        public Task ForceFlushAsync() => FlushAsync();
 
         /// <summary>
         /// Writes the specified <see cref="Span"/> collection to the agent writer.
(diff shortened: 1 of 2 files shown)


src/Datadog.Trace/Tracer.cs
```diff
@@ -372,6 +372,13 @@ public Span StartSpan(string operationName, ISpanContext parent = null, string s
             return StartSpan(operationName, tags: null, parent, serviceName, startTime, ignoreActiveScope, spanId: null);
         }
 
+        /// <summary>
+        /// Forces the tracer to immediately flush pending traces and send them to the agent.
+        /// To be called when the appdomain or the process is about to be killed in a non-graceful way.
+        /// </summary>
+        /// <returns>Task used to track the async flush operation</returns>
+        public Task ForceFlush() => FlushAsync();
```
andrewlock, line 380: Should this be `ForceFlushAsync`? Given we have the `async` suffix elsewhere.

andrewlock (approved):

src/Datadog.Trace/Tracer.cs
kevingosse: Right

tonyredondo (approved):

PR #1600, opened 2021-07-13
Fixes native memory leaks

tonyredondo:
This PR superseeds #1597

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
+++ b/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
@@ -241,7 +241,7 @@ HRESULT STDMETHODCALLTYPE CorProfiler::Initialize(IUnknown* cor_profiler_info_un
 
 
   // set event mask to subscribe to events and disable NGEN images
-  if (is_net46_or_greater) 
+  if (is_net46_or_greater)
   {
         hr = info6->SetEventMask2(event_mask, COR_PRF_HIGH_ADD_ASSEMBLY_REFERENCES);
 
@@ -634,6 +634,11 @@ HRESULT STDMETHODCALLTYPE CorProfiler::ModuleUnloadStarted(ModuleID module_id)
         }
 
         module_id_to_info_map_.erase(module_id);
+
+        if (rejit_handler != nullptr)
+        {
+            rejit_handler->RemoveModule(module_id);
+        }
         delete metadata;
     }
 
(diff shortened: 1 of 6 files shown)


src/Datadog.Trace.ClrProfiler.Native/il_rewriter.cpp
```diff
@@ -176,6 +176,13 @@ EHClause* ILRewriter::GetEHPointer()
 
 void ILRewriter::SetEHClause(EHClause* ehPointer, unsigned ehLength)
 {
+    if (m_pEH != nullptr)
+    {
+        // Delete previous array
+        m_nEH = 0;
```
gleocadie, line 182: you could use a `std::unique_ptr` for `m_nEh` and use the `reset` method to i) replace the old array by the new array ii) release the memory of the old array.

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -651,6 +656,8 @@ HRESULT STDMETHODCALLTYPE CorProfiler::Shutdown()
     if (rejit_handler != nullptr)
     {
         rejit_handler->Shutdown();
+        delete rejit_handler;
```
gleocadie, line 659: out of curiosity, why not a `std::unique_ptr` ?

gleocadie (commented):
Overall nice PR ;)
2 questions about.... `std::unique_ptr`

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
tonyredondo: because I'm interested in releasing all resources before the CorProfiler instance dtor... I know I can do it by calling a method on `std::unique_ptr` too but I leave it like this and leave the `std::unique_ptr` to the scenario where it follows the complete lifecycle of the parent.

src/Datadog.Trace.ClrProfiler.Native/il_rewriter.cpp
tonyredondo: I thought about it, but decide to don't change this `il_rewiter.cpp` file a lot, because the `m_nEH` member is not the only pointer. Maybe I can leave that for a bigger refactor of that class.

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
gleocadie: sure no pb, just curiosity ;)

gleocadie (approved):
LGTM

zacharycmontoya (approved):
With the caveat that I'm really new to smart pointers in C++, it looks good to me

PR #1601, opened 2021-07-13
Remove usage of non-builtin types in Attribute named arguments

zacharycmontoya:
To workaround an issue discovered in Azure Functions, remove all usages of named arguments in Attributes that are not built-in types. This means removing arguments that use our own custom enums, but keeping arguments that use `string`, etc. While rare, the failing scenario occurs when the corelib attempts to create the `Attribute` object from the metadata but the assembly that gets resolved from the default load context (or one of its handlers) does not resolve to the same assembly as the one we expected, causing type equality to fail.

Note: This workaround is only required for attributes that will be instantiated from metadata, like `DuckAttribute`, but it is easier to maintain a broader rule that we can always revisit later.

Specific changes include:
- Replacing all usages of `[Duck(Kind = DuckKind.Field)]` with a new attribute `[DuckField]`
- Replacing the one usage of `[InterceptMethod(MethodReplacementAction = MethodReplacementActionType.InsertFirst)]` with `[InsertFirstInterceptMethod]`
- Adding a unit test to continuously maintain this coding guideline

@DataDog/apm-dotnet

### DIFF
--- a/test/Datadog.Trace.ClrProfiler.Managed.Tests/AttributeTests.cs
+++ b/test/Datadog.Trace.ClrProfiler.Managed.Tests/AttributeTests.cs
@@ -1,0 +1,53 @@
+// <copyright file="AttributeTests.cs" company="Datadog">
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2 License.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using FluentAssertions;
+using Xunit;
+
+namespace Datadog.Trace.ClrProfiler.Managed.Tests
+{
+    public class AttributeTests
+    {
+        /// <summary>
+        /// This test reports instances in the Datadog.Trace.ClrProfiler.Managed and Datadog.Trace
+        /// where attributes use named arguments that are not a built-in type. Using a custom type
+        /// (especially one from the Datadog.Trace assembly) as a named argument can fail with a
+        /// System.Reflection.CustomAttributeFormatException if the owning assembly cannot be found
+        /// in the default load context and other assembly resolve events are registered, in addition
+        /// to the ones added by the Datadog .NET Tracer
+        /// </summary>
+        [Fact]
+        public void AttributesInstantiationsOnlyUseBuiltinTypes()
+        {
+            List<string> invalidAttributeUsages = new();
+            var coreAssembly = typeof(object).Assembly;
+            var managedAssembly = typeof(Instrumentation).Assembly;
+            var types = coreAssembly.GetTypes().Concat(managedAssembly.GetTypes());
+
+            foreach (var type in types)
+            {
+                foreach (var member in type.GetMembers())
+                {
+                    foreach (var attribute in member.CustomAttributes)
+                    {
+                        foreach (var namedArg in attribute.NamedArguments)
+                        {
+                            var argumentType = namedArg.TypedValue.ArgumentType;
+                            if (argumentType.Assembly != coreAssembly)
+                            {
+                                invalidAttributeUsages.Add($"{type.FullName}.{member.Name} => {attribute.AttributeType.FullName}({namedArg.MemberName} : {argumentType.FullName})");
+                            }
+                        }
+                    }
+                }
+            }
+
+            invalidAttributeUsages.Should().BeEmpty();
+        }
+    }
+}
--- a/test/Datadog.Trace.ClrProfiler.Managed.Tests/AttributeTests.cs
+++ b/test/Datadog.Trace.ClrProfiler.Managed.Tests/AttributeTests.cs
@@ -26,8 +26,10 @@ namespace Datadog.Trace.ClrProfiler.Managed.Tests
         {
             List<string> invalidAttributeUsages = new();
             var coreAssembly = typeof(object).Assembly;
+
+            var tracerAssembly = typeof(Tracer).Assembly;
             var managedAssembly = typeof(Instrumentation).Assembly;
-            var types = coreAssembly.GetTypes().Concat(managedAssembly.GetTypes());
+            var types = tracerAssembly.GetTypes().Concat(managedAssembly.GetTypes());
 
             foreach (var type in types)
             {
--- a/test/Datadog.Trace.ClrProfiler.Managed.Tests/Datadog.Trace.ClrProfiler.Managed.Tests.csproj
+++ b/test/Datadog.Trace.ClrProfiler.Managed.Tests/Datadog.Trace.ClrProfiler.Managed.Tests.csproj
@@ -31,4 +31,17 @@
   <ItemGroup Condition="'$(TargetFramework)' != 'net452'">
     <PackageReference Include="Microsoft.Data.SqlClient" Version="2.0.1" />
   </ItemGroup>
+
+  <!-- Bring in Microsoft.AspNetCore.Mvc.Testing packages so we can resolve all type references -->
+  <ItemGroup Condition=" $(TargetFramework.StartsWith('netcoreapp2')) ">
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="2.2.0" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" $(TargetFramework.StartsWith('netcoreapp3')) ">
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="3.0.0" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" $(TargetFramework.StartsWith('net5')) ">
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="5.0.0" />
+  </ItemGroup>
 </Project>
--- a/test/Datadog.Trace.ClrProfiler.Managed.Tests/Datadog.Trace.ClrProfiler.Managed.Tests.csproj
+++ b/test/Datadog.Trace.ClrProfiler.Managed.Tests/Datadog.Trace.ClrProfiler.Managed.Tests.csproj
@@ -32,16 +32,9 @@
     <PackageReference Include="Microsoft.Data.SqlClient" Version="2.0.1" />
   </ItemGroup>
 
-  <!-- Bring in Microsoft.AspNetCore.Mvc.Testing packages so we can resolve all type references -->
-  <ItemGroup Condition=" $(TargetFramework.StartsWith('netcoreapp2')) ">
-    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="2.2.0" />
-  </ItemGroup>
-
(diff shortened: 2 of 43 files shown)


test/Datadog.Trace.ClrProfiler.Managed.Tests/AttributeTests.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// (especially one from the Datadog.Trace assembly) as a named argument can fail with a
+        /// System.Reflection.CustomAttributeFormatException if the owning assembly cannot be found
+        /// in the default load context and other assembly resolve events are registered, in addition
+        /// to the ones added by the Datadog .NET Tracer
+        /// </summary>
+        [Fact]
+        public void AttributesInstantiationsOnlyUseBuiltinTypes()
+        {
+            List<string> invalidAttributeUsages = new();
+            var coreAssembly = typeof(object).Assembly;
```
lucaspimentel, line 28: Did you meant to use something like `typeof(Tracer)` here? We want attributes from `Datadog.Trace.dll` and `Datadog.Trace.ClrProfiler.Managed.dll`, right?

lucaspimentel: Oh, I guess not. I just noticed the `if (argumentType.Assembly != coreAssembly)` below. I still don't understand why we iterate over all types in `typeof(object).Assembly`, though.

zacharycmontoya: 🤦🏼 Yep exactly, I meant to do `typeof(Tracer)`. I still need the core assembly to do the assembly comparison later, so I'll just add a small fix to get the Trace assembly. Fixed in 78269a33c2f22c5e875a6a6a227797795e259bc2

test/Datadog.Trace.ClrProfiler.Managed.Tests/AttributeTests.cs
```diff
@@ -0,0 +1,55 @@
...
+            var managedAssembly = typeof(Instrumentation).Assembly;
+            var types = tracerAssembly.GetTypes().Concat(managedAssembly.GetTypes());
+
+            foreach (var type in types)
+            {
+                foreach (var member in type.GetMembers())
+                {
+                    foreach (var attribute in member.CustomAttributes)
+                    {
+                        foreach (var namedArg in attribute.NamedArguments)
```
andrewlock, line 40: Just confirming, the issue is explicitly only around _named_ arguments right?  Ordered arguments don't cause any issues, even if they're custom types?

src/Datadog.Trace.ClrProfiler.Managed.Core/InsertFirstInterceptMethodAttribute.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = true, Inherited = false)]
+    public class InsertFirstInterceptMethodAttribute : InterceptMethodAttribute
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="InsertFirstInterceptMethodAttribute"/> class.
+        /// </summary>
+        public InsertFirstInterceptMethodAttribute()
+        {
+            MethodReplacementAction = MethodReplacementActionType.InsertFirst;
```
andrewlock, line 23: I know we have a unit test for it, but should we make the `MethodReplacementAction` setter `private`/`protected`, so you can't use it directly anyway? (Same goes for `Kind` in the other attribute)

andrewlock (approved):

test/Datadog.Trace.ClrProfiler.Managed.Tests/AttributeTests.cs
tonyredondo: I thought this issue happens on all kind of arguments with a custom Type regardless of named or not.

zacharycmontoya: Good point...it definitely fails on named arguments like the following declaration:

```csharp
[Duck(Name = "_binding", Kind = DuckKind.Field)]
```

To test constructor arguments I temporarily modified `DuckAttribute` to have a constructor like the following:
```csharp
        public DuckAttribute(string nameCtorArg, DuckKind kindCtorArg)
        {
            Name = nameCtorArg;
            Kind = kindCtorArg;
        }
```

To test positional arguments on the constructor, I used the following invocation but it succeeded.
```csharp
[Duck("_binding", DuckKind.Field)]
```

To test named arguments on the constructor, I used the following invocation and used a named argument of type `System.Reflection.BindingFlags` and it succeeds.
```csharp
[Duck("_binding", DuckKind.Field, BindingFlags = DuckAttribute.DefaultFlags | BindingFlags.IgnoreCase)]
```

And for one final test, I added a `DuckKind` named argument just to make sure, and sure enough it fails again with this new custom type.
```csharp
[Duck("_binding", DuckKind.Field, BindingFlags = DuckAttribute.DefaultFlags | BindingFlags.IgnoreCase, Kind = DuckKind.Field)]
```

Are there any other potential usages I'm missing?

src/Datadog.Trace.ClrProfiler.Managed.Core/InsertFirstInterceptMethodAttribute.cs
zacharycmontoya: Unfortunately, we can't change `DuckAttribute.Kind` because it exists in `Datadog.Trace.dll` which could be referenced by an older `Datadog.Trace.ClrProfiler.Managed.dll` if the NuGet package is used, and several integrations set the `Kind` named argument 😞

I can and will change `MethodReplacementAction` since that type is defined in `Datadog.Trace.ClrProfiler.Managed.Core` and will always be in sync with the profiler because it's provided by our automatic instrumentation packages.

Great suggestion @andrewlock  👍🏼

zacharycmontoya:
> Due this is already a breaking change (instrumentations using `[DuckField]` cannot be used in olders `Datadog.Trace` versions). Did you consider to create`[DuckProperty]` as well for symmetry? Doing this we can also change `DuckAttribute` and declared as abstract (this requires to add all possible options to children attributes).

I like the suggestion of having a `[DuckProperty]` as well to make it symmetrical 😄  However I don't think I can declare `DuckAttribute` abstract, because the `Datadog.Trace.dll` can be loaded by older versions of `Datadog.Trace.ClrProfiler.Managed.dll` which expect to instantiate a `DuckAttribute`.

test/Datadog.Trace.ClrProfiler.Managed.Tests/AttributeTests.cs
lucaspimentel: What was the conclusion of your tests? I would expect both named and ordered arguments to behave the same way, since that is handled by the C# compiler and it emits the same IL in both cases (no difference at runtime).

lucaspimentel (approved):

test/Datadog.Trace.ClrProfiler.Managed.Tests/AttributeTests.cs
zacharycmontoya: My conclusion was that named arguments (i.e. object properties) are treated differently than constructor arguments. The error only occurs when a **named argument** has a custom type. Constructor arguments (both named and positional) do not cause errors when the argument type is a custom type.

lucaspimentel: Interesting, thanks!

lucaspimentel: Oops, I was thinking of named method arguments like `Foo(name: "bar")`, but these are attributes. Never mind!

colin-higgins (approved):

test/Datadog.Trace.ClrProfiler.Managed.Tests/Datadog.Trace.ClrProfiler.Managed.Tests.csproj
```diff
@@ -28,4 +28,10 @@
   <ItemGroup Condition="'$(TargetFramework)' != 'net452'">
     <PackageReference Include="Microsoft.Data.SqlClient" Version="2.0.1" />
   </ItemGroup>
+
+  <ItemGroup Condition=" !$(TargetFramework.StartsWith('net4')) ">
+    <PackageReference Include="Microsoft.AspNetCore" Version="2.2.0" />
```
lucaspimentel, line 33: I added these as well in my PR, but I don't remember why. Do you? See https://github.com/DataDog/dd-trace-dotnet/pull/1443#discussion_r678299277
cc @andrewlock

lucaspimentel: Btw, there's already a group with condition `!$(TargetFramework.StartsWith('net4'))` in line 23 above, so these references could be moved there. (level of care low)

zacharycmontoya: This is unnecessary after you added it. I had to add it because our AspNetCoreDiagnosticObserver depended on some Microsoft.AspNetCore.Http types (I forgot which one, maybe `HttpContext`?), but we only compile against ref assemblies. So when we load the types at runtime in our test framework, it can't resolve the runtime type. So by adding the implementation assemblies with these package references, the type loads succeed

test/Datadog.Trace.ClrProfiler.Managed.Tests/Datadog.Trace.ClrProfiler.Managed.Tests.csproj
```diff
@@ -29,16 +29,9 @@
...
-  <ItemGroup Condition=" $(TargetFramework.StartsWith('netcoreapp3')) ">
-    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="3.0.0" />
-  </ItemGroup>
-
-  <ItemGroup Condition=" $(TargetFramework.StartsWith('net5')) ">
-    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="5.0.0" />
+  <ItemGroup Condition=" !$(TargetFramework.StartsWith('net4')) ">
+    <PackageReference Include="Microsoft.AspNetCore" Version="2.2.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc" Version="2.2.0" />
+    <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="2.2.0" />
```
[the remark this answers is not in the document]
zacharycmontoya: See https://github.com/DataDog/dd-trace-dotnet/pull/1601#discussion_r688969747 for why I had to add the package references to begin with. Why we're referencing 2.2.0 specifically...I have no idea. I suppose we can try to change this, but I'd rather do that in a separate PR

zacharycmontoya:
I'll merge this after we put together the next (pre-)release

PR #1604, opened 2021-07-15
Update package versions + fix MSTest

andrewlock:
Ran `GeneratePackageVersions` and found that it broke an MSTest integration test in the latest patch release.

Previously, we were receiving only one `SimpleSkipParameterizedTest` because the skip was executed before iterating the parameters. That changed in 2.2.5, so each parameterised run gets a span.

@DataDog/apm-dotnet


tonyredondo (approved):
Thanks for the fix!

zacharycmontoya (approved):

PR #1605, opened 2021-07-15
Merge tools into Nuke

andrewlock:
This PR merges all* the separate tools we have into Nuke.

Some of these were a simple case of moving the code into the Nuke build project and creating a new target:
* `GeneratePackageVersion` (generate the multi-package test versions code)
* `UpdateVendors` (update the vendored code and dependabot honeypot)
* `PrepareRelease` - msi and versions targets
* `CIAppFeatureTracking` - Had to switch to using reflection for this one

Moving the `PrepareRelease` integrations target was a PITA, as needed to switch to using reflection to find the types. I also now search _all_ the target versions of the managed libraries, so that we don't miss integrations that are target-framework specific. 

I also removed the `Datadog.Core.Tools` project, as it only had 2 files in it, one of which (`EnvironmentTools`) was only used in test helpers, and the other (`TcpPortProvider`) was only used in tests and samples. 

*~The only "tool" project I _didn't_ move is the DistributedTrace project, as that's hard to do without either referencing Datadog.Trace or using a lot of reflection. I'm thinking either:~ As it's a very specific, simple, and (it sounds like, rarely used) tool, I removed it in https://github.com/DataDog/dd-trace-dotnet/pull/1605/commits/8052a9eb0ef5f91ea35675863217faef99b9e4d6
* ~Create a "customer sample" that uses similar code~
* ~Create a "test sample" that uses similar code~

And then invoke that sample from Nuke? Preferences?

Once merged I'll also
- [ ] Create GitHub workflows to automate the release process using Nuke
- [ ] Update the Wiki with the new commands etc

@DataDog/apm-dotnet


zacharycmontoya:
> *The only "tool" project I didn't move is the DistributedTrace project, as that's hard to do without either referencing Datadog.Trace or using a lot of reflection.

I don't think anyone else has expressed interest in using this tool so maybe we can just remove it. I created this tool as an easy way to test distributed tracing functionality, but spinning up a new program to do this takes very little time.

build/_build/_build.csproj
```diff
@@ -24,4 +25,9 @@
     <Content Include="..\..\.azure-pipelines\ultimate-pipeline.yml" Link="ci\ultimate-pipeline.yml" />
     <Content Include="..\..\build_in_docker.sh" Link="boot\build_in_docker.sh" />
   </ItemGroup>
+
+  <ItemGroup>
+    <None Remove="UpdateVendors\Datadog.Dependabot.Honeypot.template" />
+    <EmbeddedResource Include="UpdateVendors\Datadog.Dependabot.Honeypot.template" />
```
zacharycmontoya, line 31: Neat! 👍🏼

build/_build/Build.Utilities.cs
```diff
@@ -168,4 +173,91 @@ partial class Build
...
+       .Description("Update the version number for the tracer")
+       .Before(Clean, BuildTracerHome)
+       .Requires(() => Version)
+       .Executes(() =>
+        {
+            new SetAllVersions(RootDirectory, Version, IsPrerelease).Run();
+        });
+
+    Target SyncMsiContents => _ => _
+       .Description("Update the version number for the tracer")
```
zacharycmontoya, line 224:
```suggestion
       .Description("Update the contents of the MSI")
```

zacharycmontoya (commented):
Before any changes to the `StartDistributedTrace` project, this looks good to me

zacharycmontoya (approved):

build/_build/Build.Utilities.cs
```diff
@@ -168,4 +173,91 @@ partial class Build
...
+       .Description("Regenerate the PackageVersions props and .cs files")
+       .Executes(async () =>
+       {
+           var testDir = Solution.GetProject(Projects.ClrProfilerIntegrationTests).Directory;
+
+           var versionGenerator = new PackageVersionGenerator(RootDirectory, testDir);
+           await versionGenerator.GenerateVersions();
+       });
+
+    Target UpdateVendors => _ => _
```
lucaspimentel, line 187: Can we rename this to `UpdateVendoredCode`? It updates "code", not "vendors."

Even better, maybe something like `UpdateThirdPartyCode` or `UpdateImportedCode`? I'm not a fan of using "vendor" as a verb.

build/_build/Build.Utilities.cs
```diff
@@ -168,4 +173,91 @@ partial class Build
...
+    Target SetVersion => _ => _
+       .Description("Update the version number for the tracer")
+       .Before(Clean, BuildTracerHome)
+       .Requires(() => Version)
+       .Executes(() =>
+        {
+            new SetAllVersions(RootDirectory, Version, IsPrerelease).Run();
+        });
+
+    Target SyncMsiContents => _ => _
```
lucaspimentel, line 223: Thanks for removing the word `Synchronize` from `SetVersion`. Can we do the same here? `UpdateMsiContents`?

build/_build/Build.cs
```diff
@@ -53,6 +53,9 @@ partial class Build : NukeBuild
     [Parameter("The build version (for packaging purposes). Default is latest")]
```
lucaspimentel, line 53: Version is no longer used only for "for packaging purposes" anymore, right?
```suggestion
    [Parameter("The build version. Default is latest")]
```

build/_build/PrepareRelease/GenerateIntegrationDefinitions.cs
```diff
@@ -0,0 +1,494 @@
...
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+
+namespace PrepareRelease
+{
+    public static class GenerateIntegrationDefinitions
+    {
+        const string InstrumentMethodAttributeName = "Datadog.Trace.ClrProfiler.InstrumentMethodAttribute";
+        const string InterceptMethodAttributeName = "Datadog.Trace.ClrProfiler.InterceptMethodAttribute";
+        public static void Run(ICollection<string> assemblyPaths, params string[] outputDirectories)
```
lucaspimentel, line 23: whitespace nitpicking 
```suggestion
        const string InterceptMethodAttributeName = "Datadog.Trace.ClrProfiler.InterceptMethodAttribute";

        public static void Run(ICollection<string> assemblyPaths, params string[] outputDirectories)
```

build/_build/PrepareRelease/SetAllVersions.cs
```diff
@@ -7,14 +7,17 @@
...
 using System.Text;
 using System.Text.RegularExpressions;
+using Nuke.Common.IO;
 
 namespace PrepareRelease
 {
     public class SetAllVersions
     {
-        public SetAllVersions(string solutionDirectory)
+        public SetAllVersions(string solutionDirectory, string tracerVersion, bool isPrerelease)
```
lucaspimentel, line 16: Nice. I will be able to clean up some code in my branch. I had to do something similar to pass versions from the command line:
https://github.com/DataDog/dd-trace-dotnet/blob/ab559b6be4742360d9265a4329a2505270842abb/build/tools/PrepareRelease/SetAllVersions.cs#L15-L20

lucaspimentel (approved):
Thank you! Just lefts some opinions on task names and such. Nothing blocking (we can always rename them later if we want to).

build/_build/Build.Utilities.cs
zacharycmontoya: I was going to suggest changing all of the "release" tools to have the same prefix but I thought it could be done in a separate PR:
- `UpdateIntegrationsJson`
- `UpdateVersion`
- `UpdateMsiContents`

build/_build/Build.Utilities.cs
andrewlock: I agree with you that imported/third party code is more descriptive, but currently we use "vendor" in a number of classes and namespaces around this, so I think it makes sense to update everywhere rather than just this target. That will require touching quite a lot of files though, so should probably be done separately (to the backlog!)

I updated this to `UpdateVendoredCode` as a compromise for now.

build/_build/Build.Utilities.cs
andrewlock: I agree, these make sense. I updated to use these suggestions in https://github.com/DataDog/dd-trace-dotnet/pull/1605/commits/ce91112ec00649f3bec6a87655d1b3e6653dd049

build/_build/Build.Utilities.cs
lucaspimentel: Thanks!

build/_build/Build.Utilities.cs
lucaspimentel: Thanks!

PR #1607, opened 2021-07-16
Add fallback for logs injection for NLog versions < 4.1.0

zacharycmontoya:
We advertise support beginning with NLog 4.0. Until NLog 4.1.0, there is only one Set API in the MappedDiagnosticsContext class: `Set(string, string)`. Our new `CustomNLogLogProvider` requires `Set(string, object)` API, which was introduced in NLog 4.1.0, to use our optimized `LogEnricher` type. Without this API present, we can't use our `CustomNLogLogProvider` type so revise the `IsLoggerAvailable` check to fail when the API is not found.

Note: I discovered this while testing cross-AppDomain calls with the various logging frameworks. I plan to post that PR separately, but I've verified that we find and use the fallback API in this scenario.

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace/Logging/CustomNLogLogProvider.cs
+++ b/src/Datadog.Trace/Logging/CustomNLogLogProvider.cs
@@ -13,6 +13,9 @@ namespace Datadog.Trace.Logging
     {
         public ILogEnricher CreateEnricher() => new LogEnricher(this);
 
+        internal static new bool IsLoggerAvailable() =>
+            NLogLogProvider.IsLoggerAvailable() && IsSetObjectAvailable();
+
         protected override OpenMdc GetOpenMdcMethod()
         {
             // This is a copy/paste of the base GetOpenMdcMethod, but calling Set(string, object) instead of Set(string, string)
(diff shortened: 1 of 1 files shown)


andrewlock (approved):

src/Datadog.Trace/Logging/CustomNLogLogProvider.cs
```diff
@@ -43,23 +43,48 @@ protected override OpenMdc GetOpenMdcMethod()
...
+                var set = Expression
+                    .Lambda<Action<string, string>>(setMethodCall, keyParam, valueParam)
+                    .Compile();
+                var remove = Expression
+                    .Lambda<Action<string>>(removeMethodCall, keyParam)
+                    .Compile();
+
+                return (key, value, _) =>
+                {
+                    set(key, value.ToString());
```
kevingosse, line 84: I wonder if we should just use this one all the time? 🤔

kevingosse (approved):

src/Datadog.Trace/Logging/CustomNLogLogProvider.cs
zacharycmontoya: That sounds fine to me. The only values we pass to this method are string objects anyways, so the impact of the added `ToString()` should be minimal

zacharycmontoya: 🤦🏼 we can't invoke `ToString()` here because we would immediately call `ToString()` on our `LogEnricher` object when we try to add it at the beginning of the trace. NLog needs the exact string representation when the API is called so I think the fallback is to actually avoid using our optimized `LogEnricher` in this case. I'll need to re-do the PR

zacharycmontoya:
Re-requesting reviews as I completely changed the solution to the problem

lucaspimentel (approved):

andrewlock (approved):

PR #1608, opened 2021-07-16
Make the benchmark step optional

kevingosse:
Failure of this job shouldn't prevent from merging a PR.


andrewlock (approved):
Should we make throughput optional too? Shame we can't use the AzDo app in Github to control this :(

lucaspimentel (approved):

PR #1614, opened 2021-07-19
Add additional logs injection fallback for NLog 1.x

zacharycmontoya:
This PR is based on the work from #1607, so that must be merged first.

The previous PR allows logs injection to fallback on LibLog's built-in log provider for NLog for versions < 4.0.0. However, the built-in log provider only looks for the MDC class `NLog.MappedDiagnosticsContext` to find the `Set`/`Remove` API's. These API's were introduced in NLog 2.x, so the lookup fails on NLog 1.x. This PR introduces a new custom NLog log provider to add a fallback path to search on the type `NLog.MDC` for the methods, which is the predecessor of the `NLog.MappedDiagnosticsContext` class.

Result: We no longer throw an unhandled exception when an application enables logs injection when using NLog 1.x.

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace/Logging/FallbackNLogLogProvider.cs
+++ b/src/Datadog.Trace/Logging/FallbackNLogLogProvider.cs
@@ -1,4 +1,4 @@
-// <copyright file="CustomNLogLogProvider.cs" company="Datadog">
+// <copyright file="FallbackNLogLogProvider.cs" company="Datadog">
 // Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2 License.
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
 // </copyright>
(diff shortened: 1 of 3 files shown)


lucaspimentel:
Do we need regression tests for this change?

zacharycmontoya:
Yes we definitely need regression tests for this change (and the other fixes related to log injection). I would like to add the regression tests in a separate PR once I'm able to polish them up, but in the meantime I've been finding a number of issues and would like to get them into the product ASAP because they can throw unhandled exceptions.

src/Datadog.Trace/Logging/LibLogScopeEventSubscriber.cs
```diff
@@ -271,13 +271,19 @@ private static void InitResolvers()
...
                 Tuple.Create<LogProvider.IsLoggerAvailable, LogProvider.CreateLogProvider>(
                     CustomNLogLogProvider.IsLoggerAvailable,
                     () => new CustomNLogLogProvider()));
 
+            LogProvider.LogProviderResolvers.Insert(
+                1,
+                Tuple.Create<LogProvider.IsLoggerAvailable, LogProvider.CreateLogProvider>(
+                    FallbackNLogLogProvider.IsLoggerAvailable,
+                    () => new FallbackNLogLogProvider()));
+
```
andrewlock, line 286: It's a minor thing, but I find it a bit confusing to be using index `1` here when we use `0` for all the other providers (meaning that they are added in increasing precedence order). The fallback provider is the only one that breaks that pattern.

That said, I totally understand why you went for that approach, as the fact we have to do Insert 0 is a bit annoying!

src/Datadog.Trace/Logging/FallbackNLogLogProvider.cs
```diff
@@ -0,0 +1,73 @@
...
+// </copyright>
+
+using System;
+using System.Linq.Expressions;
+using Datadog.Trace.Logging.LogProviders;
+
+namespace Datadog.Trace.Logging
+{
+    /// <summary>
+    /// Log provider for all versions of NLog lower than 4.1. This is required
```
andrewlock, line 13: Might be nice to clarify that this provider _supports_ all versions of NLog lower than 4.1, but it will only be _used_ with NLog version 1.x?

andrewlock (approved):

src/Datadog.Trace/Logging/LibLogScopeEventSubscriber.cs
kevingosse: I second this. Since the order of the providers is important, we need to make this code consistent. I don't mind if we use `Insert(0)` or `Insert(0,1,2,3,...)`, but let's use the same for all providers

zacharycmontoya: I went with the `Insert(0,1,2,3,...)` approach in my latest commit. Let me know if that looks good to you

src/Datadog.Trace/Logging/FallbackNLogLogProvider.cs
zacharycmontoya: Added more documentation in my latest commit. And it also clarifies that this is actually being used for all versions of NLog lower than 4.1. So it's the `CustomNLogLogProvider` for the most optimized performance in NLog >= 4.1 and `FallbackNLogLogProvider` for most compatible fallback for NLog < 4.1 😄

lucaspimentel (approved):

PR #1618, opened 2021-07-20
Fix throughtput/crank pipeline.

tonyredondo:
- Includes the project reference to the crank sample project to avoid the error trying to locate the managed assembly.

Due the crank pipeline does not use the home folder artifact but manually copies the native build to each crank agent, we need to add the managed profiler reference to the sample project, that way we ensure the required managed assemblies will be found at runtime.

@DataDog/apm-dotnet


lucaspimentel (approved):

lucaspimentel:
Remind me to change `Datadog.Trace.ClrProfiler.Managed` to `Datadog.Trace` in my branch when you merge this 😄

zacharycmontoya (approved):

PR #1619, opened 2021-07-20
don't include code analyzer when publishing projects

lucaspimentel:
Don't include the analyzer (added in #1325) when publishing projects. This makes the analyzer dlls and all the `System.*` it depends on to be included in the tracer home directory, the MSI installer, installed in the GAC, etc.

See https://github.com/dotnet/roslyn/issues/18093#issuecomment-405702631

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -30,7 +30,7 @@
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' != 'net45' ">
-    <ProjectReference Include="$(MSBuildThisFileDirectory)Datadog.Trace.Tools.Analyzers\Datadog.Trace.Tools.Analyzers.csproj" OutputItemType="Analyzer" PrivateAssets="All" />
+    <ProjectReference Include="$(MSBuildThisFileDirectory)Datadog.Trace.Tools.Analyzers\Datadog.Trace.Tools.Analyzers.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
   </ItemGroup>
 
 </Project>


zacharycmontoya (approved):

src/Directory.Build.props
```diff
@@ -30,7 +30,7 @@
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' != 'net45' ">
-    <ProjectReference Include="$(MSBuildThisFileDirectory)Datadog.Trace.Tools.Analyzers\Datadog.Trace.Tools.Analyzers.csproj" OutputItemType="Analyzer" PrivateAssets="All" />
+    <ProjectReference Include="$(MSBuildThisFileDirectory)Datadog.Trace.Tools.Analyzers\Datadog.Trace.Tools.Analyzers.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
```
andrewlock, line 33: I think we actually want both of these 
```suggestion
    <ProjectReference Include="$(MSBuildThisFileDirectory)Datadog.Trace.Tools.Analyzers\Datadog.Trace.Tools.Analyzers.csproj" OutputItemType="Analyzer" PrivateAssets="All" ReferenceOutputAssembly="false" />
```

andrewlock (approved):

src/Directory.Build.props
lucaspimentel: I don't know if that does anything for project references. [The docs](https://docs.microsoft.com/en-us/visualstudio/msbuild/common-msbuild-project-items#projectreference) for `ProjectReference` don't mention `PrivateAssets`. I think that only applies to `PackageReference` ([docs](https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#controlling-dependency-assets)).

andrewlock: I was going on the trusted advice of [Nick Craver](https://twitter.com/Nick_Craver/status/1256365611455840256) and [others](https://www.meziantou.net/referencing-an-analyzer-from-a-project.htm)😛 You don't think the docs are actually correct do you? 😉

lucaspimentel: Haha. Well, it's working correctly as-is and I think I trust the people working on .NET a little more (see the comment I linked in the PR description).

PR #1620, opened 2021-07-21
Include Sqreen.dll in the msi

robertpi:


lucaspimentel (approved):

build/_build/Build.Steps.cs
```diff
@@ -37,6 +37,7 @@ partial class Build
     AbsolutePath ArtifactsDirectory => Artifacts ?? (OutputDirectory / "artifacts");
     AbsolutePath WindowsTracerHomeZip => ArtifactsDirectory / "windows-tracer-home.zip";
     AbsolutePath BuildDataDirectory => RootDirectory / "build_data";
+    AbsolutePath LibSqreenDirectory => RootDirectory / "packages" / "libsqreen.1.1.2.2";
```
lucaspimentel, line 40: I assume we have a package reference somewhere that pulls down the nuget package?

andrewlock, line 40: Assuming that this is pulled in using a normal nuget restore, we should probably use something like this:
```suggestion
    AbsolutePath LibSqreenDirectory => (NugetPackageDirectory ?? (RootDirectory / "packages")) / "libsqreen.1.1.2.2";
```

andrewlock (approved):

src/WindowsInstaller/Product.wxs
```diff
@@ -125,17 +125,25 @@
...
-    <!-- For the 64-bit installer, also install the 32-bit profiler -->
+    <!-- For the 64-bit installer, also install the 32-bit binaries -->
     <ComponentGroup Id="Files.Native.32" Directory="INSTALLFOLDER.32">
       <Component Win64="no" Id="Datadog.Trace.ClrProfiler.Native.32">
         <File Id="Datadog.Trace.ClrProfiler.Native.32"
               Source="$(var.TracerHomeDirectory)\win-x86\Datadog.Trace.ClrProfiler.Native.dll"
               Checksum="yes">
           <Class Id="$(var.ProfilerCLSID)" Context="InprocServer32" ThreadingModel="both" Description="$(var.ProductNamePlatformAgnostic)"/>
         </File>
+        <File Id="LibSqreen.32"
```
zacharycmontoya, line 143: How is Sqreen located by the Tracer? The only other file in this 32-bit directory is Datadog.Trace.ClrProfiler.Native.dll, which is usually found via the Profiler CLSID

robertpi: Yes, this is a bit tricky. There are several approaches, but after discussing it at Tuesday evening meeting we ended up deciding to use the DD env var to search for the dll and load it using system calls. It's implemented here: https://github.com/DataDog/dd-trace-dotnet/blob/appsec/v0-integration/src/Datadog.Trace/AppSec/Waf/NativeBindings/Native.cs#L38

build/_build/Build.Steps.cs
robertpi: It's a good point, it was only referenced in the samples, as we only need it at runtime, on to build against, I've added it to packages.config in Datadog.Trace.ClrProfiler.Native. Another way of accessing may be need for the Linux installer, but that's the next task.

PR #1625, opened 2021-07-22
Logger API refactoring

gleocadie:
The goal of this PR is to allow having one logger per component (Tracer, Profiler, Common libs, ...)

* Move logging free functions (Info, Debug, Warn) to static function member of Logger class
* Rewrite the logger implementation to allow one logger for the Tracer, for the Profiler,
  for the Common libraries, and more... 

@DataDog/apm-dotnet


src/Datadog.Trace.ClrProfiler.Native/logger.h
```diff
@@ -0,0 +1,88 @@
...
+#ifdef _WIN32
+    // this field will be removed once merged with the profiler in order to have
+    // the same product folder name
+    inline static const WSTRING product_folder_name = WStr("Datadog .NET Tracer");
+#endif
+    inline static const std::string pattern = "%D %I:%M:%S.%e %p [%P|%t] [%l] %v";
+    struct logging_environment
+    {
+        // cannot reuse environment::log_path variable. On alpine, test fails
+        inline static const WSTRING log_path = WStr("DD_TRACE_LOG_PATH");
```
gleocadie, line 25: I do not know why on Alpine the test `Datadog.Trace.ClrProfiler.IntegrationTests.NoMultiLoaderTests.SingleLoaderTest` if we have this line instead:
`inline static const WSTRING log_path = environment::log_path;`

src/Datadog.Trace.ClrProfiler.Native/pal.h
```diff
@@ -37,34 +40,29 @@ inline WSTRING DatadogLogFilePath(const std::string& file_name_suffix)
...
-        program_data = R"(C:\ProgramData)";
+        program_data_path = WStr(R"(C:\ProgramData)");
     }
 
-    return ToWSTRING(program_data + R"(\Datadog .NET Tracer\logs\dotnet-tracer-native)" + file_name_suffix + ".log");
+    // on Windows WSTRING == wstring
+    return (program_data_path / TLoggerPolicy::product_folder_name / WStr("logs") / file_name).wstring();
 #else
-    return ToWSTRING("/var/log/datadog/dotnet/dotnet-tracer-native" + file_name_suffix + ".log");
+    // TODO: should we consider having a logs folder on non-Windows ?
```
lucaspimentel, line 64:
```
// TODO: should we consider having a logs folder on non-Windows ?
```
In Windows we create a `logs` folder because there could potentially be other files in the parent folder (`C:\ProgramData\Datadog .NET Tracer`), like config files or other data. See `C:\ProgramData\Datadog` and `C:\ProgramData\Datadog\logs`, for example.

On Linux, this folder is in `/var/log`, so it's expected to only contain log files, right?

gleocadie: 🤔 good one. `/var/log` is expected to contain log files

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -50,7 +53,7 @@ HRESULT STDMETHODCALLTYPE CorProfiler::Initialize(IUnknown* cor_profiler_info_un
     // check if tracing is completely disabled
     if (IsTracingDisabled())
     {
-        Info("DATADOG TRACER DIAGNOSTICS - Profiler disabled in ", environment::tracing_enabled);
+        Logger::Logger::Info("DATADOG TRACER DIAGNOSTICS - Profiler disabled in ", environment::tracing_enabled);
```
tonyredondo, line 56: `Logger::Logger::Info` here and `Logger::Info` in the same file, what's the difference?

gleocadie: Oops

gleocadie: done in https://github.com/DataDog/dd-trace-dotnet/pull/1625/commits/d419680b0e56138ce4a9e164f75155a83b7516c2

tonyredondo (approved):
LGTM, only thing that bothers me are the inlines to make it work in Alpine, but overall great work!

lucaspimentel (approved):

PR #1626, opened 2021-07-22
Add lib sqreen to unix packages

robertpi:
Fixes #

@DataDog/apm-dotnet

### DIFF
--- a/build/_build/Build.Steps.cs
+++ b/build/_build/Build.Steps.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Net;
 using System.Runtime.InteropServices;
 using System.Text.RegularExpressions;
 using Nuke.Common;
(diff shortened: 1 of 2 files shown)


build/_build/Build.Steps.cs
```diff
@@ -254,6 +255,52 @@ partial class Build
...
 
+    Target DownloadLibSqreen => _ => _
+        .Unlisted()
+        .After(CreateRequiredDirectories)
+        .OnlyWhenStatic(() => IsLinux)
+        .Executes(() =>
+        {
+            var wc = new WebClient();
+            var libSqreenUri = new Uri("https://www.nuget.org/api/v2/package/libsqreen/1.1.2.2");
+            var libSqreenZip = RootDirectory / "libsqreen.zip";
```
andrewlock, line 266: Can we download this to `TempDirectory` instead?
```suggestion
            var libSqreenZip = TempDirectory / "libsqreen.zip";
```

build/_build/Build.Steps.cs
```diff
@@ -254,6 +255,52 @@ partial class Build
...
+            Console.WriteLine($"{libSqreenZip} downloaded. Extracting to {LibSqreenDirectory}...");
+
+            UncompressZip(libSqreenZip, LibSqreenDirectory);
+        });
+
+    Target CopyLibSqreen => _ => _
+        .Unlisted()
+        .After(Clean)
+        .After(DownloadLibSqreen)
+        .OnlyWhenStatic(() => IsWin || LinuxArchitectureIdentifier != "arm64")
```
andrewlock, line 278:
```suggestion
        .OnlyWhenStatic(() => IsWin || !IsArm64)
```

build/_build/Build.Steps.cs
```diff
@@ -254,6 +255,52 @@ partial class Build
...
+                    CopyFileToDirectory(source, dest, FileExistsPolicy.Overwrite);
+                }
+            }
+            else
+            {
+                var (architecture, ext) = GetUnixArchitectureAndExtention();
+                var sqreenFileName = $"libSqreen.{ext}";
+
+                var source = LibSqreenDirectory / "runtimes" / architecture / "native" / sqreenFileName;
+                var dest = TracerHomeDirectory;
```
andrewlock, line 297: For windows we write to `TracerHomeDirectory / architecture`, is that what you mean here too?
```suggestion
                var dest = TracerHomeDirectory / architecture;
```

build/_build/Build.Steps.cs
```diff
@@ -254,6 +255,52 @@ partial class Build
...
         });
 
+    Target DownloadLibSqreen => _ => _
+        .Unlisted()
+        .After(CreateRequiredDirectories)
+        .OnlyWhenStatic(() => IsLinux)
+        .Executes(() =>
+        {
+            var wc = new WebClient();
+            var libSqreenUri = new Uri("https://www.nuget.org/api/v2/package/libsqreen/1.1.2.2");
```
andrewlock, line 265: Given we have to manually download this for Linux, doesn't it make sense to remove it from the packages.config, and do the same thing on both platforms? That way we don't have to keep version numbers in sync etc?

Also, I'm thinking we should make this version number a const to be obvious?

build/_build/Build.Steps.cs
robertpi: This is to tracer-home, my understanding is that this directory should contain the binaries in the structure they're need for the Unix installers. As we only support 64-bit under *nix OSs we place the libraries in a root. That why `CreateDdTracerHome` has actions to move them to architecture specific folders.

(Which means the changes in `CreateDdTracerHome` could be wrong, since we copy, rather than move, the sqreen library here).

andrewlock: Yeah, it should mirror the tracer home, but I'd be strongly inclined to use the dd-trace layout nevertheless. Especially given we're considering switching to that in 2.0...

build/_build/Build.Steps.cs
robertpi: I've removed the predicate, so it'll run both under Windows and Linux, as I agree this will help build stability. I think we should keep the package.config reference as this will help people that want to use Visual Studio to build and run the tests.

Moved the version over to a constant too.

build/_build/Build.Steps.cs
robertpi: I prefer not to mix the two layout styles, If we do migrate, I think it would be easier migrate both libraries to a different location, rather than trying to understand why the two libraries are stored in different location.

PR #1627, opened 2021-07-23
Run the security tests as part of the CI

robertpi:
Fixes #

@DataDog/apm-dotnet


anna-git (approved):

test/Datadog.Trace.Security.IntegrationTests/AspNetCore5.cs
```diff
@@ -22,6 +22,8 @@ public AspNetCore5(ITestOutputHelper outputHelper)
         [Theory]
         [InlineData(true, HttpStatusCode.Forbidden)]
         [InlineData(false, HttpStatusCode.OK)]
+        [Trait("RunOnWindows", "True")]
+        [Trait("Category", "ArmUnsupported")]
```
andrewlock, line 26: Is the appsec stuff in general not supported on ARM? or is it something about this test?

build/_build/Build.Steps.cs
```diff
@@ -75,6 +75,7 @@ partial class Build
     Project[] ParallelIntegrationTests => new[]
     {
         Solution.GetProject(Projects.TraceIntegrationTests),
+        Solution.GetProject(Projects.AppSecIntegrationTests),
```
andrewlock, line 78: Is this definitely safe to run in parallel? It uses IIS, and touches various samples, so just wondered. It shouldn't be a problem as the ClrProfiler tests aren't running in parallel with it, just wanted to double check :)

test/Datadog.Trace.Security.IntegrationTests/AspNetCore5.cs
robertpi: We don't yet have an arm build of the C++ Sqreen Library that does actually analysis, so we can't support it yet. I checked on the status of this the other day, the code cross compiles, it just needs releasing. So, support is coming soon.

andrewlock: Cool, thanks 👍

build/_build/Build.Steps.cs
robertpi: It does use IIS, I wasn't sure of the criteria for running in parallel, so I just guessed that it would be okay.

I guess will just see if it works for now. I think we'll probably move it to it's own build step long term, but we have too few tests to justify that at the moment.

andrewlock: Yeah, that makes sense. There isn't hard and fast criteria for running in parallel AFAIK (the approach here is what existed pre-Nukeification), but if it's using IIS it definitely _can't_ run at the same time as the ClrProfiler tests.

robertpi:
merged to https://github.com/DataDog/dd-trace-dotnet/tree/appsec/v0-integration, as part of https://github.com/DataDog/dd-trace-dotnet/pull/1638.

PR #1628, opened 2021-07-23
Enable experimental NGEN support

tonyredondo:
This PR enables an experimental support for NGEN images.

@DataDog/apm-dotnet


src/Datadog.Trace.ClrProfiler.Native/rejit_handler.cpp
```diff
@@ -168,7 +249,26 @@ void RejitHandler::EnqueueThreadLoop(RejitHandler* handler)
             break;
         }
 
-        hr = profilerInfo->RequestReJIT((ULONG) item->m_length, item->m_modulesId.get(), item->m_methodDefs.get());
+        if (profilerInfo10 != nullptr)
+        {
+            // RequestReJITWithInliners is currently always failing...
+            // more research is required, meanwhile we fallback to the normal RequestReJIT and manual track of inliners.
```
zacharycmontoya, line 255: If might be good to send an email to David Mason to ask any questions about the API and if there are any results that look wrong

src/Datadog.Trace.ClrProfiler.Native/rejit_handler.cpp
```diff
@@ -75,6 +74,79 @@ void RejitHandlerModuleMethod::SetMethodReplacement(const MethodReplacement& met
...
+                }
+                methodEnum->Release();
+                methodEnum = nullptr;
+                if (total > 0)
+                {
+                    handler->EnqueueForRejit(modules, methods);
+                    Info("NGEN:: Processed with ", total, " inliners [ModuleId=", currentModuleId,
+                         ",MethodDef=", currentMethodDef, "]");
+                }
+                m_ngenModules[moduleId] = true;
```
zacharycmontoya, line 127: On https://docs.microsoft.com/en-us/dotnet/framework/unmanaged-api/profiling/icorprofilerinfo6-enumngenmodulemethodsinliningthismethod-method#remarks, it says that the `incompleteData` result will return true if the enum does not contain all of the caller methods. Should we do the ReJIT regardless of the result and then only save this `m_ngenModules[moduleId] = true` result at the end of this `!incompleteData` block? I'm concerned about a scenario where we always get `incompleteData==true` so we never ReJIT a group of methods that in-line a call we care about

zacharycmontoya (approved):
Overall this looks good! Nice work!

src/Datadog.Trace.ClrProfiler.Native/rejit_handler.cpp
tonyredondo: Yes, makes sense to me. I wonder if we should track at method level then? Or it doesn’t matter if we call ReJIT multiple times? WDYT?

zacharycmontoya: Perhaps we can start by calling ReJIT each time and if we see large overheard we can revisit it?

tonyredondo: Agreed. Done in ae2b27d

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -3025,23 +3025,12 @@ HRESULT STDMETHODCALLTYPE CorProfiler::JITCachedFunctionSearchStarted(FunctionID
 
     if (!has_loader_injected_in_appdomain)
     {
+        Debug("Disabling NGEN due missing loader...");
         // The loader is missing in this AppDomain, we skip the NGEN image to allow the JITCompilationStart inject it.
         *pbUseCachedFunction = false;
         return S_OK;
     }
 
-    RejitHandlerModule* handlerModule = nullptr;
```
zacharycmontoya: How did this fix issue? It looks like we only set `*pbUseCachedFunction = false;` when the loader is missing so I don't understand how this works

tonyredondo: Yes, I found that RequestReJIT is doing that work already, so we only need to ReJIT an NGEN method to do the rewriting. If I set the * pbUseCachedFunction flag to false that triggers a normal JITCompilation with the original IL that replaces our rewrite. That’s why we weren’t instrumenting in that case.

zacharycmontoya: Oh I see, so we want to set the flag to true and let RequestReJIT take care of the rewriting. That makes sense, thanks for explaining!

tonyredondo: That’s right and keep it disabled until the loader inject itself in the JITCompilationStarted callback.

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -233,6 +248,17 @@ HRESULT STDMETHODCALLTYPE CorProfiler::Initialize(IUnknown* cor_profiler_info_un
...
+    if (!is_calltarget_enabled || !is_net46_or_greater || !IsNGENEnabled())
+    {
+        Info("NGEN is disabled.");
+        event_mask |= COR_PRF_DISABLE_ALL_NGEN_IMAGES;
+    }
+    else
+    {
+        Info("NGEN is enabled.");
+        event_mask |= COR_PRF_MONITOR_CACHE_SEARCHES;
+    }
```
lucaspimentel, line 260: not a blocker, but I find it easier to understand the logic if we reverse it:

```suggestion
    if (is_calltarget_enabled && is_net46_or_greater && IsNGENEnabled())
    {
        Info("NGEN is enabled.");
        event_mask |= COR_PRF_MONITOR_CACHE_SEARCHES;
    }
    else
    {
        Info("NGEN is disabled.");
        event_mask |= COR_PRF_DISABLE_ALL_NGEN_IMAGES;
    }
```

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -2945,6 +2977,64 @@ HRESULT STDMETHODCALLTYPE CorProfiler::ReJITError(ModuleID moduleId, mdMethodDef
...
+        return S_OK;
+    }
+
+    const bool has_loader_injected_in_appdomain =
+        first_jit_compilation_app_domains.find(module_metadata->app_domain_id) !=
+        first_jit_compilation_app_domains.end();
+
+    if (!has_loader_injected_in_appdomain)
+    {
+        Debug("Disabling NGEN due missing loader...");
```
lucaspimentel, line 3028:
```suggestion
        Debug("Disabling NGEN due to missing loader.");
```

lucaspimentel (approved):

kevingosse (approved):

PR #1630, opened 2021-07-26
Clear metadata when an appdomain unloads

kevingosse:
We keep some metadata to know in which appdomains the tracer has been loaded. Unfortunately we don't clear that metadata when an appdomain is unloaded, so if a new appdomain is created with the same id then the tracer won't be loaded.

This scenario occasionally occurs in the MultiDomainHost tests.


tonyredondo (approved):
Weird it uses the same AppDomainId...

lucaspimentel (approved):

zacharycmontoya (approved):

PR #1631, opened 2021-07-26
replace "minimal" solution with a solution filter

lucaspimentel:
- ~add missing `Datadog.Trace.Tools.Analyzers`~
- ~remove deleted `Datadog.Core.Tools`~

~Does anyone still use this solution? If not, we should delete it in favor of solution filters (`.slnf` file), instead of going through the trouble of updating.~

Change of plans: replace `Datadog.Trace.Minimal.sln` with a [solution filter](https://docs.microsoft.com/en-us/visualstudio/ide/filtered-solutions).


andrewlock (commented):
I don't use it, so +1 for deleting from me

robertpi:
I have used it, but I was unaware of .slnf files, if they work, they would seem like a better idea that separate solution files.

zacharycmontoya (approved):

andrewlock (approved):

PR #1636, opened 2021-07-28
Fix MockTracerAgent listener thread not closing

andrewlock:
As noted elsewhere, `listener.GetContext()` doesn't throw on .NET Core when the listener is stopped, which means the thread never ends. `GetContextAsync()` _does_ throw however, so refactored to use a task based approach instead. 

Without this, the Rider test process hangs after running integration tests.

### DIFF
--- a/test/Datadog.Trace.TestHelpers/MockTracerAgent.cs
+++ b/test/Datadog.Trace.TestHelpers/MockTracerAgent.cs
@@ -192,7 +192,7 @@ namespace Datadog.Trace.TestHelpers
 
         public void Dispose()
         {
-            _listener?.Stop();
+            _listener?.Abort();
             _cancellationTokenSource.Cancel();
             _udpClient?.Close();
         }


test/Datadog.Trace.TestHelpers/MockTracerAgent.cs
```diff
@@ -195,6 +195,7 @@ public void Dispose()
             _listener?.Stop();
             _cancellationTokenSource.Cancel();
             _udpClient?.Close();
+            _listenerTask.Wait();
```
zacharycmontoya, line 198: Will this Wait call throw an exception?

andrewlock: Not in normal usage, no. In normal usage, the `GetContextAsync()` call throws when we call `_listener.Stop()`, but we swallow it, and exit the loop as `_listener.IsListening`. We _should_ probably have a `?` in there though.

zacharycmontoya (approved):

lucaspimentel (approved):

andrewlock:
Well, that doesn't work 🤔 Apparently this causes lock ups in CI (despite _fixing_ that exact issue for me locally with Rider) 🤷

zacharycmontoya:
Is it time to close this PR @andrewlock ?

andrewlock:
> Is it time to close this PR @andrewlock ?

Yeah, definitely. Wish I knew what was going on though 🙄

PR #614, opened 2020-01-24
Fix Trace Id Injection NullReferenceException with NLog 1.0.0.505

zacharycmontoya:
Issue:
Currently using the Trace Id Injection fails with a NullReferenceException when the instrumented application is using NLog 1.0.0.505. This occurs because the LibLog library we use to detect the instrumented application's logging framework assumes that all versions of NLog have the type `NLog.MappedDiagnosticsContext` for using the Mdc features. Unfortunately this is incorrect in one case: the very first NLog library v1.0.0.505 had its Mdc features in a type named `NLog.MDC`. Rather than vendor LibLog to support this, for now we will catch the issue and disable Trace Id Injection for the application.

@DataDog/apm-dotnet

PR #551, opened 2019-11-11
Fix DD_LOGS_INJECTION crash with the ASP.NET integration

zacharycmontoya:
Fixes:
`SerializationException: Type is not resolved for member 'log4net.Util.PropertiesDictionary,log4net, Version=2.0.8.0, Culture=neutral, PublicKeyToken=669e0ddf0bb1aa2a'`

Scenario:
This crash occurs when 3 conditions are met:
1. The user is running an ASP.NET application in IIS
1. The user references the `Datadog.Trace.ClrProfiler.Managed` NuGet package (which is required for the `AspNet` integration) up to and including `1.9.0`
1. `DD_LOGS_INJECTION=true`

This occurs because at the time of the Asp.Net HttpModule startup, we initialize the Tracer and set default values of `"0"` for `"dd.trace_id"` and `"dd.span_id"`. When the initialization is finished and control is passed back to IIS, an AppDomain transition occurs. At this time, the `log4net.Util.PropertiesDictionary` that held the `dd.trace_id` and `dd.span_id` properties must be passed across AppDomains because it was stored inside the `System.Runtime.Remoting.Messaging.CallContext` structure. When IIS receives the object, it must be deserialized, at which point it crashes because the `log4net` assembly cannot be found: it was app-local.

The solution is to not assign default values when the Tracer is created, only assigning values to `dd.trace_id` and `dd.span_id` in response to the changing state of active spans in the application.

Changes proposed in this pull request:
- When `DD_LOGS_INJECTION=true`, do not set default values for `dd.trace_id` and `dd.span_id` when the Tracer is created, because it may not be safe to do so at that time.
- Add unit tests to ensure this behavior
- Add smoke test to replicate the IIS scenario of running another application in a separate AppDomain, with the `log4net` dependency, with `DD_LOGS_INJECTION=true`

@DataDog/apm-dotnet

PR #1637, opened 2021-07-28
Add end-to-end integration tests for logs injection

zacharycmontoya:
New Tests
- `LogsInjection.Log4Net`, `LogsInjection.NLog`, `LogsInjection.Serilog`
- Behaviors
  - Logs injection occurs with raw logs and JSON logs
  - After logs injection setup, a method is invoked another AppDomain to test LogicalCallContext storage
- Libraries tested
  - log4net
    - .NET Framework: 1.2 - 2.x
    - .NET Core: 2.0.6+
  - NLog
    - .NET Framework: 1.x - 4.x
    - .NET Core: 4.5+
  - Serilog
    - .NET Framework: 1.x - 2.x
    - .NET Core: 2.x
- Caveats
  - **Note**: Windows testing currently does not test multiple package versions. That will need to be enabled at a later point in time
  - All log4net versions are still vulnerable to serialization issues when crossing AppDomains, so the cross-AppDomain scenario is skipped in the log4net test application

Deleted Tests
- `Log4Net.SerializationException` regression test (and helper library ApplicationWithLog4Net)
  - The regression test was added in https://github.com/DataDog/dd-trace-dotnet/pull/551 but is now solved by adding new sample applications that directly cause calls to cross AppDomain boundaries
- `NLog10LogsInjection.NullReferenceException` regression test
  - The regression test was added in https://github.com/DataDog/dd-trace-dotnet/pull/614 but is now solved by the new type FallbackNLogLogProvider

Fixes
- Disable Serilog logs injection support when the detected Serilog version is earlier than 2.0. This is enforced by adding a new `NoOpSerilogLogProvider` log provider to perform no-ops in this scenario
- Add better exception handling in `CustomLog4NetLogProvider.ExtJsonAssemblySupported()`

@DataDog/apm-dotnet


build/PackageVersionsGeneratorDefinitions.json
```diff
@@ -121,5 +121,29 @@
...
+    "MinVersionNetCore": "4.5.0",
+    "MaxVersionExclusive": "5.0.0"
+  },
+  {
+    "IntegrationName": "log4net",
+    "SampleProjectName": "LogsInjection.Log4Net",
+    "NugetPackageSearchName": "log4net",
+    "MinVersion": "1.0.0",
+    "MinVersionNetCore": "2.0.6",
+    "MaxVersionExclusive": "3.0.0"
```
andrewlock, line 147: We should add these sample names to the `multiApiProjects` collection in `CompileSamplesLinux`, so we don't build them multiple times (I'll update that target to automatically handle that soon!)

src/Datadog.Trace/Logging/NoOpSerilogLogProvider.cs
```diff
@@ -0,0 +1,39 @@
...
+        }
+#pragma warning restore SA1313 // Parameter names should begin with lower-case letter
+
+        private static bool SerilogAssemblySupported()
+        {
+            var ndcContextType = FindType("Serilog.Context.LogContext", new[] { "Serilog", "Serilog.FullNetFx" });
+            var serilogVersion = ndcContextType.Assembly.GetName().Version;
+            return serilogVersion >= SupportedSerilogAssemblyVersion;
+        }
+    }
```
andrewlock, line 38: Shouldn't this support Serilog 1.0 too, not just 2.0+?

test/Datadog.Trace.ClrProfiler.IntegrationTests/SerilogTests.cs
```diff
@@ -0,0 +1,80 @@
...
+                Assert.True(spans.Count >= 1, $"Expecting at least 1 span, only received {spans.Count}");
+
+                if (string.IsNullOrWhiteSpace(packageVersion) || new Version(packageVersion) >= new Version("2.0.0"))
+                {
+                    ValidateLogCorrelation(spans, _log200FileTests);
+                }
+                else
+                {
+                    // We do not expect logs injection for Serilog versions < 2.0.0 so filter out all logs
+                    ValidateLogCorrelation(spans, _logPre200FileTests, additionalInjectedLogFilter: (_) => false);
```
andrewlock, line 75: I wonder if we should explicitly assert that we _don't_ have logs injection in this case? Otherwise it's not confirming that the noop logger is used?

test/test-applications/integrations/LogsInjection.Log4Net/LogsInjection.Log4Net.csproj
```diff
@@ -0,0 +1,49 @@
...
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(ApiVersion)' &gt;= '2.0.5'">
+    <PackageReference Include="log4net.Ext.Json" Version="1.2.15.14586" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\dependency-libs\LogsInjectionHelper\LogsInjectionHelper.csproj" />
+    <ProjectReference Include="..\dependency-libs\PluginApplication\PluginApplication.csproj" />
+    <ProjectReference Include="..\..\..\..\src\Datadog.Trace\Datadog.Trace.csproj" />
```
andrewlock, line 27: Do we need/want this reference? It's currently referenced transitively, and isn't used directly by this app?

test/test-applications/integrations/LogsInjection.NLog/LogsInjection.NLog.csproj
```diff
@@ -0,0 +1,45 @@
...
+
+  <ItemGroup>
+    <PackageReference Include="NLog" Version="$(ApiVersion)" />
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\dependency-libs\LogsInjectionHelper\LogsInjectionHelper.csproj" />
+    <ProjectReference Include="..\dependency-libs\PluginApplication\PluginApplication.csproj" />
+    <ProjectReference Include="..\..\..\..\src\Datadog.Trace\Datadog.Trace.csproj" />
```
andrewlock, line 23: Do we need/want this reference? It's currently referenced transitively, and isn't used directly by this app?

test/test-applications/integrations/LogsInjection.Serilog/LogsInjection.Serilog.csproj
```diff
@@ -0,0 +1,54 @@
...
+  <ItemGroup Condition="'$(ApiVersion)' &gt;= '2.0.0' and '$(TargetFramework)' != 'net452'">
+    <!-- Directly reference dependencies of Serilog.Sinks.File to avoid package downgrade errors -->
+    <PackageReference Include="System.IO.FileSystem" Version="4.3.0" />
+    <PackageReference Include="System.IO.FileSystem.Primitives" Version="4.3.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\dependency-libs\LogsInjectionHelper\LogsInjectionHelper.csproj" />
+    <ProjectReference Include="..\dependency-libs\PluginApplication\PluginApplication.csproj" />
+    <ProjectReference Include="..\..\..\..\src\Datadog.Trace\Datadog.Trace.csproj" />
```
andrewlock, line 41: Do we need/want this reference? It's currently referenced transitively, and isn't used directly by this app?

test/test-applications/integrations/dependency-libs/LogsInjectionHelper/LoggingMethods.cs
```diff
@@ -0,0 +1,115 @@
...
+            // Create an AppDomain with that directory as the appBasePath
+            var entryDirectory = Directory.GetParent(Assembly.GetEntryAssembly().Location);
+            var applicationFilesDirectory = Path.Combine(entryDirectory.FullName, "ApplicationFiles");
+            var applicationAppDomain = AppDomain.CreateDomain("ApplicationAppDomain", null, applicationFilesDirectory, applicationFilesDirectory, false);
+#endif
+
+            // Set up Tracer and start a trace
+            var settings = TracerSettings.FromDefaultSources();
+            settings.LogsInjectionEnabled = true;
+            settings.Environment ??= "dev"; // Later we can test when Environment=null / dd.env=null
```
andrewlock, line 41: FYI, by default, when starting sample apps, we set `DD_ENV` to `integration_tests` for consistency: 
https://github.com/DataDog/dd-trace-dotnet/blob/master/test/Datadog.Trace.TestHelpers/EnvironmentHelper.cs#L217

test/test-applications/integrations/dependency-libs/LogsInjectionHelper/LogsInjectionHelper.csproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net45;net461;netstandard2.0</TargetFrameworks>
+    <LangVersion>latest</LangVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\..\..\..\src\Datadog.Trace\Datadog.Trace.csproj" />
```
andrewlock, line 9: I thought there were issues with code coverage when referencing the Datadog.Trace.dll directly, but I see some other test-apps do it, so I'm not sure now. @kevingosse?

andrewlock (approved):
LGTM👍 mostly minor comments. The test helpers should be useful for the ILogger work too🎉

build/PackageVersionsGeneratorDefinitions.json
andrewlock: Did a small PR to handle this automatically here https://github.com/DataDog/dd-trace-dotnet/pull/1646

andrewlock:
Minor thing really, but we now have a ton of errors in the build output 
```
\test\test-applications\integrations\LogsInjection.Log4Net\LogsInjection.Log4Net.csproj : warning NU1701: Package 'log4net.Ext.Json 1.2.15.14586' was restored using '.NETFramework,Version=v4.6.1, .NETFramework,Version=v4.6.2, .NETFramework,Version=v4.7, .NETFramework,Version=v4.7.1, .NETFramework,Version=v4.7.2, .NETFramework,Version=v4.8' instead of the project target framework '.NETCoreApp,Version=v2.1'. This package may not be fully compatible with your project.
```

Can we supress those errors (pretty sure we/I did for other similar complaints)?

test/Datadog.Trace.ClrProfiler.IntegrationTests/LogsInjectionTestBase.cs
```diff
@@ -0,0 +1,141 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Datadog.Trace.TestHelpers;
+using FluentAssertions;
+using FluentAssertions.Execution;
+using Xunit.Abstractions;
+
+namespace Datadog.Trace.ClrProfiler.IntegrationTests
+{
+    public abstract class LogsInjectionTestBase : TestHelper
```
andrewlock, line 20: Could we move the bulk of this to a helper class instead? I'd like to use it in the aspnetcore tests too, but this has different setup around finding the sample name etc and the helpers are static anyway as far as I can see?

src/Datadog.Trace/Logging/NoOpSerilogLogProvider.cs
zacharycmontoya: I just realized that this `SerilogAssemblySupported()` implementation is the same as the definition in `CustomSerilogLogProvider` except our `IsLoggerAvailable()` implementation checks for `!SerilogAssemblySupported()`. Which style do you think more clearly represents that we want to disable Serilog injection on assemblies lower than 2.0?
- Keep the `SerilogAssemblySupported()` logic and assert `!SerilogAssemblySupported()`
- Rename and reimplement `SerilogAssemblyUnsupported` and assert `SerilogAssemblyUnsupported()`

build/PackageVersionsGeneratorDefinitions.json
zacharycmontoya: It looks like this is handled. If there's anything else for me to do, let me know, or you can resolve this

test/test-applications/integrations/dependency-libs/LogsInjectionHelper/LogsInjectionHelper.csproj
zacharycmontoya: @kevingosse we recently merged everything into one `Datadog.Trace.dll` and we have a bunch of applications that directly reference it. Is our code coverage okay or do we need to change our test applications somehow?

test/test-applications/integrations/dependency-libs/LogsInjectionHelper/LoggingMethods.cs
zacharycmontoya: Yep 👍🏼 I will add a comment to remove any confusion. I wrote this to make sure local dev testing had it assigned.

Fixed in 58967e1c9556e1eda67a978f2acb5a2a7713381f

test/test-applications/integrations/LogsInjection.NLog/LogsInjection.NLog.csproj
zacharycmontoya: Good catch, we don't need this anymore. Before, the app was using this directly but now it's being used transitively. I'll remove the project reference 👍🏼 Fixed in a1c64bfee75f3e355c2262dc33fc5dab3846e54e

test/test-applications/integrations/LogsInjection.Serilog/LogsInjection.Serilog.csproj
zacharycmontoya: Fixed in a1c64bfee75f3e355c2262dc33fc5dab3846e54e

test/test-applications/integrations/LogsInjection.Log4Net/LogsInjection.Log4Net.csproj
zacharycmontoya: Fixed in a1c64bfee75f3e355c2262dc33fc5dab3846e54e

test/Datadog.Trace.ClrProfiler.IntegrationTests/SerilogTests.cs
zacharycmontoya: Good idea. Done in ebbe57057684f7e18e04146cba343448d8d3c64c . I figured we should probably also test that we don't have correlated logs when DD_LOGS_INJECTION=false, so I tested that as well

robertpi (approved):
Looks good to me.

src/Datadog.Trace/Logging/CustomLog4NetLogProvider.cs
```diff
@@ -82,10 +82,18 @@ protected override OpenMdc GetOpenMdcMethod()
...
+            try
             {
-                return extJsonAssembly.GetName().Version >= SupportedExtJsonAssemblyVersion;
+                Assembly extJsonAssembly = Assembly.Load("log4net.Ext.Json");
+                if (extJsonAssembly is not null)
+                {
+                    return extJsonAssembly.GetName().Version >= SupportedExtJsonAssemblyVersion;
+                }
+            }
+            catch
```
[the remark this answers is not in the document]
zacharycmontoya: If the instrumented application doesn't use the JSON library then we'll always enter the catch block once during logs injection setup, but it doesn't necessarily mean that anything is wrong. So I preferred not writing any logs because that might cause confusion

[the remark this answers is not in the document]
zacharycmontoya: Maybe this could be a debug-level log?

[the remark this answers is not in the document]
zacharycmontoya: Added a log at the debug log level in a405f0d6c62581810788237a4e01168aacc4cb9a

PR #1639, opened 2021-07-29
Ignore empty values when parsing headers

kevingosse:
There has been one case of customer seeing a lot of warning logs because of an empty `x-datadog-sampling-priority`. While the provenance of this header is unclear, we can save ourselves a lot of trouble if we ignore empty values instead of treating them as invalid.

### DIFF
--- a/src/Datadog.Trace/SpanContextPropagator.cs
+++ b/src/Datadog.Trace/SpanContextPropagator.cs
@@ -238,6 +238,11 @@ namespace Datadog.Trace
 
             foreach (string headerValue in headerValues)
             {
+                if (string.IsNullOrWhiteSpace(headerValue))
+                {
+                    continue;
+                }
+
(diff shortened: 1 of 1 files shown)


lucaspimentel (approved):
I'm not sure we should ignore this completely (it's an indication that there is an issue somewhere else), but it does seem better than spamming the logs.

andrewlock (approved):
Approved, but I agree with @lucaspimentel, this feels like it probably _should_ be a warning 🤔

kevingosse:
> Approved, but I agree with @lucaspimentel, this feels like it probably _should_ be a warning 🤔

I agree it's a gray area. But while a "wrong" value for the header is clearly a warning, I think we can reasonably expect the header to be empty in some scenarios.

lucaspimentel:
> I think we can reasonably expect the header to be empty in some scenarios.

If one of the tracers is injecting this header into http requests with no value, that's a bug. AFAIK this value is required for distributed tracing to work correctly.

PR #1641, opened 2021-07-29
Add better error handling for the Header Tags feature accessing System.Web.HttpResponse.Headers

zacharycmontoya:
Fixes `System.PlatformNotSupportedException: This operation requires IIS integrated pipeline mode`

This scenario can occur even when the IIS application is running in the Integrated pipeline mode, which seems to be related to Sitefinity CMS widgets. Without having a reproduction available, this doesn't prevent the bad access from happening. Instead, we have better handling of the scenario:
1. Only attempt to access the System.Web.HttpResponse.Headers field when the application has explicitly configured the `DD_TRACE_HEADER_TAGS`/`TracerSettings.HeaderTags` feature, which avoids the issue in some cases
2. Catch the exception and prevents further field accesses when the exception is observed

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace.AspNet/TracingHttpModule.cs
+++ b/src/Datadog.Trace.AspNet/TracingHttpModule.cs
@@ -28,6 +28,8 @@ namespace Datadog.Trace.AspNet
         private readonly string _httpContextScopeKey;
         private readonly string _requestOperationName;
 
+        private static bool CanReadHttpResponseHeaders = true;
+
         /// <summary>
         ///     Initializes a new instance of the <see cref="TracingHttpModule" /> class.
         /// </summary>
@@ -151,10 +153,7 @@ namespace Datadog.Trace.AspNet
                 {
                     try
                     {
-                        if (HttpRuntime.UsingIntegratedPipeline)
-                        {
-                            scope.Span.SetHeaderTags<IHeadersCollection>(app.Context.Response.Headers.Wrap(), Tracer.Instance.Settings.HeaderTags, defaultTagPrefix: SpanContextPropagator.HttpResponseHeadersTagPrefix);
-                        }
+                        AddHeaderTagsFromHttpResponse(app.Context, scope);
 
                         scope.Span.SetHttpStatusCode(app.Context.Response.StatusCode, isServer: true);
 
@@ -197,10 +196,7 @@ namespace Datadog.Trace.AspNet
 
                 if (httpContext.Items[_httpContextScopeKey] is Scope scope)
                 {
-                    if (HttpRuntime.UsingIntegratedPipeline)
-                    {
-                        scope.Span.SetHeaderTags<IHeadersCollection>(httpContext.Response.Headers.Wrap(), Tracer.Instance.Settings.HeaderTags, defaultTagPrefix: SpanContextPropagator.HttpResponseHeadersTagPrefix);
-                    }
+                    AddHeaderTagsFromHttpResponse(httpContext, scope);
 
                     if (exception != null && !is404)
                     {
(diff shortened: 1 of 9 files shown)


src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AspNet/ApiController_ExecuteAsync_Integration.cs
```diff
@@ -120,7 +120,7 @@ public class ApiController_ExecuteAsync_Integration
                 {
                     scope.Span.SetHeaderTags<IHeadersCollection>(httpContext.Response.Headers.Wrap(), Tracer.Instance.Settings.HeaderTags, defaultTagPrefix: SpanContextPropagator.HttpResponseHeadersTagPrefix);
                 }
-
+                HttpContextHelpers.AddHeaderTagsFromHttpResponse(httpContext, scope);
```
lucaspimentel, line 123: Shouldn't this new line replace the few lines above?

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/HttpContextHelpers.cs
```diff
@@ -0,0 +1,37 @@
+#if NETFRAMEWORK
+using System;
+using System.Web;
+using Datadog.Trace.ExtensionMethods;
+using Datadog.Trace.Headers;
+using Datadog.Trace.Logging;
+
+namespace Datadog.Trace.ClrProfiler.Integrations.AspNet
+{
+    internal static class HttpContextHelpers
```
lucaspimentel, line 10: The usual naming is singular: `HttpContextHelper`

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetWebApi2Integration.cs
```diff
@@ -41,6 +41,8 @@ public static class AspNetWebApi2Integration
         private static readonly IntegrationInfo IntegrationId = IntegrationRegistry.GetIntegrationInfo(nameof(IntegrationIds.AspNetWebApi2));
         private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(AspNetWebApi2Integration));
 
+        private static bool CanReadResponseHeaders = true;
```
lucaspimentel, line 44: I don't see this field being used anywhere.

src/Datadog.Trace.AspNet/TracingHttpModule.cs
```diff
@@ -225,5 +221,26 @@ private void TryClearContext(HttpContext context)
                 Log.Error(ex, "Error while clearing the HttpContext");
             }
         }
+
+        private static void AddHeaderTagsFromHttpResponse(System.Web.HttpContext httpContext, Scope scope)
+        {
```
lucaspimentel, line 226: Can't `TracingHttpModule` use `HttpContextHelpers.AddHeaderTagsFromHttpResponse()` instead of duplicating the code?

zacharycmontoya: I wanted to place the helper code in the instrumentation assembly since that's where it's mostly used. Unfortunately, the AspNet assembly does not reference that, it only references the base Datadog.Trace assembly, so I figured it could be duplicated for now.

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/AspNetWebApi2Integration.cs
zacharycmontoya: Good catch, this got refactored out. Fixed in f9debd5b8032b91f44e8d852ac9f8f7c475cec3a

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/AspNet/ApiController_ExecuteAsync_Integration.cs
zacharycmontoya: Yep you're right, fixed in f9debd5b8032b91f44e8d852ac9f8f7c475cec3a

src/Datadog.Trace.ClrProfiler.Managed/Integrations/AspNet/HttpContextHelpers.cs
zacharycmontoya: Sounds good, renamed to `HttpContextHelper` in 59123778f187fd5034b83c472b32b53278f5b1e4

lucaspimentel (approved):

src/Datadog.Trace.AspNet/TracingHttpModule.cs
andrewlock: Agreed, especially since this issue goes away once #1443 is merged

src/Datadog.Trace/ExtensionMethods/DictionaryExtensions.cs
```diff
@@ -79,5 +79,7 @@ public static bool TryGetValue<TValue>(this IDictionary dictionary, object key,
 
             return dictionary.Count == 0;
         }
+
+        public static bool IsNullOrEmpty<TKey, TValue>(this IDictionary<TKey, TValue> dictionary) => dictionary?.IsEmpty() ?? true;
```
andrewlock, line 83: Is it just me that hates the "negative" of `IsNullOrEmpty()`, especially as we always do `!string.IsNullOrEmpty()`?!

I wish `string.HasValue()` was a thing, so that this would be `HasValue()` 😄

andrewlock (approved):

src/Datadog.Trace/ExtensionMethods/DictionaryExtensions.cs
kevingosse: @andrewlock Just add an extension method with aggressive inlining if that makes you feel better 😄

kevingosse (approved):

PR #1651, opened 2021-08-02
Read performance counter values from environment variables in AAS

kevingosse:
In AAS, the IIS application pool process does not have sufficient permissions to read performance counters. However, the values are pushed [as environment variables](https://docs.microsoft.com/en-us/azure/app-service/reference-app-settings?tabs=kudu%2Cdotnet#performance-counters).

Unfortunately, "contention count" is missing.

### DIFF
--- a/src/Datadog.Trace/RuntimeMetrics/AzurePerformanceCountersListener.cs
+++ b/src/Datadog.Trace/RuntimeMetrics/AzurePerformanceCountersListener.cs
@@ -3,6 +3,8 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
 // </copyright>
 
+#if NETFRAMEWORK
+
 using System;
 using Datadog.Trace.Util;
 using Datadog.Trace.Vendors.Newtonsoft.Json;
@@ -79,3 +81,5 @@ namespace Datadog.Trace.RuntimeMetrics
         }
     }
 }
+
+#endif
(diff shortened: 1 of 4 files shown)


src/Datadog.Trace/RuntimeMetrics/AzurePerformanceCountersListener.cs
```diff
@@ -0,0 +1,81 @@
...
+// </copyright>
+
+using System;
+using Datadog.Trace.Util;
+using Datadog.Trace.Vendors.Newtonsoft.Json;
+using Datadog.Trace.Vendors.StatsdClient;
+
+namespace Datadog.Trace.RuntimeMetrics
+{
+    internal class AzurePerformanceCountersListener : IRuntimeMetricsListener
```
andrewlock, line 13: PerformanceCounters is .NET Framework only right? Should this be .NET Framework only too? I know we _currently_ don't support Linux in AAS anyway, but may be better to be explicit anyway.

Also, is there value in also extracting other aspnet releted metrics ([as we do in `RuntimeEventListener`](https://github.com/DataDog/dd-trace-dotnet/blob/master/src/Datadog.Trace/RuntimeMetrics/RuntimeEventListener.cs#L49-L58)) seeing as they're exposed here (i.e. use `WEBSITE_COUNTERS_ALL`?

andrewlock (commented):
Looks good, just a couple of questions

src/Datadog.Trace/RuntimeMetrics/AzurePerformanceCountersListener.cs
kevingosse: > Should this be .NET Framework only too?
Good point. Done in https://github.com/DataDog/dd-trace-dotnet/pull/1651/commits/dc217219603dc15b65cf4417e713c1a3de64e81e

> Also, is there value in also extracting other aspnet releted metrics
Currently we rely on the agent to collect aspnet metrics. Maybe it would be worth changing that, but it's beyond the scope of this PR.

andrewlock (approved):

src/Datadog.Trace/RuntimeMetrics/AzurePerformanceCountersListener.cs
colin-higgins: May want to name this `AzureAppServicePerformanceCounters` for less ambiguity

zacharycmontoya (approved):

PR #1660, opened 2021-08-04
Enable security only if initialization process OK

anna-git:
Fixes #

Enable security only if initialization process OK
Small fixes

@DataDog/apm-dotnet


src/Datadog.Trace/AppSec/Security.cs
```diff
@@ -56,12 +56,18 @@ private Security(InstrumentationGateway instrumentationGateway = null, IPowerWaf
...
+                        _agentWriter = agentWriter ?? new AgentWriter();
+                        _instrumentationGateway.InstrumentationGatewayEvent += InstrumentationGatewayInstrumentationGatewayEvent;
+                    }
+                    else
+                    {
+                        Enabled = false;
+                    }
                 }
             }
             catch (Exception ex)
```
robertpi, line 73: Should probably set `Enabled = false` in this catch block too.

robertpi (approved):
One small comment, otherwise it's good.

src/Datadog.Trace/AppSec/Security.cs
anna-git: good point!

PR #1663, opened 2021-08-05
Add Microsoft.Extensions.Logging.ILogger instrumentation

andrewlock:
This PR uses an alternative approach to add `ILogger` logs injection. The approach in #1549 was limited to using logs injection in ASP.NET Core requests, whereas this approach handles other cases (such as manual instrumentation) as well by instrumenting `LoggerExternalScopeProvider` and `LoggerFactoryScopeProvider`.

## Approach

When a logger iterates the scopes for a log message we insert ourselves as the first entry. Note that we don't _really_ add a scope (by instrumenting `BeginScope()` or similar) as that would insert our scope in the "scope stack", and make us responsible for removing it at the appropriate time. The approach here is much simpler, and works for all types of instrumentation that are using the standard ILogger infrastructure This essentially works the same way as the "activity" `ILogger` integration does in the runtime: 

https://github.com/dotnet/runtime/blob/cf258a14b70ad9069470a108f13765e0e5988f51/src/libraries/Microsoft.Extensions.Logging/src/LoggerFactoryScopeProvider.cs#L36-L52

Note that we're instrumenting `Microsoft.Extensions.Logging.*` assemblies, so [had to update the "ignore prefixes" section in _dd_profiler_constants.h _](https://github.com/DataDog/dd-trace-dotnet/commit/eb5d398c26295dc92b9ccd74bf04fd134a97e9e0) . 

## Limitations

If the app is using a custom `ILoggerFactory` implementation, or if the logging providers are using custom `IExternalScopeProvider` implementations, or otherwise not delegating to the insturmented types, then this obviously won't work. Serilog, for example, uses a custom ILoggerFactory (which doesn't matter generally, as we have direct Serilog integration, but it's worth noting).

## Testing

Added tests that logs injection works for aspnetcore requests and manual traces, based on the work in #1637. Note that we also get Log injection on .NET Framework, though as we don't have aspnetcore support, logs written outside of a manual trace are not log injected

> To remain independent from #1637, I copied [the logs injection helper](https://github.com/DataDog/dd-trace-dotnet/blob/a3528e0a575ff695be4554a3fdcc35aeb2d2aed6/test/Datadog.Trace.ClrProfiler.IntegrationTests/LogsInjectionTestBase.cs) from that PR.


src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/ILogger/DatadogLoggingScope.cs
```diff
@@ -0,0 +1,96 @@
...
+                span.TraceId);
+        }
+
+        public IEnumerator<KeyValuePair<string, object>> GetEnumerator()
+        {
+            var span = _tracer.ActiveScope?.Span;
+            if (span is null)
+            {
+                yield break;
+            }
```
andrewlock, line 83: I'm undecided whether we should still add the `dd_service`, `dd_env`, and `dd_version` tags if we don't have a trace. I'm inclined to say we should, but our other logs injection enrichers _don't_ do this currently, so I took this approach for consistency.

Does anyone else have opinions on which approach to take? Should we always add the `dd_service` etc tags?

andrewlock: Established that we _should_ be doing this, so implemented in https://github.com/DataDog/dd-trace-dotnet/pull/1663/commits/0ba8588b8feb0b6106ea2d1c1e9b90e786ccd32e and updated the tests in https://github.com/DataDog/dd-trace-dotnet/pull/1663/commits/893deff858a845a23d06df32b845b63c9e5681bb

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/ILogger/DatadogLoggingScope.cs
```diff
@@ -0,0 +1,91 @@
...
+                {
+                    return new KeyValuePair<string, object>("dd_env", _env);
+                }
+                else if (index == 2)
+                {
+                    return new KeyValuePair<string, object>("dd_version", _version);
+                }
+                else if (index == 3)
+                {
+                    return new KeyValuePair<string, object>("dd_trace_id", _tracer.ActiveScope?.Span.TraceId);
```
zacharycmontoya, line 53: I know we determined that the span id wasn't strictly necessary for the correlation, but I think we should add that here since we have the ability to do so now

test/Datadog.Trace.ClrProfiler.IntegrationTests/LogsInjectionTestBase.cs
```diff
@@ -0,0 +1,161 @@
...
+            /// <summary>
+            /// UnTraced logs include the dd_ properties, but without any values
+            /// </summary>
+            EmptyProperties,
+
+            /// <summary>
+            /// UnTraced logs include dd_service, dd_env, and dd_version with their correct values
+            /// but no dd_trace_id
+            /// </summary>
+            TWOLPropertiesOnly,
```
zacharycmontoya, line 51: nit: I thought this was "Unified Service Tagging", not "Tracing Without Limits"?

zacharycmontoya (commented):
This is great! I left a comment requesting span_id's now that we're capable of getting the span id on each call. Once that's figured out, this is ready to go 👍🏼

test/Datadog.Trace.ClrProfiler.IntegrationTests/LogsInjectionTestBase.cs
andrewlock: Oops, you're right. Updated the name

src/Datadog.Trace.ClrProfiler.Managed/AutoInstrumentation/ILogger/DatadogLoggingScope.cs
andrewlock: Yep, good plan, done in https://github.com/DataDog/dd-trace-dotnet/pull/1663/commits/55a06d422431af54be0b5e21e30e58620162b2a8

zacharycmontoya (approved):
LGTM. Good work!

src/Datadog.Trace/ClrProfiler/AutoInstrumentation/ILogger/DatadogLoggingScope.cs
```diff
@@ -0,0 +1,97 @@
...
+                    return new KeyValuePair<string, object>("dd_trace_id", _tracer.ActiveScope?.Span.TraceId);
+                }
+                else if (index == 4)
+                {
+                    return new KeyValuePair<string, object>("dd_span_id", _tracer.ActiveScope?.Span.SpanId);
+                }
+
+                throw new ArgumentOutOfRangeException(nameof(index));
+            }
+        }
```
kevingosse, line 62: NIT: should probably be a switch case. The compiler has special optimizations for consecutive numerical cases

kevingosse (approved):
LGTM

src/Datadog.Trace/ClrProfiler/AutoInstrumentation/ILogger/DatadogLoggingScope.cs
andrewlock: Done in https://github.com/DataDog/dd-trace-dotnet/pull/1663/commits/039f1a3f0085e6215bed7ecb1460019bf7f9e1ac

kevingosse (approved):

PR #1665, opened 2021-08-05
Cache native AssemblyReference instances

tonyredondo:
Currently the profiler Initialization method is creating `AssemblyReferences` for each item in the `integrations.json` and each `AssemblyReference` calls the `regex` library 3 times (`GetVersionFromAssemblyReferenceString`, `GetLocaleFromAssemblyReferenceString`, `GetPublicKeyFromAssemblyReferenceString`).

Because our integrations are contained in a single assembly, this PR uses a `AssemblyReference` cache as an optimization.

#### Initialization time in milliseconds in a sample application:

image

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace.ClrProfiler.Native/calltarget_tokens.cpp
+++ b/src/Datadog.Trace.ClrProfiler.Native/calltarget_tokens.cpp
@@ -136,7 +136,7 @@ HRESULT CallTargetTokens::EnsureBaseCalltargetTokens()
     // *** Ensure profiler assembly ref
     if (profilerAssemblyRef == mdAssemblyRefNil)
     {
-        const AssemblyReference assemblyReference = trace::AssemblyReference(managed_profiler_full_assembly_version);
+        const AssemblyReference assemblyReference = trace::AssemblyReference::GetFromCache(managed_profiler_full_assembly_version);
         ASSEMBLYMETADATA assembly_metadata{};
 
         assembly_metadata.usMajorVersion = assemblyReference.version.major;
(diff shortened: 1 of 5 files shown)


lucaspimentel (approved):
Nice! If you merge this first, I'll have to adopt this usage in #1443.

src/Datadog.Trace.ClrProfiler.Native/calltarget_tokens.cpp
```diff
@@ -136,7 +136,7 @@ HRESULT CallTargetTokens::EnsureBaseCalltargetTokens()
     // *** Ensure profiler assembly ref
     if (profilerAssemblyRef == mdAssemblyRefNil)
     {
-        const AssemblyReference assemblyReference = managed_profiler_full_assembly_version;
+        const AssemblyReference assemblyReference = trace::AssemblyReference(managed_profiler_full_assembly_version);
```
zacharycmontoya, line 139: Should this also get the cached AssemblyReference, like this?

```suggestion
        const AssemblyReference assemblyReference = trace::AssemblyReference::GetFromCache(managed_profiler_full_assembly_version);
```

zacharycmontoya (approved):

src/Datadog.Trace.ClrProfiler.Native/calltarget_tokens.cpp
tonyredondo: Oh!, you're right, I missed the `GetFromCache` part 🤦🏻

PR #1647, opened 2021-07-30
The v0 version of App Sec

robertpi:
This PR implements a first version of App Sec. Please forgive the message history, it will be squashed away :)

## Activation

The App Sec module introduces two new environment variables to activate App Sec features:

`DD_APPSEC_ENABLED` – enables the App Sec feature, any HTTP requests will be examined by the WAF (Web Application Firewall). By default, this feature is disabled.

`DD_APPSEC_BLOCKING_ENABLED` – this controls whether the App Sec feature will block suspicious requests. By default, this feature is disabled.

The app sec uses the ‘call target’ instrumentation to gather the data about the HTTP request that will be passed to the WAF. It is has not been implemented for ‘call site’ instrumentation and currently there is not plan to support this style of instrumentation.

## Implementation and Data Flow

Data is gathered in the instrumentations for ASP.NET and ASP.NET Core and then passed to the ‘instrumentation gateway’ which starts the security flow. Currently the instrumentation gateway simply creates and event with the incoming data. In future version this component will be used to route the data to different security components based on the addresses associated with the data, but this functionality is not necessary for the first version.

After the instrumentation gateway the data flow is as follows:

```
InstrumentationGateway -> Security -> PowerWaf -> Native -> Sqreen.dll
```

`Security` – this class coordinates the reaction to security events
`PowerWaf` – this class provides .NET idiomatic wrapper to the WAF, as the WAF is implemented in C++ with C style interface
`Native` – this class is responsible for finding, loading and providing low level bindings with the WAF
`[lib]Sqreen.[dll|so|dylib]` – this are the native binaries containing the WAF implementation.

Security events are reported to the agent and backend by `AgentWriter` class, which implements a basic queuing system.

## Testing

Two new samples are provided, that form the basis for the security integration tests. These are `Samples.AspNetMvc5` and `Samples.AspNetCore5` which are used to test ASP.NET and ASP.NET Core respectively.

Integration tests based on similar principals to `Datadog.Trace.ClrProfiler.IntegrationTests` have been provided in `Datadog.Trace.Security.IntegrationTests` and are run as part of the integration tests in the `consolidated-pipeline` CI.

The main focus of our work after this PR will to provide more testing so that we can confidently onboard the first private beta clients. The scope of the testing is laid out in this document: https://datadoghq.atlassian.net/wiki/spaces/APS/pages/2014610699/IN+REVIEW+Acceptance+Criterias

## Supported Environments

[...]

PR #1685, opened 2021-08-12
update integrations.json

lucaspimentel:
Regenerate `integrations.json` which got out of sync in #1647.
@robertpi and @anna-git please confirm since this is used for AppSec, I believe.

### DIFF
--- a/integrations.json
+++ b/integrations.json
@@ -5180,33 +5180,6 @@
     ]
   },
   {
-    "name": "AspNetCore",
-    "method_replacements": [
-      {
-        "caller": {},
-        "target": {
-          "assembly": "Microsoft.AspNetCore.Http",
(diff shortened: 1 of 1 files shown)


andrewlock (approved):
For reference, this integration is for 2.x+, and returns `void`. That was split into two integrations - a 3.x+ integration that returns `void`, and a 2.x-3 which returns `RequestDelegate`.

anna-git (approved):

anna-git:
Indeed there's one too many AspNetCore, I checked on the former appsec branch and it wasn't regenerated 🙇

colin-higgins:
This is handled in the 1.28.3-prerelease PR: https://github.com/DataDog/dd-trace-dotnet/pull/1688

PR #1686, opened 2021-08-13
Proxy calls to dl

robertpi:
Some Linux distros don't have a version of libdl.so without a version prefix. Call from manage code will fail to
find the library on these distros. Proxying via a C++ library should cure this, because the C libraries will
handle loading in this case.

Fixes #

@DataDog/apm-dotnet


tonyredondo (approved):
LGTM!

zacharycmontoya (approved):
LGTM, thanks!

PR #1687, opened 2021-08-13
Adds Execution time benchmarks

tonyredondo:
This PR adds the execution time benchmarks and dashboard.

image

@DataDog/apm-dotnet


zacharycmontoya (approved):

robertpi (approved):
Looks good, any particular reason for using these two samples?

tonyredondo:
> Looks good, any particular reason for using these two samples?

Thanks, those two samples doesn't require any external resource as the db ones, also I omit the AppDomain related ones because .NET Core. Other than that there's no particular reason.

PR #1690, opened 2021-08-18
Add minimal test applications that use service bus libraries

zacharycmontoya:
Adds a test application and smoke test for the following service bus libraries. This tests that automatic instrumentation does not encounter issues with the core components of the service bus libraries.
- MassTransit
- NServiceBus
- Rebus

@DataDog/apm-dotnet


zacharycmontoya:
Actually, I will be making additional changes to this so that they use SqlServer as a transport so that our automatic instrumentation does some amount of instrumentation during the application run...

PR #1692, opened 2021-08-20
New version of the WAF

robertpi:
The latest version of the C++ binary that analyses the incoming web requests. The configs have moved to a yaml, I have included a test rule, but there will be another later PR that will add yaml parsing and a real default rule set.

@DataDog/apm-dotnet


anna-git (approved):

src/Datadog.Trace/AppSec/Waf/Encoder.cs
```diff
@@ -167,5 +167,88 @@ private static Args CreateNativeString(string s)
             s = TrunacteLongString(s);
             return new Args(Native.pw_createStringWithLength(s, Convert.ToUInt64(s.Length)));
         }
+
+        public static string FormatArgs(object o)
+        {
+            var sb = new StringBuilder();
```
tonyredondo, line 173: Maybe you can use: https://github.com/DataDog/dd-trace-dotnet/blob/master/src/Datadog.Trace/Util/StringBuilderCache.cs to remove a couple of allocations.

tonyredondo (approved):
LGTM just a small allocation nit.

PR #1693, opened 2021-08-20
Remove CallTarget Integrations from json file.

tonyredondo:
This PR removes all CallTarget integrations from the `integrations.json` file and replace it by a PInvoke on initialization.

#### Limitations:
- Works only for CallTarget integrations because we use ReJIT for rewriting.
- ~This PR uses reflection to get all the CallTarget integrations, this must be changed to use Source Generators or similar and avoid the reflection costs.~

#### Performance:
Removing all the json parsing from CorProfiler::Initialize callback allows to remove almost 90% of initialization cost. ~But because we are using reflection later on when the loader initializes the managed side this improvement vanishes, although this can be improved removing the reflection calls and use source generators or similar.~ A generator is implemented.

@DataDog/apm-dotnet


kevingosse:
> But because we are using reflection later on when the loader initializes the managed side this improvement vanishes, although this can be improved removing the reflection calls and use source generators or similar.

Source generators would be great, but as a shorter-term solution we can update the `PrepareRelease` tool to generate a .cs file instead of a .json (basically, making our own source generator).

andrewlock:
> but as a shorter-term solution we can update the PrepareRelease tool to generate a .cs file instead of a .json (basically, making our own source generator).

Great idea. RE source generators, we need to remove .NET 4.5 support before we can use them (as they can only be used in projects that support netstandard 2.1)

tonyredondo:
>Source generators would be great, but as a shorter-term solution we can update the PrepareRelease tool to generate a .cs file instead of a .json (basically, making our own source generator).

Done.

build/_build/PrepareRelease/GenerateIntegrationDefinitions.cs
```diff
@@ -73,18 +66,77 @@ public static void Run(ICollection<string> assemblyPaths, params string[] output
...
+                Console.WriteLine($"Writing {calltargetPath}...");
+                File.WriteAllText(calltargetPath, callTargetFileContent, utf8NoBom);
             }
         }
 
-        static IEnumerable<Integration> GetCallTargetIntegrations(ICollection<Assembly> assemblies)
+        static string CreateCallTargetDefinitionFileContent(IEnumerable<CallTargetDefinitionSource> callTargetIntegrations)
+        {
+            var cTargetIntegrations = callTargetIntegrations.Distinct().ToList();
+            var sBuilder = new StringBuilder();
```
lucaspimentel, line 88: Is it worth changing this to write directly to a (file) stream instead of building the entire string in memory first? I would request the change if this was production code, but it's just a CI tool, so level of care low.

tonyredondo: Sure, I like the idea, will do the change.

src/Datadog.Trace/ClrProfiler/Instrumentation.cs
```diff
@@ -66,19 +66,25 @@ public static void Initialize()
                 return;
             }
 
+            Log.Debug("Initialization started.");
+
             try
             {
-                // ensure global instance is created if it's not already
-                _ = Tracer.Instance;
+                Log.Debug("Sending CallTarget integrations definitions to native profiler.");
```
lucaspimentel, line 73: Let's try to not use "profiler" to avoid confusion with the Continuous Profiler product. We can say the tracer has a managed library and a native library. Or maybe "component" instead of "library"?
```suggestion
                Log.Debug("Sending CallTarget integrations definitions to native library.");
```

src/Datadog.Trace/ClrProfiler/Instrumentation.cs
```diff
@@ -66,19 +66,25 @@ public static void Initialize()
...
+                NativeMethods.InitializeProfiler(InstrumentationDefinitions.GetAllDefinitions());
+                Log.Debug("CallTarget integrations definitions sent.");
```
lucaspimentel, line 75:
```suggestion
                Log.Debug("CallTarget integration definitions sent.");
```

src/Datadog.Trace/ClrProfiler/InstrumentationDefinitions.cs
```diff
@@ -0,0 +1,234 @@
...
+// </copyright>
+
+namespace Datadog.Trace.ClrProfiler
+{
+    internal static class InstrumentationDefinitions
+    {
+        internal static NativeCallTargetDefinition[] GetAllDefinitions()
+        {
+            var defsArray = new NativeCallTargetDefinition[218];
+            defsArray[0] = new NativeCallTargetDefinition("AerospikeClient", "Aerospike.Client.AsyncCommand", "ExecuteCommand",  new string[] { "System.Void" }, 4, 0, 0, 4, 65535, 65535, "Datadog.Trace, Version=1.28.6.0, Culture=neutral, PublicKeyToken=def86d061d0d2eeb", "Datadog.Trace.ClrProfiler.AutoInstrumentation.Aerospike.AsyncCommandIntegration");
```
lucaspimentel, line 13: Non-blocking, but each line here can be 
```
defsArray[0] = new(....);
```

or we could go further and use
```
var defsArray = new NativeCallTargetDefinition
{
    new(....),
    new(....),
    new(....),
    // ...
};

File: build/_build/PrepareRelease/GenerateIntegrationDefinitions.cs [hunk 5]
```diff
@@ -73,18 +66,77 @@ public static void Run(ICollection<string> assemblyPaths, params string[] output
...
+            sBuilder.AppendLine("namespace Datadog.Trace.ClrProfiler");
+            sBuilder.AppendLine("{");
+            sBuilder.AppendLine("\tinternal static class InstrumentationDefinitions");
+            sBuilder.AppendLine("\t{");
+            sBuilder.AppendLine("\t\tinternal static NativeCallTargetDefinition[] GetAllDefinitions()");
+            sBuilder.AppendLine("\t\t{");
+            sBuilder.AppendLine($"\t\t\tvar defsArray = new NativeCallTargetDefinition[{cTargetIntegrations.Count}];");
+            for (var i = 0; i < cTargetIntegrations.Count; i++)
+            {
+                sBuilder.Append($"\t\t\tdefsArray[{i}] = new NativeCallTargetDefinition(");
```
lucaspimentel, line 102:
```suggestion
                sBuilder.Append($"\t\t\tdefsArray[{i}] = new(");
```

build/_build/PrepareRelease/GenerateIntegrationDefinitions.cs
```diff
@@ -73,18 +66,77 @@ public static void Run(ICollection<string> assemblyPaths, params string[] output
...
+                sBuilder.Append($"\"{cTargetIntegrations[i].TargetAssembly}\", ");
+                sBuilder.Append($"\"{cTargetIntegrations[i].TargetType}\", ");
+                sBuilder.Append($"\"{cTargetIntegrations[i].TargetMethod}\", ");
+
+                sBuilder.Append($" new string[] {{ ");
```
lucaspimentel, line 107:
```suggestion
                sBuilder.Append($" new[] {{ ");
```

src/Datadog.Trace/ClrProfiler/NativeMethods.cs
```diff
@@ -23,19 +23,37 @@ public static bool IsProfilerAttached()
...
+
         // the "dll" extension is required on .NET Framework
         // and optional on .NET Core
         private static class Windows
         {
             [DllImport("Datadog.Trace.ClrProfiler.Native.dll")]
             public static extern bool IsProfilerAttached();
+
+            [DllImport("Datadog.Trace.ClrProfiler.Native.dll")]
+            public static extern void InitializeProfiler([In, Out] NativeCallTargetDefinition[] methodArrays, int size);
```
lucaspimentel, line 46: Why do we need `[Out]` here? Isn't `[In]` enough? (same comment below)

build/_build/PrepareRelease/GenerateIntegrationDefinitions.cs
lucaspimentel: (don't bother changing the call site json to do this, it will go away soon)

tonyredondo: done in 73f5a81

src/Datadog.Trace/ClrProfiler/Instrumentation.cs
tonyredondo: done in 73f5a81

src/Datadog.Trace/ClrProfiler/Instrumentation.cs
tonyredondo: done in 73f5a81

src/Datadog.Trace/ClrProfiler/InstrumentationDefinitions.cs
tonyredondo: done in 73f5a81

build/_build/PrepareRelease/GenerateIntegrationDefinitions.cs
tonyredondo: done in 73f5a81

build/_build/PrepareRelease/GenerateIntegrationDefinitions.cs
tonyredondo: done in 73f5a81

src/Datadog.Trace/ClrProfiler/NativeMethods.cs
tonyredondo: You're right, done in 73f5a81

lucaspimentel (approved):
Nice!

src/Datadog.Trace/ClrProfiler/InstrumentationDefinitions.cs
```diff
@@ -0,0 +1,235 @@
...
+
+namespace Datadog.Trace.ClrProfiler
+{
+    internal static class InstrumentationDefinitions
+    {
+        internal static NativeCallTargetDefinition[] GetAllDefinitions()
+        {
+            return new NativeCallTargetDefinition[]
+            {
+                new("AerospikeClient", "Aerospike.Client.AsyncCommand", "ExecuteCommand",  new[] { "System.Void" }, 4, 0, 0, 4, 65535, 65535, "Datadog.Trace, Version=1.28.6.0, Culture=neutral, PublicKeyToken=def86d061d0d2eeb", "Datadog.Trace.ClrProfiler.AutoInstrumentation.Aerospike.AsyncCommandIntegration"),
```
zacharycmontoya, line 14: I think I know the answer, but figured I'd ask anyways. I noticed that there are no longer integration names for each integration. Is that because the instrumentation methods are supposed to be checking whether the integration is enabled or not?

tonyredondo: Yes, right now we are doing the check in the managed side (eg. https://github.com/DataDog/dd-trace-dotnet/blob/master/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/SQS/AwsSqsCommon.cs#L27). But now you mention this, maybe we can improve it and move it to this instrumentation definitions file by filtering the array before returning (or avoiding the instance generation), so we avoid the rewriting completely. That will reduce memory (ducktyping and dynamic methods to run the integration).  But maybe that's outside this PR? WDYT?

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -1101,6 +1085,104 @@ HRESULT STDMETHODCALLTYPE CorProfiler::JITInlining(FunctionID callerId, Function
     return S_OK;
 }
 
+
+//
+// InitializeProfiler method
+//
+void CorProfiler::InitializeProfiler(CallTargetDefinition* items, int size)
```
zacharycmontoya, line 1092: nit: If this is just going to initialize the CallTarget methods, maybe this can be renamed to `InitializeCallTargetIntegration` or something similar

src/Datadog.Trace/ClrProfiler/InstrumentationDefinitions.cs
zacharycmontoya: Yeah I think that is out of scope for this PR. However we decide to implement it, I think it would be helpful to have a built-in mechanism to filter out integrations so developers don't need to manually write the check `if (tracer.Settings.IsIntegrationEnabled())`

src/Datadog.Trace/ClrProfiler/Instrumentation.cs
```diff
@@ -66,19 +66,25 @@ public static void Initialize()
...
             }
 
+            Log.Debug("Initialization started.");
+
             try
             {
-                // ensure global instance is created if it's not already
-                _ = Tracer.Instance;
+                Log.Debug("Sending CallTarget integration definitions to native library.");
+                NativeMethods.InitializeProfiler(InstrumentationDefinitions.GetAllDefinitions());
```
zacharycmontoya, line 74: Should we have this method free the native memory that was allocated for each `NativeCallTargetDefinition`?

zacharycmontoya (approved):
Awesome work!! LGTM

src/Datadog.Trace/ClrProfiler/Instrumentation.cs
tonyredondo: done in c72171a

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
andrewlock: I guess soon (v2.0) there will only _be_ call target, so that may not be necessary?

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.h
```diff
@@ -142,6 +142,11 @@ class CorProfiler : public CorProfilerBase
     //
     HRESULT STDMETHODCALLTYPE GetAssemblyReferences(const WCHAR* wszAssemblyPath,
                                                     ICorProfilerAssemblyReferenceProvider* pAsmRefProvider) override;
+
+    //
+    // Add Integrations methods
+    //
+    void InitializeProfiler(CallTargetDefinition* items, int size);
```
andrewlock, line 149: The comment implies this should probably be called `AddIntegrationsMethods()` 😉

src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AspNet/ApiController_ExecuteAsync_Integration.cs
```diff
@@ -3,21 +3,20 @@
...
-using Datadog.Trace.ClrProfiler.Emit;
 using Datadog.Trace.ClrProfiler.Integrations;
+#if NETFRAMEWORK
 using Datadog.Trace.ClrProfiler.Integrations.AspNet;
+#endif
 using Datadog.Trace.Configuration;
 using Datadog.Trace.DuckTyping;
 using Datadog.Trace.ExtensionMethods;
-using Datadog.Trace.Headers;
 using Datadog.Trace.Tagging;
```
andrewlock, line 20: I assume this change is related to generated the integrations.json/cs file? I don't think this is necessary, as we look through all of the dlls for the different platforms and do a `Distinct()`? (I prefer the #if def over the whole file, makes more sense I think)

andrewlock (commented):
🎉

src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AspNet/ApiController_ExecuteAsync_Integration.cs
tonyredondo: I think I can rollback this, this was done because before I was extracting the integrations using reflection, and depending on the Tracer dll loaded I was having different numbers of integrations.... Not a problem anymore.

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
tonyredondo: I wanted to have this generic name, because I plan to do some more experimentation on passing more info between both unmanaged/managed world, and I don't want to pay the PInvoke performance penalty multiple times.

src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AspNet/ApiController_ExecuteAsync_Integration.cs
tonyredondo: done in 7dcf857

PR #1695, opened 2021-08-20
Sending more relevant data to backend

anna-git:
Fixes #

Removing cookie value from waf address header no cookie

Sending more relevant data to backend

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace/AppSec/Transports/Http/HttpTransport.Framework.cs
+++ b/src/Datadog.Trace/AppSec/Transports/Http/HttpTransport.Framework.cs
@@ -43,14 +43,14 @@ namespace Datadog.Trace.AppSec.Transport.Http
             return context.Items[WafKey] as IAdditiveContext;
         }
 
-        public Request Request()
+        public Request Request() => new()
         {
-            return new Request
-            {
-                Url = context.Request.Url,
-                Method = context.Request.HttpMethod,
-            };
-        }
+            Url = context.Request.Url,
+            Method = context.Request.HttpMethod,
+            Scheme = context.Request.Url.Scheme,
+            RemoteIp = context.Request.ServerVariables["HTTP_X_FORWARDED_FOR"] ?? context.Request.ServerVariables["REMOTE_ADDR"],
+            Host = context.Request.UserHostAddress,
+        };
 
         public Response Response(bool blocked) => new()
         {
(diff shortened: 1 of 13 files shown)


src/Datadog.Trace/Util/Http/HttpRequestExtensions.Core.cs
```diff
@@ -19,9 +19,9 @@ internal static partial class HttpRequestExtensions
         internal static Dictionary<string, object> PrepareArgsForWaf(this HttpRequest request, RouteData routeDatas = null)
         {
             var url = GetUrl(request);
-
             var headersDic = new Dictionary<string, string>();
-            foreach (var k in request.Headers.Keys)
+            var headerKeys = request.Headers.Keys.Where(k => k != "Cookie");
```
andrewlock, line 23: Don't know if you want to use LINQ here, and instead move to be an `if` inside the loop (as you do in the other implementation for FX?). Also, probably want to make the comparison case insensitive?

andrewlock (approved):

src/Datadog.Trace/AppSec/Transports/Http/HttpTransport.Framework.cs
```diff
@@ -43,14 +43,14 @@ public IAdditiveContext GetAdditiveContext()
...
-            return new Request
-            {
-                Url = context.Request.Url,
-                Method = context.Request.HttpMethod,
-            };
-        }
+            Url = context.Request.Url,
+            Method = context.Request.HttpMethod,
+            Scheme = context.Request.Url.Scheme,
+            RemoteIp = context.Request.ServerVariables["HTTP_X_FORWARDED_FOR"] ?? context.Request.ServerVariables["REMOTE_ADDR"],
```
robertpi, line 51: How did we decide what's the right value, between `request.RemoteIp`, `context.Request.ServerVariables["HTTP_X_FORWARDED_FOR"]`, and `context.Request.ServerVariables["REMOTE_ADDR"]`? I think what you have is okay for now, but I think there's going to be an official algorithm to follow about how we should collect the IP address, so we may have to change this at somepoint.

src/Datadog.Trace/Util/Http/HttpRequestExtensions.Core.cs
anna-git: thanks, done in 4b5c2315584c64409bd566bde67b81b8489e2807

src/Datadog.Trace/AppSec/Transports/Http/HttpTransport.Framework.cs
anna-git: yes! let's confirm with the team how we get the remote ip later

PR #1715, opened 2021-08-30
Fix docker-compose command line for arm64 integration tests

zacharycmontoya:
In the integration_tests_arm64 stage, replace COMPOSE_PROJECT_NAME with the projectName input on the DockerCompose Azure DevOps task. This should correctly set the project name because the task overrides the environment variable with the -p option.

This fixes the issue where the arm64 integration test run cannot stop the docker-compose containers because parallel runs use the same "project name", resulting in re-using the same container names.

@DataDog/apm-dotnet

### DIFF
--- a/.azure-pipelines/ultimate-pipeline.yml
+++ b/.azure-pipelines/ultimate-pipeline.yml
@@ -603,7 +603,6 @@ stages:
         TestAllPackageVersions: true
         publishTargetFramework: net5.0
         baseImage: debian
-        COMPOSE_PROJECT_NAME: ddtrace_$(Build.BuildNumber)
 
       pool:
         name: Arm64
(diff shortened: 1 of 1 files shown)


lucaspimentel (approved):

andrewlock (commented):
Doh, I introduced this specifically to _fix_ that issue. I don't understand why it didn't work though? Parallel runs should have a different `Build.BuildNumber` shouldn't they? And doesn't this fix suffer from the same "problem" (whatever that problem is)?

zacharycmontoya:
> Doh, I introduced this specifically to _fix_ that issue. I don't understand why it didn't work though? Parallel runs should have a different `Build.BuildNumber` shouldn't they? And doesn't this fix suffer from the same "problem" (whatever that problem is)?

This fixes the underlying problem because Azure DevOps sets a default in the DockerCompose@0 task. So their default was overriding the environment variable you set

andrewlock (approved):

andrewlock:
> This fixes the underlying problem because Azure DevOps sets a default in the DockerCompose@0 task. So their default was overriding the environment variable you set

oh, I see 😞 thanks!

PR #1716, opened 2021-08-31
Add tests for ASP.NET Core on .NET 5.0 and .NET Framework 4.6.1

andrewlock:
This ports the additional ASP.NET Core tests that were originally part of #1549 (which was closed without merging).

* Adds explicit ASP.NET Core 5.0 tests (identical to the 3.1 tests in practice, so just reused that app)
* Adds tests for .NET Framework on ASP.NET Core (currently confirms we _don't_ see any traces)

I trimmed down the test cases considerably given how similar netcoreapp31 and net5 are (calltarget only, feature flag enabled only), so as to not increase testing time to much.

Don't know if we actually want to include these or not, not passionate about it either way, as doesn't add a huge amount, just closes a small gap

@DataDog/apm-dotnet

### DIFF
--- a/tracer/build/_build/Build.Steps.cs
+++ b/tracer/build/_build/Build.Steps.cs
@@ -1098,7 +1098,7 @@ partial class Build
                     {
                         "Samples.AspNetCoreMvc21" => Framework == TargetFramework.NETCOREAPP2_1,
                         "Samples.AspNetCoreMvc30" => Framework == TargetFramework.NETCOREAPP3_0,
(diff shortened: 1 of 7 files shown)


zacharycmontoya (approved):

tonyredondo (approved):

zacharycmontoya:
@andrewlock should we keep this PR or close it?

andrewlock:
Closing in preference of #2398

PR #1717, opened 2021-08-31
Add support for Aerospike

kevingosse:
Enables calltarget instrumentation of the Aerospike.Client nuget package in versions `4.*.*`.

span name: aerospike.command
span type: aerospike
span kind: client
resource name: read/write/delete/...
tags:
 - aerospike.key
 - aerospike.setname
 - aerospike.namespace


src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Aerospike/AerospikeCommon.cs
```diff
@@ -0,0 +1,85 @@
...
+            Scope scope = null;
+
+            try
+            {
+                tags = new AerospikeTags();
+                var serviceName = tracer.Settings.GetServiceName(tracer, ServiceName);
+                scope = tracer.StartActiveWithTags(OperationName, tags: tags, serviceName: serviceName);
+                var span = scope.Span;
+
+                if (target.TryGetFieldValue<object>("key", out var key) && key != null)
```
tonyredondo, line 39: Aren't we moving from this method to DuckTyping?

kevingosse: Here I'm just testing if the "key" field is defined. It will be defined on some types and not on others. I find the ducktyping API a bit cumbersome for those scenarios 🤔

tonyredondo: 🤔   we have the `TryDuckCast` method but fair enough your point.

kevingosse: Done in https://github.com/DataDog/dd-trace-dotnet/pull/1717/commits/5c7738a76002e45f9a89d3db6c793336ef3a6d17, but I wonder if we could build a helper to avoid having to define a new type when only interested in one single field.

src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Aerospike/AerospikeCommon.cs
```diff
@@ -0,0 +1,86 @@
...
+            try
+            {
+                tags = new AerospikeTags();
+                var serviceName = tracer.Settings.GetServiceName(tracer, ServiceName);
+                scope = tracer.StartActiveWithTags(OperationName, tags: tags, serviceName: serviceName);
+                var span = scope.Span;
+
+                if (target.TryDuckCast<HasKey>(out var key) && key.Key != null)
+                {
+                    tags.Key = key.Key.ToString();
```
lucaspimentel, line 42: Are there other tags that could be useful for users like the server name or address? Are there command types like insert, update, delete?

lucaspimentel: I just saw `Write`, `Read`, `Exists`, and `Delete` in the test app.)

kevingosse: That's hard to tell since I'm not familiar with aerospike 😄 The client talks to a cluster and I don't think I can extract what server in the cluster has accepted the request 🤔

lucaspimentel: Discussed offline. Looks like it's not practical for now to iterate through the list of nodes to build some time of "remote host" information. We can come back and add it later if needed.

lucaspimentel (approved):

zacharycmontoya (approved):
LGTM

tonyredondo (approved):
LGTM

test/Datadog.Trace.ClrProfiler.IntegrationTests/AerospikeTests.cs
```diff
@@ -0,0 +1,77 @@
...
+                    Assert.Equal("aerospike.command", span.Name);
+                    Assert.Equal("Samples.Aerospike-aerospike", span.Service);
+                    Assert.Equal("aerospike", span.Type);
+                    Assert.Equal(SpanKinds.Client, span.Tags[Tags.SpanKind]);
+                    Assert.Equal("test:myset:mykey:56cfdb28a0a21ba119f76e0ea4e528a1f406dd94", span.Tags[Tags.AerospikeKey]);
+
+                    Assert.True(expectedSpans.Remove(span.Resource), $"Unexpected span resource: {span.Resource}");
+                }
+
+                Assert.Empty(expectedSpans);
```
andrewlock, line 73: It's just personal preference, but I'd prefer to write this as the following, as it gives you the whole context in case of failures

```suggestion
                spans.Should()
                     .OnlyContain(span => span.Name == "aerospike.command")
                     .And.OnlyContain(span => span.Service == "Samples.Aerospike-aerospike")
                     .And.OnlyContain(span => span.Tags[Tags.SpanKind] == SpanKinds.Client)
                     .And.OnlyContain(span => span.Tags[Tags.AerospikeKey] == "test:myset:mykey:56cfdb28a0a21ba119f76e0ea4e528a1f406dd94");
                spans.Select(span => span.Resource).Should().ContainInOrder(expectedSpans);
```

src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Aerospike/AerospikeCommon.cs
```diff
@@ -0,0 +1,86 @@
...
+namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Aerospike
+{
+    internal class AerospikeCommon
+    {
+        private const string ServiceName = "aerospike";
+        private const string OperationName = "aerospike.command";
+        public const string IntegrationName = nameof(IntegrationIds.Aerospike);
+        internal static readonly IntegrationInfo IntegrationId = IntegrationRegistry.GetIntegrationInfo(IntegrationName);
+
+        public static Scope CreateScope<TTarget>(Tracer tracer, out AerospikeTags tags, TTarget target)
```
andrewlock, line 21: AFAICT, the `out tags` are never used - does it need to be an `out` parameter?

andrewlock (approved):
Nice taming of the API 😉

src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Aerospike/AerospikeCommon.cs
kevingosse: - [ ] You're right. It bugged me during the night, then I forgot again in the morning 😄

test/Datadog.Trace.ClrProfiler.IntegrationTests/AerospikeTests.cs
kevingosse: Done in https://github.com/DataDog/dd-trace-dotnet/pull/1717/commits/465829a96c37b48ff3d2c7faf0f1e4c3ae96356a

PR #1719, opened 2021-08-31
Additional fix for ARM64 docker-compose services in CI

zacharycmontoya:
The previous PR (https://github.com/DataDog/dd-trace-dotnet/pull/1715) succeeded in letting each integration test run create its own set of docker-compose services. However, those services all tried to bind to the host via `127.0.0.1:<service-port>`. This resulted in the first set of services starting up fine, but the second set of services would fail to start because they could not bind to `127.0.0.1:<service-port>`, which was still occupied by the first set of services.

To fix this for ARM64 services specifically, I propose duplicating only the services needed by the ARM64 integration tests and only exposing the ports to linked services, not the hosts.

Below, you can see two builds from this branch that successfully ran in parallel with this change:
1. https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=73649&view=results
2. https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=73650&view=results

@DataDog/apm-dotnet

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,72 @@
 version: '3'
 services:
+  # ARM64 dependencies
+  aws_sqs_arm64:
+    image: softwaremill/elasticmq
+    ports:
+    - "9324"
+
+  elasticsearch7_arm64:
+    image: elasticsearch:7.10.1
+    ports:
+    - "9200"
+    - "9300"
+    environment:
+    - discovery.type=single-node
+
+  mongo_arm64:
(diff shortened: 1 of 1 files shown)


andrewlock (approved):

lucaspimentel (approved):

zacharycmontoya:
We can decide later if we want this to be the long-term fix, but this should fix the immediate issue so I'll go ahead and merge. Thanks for the reviews!

PR #1723, opened 2021-09-01
Native memory improvements

tonyredondo:
This PR improves the native memory consumption in the profiler by avoiding unnecessary copies and process.

#### Before: 10.63MB / 12.50MB / 84.07MB
image

#### After: 0.91MB / 4.15MB / 18.54MB
image

#### Before: 6.64MB / 8.40MB / 1386.55MB
image

#### After: 1.03MB / 4.15MB / 19.96MB
image

@DataDog/apm-dotnet


src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -629,81 +631,140 @@ HRESULT STDMETHODCALLTYPE CorProfiler::ModuleLoadFinished(ModuleID module_id, HR
...
+        const auto assembly_import = metadata_interfaces.As<IMetaDataAssemblyImport>(IID_IMetaDataAssemblyImport);
+        const auto assembly_emit = metadata_interfaces.As<IMetaDataAssemblyEmit>(IID_IMetaDataAssemblyEmit);
 
-    // store module info for later lookup
-    module_id_to_info_map_[module_id] = module_metadata;
+        // don't skip Microsoft.AspNetCore.Hosting so we can run the startup hook and
+        // subscribe to DiagnosticSource events.
+        // don't skip Dapper: it makes ADO.NET calls even though it doesn't reference
+        // System.Data or System.Data.Common
+        if (module_info.assembly.name != WStr("Microsoft.AspNetCore.Hosting") && module_info.assembly.name != WStr("Dapper"))
```
zacharycmontoya, line 729: I think these strings could also be replaced with constants so they're not allocated on every module load

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -3212,7 +3273,6 @@ size_t CorProfiler::CallTarget_RequestRejitForModule(ModuleID module_id, ModuleM
     if (!vtMethodDefs.empty())
     {
         this->rejit_handler->EnqueueForRejit(vtModules, vtMethodDefs);
-        this->rejit_handler->RequestRejitForNGenInliners();
```
zacharycmontoya, line 3215: What was the issue with RequestRejitForNGenInliners being called here?

src/Datadog.Trace.ClrProfiler.Native/module_metadata.h
```diff
@@ -45,15 +48,56 @@ class ModuleMetadata
...
+                   const WSTRING& assembly_name, const AppDomainID app_domain_id, const GUID module_version_id,
+                   const std::vector<IntegrationMethod> integrations, const AssemblyProperty* corAssemblyProperty) :
+        metadata_import(metadata_import),
+        metadata_emit(metadata_emit),
+        assembly_import(assembly_import),
+        assembly_emit(assembly_emit),
+        assemblyName(assembly_name),
+        app_domain_id(app_domain_id),
+        module_version_id(module_version_id),
+        integrations(new std::vector<trace::IntegrationMethod>(integrations)),
```
zacharycmontoya, line 66: I don't understand why you need to allocate a new object on this line, can you explain? From reading the other code, the `integrations` argument that's passed into the constructor is a new vector produced by `FilterIntegrationsByTarget` so it seems like it's not shared

tonyredondo: I do this because now we store a pointer in the module_metadata class, if the `FilterIntegrationsByTarget` creates the vector in the stack, the pointer will be invalid once the ModuleFinished callback returns. So to avoid this I force the creation in the heap and store that pointer to the heap. I have to check if the vector behave different and is already in the heap, but that's the reason is done this way.

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
tonyredondo: Oh right, that modification belongs to the `remove the json` PR, and the reason for that is we need to call the RequestReJIT on proper threads or callbacks, for the ModuleFinished callback is ok to call it directly, but if we do a PInvoke (like in the remove the json PR) then a direct call to RequestReJIT will fail. I'll revert this change on this PR.

src/Datadog.Trace.ClrProfiler.Native/module_metadata.h
gleocadie: Same as Zach.
If you need a new instance, why using new ? why not using `std::unique_ptr`. It would simplify the dtor (no dtor at all)

src/Datadog.Trace.ClrProfiler.Native/module_metadata.h
```diff
@@ -89,23 +143,41 @@ class ModuleMetadata
 
     void SetWrapperMemberRef(const WSTRING& keyIn, const mdMemberRef valueIn)
     {
-        wrapper_refs[keyIn] = valueIn;
+        std::lock_guard<std::mutex> guard(wrapper_mutex);
+        if (wrapper_refs == nullptr)
+        {
+            wrapper_refs = new std::unordered_map<WSTRING, mdMemberRef>();
```
gleocadie, line 149: same here, why `new` ?

gleocadie (commented):
I'm back with `std::unique_ptr` :P

src/Datadog.Trace.ClrProfiler.Native/module_metadata.h
tonyredondo: I didn't use `std::unique_ptr` in this case because the 1module metadata1 class behave differently depending if is CallTarget or CallSite. 

In the CallTarget case I want to use the already heap allocated integrations vector loaded in the Initialization without a copy and without calling the dtor or deleting the allocation, so in that case I just want a copy of the pointer. 

Meanwhile in CallSite we first filter the global integrations vector and return a new vector scoped to that module so... in that case I want to copy the vector to the heap and also call the dtor on finalization, because no other module we have the same copy.

src/Datadog.Trace.ClrProfiler.Native/module_metadata.h
tonyredondo: In this case yes, we can use `std::unique_ptr` the behavior is the same for both integration methods (CallTarget and CallSite), will do the change ... Although Is not that bad to use `new` here isn't?  XD for modules with any rewriting I will allocate 0 bytes, not true if I'm using smart pointers :D

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
tonyredondo: done in 59ae73f

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
tonyredondo: done in 988f718

src/Datadog.Trace.ClrProfiler.Native/module_metadata.h
tonyredondo: done in 79804dd

src/Datadog.Trace.ClrProfiler.Native/module_metadata.h
tonyredondo: I made changes to remove the different behaviours of the module_metadata class in 79804dd, that allowed me to replace the pointers to `unique_ptr`

src/Datadog.Trace.ClrProfiler.Native/module_metadata.h
gleocadie: XD no it's not that bad. I mean, it's just that if it's shared at some point, or we do not handle the lifecycle, it can become a nightmare. I know I can be extreme on that (sorry)
Yeah, this is my thinking:
1. can we use smart pointers first
2. If it's better (for perf reason ...) to use raw pointer, so be it

tonyredondo: no worries :) I've tested the change with `unique_ptr` and we are ok.

zacharycmontoya (approved):
LGTM and thanks for addressing the feedback

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -551,15 +558,18 @@ HRESULT STDMETHODCALLTYPE CorProfiler::ModuleLoadFinished(ModuleID module_id, HR
     if (Logger::IsDebugEnabled())
     {
         Logger::Debug("ModuleLoadFinished: ", module_id, " ", module_info.assembly.name, " AppDomain ",
-                      module_info.assembly.app_domain_id, " ", module_info.assembly.app_domain_name, " | IsNGEN = ", (module_info.IsNGEN() ? "true" : "false"));
+                      module_info.assembly.app_domain_id, " ", module_info.assembly.app_domain_name,
+                      " | IsNGEN = ", (module_info.IsNGEN() ? "true" : "false"),
+                      " | IsDynamic = ", (module_info.IsDynamic() ? "true" : "false"),
+                      " | IsResource = ", (module_info.IsResource() ? "true" : "false"));
     }
```
gleocadie, line 565: FYI, if `IsNGEN` and others are `bool` ,you can use `std::boolalpha` to express that the variable are boolean and the ouput should output true or false and `std::noboolalpha` to express that the next one are not to be considered as `bool`
https://en.cppreference.com/w/cpp/io/manip/boolalpha

src/Datadog.Trace.ClrProfiler.Native/module_metadata.h
```diff
@@ -89,23 +121,46 @@ class ModuleMetadata
 
     void SetWrapperMemberRef(const WSTRING& keyIn, const mdMemberRef valueIn)
     {
-        wrapper_refs[keyIn] = valueIn;
+        std::lock_guard<std::mutex> guard(wrapper_mutex);
```
gleocadie, line 124: FYI, since C++17, `lock_guard` should be replaced by `scoped_lock` (https://en.cppreference.com/w/cpp/thread/scoped_lock)

gleocadie (approved):
LGTM
comments are just tips ;)

src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
tonyredondo: TIL, done in f18e9da

src/Datadog.Trace.ClrProfiler.Native/module_metadata.h
tonyredondo: TIL, done in f18e9da

tonyredondo:
> Do you know what's the impact on the working set? Basically, how much memory is the sample using in total?

Testing using `Samples.HttpMessageHandler` on linux 4 times (mmem = M Maximum resident set size of the process during its lifetime, in Kilobytes.) :

### Before
```
 0:02.20 real time
 33% percentage
 0.69 user
 0.05 sys
 0 amem
 87452 mmem
 0 page faults
 12374 other page faults
 4096 context-switches involutarily
 1020 waits

---

 0:02.18 real time
 33% percentage
 0.65 user
 0.08 sys
 0 amem
 87848 mmem
 0 page faults
 12387 other page faults
 4280 context-switches involutarily
 1147 waits

---

 0:02.21 real time
 34% percentage
 0.69 user
 0.07 sys
 0 amem
 87308 mmem
 0 page faults
 12390 other page faults
 3432 context-switches involutarily
 1008 waits

 ---

 0:02.12 real time
 33% percentage
 0.63 user
 0.08 sys
 0 amem
 87712 mmem
 0 page faults
 12434 other page faults
 4598 context-switches involutarily
 1208 waits
```

### After
```
 0:02.05 real time
 34% percentage
 0.62 user
 0.09 sys
 0 amem
 77796 mmem
 0 page faults
 9909 other page faults
 2212 context-switches involutarily
 1098 waits

---

 0:02.29 real time
 35% percentage
 0.71 user
 0.09 sys
 0 amem
 78032 mmem
 0 page faults
 9972 other page faults
 5245 context-switches involutarily
 1163 waits

 ---

 0:02.24 real time
 35% percentage
 0.69 user
 0.09 sys
 0 amem
 77876 mmem
 0 page faults
 9960 other page faults
 3303 context-switches involutarily
 1080 waits

---

 0:02.25 real time
 34% percentage
 0.73 user
 0.04 sys
 0 amem
 77936 mmem
 0 page faults
 9990 other page faults
 2617 context-switches involutarily
 1130 waits
```

gleocadie (approved):
LGTM. TIL

PR #1725, opened 2021-09-02
Add FmtLib debug files that were previously forgotten during import of shared files

macrogreg:
The recent merge of the `dd-shared-components-dotnet` repo included the native [fmt library](https://github.com/DataDog/dd-shared-components-dotnet/tree/master/src/Shared-Libs/Native). During the merge, we included the release version of that library, but we forgot the [debug files](https://github.com/DataDog/dd-shared-components-dotnet/tree/master/src/Shared-Libs/Native/fmt_x64-windows-static/debug/lib).

Now, we need to have the Profiler use the shared files from their new location in `dd-trace-dotnet`. A [PR](https://github.com/DataDog/dd-continuous-profiler-dotnet/pull/135) for this is open, however, that PR is currently not building correctly, because the fmt debug files are missing.

This change adds the missing files.
There is no other functional change, beyond unblocking the Profiler build.


zacharycmontoya (approved):

lucaspimentel:
Is this what we want? We don't normally push compiled binaries to the repository. In the tracer today we pull the sources and compile them as part of our build process.

lucaspimentel (approved):

macrogreg:
@lucaspimentel I agree and prefer to remove it, including from history. But it would need to be a separate exercise: right now I am simply moving the code as-is.

PR #1728, opened 2021-09-02
Add a beta suffix to the Datadog.Monitoring.Distribution NuGet package

zacharycmontoya:
By adding an additional `<Version>$(Version)-beta01</Version>` property to the NuGet package, we can get a `-beta01` suffix regardless of the version of the release. This achieves the following results:
1. Running the version update with the existing version does not change the version number
```powershell
$env:Version = "1.28.5"
build.ps1 UpdateVersions
```

2. Running the version update with a newer version changes the version number
```powershell
$env:Version = "1.28.6"
$env:IsPrerelease = "false"
build.ps1 UpdateVersions
```

3. Building the NuGet package results in Datadog.Monitoring.Distribution.1.28.6-beta01
```powershell
.\build.ps1 BuildTracerHome CreateDistributionHome BuildDistributionNuget
```

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Monitoring.Distribution/Datadog.Monitoring.Distribution.csproj
+++ b/src/Datadog.Monitoring.Distribution/Datadog.Monitoring.Distribution.csproj
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <Version>1.28.5-prerelease</Version>
+    <Version>$(Version)-beta.1</Version> <!-- Unconditionally add a beta suffix to the package, but keep the rest of the name so we can associate it with the rest of the release -->
     <Title>Datadog APM Auto-instrumentation Assets</Title>
     <Description>Auto-instrumentation assets for Datadog APM</Description>
     <IncludeBuildOutput>false</IncludeBuildOutput>
--- a/src/Datadog.Monitoring.Distribution/Datadog.Monitoring.Distribution.csproj
+++ b/src/Datadog.Monitoring.Distribution/Datadog.Monitoring.Distribution.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <Version>1.28.5-prerelease</Version>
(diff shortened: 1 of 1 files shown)


andrewlock (approved):
I wonder if we should use -beta01 [to support older NuGet clients](https://docs.microsoft.com/en-us/nuget/create-packages/prerelease-packages#semantic-versioning)

image

zacharycmontoya:
> I wonder if we should use -beta01 [to support older NuGet clients](https://docs.microsoft.com/en-us/nuget/create-packages/prerelease-packages#semantic-versioning)
> 
> image

Yeah that's a good idea, will do

src/Datadog.Monitoring.Distribution/Datadog.Monitoring.Distribution.csproj
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <Version>1.28.5-prerelease</Version>
+    <Version>$(Version)-beta01</Version> <!-- Unconditionally add a beta suffix to the package, but keep the rest of the name so we can associate it with the rest of the release -->
```
lucaspimentel, line 5: Could this be done with `<VersionSuffix>` instead? Seems a little cleaner that using `<Version>` twice.

See this blog some "some guy" wrote: https://andrewlock.net/version-vs-versionsuffix-vs-packageversion-what-do-they-all-mean/

lucaspimentel: >setting Version explicitly will override any `VersionPrefix` or `VersionSuffix` settings.

Ehh. Looks like we would need to use  `<VersionPrefix>` and `<VersionSuffix>` for this to work.

zacharycmontoya: Yeah, this approach works and it creates the smallest amount of changes, so I'd like to keep this approach

lucaspimentel (approved):

PR #1734, opened 2021-09-03
Add exceptions to active span during ASP.NET Web API 2 message handler exception

zacharycmontoya:
## Scenario
At least one scenario has been identified today where an exception is thrown during while handling an ASP.NET Web API 2 request and we are unable to attach the error to our automatic instrumentation spans. This occurs when the application registers custom message handlers in the ASP.NET Web API 2 and the message handlers throw exceptions (before reaching the Api controllers). The result is that there was no `aspnet-webapi.request` span and the `aspnet.request` root span only sees the 500 status code because the exception was handled by the ASP.NET Web API 2 pipeline before returning.

## Fix
Add an automatic instrumentation method into the global error handling of ASP.NET Web API 2 applications. When an exception is detected by the Web API 2 pipeline, the registered `IExceptionHandler` singleton is invoked before returning the request to the server, so we hook into that call to grab the exception object and attach it to the active span.

Note: This exception handling also applies to OWIN applications running Web API 2 endpoints, so once we automatically instrument OWIN applications to generate root spans this feature will come along for free.

### Before
```
Tags: {
  error.msg: The HTTP response has status code 500.
}
```

### After
```
Tags: {
      error.msg: Passed in value was not 'true': false,
      error.stack: 
System.ArgumentException: Passed in value was not 'true': false
at Samples.AspNetMvc5.Controllers.ApiController.TransientFailure(String value),
      error.type: System.ArgumentException,
}
```

## Testing
This has been confirmed with the following test cases, on both ASP.NET Web API 2 applications and OWIN Web API 2 applications:
1. Adding a global message handler
2. Adding a per-route message handler

## Resources:
[1] https://docs.microsoft.com/en-us/aspnet/web-api/overview/advanced/http-message-handlers
[2] https://docs.microsoft.com/en-us/aspnet/web-api/overview/error-handling/web-api-global-error-handling

@DataDog/apm-dotnet


lucaspimentel (approved):

andrewlock (approved):

zacharycmontoya:
I need to revisit this. I just tested two scenarios with the existing code, and here's what currently happens:
- Scenario: Add a global message handler and error in the handler. Result: Does not work
- Scenario: Insert a per-route message handler and error in the handler. Result: Works

src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AspNet/DelegatingHandlerIntegration.cs
```diff
@@ -12,11 +12,11 @@
...
 {
     /// <summary>
-    /// System.Web.Http.Dispatcher.HttpRoutingDispatcher calltarget instrumentation
+    /// System.Net.Http.DelegatingHandler calltarget instrumentation
     /// </summary>
     [InstrumentMethod(
-        AssemblyName = "System.Web.Http",
-        TypeName = "System.Web.Http.Dispatcher.HttpRoutingDispatcher",
+        AssemblyName = "System.Net.Http",
+        TypeName = "System.Net.Http.DelegatingHandler",
```
andrewlock, line 19: This seems a bit far reaching to me 🤔 `DelegatingHandler` makes up the pipeline of handlers in `HttpClient`, so we'd be instrumenting all of those too. I haven't looked to see if there's other options, but instrumenting this one concerns me somewhat 😟

zacharycmontoya: I understand your concern. Maybe we can add a new message handler to the beginning of the pipeline using automatic instrumentation. I'm going to experiment with that approach and see if it works

tracer/src/Datadog.Trace/ClrProfiler/ClrNames.cs
```diff
@@ -45,5 +45,7 @@ internal static class ClrNames
         public const string GenericParameterTask = "System.Threading.Tasks.Task`1<T>";
         public const string ObjectTask = "System.Threading.Tasks.Task`1<System.Object>";
         public const string Int32Task = "System.Threading.Tasks.Task`1<System.Int32>";
+
+        public const string HttpConfigurationAction = "System.Action`1<System.Web.Http.HttpConfiguration>";
```
lucaspimentel, line 49: Are you using this anywhere? (I couldn't find it.)

lucaspimentel (approved):
LGTM, aside from one non-blocking comment.

tracer/src/Datadog.Trace/ClrProfiler/ClrNames.cs
zacharycmontoya: Whoops this is no longer needed, good catch! Removed in b5afa0200db546dca90dc571b52065c20e6d4d83

src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AspNet/DelegatingHandlerIntegration.cs
zacharycmontoya: I updated the instrumentation in 7c01174399bcaa4e67728a2a379b54340e8e7ac0 to instrument the exception handling pipeline instead, so it should be a much more constrained solution 👍🏼

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AspNet/ExceptionHandlerExtensions_HandleAsync_Integration.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// OnAsyncMethodEnd callback
+        /// </summary>
+        /// <typeparam name="TTarget">Type of the target</typeparam>
+        /// <typeparam name="TResponse">Type of the response, in an async scenario will be T of Task of T</typeparam>
+        /// <param name="instance">Instance value, aka `this` of the instrumented method. This will be null because the method is static</param>
+        /// <param name="response">Response instance</param>
+        /// <param name="exception">Exception instance in case the original code threw an exception.</param>
+        /// <param name="state">Calltarget state value</param>
+        /// <returns>A response value, in an async scenario will be T of Task of T</returns>
+        public static TResponse OnAsyncMethodEnd<TTarget, TResponse>(TTarget instance, TResponse response, Exception exception, CallTargetState state)
```
andrewlock, line 73: I believe you need to add `[PreserveContext]` to this function, but @kevingosse will confirm

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AspNet/ExceptionHandlerExtensions_HandleAsync_Integration.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            // Do nothing. Do I need OnAsyncMethodEnd?
```
andrewlock, line 75: I don't think you would _except_ that we need the `[PreserveContext]` attribute?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AspNet/HttpConfigurationStruct.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// </summary>
+    [DuckCopy]
+    [Browsable(false)]
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public struct HttpConfigurationStruct
+    {
+        /// <summary>
+        /// Gets the MessageHandlers collection
+        /// </summary>
+        [Duck(BindingFlags = DuckAttribute.DefaultFlags | BindingFlags.IgnoreCase)]
```
andrewlock, line 24: `DuckAttribute.DefaultFlags` doesn't hit the same problem you found with `DuckField` in #1601 does it?

tracer/test/snapshots/AspNetWebApi2Tests.CallSite.Classic.NoFF.__path=_api2_transientfailure_false_statusCode=500.verified.txt
```diff
@@ -2,14 +2,31 @@
...
+    ParentId: Id_3,
+    Tags: {
+      env: integration_tests,
+      span.kind: server,
+      version: 1.0.0
+    }
+  },
+  {
+    TraceId: Id_1,
+    SpanId: Id_3,
```
andrewlock, line 17: ~Were we expecting an _extra_ span in this unrelated endpoint?~ 🤔 

EDIT: Oh I see, it's because we've added an extra exception handler. I imagine in practice the exception handler would _also_ set the exception in this span too

andrewlock (commented):
Thanks for finding a better hook! I think we might need the `[PreserveContext]` attribute though

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AspNet/ExceptionHandlerExtensions_HandleAsync_Integration.cs
andrewlock: >Kevin: Well the attribute isn't needed if the code in the integration doesn't use the HttpContext. Here it doesn't look like it does

So ignore me 🙂

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AspNet/ExceptionHandlerExtensions_HandleAsync_Integration.cs
andrewlock: As we don't need the attribute, I believe you can remove the method (you can certainly remove the begin method when it's not needed)

andrewlock (approved):

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AspNet/HttpConfigurationStruct.cs
zacharycmontoya: No it doesn't hit the same issue because `BindingFlags` is defined in corlib, not in a custom assembly

tracer/test/snapshots/AspNetWebApi2Tests.CallSite.Classic.NoFF.__path=_api2_transientfailure_false_statusCode=500.verified.txt
zacharycmontoya: Ah yes, I could do probably set the same exception in the exception handler span. I mostly just wanted to make sure that exception handler was invoked so I just created a minimal span. Hope that implementation is cool with you

PR #1740, opened 2021-09-06
Refactor ILogger integration

andrewlock:
Changes in this PR:

* Moves the  `ILogger` instrumentation into Logging sub-folder. We're going to have a lot of common code for the different ILogger integrations for direct log shipping, so makes sense to do this I think
* Call `ToString()` on `SpanId` and `TraceId`. Ensures they're written as strings by log providers to hopefully circumvent parsing issues when using custom log parsing.
* Addd a benchmark for logs injection (uninjected vs injected results shown below, but only the injected benchmark included, in line with other log injection)
* Added a sample app demonstrating the ILogger logs injection
* Output a better format string from DatadogScope. Some simplistic logger providers just call `ToString()` on an injected scope, instead of enumerating the values. This change ensures there's a valid json blob in the output for the logs agent to parse.

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Logging/ILogger/DatadogLoggingScope.cs
+++ b/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Logging/ILogger/DatadogLoggingScope.cs
@@ -8,7 +8,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Globalization;
 
-namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.ILogger
+namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Logging.ILogger
 {
     internal class DatadogLoggingScope : IReadOnlyList<KeyValuePair<string, object>>
     {
--- a/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Logging/ILogger/DatadogLoggingScope.cs
+++ b/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Logging/ILogger/DatadogLoggingScope.cs
@@ -41,8 +41,8 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Logging.ILogger
                     0 => new KeyValuePair<string, object>("dd_service", _tracer.ActiveScope?.Span.ServiceName ?? _service),
                     1 => new KeyValuePair<string, object>("dd_env", _env),
                     2 => new KeyValuePair<string, object>("dd_version", _version),
-                    3 => new KeyValuePair<string, object>("dd_trace_id", _tracer.ActiveScope?.Span.TraceId),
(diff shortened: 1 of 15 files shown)


src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Logging/ILogger/DatadogLoggingScope.cs
```diff
@@ -51,15 +58,19 @@ internal DatadogLoggingScope(Tracer tracer)
...
             var span = _tracer.ActiveScope?.Span;
+            if (span is null)
+            {
+                return _cachedFormat;
+            }
 
             return string.Format(
                 CultureInfo.InvariantCulture,
-                "dd_service:{0}, dd_env:{1}, dd_version:{2}, dd_trace_id:{3}, dd_span_id:{4}",
+                "{{\"dd_service\":\"{0}\", \"dd_env\":\"{1}\", \"dd_version\":\"{2}\", \"dd_trace_id\":\"{3}\", \"dd_span_id\":\"{4}\"}}",
```
andrewlock, line 68: I'm not 100% sure if this is a good idea. It should help with log parsing, but also isn't exactly "standard". Any "good" `ILoggerProvider` implementation will be calling `GetEnumerator`, but that isn't guaranteed...

zacharycmontoya: Yeah I'm also not sure about that. What if the only change at this moment in time is to quote the property values? It's not as opinionated but I think it may be a small improvement over what we have today

src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Logging/ILogger/DatadogLoggingScope.cs
```diff
@@ -51,10 +58,14 @@ internal DatadogLoggingScope(Tracer tracer)
         public override string ToString()
         {
             var span = _tracer.ActiveScope?.Span;
+            if (span is null)
+            {
+                return _cachedFormat;
```
zacharycmontoya, line 63: Nice!

zacharycmontoya (approved):
Looks good! I left one small comment on the ToString result

lucaspimentel (approved):

src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Logging/ILogger/DatadogLoggingScope.cs
andrewlock: Yep, I'm more comfortable with that too. done in https://github.com/DataDog/dd-trace-dotnet/pull/1740/commits/1768e10368deb90ab9e43b8c98d22a4dadedf99d

PR #1729, opened 2021-09-02
Minor tweaks to shared files

macrogreg:
As we are cleaning up the Profiler build to reference `dd-trace-dotnet` instead of `dd-shared-components-dotnet`, we discover the need to minor tweaks. This PR is the collection of such tweaks which I collected into a single PR to minimize churn.

The Profiler work is now done (https://github.com/DataDog/dd-continuous-profiler-dotnet/pull/135).
Note that while this PR is being reviewed and merged, the Profiler builds cleanly while referencing `dd-trace-dotnet / macrogreg/SharedItemsCatchup`. Once this PR is merged, we must point the profiler to `dd-trace-dotnet / master`.

Changes here:

* Make packages path the same for all shared items.
* Props files must use relative paths to construct shared file references.
* Fix upward walking chains in `Directory.Build.props` files.

PR #1741, opened 2021-09-06
Fix profiler build issue

gleocadie:
When compiling the profiler, props files that uses EnlistmentRoot variable will point to the profiler repository.
But in the shared logging `props` files, EntlistmentRoot variable is considered as being the tracer repository.
This prevents the profiler to build correctly its managed libraries.

Fixes #

For now, in the shared logging `props` files, we will consider EntlistmentRoot variable as the path to the profiler repository and computer the other path relative to it.

@DataDog/apm-dotnet

### DIFF
--- a/shared/src/managed-src/Datadog.Logging/Datadog.Logging.Composition.props
+++ b/shared/src/managed-src/Datadog.Logging/Datadog.Logging.Composition.props
@@ -5,7 +5,7 @@
   </PropertyGroup>
   
   <PropertyGroup>
-    <SharedSrcBaseDir>$(EnlistmentRoot)\shared\src\managed-src</SharedSrcBaseDir>
+    <SharedSrcBaseDir>$(EnlistmentRoot)\..\dd-trace-dotnet\shared\src\managed-src</SharedSrcBaseDir>
     <SharedSrcBaseLabel>Shared-Src</SharedSrcBaseLabel>
   </PropertyGroup>
   
--- a/shared/src/managed-src/Datadog.Logging/Datadog.Logging.Emission.props
+++ b/shared/src/managed-src/Datadog.Logging/Datadog.Logging.Emission.props
@@ -5,7 +5,7 @@
   </PropertyGroup>
   
   <PropertyGroup>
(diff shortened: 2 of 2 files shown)


tonyredondo (approved):

macrogreg (changes requested):
Hi @gleocadie ,

My apologies for causing this problem, I have noted its existence of this issue in the description of https://github.com/DataDog/dd-continuous-profiler-dotnet/pull/135, but I should have mentioned it in Slack too, so that it does not get missed.

The background is that as I was doing the work for the Profiler, I was discovering ever more tiny changes required in the Tracer repo, and to minimize churn, I collected all the required changes into [a single PR](https://github.com/DataDog/dd-trace-dotnet/pull/1729) and I was waiting to merge the Profiler changes, which would mean that we know that no more changes to the Tracer are required, before actually getting the Tracer tweaks reviewed and merged.

In the meantime, I have pointed the Profiler CI to use the Tracer branch `macrogreg/SharedItemsCatchup`. So the CI will pass and to work-around locally you need to check out that branch for the Tracer repo.

Now that you have approved the Profiler work (thanks for the reviews!) I have moved https://github.com/DataDog/dd-trace-dotnet/pull/1729 from Draft to Ready and will merge it as soon as it is approved (please take a look).

I would prefer to use https://github.com/DataDog/dd-trace-dotnet/pull/1729 over _this_ change for the following reasons:
- They contain the same logical change;
- However, the way of formulating the path in _this_ PR will only work for consuming the affected shared files from the Profiler repo, but https://github.com/DataDog/dd-trace-dotnet/pull/1729 will work for consuming those files from any repo;
- The Profiler CI points to https://github.com/DataDog/dd-trace-dotnet/pull/1729, thus - consistency.

Hope this helps. Apologies, again, for not notifying you of the issue more proactively. @chrisnas , please take note. Cheers!

macrogreg:
I will close this, because the the changes contained here were just merged as a part of https://github.com/DataDog/dd-trace-dotnet/pull/1729

PR #1743, opened 2021-09-07
Header keys should be lower case

robertpi:
Fixes #

@DataDog/apm-dotnet

### DIFF
--- a/src/Datadog.Trace/Util/Http/HttpRequestExtensions.Core.cs
+++ b/src/Datadog.Trace/Util/Http/HttpRequestExtensions.Core.cs
@@ -5,6 +5,7 @@
 
 #if !NETFRAMEWORK
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using Datadog.Trace.AppSec;
 using Microsoft.AspNetCore.Http;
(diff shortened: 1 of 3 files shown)


andrewlock (approved):

src/Datadog.Trace/Util/Http/HttpRequestExtensions.Core.cs
```diff
@@ -24,7 +25,7 @@ internal static partial class HttpRequestExtensions
             {
                 if (!k.Equals("cookie", System.StringComparison.OrdinalIgnoreCase))
                 {
-                    headersDic.Add(k, request.Headers[k]);
+                    headersDic.Add(k.ToLower(CultureInfo.InvariantCulture), request.Headers[k]);
```
andrewlock, line 28: Any reason to not use `ToLowerInvariant()`?
```suggestion
                    headersDic.Add(k.ToLowerInvariant(), request.Headers[k]);
```

robertpi: have to admit the main reason I went with ToLower is that it's higher in the autocomplete list :)

robertpi: fixed in a2e44b78

PR #1749, opened 2021-09-07
Fix Benchmarks build break from removing Datadog.Trace.Abstractions namespace

zacharycmontoya:
@DataDog/apm-dotnet


lucaspimentel (approved):
Thanks!

PR #1757, opened 2021-09-09
Env var to override default AppSec rules

robertpi:
Fixes #

Changes proposed in this pull request:

@DataDog/apm-dotnet

PR #1761, opened 2021-09-10
Appsec/robert/peer testing

robertpi:
Fixes #

@DataDog/apm-dotnet

### DIFF
--- a/docs/peer-testing.md
+++ b/docs/peer-testing.md
@@ -1,6 +1,6 @@
 # [Datadog .NET AppSec] Peer-testing
 
-This document describes how to test the installation of the .NET App Sec Library
+Peer-testing is a test script that allows engineers outside of the .NET group to test the installation and basic functionality of the app sec components.
 
 ## Testing Locally
 
--- a/docs/peer-testing.md
+++ b/docs/peer-testing.md
@@ -1,0 +1,170 @@
+# [Datadog .NET AppSec] Peer-testing
+
+This document describes how to test the installation of the .NET App Sec Library
+
+## Testing Locally
+
+This assumes the agent is already running, it requires agent 7.31.0-rc.6 or later.
+
+Install .NET (yes, it does run on Mac and Linux): https://dotnet.microsoft.com/download
+
+Create the app
+
+```
+dotnet new webapp -n DotnetPeerTest
+```
+
+Run the app:
+
+```
+cd DotnetPeerTest
+dotnet run DotnetPeerTest.csproj
+```
+
+Check it works:
+
+```
+curl -k https://localhost:5001
+```
+
+Add the Datadog package:
+
+```
+dotnet add package Datadog.Monitoring.Distribution --prerelease
+```
+
+Set the env var.
+
+Windows
+
+```
+SET DD_ENV=peer-test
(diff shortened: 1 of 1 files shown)


docs/peer-testing.md
```diff
@@ -0,0 +1,170 @@
+# [Datadog .NET AppSec] Peer-testing
+
+This document describes how to test the installation of the .NET App Sec Library
+
```
andrewlock, line 4: A little blurb about what peer-testing is might be nice (for me 😅 )

docs/peer-testing.md
```diff
@@ -0,0 +1,170 @@
...
+This assumes the agent is already running, it requires agent 7.31.0-rc.6 or later.
+
+Install .NET (yes, it does run on Mac and Linux): https://dotnet.microsoft.com/download
+
+Create the app
+
+```
+dotnet new webapp -n DotnetPeerTest
+```
```
andrewlock, line 15: Rather than having these instructions, can we create the sample app directly? That way we can setup all the env vars and everything automatically too (like we do in this sample for example https://github.com/DataDog/dd-trace-dotnet/tree/master/tracer/samples/AutomaticTraceIdInjection/MicrosoftExtensionsExample)

tracer/src/Datadog.Trace/AppSec/Security.cs
```diff
@@ -53,7 +53,7 @@ private Security(SecuritySettings settings = null, InstrumentationGateway instru
                 _settings.Enabled = _settings.Enabled && AreArchitectureAndOsSupported();
                 if (_settings.Enabled)
                 {
-                    _powerWaf = powerWaf ?? Waf.Waf.Initialize();
+                    _powerWaf = powerWaf ?? Waf.Waf.Initialize(_settings.Rules);
```
andrewlock, line 56: This is all in https://github.com/DataDog/dd-trace-dotnet/pull/1757 right? shouldn't be in this pr?

robertpi: Yes, git accident, I was hoping to correct it before anyone noticed, it seems this was not possible 😄

docs/peer-testing.md
robertpi: fixed in a92010b64534fc8ef02f7ad7689d58fb430b4232

docs/peer-testing.md
robertpi: The idea is to have someone outside of the .NET group test the installation and basic functionality, we want to simulate the steps that a customer would go through. I therefore think have a prebuilt app would make that too easy.

robertpi:
I wasn't sure where the best place to put this was, but as I wrote it markdown, I thought it may as well go here.

docs/peer-testing.md
andrewlock: Yeah, makes sense. I think having a sample _as well_ makes for better documentation 🙂

robertpi:
Decide close in favour of using Confluence: https://datadoghq.atlassian.net/wiki/spaces/APS/pages/2159445754/Dotnet+Peer+Testing+Instuctions 

The sample is a good idea, I thought @andrewlock already added a sample based on the new nuget, but I can't find it anymore.

PR #1805, opened 2021-09-22
Clean the workspace before building on ARM64

kevingosse:


colin-higgins (approved):

zacharycmontoya (approved):

kevingosse:
Closing this as it would make the build significantly slower. We can bring it back if the issue occurs once more.

PR #1807, opened 2021-09-22
Exclude the profiler from the regression sample apps

andrewlock:
It's just a temporary fix, but it saves ~1GB of drive space which is sufficient to get the regression samples running again

@DataDog/apm-dotnet


robertpi (approved):
Don't know if you're still undecided, but I think https://github.com/DataDog/dd-trace-dotnet/pull/1808 is better.

andrewlock:
Closing in favour of #1808 as it doesn't actually solve the issue anyway

PR #1820, opened 2021-09-27
Managed loader cannot resolve profiler managed libraries when installed by the MSI installer

gleocadie:
### Context

When installing the profiler and the tracer using the MSI installer, the profiler managed library cannot be resolved.

To resolve the profiler managed library, the managed loader uses the `COR_PROFILER_PATH_XX` or `CORECLR_PROFILER_PATH_XX` to locate the profiler home directory.

But, the MSI installer does not populate those environment variables, instead it registers the CLR profiler CLSID and the path in the registry.

### Fixes

To resolve the tracer managed library, the managed loader relies on the `DD_DOTNET_TRACER_HOME` environment variable. (even if today the tracer does not use the managed loader).

We can introduction a `DD_DOTNET_PROFILER_HOME` environment variable to replicate the same pattern used for the tracer. 

@DataDog/apm-dotnet


shared/src/managed-lib/ManagedLoader/Datadog.AutoInstrumentation.ManagedLoader/public/AssemblyLoader.cs
```diff
@@ -412,104 +412,62 @@ private static void GetTracerManagedBinariesDirectories(List<string> binaryDirs)
...
-
-                string profilerHomeDirectory = nativeProfilerLoaderAssemblyDirectory;                           // Profiler-HOME
-                string managedBinariesDirectory = Path.Combine(profilerHomeDirectory, managedBinariesSubdir);   // Managed binaries are in <Profiler-HOME>/net-ver-moniker/
-
-                if (binaryDirs != null && !String.IsNullOrWhiteSpace(managedBinariesDirectory))
-                {
-                    binaryDirs.Add(managedBinariesDirectory);
-                }
-            }
+#if DEBUG
```
chrisnas, line 436: Maybe use System.Diagnostics.Debugger.IsAttached so it would be possible to debug a RELEASE build

chrisnas (changes requested):

shared/src/managed-lib/ManagedLoader/Datadog.AutoInstrumentation.ManagedLoader/public/AssemblyLoader.cs
gleocadie: ✔️

zacharycmontoya (approved):
LGTM

macrogreg (approved):
Looks good. Please see one change request in the PR about F5/Ctrl-F5 mode. Thanks!

tonyredondo (approved):

zacharycmontoya:
I ran a new pipeline run that passed, which is here: https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=76828&view=results

If there are no further code changes, we should have an admin force merge this PR.

chrisnas (approved):

PR #1850, opened 2021-10-01
Fix performance pipeline

andrewlock:
With the additional triggers in the performance pipeline, it was being triggered twice
- Immediately when a push to a monitored branch
- When the ci pipeline had finished building

We only want the latter one

Also removes `ContinueOnError` from benchmarks job, as failures in this pipeline don't block the ability to merge, so they should be counted as real failures.

~There's still an issue with the throughput tests I don't entirely understand (it's unable to checkout the merge commit GitHub creates) but this fixes benchmark overcrowding at least~

The issue with the throughput tests was because we rely on the `SYSTEM_PULLREQUEST_SOURCECOMMITID` being set to determine which commit to checkout in the agents, but that isn't set when triggered by a separate build. 

@DataDog/apm-dotnet


andrewlock:
Something's gone wonky, and I can't push to this any more, so closing in favour of a new one

PR #279, opened 2019-03-06
Distributed tracing and sampling priority propagation

lucaspimentel:
Use Trace Agent endpoint v0.4 to retrieve sample rate map and set sampling priorities for distributed traces.

Other fixes:
- add missing `ConfigureAwait(false)` in `AgentWriter` and `Api`
- general code clean up (readonly fields and properties, autoproperties, comments, etc)
- instrument only `HttpClientHandler` and no other `HttpMessageHandler` implementations

Note:
- this PR adds a dependency to `Newtonsoft.Json`

EDIT: ended up refactoring a lot of the core classes (`Span`, `SpanContext`, `TraceContext`, etc) to support propagating different values at the trace level and to remove some dependencies between classes.

PR #1867, opened 2021-10-05
Kill the old Samples.Shared project

kevingosse:
This project wasn't actually shared, and can be confused with `tracer\test\test-applications\Samples.Shared`.


lucaspimentel (approved):

andrewlock (approved):
FWIW, those types are _used_ in Samples.AspNetMvc5 (in `DistributedTracingApiController`) but AFAICT, we don't actually test anything around them, so maybe they should just be removed?

kevingosse:
> FWIW, those types are _used_ in Samples.AspNetMvc5 (in `DistributedTracingApiController`) but AFAICT, we don't actually test anything around them, so maybe they should just be removed?

I suspect they may be here for manual testing. @lucaspimentel @zacharycmontoya  does it ring a bell?

lucaspimentel:
Looks like something I used a long time ago to test distributed tracing (back when I first implemented it in #279). I don't think we ever wrote automated tests using that specific endpoint.

PR #1921, opened 2021-10-19
Include the package version in all the MySqlConnector tests

andrewlock:
We have to include the package version in _all_ the tests. When the MySqlConnector sample is built for the multi-api packages, it's _not_ built for the default package, which causes failures on master only

@DataDog/apm-dotnet


andrewlock:
[Confirmed this fixes it when running all package versions too](https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=78756&view=results)

PR #1966, opened 2021-10-29
[2.0] Additional cleanup from CallSite removal

andrewlock:
Follow on cleanup from #1958 

- Clean up empty xml in installer files
- Remove remaining references to call site in Nuke helpers
- Delete the CallSite snapshots for tests removed in #1958
- Remove the .CallTarget suffix from remaining snapshots, and rename files as appropriate
- Remove references to DD_TRACE_NETSTANDARD_ENABLED and DD_TRACE_CALLTARGET_ENABLED
- Other minor cleanup

@DataDog/apm-dotnet


lucaspimentel:
Doesn't have to be in this PR, but `tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/SmokeTests/EntityFramework6xMdTokenLookupFailure.cs` was used to test a code path in call-site instrumentation. If we remove call site, we don't need this test anymore.

PR #1991, opened 2021-11-04
merge branch `hotfix/1.29.1-prerelease` branch into master

lucaspimentel:
[Release 1.29.1-prerelease](https://github.com/DataDog/dd-trace-dotnet/releases/tag/v1.29.1-prerelease) was created last week. We don't need this branch anymore, but we should merge its changes into `master` and delete it.


andrewlock:
I don't think we should merge this into master, the commit is already there right? I think we should cherry-pick the version bump commit as a new PR instead?

zacharycmontoya:
I agree with Andrew. That version bump commit is the only difference between the hotfix branch and master

PR #2009, opened 2021-11-09
Fix StartActiveWithTags to start span with given SpanID

pellared:
## What

Fix `Tracer.StartActiveWithTags` to start span with given `spanId`.

## Why

Right now the `traceId` is assigned instead. 

@DataDog/apm-dotnet

### DIFF
--- a/tracer/src/Datadog.Trace/Tracer.cs
+++ b/tracer/src/Datadog.Trace/Tracer.cs
@@ -419,7 +419,7 @@ namespace Datadog.Trace
 
         internal Scope StartActiveWithTags(string operationName, ISpanContext parent = null, string serviceName = null, DateTimeOffset? startTime = null, bool ignoreActiveScope = false, bool finishOnClose = true, ITags tags = null, ulong? spanId = null)
         {
-            var span = StartSpan(operationName, tags, parent, serviceName, startTime, ignoreActiveScope, spanId);
+            var span = StartSpan(operationName, tags, parent, serviceName, startTime, ignoreActiveScope, spanId: spanId);
             return _scopeManager.Activate(span, finishOnClose);
         }
 


andrewlock (approved):
Eek, good spot, thanks!

robertpi:
At my last place we prevent these kind of errors with strongly typed ids, using this tool: https://github.com/vbfox/stidgen

As we don't have that many ids, we could probably do something similar without the tool.

andrewlock:
> At my last place we prevent these kind of errors with strongly typed ids, using this tool: https://github.com/vbfox/stidgen
> 
> As we don't have that many ids, we could probably do something similar without the tool.

Source generators 💪 

- https://github.com/andrewlock/StronglyTypedId
- https://andrewlock.net/using-strongly-typed-entity-ids-to-avoid-primitive-obsession-part-1/

I expect we don't want the small overhead but more annoying serialization issues etc

kevingosse:
Looks like the `spanId` parameter on `StartActiveWithTags` is not used anywhere (I guess that's why I missed it). Let's remove it entirely.

andrewlock:
> Looks like the `spanId` parameter on `StartActiveWithTags` is not used anywhere (I guess that's why I missed it). Let's remove it entirely.

It _was_ used here: https://github.com/DataDog/dd-trace-dotnet/blob/release/2.0/tracer/src/Datadog.Trace/ClrProfiler/ScopeFactory.cs#L132

Which is used in `HttpWebRequest_EndGetResponse_Integration`.

But looks like you refactored all that in #1947 and removed the problematic usage

kevingosse:
I pushed the change in https://github.com/DataDog/dd-trace-dotnet/pull/2017. Thanks!

PR #2177, opened 2021-12-16
Minor improvements

tonyredondo:
This PR contains minor improvements to the source code to reduce an allocation and improve inlining.

@DataDog/apm-dotnet

### DIFF
--- a/tracer/src/Datadog.Trace/Util/ThrowHelper.cs
+++ b/tracer/src/Datadog.Trace/Util/ThrowHelper.cs
@@ -1,0 +1,71 @@
+// <copyright file="ThrowHelper.cs" company="Datadog">
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2 License.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
+
+namespace Datadog.Trace.Util
+{
+    internal class ThrowHelper
+    {
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        [DebuggerHidden]
+        [DoesNotReturn]
+        internal static void ThrowArgumentNullException(string paramName) => throw new ArgumentNullException(paramName);
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        [DebuggerHidden]
+        [DoesNotReturn]
+        internal static void ThrowArgumentOutOfRangeException(string paramName) => throw new ArgumentOutOfRangeException(paramName);
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        [DebuggerHidden]
+        [DoesNotReturn]
+        internal static void ThrowArgumentOutOfRangeException(string paramName, string message) => throw new ArgumentOutOfRangeException(paramName, message);
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        [DebuggerHidden]
+        [DoesNotReturn]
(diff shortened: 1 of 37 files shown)


tracer/src/Datadog.Trace/Util/ThrowHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
+
+namespace Datadog.Trace.Util
+{
+    internal class ThrowHelper
```
lucaspimentel, line 14: We should think about adding analyzers for throwing exceptions, at least for common cases `ArgumentNullException`. (not now, just for the backlog)

lucaspimentel (approved):

tracer/src/Datadog.Trace/Util/ThrowHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+    {
+        [MethodImpl(MethodImplOptions.NoInlining)]
```
zacharycmontoya, line 16: How come these are marked as `NoInlining`?

tracer/src/Datadog.Trace/Util/ThrowHelper.cs
tonyredondo: I agree.

tracer/src/Datadog.Trace/Util/ThrowHelper.cs
tonyredondo: Currently a method with throw doesn't get inlined, that's why we use the helper, to let the caller (with the non exceptional code) to be inlined, so I just wanted to annotate that with the attribute.

Surprisingly, affects the code gen: 
https://sharplab.io/#v2:EYLgHgbALANALiAhgZwLYB8ACAmAjAWAChMAGAAk1wDoAlAVwDs4BLVAUyoGEB7VAB2YAbNgCcAyqIBuzAMZtkAbiJFMAZgrYynMgG8iZA2X2G1FKGQCyuABTMmZZgEpdxw4eYBqD0sJu3ASRsnHz93LxDDAF9XI18TdUxzC2xbeycXONDPbxiAlODcsJzMsmiSmNNEskDUuAdnPRK3ZgAzMlsyAF5OshIGwtCAFQALEW4Ad2tHCNCy0IMBilwATmsAIgAJNkFBbjIAdW4RQQATNemYucMKhPN/fLT+pvc2ju7ep/m/OFGJsgY2OMyABRMByPgsbgMKYzPxXUKLSirTbbXYHI6nc6w+GxNwAbQsbB+3BO/n4gmshOJpPJAHkIcwocgqAA5bj+BiCOx2ADmjgAujczGQRmNJp95j8xf9ASCwWwGVCYZciJEgA

Try commenting the attribute :).

The ASM is very similar but with the attribute is still a bit shorter...

The weird thing to notice is the change from `je` to `jne` or `jz` to `jnz` . I don't know how this can affect the branch predictor. Any idea @kevingosse ?

zacharycmontoya (approved):
LGTM

tracer/src/Datadog.Trace/Util/ThrowHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+        [DebuggerHidden]
+        [DoesNotReturn]
+        internal static void ThrowArgumentNullException(string paramName) => throw new ArgumentNullException(paramName);
```
andrewlock, line 19: [Why no `[CallerArgumentExpression]` 😢 ](https://andrewlock.net/exploring-dotnet-6-part-11-callerargumentexpression-and-throw-helpers/)

tonyredondo: Being handled in another PR by lucas. 😊

PR #2208, opened 2021-12-21
Add a new solution filter including most often used tests projects

pierotibou:
IIRC, Kevin brought this up at some point. If not relevant to other members of the team, I'll just keep it locally.

@DataDog/apm-dotnet


andrewlock (commented):
Personally I don't feel like it's worth adding these. Arguably different people might want completely different filters (and I just load the full solution). I think I would favour adding a _*.slnf_ exception to the gitignore, then everyone can have their own preferred list of projects?

PR #2287, opened 2022-01-10
Don't run the CI when only changing docs.

pierotibou:
@DataDog/apm-dotnet

PR #2296, opened 2022-01-12
A no-op pipeline for non code PRs

pierotibou:
Following #2287, the no-op pipeline is broken.
This pipeline is supposed to run to make sure we can merge commits where the ultimate-pipeline wouldn't run (PRs with no code) without force merge.
This no-op pipeline will then send a success status for all jobs that are required to merge on master.  

@DataDog/apm-dotnet

### DIFF
--- a/.azure-pipelines/noop-pipeline.yml
+++ b/.azure-pipelines/noop-pipeline.yml
@@ -3,6 +3,7 @@ pr:
     exclude:
       - .
     include:
+      - tracer/dependabot/
       - blog/
       - docs/
       - tracer/README.md
--- a/.azure-pipelines/noop-pipeline.yml
+++ b/.azure-pipelines/noop-pipeline.yml
@@ -1,8 +1,18 @@
 pr:
-  branches:
+  paths:
+    exclude:
+      - .
     include:
-      - docs/*
+      - blog/
+      - docs/
       - tracer/README.md
+      - tracer/src/Datadog.Trace/DuckTyping/README.md
+      - tracer/samples
+      - .github/
+      - LICENSE
+      - LICENSE-3rdparty.csv
+      - NOTICE
+      
 
 # Global variables
 variables:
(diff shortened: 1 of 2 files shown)


.azure-pipelines/noop-pipeline.yml
```diff
@@ -1,28 +1,95 @@
-trigger:
-
 pr:
   branches:
     include:
       - docs/*
       - tracer/README.md
```
andrewlock, line 5: I don't think we can just use this approach. For example, this pipeline will run if there are changes to the readme, even if there are also changes to the source code. The checks will then clobber the checks from the "real" pipeline. Also note that wildcards are not supported. And that you're using `branches:` not `paths:` 😉 

Instead, I think we need something like this, though this will also need testing (and we should set up the reverse in the ultimate-pipeline)

```suggestion
pr:
  paths:
    exclude:
      - .
    include:
      - blog/
      - docs/
      - tracer/README.md
      - tracer/src/Datadog.Trace/DuckTyping/README.md
      - tracer/samples
      - .github/
      - LICENSE
      - LICENSE-3rdparty.csv
      - NOTICE
      
```

andrewlock (changes requested):

.azure-pipelines/noop-pipeline.yml
pierotibou: Thanks, I think I have other issues as well as the status wasn't reported below. Looking into it

.azure-pipelines/noop-pipeline.yml
```diff
@@ -3,6 +3,7 @@ pr:
     exclude:
       - .
     include:
+      - tracer/dependabot/
```
andrewlock, line 6: I'm not sure whether we need this one tbh, as we don't ever want to merge those PRs directly 🤔

andrewlock (approved):
Certainly better than a broken one, I guess we'll find out whether it works as we want it to!

lucaspimentel (approved):

zacharycmontoya (approved):

PR #2315, opened 2022-01-14
[DEPENDABOT] Ignore Patches

colin-higgins:
Reduces Dependabot noise for our integrations tracking.
We only really care about ~major~ minor versions where the API may change.
https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#specifying-dependencies-and-versions-to-ignore

@DataDog/apm-dotnet

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,8 +10,8 @@ updates:
       - "area:dependabot"
     ignore:
       - dependency-name: "MessagePack" # Locked at a version that supports our net452 build target
-      - dependency-name: "*" # Ignore minors and patches for all integrations, majors change the signature
-        update-types: ["version-update:semver-minor", "version-update:semver-patch"]
+      - dependency-name: "*" # Ignore patches for all integrations
+        update-types: ["version-update:semver-patch"]
   # Src libraries
   - package-ecosystem: "nuget"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -10,6 +10,8 @@ updates:
       - "area:dependabot"
     ignore:
       - dependency-name: "MessagePack" # Locked at a version that supports our net452 build target
+      - dependency-name: "*" # Ignore minors and patches for all integrations, majors change the signature
+        update-types: ["version-update:semver-minor", "version-update:semver-patch"]
```
zacharycmontoya, line 14: I agree on ignoring patches (gets rid of the obnoxious AWSSDK updates), but I think we should still keep track of minor versions since we do not always instrument public API's. Can we ignore just patches for now and later drop minor if it's still too much noise?

colin-higgins: Done

zacharycmontoya (approved):
LGTM, this should really help with noise!

colin-higgins:
Seems like we need to have a broader discussion on whether we want the noise or not before we merge this.

andrewlock (approved):

lucaspimentel (approved):

PR #2318, opened 2022-01-17
Try fix master CI pipeline

andrewlock:
After https://github.com/DataDog/dd-trace-dotnet/commit/84aff95c56e78e9c802a6186d554c8993b897cf3, CI runs for `master` stopped happening. The scheduled runs kept going, and PR validation is still working, it's just the merges to master that are broken. 

I can't see any reason that commit should have broken things, so trying some things:

- Don't use wildcards in paths ([not supported](https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#paths))
- Try remove `NOTICE`/`LICENSE` from `exclude` - maybe they're being interpreted as folders and confusing something? We never change them anyway, so no biggy
- Try replacing `trigger:none` in the `noop-pipeline`. _No way_ this should affect the main pipeline, but I don't trust AzDo.

We won't know if it's worked till we merge this though 🙄 Yay CI.

@DataDog/apm-dotnet

### DIFF
--- a/.azure-pipelines/noop-pipeline.yml
+++ b/.azure-pipelines/noop-pipeline.yml
@@ -1,4 +1,7 @@
-trigger: none
+trigger:
+  branches:
+    exclude:
+      - "*"
 
 pr:
   paths:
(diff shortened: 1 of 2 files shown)


andrewlock:
Well, looks like we have have master builds again, I guess this isn't necessary.

> When troubleshooting, in AzDo, I noticed a banner saying "there's a problem with the webhooks", so I clicked resolve, and to my extreme surprise, it seems to have worked 🤷‍♂️

PR #2354, opened 2022-01-26
Implement pooling mechanism for StringBuilder

### LINKED ISSUE #2353: Memory Leak
**Describe the bug**
Memory leak due to `[ThreadStatic]` when no synchronization context is present.

**To Reproduce**
1. Enable tracer for any long running ASP.NET Core application with a large threadpool, or high threadpool churn
Due to continuations running on any threadpool thread:
- The cached `StringBuilder` can be returned to any thread
- The thread static memory persists longer than expected (Possibly forever).

**Expected behavior**
Stable memory usage

**Screenshots**
N/A

**Runtime environment (please complete the following information):**
- Instrumentation mode: CLR profiler on linux
- Tracer version: 2.1.1
- OS: Cent OS 8.2
- CLR: .NET 6.0

**Additional context**
The use of `[ThreadStatic]` in `StringBuilderCache` results in a memory leak with high thread pool worker counts, or churn.
We have several high load/performance applications that override the default threadpool configuration, and in each of them StringBuilder uses an excessive amount of memory.

Pooling with very short lived locks should be used in place of `[ThreadStatic]`

I will submit a PR to replace this code with a pooling mechanism instead.

<!--
NOTE: To contact our support team, see https://www.datadoghq.com/support/
When contacting support, please include the information requested above.
If possible, also include the tracer log files found (by default) in:
Windows: %ProgramData%\Datadog .NET Tracer\logs\
Linux: /var/log/datadog/dotnet/
--!>

jasoncouture:
Fixes #2353 

StringBuilderCache used [ThreadStatic] to provide a reusable StringBuilder.
This results in a memory leak and excessive memory usage on applications
that utilize large thread pools, or where thread pool threads are created
and destroyed often.

By using this method instead, we ensure that we never allocate more than a set maximum number of StringBuilder objects.
We also ensure that StringBuilder objects have a semi-deterministic lifetime.

@DataDog/apm-dotnet

### DIFF
--- a/tracer/src/Datadog.Trace/Util/StringBuilderCache.cs
+++ b/tracer/src/Datadog.Trace/Util/StringBuilderCache.cs
@@ -4,53 +4,61 @@
 // </copyright>
 
 using System;
+using System.Collections.Generic;
 using System.Text;
 
 namespace Datadog.Trace.Util
 {
     /// <summary>
-    /// Provide a cached reusable instance of StringBuilder per thread.
+    /// Provide a pooled reusable instance of StringBuilder.
     /// </summary>
-    /// <remarks>
-    /// Based on https://source.dot.net/#System.Private.CoreLib/StringBuilderCache.cs,a6dbe82674916ac0
-    /// </remarks>
     internal static class StringBuilderCache
     {
         internal const int MaxBuilderSize = 360;
-
-        [ThreadStatic]
-        private static StringBuilder _cachedInstance;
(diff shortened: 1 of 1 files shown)


lucaspimentel:
Thanks, Jason. As discussed in the linked github issue, we use `[ThreadStatic]` intentionally and we don't intend to switch to lock-based pooling, so we will close this PR.

PR #2356, opened 2022-01-27
Small changes around pipeline artifacts

pierotibou:
Opening this PR to get feedback around these 2 topics:

- I've you tried to use `download` pipeline tasks. it seems download build task is legacy (cf https://docs.microsoft.com/en-us/azure/devops/pipelines/artifacts/pipeline-artifacts?view=azure-devops&tabs=yaml#migrate-from-build-artifacts)
- And maybe we could reduce a tiny bit the size of the artifacts by publishing only what we need using a [.artifactignore file](https://docs.microsoft.com/en-us/azure/devops/artifacts/reference/artifactignore?view=azure-devop)

@DataDog/apm-dotnet

### DIFF
--- a/.artifactignore
+++ b/.artifactignore
@@ -1,0 +1,9 @@
+**/*
+!**/@(bin|obj|packages)/**
+!**/tracer/src/bin/**
+!**/tracer/build_data/**
+!**/tracer/test/**/test*.xml
+!**/Datadog.Trace.Tools.Runner/bin/**/Tool
+!**/publish/
+!**/Cobertura.xml
+!**/system-tests/logs/**
--- a/.azure-pipelines/steps/restore-working-directory.yml
+++ b/.azure-pipelines/steps/restore-working-directory.yml
@@ -4,10 +4,9 @@ parameters:
     default: 'build-windows-working-directory'
 
 steps:
-- task: DownloadPipelineArtifact@2
+- download: current
   displayName: Download working dir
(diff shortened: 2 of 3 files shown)


andrewlock:
> I've you tried to use download pipeline tasks. it seems download build task is legacy

I think you're mistaking the download artifact _Task_ with the legacy "download build task". We've never used the download build task, the change you've made is just sugar. You can see that if you choose the `Yaml(Task)` tab instead ([from the page you linked](https://docs.microsoft.com/en-us/azure/devops/pipelines/artifacts/pipeline-artifacts?view=azure-devops&tabs=yaml#migrate-from-build-artifacts))

image

That said, there's no harm in it so it's fine 🤷 

> And maybe we could reduce a tiny bit the size of the artifacts by publishing only what we need using a .artifactignore file

Maybe see how much this actually reduces the artifact sizes? If it's significant, it could be worth it (How long does it take to upload the reduced size artifact compared to _without_ the ignore file). But I can also see it causing weird, hard to debug issues as we evolve the pipeline, so if it's not a big difference, I'm inclined to not add it.

pierotibou:
> I think you're mistaking the download artifact Task with the legacy "download build task"
Thanks for the feedback. I misunderstood for the download.

> Maybe see how much this actually reduces the artifact sizes
it will be negligible clearly. The big chunk is the packages:
image

PR #2387, opened 2022-02-02
[Release] A few enhancements after last release

pierotibou:
A few enhancements:
- Forgot the $ for variables in nuget upload
- Auto close the milestone after the release has been published

@DataDog/apm-dotnet

### DIFF
--- a/.github/workflows/auto_create_version_bump_pr.yml
+++ b/.github/workflows/auto_create_version_bump_pr.yml
@@ -77,7 +77,7 @@ jobs:
           body: "${{steps.changes.outputs.release_notes}}"
 
       - name: "Close previous milestone"
-        run: ./tracer/build.sh CloseMilestone
+        run: .\tracer\build.ps1 CloseMilestone
         env:
           Version: ${{steps.versions.outputs.previous_version}}
 
--- a/.github/workflows/auto_create_version_bump_pr.yml
+++ b/.github/workflows/auto_create_version_bump_pr.yml
@@ -77,9 +77,7 @@ jobs:
           body: "${{steps.changes.outputs.release_notes}}"
 
       - name: "Close previous milestone"
-        # We don't close milestones for hotfix releases
-        if: ${{ !contains(github.event.ref, 'hotfix') }}
-        run: ./tracer/build.sh ClosePreviousMilestone
+        run: ./tracer/build.sh CloseMilestone
         env:
           Version: ${{steps.versions.outputs.previous_version}}
 
--- a/.github/workflows/auto_create_version_bump_pr.yml
+++ b/.github/workflows/auto_create_version_bump_pr.yml
@@ -76,6 +76,13 @@ jobs:
           reviewers: "DataDog/apm-dotnet"
           body: "${{steps.changes.outputs.release_notes}}"
 
+      - name: "Close previous milestone"
+        # We don't close milestones for hotfix releases
+        if: ${{ !contains(github.event.ref, 'hotfix') }}
+        run: ./tracer/build.sh ClosePreviousMilestone
+        env:
+          Version: ${{steps.versions.outputs.previous_version}}
+
       - name: Display output
         run: |
           echo "Pull Request Number - ${{ steps.pr.outputs.pull-request-number }}"
--- a/tracer/build/_build/Build.GitHub.cs
+++ b/tracer/build/_build/Build.GitHub.cs
@@ -81,7 +81,7 @@ partial class Build
             Console.WriteLine($"PR assigned");
         });
 
-    Target ClosePreviousMilestone => _ => _
+    Target CloseMilestone => _ => _
        .Unlisted()
        .Requires(() => GitHubToken)
        .Requires(() => Version)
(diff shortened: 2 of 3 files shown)


anna-git (approved):

.github/workflows/auto_create_version_bump_pr.yml
```diff
@@ -76,6 +76,13 @@ jobs:
           reviewers: "DataDog/apm-dotnet"
           body: "${{steps.changes.outputs.release_notes}}"
 
+      - name: "Close previous milestone"
+        # We don't close milestones for hotfix releases
+        if: ${{ !contains(github.event.ref, 'hotfix') }}
```
andrewlock, line 81: We should close milestones for hotfixes, no? 🤔  We have separate milestones for hotfix vs mainline

tracer/build/_build/Build.GitHub.cs
```diff
@@ -74,6 +74,40 @@ partial class Build
             Console.WriteLine($"PR assigned");
         });
 
+    Target ClosePreviousMilestone => _ => _
```
andrewlock, line 77: Wait, have we never been closing them before? 🤔  That seems strange 🤷‍♂️

andrewlock, line 77: nit: can probably rename this to `CloseMilestone`, this target doesn't have any sense of "previous"

tracer/build/_build/Build.GitHub.cs
```diff
@@ -359,8 +396,7 @@ partial class Build
 
             if (previousRelease is not null)
             {
-                sb.AppendLine($"[Changes since {previousRelease.Name}](https://github.com/DataDog/dd-trace-dotnet/compare/v{previousRelease.Name}...v{nextVersion})")
-                  .AppendLine();
+                sb.AppendLine($"[Changes since {previousRelease.Name}](https://github.com/DataDog/dd-trace-dotnet/compare/v{previousRelease.Name}...v{nextVersion})");
```
andrewlock, line 399: If you trim above, we need to insert a line break beforehand
```suggestion
                sb.AppendLine()
                  .AppendLine($"[Changes since {previousRelease.Name}](https://github.com/DataDog/dd-trace-dotnet/compare/v{previousRelease.Name}...v{nextVersion})");
```

.github/workflows/auto_create_version_bump_pr.yml
```diff
@@ -76,6 +76,13 @@ jobs:
           reviewers: "DataDog/apm-dotnet"
           body: "${{steps.changes.outputs.release_notes}}"
 
+      - name: "Close previous milestone"
```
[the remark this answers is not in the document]
pierotibou: I put it there as it's less critical than the others, easy to fix in case it fails. I'd rather leave it there even though that may be only because I've broken too many things lately :/ (I tested it locally though, not saying it will work as I've learned the hard way recently :p)

.github/workflows/auto_create_version_bump_pr.yml
pierotibou: We don't do renaming of vNext milestone for hotfixes, so I copy pasted it. But we don't do version bumps for hotfixes either (see the filter above) so this condition is useless actually. I'll remove it

tracer/build/_build/Build.GitHub.cs
pierotibou: There was a step in the release process to do it manually.

.github/workflows/auto_create_version_bump_pr.yml
pierotibou: done in 670ff1e7ec78fd6da6c207a9d6798fa99b4f630e

tracer/build/_build/Build.GitHub.cs
pierotibou: done in 670ff1e7ec78fd6da6c207a9d6798fa99b4f630e

tracer/build/_build/Build.GitHub.cs
pierotibou: Hmm, it's not the same. here the code you are modifying inserts an extra line in the release notes, after the issues. 
It' still relevant though so I'm adding it anyway in 670ff1e7ec78fd6da6c207a9d6798fa99b4f630e

tracer/build/_build/Build.GitHub.cs
```diff
@@ -81,6 +81,40 @@ partial class Build
...
+            {
+                await client.Issue.Milestone.Update(
+                    owner: GitHubRepositoryOwner,
+                    name: GitHubRepositoryName,
+                    number: milestone.Number,
+                    new MilestoneUpdate { State = ItemState.Closed });
+            }
+            catch (ApiValidationException)
+            {
+                Console.WriteLine($"Unable to close {milestone.Title}.");
```
kevingosse, line 111: Maybe write the exception message?

.github/workflows/auto_create_version_bump_pr.yml
```diff
@@ -76,6 +76,11 @@ jobs:
...
+        run: ./tracer/build.sh CloseMilestone
```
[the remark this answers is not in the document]
pierotibou: Indeed, thanks a lot. Fixed in a0840de92f25173116e71b837a45e91dc6ad66a4

tracer/build/_build/Build.GitHub.cs
pierotibou: Added in a0840de92f25173116e71b837a45e91dc6ad66a4

kevingosse (approved):

PR #2395, opened 2022-02-03
Use IsShutdownRequested to avoid recursive lock acquisition

### LINKED ISSUE #2366: Don't recursively acquire std::shared_mutex in rejit_handler
Hi!

In the file `rejit_handler.cpp` there is extensive use of a reader/writer lock
set around `m_shutdown_lock`.  The logic there looks generally sound except for
one thing: msvc `std::shared_mutex`es use [Slim Reader/Writer locks 
(SRW)](https://docs.microsoft.com/en-us/windows/win32/sync/slim-reader-writer--srw--locks), and these are not reentrant/recursive. (Another source for msvc shared_mutex being non-recursive is 
https://docs.microsoft.com/en-us/cpp/standard-library/shared-mutex?view=msvc-170 )

There are several paths in `RejitHandler` where a snippet like the following happens:
```
void RejitHandler::SomeMethod() 
{
  ReadLock r_lock(m_shutdown_lock);
  if (m_shutdown) 
  {
     return;
  }
  ...
  // note that r_lock is still in-scope and active here
  OtherRejitHandlerMethod(); // which will also check m_shudown_lock and m_shutdown
}
```
One current example is `AddNGenModule` calling `RequestRejitForInlinersInModule` - after I found 3 of these I stopped counting.

The cleanest fix I see for this is to scope the lock ownership to just what's needed to check shutdown, most likely 
by the addition of a helper method:
```
bool RejitHandler::ShuttingDown()
{
    ReadLock r_lock(m_shutdown_lock);
    return m_shutdown;
}

void RejitHandler::SomeMethod()
{
  if (ShuttingDown())
  {
    // variations exist returning nullptr, false, etc. so this can't easily be a macro
    return;
  }
  ...
  OtherRejitHandlerMethod();
}
```

I'm happy to submit such a PR; just wanted to open a discussion first to see if this was OK or if anybody had a better idea.

johnbley:
Fixes #2366

Use IsShutdownRequested everywhere to avoid recursive std::shared_mutex acquisition.

@DataDog/apm-dotnet

### DIFF
--- a/tracer/src/Datadog.Trace.ClrProfiler.Native/rejit_handler.cpp
+++ b/tracer/src/Datadog.Trace.ClrProfiler.Native/rejit_handler.cpp
@@ -235,8 +235,7 @@ void RejitHandlerModule::RequestRejitForInlinersInModule(ModuleID moduleId)
 
 void RejitHandler::RequestRejitForInlinersInModule(ModuleID moduleId)
 {
-    ReadLock r_lock(m_shutdown_lock);
-    if (m_shutdown)
+    if (IsShutdownRequested())
     {
         return;
     }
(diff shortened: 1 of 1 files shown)


tonyredondo (approved):
LGTM

zacharycmontoya (approved):

PR #2397, opened 2022-02-03
[Gitlab] More explanations on the --grant option used to upload artifacts

pierotibou:
Switch to a comment, it's more verbose and it will work

@DataDog/apm-dotnet

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -6,7 +6,6 @@ stages:
 variables:
   DATADOG_AGENT_WINBUILDIMAGES: v5996510-a9e6a7d
   GIT_PROFILER_REF: master
-  GRANT_PUBLIC_ACCESS_RIGHTS: --grants read=uri=http://acs.amazonaws.com/groups/global/AllUsers full=id=3a6e02b08553fd157ae3fb918945dd1eaae5a1aa818940381ef07a430cf25732
     
 build:
   only:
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -45,12 +44,13 @@ publish:
...
+    - aws s3 cp artifacts/ s3://dd-windowsfilter/builds/tracer/${CI_COMMIT_SHA} --recursive --region us-east-1 --exclude "*" --include "*.zip" --include "*.msi" --grants read=uri=http://acs.amazonaws.com/groups/global/AllUsers full=id=3a6e02b08553fd157ae3fb918945dd1eaae5a1aa818940381ef07a430cf25732
     - |
       if( -not $? )
       {
         $msg = $Error[0].Exception.Message
         Write-Output "Encountered error during while publishing to S3. Error Message is $msg."
         Write-Output "Retrying..."
-        aws s3 cp artifacts/ s3://dd-windowsfilter/builds/tracer/${CI_COMMIT_SHA} --recursive --region us-east-1 --exclude "*" --include "*.zip" --include "*.msi" ${GRANT_PUBLIC_ACCESS_RIGHTS}
+        aws s3 cp artifacts/ s3://dd-windowsfilter/builds/tracer/${CI_COMMIT_SHA} --recursive --region us-east-1 --exclude "*" --include "*.zip" --include "*.msi" --grants read=uri=http://acs.amazonaws.com/groups/global/AllUsers full=id=3a6e02b08553fd157ae3fb918945dd1eaae5a1aa818940381ef07a430cf25732
       }
```
andrewlock, line 56: Unrelated, but would be nice to remove the duplication from this. Maybe something like: 

```powershell
$i = 0
do {
    try {
        # The grants option at the end is used to allow public access on the files we upload as the acls only aren't enough.
        aws s3 cp artifacts/ s3://dd-windowsfilter/builds/tracer/${CI_COMMIT_SHA} --recursive --region us-east-1 --exclude "*" --include "*.zip" --include "*.msi" --grants read=uri=http://acs.amazonaws.com/groups/global/AllUsers full=id=3a6e02b08553fd157ae3fb918945dd1eaae5a1aa818940381ef07a430cf25732
        return
    } catch {
        $msg = $Error[0].Exception.Message
        Write-Output "Encountered error during while publishing to S3. Error Message is $msg."
        Write-Output "Retrying..."
        $i++
        Start-Sleep -Milliseconds 100
    }
} while ($i -lt 3)
```

andrewlock (approved):

.gitlab-ci.yml
pierotibou: Thanks. Done in d72583b4b66b5fbb425fc276575683effe57e0f3
Tried to test it but my 3 attempts didn't fail, as when I tested the first time

PR #2455, opened 2022-02-17
Review Exporter Settings

### LINKED ISSUE #2426: Configuration parsing issue with UDS (DD_AGENT_HOST+DD_TRACE_AGENT_URL)
**Describe the bug**
The tracing using UDS doesn't work when DD_AGENT_HOST and DD_TRACE_AGENT_URL is configured.
Then there won't be any traces submitted and the logfile logs:
```
2022-02-11 12:56:48.626 +00:00 [ERR] An error occurred while sending 1 traces to the agent at unix:///v0.4/traces System.NotSupportedException: The 'unix' scheme is not supported.
   at Datadog.Trace.ClrProfiler.CallTarget.Handlers.Continuations.TaskContinuationGenerator`4.ContinuationAction(Task`1 previousTask, TTarget target, CallTargetState state)
   at Datadog.Trace.ClrProfiler.CallTarget.Handlers.Continuations.TaskContinuationGenerator`4.ContinuationAction(Task`1 previousTask, TTarget target, CallTargetState state)
   at System.Net.Http.HttpClient.<SendAsync>g__Core|83_0(HttpRequestMessage request, HttpCompletionOption completionOption, CancellationTokenSource cts, Boolean disposeCts, CancellationTokenSource pendingRequestsCts, CancellationToken originalCancellationToken)
   at Datadog.Trace.Agent.Transports.HttpClientRequest.PostAsync(ArraySegment`1 bytes, String contentType)
   at Datadog.Trace.Agent.Api.SendTracesAsync(ArraySegment`1 traces, Int32 numberOfTraces, IApiRequest request, Boolean finalTry)
   at Datadog.Trace.Agent.Api.SendTracesAsync(ArraySegment`1 traces, Int32 numberOfTraces)
 { MachineName: ".", Process: "[1 dotnet]", AppDomain: "[1 ConsoleApp]", AssemblyLoadContext: "\"Default\" System.Runtime.Loader.DefaultAssemblyLoadContext #2", TracerVersion: "2.3.0.0" }
```

The root cause seems to be `ExporterSettings.ConfigureTraceTransport`, because it reads the agentHost from DD_AGENT_HOST, then checks if it starts with unix:// and sets `TracesTransportType` to Default. Later on the agentUri will be read from DD_TRACE_AGENT_URL.
Which causes the `TracesTransportStrategy` to return a `HttpClientRequestFactory`, but the configuration is set to a UDS uri.
https://github.com/DataDog/dd-trace-dotnet/blob/57f8232c2eb7837320a6ee8f2a7a4db86e37cbe2/tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs#L231-L303

**To Reproduce**
A repository with a dockerfile to reproduce can be found here: https://github.com/hiiru/dd-net-unix-issue

Basically it prepares a linux container with a .NET app and the tracer with these ENV vars set:
DD_AGENT_HOST=localhost
DD_TRACE_AGENT_URL unix:///var/run/datadog/apm.socket

When DD_AGENT_HOST is set, the NotSupportedException gets thrown.
Otherwise the correct SocketExceptionFactory+ExtendedSocketException gets thrown (since there is no socket in the example docker).

**Expected behavior**
I except that the `TracesTransportType` will be set from DD_TRACE_AGENT_URL and the tracer uses the UDS transfer.
So that it's possible to use UDS for trace and HTTP for Dogstatsd.

pierotibou:
## Summary of changes
I've removed `DD_APM_RECEIVER_SOCKET` as it should not have been added. As it is not documented, we agreed with Lucas that it was ok to break the public API. I've also cleaned the settings to reintroduce `TraceAgentUri` as the first parameter to take into account and make more explicit the order of precedence. Finally, I've removed the TCP Strategy as it wasn't used and as it doesn't seem we want to go further down that road.

## Reason for change
A regression had been introduced as we were not using the Trace Agent Uri as the primary parameter to take into account. We've also seen that we may be able to improve how we configure the transport when Kevin developed the Agent checks in the CLI tool.

## Implementation details
Regarding the settings refactoring:
- I had to introduce a builder class to do build the immutable settings as it seemed clearer (I could have done everything from the .ctor though).
- Removed `DD_APM_RECEIVER_SOCKET`
- Reintroduced `TraceAgentUri` as the first parameter to take into account
- made clearer the order of precedence of the parameters in the code
- handled possible wrongly set parameters that would lead to throwing an exception
- allow falling back on other configuration when fixing the issue would need to restart the app
- moved the settings construction in ImmutableSettings (through a builder, unfortunately this class is public, I couldn't do all the changes I would have liked to do)
- Added a bit of logs to explain the issues
- Removed the TCP strategy

## Testing

Tried to add as much testing as I could by testing both `ExporterSettings`, `ImmutableExporterSettings` checking both the modified values and the ones that should have stayed with their default values.

Fixes #2426 (hopefully:))


tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -6,6 +6,7 @@
 using System;
 using System.IO;
 using Datadog.Trace.Agent;
+using Datadog.Trace.Vendors.Serilog;
```
andrewlock, line 9: Not sure why this has appeared, but it almost certainly shouldn't be here 😉

pierotibou: I thought it's what is used to Log early when we're in the setup phase.  cause it's what is used in `TracerSettings` (cf https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/src/Datadog.Trace/Configuration/TracerSettings.cs#L475). I kinda recalled a discussion saying I shouldn't use a Logger early in the setup, but I haven't checked if that's true.

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -177,59 +185,50 @@ public int PartialFlushMinSpans
...
+            var metricsPipeName = source?.GetString(ConfigurationKeys.MetricsPipeName);
+            var metricsUnixDomainSocketPath = source?.GetString(ConfigurationKeys.MetricsUnixDomainSocketPath);
 
-            MetricsPipeName = source?.GetString(ConfigurationKeys.MetricsPipeName);
-
-            // Agent port is set to zero in places like AAS where it's needed to prevent port conflict.
-            // The agent will fail to start if it can not bind a port.
-            // If the dogstatsd port isn't explicitly configured, check for pipes or sockets.
-            if (!forceMetricsOverUdp && (dogStatsdPort == 0 || dogStatsdPort == null))
+            if ((dogStatsdPort != null && dogStatsdPort != 0) || agentHost != null)
```
lucaspimentel, line 195: If we're ignoring `0`, do we want to ignore negative numbers as well`?
(and out of scope here, but should we log invalid settings?)
```suggestion
            if ((dogStatsdPort != null && dogStatsdPort > 0) || agentHost != null)
```

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -238,81 +237,142 @@ private void ConfigureTraceTransport(IConfigurationSource source, out bool force
...
 #else
                     ?? 500;
+
 #endif
+                SetAgentUriAnyway(agentHost ?? DefaultAgentHost, agentPort ?? DefaultAgentPort); // this one can throw
+                return;
             }
 
-            if (traceTransport == null)
+            if ((agentPort != null && agentPort != 0) || agentHost != null)
```
lucaspimentel, line 273:
```suggestion
            if (agentPort > 0 || agentHost != null)
```

tracer/test/Datadog.Trace.Tests/Configuration/ExporterSettingsTests.cs
```diff
@@ -142,8 +194,8 @@ private ExporterSettings Setup(Func<string, bool> fileExistsMock, params string[
 
             foreach (var item in config)
             {
-                var parts = item.Split(':');
-                configNameValues.Add(parts[0], parts[1]);
+                var separatorIndex = item.IndexOf(':');
```
lucaspimentel, line 197: Since test data (urls!) can clearly have `:` chars, do you think we should switch this to use `=` instead? Or better yet, use separate strings so we don't need to split them as part of the test?

```csharp
private ExporterSettings Setup(Func<string, bool> fileExistsMock, params (string Key, string Value)[] config)
```

This could have led to subtle bugs in the tests. Something like `A:B:C` which was meant to be `A` and `B:C` would have been interpreted as `A` and `B` (and ignored `C`). A test could have passed by accident, for example, giving us false confidence.

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
pierotibou: In the current state of this PR, negative ports will be logged indirectly as we will log a warning when trying to create the URL.
But you're right maybe I shall start by checking all variables, see if they are valid, log and ignore them if not valid 🤔

tracer/test/Datadog.Trace.Tests/Configuration/ExporterSettingsTests.cs
pierotibou: > Since test data (urls!) can clearly have : chars

Yes that's why I made this change, the previous implementation didn't work. 
But I considered it was enough as I don't see us putting a `:` in the key name in the tests. We're not that 😈 . 
But I'll look into using your proposal, as it's even simpler

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
pierotibou: Fixed with latest commits

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -238,81 +237,142 @@ private void ConfigureTraceTransport(IConfigurationSource source, out bool force
...
+                TracesTransport = TracesTransportType.WindowsNamedPipe;
+                TracesPipeName = tracesPipeName;
                 TracesPipeTimeoutMs = source?.GetInt32(ConfigurationKeys.TracesPipeTimeoutMs)
 #if DEBUG
-                    ?? 20_000;
+                                   ?? 20_000;
 #else
                     ?? 500;
+
 #endif
```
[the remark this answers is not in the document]
pierotibou: Removed it for now, I shall get back to this one later

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
pierotibou: Fixed in 96dda0983f4eb26e0bc12bb0078c855df838b7c4. 
Switched to the same system you used in DirectLogsSubmission. ie storing the data in a list appended in the diagnosis log.

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -66,37 +59,50 @@ public ExporterSettings()
...
+                }
+                else
+                {
+                    AgentUri = uri;
+                }
+            }
 
-            ConfigureTraceTransport(source, out var shouldUseUdpForMetrics);
-            ConfigureMetricsTransport(source, shouldUseUdpForMetrics);
+            TracesUnixDomainSocketPath = source?.GetString(ConfigurationKeys.TracesUnixDomainSocketPath);
```
lucaspimentel, line 80: I just noticed while looking at the documentation PR (https://github.com/DataDog/documentation/pull/13131) that this setting maps to `DD_APM_RECEIVER_SOCKET`. That is an Agent setting and shouldn't be a Tracer settings. I don't see anything about using `DD_APM_RECEIVER_SOCKET` in the RFC, it doesn't follow the `DD_TRACE_*` convention, and other Tracers seem to only use `DD_TRACE_AGENT_URL` to configure Unix Domain Sockets.

If it makes the logic here easier, we should consider removing this setting. Otherwise, we should at least rename it (and I guess keep supporting the older name if anyone is already using it).

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -66,37 +59,50 @@ public ExporterSettings()
...
+            MetricsPipeName = source?.GetString(ConfigurationKeys.MetricsPipeName);
+            MetricsUnixDomainSocketPath = source?.GetString(ConfigurationKeys.MetricsUnixDomainSocketPath);
 
-            if ((partialFlushMinSpans ?? 0) <= 0)
+            PartialFlushEnabled = source?.GetBool(ConfigurationKeys.PartialFlushEnabled) ?? false;
+            var partialFlushMinSpans = source?.GetInt32(ConfigurationKeys.PartialFlushMinSpans);
+            if (partialFlushMinSpans > 0)
             {
-                partialFlushMinSpans = 500;
+                PartialFlushMinSpans = partialFlushMinSpans.Value;
```
lucaspimentel, line 100: Looks like the default of 500 was lost?

pierotibou: default values are set in ImmutableExporterSetting now ie if it is set to 0 there, we default to 500

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
pierotibou: Fixed in e445c9959f19dcc8ec0607a401153d98f9803479

lucaspimentel (approved):

tracer/test/Datadog.Trace.Tests/Snapshots/PublicApiTests.PublicApiHasNotChanged.verified.txt
```diff
@@ -33,7 +33,6 @@ namespace Datadog.Trace.Configuration
         public int PartialFlushMinSpans { get; set; }
         public string TracesPipeName { get; set; }
         public int TracesPipeTimeoutMs { get; set; }
-        public string TracesUnixDomainSocketPath { get; set; }
```
andrewlock, line 36: This shouldn't have gone I don't think 🤔

zacharycmontoya: This is correct, `TracesUnixDomainSocketPath` got removed from `ExporterSettings` here but it is still present on the `ImmutableExporterSettings` type if you expand below

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -96,14 +97,21 @@ internal ExporterSettings(IConfigurationSource source, Func<string, bool> fileEx
...
         /// <seealso cref="ConfigurationKeys.AgentHost"/>
         /// <seealso cref="ConfigurationKeys.AgentPort"/>
         public Uri AgentUri { get; set; }
 
+        /// <summary    >
+        /// Gets or sets the agent host that can be used to reach dogStatsD
+        /// Default is <c>"localhost"</c>.
+        /// </summary>
+        /// <seealso cref="ConfigurationKeys.AgentHost"/>
+        public string AgentHost { get; set; }
```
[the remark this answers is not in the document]
zacharycmontoya: It looks like this property is also used to calculate the traces endpoint when both `AgentUri` and `TracesPipeName` are not set. If my understanding is correct, then it still encompasses both metrics and traces

tracer/test/Datadog.Trace.Tests/TracerSettingsTests.cs
```diff
@@ -99,7 +99,7 @@ public void ReplaceLocalhost(string original, string expected)
                 { ConfigurationKeys.AgentUri, original }
             };
 
-            var tracerSettings = new TracerSettings(new NameValueConfigurationSource(settings));
+            var tracerSettings = new ImmutableTracerSettings(new TracerSettings(new NameValueConfigurationSource(settings)));
```
zacharycmontoya, line 102: I'm a little on the fence, but I feel like this test belongs in `ImmutableExporterSettingsTests` since that is where the replacement of localhost occurs

tracer/test/Datadog.Trace.Tests/Configuration/ImmutableExporterSettingsTests.cs
```diff
@@ -61,46 +62,281 @@ public void HasSamePropertiesAsExporterSettings()
...
+            var settings = new ExporterSettings { AgentHost = "someotherhost" };
+            var immutableExporterSettings = Setup(settings);
+
+            AssertHttpIsConfigured(immutableExporterSettings, "http://someotherhost:8126");
+        }
+
+        [Fact]
+        public void Traces_SocketFilesExist_ExplicitTraceAgentPort_UsesDefaultHttp()
+        {
+            // AgentPort isn't settable directly, so goinng a config source as a user would
```
zacharycmontoya, line 116:
```suggestion
            // AgentPort isn't settable directly, so setting via a config source as a user would
```

tracer/test/Datadog.Trace.Tests/Configuration/ImmutableExporterSettingsTests.cs
```diff
@@ -61,46 +62,281 @@ public void HasSamePropertiesAsExporterSettings()
...
+            CheckDefaultValues(immutableExporterSettings, "PartialFlushEnabled", "PartialFlushMinSpans");
+        }
+
+        [Fact]
+        public void Metrics_SocketFilesExist_NoExplicitConfig_UsesMetricsSocket()
+        {
+            var immutableExporterSettings = Setup(FileExistsMock(ExporterSettings.DefaultMetricsUnixDomainSocket), new ExporterSettings());
+            Assert.Equal(expected: MetricsTransportType.UDS, actual: immutableExporterSettings.MetricsTransport);
+            Assert.Equal(expected: ExporterSettings.DefaultMetricsUnixDomainSocket, actual: immutableExporterSettings.MetricsUnixDomainSocketPath);
+            CheckDefaultValues(immutableExporterSettings, "MetricsTransport", "MetricsUnixDomainSocketPath", "DogStatsdPort");
```
zacharycmontoya, line 189: Should we assert DogStatsdPort to be 0 in this instance?

zacharycmontoya (approved):
LGTM with some minor comments

tracer/test/Datadog.Trace.Tests/Snapshots/PublicApiTests.PublicApiHasNotChanged.verified.txt
pierotibou: This was removed intentionlaly as it had been added only for .NET. (cf PR description)
This is the reason why Lucas rightfully asked me to document these settings (cf https://github.com/DataDog/architecture/pull/1000)

PR #2227, opened 2021-12-23
Disable parallelization in runtime metrics tests

kevingosse:
This is done to reduce the probability of running into https://github.com/dotnet/runtime/issues/51579

If it still happens despite this mitigation, then we'll have to disable this test on 3.1.

PR #2457, opened 2022-02-17
Disable runtime metrics tests for netcoreapp 3.1

kevingosse:
It looks like https://github.com/DataDog/dd-trace-dotnet/pull/2227 wasn't enough, the tests are still occasionally deadlocking.

### DIFF
--- a/tracer/test/Datadog.Trace.Tests/RuntimeMetrics/RuntimeEventListenerTests.cs
+++ b/tracer/test/Datadog.Trace.Tests/RuntimeMetrics/RuntimeEventListenerTests.cs
@@ -3,6 +3,7 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
 // </copyright>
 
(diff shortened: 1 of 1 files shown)


andrewlock (approved):
❤️ Fighting the good fight against CI

tracer/test/Datadog.Trace.Tests/RuntimeMetrics/RuntimeEventListenerTests.cs
```diff
@@ -3,7 +3,7 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
 // </copyright>
 
-#if NETCOREAPP3_1 || NET5_0
+#if NET5_0_OR_GREATER
```
pierotibou, line 6:
```suggestion
// Disabled on .NET Core 3.1 as we were running into this issue: https://github.com/dotnet/runtime/issues/51579
#if NET5_0_OR_GREATER
```

pierotibou (approved):
Maybe add a comment for posterity.

PR #2461, opened 2022-02-17
Make check iis command more tolerant

kevingosse:
Before this change, arguments must be wrapped in quotes when they have spaces in them. For instance:

```
dd-trace check iis "Default Web Site/WebApplication1"
```

It's likely that many customers are running apps in the default web site, so this change preventively makes the syntax more tolerant. This is now valid:

```
dd-trace check iis Default Web Site/WebApplication1
````

### DIFF
--- a/tracer/test/Datadog.Trace.Tools.Runner.IntegrationTests/Checks/IisCheckTests.cs
+++ b/tracer/test/Datadog.Trace.Tools.Runner.IntegrationTests/Checks/IisCheckTests.cs
@@ -54,7 +54,7 @@ namespace Datadog.Trace.Tools.Runner.IntegrationTests.Checks
 
                 using var console = ConsoleHelper.Redirect();
 
-                var result = await CheckIisCommand.ExecuteAsync(new CheckIisSettings { SiteName = siteName }, _iisFixture.IisExpress.ConfigFile, _iisFixture.IisExpress.Process.Id);
+                var result = await CheckIisCommand.ExecuteAsync(new CheckIisSettings { SiteName = new[] { siteName } }, _iisFixture.IisExpress.ConfigFile, _iisFixture.IisExpress.Process.Id);
 
                 result.Should().Be(0);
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by andrewlock — COMMENTED** [review]
My concern with this is that it means we can't add other arguments later. I'd rather we flagged the command as an error if they provided "additional arguments". Can we do that?

**Comment by kevingosse** [comment]
> My concern with this is that it means we can't add other arguments later. I'd rather we flagged the command as an error if they provided "additional arguments". Can we do that?

It doesn't prevent us from adding arguments of type `--flag true`, like for the run command.

But I can display an error message instead. I still have to change the argument type for `string[]` otherwise Spectre.Console doesn't match it at all and executes the "old syntax" fallback, but I can show an error message if there's more than one element in the array.

**Comment by kevingosse** [comment]
> Means we can "safely" change the behaviour later?

Well we can always change the behavior later. Our hands are tied for the `run` and `ci` command because changing them might break something in a script or a CI, but I don't think we have such constraints with the `check` command.

File: tracer/test/Datadog.Trace.Tools.Runner.IntegrationTests/Checks/IisCheckTests.cs [hunk 1]
```diff
@@ -124,7 +124,7 @@ public async Task ListApplications()
 
             using var console = ConsoleHelper.Redirect();
 
-            var result = await CheckIisCommand.ExecuteAsync(new CheckIisSettings { SiteName = "sample/dummy" }, _iisFixture.IisExpress.ConfigFile, _iisFixture.IisExpress.Process.Id);
+            var result = await CheckIisCommand.ExecuteAsync(new CheckIisSettings { SiteName = new[] { "sample/dummy" } }, _iisFixture.IisExpress.ConfigFile, _iisFixture.IisExpress.Process.Id);
```
andrewlock, line 127: Worth adding a check for the site name joining, e.g. `new[] { "sample", "dummy" }` I'm undecided if it's worth it 🤷‍♂️

andrewlock (approved):

kevingosse:
Abandoned in favor of https://github.com/DataDog/dd-trace-dotnet/pull/2463

PR #2476, opened 2022-02-20
Use mermaid for flow chart

andrewlock:
## Summary of changes

Convert flow chart to mermaid

## Reason for change

Whimsy.

## Implementation details

GitHub supports mermaidjs now

## Other details

image


kevingosse (approved):
Nice. I wonder if it would look even better if you merged the `result1` and `result2` states.

andrewlock:
> Nice. I wonder if it would look even better if you merged the `result1` and `result2` states.

Yeah, I think so: 
image

PR #2537, opened 2022-03-07
[Appsec]Benchmarks

anna-git:

### DIFF
--- a/tracer/build/_build/Build.Steps.cs
+++ b/tracer/build/_build/Build.Steps.cs
@@ -324,36 +324,51 @@ partial class Build
         .Executes(() =>
         {
             var project = Solution.GetProject(Projects.AppSecUnitTests);
-            var directory = project.Directory;
-            var targetFrameworks = project.GetTargetFrameworks();
-            if (IsWin)
-            {
-                foreach (var architecture in WafWindowsArchitectureFolders)
-                {
-                    CopyWaf(architecture, targetFrameworks, directory, "ddwaf", "dll");
-                }
-            }
-            else
-            {
-                var (architecture, ext) = GetUnixArchitectureAndExtension();
-                CopyWaf(architecture, targetFrameworks, directory, "libddwaf", ext);
-            }
-
-            void CopyWaf(string architecture, IEnumerable<string> frameworks, AbsolutePath absolutePath, string wafFileName, string extension)
-            {
-                var source = LibDdwafDirectory / "runtimes" / architecture / "native" / $"{wafFileName}.{extension}";
-                var nativeDir = DDTracerHomeDirectory / architecture / $"Datadog.Trace.ClrProfiler.Native.{extension}";
-                foreach (var fmk in frameworks)
-                {
-                    var dest = absolutePath / "bin" / BuildConfiguration / fmk / architecture;
-                    CopyFileToDirectory(source, dest, FileExistsPolicy.Overwrite);
-                    if (!IsWin)
-                    {
-                        CopyFileToDirectory(nativeDir, absolutePath / "bin" / BuildConfiguration / fmk, FileExistsPolicy.Overwrite);
-                    }
-                }
-            }
+            CopyWafForProject(project);
         });
+
+    Target CopyLibDdwafForBenchmarks => _ => _
+       .Unlisted()
+       .After(Clean)
+       .After(DownloadLibDdwaf)
+       .Executes(() =>
+       {
+           var project = Solution.GetProject(Projects.BenchmarksTrace);
+           CopyWafForProject(project);
+       });
+
+    private void CopyWafForProject(Project project)
+    {
+        var directory = project.Directory;
+        var targetFrameworks = project.GetTargetFrameworks();
+        if (IsWin)
+        {
+            foreach (var architecture in WafWindowsArchitectureFolders)
+            {
+                CopyWaf(architecture, targetFrameworks, directory, "ddwaf", "dll");
+            }
+        }
+        else
+        {
+            var (architecture, ext) = GetUnixArchitectureAndExtension();
+            CopyWaf(architecture, targetFrameworks, directory, "libddwaf", ext);
+        }
+
+        void CopyWaf(string architecture, IEnumerable<string> frameworks, AbsolutePath absolutePath, string wafFileName, string extension)
+        {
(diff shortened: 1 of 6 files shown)


tracer/build/_build/Build.Steps.cs
```diff
@@ -1231,7 +1246,7 @@ void CopyWaf(string architecture, IEnumerable<string> frameworks, AbsolutePath a
         .Executes(() =>
         {
             // Build the actual integration test projects for Any CPU
-            var integrationTestProjects = TracerDirectory.GlobFiles("test/*.IntegrationTests/*.csproj");
+            var integrationTestProjects = TracerDirectory.GlobFiles("test/*.Security.IntegrationTests/*.csproj");
```
robertpi, line 1249: Looks like an error.

tracer/test/benchmarks/Benchmarks.Trace/datadog.json
```diff
@@ -0,0 +1,5 @@
+{
+  "DD_APPSEC_ENABLED": true,
+  "DD_DOTNET_TRACER_HOME": "C:\Repositories\dd-trace-dotnet\tracer\bin\dd-tracer-home\win-x64",
```
andrewlock, line 3: Should this really be hard-coded? 🤔

tracer/test/benchmarks/Benchmarks.Trace/AppSecBodyBenchmark.cs
```diff
@@ -0,0 +1,247 @@
...
+#if !NETFRAMEWORK
+    public class HttpContextMock : HttpContext
+    {
+        private readonly HttpResponseMock responseMock;
+
+        public HttpContextMock(HttpResponseMock responseMock)
+        {
+            this.responseMock = responseMock;
+        }
+        public override ConnectionInfo Connection => throw new NotImplementedException();
```
andrewlock, line 118: I'm not sure all this is necessary? I think you can use `DefaultHttpContext`?

tracer/test/benchmarks/Benchmarks.Trace/AppSecBodyBenchmark.cs
```diff
@@ -0,0 +1,247 @@
...
+                    new Dog { Name = "toto", Dogs = new List<Dog> { new Dog { Name = "tata" }, new Dog { Name = "tata" } } },
+                    new Dog { Name = "tata", Dogs = new List<Dog> { new Dog { Name = "titi" }, new Dog { Name = "titi" }, new Dog { Name = "tutu" } } }
+                    }
+        };
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            security = Security.Instance;
+        }
```
andrewlock, line 71: nit: rather than using `GlobalSetup` we typically use a static constructor. Nothing wrong with this AFAIK, just wondered if there was a specific reason for this approach?

tracer/test/benchmarks/Benchmarks.Trace/AppSecBodyBenchmark.cs
```diff
@@ -0,0 +1,247 @@
...
+            Debugger.Break();
+            using var ms = new MemoryStream();
+            using var sw = new StreamWriter(ms);
+            var httpContextMock = new HttpContext(new HttpRequest(string.Empty, string.Empty, string.Empty), new HttpResponse(sw));
+            security.InstrumentationGateway.RaiseBodyAvailable(httpContextMock, new Span(new SpanContext(1, 1), DateTimeOffset.UtcNow), new { });
+#else
+            var httpContextMock = new HttpContextMock(new HttpResponseMock());
+            security.InstrumentationGateway.RaiseBodyAvailable(httpContextMock, new Span(new SpanContext(1, 1), DateTimeOffset.UtcNow), new { });
+#endif
+        }
```
andrewlock, line 86: It would be nice to be able to compare these to benchmarks where security is disabled, do we do that currently?

tracer/test/benchmarks/Benchmarks.Trace/AppSecBodyBenchmark.cs
```diff
@@ -0,0 +1,247 @@
...
+
+        [Benchmark]
+        public void AllCycleMoreComplexBody()
+        {
+#if NETFRAMEWORK
+            using var ms = new MemoryStream();
+            using var sw = new StreamWriter(ms);
+            var httpContextMock = new HttpContext(new HttpRequest(string.Empty, string.Empty, string.Empty), new HttpResponse(sw));
```
andrewlock, line 94: I wonder if this setup should be outside the benchmark itself? Probably not necessary though.

andrewlock (approved):
Just some minor suggestions/questions

tracer/test/benchmarks/Benchmarks.Trace/AppSecBodyBenchmark.cs
anna-git: there was no reason , as I had problems initializing Security, I tried this and left it 😅

tracer/test/benchmarks/Benchmarks.Trace/AppSecBodyBenchmark.cs
anna-git: thanks! Fixed  in 8eda4fec8c9b51c954871e7bc66197dfb77ec7d9, resources are then disposed after the execution

tracer/test/benchmarks/Benchmarks.Trace/AppSecBodyBenchmark.cs
anna-git: we have that in the crank but this was more to benchmark the body, and in particular, the body reflection algorithm.

tracer/test/benchmarks/Benchmarks.Trace/AppSecBodyBenchmark.cs
anna-git: fixed in https://github.com/DataDog/dd-trace-dotnet/pull/2537/commits/2dd558752b586980c3b56b32ef30423f75f009c1

tracer/test/benchmarks/Benchmarks.Trace/AppSecBodyBenchmark.cs
anna-git: yes I wasnt sure how to make sure the response was disposing the waf pointers, but with disposing them manually now, no need for all this anymore thanks

PR #2568, opened 2022-03-14
Upgrade Nuke version

andrewlock:
## Summary of changes
- Upgrades Nuke to 5.3.0

## Reason for change

Maintenance, adds built-in support for VS2022, could use project source generator (found this to be a bit flaky though, so haven't used it)

## Implementation details

I haven't updated to 6.x as this caused weird build issues, and we don't _need_ it


andrewlock:
Pffft, it fails on linux for some reason, probably not worth the effort

PR #2570, opened 2022-03-14
[Profiler] Use smart_pointer to manage services lifetime

gleocadie:
## Summary of changes

Use smart pointers to automatically managed services lifecycle.

## Reason for change

Today we use new/delete statements to allocate and deallocate memory. This can be problematic and can lead to memory leak
Instead we should rely on smart pointers.

## Implementation details

Use smart pointers for the services and fix few programming bugs.

## Test coverage

* ASAN tests will detect misusage
* Current integration tests will check if it's still working


chrisnas (approved):
LGTM
As seen together (checked the public/private destructor)

PR #2573, opened 2022-03-15
Avoid WebRequest deadlock in Telemetry

tonyredondo:
In some scenarios the Telemetry worker triggers a deadlock when using the HttpWebRequest implementation (.NET Framework) on process shutdown.

The deadlock appear to be a bug in the HttpWebRequest implementation according to: https://social.msdn.microsoft.com/Forums/en-US/2ab06a91-c8fd-474c-81ea-61b78c8fbb56/httpwebrequestgetrequeststream-hangs-bug-in-systemnet-stack-?forum=ncl

To avoid the deadlock, the `JsonWebRequestTelemetryTransport` implementation changed from Async to Sync  and wrapped by a `lock` mechanism, as suggested by the forum response:

image


tracer/src/Datadog.Trace/LifetimeManager.cs
```diff
@@ -75,12 +76,14 @@ private void CurrentDomain_UnhandledException(object sender, UnhandledExceptionE
 
         private void Console_CancelKeyPress(object sender, ConsoleCancelEventArgs e)
         {
+            Log.Debug("Running Shutdown tasks due CancelKeyPress");
```
andrewlock, line 79:
```suggestion
            Log.Debug("Running Shutdown tasks due to CancelKeyPress");
```

tracer/src/Datadog.Trace/LifetimeManager.cs
```diff
@@ -75,12 +76,14 @@ private void CurrentDomain_UnhandledException(object sender, UnhandledExceptionE
...
             RunShutdownTasks();
             Console.CancelKeyPress -= Console_CancelKeyPress;
         }
 
         private void CurrentDomain_DomainUnload(object sender, EventArgs e)
         {
+            Log.Debug("Running Shutdown tasks due DomainUnload");
```
andrewlock, line 86:
```suggestion
            Log.Debug("Running Shutdown tasks due to DomainUnload");
```

tracer/src/Datadog.Trace/TracerManager.cs
```diff
@@ -439,17 +439,49 @@ private static void OneTimeSetup()
...
+                    _heartbeatTimer?.Dispose();
+
+                    var instance = _instance;
+                    if (instance is not null)
+                    {
+                        Log.Debug("Shutdown agent writer.");
+                        await _instance.AgentWriter.FlushAndCloseAsync().ConfigureAwait(false);
+
+                        Log.Debug("Shutdown direct log submission.");
+                        _instance.DirectLogSubmission?.Dispose();
```
andrewlock, line 471: I wonder if there's a similar path in here we need to make async?

tonyredondo:
The deadlock issue we have seen was not related to the forum post, but for missing the disposal of every telemetry response.

So, this PR has been superseeded by https://github.com/DataDog/dd-trace-dotnet/pull/2574

PR #2611, opened 2022-03-29
Adds additional required Visual Studio components to tracer-ReadMe

cwe1ss:
## Summary of changes
Adds missing .NET targeting packs and the VS 2019 build tools to the "minimum requirements"-section of the tracer ReadMe.

## Reason for change
When trying to open and build the tracer folder, I received several errors due to missing VS 2022 components that aren't listed in the current ReadMe. Adding these to the ReadMe would simplify the initial setup for contributors.

The component `MSVC v142 – VS 2019 C++ x64/x86 build tools (v14.29-16.11)` was necessary to fix this error when calling `.\build.cmd Clean BuildTracerHome`:
```
"C:\src\cwe1ss\dd-trace-dotnet\tracer\src\Datadog.Trace.ClrProfiler.Native\Datadog.Trace.ClrProfiler.Native.vcxproj" (Standardziel) (4) ->
  C:\Program Files\Microsoft Visual Studio\2022\Professional\MSBuild\Microsoft\VC\v170\Microsoft.CppBuild.targets(440,5): error MSB8020: The build tools for Visual Studio 2019 (Platform Toolset = 'v142') cannot be found. To build using the v142 build tools, please install Visual Studio 2019 build tools.  Alternatively, you may upgrade to the current Visual Studio tools by selecting the Project menu or right-click the solution, and then selecting "Retarget solution". [C:\src\cwe1ss\dd-trace-dotnet\tracer\src\Datadog.Trace.ClrProfiler.Native\Datadog.Trace.ClrProfiler.Native.vcxproj]
```

The .NET targeting packs were required to fix the following errors:

image

image

## Implementation details
N/A

## Test coverage
N/A

## Other details
N/A


andrewlock:
Thanks @cwe1ss! Theoretically, the [`.vsconfig`](https://github.com/DataDog/dd-trace-dotnet/blob/master/.vsconfig) file is _supposed_ to prompt you to install all these automatically when you open the solution in Visual Studio. Just wondering if:

1. It didn't prompt you to install these in Visual Studio (i.e. it ignored the .vsconfig)
2. It prompted you, but it wasn't obvious that this was required (our docs don't mention it)
3. You weren't using Visual Studio initially

I use Rider, and already have the components installed, so I'm not _entirely_ sure how this works in 2022 😄 We _may_ want to update the Readme to prompt you to to use the `.vsconfig` directly, rather than listing the components in both places 🙂

cwe1ss:
Thanks for the reply, Andrew! It was a mix of 2) and 3). 

I went straight for the `.\build.cmd` commands first. I later opened Visual Studio and got an informational message in my solution explorer to install some things. I didn't know about this `.vsconfig`-feature though, so I thought this was just some built-in "guess" by VS as it suggested some things that I had already installed separately (Git, .NET 3.1 runtime, .NET 5.0 runtime). I therefore just manually installed the necessary components.

I now went through this prompt again and let VS install everything that was suggested. I guess it suggested the .NET 3.1 runtime & .NET 5.0 runtime for me, because I didn't have the most recent versions installed.

This is what it looked like (sorry for it being in German, couldn't see a quick way to change the language in the installer):
image

Given this `.vsconfig`-feature exists I also think it would be better to just mention this in the readme instead of listing the components again! Do you want me to update the PR with a proposal?

cwe1ss:
@andrewlock I've updated the text. Just let me know if you'd like any further changes, or feel free to modify it however you see fit!

andrewlock (approved):
Thank @cwe1ss, looks good to me! 🙂

cwe1ss:
Regarding the failing add-labels workflow: Could it be, that the "allFilesIn"-match in `labeller.yml` is case-sensitive? The file I've changed is called README.MD (uppercase MD) and labeller.yml currently contains this:
```yaml
  - name: "area:docs"
    allFilesIn: "docs\/.*|.*README.*|.*\\.md"
```

If you want, I can rename README.MD to README.md - having uppercase MD is unusual anyway AFAIK.

cwe1ss:
I looked into this in more detail:
* The labeller did match `tracer/README.MD` for `.*README.*` so it correctly applied `area:docs`
  * https://github.com/DataDog/dd-trace-dotnet/runs/5993959170?check_suite_focus=true#step:4:37
* If it weren't named **README**.MD, it would NOT have matched `.*\\.md` (it's case sensitive)
* The error seems to be a permission error [when updating the PR](https://github.com/DataDog/dd-trace-dotnet/blob/a210ddef724906ff8b93e569c48eaea212daa8a6/tracer/build/_build/Build.GitHub.cs#L113-L117) (which currently is invoked even if nothing has changed). The token seems to not have write-permissions if the PR is coming from a fork: 
  * https://github.com/actions/first-interaction/issues/10
  * A comment suggests that you can explicitly set write permissions: https://github.com/actions/first-interaction/issues/10#issuecomment-1041402989

Hope this helps! 

(And sorry again for creating the test-PR before. I created it in my fork but GitHub changed the target to upstream and I missed that)

pierotibou:
Thanks a lot for the investigation on the labeler @cwe1ss. It's something I've just added, and it seems it needs a bit more work indeed :)

pierotibou (approved):

PR #2607, opened 2022-03-28
[Release] Remove Tracer MSI and Improve release notes categorization

pierotibou:
## Summary of changes
I added a new split in the release notes, in order to categorize by components (ie Tracer, Ci App..., by historical order). It's not perfect as it mixes fixes and changes when there is a component, but I wanted to get feedback before doing more complicated.
Doing so, I realized that we still had the Tracer Msi in the code base and that we needed to update the release processe with a few last changes.

## Reason for change
Last release notes were loooonnnngggg. I really think we need a better categorization so I rewrote it manually: https://github.com/DataDog/dd-trace-dotnet/releases/tag/v2.5.1

## Implementation details
The categorization relies on the label that is present on the PR.
Tracer msi removed and 'UpdateMsiContent' target reworked

## Test coverage
None

## Other details
Not sure I'm not missing something for the profiler-beta msi :)

PR #2212, opened 2021-12-21
add GitHub action to automatically label PRs based on file paths

lucaspimentel:
> Automatically label new pull requests based on the paths of files being changed.

https://github.com/actions/labeler#pull-request-labeler

PR #2662, opened 2022-04-09
Add simplistic labeller

pierotibou:
## Summary of changes
Adding a simplistic labeler that allows adding label when
- PR title matches a string (eg `[CIApp]` or `CIApp`)
- All modified files path matches a string (eg `profiler/*`)

Currently, this allows to add `CIApp`, `Serverless`, `AppSec`, `Profiler` and `Debugger` labels easily.
In a second commit, I've also added more labels taken from Lucas' PR (https://github.com/DataDog/dd-trace-dotnet/pull/2212) and in a third, modified the release notes to contain Serverless as well

## Reason for change
Following https://github.com/DataDog/dd-trace-dotnet/pull/2607, we now categorize release notes based on labels, using the `area-{component}` label. So we thought we could revisit having a labeler. I spent a bit of time trying to find one already available but they do not match our needs (matching PR description, matching all files, not removing labels). So I decided to implement a simplistic one as we already have all the tools to do so.

## Implementation details
Added an action, its configuration and a Nuke target.

## Test coverage
Tested on my fork (mainly the first commit)

## Next steps
We could imagine also:
- Adding the `type:bug` when the PR summary `Fixes` section is filled
- Split regexp on multiple lines for readability
- Discuss with other tracer team to see if this target could interest them and make it a real action?

### DIFF
--- a/.github/labeller.yml
+++ b/.github/labeller.yml
@@ -1,0 +1,21 @@
+labels:
+  - name: "area-ciapp"
+    title: "^\\[?(?i)ciapp"
+
+  - name: "area-serverless"
+    title: "^\\[?(?i)serverless"
+
+  - name: "area-appsec"
+    title: "^\\[?(?i)appsec"
+
+  - name: "area-profiler"
+    title: "^\\[?(?i)profiler"
+
+  - name: "area-debugger"
+    title: "^\\[?(?i)debugger"
+
+  - name: "area-profiler"
+    allFilesIn: "profiler\/.*"
+
+  - name: "area-debugger"
+    allFilesIn: "debugger\/.*" 
--- a/.github/labeller.yml
+++ b/.github/labeller.yml
@@ -39,7 +39,7 @@ labels:
     allFilesIn: ".*OpenTracing.*"
 
   - name: "area:shared-components"
-    allFilesIn: "shared\/.*"
+    allFilesIn: "(?!.*msi-installer.*)(shared\/.*)"
 
   - name: "area:tests"
     allFilesIn: "tracer/test/.*"
--- a/.github/labeller.yml
+++ b/.github/labeller.yml
@@ -33,7 +33,7 @@ labels:
     allFilesIn: "shared\/src\/msi-installer\/.*"
 
   - name: "area:integrations"
-    allFilesIn: "tracer\/src\/Datadog.Trace\/ClrProfiler\/.*|tracer\/src\/Datadog\\.Trace\\.AspNet\/.*|tracer\/test\/Datadog\\.Trace\\.ClrProfiler\\.IntegrationTests\/.*|tracer\/test\/Datadog\\.Trace\\.ClrProfiler\\.Managed\\.Tests\/.*"
+    allFilesIn: "tracer\/src\/Datadog\\.Trace\/ClrProfiler\/AutoInstrumentation\/.*|tracer\/src\/Datadog\\.Trace\\.AspNet\/.*|tracer\/test\/.*"
 
   - name: "area:opentracing"
     allFilesIn: ".*OpenTracing.*"
@@ -43,9 +43,6 @@ labels:
 
   - name: "area:tests"
     allFilesIn: "tracer/test/.*"
-
-  - name: "area:test-apps"
-    allFilesIn: ".*\/test-applications\/.*"
 
   # if any modified file is in tracer/src/Datadog.Trace, excluding files in ClrProfiler, AppSec or Ci
   - name: "area:tracer"
--- a/.github/labeller.yml
+++ b/.github/labeller.yml
@@ -18,4 +18,35 @@ labels:
     allFilesIn: "profiler\/.*"
 
   - name: "area-debugger"
-    allFilesIn: "debugger\/.*" 
+    allFilesIn: "debugger\/.*"
+
+  - name: "area:builds"
+    allFilesIn: "\\.github\/.*|tracer\/build\/.*|\\.azure-pipelines\/.*|gitlab-ci\\.yml"
+
+  - name: "area:vendors"
+    allFilesIn: "tracer\/src\/Datadog\\.Trace\/vendors\/.*"
+
+  - name: "area:docs"
+    allFilesIn: "docs\/.*|.*README.*|.*\\.md"
+
+  - name: "area:installers"
+    allFilesIn: "shared\/src\/msi-installer\/.*"
+
+  - name: "area:integrations"
+    allFilesIn: "tracer\/src\/Datadog.Trace\/ClrProfiler\/.*|tracer\/src\/Datadog\\.Trace\\.AspNet\/.*|tracer\/test\/Datadog\\.Trace\\.ClrProfiler\\.IntegrationTests\/.*|tracer\/test\/Datadog\\.Trace\\.ClrProfiler\\.Managed\\.Tests\/.*"
+
+  - name: "area:opentracing"
+    allFilesIn: ".*OpenTracing.*"
+
+  - name: "area:shared-components"
+    allFilesIn: "shared\/.*"
+
+  - name: "area:tests"
+    allFilesIn: "tracer/test/.*"
+
+  - name: "area:test-apps"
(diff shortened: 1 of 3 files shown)


.github/labeller.yml
```diff
@@ -0,0 +1,52 @@
...
+    title: "^\\[?(?i)debugger"
+
+  - name: "area-profiler"
+    allFilesIn: "profiler\/.*"
+
+  - name: "area-debugger"
+    allFilesIn: "debugger\/.*"
+
+  - name: "area:builds"
+    allFilesIn: "\\.github\/.*|tracer\/build\/.*|\\.azure-pipelines\/.*|gitlab-ci\\.yml"
```
andrewlock, line 24: We should remember to update this label if/when we move Nuke to be top-level

.github/labeller.yml
```diff
@@ -0,0 +1,52 @@
...
+
+  - name: "area:vendors"
+    allFilesIn: "tracer\/src\/Datadog\\.Trace\/vendors\/.*"
```
andrewlock, line 27: Does the profiler have any vendored code? I couldn't see any but seems likely 🤔

.github/labeller.yml
```diff
@@ -0,0 +1,52 @@
...
+
+  - name: "area:docs"
+    allFilesIn: "docs\/.*|.*README.*|.*\\.md"
+
+  - name: "area:installers"
+    allFilesIn: "shared\/src\/msi-installer\/.*"
+
+  - name: "area:integrations"
+    allFilesIn: "tracer\/src\/Datadog.Trace\/ClrProfiler\/.*|tracer\/src\/Datadog\\.Trace\\.AspNet\/.*|tracer\/test\/Datadog\\.Trace\\.ClrProfiler\\.IntegrationTests\/.*|tracer\/test\/Datadog\\.Trace\\.ClrProfiler\\.Managed\\.Tests\/.*"
```
andrewlock, line 36: Not sure about this one exactly. ClrProfiler (for example) contains stuff not specific to "integrations" per-se, and is more of a generalised "tracer" folder, but _sometimes_ contains stuff related to integrations. I would probably limit it to Datadog.Trace.ClrProfiler.AutoInstrumentation.

Also, some of the `.` aren't escaped, not that it _really_ matters as it will still match the char `.` 😉

.github/labeller.yml
```diff
@@ -0,0 +1,52 @@
...
+
+  - name: "area:opentracing"
+    allFilesIn: ".*OpenTracing.*"
+
+  - name: "area:shared-components"
+    allFilesIn: "shared\/.*"
+
+  - name: "area:tests"
+    allFilesIn: "tracer/test/.*"
```
andrewlock, line 45: arguably stuff in `tracer/test/test-applications` is about `area:tests` not `area:test-apps`. it could also be about benchmarks.

.github/labeller.yml
```diff
@@ -0,0 +1,52 @@
...
+
+  - name: "area:test-apps"
+    allFilesIn: ".*\/test-applications\/.*"
+
+  # if any modified file is in tracer/src/Datadog.Trace, excluding files in ClrProfiler, AppSec or Ci
+  - name: "area:tracer"
+    allFilesIn: "(?!.*ClrProfiler.*|.*\/Ci\/.*|.*\/AppSec\/.*)(src\/Datadog\\.Trace\/.*)|tracer\/test\/.*"
```
andrewlock, line 52: Plenty of stuff in ClrProfiler falls under "tracer" IMO. But then that's why these automatic labellers always bug me 😉

.github/labeller.yml
```diff
@@ -0,0 +1,52 @@
+labels:
```
andrewlock, line 1: IMO, if this is going to be ingested and converted in Nuke, we should just define them in Nuke, instead of having to parse and interpret from YAML. If you defined them in Nuke directly, you get extra IDE support for the regexes, and don't limit yourself to what can be defined in YAML. I don't think it gains a lot.

tracer/build/_build/Build.GitHub.cs
```diff
@@ -84,6 +86,88 @@ partial class Build
...
+       .Executes(async() =>
+        {
+            var client = GetGitHubClient();
+
+            var pr = await client.PullRequest.Get(
+                owner: GitHubRepositoryOwner,
+                name: GitHubRepositoryName,
+                number: PullRequestNumber.Value);
+
+            GitTasks.Git("fetch origin master:master", logOutput: false);
```
andrewlock, line 103: I think we should remove this line? Seems unnecessary given you've already done a checkout?

```suggestion
            GitTasks.Git("fetch origin master:master", logOutput: false);
```

.github/workflows/auto_label_prs.yml
```diff
@@ -0,0 +1,23 @@
+name: Label PRs
+
+on:
+- pull_request
```
andrewlock, line 4: I like that this only runs when a PR is `opened`, so we can change it afterwards if we don't like the labelling 👍

andrewlock (approved):
LGTM 👍 Feel free to ignore all the comments, I got carried away, and this will do the job as-is IMO 😂

.github/labeller.yml
pierotibou: Agreed. I even hesitated adding the tracer in auto labelling. I left cause I honestly dunno what's n'est (or least worse;))

.github/labeller.yml
pierotibou: Good point. I didit as a reflex to split conf and code. But I don't mind either way.

tracer/build/_build/Build.GitHub.cs
pierotibou: So my first try was without this line. And it failed `fatal: ambiguous argument 'master': unknown revision or path not in the working tree.`. 
This is me fixing the issue after a quick look at StackOverflow :p So I'd happily change to a better solution (or at least adding a comment)

.github/workflows/auto_label_prs.yml
pierotibou: Yes. Also I made it mandatory in my spec to disallow removing labels. Because, I could imagine seeing myself fighting a bot :p

.github/labeller.yml
pierotibou: Indeed, I think it's better too. Fixed it in e80d75615202e9d91f1a3c193729d9de4e17a91b

.github/labeller.yml
pierotibou: Fixed in e80d75615202e9d91f1a3c193729d9de4e17a91b

tracer/build/_build/Build.GitHub.cs
andrewlock: ah, the next line should probably `origin/master`. The fetch here is creating a local `master` branch and associating it with the upstream, which you don't want/need to do 🙂

pierotibou: Thanks a lot. Fixed in 11c3cb345eb279dba2a0df0166e1e845a683e54d

.github/labeller.yml
```diff
@@ -0,0 +1,49 @@
...
+    allFilesIn: "\\.github\/.*|tracer\/build\/.*|\\.azure-pipelines\/.*|gitlab-ci\\.yml"
+
+  - name: "area:vendors"
+    allFilesIn: "tracer\/src\/Datadog\\.Trace\/vendors\/.*"
+
+  - name: "area:docs"
+    allFilesIn: "docs\/.*|.*README.*|.*\\.md"
+
+  - name: "area:installers"
+    allFilesIn: "shared\/src\/msi-installer\/.*"
```
zacharycmontoya, line 33: Would files here also receive the `area:shared-components` label below? Ideally I would only want this label and not both

zacharycmontoya (approved):

tracer/build/_build/Build.GitHub.cs
pierotibou: Actually, it didn't work. I've roll backed and will merge :) I can revisit later

PR #2663, opened 2022-04-09
Explicit public changes in PR description

pierotibou:
## Summary of changes
Change the PR Template to add a section for public changes (eg Configuration keys, API changes)

## Reason for change
The aim is to make more visible public changes. Indeed, we may not want to review all PRs but seeing in our notifications the public changes could help for visibility.

## Implementation details
Just added a section here, in 2nd place, for better visibility

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,5 +1,8 @@
 ## Summary of changes
 
+## Public changes
+<!-- Public API changes, new configuration keys... -->
+
 ## Reason for change
 
 ## Implementation details


pierotibou:
@andrewlock , Indeed I thought of automating that in a second step (Config keys aren't in the Snapshots though IIRC, we should add them)

andrewlock:
> Config keys aren't in the Snapshots though IIRC

Well, _public_ config keys are 😉

lucaspimentel:
> ConfigurationKeys are internal in code but public in our public doc... So somehow public.

Most are actually public:
https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs

PR #2664, opened 2022-04-11
[Profiler] Add beta revision in the profiler version

gleocadie:
## Reason for change

Today, the beta revision is not part of the version of the profiler (when sending the pprof). We would like it to be part of the version so we can track issues/features.

## Implementation details

Add a `PROFILER_BETA_REVISION` variable in `dd_profiler_version.h`

### DIFF
--- a/.github/scripts/package_and_deploy.sh
+++ b/.github/scripts/package_and_deploy.sh
@@ -6,8 +6,9 @@ ddprof_deploy_folder=$1
 commit_sha=$2
 commit_author=$3
 
-
-profiler_version=v2.5.0.0_$(date -u +%G%m%d%H%M%S)
+current_profiler_version=$(sed -rn 's/constexpr auto PROFILER_VERSION = "([^"]+)";/\1/p' ../profiler/src/ProfilerEngine/Datadog.Profiler.Native/dd_profiler_version.h)
+current_profiler_beta_version=$(sed -rn 's/constexpr auto PROFILER_BETA_REVISION = "([^"]+)";/\1/p' ../profiler/src/ProfilerEngine/Datadog.Profiler.Native/dd_profiler_version.h)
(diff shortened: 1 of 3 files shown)


luhenry (approved):

.github/scripts/package_and_deploy.sh
```diff
@@ -6,8 +6,9 @@ ddprof_deploy_folder=$1
 commit_sha=$2
 commit_author=$3
 
-
-profiler_version=v2.5.0.0_$(date -u +%G%m%d%H%M%S)
+current_profiler_version=$(sed -rn 's/constexpr auto PROFILER_VERSION = "([^"]+)";/\1/p' ../profiler/src/ProfilerEngine/Datadog.Profiler.Native/dd_profiler_version.h)
```
pierotibou, line 9: Interesting, this one wasn't well bumped. Have you checked if we have forgotten other instances in the VersionBump process?

But also, shouldn't we just add this version in the [bump process](https://cs.github.com/DataDog/dd-trace-dotnet/blob/a908775dbf6c402aa45f19c447aa252cce0b7754/tracer/build/_build/PrepareRelease/SetAllVersions.cs?q=SetAllVersions#L41) instead of doing a sed?

gleocadie: Yep that one the last (🤞 ) occurrence.

>But also, shouldn't we just add this version in the [bump process](https://cs.github.com/DataDog/dd-trace-dotnet/blob/a908775dbf6c402aa45f19c447aa252cce0b7754/tracer/build/_build/PrepareRelease/SetAllVersions.cs?q=SetAllVersions#L41) instead of doing a sed?

We can, but I will still need to do a `sed` to get my `PROFILER_BETA_REVISION`. So I prefer having one place for the bump, and the rest just extract it from there.

PR #2678, opened 2022-04-12
Pass the environment variable to the docker container

gleocadie:
## Reason for change

The environment variable was set in the build process and not in the docker container.

## Implementation details

Just pass the environment variable on the command-line that runs the Docker container.


andrewlock (approved):
Doh, good spot 👍

pierotibou (approved):

PR #2683, opened 2022-04-13
Labeler shouldn't check test folders for tracer and integrations areas

pierotibou:
## Summary of changes
Not checking if all files are within test folders when checking tracer and integrations areas. 

## Reason for change
It was over labelling obviously. The labeler should actually handle tests differently, but this can be improved later as those areas were already best effort.


andrewlock (approved):
I have no idea what these Regexes are doing so LGTM 😉

zacharycmontoya (approved):

PR #2687, opened 2022-04-14
[Profiler] Add new job: Address sanitizer on Linux

gleocadie:
## Summary of changes

Add a new job to build the profiler with Clang Address sanitizer.

## Reason for change

This will help us in finding memory misusages.

## Implementation details

Add new job:
- Ensure libasan is installed
- Run the application with and without the new pipeline

Disable for now memory leak detection. This will be done in another PR since there are a lot of noise due external libraries.


chrisnas (approved):
LGTM

PR #2690, opened 2022-04-15
[Profiler] Add Undefined-Behavior sanitizer on the profiler

gleocadie:
## Summary of changes

Add a new job to build the profiler with Clang Undefined-Behavior sanitizer.

## Reason for change

This will help us in finding various issue:
*  Array subscript out of bounds
* Misalignment

More [here](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html)

## Implementation details

Add new job:
- Ensure libubsan is installed
- Run the application with and without the new pipeline
/!\ for now we do not fail the job if with the old pipeline there is an issue. This pipeline is deprecated and will be soon removed


chrisnas (approved):
LGTM

PR #2701, opened 2022-04-19
[Profiler] Do not export empty profiles

gleocadie:
## Reason for change

When running on IIS, we can end up in a situation where no thread is running in an AppDomain, so no callstack will be collected for this AppDomain/Application.
But still, every minute we send one profile per application to the back-end and we have empty profiles in the dashboard.

## Implementation details

Keep the count of samples added to each profile. Check that count to decide if the profile has to be sent or not.

## Test coverage

Adding unit test.

## Other details
This is a simple implementation. A refactoring is needed and will be done after the 2nd beta release.


chrisnas (approved):
LGTM

PR #2765, opened 2022-05-06
[Profiler] Update CPU profiling implementation

chrisnas:
## Summary of changes
Update CPU profiling algorithm to take into account only running threads

## Reason for change
The current CPU profiling implementation samples like the walltime : look at 5 threads every 9ms and if the thread has run on a CPU since the last check, a sample is taken. However, if the thread was not running on a core at that particular time, the callstack might not represent CPU-bound code.

## Implementation details
One solution to this issue is to check if the thread is running (with a tiny race condition) before generating a sample.

## Test coverage
Tests corresponding to the refactoring have been added


profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/OsSpecificApi.cpp
```diff
@@ -18,4 +19,84 @@ StackFramesCollectorBase* CreateNewStackFramesCollectorInstance(ICorProfilerInfo
...
+//
+// Another solution would be to use clock_gettime but without the Running status available
+//    pthread_getcpuclockid(pthread_self(), &clockid);
+//    if (clock_gettime(clockid, &cpu_time)) { ... }
+//
+
+bool GetCpuInfo(pid_t tid, bool& isRunning, uint64_t& cpuTime)
+{
+    char statPath[64];
+    snprintf(statPath, sizeof(statPath), "/proc/self/task/%d/stat", (int)syscall(SYS_gettid));
```
luhenry, line 53: Need to use `tid` instead of current thread's tid.

profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/OsSpecificApi.cpp
```diff
@@ -18,4 +19,84 @@ StackFramesCollectorBase* CreateNewStackFramesCollectorInstance(ICorProfilerInfo
...
+        fscanf(file, "%*s %*s %c %*s %*s %*s %*s %*s %*s %*s %*s %*s %*s %d %d",
+               &state, &userTime, &kernelTime) == 3;
+    fclose(file);
+    if (!success)
+    {
+        return false;
+    }
+
+    cpuTime = ((userTime + kernelTime) * 1000) / sysconf(_SC_CLK_TCK);
+    isRunning = state == 'R';
```
luhenry, line 74: async-profiler checks for `D` as well which is for `Waiting in uninterruptible disk sleep`. I'm not sure what's that case but might be worth understanding why.

profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/OsSpecificApi.cpp
chrisnas: Good catch!
✅

profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/OsSpecificApi.cpp
chrisnas: ✅

profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/OsSpecificApi.cpp
```diff
@@ -18,4 +19,84 @@ StackFramesCollectorBase* CreateNewStackFramesCollectorInstance(ICorProfilerInfo
 {
     return new LinuxStackFramesCollector(const_cast<ICorProfilerInfo4* const>(pCorProfilerInfo));
 }
+
+// TODO: use /proc/<pid>/task/<tid>/stat
```
gleocadie, line 23:
```suggestion
```

profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/OsSpecificApi.cpp
```diff
@@ -18,4 +19,84 @@ StackFramesCollectorBase* CreateNewStackFramesCollectorInstance(ICorProfilerInfo
...
+    cpuTime = ((userTime + kernelTime) * 1000) / sysconf(_SC_CLK_TCK);
+    isRunning = (state == 'R') || (state == 'D') || (state == 'W');
+    return true;
+}
+
+uint64_t GetThreadCpuTime(ManagedThreadInfo* pThreadInfo)
+{
+    bool isRunning = false;
+    uint64_t cpuTime = 0;
+    if (!GetCpuInfo(pThreadInfo->GetOsThreadId(), std::ref(isRunning), std::ref(cpuTime)))
```
gleocadie, line 82: why using `std::ref` here ? `std::ref` will create an object wrapper.
you `GetCpuInfo` use them as reference ?

profiler/src/ProfilerEngine/Datadog.Profiler.Native.Windows/OsSpecificApi.cpp
```diff
@@ -36,4 +38,133 @@ StackFramesCollectorBase* CreateNewStackFramesCollectorInstance(ICorProfilerInfo
     return new Windows32BitStackFramesCollector(pCorProfilerInfo);
 #endif
 }
+
+uint64_t GetThreadCpuTime(ManagedThreadInfo* pThreadInfo)
+{
+    uint64_t duration = 0;
```
gleocadie, line 44: nitpick: this variable is not used. you can remove duration and `return 0` at line 56

profiler/src/ProfilerEngine/Datadog.Profiler.Native/ManagedThreadList.cpp
```diff
@@ -226,149 +183,99 @@ bool ManagedThreadList::SetThreadName(ThreadID clrThreadId, const shared::WSTRIN
...
-                (*ppThreadInfo)->AddRef(); // caller must release
-            }
-            return true;
-        }
+        return nullptr;
     }
+
+    ManagedThreadInfo* pInfo = elem->second;
+    pInfo->AddRef(); // caller must release
+    return pInfo;
```
gleocadie, line 265: I like the simplicity ^^ . thx for that ;)

profiler/src/ProfilerEngine/Datadog.Profiler.Native/StackSamplerLoop.cpp
```diff
@@ -510,79 +576,43 @@ void StackSamplerLoop::LogEncounteredStackSnapshotResultStatistics(std::int64_t
...
+        rawSample.SpanId = pSnapshotResult->GetSpanId();
+        rawSample.AppDomainId = pSnapshotResult->GetAppDomainId();
+        pSnapshotResult->CopyInstructionPointers(rawSample.Stack);
+        rawSample.ThreadInfo = pThreadInfo;
+        pThreadInfo->AddRef();
+        rawSample.Duration = pSnapshotResult->GetRepresentedDurationNanoseconds();
+        _pWallTimeCollector->Add(std::move(rawSample));
+    }
+    else
+    if ((profilingType == PROFILING_TYPE::CpuTime) && (_pConfiguration->IsCpuProfilingEnabled()))
```
gleocadie, line 604: do we really need to check if it's enabled ?
if not enabled, we won't have Cpu samples, right ?

gleocadie (commented):
Few comment

profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/OsSpecificApi.cpp
chrisnas: ✅

profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/OsSpecificApi.cpp
chrisnas: I used std::ref to make explicit the "ref" usage
It is now removed.
✅

profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/OsSpecificApi.cpp
```diff
@@ -18,4 +19,84 @@ StackFramesCollectorBase* CreateNewStackFramesCollectorInstance(ICorProfilerInfo
...
+        return 0;
+    }
+
+    return cpuTime;
+}
+
+bool IsRunning(ManagedThreadInfo* pThreadInfo, uint64_t& cpuTime)
+{
+    bool isRunning = false;
+    if (!GetCpuInfo(pThreadInfo->GetOsThreadId(), std::ref(isRunning), std::ref(cpuTime)))
```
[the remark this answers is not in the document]
chrisnas: ✅

profiler/src/ProfilerEngine/Datadog.Profiler.Native.Windows/OsSpecificApi.cpp
chrisnas: ✅

profiler/src/ProfilerEngine/Datadog.Profiler.Native.Windows/OsSpecificApi.cpp
```diff
@@ -36,4 +38,133 @@ StackFramesCollectorBase* CreateNewStackFramesCollectorInstance(ICorProfilerInfo
...
+    return
+        (THREAD_STATE::Running == threadState) ||
+        (THREAD_STATE::Standby == threadState) ||
+        (THREAD_STATE::Ready == threadState) ||
+        (THREAD_STATE::DeferredReady == threadState);
+
+    // Note that THREAD_STATE::Standby, THREAD_STATE::Ready and THREAD_STATE::DeferredReady
+    // indicate that threads are simply waiting for an available core to run.
+    // If too many samples are missed with Running only, it might be interesting to return TRUE
+    // for these other states (especially if these threads consummed CPU since the last call)
```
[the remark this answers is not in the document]
chrisnas: Right: I will change the comment now that all these states are used as "Running" in case we would like to be conservative

profiler/src/ProfilerEngine/Datadog.Profiler.Native/StackSamplerLoop.cpp
chrisnas: Just in case...
But I'm removing the test
✅

profiler/src/ProfilerEngine/Datadog.Profiler.Native/ManagedThreadList.cpp
```diff
@@ -226,149 +183,99 @@ bool ManagedThreadList::SetThreadName(ThreadID clrThreadId, const shared::WSTRIN
...
-            }
-        }
-        else
-        {
-            newThreadsData->TrySet(newNextFreeIndex, *ppThreadInfo);
-            newNextFreeIndex++;
-        }
+        pos = 0;
     }
+    _iterators[iterator] = pos;
```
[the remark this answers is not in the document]
chrisnas: I prefer to keep it more "explicit"  :^)

luhenry (approved):

PR #2797, opened 2022-05-13
[Profiler] Remove dependency to Datadog.AutoInstrumentation.Profiler.Managed

gleocadie:
## Reason for change

2 things:
 - We are cleaning up profiler code: removing the old pipeline.
 - We want to run the profiler integration tests in AzDo

## Implementation details

Remove the dependency to the profiler managed project and add one to Google Protobuf

### DIFF
--- a/profiler/test/Datadog.Profiler.IntegrationTests/Datadog.Profiler.IntegrationTests.csproj
+++ b/profiler/test/Datadog.Profiler.IntegrationTests/Datadog.Profiler.IntegrationTests.csproj
@@ -36,6 +36,7 @@
 
   <ItemGroup>
     <PackageReference Include="FluentAssertions" Version="5.10.3" />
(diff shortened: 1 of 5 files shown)


profiler/test/Datadog.Profiler.IntegrationTests/Datadog.Profiler.IntegrationTests.csproj
```diff
@@ -36,6 +36,7 @@
 
   <ItemGroup>
     <PackageReference Include="FluentAssertions" Version="5.10.3" />
+    <PackageReference Include="Google.Protobuf" Version="3.20.1" />
```
andrewlock, line 39: Does this work on alpine + arm64? I know I had all sorts of issues with GRPC, but I can't remember if that was the grpc-tool or protobuf lib

PR #2805, opened 2022-05-14
[Profiler] Do not aggregate sample with empty callstack

gleocadie:
## Reason for change

Currently we do not resolve native frames. This leads to samples with empty callstacks and in the dashboard we can have profile with not callstack at all.

## Implementation details

Filter out samples with empty callstack.

## Test coverage

- Added a unit test


chrisnas (approved):
LGTM

PR #2814, opened 2022-05-17
[Profiler] Flush samples and export a last .pprof on exit

chrisnas:
## Summary of changes
A final .pprof will be sent on application exit

## Reason for change
The samples taken since the last export are lost when the application exits.
A last .pprof containing the collected samples should be sent when the CLR shutdown

## Implementation details
- Allow providers to support multiple calls to Stop() and transforms all raw samples when stopping
- Update aggregator tests to explicitly check final .pprof


gleocadie (approved):
LGTM

PR #2819, opened 2022-05-18
[Profiler] Add a profile_seq tag to each uploaded .pprof

chrisnas:
## Summary of changes
Add an incrementing "profile_seq" tag to the HTTP request used to upload .pprof files

## Reason for change
It is not easy to identify the first uploaded profile for a given service. With this new tag, the identification will be easy. 
image
In addition, the backend could use it to provide better analysis by focusing only on profiles of a single execution run

## Implementation details
Add an export count information to the data associated to each application in the exporter
Refactor the corresponding code that was using std::pair

## Test coverage
Validated manually 
It might be possible to add it as a specific integration test by parsing the multipart format of the request.


gleocadie (approved):
LGTM

PR #2823, opened 2022-05-20
[Profiler] Implement sampling for exceptions

kevingosse:
## Summary of changes

Add sampling to the exception profiler.

## Reason for change

Walking the stack of a thread can be costly, so we need to put some kind of limit.

## Implementation details

This is a 1:1 port of the AdaptiveSample algorithm used in the Java profiler: https://github.com/DataDog/dd-trace-java/blob/661b886b0b80474326072f072fe0584c281d3198/internal-api/internal-api-8/src/main/java/datadog/trace/api/sampling/AdaptiveSampler.java

On top of it, we keep track of what types of exception have been seen in the sampling window. If a type if seen for the first time, we override the sampling decision and we keep the sample.

## Test coverage

The unit tests have been ported from the Java codebase.
The integration test makes sure the number of sampled exceptions is within the expected order of magnitude.


profiler/src/ProfilerEngine/Datadog.Profiler.Native/AdaptiveSampler.cpp
```diff
@@ -0,0 +1,194 @@
...
+    int32_t samplesPerWindow,
+    int32_t averageLookback,
+    int32_t budgetLookback,
+    std::function<void()> rollWindowCallback) :
+    _timer([this] { RollWindow(); }, windowDuration),
+    _totalCountRunningAverage(0),
+    _rollWindowCallback(std::move(rollWindowCallback)),
+    _avgSamples(0)
+{
+    if (averageLookback < 1)
```
chrisnas, line 57: Would it be possible to have a default value and log an error instead of throwing an exception?

kevingosse: Done

profiler/src/ProfilerEngine/Datadog.Profiler.Native/AdaptiveSampler.cpp
```diff
@@ -0,0 +1,194 @@
...
+    {
+        throw std::invalid_argument("'averageLookback' argument must be at least 1");
+    }
+
+    if (budgetLookback < 1)
+    {
+        throw std::invalid_argument("'budgetLookback' argument must be at least 1");
```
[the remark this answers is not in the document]
kevingosse: Done

chrisnas (approved):
LGTM

PR #2837, opened 2022-05-24
[Profiler] Fix profiler throughput test on master

kevingosse:
## Summary of changes

`GITHUB_HEAD_REF` is not defined when running the job on master. In that case, use the commit SHA instead (and prepend it with `#` because that's the syntax expected by Crank).


andrewlock (approved):
Looks good to me, what good possibli go wrong

gleocadie (approved):
🤞 LGTM

PR #2845, opened 2022-05-25
[CIApp] - Refactor CodeOwner feature + Skip Test with Zero duration.

tonyredondo:
## Summary of changes

The PR refactors the CodeOwner feature into the Common Test instrumentation class and remove any duration to the skipped tests.

## Reason for change

Code cleaning and correctness.

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Testing/MsTestV2/TestMethodRunnerExecuteIntegration.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Testing/MsTestV2/TestMethodRunnerExecuteIntegration.cs
@@ -65,6 +65,7 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Testing.MsTestV2
                             scope.Span.SetTag("test.coverage", Datadog.Trace.Vendors.Newtonsoft.Json.JsonConvert.SerializeObject(coverageSession));
                         }
 
+                        scope.Span.Finish(TimeSpan.Zero);
                         scope.Dispose();
                     }
                 }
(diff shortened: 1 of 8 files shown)


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Testing/MsTestV2/TestMethodRunnerExecuteIntegration.cs
```diff
@@ -65,6 +65,7 @@ public static class TestMethodRunnerExecuteIntegration
                             scope.Span.SetTag("test.coverage", Datadog.Trace.Vendors.Newtonsoft.Json.JsonConvert.SerializeObject(coverageSession));
                         }
 
+                        scope.Span.Finish(TimeSpan.Zero);
```
andrewlock, line 68: We may not be able to do anything about it, but `Inconclusive` doesn't mean it didn't run AFAIK. The duration could definitely be non-zero. e.g. if it's killed while running. If the data isn't available, then maybe that's fine, but I'm not sure.

andrewlock (approved):

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Testing/MsTestV2/TestMethodRunnerExecuteIntegration.cs
tonyredondo: Well the thing is that we are not able to measure that correctly, in this block we are creating and closing the scope, so the actual duration being measured is the one during setting tags and creating a coverage session. So I prefer to report Zero here.

zacharycmontoya (approved):

PR #2861, opened 2022-05-31
[Profiler] Set _GLIBCXX_USE_CXX11_ABI to 0 and remove liblzma dependency

gleocadie:
## Reason for change

In the past, the tracer suffered from a bug where the customer application got stuck due to an `ABI`  issue.
By setting the flag `_GLIBCXX_USE_CXX11_ABI` to 0, the problem was solved.

In that case, we compile the profiler and its dependency with that flag set to `0`.

We also remove the `liblzma` dependency, to avoid having the profiler not working on alpine (base/default installation)
## Implementation details

Set the flag `_GLIBCXX_USE_CXX11_ABI` to 0
Deactivate `minidebuginfo` in libunwind to remove the liblzma dependency

### DIFF
--- a/profiler/CMakeLists.txt
+++ b/profiler/CMakeLists.txt
@@ -83,6 +83,8 @@ if (DEFINED RUN_ANALYSIS AND RUN_ANALYSIS EQUAL 1)
     endif()
 endif()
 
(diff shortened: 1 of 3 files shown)


chrisnas (approved):
LGTM

PR #2866, opened 2022-06-01
[Profiler] Prevent from blocking the application

gleocadie:
## Reason for change

Here we have two cases:
- on alpine dlopen has a lock which is not recursive and if the application thread is hijacked while holding the lock, the thread will stay blocked. By wrapping we prevent the thread from being hijacked while in a dlopen call.
- There are times in AzDo when the sampler thread was blocked, waiting for the application thread. But the application thread was dead (no notification from the CLR).

## Implementation details

For the first case, we just replace dlopen by our implementation (which just wrap and prevent signals from interrupting its execution)
For the second case, we check if the thread is alive by sending a  `0` signal. In reality, no signal is sent to the thread, only sanity check (is the thread alive?)

### DIFF
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/LinuxStackFramesCollector.cpp
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/LinuxStackFramesCollector.cpp
@@ -4,6 +4,7 @@
 #include "LinuxStackFramesCollector.h"
 
 #include <cassert>
+#include <chrono>
 #include <errno.h>
 #include <mutex>
 #include <signal.h>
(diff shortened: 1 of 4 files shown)


profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/LinuxStackFramesCollector.cpp
```diff
@@ -82,7 +95,15 @@ StackSnapshotResultBuffer* LinuxStackFramesCollector::CollectStackSampleImplemen
         }
         else
         {
-            _stackWalkInProgressWaiter.wait(stackWalkInProgressLock);
+            do
```
chrisnas, line 98: I would add a small comment here to explain why we needed this code; something like:
   // When the application ends and the CLR shuts down, it might happen that 
   // the currently walked thread gets terminated without notice for us.
   // This loop ensures that the code does not stay stuck waiting for a lock that will
   // never be released. It will exit if the stack walked thread does not run anymore or 
   // the stack walk finishes as expected.
Feel free to adapt  :^)

chrisnas (approved):
LGTM with just a tiny comment to add

PR #2867, opened 2022-06-01
[Profiler] Use uws_backtrace for stackwalking on Linux

kevingosse:
## Summary of changes

Use libunwind `uws_backtrace` for stackwalking instead of manually using the cursor.

`uws_backtrace` is supposed to be significantly faster because it relies on an internal cache.

## Implementation details

Since `uws_backtrace` directly writes to the buffer, this PR also changes the underlying storage of `StackSnapshotResultBuffer` to use a `std::array` instead of a `std::vector`, so we can safely get a pointer to the underlying array.

### DIFF
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/LinuxStackFramesCollector.cpp
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/LinuxStackFramesCollector.cpp
@@ -24,7 +24,7 @@ error("unsupported architecture")
 #include "ManagedThreadInfo.h"
 #include "OpSysTools.h"
 #include "ScopeFinalizer.h"
-#include "StackSnapshotResultReusableBuffer.h"
+#include "StackSnapshotResultBuffer.h"
 
 using namespace std::chrono_literals;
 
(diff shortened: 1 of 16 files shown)


profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/LinuxStackFramesCollector.cpp
```diff
@@ -253,54 +253,22 @@ std::int32_t LinuxStackFramesCollector::CollectCallStackCurrentThread()
...
-                if (!AddFrame(nativeInstructionPointer))
-                {
-                    return S_FALSE;
-                }
-
-                resultErrorCode = unw_step(&cursor);
-            }
+        if (count == 0)
+        {
+            AddFakeFrame();
```
gleocadie, line 266: if we had a fake frame, the number of frames will be 1 and in the stacksampler loop  (for example), we will process this `empty` callstack. (the return value `S_FALSE` and al is kept only for statistic. We process a sample if the callstack is not empty)

maybe we can just skip adding a fake frame in that case, no?

gleocadie (approved):
Except the comment about `adding a fake frame`, the rest looks good

profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/LinuxStackFramesCollector.cpp
kevingosse: Right. I was mimicking the "collection aborted" case of the previous code, but I guess it's closer to the "'error" case, where didn't add fake frames. I'll fix this.

PR #2868, opened 2022-06-01
[Profiler] Do not resolve native frame if not needed

gleocadie:
## Reason for change

Native frames are not displayed today in the flamegraph and in the profiler, we do not fully resolve the symbol (only the module name).
This has a performance impact on the application (linux `dladdr` incurs a non-negligible impact)

## Implementation details

- Pass the configuration class to the `FrameStore` and check in `GetFrame` if we need to resolve native frames.
- Cleanup

### DIFF
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native/FrameStore.cpp
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native/FrameStore.cpp
@@ -21,7 +21,7 @@ FrameStore::FrameStore(ICorProfilerInfo4* pCorProfilerInfo, IConfiguration* pCon
 std::tuple<bool, std::string, std::string> FrameStore::GetFrame(uintptr_t instructionPointer)
 {
     static const std::string NotResolvedModuleName("NotResolvedModule");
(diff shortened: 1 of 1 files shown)


profiler/src/ProfilerEngine/Datadog.Profiler.Native/FrameStore.cpp
```diff
@@ -10,13 +12,17 @@
...
+FrameStore::FrameStore(ICorProfilerInfo4* pCorProfilerInfo, IConfiguration* pConfiguration) :
+    _pCorProfilerInfo{pCorProfilerInfo},
+    _resolveNativeFrames{pConfiguration->IsNativeFramesEnabled()}
 {
 }
 
 std::tuple<bool, std::string, std::string> FrameStore::GetFrame(uintptr_t instructionPointer)
 {
+    static const std::string NotResolvedModuleName("NotResolvedModule");
+    static const std::string NotResolvedFrame("NotResolvedModule");
```
chrisnas, line 24:
```suggestion
    static const std::string NotResolvedFrame("NotResolvedFrame");
```

chrisnas (changes requested):

PR #2889, opened 2022-06-09
[Profiler] Add iterator scenario in computer01 demo

chrisnas:
## Summary of changes
Add iterator code in demo application

## Reason for change
Need to have iterator and constructor frames in call stacks

## Implementation details
Add a new scenario in computer01 demo app


gleocadie (approved):
LGTM

PR #2899, opened 2022-06-14
[Profiler] use string view to avoid string copies

gleocadie:
## Reason for change

When resolving symbols, we return `std::string`, and those string are copied multiple times, until added to the profile..
By using `std::string_view` wherever we can we avoid string copies and memory/cpu overhead.

## Implementation details

Use `std::string_view` when resolving symbols.

## Test coverage

Ensure that current test are working correctly.


profiler/src/ProfilerEngine/Datadog.Profiler.Native/Sample.h
```diff
@@ -56,10 +56,11 @@ static constexpr size_t array_size = sizeof(SampleTypeDefinitions) / sizeof(Samp
 //---------------------------------------------------------------
 
 typedef std::array<int64_t, array_size> Values;
-typedef std::pair<std::string, std::string> Label; // TODO: use stringview to avoid copy
+typedef std::pair<std::string_view, std::string> Label; // TODO: use stringview to avoid copy
```
chrisnas, line 59: Remove the comment since you are using stringview for the label text (keeping  string for its value seems reasonable)

profiler/src/ProfilerEngine/Datadog.Profiler.Native/Sample.h
```diff
@@ -108,7 +109,7 @@ class Sample
 
 private:
     uint64_t _timestamp;
-    std::vector<std::pair<std::string, std::string>> _callstack; // TODO: use stringview to avoid copy
+    CallStack _callstack; // TODO: use stringview to avoid copy
```
chrisnas, line 112: remove comment because you are making the change

chrisnas (approved):
LGTM with comments removal

profiler/src/ProfilerEngine/Datadog.Profiler.Native/FrameStore.cpp
```diff
@@ -146,7 +146,12 @@ std::pair<std::string, std::string> FrameStore::GetManagedFrame(FunctionID funct
         // try to get the type description
         if (!GetTypeDesc(pMetadataImport.Get(), classId, moduleId, mdTokenType, typeDesc))
         {
-            return {UnknownManagedAssembly, UnknownManagedType + " |fn:" + methodName};
+            // This should never happen but in case it happens, we cached the module/frame value.
```
kevingosse, line 149: I know it's just a small typo, but in this context it actually changes the meaning quite a bit
```suggestion
            // This should never happen but in case it happens, we cache the module/frame value.
```

kevingosse (approved):
LGTM + 1 nit

PR #2902, opened 2022-06-14
[Profiler] Do not print stack walker error messages everytime

gleocadie:
## Reason for change

Time to time the stack sampler can encounter errors. We log those messages each time. The log file size will continue increasing and this is an issue on customer using alpine with little storage allocated to the application.

## Implementation details

Collect the different errors as statistics and log them every 10min.


chrisnas (approved):
LGTM

PR #2903, opened 2022-06-15
[Profiler] Don't sample if no CPU time was consumed nor in Ready state

chrisnas:
## Summary of changes
Filter out more samples based on CPU-related criteria

## Reason for change
Try to avoid waiting frames on call stacks for CPU profiling


gleocadie (approved):
LGTM

PR #2917, opened 2022-06-23
[Profiler] Refactor raw sample processing

kevingosse:
## Summary of changes

- Move the transform method to the implementation of RawSample
- Have a single thread in SampleAggregator process the raw samples, instead of one thread per provider

## Reason for change

Having one thread per provider wouldn't scale as we keep adding new providers.


gleocadie (approved):
LGTM

PR #2924, opened 2022-06-29
[Profiler] Change PiComputation to run a fixed amount of time

kevingosse:
## Summary of changes

Convert PiComputation to run for a set amount of time. The number of iterations is too dependent on the hardware and the version of the runtime.

## Other details
Also replaced the ManualResetEvent with a CancellationToken.


chrisnas (approved):
LGTM

PR #2925, opened 2022-06-29
[Profiler] POC to receive CLR events synchronously via ICorProfilerCallback

chrisnas:
## Summary of changes
Create a session to receive CLR events via ICorProfilerCallback
Handle AllocationTickV4 (Allocation profiling) and ContentionStopV1 (Contention profiling)

## Reason for change
Validate the solution


chrisnas:
Used only for R&D Week

PR #2929, opened 2022-06-30
[Profiler] Fix parsing thread information

chrisnas:
## Summary of changes
Fix parsing of /proc/self/task/%d/stat to get thread running state and cpu consumption

## Reason for change
Test were failing in .NET 6

## Implementation details
Take into account thread names with SPACE and ()

### DIFF
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/OsSpecificApi.cpp
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/OsSpecificApi.cpp
@@ -8,6 +8,7 @@
 
 #include <sys/syscall.h>
 #include "OsSpecificApi.h"
+#include "OpSysTools.h"
 
 #include "Log.h"
 #include "LinuxStackFramesCollector.h"
@@ -67,14 +68,10 @@ bool GetCpuInfo(pid_t tid, bool& isRunning, uint64_t& cpuTime)
         return false;
     }
 
-    // based on https://linux.die.net/man/5/proc
-    char state = ' ';   // 3rd position  and 'R' for Running
-    int userTime = 0;   // 14th position in clock ticks
-    int kernelTime = 0; // 15th position in clock ticks
-
-    auto pos = sline.find_last_of(")");
-    const char* pEnd = sline.c_str() + pos + 1;
-    bool success = sscanf(pEnd, " %c %*s %*s %*s %*s %*s %*s %*s %*s %*s %*s %d %d", &state, &userTime, &kernelTime) == 3;
+    char state = ' ';
+    int userTime = 0;
+    int kernelTime = 0;
+    bool success = OpSysTools::ParseThreadInfo(sline, state, userTime, kernelTime);
     if (!success)
     {
         // log the first error to be able to analyze unexpected string format
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/OsSpecificApi.cpp
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/OsSpecificApi.cpp
@@ -3,6 +3,9 @@
 
 // OsSpecificApi for LINUX
 
+#include <fstream>
+#include <string>
+
 #include <sys/syscall.h>
 #include "OsSpecificApi.h"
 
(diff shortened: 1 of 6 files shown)


profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/OsSpecificApi.cpp
```diff
@@ -60,10 +69,10 @@ bool GetCpuInfo(pid_t tid, bool& isRunning, uint64_t& cpuTime)
     char state = ' ';   // 3rd position  and 'R' for Running
     int userTime = 0;   // 14th position in clock ticks
     int kernelTime = 0; // 15th position in clock ticks
-    bool success =
-        fscanf(file, "%*s %*s %c %*s %*s %*s %*s %*s %*s %*s %*s %*s %*s %d %d",
-               &state, &userTime, &kernelTime) == 3;
-    fclose(file);
+
+    auto pos = sline.find_last_of(")");
```
gleocadie, line 73: for the question (I asked it myself): are parenthesis always used ?
answerd: This is the code in the kernel that write in that file and there will always be parenthesis https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/proc/array.c#n567

gleocadie (approved):

profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/OsSpecificApi.cpp
```diff
@@ -50,8 +53,14 @@ bool GetCpuInfo(pid_t tid, bool& isRunning, uint64_t& cpuTime)
 {
     char statPath[64];
     snprintf(statPath, sizeof(statPath), "/proc/self/task/%d/stat", tid);
-    FILE* file = fopen(statPath, "r");
-    if (file == nullptr)
+
+    // load the line to be able to parse it in memory
+    std::ifstream file;
+    file.open(statPath);
```
andrewlock, line 59: Previously it looks like `fopen` may possible fail `(file == nullptr)` is that no longer the case?

profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/OsSpecificApi.cpp
```diff
@@ -60,10 +69,10 @@ bool GetCpuInfo(pid_t tid, bool& isRunning, uint64_t& cpuTime)
     char state = ' ';   // 3rd position  and 'R' for Running
     int userTime = 0;   // 14th position in clock ticks
     int kernelTime = 0; // 15th position in clock ticks
-    bool success =
-        fscanf(file, "%*s %*s %c %*s %*s %*s %*s %*s %*s %*s %*s %*s %*s %d %d",
-               &state, &userTime, &kernelTime) == 3;
-    fclose(file);
+
+    auto pos = sline.find_last_of(")");
+    const char* pEnd = sline.c_str() + pos + 1;
```
kevingosse, line 74: I believe you should check that `pos + 1` is within the string

kevingosse (commented):
I think it would be nice to have a unit test for this

profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/OsSpecificApi.cpp
```diff
@@ -60,10 +69,10 @@ bool GetCpuInfo(pid_t tid, bool& isRunning, uint64_t& cpuTime)
...
+    bool success = sscanf(pEnd, " %c %*s %*s %*s %*s %*s %*s %*s %*s %*s %*s %d %d", &state, &userTime, &kernelTime) == 3;
```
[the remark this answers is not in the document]
andrewlock: image

andrewlock:
[Woohoo, it fixes the integration tests](https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=103531&view=results)! 🎉

profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/OsSpecificApi.cpp
chrisnas: > Previously it looks like `fopen` may possible fail `(file == nullptr)` is that no longer the case?

It is no more the case: the line will be empty (no exception)

profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/OsSpecificApi.cpp
chrisnas: if final ), then points to the \0

chrisnas:
> [Woohoo, it fixes the integration tests](https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=103531&view=results)! 🎉

We are doing that just for you @andrewlock ;^)

kevingosse (approved):

PR #2935, opened 2022-07-01
Disable cpu profiler during exception tests

kevingosse:
## Summary of changes

There is a known race-condition where the callstacks of the exception is not collected if the cpu/walltime profiler is sampling the same thread at the same time. This makes the GetExceptionSamples test flaky.

This PR adds a setting to disable the CPU profiler to make the test more reliable.


gleocadie (approved):
LGTM

chrisnas (approved):
LGTM

PR #2940, opened 2022-07-04
[Profiler] Do not dereference nullptr to avoid crashing

gleocadie:
## Reason for change
When deactivating the walltime profiler, the walltime instance variable is null. If we do not protect it access, we will dereference a nullptr and will crash the application.

## Implementation details

Check if the walltime provider is activated or not.


andrewlock (approved):

andrewlock:
[Confirmed this fixes the tests again](https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=103733&view=results) 👍

chrisnas (approved):
LGTM

PR #2958, opened 2022-07-07
[Profiler] Bump libddprof version to 0.6.0

gleocadie:
## Reason for change

We use an old version of libddprof. In order to build the profiler on Arm64, we need to update it.

## Implementation details

Bump nuget version to 0.6.0.


kevingosse (approved):
I wouldn't say it looks _good_ but at least it looks correct 😛

PR #2967, opened 2022-07-12
[Profiler] Deactivate profiling on linux if wrapper library is not used

gleocadie:
## Summary of changes

On Linux, to avoid deadlocking the profiler (and app threads), we use a dynamic loader trick that allows us to force the application to use our custom implementation for functions that might deadlock (`dl_iterate_phdr`, `dlopen`...)

To use this trick, we need a special library that has the same signature of the ones we want to wrap and pass the path of this library in `LD_PRELOAD` environment variable.

If the value of this environment variable does not contain the path to our library, we might end up blocking the customer application.

## Reason for change

We need to check if the value of `LD_PRELOAD` environment variable contains our library. If it's not the case, we just disable the profiler.

## Implementation details

Check if the library file exists and if one of our custom symbol (`dl_iterate_phdr`) is the one used by default.

## Test coverage

Add integration tests for linux only


kevingosse (approved):
LGTM

PR #2977, opened 2022-07-13
[Profiler] Minor cleanup

kevingosse:
## Summary of changes

- Remove `DisplayName` attribute from tests to use the test name instead
- Replace usage of `int`, `long`, and `long long` by their explicit equivalent
- Fix warnings

## Reason for change

Clean is good.


gleocadie (approved):
LGTM

PR #2981, opened 2022-07-14
[Profiler] Replace usage of refcounted object by shared_ptr

gleocadie:
## Summary of changes

Remove homemade `RefCounted` class and use `std::shared_ptr` instead.

## Reason for change

Our refcounted class does not benefit from RAII to acquire and release pointer. We have to do it by hand and this ca be error-prone (e.g.: memory leak if we forget to call Release)

Instead of using/implementing our own, we will reuse the `std::shared_ptr` class which does the exact same thing and rely on RAII for safety.

## Implementation details

Replace Refcounted by std::shared_ptr.


kevingosse (approved):

profiler/src/ProfilerEngine/Datadog.Profiler.Native/StackSamplerLoop.cpp
```diff
@@ -246,20 +241,13 @@ void StackSamplerLoop::CpuProfilingIteration(void)
...
-            _targetThread = nullptr;
-
+            _targetThread.reset();
         }
     }
 }
 
 void StackSamplerLoop::CollectOneThreadStackSample(
-    ManagedThreadInfo* pThreadInfo,
+    std::shared_ptr<ManagedThreadInfo>& pThreadInfo,
```
chrisnas, line 250: I'm not sure we need to pass the smart pointer (especially by ref so not triggering an AddRef?): we could have kept the signature and pass .get() like for CollectStackSample().
Note that the function only uses .get()

profiler/src/ProfilerEngine/Datadog.Profiler.Native/StackSamplerLoopManager.cpp
```diff
@@ -472,8 +472,7 @@ bool StackSamplerLoopManager::AllowStackWalk(ManagedThreadInfo* pThreadInfo)
         return false;
     }
 
-    pThreadInfo->AddRef();
-    _pTargetThread = pThreadInfo;
+    _pTargetThread = std::move(pThreadInfo);
```
chrisnas, line 475: Since pThreadInfo is copied (i.e. AddRef) when passed to this function, does this move forbid the destructor to Release when exiting the function?

profiler/src/ProfilerEngine/Datadog.Profiler.Native/StackSamplerLoopManager.cpp
```diff
@@ -519,8 +518,7 @@ void StackSamplerLoopManager::NotifyIterationFinished()
     std::lock_guard<std::mutex> guardedLock(_watcherActivityLock);
 
     _pTargetThread->GetStackWalkLock().Release();
-    _pTargetThread->Release();
-    _pTargetThread = nullptr;
+    _pTargetThread.reset();
```
chrisnas, line 521: Is this what will make the nullptr test in line 393 work?
`if (LogDuringStackSampling_Unsafe || _pTargetThread == nullptr)`

profiler/test/Datadog.Profiler.Native.Tests/ManagedThreadListTest.cpp
```diff
@@ -29,15 +29,15 @@ TEST(ManagedThreadListTest, CheckRemove)
     ASSERT_EQ(threads.Count(), 3);
 
     bool found = false;
-    ManagedThreadInfo* pInfo = nullptr;
+    std::shared_ptr<ManagedThreadInfo> pInfo = nullptr;
```
chrisnas, line 32: Is it the same as calling reset()?

chrisnas (approved):
LGTM

profiler/src/ProfilerEngine/Datadog.Profiler.Native/StackSamplerLoop.cpp
gleocadie: Later we call a persisting function and we need a copy. So not possible to pass a raw pointer.

profiler/src/ProfilerEngine/Datadog.Profiler.Native/StackSamplerLoopManager.cpp
gleocadie: This move allows to transfer ownership to the `_pThreadInfo` field. So no additional release.

profiler/src/ProfilerEngine/Datadog.Profiler.Native/StackSamplerLoopManager.cpp
gleocadie: exactly.

gleocadie: Calling reset will call Release(). FYI

profiler/test/Datadog.Profiler.Native.Tests/ManagedThreadListTest.cpp
gleocadie: No, we just initialize the variable to nullptr.

PR #2994, opened 2022-07-20
[Profiler] Deactivate profiler on ARM64

gleocadie:
## Reason for change

The preliminary tests on ARM64 shows we have issue (stackwalking...) To avoid customer hitting the same problem and until we fix those issues, the profiler will be disabled on ARM64/Linux.

### DIFF
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native/DllMain.cpp
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native/DllMain.cpp
@@ -117,6 +117,11 @@ extern "C" HRESULT STDMETHODCALLTYPE DllGetClassObject(REFCLSID rclsid, REFIID r
         return CORPROF_E_PROFILER_CANCEL_ACTIVATION;
     }
 
+#ifdef ARM64
+    Log::Warn("Profiler is deactivated because it runs on an unsupported architecture.");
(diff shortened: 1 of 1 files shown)


chrisnas (approved):
LGTM

profiler/src/ProfilerEngine/Datadog.Profiler.Native/DllMain.cpp
```diff
@@ -44,6 +44,11 @@ extern "C" BOOL STDMETHODCALLTYPE DllMain(HINSTANCE hInstDll, DWORD reason, PVOI
 
 bool CheckProfilingEnabledEnvironmentVariable()
 {
+#ifdef ARM64
+    Log::Warn("Profiler is deactivated because it runs on an unsupported architecture.");
```
kevingosse, line 48: NIT: I think we should evaluate the environment variable like we do in the other platforms, and only log the warning if the result is "profiler should be enabled". This way we won't add non-actionable warnings to customers who didn't enable the profiler anyway

kevingosse (approved):

profiler/src/ProfilerEngine/Datadog.Profiler.Native/DllMain.cpp
sciutand: Maybe should have also updated the official datadog documentation. There is no mention anywhere that arm is not supported.

PR #3046, opened 2022-08-03
[CIVisibility] Change build file and CODEOWNERS

tonyredondo:
## Summary of changes

Changes to reflect the new label name `ci-visibility` and the new codeowners team.

### DIFF
--- a/tracer/build/_build/Build.GitHub.cs
+++ b/tracer/build/_build/Build.GitHub.cs
@@ -444,7 +444,7 @@ partial class Build
             const string buildAndTest = "Build / Test";
             const string misc = "Miscellaneous";
             const string tracer = "Tracer";
-            const string ciApp = "CI App";
+            const string ciVisibility = "CI Visibility";
             const string appSecMonitoring = "ASM";
             const string profiler = "Continuous Profiler";
             const string debugger = "Debugger";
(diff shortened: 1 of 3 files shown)


tracer/build/_build/Build.GitHub.cs
```diff
@@ -535,7 +535,7 @@ LabbelerConfiguration GetLabellerConfiguration()
                 var fixIssues = new[] { "type:bug", "type:regression", "type:cleanup" };
                 var areaLabelToComponentMap = new Dictionary<string, string>() {
                     { "area:tracer", tracer },
-                    { "area:ci-app", ciApp },
+                    { "area:ci-visibility", ciVisibility },
```
pierotibou, line 538: you should change the [labeller config](https://github.com/DataDog/dd-trace-dotnet/blob/master/.github/labeller.yml#L2) as well

pierotibou (approved):

tracer/build/_build/Build.GitHub.cs
tonyredondo: oh, thanks!, done in [47385b2](https://github.com/DataDog/dd-trace-dotnet/pull/3046/commits/47385b26ad61ee06fa901c36b622f48cfbe10c8b)

andrewlock (approved):

PR #3057, opened 2022-08-05
[Profiler] Try to use default UDS path on Linux

kevingosse:
## Summary of changes

When no agent url is provided, the tracer tries to bind to the default UDS path (`/var/run/datadog/apm.socket`), before falling back to the default HTTP endpoint (`http://localhost:8126`).

This PR makes the profiler behavior consistent with the tracer.

## Reason for change

Consistency :yay2:

## Test coverage

:awkward-monkey:

### DIFF
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native/LibddprofExporter.cpp
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native/LibddprofExporter.cpp
@@ -144,11 +144,29 @@ ddprof_ffi_EndpointV3 LibddprofExporter::CreateEndpoint(IConfiguration* configur
     }
     else
     {
-        // agent mode
-        std::stringstream oss;
-        oss << "http://" << configuration->GetAgentHost() << ":" << configuration->GetAgentPort();
-        _agentUrl = oss.str();
+        // Agent mode
+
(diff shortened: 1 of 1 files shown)


chrisnas (approved):
LGTM

profiler/src/ProfilerEngine/Datadog.Profiler.Native/LibddprofExporter.cpp
```diff
@@ -18,6 +18,10 @@
 #include <string.h>
 #include <time.h>
 
+#ifndef _WINDOWS
+#include <filesystem>
+#endif
```
gleocadie, line 23: not needed
this is included by the `dd_filesystem.hpp` line 25

gleocadie (commented):
one nit

profiler/src/ProfilerEngine/Datadog.Profiler.Native/LibddprofExporter.cpp
gleocadie: https://github.com/DataDog/dd-trace-dotnet/blob/master/shared/src/native-src/dd_filesystem.hpp
and you can use `fs::` for the namespace

PR #3058, opened 2022-08-05
[Profiler] Fix generic types name for exception and allocation profilers

chrisnas:
## Summary of changes
Reuse the existing symbol resolution code used by frames for exception and allocated objects types

## Reason for change
The CLR provides generic types name with the `xx syntax instead of <>

## Implementation details
Refactor to allow backend encoding and C# representation


profiler/src/ProfilerEngine/Datadog.Profiler.Native/AllocationsProvider.cpp
```diff
@@ -71,7 +64,13 @@ void AllocationsProvider::OnAllocation(uint32_t allocationKind,
     rawSample.ThreadInfo = threadInfo;
     threadInfo->AddRef();
     rawSample.AllocationSize = objectSize;
-    rawSample.AllocationClass = std::move(shared::ToString(shared::WSTRING(typeName)));
+
+    // The provided type name contains the metadata-based `xx syntax for generics instead of <>
+    // So rely on the frame store to get a C#-like representation like what is done for frames
+    if (!_pFrameStore->GetTypeName(classId, rawSample.AllocationClass))
+    {
+        rawSample.AllocationClass = std::move(shared::ToString(shared::WSTRING(typeName)));
```
gleocadie, line 72: Why calling move here?
From what I see, the return value of `shared::ToString` is a rvalue-ref  so no need to call move on it

gleocadie (approved):
One minor thing

profiler/src/ProfilerEngine/Datadog.Profiler.Native/AllocationsProvider.cpp
chrisnas: ✅ done

PR #3068, opened 2022-08-11
Add a very simplistic snapshots checker

pierotibou:
## Summary of changes
Adding a github action that will print a summary of the diffs in the snapshots in a PR comment if we have more than 50 snapshots modified. The comment would look like something like this:

## Snapshots difference summary
The following differences have been observed in snapshots. So diff is simplistic, so please check some files anyway while we improve it.

950 occurrences of:
```diff
+    attribute: value
   [...]
+        tag:value
```

1 occurrences of:
```diff
+    attribute: value
   [...]
+        tag:value
-        ohboy: there's a bug
```

This shows the place in the hierarchy thanks to the number of spaces. So not great. 

## Reason for change
To start simplifying snapshots review.

## Implementation details
- Diff on snapshots file
- Foreach file check the change
- Count occurences

## Test coverage
Quick local tests

## Other details
I think we clearly need to improve this. It may be too simplistic. A few ideas:
- show the full path eg `tags.tagname: value`, `metrics.metricname: value`
- identify where the change happens: rootspan only, all spans, subset of spans

### DIFF
--- a/tracer/build/_build/Build.GitHub.cs
+++ b/tracer/build/_build/Build.GitHub.cs
@@ -161,8 +161,8 @@ partial class Build
                     markdown.Append("```").AppendLine();
                 }
 
-                Console.WriteLine(markdown.ToString());
-                // await PostCommentToPullRequest(PullRequestNumber.Value, markdown.ToString());
+                // Console.WriteLine(markdown.ToString());
+                await PostCommentToPullRequest(PullRequestNumber.Value, markdown.ToString());
 
                 void RecordChange(StringBuilder diffsInFile, Dictionary<string, int> diffCounts)
                 {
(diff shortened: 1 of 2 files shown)


tracer/build/_build/Build.GitHub.cs
```diff
@@ -86,6 +86,102 @@ partial class Build
...
+                RecordChange(diffsInFile, diffCounts);
+
+                var markdown = new StringBuilder();
+                markdown.AppendLine("## Snapshots difference summary").AppendLine();
+                markdown.AppendLine("The following differences have been observed in snapshots. So diff is simplistic, so please check some files anyway while we improve it.").AppendLine();
+
+                foreach (var diff in diffCounts)
+                {
+                    markdown.AppendLine($"{diff.Value} occurrences of : ");
+                    markdown.AppendLine("```");
```
andrewlock, line 159: If this is printing in diff format (i.e. with `+`/`-` at the start) then you can add the language:

```suggestion
                    markdown.AppendLine("```diff");
```

andrewlock (approved):
:shipit: we can always iterate!

tracer/build/_build/Build.GitHub.cs
pierotibou: :wow: thanks

PR #3089, opened 2022-08-17
[Profiler] Improve error reported by libddprof

gleocadie:
## Reason for change

There was an overflow when retrieving the error string from libddprof. Until now it was not possible to know what was the issue.
 
In addition to that, we forgot to release the `SendResult` which would leak memory.

## Implementation details

Get the the error string and log it
+ add `on_leave` macro to simplify the use of RAII


chrisnas (approved):
LGTM

PR #3090, opened 2022-08-17
[Profiler] Fix Profiler integration tests in VS

gleocadie:
## Reason for change

Profiler integration tests cannot run in VS. Make sure they can run :)

### DIFF
--- a/profiler/test/Datadog.Profiler.IntegrationTests/Helpers/EnvironmentHelper.cs
+++ b/profiler/test/Datadog.Profiler.IntegrationTests/Helpers/EnvironmentHelper.cs
@@ -109,7 +109,7 @@ namespace Datadog.Profiler.IntegrationTests.Helpers
 
         internal string GetProfilerNativeLibraryPath()
         {
-            var profilerHome = GetMonitoringHome();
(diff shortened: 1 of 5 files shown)


chrisnas (approved):
LGTM

profiler/test/Datadog.Profiler.IntegrationTests/Helpers/EnvironmentHelper.cs
```diff
@@ -268,6 +268,26 @@ private static string GetMonitoringHome()
             return s;
         }
 
+        private static string GetProfilerHomeDirectory()
+            => IsRunningInCi()
+                    ? Path.Combine(GetMonitoringHome(), GetArchitectureSubfolder())
+                    : GetDeployDir();
```
andrewlock, line 274: 🙁 I was trying to get rid of the entire concept of "profiler home", as it just adds confusion IMO 🙁

gleocadie: I know, but it prevents the tests from running in VS.
We want the freshly compiled binary in VS used in the integration tests in VS.

PR #3093, opened 2022-08-17
[Profiler] Add new profile tags

chrisnas:
## Summary of changes
When uploading a .pprof , add 
- .NET version as "runtime_version" tag (framework-4.8 or core-6.0 for example)
- OS as "runtime_os" tag (linux or windows)
- list of enabled profilers tag (walltime_cpu_exceptions)

## Reason for change
Figure out which version of .NET is running profiled applications on which OS

## Implementation details
Use ICorProfilerInfo::GetRuntimeInformation() to get major/minor versions with Framework or Core

## Test coverage
Tags generation


profiler/src/ProfilerEngine/Datadog.Profiler.Native/LibddprofExporter.cpp
```diff
@@ -113,9 +118,27 @@ LibddprofExporter::Tags LibddprofExporter::CreateTags(IConfiguration* configurat
...
+    }
+    buffer << "-" << std::dec << runtimeInfo->GetDotnetMajorVersion() << "." << runtimeInfo->GetDotnetMinorVersion();
+    tags.Add("runtime_version", buffer.str());
+
+    // runtime_platform (os and version later)
+#ifdef LINUX
+    tags.Add("runtime_os", "linux");
+#else
+    tags.Add("runtime_os", "windows");
+#endif
```
gleocadie, line 141: maybe add `GetOs` member function to `IRuntimeInfo` and call it from there ?

chrisnas: ✅ done

profiler/src/ProfilerEngine/Datadog.Profiler.Native/LibddprofExporter.cpp
```diff
@@ -125,6 +148,56 @@ LibddprofExporter::Tags LibddprofExporter::CreateTags(IConfiguration* configurat
...
+    }
+
+    if (emptyList)
+    {
+        return "";
+    }
+    else
+    {
+        return buffer.str();
+    }
```
gleocadie, line 197: you can just `return buffer.str()` to simplify, no?

gleocadie (approved):
one comment otherwise LGTM

profiler/src/ProfilerEngine/Datadog.Profiler.Native/LibddprofExporter.cpp
chrisnas: ✅ done

andrewlock (commented):
Are these tags standardised for the profiler, as they differ significantly from the values used in the tracer. 

For the tracer, for example, we send the following headers:

`Datadog-Meta-Lang-Interpreter`: `.NET Framework`/`.NET Core`/`.NET`
`Datadog-Meta-Lang-Version`: `4.8.1`/`2.1`/`6.0`

(we don't track os on spans directly AFAICT).

CI-visibility tags a lot more with span-level tags, e.g:
`runtime.name: `.NET Framework`
`runtime.version`: `4.8.1`
`os.platform`: `Windows`
`os.version`:

If these _aren't_ standardised for profiling, maybe we should look at not introducing a whole new set of descriptions?

chrisnas:
> You should wait for a manual run of the profiler integration tests before merging this - I've just triggered one

Thanks!
I'm struggling with the CI since last week...

> Are these tags standardised for the profiler, as they differ significantly from the values used in the tracer.
> 
> For the tracer, for example, we send the following headers:
> 
> `Datadog-Meta-Lang-Interpreter`: `.NET Framework`/`.NET Core`/`.NET` `Datadog-Meta-Lang-Version`: `4.8.1`/`2.1`/`6.0`
> 
> (we don't track os on spans directly AFAICT).
> 
> CI-visibility tags a lot more with span-level tags, e.g: `runtime.name`: `.NET Framework` `runtime.version`: `4.8.1` `os.platform`: `Windows` `os.version`:
> 
> If these _aren't_ standardised for profiling, maybe we should look at not introducing a whole new set of descriptions?

They are defined among profiler runtimes already.
I don't know why the tracer tags were not used  :^(

PR #3112, opened 2022-08-22
[Profiler] Measure exceptions and allocations profilers

chrisnas:
## Summary of changes
Add dedicated scenarios to measure exception and allocation profilers 

## Reason for change
Need to measure worst case scenario impact for exception and allocation profilers

## Implementation details
Update timeit .json for exceptions and allocations


gleocadie (approved):
LGTM

PR #3121, opened 2022-08-24
[Profiler] Activate the CPU profiling by default

gleocadie:
## Summary of changes

Activate the CPU profiling by default.


chrisnas (approved):
LGTM

PR #3128, opened 2022-08-26
[Profiler] Disable tracer for profiler timeit usage

chrisnas:
## Summary of changes
Disable tracer during timeit profiler runs

## Reason for change
Fix CI issues

## Implementation details
Set DD_TRACE_ENABLED to 0 in json configurations


gleocadie (approved):
LGTM
nit: we can move the definition of `DD_TRACE_ENABLED` in the global env. next PR ;)

PR #3179, opened 2022-09-07
[Profiler] Ensure that all API HRESULT are checked

chrisnas:
## Summary of changes
Check GetClassIDInfo2 return value in all cases

## Reason for change
Crash in the CLR

## Implementation details
Check HRESULT and 0 CLASSID


gleocadie (approved):
LGTM: can't we add a test to make sure that it won't crash for this case in the future ? (in Computer01 for example )

chrisnas:
> LGTM: can't we add a test to make sure that it won't crash for this case in the future ? (in Computer01 for example )

I tried to reproduce by allocating Generic<string> instead of Generic<int> but without success  :^(

PR #3181, opened 2022-09-07
[Profiler] Fix sanitizer jobs

gleocadie:
## Summary of changes

Fix Address sanitizer and Undefined behavior jobs.

## Reason for change

Changes in the profiler silently broke the sanitizers jobs.

## Implementation details

- Add a `DD_SANITIZERS` compilation flag to enable the profiler even if the LD_PRELOAD does not contain the `anti-deadlock` library.
- deactivate the alignment check for `ParseGcEvent` function
- Run other scenarios in the sanitizers tests.


chrisnas (approved):
LGTM

PR #3188, opened 2022-09-07
[Profiler] Make exceptions profiler more real for BuggyBits

chrisnas:
## Summary of changes
Add Parse/TryParse implementations to generate or not exceptions

## Reason for change
Add different implementation to show exceptions impacts

## Implementation details
Update data layer and controller

### DIFF
--- a/profiler/src/Demos/Samples.BuggyBits/Controllers/ProductsController.cs
+++ b/profiler/src/Demos/Samples.BuggyBits/Controllers/ProductsController.cs
@@ -167,7 +167,13 @@ namespace BuggyBits.Controllers
         {
             var sw = new Stopwatch();
             sw.Start();
+
+            // Exception-driven code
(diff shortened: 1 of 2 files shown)


profiler/src/Demos/Samples.BuggyBits/Controllers/ProductsController.cs
```diff
@@ -167,7 +167,13 @@ public IActionResult Sales()
         {
             var sw = new Stopwatch();
             sw.Start();
+
+            // Exception-driven code
             var products = dataLayer.GetProductsOnSale();
+
+            // Fix:
+            //var products = dataLayer.GetProductsOnSaleEx();
```
kevingosse, line 175:
```suggestion
            // var products = dataLayer.GetProductsOnSaleEx();
```

kevingosse (approved):

PR #3208, opened 2022-09-12
Fix WindowsContainer tracing sample

zacharycmontoya:
## Summary of changes
Fix the C# build error and make required changes for datadog-agent to start properly on the latest Windows 11. After the changes, I am able to see traces in Datadog after following the steps in the sample's `README.md`

### DIFF
--- a/tracer/samples/WindowsContainer/docker-compose.yml
+++ b/tracer/samples/WindowsContainer/docker-compose.yml
@@ -13,8 +13,9 @@ services:
       ASPNETCORE_ENVIRONMENT: Production
 
   datadog-agent:
-    image: "gcr.io/datadoghq/agent:7"
+    image: "gcr.io/datadoghq/agent:latest"
     environment:
+      DD_HOSTNAME: "windowscontainer-datadog-agent"
       DD_API_KEY: ${DD_API_KEY}
       DD_APM_ENABLED: "true"
       DD_APM_NON_LOCAL_TRAFFIC: "true"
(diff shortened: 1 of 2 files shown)


pierotibou (approved):

tracer/samples/WindowsContainer/docker-compose.yml
```diff
@@ -13,8 +13,9 @@ services:
       ASPNETCORE_ENVIRONMENT: Production
 
   datadog-agent:
-    image: "gcr.io/datadoghq/agent:7"
+    image: "gcr.io/datadoghq/agent:latest"
```
andrewlock, line 16: Just FYI, these images are synonymous, so not required 🙂

tracer/samples/WindowsContainer/docker-compose.yml
```diff
@@ -13,8 +13,9 @@ services:
...
     environment:
+      DD_HOSTNAME: "windowscontainer-datadog-agent"
```
andrewlock, line 18: These is the "real" issue - it's a recent requirement in the latest version of the agent. We should ensure that all other docker-compose files that use it similarly have a hostname, e.g.: https://github.com/DataDog/dd-trace-dotnet/blob/a72b74fc8e67d8d8a6430628fe8643b3a693d2bc/tracer/samples/NugetDeployment/docker-compose.yml#L7

andrewlock (approved):
Good catch, this was a recent change to the agent to make `DD_HOSTNAME` required IIRC

PR #3246, opened 2022-09-22
[Profiler] Fix throughput tests on Linux

kevingosse:
## Summary of changes

After the structure of the monitoring-home folder was changed, the throughput tests weren't updated. So the linux tests were actually running without the profiler.

### DIFF
--- a/profiler/build/crank/os.profiles.yml
+++ b/profiler/build/crank/os.profiles.yml
@@ -38,18 +38,16 @@ profiles:
         endpoints:
           - "{{ linuxEndpoint }}"
         environmentVariables:
-          COR_PROFILER: "{846F5F1C-F9AE-4B07-969E-05C26BC060D8}"
-          COR_PROFILER_PATH: "{{ linuxProfilerPath }}/{{ commit_hash }}/monitoring-home-linux/Datadog.Trace.ClrProfiler.Native.so"
           CORECLR_PROFILER: "{846F5F1C-F9AE-4B07-969E-05C26BC060D8}"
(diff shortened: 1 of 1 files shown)


andrewlock (approved):
In the tracer throughput tests we have a check in the app that the profiler is loaded when it's expected to be there, and it throws if not. May be a good idea to add the same to the profiler: https://github.com/DataDog/dd-trace-dotnet/blob/94b2d1dde358e9ca3b6affe5c7d7b8a730a53d97/tracer/test/test-applications/throughput/Samples.AspNetCoreSimpleController/Program.cs#L12-L32

kevingosse:
> In the tracer throughput tests we have a check in the app that the profiler is loaded when it's expected to be there, and it throws if not. May be a good idea to add the same to the profiler:

Yeah but the same app is used for tracer/profiler throughput tests, and AFAIK the profiler is not loaded during the tracer throughput tests. So the test would have to recognize what scenario it's running. Definitely doable but that's a bit more work.

andrewlock:
> Yeah but the same app is used for tracer/profiler throughput tests, and AFAIK the profiler is not loaded during the tracer throughput tests. So the test would have to recognize what scenario it's running. Definitely doable but that's a bit more work.

So that test app checks that the `Tracer` is loaded, which happens in _both_ scenarios right? And would have caught this issue. Or do you explicitly not load the tracer in the profiler tests? 

> So the test would have to recognize what scenario it's running. Definitely doable but that's a bit more work.

Would require checking for the `DD_PROFILING_ENABLED` env var, and invoking `Profiler.Instance.Status.IsProfilerReady` via reflection, right? We could add that as a helper to the existing `SampleHelpers`?

chrisnas (approved):
LGTM

PR #3250, opened 2022-09-22
[Snapshots] Diff on base branch

pierotibou:
## Summary of changes
- Doing the snapshots diff on base branch.
- Increased the scope of the diff to debugger and smoke_tests (for a file to be considered in the diff, they have to be in a folder containing `snapshots`)
- Doing the diff only on modified files

## Test coverage
Tested with a child PR.

### DIFF
--- a/tracer/build/_build/Build.GitHub.cs
+++ b/tracer/build/_build/Build.GitHub.cs
@@ -100,16 +100,12 @@ partial class Build
                 // This is a dumb implementation that just show the diff
                 // We could imagine getting the whole context with -U1000 and show the differences including parent name
                 // eg now we show -oldAttribute: oldValue, but we could show -tag.oldattribute: oldvalue
-                var changes = GitTasks.Git($"diff \"{baseCommit}\" -- */*snapshots*/*.*")
+                var changes = GitTasks.Git($"diff --diff-filter=M \"{baseCommit}\" -- */*snapshots*/*.*")
                                    .Select(f => f.Text);
 
-                const int minFiles = 10;
-                var nbSnapshotsModified = changes.Count(f => f.Contains("@@ "));
-                if (nbSnapshotsModified < minFiles)
+                if (!changes.Any())
                 {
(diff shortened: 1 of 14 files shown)


tracer/build/_build/Build.GitHub.cs
```diff
@@ -91,29 +91,23 @@ partial class Build
...
-                GitTasks.Git("fetch origin master:master", logOutput: false);
+                // This assumes that we're running in a pull request, so we compare against the target branch
+                var baseCommit = GitTasks.Git($"merge-base origin/{TargetBranch} HEAD").First().Text;
 
                 // This is a dumb implementation that just show the diff
                 // We could imagine getting the whole context with -U1000 and show the differences including parent name
                 // eg now we show -oldAttribute: oldValue, but we could show -tag.oldattribute: oldvalue
-                var changes = GitTasks.Git("diff master -- tracer/test/snapshots")
+                var changes = GitTasks.Git($"diff \"{baseCommit}\" -- */*snapshots*/*.*")
                                    .Select(f => f.Text);
```
andrewlock, line 104: I have an alternative suggestion to the "exclude count". The intention there (IIUC) is to not flood the PR with comments when new snapshots have been added? In which case, we could limit the analysis to _only_ snapshots that have been modified by adding a `--diff-filter`:

```suggestion
                var changes = GitTasks.Git($"diff --diff-filter=M \"{baseCommit}\" -- */*snapshots*/*.*")
                                   .Select(f => f.Text);
```

This excludes renamed/deleted/added snapshots, so I think you could probably remove the following `nbSnapshotsModified` logic entirely?

andrewlock (approved):
Looks good to me, one suggestion for an improvement

tracer/build/_build/Build.GitHub.cs
pierotibou: Thanks for the suggestion, it's way better. Done in 4a4cb032df0f3a8c9b72787dea79e2b1d07fd682

PR #3273, opened 2022-09-28
[Tracer] Use `Datadog.Trace.Bundle` in `NugetDeployment` samples

pierotibou:
## Summary of changes
Use `Datadog.Trace.Bundle`in `NugetDeployment` samples

## Reason for change
We need to change the samples when we do the release.

## Test coverage
Tested locally with a package from the CI.

## Other details
To be merged after next release

### DIFF
--- a/tracer/samples/NugetDeployment/HttpListenerExample.csproj
+++ b/tracer/samples/NugetDeployment/HttpListenerExample.csproj
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Datadog.Monitoring.Distribution" Version="1.28.2" />
+    <PackageReference Include="Datadog.Trace.Bundle" Version="2.16.0" />
   </ItemGroup>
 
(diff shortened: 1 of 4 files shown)


tracer/samples/NugetDeployment/HttpListenerExample.csproj
```diff
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Datadog.Monitoring.Distribution" Version="1.28.2" />
+    <PackageReference Include="Datadog.Trace.Bundle" Version="2.16.0" />
```
andrewlock, line 11: Should we keep this up to date with the version bump target?

andrewlock (approved):

tracer/samples/NugetDeployment/HttpListenerExample.csproj
pierotibou: I asked myself the question. I don't see the added value honestly as we don't compile or run them. But I'm ok either way honestly

zacharycmontoya (approved):

PR #3288, opened 2022-10-03
[Profiler] Fix deadlock on alpine on thread creation

gleocadie:
## Summary of changes

Prevent from deadlock on alpine

## Reason for change

We knew that `dl_open` and `dl_iterate_phdr` were sharing a lock. But we discovered that on Alpine, `dl_open` and `__pthread_create` share a `ptc` lock (different from the previous one).
Steps:
- The profiler thread interrupts a thread who was calling `__pthread_create`. This thread acquired the `ptc` lock.
- Another thread calls `dl_open` and acquires the lock shared between `dl_open` and `dl_iterate_phdr`. This thread tries acquiring the `ptc` lock but stays blocked.
- The profiler continues its execution and calls `dl_iterate_phdr`, but gets blocked because the lock is already taken by the thread  which called `dl_open`.

## Implementation details

Add new functions to the Wrapper library:
- __pthread_create
- pthread_attr_init
- pthread_getattr_default_np
- pthread_setattr_default_np
- fork


profiler/src/ProfilerEngine/Datadog.Linux.ApiWrapper/functions_to_wrap.c
```diff
@@ -122,3 +120,34 @@ int dladdr(const void* addr_arg, Dl_info* info)
...
+
+#define INHIBIT_SIGNALS_FOR(fn_name)                                            \
+    /* Function pointers to hold the value of the musl libc functions */            \
+    static void (*__real_##fn_name)() = NULL;                                       \
+                                                                                    \
+    void fn_name()                                                                  \
+    {                                                                               \
+        if (__real_##fn_name == NULL)                                               \
+        {                                                                           \
+            __real_##fn_name = dlsym(RTLD_NEXT, #fn_name);                          \
```
chrisnas, line 132: Maybe check that __real_##fn_name is different from nullptr
Would it be possible that a certain distrib does not contains the entry point we want to highjack?

gleocadie: Good point, this is only for alpine

gleocadie: but if it does not exist it will never be called

chrisnas: ✅ ok

chrisnas (approved):
LGTM

kevingosse (approved):

profiler/src/ProfilerEngine/Datadog.Linux.ApiWrapper/functions_to_wrap.c
```diff
@@ -122,3 +120,143 @@ int dladdr(const void* addr_arg, Dl_info* info)
...
+    // restore the previous state for signals
+    pthread_sigmask(SIG_SETMASK, &oldOne, NULL);
+
+    return result;
+}
+
+/* Function pointers to hold the value of the glibc functions */
+static int (*__real_pthread_setattr_default_np)(const pthread_attr_t* attrp) = NULL;
+
+int pthread_setattr_default_np(pthread_attr_t* a)
```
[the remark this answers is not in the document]
gleocadie: good catch

Issue #3209, opened 2022-09-12
Excessive logging from Continuous Profiler


mikesigs:
**Describe the bug**
The log file `DD-DotNet-Profiler-Native-dotnet-1.log` is filled with millions of lines like the following:

```
[2022-09-12 20:24:35.909 | info | PId: 1 | TId: 17] The profile was sent. Success?true, Http code: 200
```

This kind of logging should be reserved for Debug level shouldn't it?

**Runtime environment:**
- Instrumentation mode: automatic with NuGet installer
- Tracer version: 2.14.0
- OS: Debian GNU/Linux 10 (buster)
- CLR: .NET Core 3.1

chrisnas:
Thanks for the feedback @mikesigs.

> **Describe the bug** The log file `DD-DotNet-Profiler-Native-dotnet-1.log` is filled with millions of lines like the following:
> 
> ```
> [2022-09-12 20:24:35.909 | info | PId: 1 | TId: 17] The profile was sent. Success?true, Http code: 200
> ```
> 
> This kind of logging should be reserved for Debug level shouldn't it?

Only errors should be logged indeed.
This change has been added into our backlog

PR #3290, opened 2022-10-03
[Profiler] Change message when exporting profile

gleocadie:
## Summary of changes

This follows the issue https://github.com/DataDog/dd-trace-dotnet/issues/3209

## Reason for change

Even if we log this message once every minute, after long hours of running, the log file is cluttered with this information. 

## Implementation details

From now, we will log the message only if there is an error.


andrewlock (approved):

chrisnas (approved):
LGTM

PR #3298, opened 2022-10-04
[Profiler] Change lock-duration to lock-time

chrisnas:
## Summary of changes
The contention lock duration column is renamed to "lock-time"

## Reason for change
"-duration" suffix cannot be used here due to special backend processing

## Test coverage
Generated .pprof validated with backend


gleocadie (approved):
LGTM

PR #3309, opened 2022-10-05
[Profiler] Log INVOKE and CALL warning message only once

gleocadie:
## Summary of changes

Log message from INVOKE and CALL macros only once per call site.

## Reason for change

Too avoid cluttering the log files.

## Implementation details

Just add a static in the if body.


chrisnas (approved):
LGTM

PR #3313, opened 2022-10-05
[Profiler] Bump libdatadog version to 0.9.0

gleocadie:
## Summary of changes

Bump libdatadog version to 0.9.0.

## Reason for change

Get support for Named pipe and AAS tags.

## Implementation details

- Update nuget package version
- Update cmake file

### DIFF
--- a/profiler/test/Datadog.Profiler.IntegrationTests/ApplicationInfoTest.cs
+++ b/profiler/test/Datadog.Profiler.IntegrationTests/ApplicationInfoTest.cs
@@ -62,6 +62,18 @@ namespace Datadog.Profiler.IntegrationTests
                 text = reader.ReadToEnd();
             }
 
+            /*
+             we want to extract the "tags_profiler" attribute from the http body. The format is as followed:
+             {
+                 "start":"<START DATE>",
+                 "end":"<END DATE>",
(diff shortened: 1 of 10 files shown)


profiler/test/Datadog.Profiler.IntegrationTests/ApplicationInfoTest.cs
```diff
@@ -63,17 +62,19 @@ private static (string ServiceName, string Environment, string Version) ExtractS
                 text = reader.ReadToEnd();
             }
 
-            return (
-                ServiceName: ExtractTag("service", text),
-                Environment: ExtractTag("env", text),
-                Version: ExtractTag("version", text));
-        }
+            var match_tags = Regex.Match(text, "\"tags_profiler\":\"(?<tags>[^\"]*)\"", RegexOptions.Compiled);
```
chrisnas, line 65: Would it be possible to show an example of the expected format as a comment?
Easier to read than regex  :^)

gleocadie: good point :) will do

chrisnas (approved):
LGTM

PR #3314, opened 2022-10-05
[Profiler] Change log level from error to info in StackSamplerLoopManager

gleocadie:
## Summary of changes

Change the level of a log message from Error to Info to report the state of the problem in case we have a Customer issue.

We do not put it to Debug because we want the information information in case of an investigation and this will happen very rarely (and not asking to restart with the Debug flag, and never reproduce it).

### DIFF
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native/StackSamplerLoopManager.cpp
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native/StackSamplerLoopManager.cpp
@@ -231,7 +231,7 @@ void StackSamplerLoopManager::WatcherLoopIteration()
     if (_deadlockInterventionInProgress >= 1)
     {
         _deadlockInterventionInProgress++;
-        Log::Error("StackSamplerLoopManager::WatcherLoopIteration - Deadlock intervention still in progress for thread ", _pTargetThread->GetOsThreadId(),
+        Log::Info("StackSamplerLoopManager::WatcherLoopIteration - Deadlock intervention still in progress for thread ", _pTargetThread->GetOsThreadId(),
                    std::hex, " (= 0x", _pTargetThread->GetOsThreadId(), ")");
         // TODO: Validate that calling resuming again (and again) could unlock the situation.
         // The previous call to ResumeThread failed.


profiler/src/ProfilerEngine/Datadog.Profiler.Native/StackSamplerLoopManager.cpp
```diff
@@ -231,7 +231,7 @@ void StackSamplerLoopManager::WatcherLoopIteration()
     if (_deadlockInterventionInProgress >= 1)
     {
         _deadlockInterventionInProgress++;
-        Log::Error("StackSamplerLoopManager::WatcherLoopIteration - Deadlock intervention still in progress for thread ", _pTargetThread->GetOsThreadId(),
+        Log::Info("StackSamplerLoopManager::WatcherLoopIteration - Deadlock intervention still in progress for thread ", _pTargetThread->GetOsThreadId(),
```
gleocadie, line 234: maybe warning ?

andrewlock: Depends how likely it is to happen, and what we/customer should do about it IMO. If the customer can't do anything (other than get worried), and we would only need to when investigating an escalation, then Info or even Debug makes sense to me?

chrisnas: I would prefer Info : no need to see this log as a potential problem. It will probably happen for overloaded small containers as expected

chrisnas (approved):
LGTM

profiler/src/ProfilerEngine/Datadog.Profiler.Native/StackSamplerLoopManager.cpp
gleocadie: Yep, just discussed with Christophe, and I had a similar reasoning: I do not want Customer/support to be scared because there is a warning with "deadlock" in it.
I would like it to be there so if something happens, it will be a lead.
So we'll keep it info I think

PR #3370, opened 2022-10-19
[Profiler] Save and restore the value of errno in the signal handler

kevingosse:
## Summary of changes

Save and restore the value of errno in the signal handler.

## Reason for change

It seems like the Libunwind can change the value of errno when validating whether an address is valid: https://github.com/libunwind/libunwind/blob/47aef9392723e4816d94d4efda4248180aece7f5/src/x86_64/Ginit.c#L185

This can lead the CLR to preventively crash because it finds an unexpected errono value after calling epoll_wait.


gleocadie (approved):
LGTM

PR #3372, opened 2022-10-19
[Profiler] Samples supports timestamps as label

chrisnas:
## Summary of changes
Store samples timestamp as a label in .pprof

## Reason for change
Before changing the .pprof format with breakdown, support samples timestamp with the current format.

## Implementation details
Samples timestamp is stored with the "timestamp_ns" label in .pprof.
Supported under DD_INTERNAL_PROFILING_TIMESTAMPS_AS_LABEL_ENABLED environment variable with a default value of false.


profiler/test/Datadog.Profiler.IntegrationTests/WindowsOnly/NamedPipeTestcs.cs
```diff
@@ -26,7 +26,14 @@ public NamedPipeTestcs(ITestOutputHelper output)
...
         {
-             new SmokeTestRunner(appName, framework, appAssembly, commandLine: "--scenario 1", _output, TransportType.NamedPipe).RunAndCheck();
+            if (framework != "netcoreapp3.1")
+            {
+                new SmokeTestRunner(appName, framework, appAssembly, commandLine: "--scenario 1", _output, TransportType.NamedPipe).RunAndCheck();
+            }
+            else
+            {
+                // TODO: need to figure out issue in netcore3.1 for x86
+            }
```
kevingosse, line 36: I don't think that's related to the PR?

kevingosse (approved):

PR #3434, opened 2022-11-09
Upgrade build to .NET 7

andrewlock:
## Summary of changes

- Updates the SDK we build with to .NET 7
- Does _not_ update the version we test again

## Reason for change

We need to start building and testing against the latest SDK.

## Implementation details

- Bump the SDK version to 7.0.100
- Use a custom alpine base image so we keep building on alpine:3.14 for now, but use the newer SDK
- Switch to Windows 2022 base images (required for MSBuild version)
- Explicitly install .NET 6 runtime/sdk for testing

## Other details
This won't succeed until
- [ ] Create Windows 2022 VMSS images with .NET 7 installed

Once this is merged, we should 
- [ ] Update the Linux VMSS images to re-cache the new docker images
- [ ] Add tests for .NET 7
- [ ] Fix integrations that are broken in .NET 7 (most of the built-in integrations)
- [ ] Consider if we can stop testing any FX versions

### DIFF
--- a/.azure-pipelines/noop-pipeline.yml
+++ b/.azure-pipelines/noop-pipeline.yml
@@ -7,7 +7,7 @@ pr:
 
 # Global variables
 variables:
-  dotnetCoreSdkLatestVersion: 6.0.100
+  dotnetCoreSdkLatestVersion: 7.0.100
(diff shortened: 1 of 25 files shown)


andrewlock:
Closing for now because this breaks all the other builds

PR #3412, opened 2022-11-03
[Profiler] Add garbage collections in profile

chrisnas:
## Summary of changes
Add Stop the World periods due to garbage collections into profile samples

## Reason for change
Provide information for timeline

## Implementation details
Listen to GC_SUSPEND_EE_BEGIN/ EVENT_GC_RESTART_EE_END + other GC events to compute Stop The World duration

## Test coverage
Added tests

PR #3437, opened 2022-11-09
Revert "[Profiler] Add garbage collections in profile (#3412)"

lucaspimentel:
Revert #3412


zacharycmontoya (approved):
Approving so we can unblock the build break

PR #3438, opened 2022-11-10
Fix builds being broken by "quoted things" in the commit message

andrewlock:
## Summary of changes

Fixes CI being broken when the commit message contains `"` marks

## Reason for change

We want the build to work

## Implementation details

Don't write out the env vars, just pass them through


tonyredondo (approved):

pierotibou (commented):
Thanks for doing it

PR #3478, opened 2022-11-17
[Profiler] Revisit the way Sample object are used

gleocadie:
## Summary of changes

Allow to shared Sample objects around the components.

## Reason for change

`Sample` objects were moveable only until now (no copyable), because we wanted to prevent copies of them. This would have been a performance it (copying callstacks). The `SamplesCollector` is responsible for creating them and releasing them.

But now, we will have provider that will have to keep the `Sample` objects alive for a longer period of time (Heap profiler).

## Implementation details

Allocate a pointer to `Sample` on the heap and shared it with the rest of the component. This will be done by using `std::shared_ptr` smart pointer

## Other details

Cleanup up a bit of the code


profiler/src/ProfilerEngine/Datadog.Profiler.Native/CollectorBase.h
```diff
@@ -112,9 +110,9 @@ class CollectorBase
         return input;
     }
 
-    std::list<Sample> TransformRawSamples(const std::list<TRawSample>& input)
+    std::list<std::shared_ptr<Sample>> TransformRawSamples(std::list<TRawSample>&& input)
```
chrisnas, line 113: Why not keeping passing a const ref list?

profiler/src/ProfilerEngine/Datadog.Profiler.Native/Sample.cpp
```diff
@@ -13,6 +13,7 @@ const std::string Sample::SpanIdLabel = "span id";
 const std::string Sample::ExceptionTypeLabel = "exception type";
 const std::string Sample::ExceptionMessageLabel = "exception message";
 const std::string Sample::AllocationClassLabel = "allocation class";
+const std::string Sample::EndTimestampLabel = "end_timestamp_ns";
```
chrisnas, line 16: Good idea to make it a constant

profiler/src/ProfilerEngine/Datadog.Profiler.Native/Sample.h
```diff
@@ -65,10 +60,29 @@ class Sample
     }
 
     // helpers for well known mandatory labels
-    void SetPid(const std::string& pid);
-    void SetAppDomainName(const std::string& name);
-    void SetThreadId(const std::string& tid);
-    void SetThreadName(const std::string& name);
+    template <typename T>
```
chrisnas, line 63: Why would we need template here?
Shouldn't it be string&?

profiler/test/Datadog.Profiler.Native.Tests/SamplesCollectorTest.cpp
```diff
@@ -55,40 +55,40 @@ class FakeSamplesProvider : public ISamplesProvider
...
+        std::list<std::shared_ptr<Sample>> samples;
         for (int i = 0; i < nbSamples; i++)
         {
             samples.push_back(CreateSample(runtimeId));
         }
         return samples;
     }
 
-    const char* GetName()
+    const char* GetName() override
```
chrisnas, line 91: And 1 less warning!

chrisnas (approved):
LGTM

profiler/src/ProfilerEngine/Datadog.Profiler.Native/CollectorBase.h
gleocadie: mainly semantic reason:
In `GetSamples`, we create the list, pass by `const &` the list to `TransformRawSamples` and then the list is not used.
Here it's just to say `I transfer the ownership of the list to TransformRawSamples because I do not need it anymore`

profiler/src/ProfilerEngine/Datadog.Profiler.Native/Sample.cpp
gleocadie: ;)

profiler/src/ProfilerEngine/Datadog.Profiler.Native/Sample.h
gleocadie: Here `template <typename T>` + `T&&` means universal reference.
Which means that we can call those function and pass any type of reference (r-value, l-value) and we leave the compiler decide what to do:
- R-Value reference case: the object will be move
- L-Value reference case: just pass by ref

This will help us in prevented useless copies (in the r-value ref case)

profiler/test/Datadog.Profiler.Native.Tests/SamplesCollectorTest.cpp
gleocadie: ;)

PR #3503, opened 2022-11-24
[Profiler] Fix Windows Throughput tests

gleocadie:
## Summary of changes

Fix windows throughput tests.

## Reason for change

They are broken: profiler is not correctly attached since the package layout changes.

## Other

I took the opportunity to also adjust cpu test because it was containing walltime too, and I added walltime scenario


andrewlock (approved):
> since the package layout changes

:awkward_horror:

pierotibou (approved):

chrisnas (approved):
LGTM

PR #3512, opened 2022-11-25
[Profiler] Publish symbols and install procdump as postmortem debugger

gleocadie:
## Summary of changes

Publish symbols and install procdump to catch application crash

## Reason for change

When an profiled application crashes, we do not have the memory dump. In order to investigate, we install procdump as a postmortem debugger so we will get a memory dump when the crash happens.

And now, we publish also the symbols to ease the investigation.

## Implementation details

Install procdump as postmortem debugger
Publish symbols


chrisnas (approved):
LGTM

PR #3525, opened 2022-11-30
[Profiler] Disable tiered jit compilation only when Tracer is activated

gleocadie:
## Summary of changes

Disable tiered jit compilation when Tracer is activated to prevent profiler specific tests to fail.

## Reason for change

There profiler integration tests where we check the managed callstacks we collected. If we disable tiered jit compilation, the callstack will be different. Today, we want to disable tiered jit compilation only if the Tracer is activated to avoid crashing.

## Implementation details

Set `COMPlus_TieredCompilation= "0"` only when the tracer is activated.


andrewlock (approved):

chrisnas (approved):
LGTM

PR #3548, opened 2022-12-07
[Profiler] Fix little memory leak

gleocadie:
## Summary of changes

Rely on RAII to ensure releasing of resources (here we use `on_leave` construct.

While doing that, found a memory leak.

## Implementation details

Use `on_leave` construct + clean up the request object.


gleocadie:
Fixed in https://github.com/DataDog/dd-trace-dotnet/pull/4029

PR #3561, opened 2022-12-14
[Profiler] Remove per type allocation

chrisnas:
## Summary of changes
Avoid keeping track of at least one allocation per type

## Reason for change
This will reduce bias for sampling


gleocadie (approved):
LGTM

PR #3565, opened 2022-12-15
[Profiler] Remove 100+ ms threshold for lock contention

chrisnas:
## Summary of changes
Don't always sample lock contention with duration greater or equal to 100ms

## Reason for change
It could swamp the profiles as seen by the new BuggyBits parallel lock scenario

## Test coverage
Adding scenario to BuggyBits


gleocadie (approved):
LGTM

PR #3569, opened 2022-12-16
[Profiler] Remove async call for parallel buggybits

chrisnas:
## Summary of changes
Remove async calls to parallel lock scenario in BuggyBits

## Reason for change
Avoid too many spans and lock events


gleocadie (approved):
LGTM

PR #3576, opened 2022-12-19
[Profiler] Add MetricsRegistry

gleocadie:
## Summary of changes

Add metrics in the profiler to report information on the application but also on the profiler itself

## Reason for change

The goal of this PR is to collect metrics (internal to the profiler but also about the runtime, application....) and send them to the profiling Backend.

## Implementation details

- Add a metrics registry which will be used to collect metrics
- Define different kinds of metrics: Proxy, Sum, MeanMax, Counter

## Test coverage

Add unit tests + an integration test


profiler/src/ProfilerEngine/Datadog.Profiler.Native/SumMetric.cpp
```diff
@@ -0,0 +1,31 @@
...
+{
+    double_t value = 0;
+
+    {
+        std::unique_lock lock(_lock);
+
+        value = std::exchange(_value, value);
+    }
+
+    return std::list<Metric>{ {_name, value} };
```
gleocadie, line 30: for SumMetric, should we add `_sum` suffix ?

chrisnas: > for SumMetric, should we add `_sum` suffix ?

I prefered not to add such a suffix because the accumulation (i.e. sum) is the value of the metric

chrisnas (approved):
LGTM

profiler/src/ProfilerEngine/Datadog.Profiler.Native/SumMetric.cpp
gleocadie: 👍

PR #3623, opened 2023-01-04
[Profiler] Detect and fix CPU overlaps

chrisnas:
## Summary of changes
Avoid counting too much CPU usage

## Reason for change
It happens that the OS is returning more CPU usage than real one (probably due to scheduling data structures optimization - updated each time a thread is scheduled on a core)

## Implementation details
Detect such a situation and account the largest possible CPU usage


gleocadie (approved):
LGTM

PR #3624, opened 2023-01-04
[Profiler] Add numeric labels

chrisnas:
## Summary of changes
Support numeric labels in addition to string labels

## Reason for change
Converting unique numbers (such as timestamp, spans ID, or process ID) into strings increases .pprof size due to the string table.

## Implementation details
Add numeric labels to samples


gleocadie (approved):
LGTM

PR #3638, opened 2023-01-09
[Profiler] Fix using nanoseconds instead of milliseconds

chrisnas:
## Summary of changes
Convert nano-seconds into milli-seconds for CPU sampling

## Reason for change
Fix missing conversion


gleocadie (approved):
LGTM

PR #3642, opened 2023-01-10
[Profiler] Provides an alternative stackwalk

gleocadie:
## Summary of changes

Provide an alternative stackwalk mechanism.

## Reason for change

We would like to compare the old/manual way  and the new way to stackwalk a thread on linux. The goal is to continuously see improvements or degradation, and also propose an alternative to investigate customer issue.

## Implementation details

- Add a new internal env var to select the stackwalk algorithm
- Switch between the two algorithm depending on the env var.
- Add benchmark tests.

### DIFF
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
@@ -19,12 +19,12 @@
 #endif
 
 #include "AllocationsProvider.h"
-#include "ContentionProvider.h"
 #include "AppDomainStore.h"
 #include "ApplicationStore.h"
 #include "ClrEventsParser.h"
 #include "ClrLifetime.h"
 #include "Configuration.h"
+#include "ContentionProvider.h"
 #include "CpuTimeProvider.h"
 #include "EnabledProfilers.h"
 #include "EnvironmentVariables.h"
@@ -64,7 +64,6 @@ Error("unknown platform");
 #else
 #define PROFILER_LIBRARY_BINARY_FILE_NAME WStr("Datadog.Profiler.Native" LIBRARY_FILE_EXTENSION)
 #endif
-
 
 IClrLifetime* CorProfilerCallback::GetClrLifetime() const
 {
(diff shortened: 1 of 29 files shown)


profiler/test/Datadog.Profiler.Native.Tests/LinuxStackFramesCollectorTest.cpp
```diff
@@ -290,7 +293,11 @@ class LinuxStackFramesCollectorFixture : public ::testing::Test
 TEST_F(LinuxStackFramesCollectorFixture, CheckSamplingThreadCollectCallStack)
 {
     auto* signalManager = GetSignalManager();
-    auto collector = LinuxStackFramesCollector(signalManager);
+
+    auto [configuration, mockConfiguration] = CreateConfiguration();
+    EXPECT_CALL(mockConfiguration, UseBacktrace2()).WillOnce(Return(true));
```
chrisnas, line 298: Would it be possible (like in C#) to have the same test to run once with backtrace2 and once with the previous implementation?
That way, both implementations would be tested

chrisnas (approved):
See if it is possible to run Linux stack walk tests with the 2 implementations

profiler/test/Datadog.Profiler.Native.Tests/LinuxStackFramesCollectorTest.cpp
gleocadie: oh good one.
I was adding test for the configuration but forgot this one.

profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
```diff
@@ -794,6 +792,23 @@ void CorProfilerCallback::ConfigureDebugLog()
...
+        auto envVarValue = shared::GetEnvironmentValue(name); \
+        if (!envVarValue.empty())                             \
+        {                                                     \
+            Log::Info("  ", name, ": ", envVarValue);         \
+        }                                                     \
+    }
+
+void CorProfilerCallback::PrintEnvironmentVariables()
+{
+    // TODO: add more env vars values
```
chrisnas, line 806: This is a good idea: we should also provide the command line like the Tracer is doing

chrisnas (approved):
LGTM

PR #3649, opened 2023-01-11
[Profiler] Add a test scenario where short lived threads are created over and over again

chrisnas:
## Summary of changes
Add a test scenario to Computer01 that start short lived threads

## Reason for change
During a customer investigation, the logs has shown a large number of short lived threads.
It is important to reproduce this kind of scenario.

### DIFF
--- a/profiler/src/Demos/Samples.Computer01/QuicklyDeadThreads.cs
+++ b/profiler/src/Demos/Samples.Computer01/QuicklyDeadThreads.cs
@@ -1,0 +1,93 @@
+// <copyright file="QuicklyDeadThreads.cs" company="Datadog">
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2 License.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2022 Datadog, Inc.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Samples.Computer01
+{
+    public class QuicklyDeadThreads
+    {
+        private const int SleepDurationMs = 0;
+        private ManualResetEvent _stopEvent;
+        private List<Task> _activeTasks;
+
+        private const int MaxCreatedThreadCount = 1024;
+
+        public QuicklyDeadThreads()
+        {
+        }
+
+        public void Start()
+        {
+            if (_stopEvent != null)
+            {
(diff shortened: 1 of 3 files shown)


profiler/src/Demos/Samples.Computer01/QuicklyDeadThreads.cs
```diff
@@ -0,0 +1,93 @@
...
+                            Thread.CurrentThread.Name = "QuickDeadThreads";
+                        }
+
+                        while (!_stopEvent.WaitOne(SleepDurationMs))
+                        {
+                            CreateQuicklyDeadThreads();
+                        }
+                    },
+                    TaskCreationOptions.LongRunning),
+            };
```
gleocadie, line 49: why having a list of tasks (`_activeTasks`) for only one task?

profiler/src/Demos/Samples.Computer01/QuicklyDeadThreads.cs
```diff
@@ -0,0 +1,93 @@
...
+            int count = 0;
+            while (!_stopEvent.WaitOne(SleepDurationMs) && (count < MaxCreatedThreadCount))
+            {
+                Thread t = new Thread(() => { count++; });
+                t.Start();
+                t.Join();
+            }
+
+            Console.WriteLine($"   {count} threads were created.");
+        }
```
gleocadie, line 91: 2 things:
- In the case of `Start/Stop`, this method can finish before stop is called. Couldn't we have a mode where we continuously create threads ? 
- The thread creation and run is sequential. Couldn't we have something concurrent ?

gleocadie (approved):
LGTM with comment

profiler/src/Demos/Samples.Computer01/QuicklyDeadThreads.cs
chrisnas: This is due to copy/paste of existing scenario: I noted to refactor them all and build specific base classes to handle the difference cases

profiler/src/Demos/Samples.Computer01/QuicklyDeadThreads.cs
chrisnas: - We have this mode with --iterations and --timeout
- I prefered to avoid concurrent creation and stick to the sequential short lived threads creation that was seen in the logs. I can add the support of --threads on command line

PR #3655, opened 2023-01-12
[Profiler] Allow multiple scenarios to run sequentially in BuggyBits

chrisnas:
## Summary of changes
Pass a bitfield to --scenario in BuggyBits

## Reason for change
It allows different scenarios to run in the same process so in the same profile.


gleocadie (approved):
LGTM

PR #3661, opened 2023-01-13
[Profiler] Fix deadlock due to TLS initialization while unwinding

gleocadie:
## Summary of changes

Fix a malloc-deadlock case when initializing TLS objects while unwinding.

## Reason for change

When calling `unw_backtrace` or `unw_backtrace2`, libunwind will initialize TLS object for the current thread. A TLS initialization is merely a call to `malloc`. But `malloc` is not signal-safe, which means that we can end up in a deadlock situation if we call `malloc` while being in a signal handler.
Example of callstack you might have:

image

As you can see, the thread was calling malloc and then preempted by the signal. Inside libunwind, `(2)` initializes TLS objects and ends up calling `malloc` in `(3)`.

Related issue could be one of the case in https://github.com/DataDog/dd-trace-dotnet/issues/3625.

## Implementation details

- When assigning an OS thread to a managed thread, we call `unw_backtrace` to initialize the TLS'd data structures. It should be safe in that case to call `unw_backtrace`.

## Test coverage

- Add a new scenario to validate that we do not end up in a deadlock situation.

## Other details
Another alternative could have been adding `malloc`, `calloc`, `realloc` and `free` in the [wrapper](https://github.com/DataDog/dd-trace-dotnet/blob/master/profiler/src/ProfilerEngine/Datadog.Linux.ApiWrapper/functions_to_wrap.c) , but we could a significant performance degradation, because at each `malloc` (or other function) call, we would do 2 syscalls (one to prevent signal interruption and another to restore the previous signal mask)

The `pthread_create`-like alternative could have benn a good one: adding a counter to indicate if we are in malloc call.

### DIFF
--- a/profiler/src/Demos/Samples.Computer01/ComputerService.cs
+++ b/profiler/src/Demos/Samples.Computer01/ComputerService.cs
@@ -35,6 +35,7 @@ namespace Samples.Computer01
 #if NET6_0_OR_GREATER
         private LinuxSignalHandler _linuxSignalHandler;
 #endif
+        private LinuxMallocDeadLock _linuxMallockDeadlock;
 
         public void StartService(Scenario scenario, int nbThreads, int parameter)
         {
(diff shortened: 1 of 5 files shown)


chrisnas (approved):
LGTM

PR #3791, opened 2023-02-14
[Build/Test] Publish official dd-lib-dotnet-init images on tagged commit

zacharycmontoya:
## Summary of changes
Update GitLab CI to only publish the dd-lib-dotnet-init images from the GitHub Container Registry to the official Datadog repositories when a tag is added.

Note: This configuration is nearly identical to the Java GitLab pipeline so it seems like pretty reasonable logic.

## Reason for change
This publishes official container images to support Kubernetes Admission Controller library injection for .NET.

## Implementation details
When a commit is tagged, the new GitLab jobs will run in CI. The dd-lib-dotnet-init image with the tags `<SHA>` and `<SHA>-musl` are pulled from the [GitHub container registry](https://github.com/DataDog/dd-trace-dotnet/pkgs/container/dd-trace-dotnet%2Fdd-lib-dotnet-init) and republished to the production Datadog repositories using the pipeline from the DataDog/public-images repo.

## Test coverage
Tested by running the GitLab pipeline on this branch and the new jobs were skipped.

## Other details
N/A

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -95,7 +95,7 @@ deploy_to_docker_registries:
   rules:
     - if: '$POPULATE_CACHE'
       when: never
-    - if: '$CI_COMMIT_TAG =~ /^v.*/'
+    - if: '$CI_COMMIT_TAG =~ /^v[0-9]+\.[0-9]+\.[0-9]+(-prerelease)?$/'
       when: on_success
     - when: manual
       allow_failure: true
@@ -113,7 +113,7 @@ deploy_musl_tag_to_docker_registries:
   rules:
     - if: '$POPULATE_CACHE'
       when: never
-    - if: '$CI_COMMIT_TAG =~ /^v.*/'
+    - if: '$CI_COMMIT_TAG =~ /^v[0-9]+\.[0-9]+\.[0-9]+(-prerelease)?$/'
       when: on_success
     - when: manual
       allow_failure: true
@@ -131,7 +131,7 @@ deploy_latest_tag_to_docker_registries:
   rules:
     - if: '$POPULATE_CACHE'
       when: never
-    - if: '$CI_COMMIT_TAG =~ /^v.*/'
+    - if: '$CI_COMMIT_TAG =~ /^v[0-9]+\.[0-9]+\.[0-9]+(-prerelease)?$/'
       when: on_success
     - when: manual
       allow_failure: true
@@ -149,7 +149,7 @@ deploy_latest_musl_tag_to_docker_registries:
   rules:
     - if: '$POPULATE_CACHE'
       when: never
-    - if: '$CI_COMMIT_TAG =~ /^v.*/'
+    - if: '$CI_COMMIT_TAG =~ /^v[0-9]+\.[0-9]+\.[0-9]+(-prerelease)?$/'
       when: on_success
     - when: manual
       allow_failure: true
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -89,3 +89,75 @@ deploy_to_reliability_env:
     UPSTREAM_PACKAGE_JOB: build
     UPSTREAM_PROJECT_NAME: $CI_PROJECT_NAME
     FORCE_TRIGGER: $DEPLOY_TO_REL_ENV
+
+deploy_to_docker_registries:
+  stage: deploy
+  rules:
+    - if: '$POPULATE_CACHE'
+      when: never
+    - if: '$CI_COMMIT_TAG =~ /^v.*/'
+      when: on_success
+    - when: manual
+      allow_failure: true
+  trigger:
+    project: DataDog/public-images
+    branch: main
+    strategy: depend
+  variables:
+    IMG_SOURCES: ghcr.io/datadog/dd-trace-dotnet/dd-lib-dotnet-init:$CI_COMMIT_SHA
+    IMG_DESTINATIONS: dd-lib-dotnet-init:$CI_COMMIT_TAG
+    IMG_SIGNING: "false"
+
+deploy_musl_tag_to_docker_registries:
+  stage: deploy
+  rules:
+    - if: '$POPULATE_CACHE'
+      when: never
+    - if: '$CI_COMMIT_TAG =~ /^v.*/'
+      when: on_success
+    - when: manual
+      allow_failure: true
+  trigger:
+    project: DataDog/public-images
+    branch: main
+    strategy: depend
+  variables:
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -89,3 +89,75 @@ deploy_to_reliability_env:
     UPSTREAM_PACKAGE_JOB: build
     UPSTREAM_PROJECT_NAME: $CI_PROJECT_NAME
     FORCE_TRIGGER: $DEPLOY_TO_REL_ENV
+
+deploy_to_docker_registries:
+  stage: deploy
+  rules:
+    - if: '$POPULATE_CACHE'
+      when: never
+    - if: '$CI_COMMIT_TAG =~ /^v.*/'
```
pierotibou, line 98: I know I'm paranoid for nothing, but I'd rather have a stricter regex in case we decide to add a "verYBeautifulVersion" tag one day :)

Something like this (I'm the work person ever in regex, so please triple check)

```suggestion
    - if: '$CI_COMMIT_TAG =~ /^v[0-9]*\.[0-9]*\.[0-9]*/'
```

.gitlab-ci.yml
```diff
@@ -89,3 +89,75 @@ deploy_to_reliability_env:
...
+    IMG_SOURCES: ghcr.io/datadog/dd-trace-dotnet/dd-lib-dotnet-init:$CI_COMMIT_SHA
+    IMG_DESTINATIONS: dd-lib-dotnet-init:$CI_COMMIT_TAG
+    IMG_SIGNING: "false"
+
+deploy_musl_tag_to_docker_registries:
+  stage: deploy
+  rules:
+    - if: '$POPULATE_CACHE'
+      when: never
+    - if: '$CI_COMMIT_TAG =~ /^v.*/'
```
pierotibou, line 116: same, change need to change the regex

.gitlab-ci.yml
```diff
@@ -89,3 +89,75 @@ deploy_to_reliability_env:
...
+      when: on_success
+    - when: manual
+      allow_failure: true
+  trigger:
+    project: DataDog/public-images
+    branch: main
+    strategy: depend
+  variables:
+    IMG_SOURCES: ghcr.io/datadog/dd-trace-dotnet/dd-lib-dotnet-init:$CI_COMMIT_SHA
+    IMG_DESTINATIONS: dd-lib-dotnet-init:latest
```
pierotibou, line 144: I think it hasn't been discussed and not done by other teams. Couldn't we add a `v2-latest` tag in addition to latest. Because as of now, our customers can put one version, or latest. But if we deliver a new major, we don't want them to use latest, don't we?

pierotibou (approved):
Thanks. LGTM. I'd just want to change the regex please.
And I have a question around `latest` tag and breaking changes

pierotibou (commented):
Oh and can you impact the release process doc please? https://datadoghq.atlassian.net/wiki/spaces/APMINT/pages/2258338076/Release+Process

andrewlock (approved):

.gitlab-ci.yml
```diff
@@ -89,3 +89,75 @@ deploy_to_reliability_env:
...
+      when: on_success
+    - when: manual
+      allow_failure: true
+  trigger:
+    project: DataDog/public-images
+    branch: main
+    strategy: depend
+  variables:
+    IMG_SOURCES: ghcr.io/datadog/dd-trace-dotnet/dd-lib-dotnet-init:$CI_COMMIT_SHA
+    IMG_DESTINATIONS: dd-lib-dotnet-init:$CI_COMMIT_TAG
```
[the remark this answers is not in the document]
pierotibou: Very good remark. I've checked, other libraries have published with `v`. [Java](https://gallery.ecr.aws/datadog/dd-lib-java-init), [python](https://hub.docker.com/r/datadog/dd-lib-python-init/tags) and [JS](https://hub.docker.com/r/datadog/dd-lib-js-init/tags).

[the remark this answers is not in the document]
zacharycmontoya: From a docker image tag perspective, it makes sense without the `v`, but as Pierre pointed out this is what other libraries are doing so I guess we're stuck doing it that way 😞

.gitlab-ci.yml
zacharycmontoya: This is a good idea. What do you think of keeping the existing tags the way they are (just vX.Y.Z) and we come back to this discussion and possibly add this at a later time?

.gitlab-ci.yml
zacharycmontoya: It's possible we may want to publish a prerelease tags so I think we could include it? I'll make the change as Andrew suggested

.gitlab-ci.yml
```diff
@@ -89,3 +89,75 @@ deploy_to_reliability_env:
...
+    IMG_SOURCES: ghcr.io/datadog/dd-trace-dotnet/dd-lib-dotnet-init:$CI_COMMIT_SHA
+    IMG_DESTINATIONS: dd-lib-dotnet-init:latest
+    IMG_SIGNING: "false"
+
+deploy_latest_musl_tag_to_docker_registries:
+  stage: deploy
+  rules:
+    - if: '$POPULATE_CACHE'
+      when: never
+    - if: '$CI_COMMIT_TAG =~ /^v.*/'
```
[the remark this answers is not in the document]
zacharycmontoya: Done in e87558d547e2ab76b2070f58794614720a66d33c

.gitlab-ci.yml
zacharycmontoya: Done in e87558d547e2ab76b2070f58794614720a66d33c

.gitlab-ci.yml
zacharycmontoya: Done in e87558d547e2ab76b2070f58794614720a66d33c

.gitlab-ci.yml
andrewlock: > This is a good idea. What do you think of keeping the existing tags the way they are (just vX.Y.Z) and we come back to this discussion and possibly add this at a later time?

I think that's fine. Though arguably it would be better to decide on an approach _now_ before customers start using it. That way all the _initial_ implementations they use will be the "safe" version if they want it (`v2`). Otherwise, if they add it later, _every_ customer needs to update their YAML. 

It should probably be a cross team thing though if we do, all tracer languages should have that IMO

zacharycmontoya: Keep in mind Python, Node, and Java are already GA :laughcry:

PR #3803, opened 2023-02-17
[Build/Test] Add system tests for the library injection images

zacharycmontoya:
## Summary of changes
Invokes the system tests to validate the library injection images for .NET

## Reason for change
We need tests for this, and in doing so it found an issue with the existing GHCR images 😅 

## Implementation details
- Removes the `latest` and `latest-snapshot` tags, so now we only build images with the following tags:
  - `dd-lib-dotnet-init:<SHA>`
  - `dd-lib-dotnet-init:<SHA>-musl`
- Fixes an issue where the last `docker build buildx` command would overwrite the tag in GHCR. Since the ARM64 image builds were run last, this resulted in the `dd-lib-dotnet-init:<SHA>` and `dd-lib-dotnet-init:<SHA>-musl` tags referencing only the ARM64-platform image. The fix is to create and upload a new manifest at the end of the build that directly references all of the platform images that belong to that tag.
- Invokes the system-tests to run the lib-injection test suite with the following parameters
  - Test app name: `dd-lib-dotnet-init-test-app`
  - .NET Core runtimes: `7.0-bullseye-slim` and `7.0-alpine`

## Test coverage
The tests succeed! See the latest [GitHub Actions test run](https://github.com/DataDog/dd-trace-dotnet/actions/runs/4199663351/jobs/7284816400)

## Other details
Note: Waiting on the .NET test code to be merged in the system-tests repo via https://github.com/DataDog/system-tests/pull/880

### DIFF
--- a/.github/workflows/lib-injection.yml
+++ b/.github/workflows/lib-injection.yml
@@ -121,6 +121,7 @@ jobs:
         lib-injection-connection: ['network','uds']
         lib-injection-use-admission-controller: ['', 'use-admission-controller']
         weblog-variant: ['dd-lib-dotnet-init-test-app']
+        runtime: ['bullseye-slim','alpine']
       fail-fast: false
     env:
       TEST_LIBRARY: dotnet
@@ -129,11 +130,13 @@ jobs:
       LIBRARY_INJECTION_ADMISSION_CONTROLLER: ${{ matrix.lib-injection-use-admission-controller }}
       DOCKER_REGISTRY_IMAGES_PATH: ghcr.io/datadog
       DOCKER_IMAGE_TAG: ${{ github.sha }}
-      BUILDX_PLATFORMS: linux/amd64,linux/arm64/v8 
+      DOCKER_IMAGE_WEBLOG_TAG: ${{ github.sha }}-${{ matrix.runtime }}
+      RUNTIME: ${{ matrix.runtime }}
+      BUILDX_PLATFORMS: linux/amd64
     steps:
       - name: lib-injection test runner
         id: lib-injection-test-runner
-        uses: DataDog/system-tests/lib-injection/runner@b3ca75b6ce109a349f7390a9f111e6b3ef3c97ef
+        uses: DataDog/system-tests/lib-injection/runner@1282714bfd25baae821f10cf6f1e77f990a82a89 # zach/lib-injection-dotnet
         with:
           docker-registry: ghcr.io
           docker-registry-username: ${{ github.repository_owner }}
--- a/.github/workflows/lib-injection.yml
+++ b/.github/workflows/lib-injection.yml
@@ -48,13 +48,6 @@ jobs:
       id: buildx
       uses: docker/setup-buildx-action@v2
 
-    - name: lib-injection-tags
-      id: lib-injection-tags
-      uses: DataDog/system-tests/lib-injection/docker-tags@main
-      with:
-        init-image-name: 'dd-lib-dotnet-init'
-        main-branch-name: 'master'
-
     - name: Login to Docker
       run: docker login -u publisher -p ${{ secrets.GITHUB_TOKEN }} ghcr.io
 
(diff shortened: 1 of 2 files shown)


.github/workflows/lib-injection.yml
```diff
@@ -48,63 +48,117 @@ jobs:
       id: buildx
       uses: docker/setup-buildx-action@v2
 
-    - name: Generate libc image tags
+    - name: lib-injection-tags
```
pierotibou, line 51: Dumb question but we don't use the output anymore, so do we need to run it?

.github/workflows/lib-injection.yml
```diff
@@ -48,63 +48,117 @@ jobs:
...
+      LIBRARY_INJECTION_ADMISSION_CONTROLLER: ${{ matrix.lib-injection-use-admission-controller }}
+      DOCKER_REGISTRY_IMAGES_PATH: ghcr.io/datadog
+      DOCKER_IMAGE_TAG: ${{ github.sha }}${{ matrix.init-tag-suffix }}
+      DOCKER_IMAGE_WEBLOG_TAG: ${{ github.sha }}-${{ matrix.runtime }}
+      RUNTIME: ${{ matrix.runtime }}
+      BUILDX_PLATFORMS: linux/amd64
+    steps:
+      - name: lib-injection test runner
+        id: lib-injection-test-runner
+        uses: DataDog/system-tests/lib-injection/runner@69336117276b3ef31f2594d20cbd8cd8a7d26396 # zach/lib-injection-dotnet
```
pierotibou, line 159: Looks like something that will break one day :)

pierotibou (approved):
A few questions. And honestly, a bit of a :blindfold: review.

.github/workflows/lib-injection.yml
```diff
@@ -48,63 +48,117 @@ jobs:
...
+        with:
+          docker-registry: ghcr.io
+          docker-registry-username: ${{ github.repository_owner }}
+          docker-registry-password: ${{ secrets.GITHUB_TOKEN }}
+          test-script: ./lib-injection/run-manual-lib-injection.sh
```
andrewlock, line 164: Bit of a shame that we only run the tests after we've already pushed a potentially faulty image.

I guess we don't have a choice though because IIRC you _have_ to push straight away when using buildx? 

Also, means we're not testing the arm64 images I assume 🤔

andrewlock (approved):

.github/workflows/lib-injection.yml
zacharycmontoya: You don't _have_ to push straight away when using buildx, we could remove the `push` argument. Perhaps we could build the images locally, invoke the tests (they run on the same machine), and then publish the images to the repo only after the tests pass? Think that's okay to pursue in a follow-up PR?

And yes, we're only testing amd64 at the moment. I'll have to check with the reliability to see how we might test the ARM64 image

.github/workflows/lib-injection.yml
zacharycmontoya: I had to set this to my dev branch to test the changes, but we can either keep this on a `main` SHA or always use the latest commit of the repo in general.

This is why the PR is `do-not-merge` 😄

.github/workflows/lib-injection.yml
zacharycmontoya: You're absolutely right, I'll get rid of it

.github/workflows/lib-injection.yml
pierotibou: Indeed, I was thinking it's on GHCR so we don't care. 
But we could actually start a release after, and not realize the system tests are broken. It also means we should start being way more careful about a broken build when releasing (sometimes we put the sha to force using the last build even if it has partially failed, it could become riskier to do this now)

.github/workflows/lib-injection.yml
zacharycmontoya: Fixed in 3887ee97cb3dec1b25ab2636f78605526424efcb

.github/workflows/lib-injection.yml
zacharycmontoya: Since I've now merged the system-tests PR, I've updated this to reference the `main` branch 😄

.github/workflows/lib-injection.yml
andrewlock: > Perhaps we could build the images locally, invoke the tests (they run on the same machine), and then publish the images to the repo only after the tests pass? Think that's okay to pursue in a follow-up PR?

Sounds like it would be a good idea to me 👍

PR #3807, opened 2023-02-17
Update telemetry-logs implementation

andrewlock:
## Summary of changes

- Add de-duplication of telemetry logs
- Update minimum log verbosity to Warn
- Add timestamp to log messages (otherwise uses the timestamp in the batch)

## Reason for change

We want to make sure we don't overwhelm the intake. As these are sanitised logs (message templates only) there should be huge savings by deduplicating.

## Implementation details

Uses the message template (+ stack trace if present) to generate a hash. Check if the hash has already been seen before writing the log message). 

Similar to the log-rate-limiting that we already have (disabled by default), we append an "additional messages skipped" count to each log message. 

When a new batch is emitted, we reset the log count. Technically I think there's a small race condition which would cause a duplicate log message to be "dropped" entirely under the following conditions:

- The log message (`MESSAGE`) has been previously seen `3` times (it's present in the `logCounts` dictionary).
- `EnqueueLog` is called at `EmitBatch` with the message `MESSAGE`.
- `EnqueueLog` takes a reference to the dictionary
- `EmitBatch` swaps out the dictionary, and starts processing the log messages in the batch, adding the counts. It updates the log message for `MESSAGE` with the existing count (`3`)
- `EnqueueLog` checks its `logCounts` dictionary, updates the count to `4`, and discards the log
- The "wrong" count is sent

_Ideally_ we would either send the correct count, `4`, or we would emit the log in the next batch. However, I don't think this is a big enough issue to worry about (unless there's an easy solution I'm missing)! 

## Test coverage

Added unit tests for the de-dupe behaviour

## Other details
Awaiting final verification on whether we want to append the the `"{x} additional messages skipped"` message at all. If not, then the above concern is moot and we can skip it. Note that AFAICT, in Java's proposed implementation, they _don't_ add the extra message https://github.com/DataDog/dd-trace-java/pull/4521

### DIFF
--- a/tracer/src/Datadog.Trace/Logging/DirectSubmission/Formatting/EventIdHash.cs
+++ b/tracer/src/Datadog.Trace/Logging/DirectSubmission/Formatting/EventIdHash.cs
@@ -30,13 +30,15 @@ namespace Datadog.Trace.Logging.DirectSubmission.Formatting
     internal static class EventIdHash
     {
         /// <summary>
-        /// Compute a 32-bit hash of the provided <paramref name="messageTemplate"/>. The
+        /// Compute a 32-bit hash of the provided <paramref name="messageTemplate"/>
+        /// and optionally a <paramref name="stackTrace"/>. The
         /// resulting hash value can be uses as an event id in lieu of transmitting the
         /// full template string.
         /// </summary>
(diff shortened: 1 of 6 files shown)


andrewlock:
Closed in favour of #4835

PR #3950, opened 2023-03-24
Docker Base Images. Fix snapshot generation and fix draft releases

robertomonteromiguel:
There was a bug in the pipeline, and we aren't generating the snapshot base images. Fixed. 
When we create a release from "draft release pipeline" we weren't invoking to the docker base image pipeline. Fixed

### DIFF
--- a/.github/workflows/create_draft_release.yml
+++ b/.github/workflows/create_draft_release.yml
@@ -80,3 +80,17 @@ jobs:
         working-directory: ${{steps.assets.outputs.artifacts_path}}
         run: |
           dotnet nuget push "*.${{steps.versions.outputs.full_version}}*.nupkg" --api-key ${{ secrets.NUGET_API_KEY }} --source https://api.nuget.org/v3/index.json
+      
+      - name: "Trigger Docker Base Images Github Pipeline"
+        run: |
+          curl \
+          -X POST \
+          -H "Accept: application/vnd.github+json" \
+          -H "Authorization: token $GITHUB_TOKEN"\
+          -H "X-GitHub-Api-Version: 2022-11-28" \
+          https://api.github.com/repos/DataDog/dd-trace-dotnet/actions/workflows/docker-base-images.yml/dispatches \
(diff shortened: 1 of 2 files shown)


pierotibou:
Thanks for the fix. Do you have a doc somewhere on the process? I'm a bit lost now. For instance I didn't think we would need to impact the release process.

robertomonteromiguel:
> Thanks for the fix. Do you have a doc somewhere on the process? I'm a bit lost now. For instance I didn't think we would need to impact the release process.

We have written this doc:(https://docs.google.com/document/d/1HbMGgZFRnIYjfnXGJGyk0LCwWs2op12R0BE58wma448/edit#heading=h.8d3o7vtyu1y1)

To put you in context, the purpose of "system-tests" is to be able to do e2e tests in a homogeneous way for all Datadog tracer libraries (java, net, js, php...). A current pain point is how to obtain the latest snapshot and release versions of each of the libraries, since each of the teams generates/stores them differently and there are also some security restrictions (each language has its own characteristics).

With this PR we try, from a system-tests point of view, to download the different tracer libraries in the same way, simplifying the complex steps of our pipelines/code/scripts.

This same method has already been applied to PHP and Java. Certainly, it has been complicated, because sometimes each tracer is a different world. To put you in context of other teams:
- Java: GitLab pipeline that runs when we push in master or when we create a release tag.  We take advantage of the previous steps of the pipeline to extract the already compiled artifact from the current workspace.
- PHP: GitLab pipeline that runs periodically every 4 hours and downloads the latest snapshot from CircleCI and the latest release from GitHub.
- dotNet: Azure pipeline invokes the GitHub pipeline when we compile master or build a release. There is also a "draf release" GitHub pipeline that invokes the GitHub pipeline when generating the release.

This is a process that we are starting to apply, any proposal or improvement will be welcome.

pierotibou (approved):
Discussed offline, you test both HEAD and latest release in system tests. Publishing the same binaries with different tags when we release is one way of achieving this.

andrewlock:
IMO this _shouldn't_ happen on draft release. It should happen when the release is created. We've created draft releases previously and then discovered an issue, and abandoned the release.

PR #3994, opened 2023-04-01
[Tracer] Update samples for log collection, agentless logging, and logs trace ID correlation

pierotibou:
## Summary of changes

Added a new PowerShell script `BuildAndRunSample.ps1` that accepts several parameters to choose what sample to run and sets up the desired use case the user wants to test out.
Rewrote the README to provide more information about how to run the samples, how logging libraries are configured, and what to expect as output.
Went through and updated all samples related to log collection for file-tail collection and agentless collection.

Example prompt of the script:

```powershell
pwsh BuildAndRunSample.ps1 -LoggingLibrary Serilog -Framework net7.0 -Runtime x64 -Agentless -ApiKey YOUR_API_KEY_HERE
```

## Reason for change

Provide a set of samples and documentation for customers on how to configure and use the logging features.

## Implementation details

- Script works for Windows (32/64) and Linux (64)
  - Don't have ARM64/Alpine/MacOS, but could add them easily (I think)
- Script sets up all necessary environment variables
- README updated to include sections on each logging library, what the configuration is for them, how to run that sample, and what some basic expected output should be
- Opted to use `Datadog.Trace.Bundle` in all of the samples - I've added a comment about that in the README as it isn't necessarily a requirement, but made things simpler
- Opted to remove all of the various `launchSettings.json` files that were setting environment variables to rely entirely on the running the samples via the script
  - Willing to add these back with additional profiles though

## Test coverage

- Tested all on Windows locally
- Tested a couple on WSL2 64-bit
- Everything showed up as expected in the UI


tracer/samples/AutomaticTraceIdInjection/RunSampleWithAutomaticInstrumentation.sh
```diff
@@ -0,0 +1,40 @@
...
+
+## Setup automatic instrumentation
+## cf https://www.nuget.org/packages/Datadog.Trace.Bundle#readme-body-tab
+## Note that the agent must be configured if you want to receive the traces
+
+## .NET Core
+export CORECLR_ENABLE_PROFILING=1
+export CORECLR_PROFILER={846F5F1C-F9AE-4B07-969E-05C26BC060D8}
+## Replace this path by the appropriate one depending on your system (cf https://www.nuget.org/packages/Datadog.Trace.Bundle#readme-body-tab)
+export CORECLR_PROFILER_PATH=$APP_PATH/datadog/osx/Datadog.Trace.ClrProfiler.Native.dylib
```
andrewlock, line 13: given this is a bash-script we could (should?) probably auto-detect the platform and set this as appropriate?

tracer/samples/AutomaticTraceIdInjection/RunSampleWithAutomaticInstrumentation.sh
```diff
@@ -0,0 +1,40 @@
...
+export DD_DOTNET_TRACER_HOME=$APP_PATH/datadog
+
+## .NET Framework
+# export COR_ENABLE_PROFILING=1
+# export COR_PROFILER={846F5F1C-F9AE-4B07-969E-05C26BC060D8}
+## Replace this path by the appropriate one depending on your system (cf https://www.nuget.org/packages/Datadog.Trace.Bundle#readme-body-tab)
+# export COR_PROFILER_PATH=$APP_PATH\datadog\win-x64\Datadog.Trace.ClrProfiler.Native.dll
+# export DD_DOTNET_TRACER_HOME=<APP_DIRECTORY>/datadog
```
andrewlock, line 21: _Arguably_ there's not much point in having Windows configuration in a bash script. Maybe this should be a powershell script instead (given that runs on all platforms) 😛

andrewlock (approved):

zacharycmontoya (approved):
LGTM. I was able to test this on Windows for .NET Core by manually updating the environment variables and I was able to get logs into the UI

tracer/samples/AutomaticTraceIdInjection/RunSampleWithAutomaticInstrumentation.sh
bouwkast: I've removed this in favor of `BuildAndRunSample.ps1`, but only have Windows/Linux as supported platforms (others could be added easily though to it if that makes sense)

tracer/samples/AutomaticTraceIdInjection/RunSampleWithAutomaticInstrumentation.sh
bouwkast: Made this into a powershell script `BuildAndRunSample.ps1` along with some additional configuration

PR #4227, opened 2023-06-06
Update benchmarks Ubuntu and Mac versions

bouwkast:
## Summary of changes

This updates the Ubuntu and Mac versions used in the pipeline to match what we use in the `master` branch.

## Reason for change

The pipeline for this `benchmarks/2.9.0` was failing.

## Implementation details

- Andrew told me what to look for and do to try and fix this :)
- I copied the versions used in `master` and did a find/replace

## Test coverage

- If CI works then I would say that it is good

## Other details

Unsure if we do PRs to this `benchmarks/2.9.0` branch or just push directly?

### DIFF
--- a/.azure-pipelines/ultimate-pipeline.yml
+++ b/.azure-pipelines/ultimate-pipeline.yml
@@ -374,7 +374,7 @@ stages:
     timeoutInMinutes: 60 #default value
     dependsOn: [ ]
     pool:
-      vmImage: macOS-10.15
+      vmImage: macos-11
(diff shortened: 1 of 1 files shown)


andrewlock:
Oops, I accidentally merged this by pushing to the `benchmarks/2.9.0` branch 😂 

CI is still broken for this branch, but given that we only need it for running throughput tests, I don't think we should spend any time on it (hence my merge). As long as we have some throughput results, that's good enough IMO

PR #4286, opened 2023-06-19
[Release] Deploy AAS test apps twice a week to monitor memory leaks

pierotibou:
## Summary of changes
Deploy AAS test apps twice a week to monitor memory leaks

## Reason for change
AAS seems to be way more stable than Rel Env memory wise. So it should be easier to catch some memory leaks there.

## Implementation details
- Adds a composite action to avoid duplication
- Adds a workflow on schedule to trigger the AAS deployment. This one checks whether the code freeze is in place before updating to not interfere with releases.

## Test coverage
None so far. I need to merge this first version in master to be able to run it from a branch (I could do it from a fork but looks like an unnecessary hassle)

### DIFF
--- a/.github/workflows/code_freeze_start.yml
+++ b/.github/workflows/code_freeze_start.yml
@@ -9,7 +9,7 @@ jobs:
   start_code_freeze:
     if: | 
       github.event_name == 'workflow_dispatch' || 
-      (github.event.milestone.title == 'Code Freeze' && github.event.milestone.state == 'open') 
+      (github.event.milestone.title == 'Code Freeze' && github.event.milestone.state == 'open')
     runs-on: ubuntu-latest
     env:
       GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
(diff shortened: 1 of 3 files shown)


.github/workflows/code_freeze_start.yml
```diff
@@ -42,10 +42,7 @@ jobs:
...
 
-      - name: Trigger AAS deploy
-        run: |
-          curl -X POST \
-          -H "Accept: application/vnd.github.v3+json" \
-          -H "Authorization: Bearer ${{ secrets.GH_TOKEN_PIERO }}" \
-          https://api.github.com/repos/DataDog/datadog-aas-extension/dispatches \
-          -d '{"event_type": "dd-trace-code-freeze", "client_payload": {"sha":"'"$GITHUB_SHA"'" } }'
\ No newline at end of file
+      - uses: ./.github/actions/deploy-aas-dev-apps
```
andrewlock, line 45: Is this the right path - i.e. it uses the path from the root directory, not the current workflow directory?

andrewlock (approved):
LGTM, thanks!

.github/workflows/code_freeze_start.yml
pierotibou: I think it is as I did the same for the code freeze action and it worked. But we'll know soon

PR #4339, opened 2023-07-03
[Release] Move git tag creation back to before creating the release

pierotibou:
## Summary of changes

Create git tag before creating the release.

## Reason for change

Andrew suspects this won't work as GitHub releases are specifically based on tags, so either:
The release step will fail, because the tag doesn't exist
The tag step will fail, because the release step will create the tag, and the tag step will find the duplicate

## Implementation details
It wasn't mandatory for the package publication  so it should be safer to rollback


andrewlock (approved):
Thanks!

PR #4340, opened 2023-07-03
Fix aspects order

daniel-romano-DD:
## Summary of changes
Sorted the aspects list

## Reason for change
Sometimes, aspects could appear in different order, causing CI to fail


pierotibou (approved):

tonyredondo (approved):

tonyredondo (commented):
I think you should regenerate the aspects files.

PR #4349, opened 2023-07-04
Fix system-tests lib-injection

robertomonteromiguel:
Fix system-tests lib-injection

## Summary of changes

Fix system-tests lib-injection pipeline.
The last changes on lib-injection includes more verbosity in the traces when a test fails, but we need to update pipeline add env variable (system-tests/lib-injection contains this variable from some time ago)

### DIFF
--- a/.github/workflows/lib-injection.yml
+++ b/.github/workflows/lib-injection.yml
@@ -146,6 +146,7 @@ jobs:
       DOCKER_IMAGE_WEBLOG_TAG: ${{ github.sha }}-${{ matrix.runtime }}
       RUNTIME: ${{ matrix.runtime }}
       BUILDX_PLATFORMS: linux/amd64
+      MODE: manual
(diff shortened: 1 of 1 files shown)


.github/workflows/lib-injection.yml
```diff
@@ -146,6 +146,7 @@ jobs:
       DOCKER_IMAGE_WEBLOG_TAG: ${{ github.sha }}-${{ matrix.runtime }}
       RUNTIME: ${{ matrix.runtime }}
       BUILDX_PLATFORMS: linux/amd64
+      MODE: manual
```
pierotibou, line 149: Do you have more details on what it fixes and how?

robertomonteromiguel: I have just added the details in the summary

pierotibou (approved):

PR #4405, opened 2023-07-14
Add known limitations from method_rewriter.cpp to docs

bouwkast:
## Summary of changes

Copy pasted the known limitations of `CallTarget` auto instrumentation into the `AutoInstrumentation` docs

## Reason for change

Useful to have in the docs as it is also documented in `method_writer.cpp`

## Implementation details

- Added a TOC
- Added a `Current Limitations` section with the current limitations

## Test coverage

N/A

### DIFF
--- a/docs/development/AutomaticInstrumentation.md
+++ b/docs/development/AutomaticInstrumentation.md
@@ -272,7 +272,7 @@ There are some current limitations with what types/methods with our `CallTarget`
 1. Static methods in a ValueType (struct) cannot be instrumented.
 2. Generic ValueTypes (struct) cannot be instrumented.
 3. Nested ValueTypes (struct) inside a Generic parent type will not expose the type instance (the instance willbe always null).
-4. Nested types (reference types) inside a Generic parent type will not expose the type instance (the instancewill be casted as an object type).
-5. Methods in a Generic type will not expose the Generic type instance (the instance will be casted as a nongeneric base type or object type).
+4. Nested types (reference types) inside a Generic parent type will not expose the type instance (the instance will be casted as an `object` type).
+5. Methods in a Generic type will not expose the Generic type instance (the instance will be casted as a nongeneric base type or `object` type).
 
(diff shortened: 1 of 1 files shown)


docs/development/AutomaticInstrumentation.md
```diff
@@ -249,4 +263,16 @@ Interface methods can be instrumented by setting the `TypeName` to the name of t
...
+```
+
+### Current Limitations
+
+There are some current limitations with what types/methods with our `CallTarget` approach (refer to [method_rewriter.cpp](https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/src/Datadog.Tracer.Native/method_rewriter.cpp#L69) for more information):
+
+1. Static methods in a ValueType (struct) cannot be instrumented.
+2. Generic ValueTypes (struct) cannot be instrumented.
+3. Nested ValueTypes (struct) inside a Generic parent type will not expose the type instance (the instance willbe always null).
+4. Nested types (reference types) inside a Generic parent type will not expose the type instance (the instancewill be casted as an object type).
```
andrewlock, line 275:
```suggestion
4. Nested types (reference types) inside a Generic parent type will not expose the type instance (the instance will be casted as an `object` type).
```

docs/development/AutomaticInstrumentation.md
```diff
@@ -249,4 +263,16 @@ Interface methods can be instrumented by setting the `TypeName` to the name of t
...
+5. Methods in a Generic type will not expose the Generic type instance (the instance will be casted as a nongeneric base type or object type).
```
andrewlock, line 276:
```suggestion
5. Methods in a Generic type will not expose the Generic type instance (the instance will be casted as a nongeneric base type or `object` type).
```

andrewlock (approved):
Perfect, thanks! 🙏

pierotibou (approved):
Thanks a lot Steven

docs/development/AutomaticInstrumentation.md
bouwkast: Thanks!

docs/development/AutomaticInstrumentation.md
bouwkast: Thanks!

PR #4425, opened 2023-07-20
[Builds] Update Reports comments to be updates

duncanista:
## Summary of changes

Update report comments to be updated in a unique comment.
Instead of creating a new one and minimizing the old one.

## Reason for change

Report comments follow the current process:
1. Create comment.
2. Minimize previous comments, if they exist.

This is could be the cause why this repo takes so long to load PRs. Due to the amount of widgets being generated on the background.

The new process should just create a comment if none has been created, and update the existing one if new changes arise.

## Implementation details

Report comments now are created only if there is no previous comment with the same title. 
An update is done on the first instance of the comment with the same name.

## Test coverage

Screenshots of a PR using these changes.
Screenshot 2023-07-20 at 11 02 42 PM
Screenshot 2023-07-20 at 11 02 46 PM

PR: #4427


tracer/build/_build/Build.GitHub.cs
```diff
@@ -1120,48 +1088,80 @@ async Task HideCommentsInPullRequest(int prNumber, string prefix)
...
+        
+            Console.WriteLine($"Found {prComments.Count} comments for PR {prNumber}");
 
-            var count = 0;
-            foreach (var issueComment in issueComments)
+            var updated = false;
+            foreach (var prComment in prComments)
             {
-                if (issueComment.IsMinimized || ! issueComment.Body.StartsWith(prefix))
+                if (! prComment.Body.StartsWith(title))
```
andrewlock, line 1100: nit: I think we could probably tweak this loop slightly. Instead of looping through _all_ comments, we could stop as soon as we have updated one.

If we go with that approach, we would also need to skip any hidden comments, so that we only update the "visible comment". 

I think a simple approach would be to include `comment.IsMinimized` in the selector above, and just skip it if it's minimized.

```suggestion
                if (prComment.IsMinimized || !prComment.Body.StartsWith(title))
```

While I'm writing this, I'm wondering if it's worth the hassle (especially if we have _two_ visible comments due to a previous bug or something) so really don't worry about this if you don't agree / can't be bothered 😆

tracer/build/_build/Build.GitHub.cs
```diff
@@ -1120,48 +1088,80 @@ async Task HideCommentsInPullRequest(int prNumber, string prefix)
...
                     var mutation = new Mutation()
-                                  .MinimizeComment(arg)
-                                  .Select(x => new { x.MinimizedComment.IsMinimized });
+                                  .UpdateIssueComment(arg)
+                                  .Select(x => new { x.IssueComment.Id });
 
                     await connection.Run(mutation);
-                    count++;
+                    updated = true;
+                    Console.WriteLine($"Updated comment {prComment.Id} for PR {prNumber}");
```
andrewlock, line 1120: If you do decide to go with the suggestion above:
```suggestion
                    Console.WriteLine($"Updated comment {prComment.Id} for PR {prNumber}");
                    break;
```

andrewlock (approved):
LGTM, thanks! :shipit:

tracer/build/_build/Build.GitHub.cs
duncanista: Great suggestion on the stopping looping.

That could happen when a bug happens, so it's technically possible. I'll add it, since it doesn't hurt.

andrewlock: Awesome - once it's tested, I would say feel free to merge it, no need to wait for a second review on this one 🙂

PR #4446, opened 2023-07-26
Fix `create_draft_release` when not specifying a specific commit

andrewlock:
## Summary of changes

Add step to calculate sha to use in `create_draft_release` GH Action

## Reason for change

The existing script assumes you will always force a specific commit, but that's not required

## Implementation details

Added a small bash script to choose whether to use the "default" sha or the forced one, and reuse the output throughout the script

## Test coverage

Did some (safe) test runs to confirm the correct value is used when doing forced version ([here](https://github.com/DataDog/dd-trace-dotnet/actions/runs/5667140799/job/15355270557)) and non-forced version ([here](https://github.com/DataDog/dd-trace-dotnet/actions/runs/5667155733/job/15355314955))

### DIFF
--- a/.github/workflows/create_draft_release.yml
+++ b/.github/workflows/create_draft_release.yml
@@ -20,6 +20,17 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Set SHA
+        id: set_sha
+        run: |
+          if [ -z "${{ github.event.inputs.forced_commit_id }}" ]; then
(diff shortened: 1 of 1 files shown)


pierotibou (approved):
LGTM thanks.
As discussed offline, I think we should remove this from the critical path of the release and do it a posteriori, with a release trigger. We can discuss with R&P if needed

PR #4452, opened 2023-07-27
[Release] Remove the extra CI variables

pierotibou:
## Summary of changes

Removed `CI_COMMIT_TAG` and `CI_COMMIT_SHA`

## Reason for change

It broke the release process by overriding the CI variables even when not set. 
Also it is not mandatory to declare them as we can override them anyway.

## Test coverage
Ran this [pipeline](https://gitlab.ddbuild.io/DataDog/dd-trace-dotnet/-/jobs/303371221) by adding an extra job that outputs the overridden variables.

## Other details
The temporary job is visible in the intermediate commit.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -200,3 +200,15 @@ package:
   stage: deploy
   variables:
     PRODUCT_NAME: auto_inject-dotnet
+
+temporary_test:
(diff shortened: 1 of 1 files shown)


andrewlock (approved):
LGTM, thanks

PR #4463, opened 2023-07-28
Adds sql parameters as span tags

deleteLater:
This PR allows us to include SQL parameter details in our trace


deleteLater:
Never mind, wrong PR.

PR #4480, opened 2023-08-02
[Profiler] Add Sample value type provider

gleocadie:
## Summary of changes

Add Sample Value Type provider .

## Reason for change

The goal is to be able to reuse, validate Sample Value Types (for example `timeline`).

## Implementation details

- Add `SampleValueTypeProvider` which is in charge of bookkeeping  the sample values types defined and used by the profilers.
- Pass it around.
- 
## Test coverage

Add Unit tests

### DIFF
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native/SampleValueTypeProvider.cpp
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native/SampleValueTypeProvider.cpp
@@ -12,7 +12,7 @@ SampleValueTypeProvider::SampleValueTypeProvider()
     _sampleTypeDefinitions.reserve(16);
 }
 
-std::vector<SampleValueTypeProvider::Offset> SampleValueTypeProvider::Register(std::vector<SampleValueType> const& valueTypes)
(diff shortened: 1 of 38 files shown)


profiler/src/ProfilerEngine/Datadog.Profiler.Native/SampleValueTypeProvider.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include <cassert>
+#include <tuple>
+
+
+SampleValueTypeProvider::SampleValueTypeProvider()
+{
+    _sampleTypeDefinitions.reserve(16);
+}
+
+std::vector<SampleValueTypeProvider::Offset> SampleValueTypeProvider::Register(std::vector<SampleValueType> const& valueTypes)
```
chrisnas, line 15: Rename to GetOrRegister to avoid confusion (at least in my mind ;^)

chrisnas (approved):
LGTM (with a rename)

profiler/src/ProfilerEngine/Datadog.Profiler.Native/SampleValueTypeProvider.cpp
gleocadie: done

PR #4556, opened 2023-08-25
Add a GH action for forcing required checks for version-bump PR to success

andrewlock:
## Summary of changes

Allow forcing the checks on the version-bump PR to `success`

## Reason for change

The version-bump PR is created automatically. That means that the standard GitHub Actions _don't_ run automatically, even though they're required for merging. This allows a way around the issue without using admin-mode

## Implementation details

Create a new manual-dispatch workflow that forces the checks. Only allowed on `version-bump*` PRs.

## Test coverage

Can't test this until _after_ I merge it to master 🙄

### DIFF
--- a/.github/workflows/override_version_bump_pr_checks.yml
+++ b/.github/workflows/override_version_bump_pr_checks.yml
@@ -1,0 +1,38 @@
+name: Override version bump PR checks
+
+on:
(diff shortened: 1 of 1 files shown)


robertpi (approved):
Seems like it could work 😉

PR #4716, opened 2023-10-11
Fix various GitHub Actions

andrewlock:
## Summary of changes
- Fix the start code freeze action
- Fix the version bump action
- Fix the rel env release image creation actions

## Reason for change

They were all broken/incomplete

## Implementation details

- The code freeze action wasn't opening the code_freeze milestone like it was supposed to
- The version bump action now expects changes in dd-dotnet.sh
- The rel env release image creation action wasn't expanding a variable correctly

## Test coverage

Nope

### DIFF
--- a/.github/workflows/code_freeze_start.yml
+++ b/.github/workflows/code_freeze_start.yml
@@ -18,6 +18,18 @@ jobs:
       - name: Clone dd-trace-dotnet repository
         uses: actions/checkout@v3
 
+      - uses: octokit/request-action@v2.x
+        name: 'Open Code Freeze Milestone'
+        id: milestones
+        if: github.event_name == 'workflow_dispatch'
+        with:
+          route: PATCH /repos/{owner}/{repo}/milestones/115
+          owner: DataDog
+          repo: dd-trace-dotnet
+          state: open
+        env:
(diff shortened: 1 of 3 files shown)


.github/workflows/code_freeze_start.yml
```diff
@@ -18,6 +18,18 @@ jobs:
       - name: Clone dd-trace-dotnet repository
         uses: actions/checkout@v3
 
+      - uses: octokit/request-action@v2.x
+        name: 'Open Code Freeze Milestone'
+        id: milestones
+        if: github.event_name == 'workflow_dispatch'
+        with:
+          route: PATCH /repos/{owner}/{repo}/milestones/115
```
anna-git, line 26: 115 is hardcoded?

anna-git (approved):
just little comment on 115 being hardcoded but it's probably just my ignorance :XD

.github/workflows/code_freeze_start.yml
andrewlock: Yup :D it's a specific milestone that we open and close to indicate when a code freeze is in place 🙂 Yes, it's very hacky 😂

PR #4744, opened 2023-10-23
Handle potential `NullReferenceException` in DSM

andrewlock:
## Summary of changes

Add nullable annotations and fix the warnings

## Reason for change

A recent escalation noted an incidental `NullReferenceException` in the logs:

```
[ERR] Error creating or populating scope. System.NullReferenceException: Object reference not set to an instance of an object.
at Datadog.Trace.ClrProfiler.AutoInstrumentation.Kafka.KafkaHelper.GetMessageSize[T](T message)
at Datadog.Trace.ClrProfiler.AutoInstrumentation.Kafka.KafkaHelper.CreateConsumerScope(Tracer tracer, DataStreamsManager dataStreamsManager, Object consumer, String topic, Nullable`1 partition, Nullable`1 offset, IMessage message)
```

The `GetMessageSize()` method is quite careful about null-checking, so my best guess is that the `message` object passed in itself is `null` (which the nullable annotations also suggested).

I think this is partly due to the difference in how duck typing works for duck-typing in instrumentations (where the duck-type object will _never_ be `null`, as it's implemented as a `struct`, even if using interfaces) and when using duck-chaining (where the duck-type object _can_ be `null` if you're using interfaces)

## Implementation details

Add `#nullable enable`, add some null checks

## Test coverage

Not easy, just checking there's no regressions

## Other details

Maybe we should make it a habit of marking duck-chained interface properties as nullable reference types? 🤔 I think that would help avoid this ambiguity?

PR #4761, opened 2023-10-25
Update duck-typing documentation with best-practice recommendations

andrewlock:
## Summary of changes

- Update docs on how errors happen in duck-typing
- Update docs about how nullability works in duck-typing
- Add recommendations for working with `null` in duck-typing

## Reason for change

[We recently had a bug in an integration](https://github.com/DataDog/dd-trace-dotnet/pull/4744) causing a `NullReferenceException` due to differences in how `null` works depending on how you do duck-typing. That's embarrassing. We shouldn't have `NullReferenceException` 😬 

## Implementation details

This is just a documentation update for now, but there's a recommendation at the end for concrete steps we can take to avoid/find/fix these.

## Other details

Pro tip - use "Display the Rich Diff" to review the docs as they're rendered instead of trying to grok the markdown 🙂 

If everyone agrees with the recommendations at the end, we should prioritize applying this to our existing integrations to chase down any other potential `NullReferenceException`s

### DIFF
--- a/docs/development/DuckTyping.md
+++ b/docs/development/DuckTyping.md
@@ -5,7 +5,7 @@ The duck typing library allows us to get and set data from fields and properties
 The goal of the library is to have an unify code to access unknown types as fastest and with the minimum allocations as possible.
 
 > [!NOTE]  
-> Please ensure you check the [Best Practices](#best-practices) section for details on how to choose the type of 
+> Please ensure you check the [Best Practices](#best-practices) section for details on how to choose the type of proxy.
 
 ### Example
 Given the following scenario, where we want to access the data from an anonymous class instance in another method, a code example to do that would be:
@@ -686,7 +686,7 @@ In the above example, the `ProxyMyHandlerStruct.Configuration` field is of type 
 
 What's more, the `[DuckCopy]` `struct` does _not_ implement `IDuckType`. So with the above duck-types, there's no way to detect when `originalObject.Configuration` is `null`! This is a problem, as it could mean accidentally using the `default` value for `MaxConnections` (i.e. `0`) when in practice the `Configuration` property was `null`. 
 
-The solution to this problem, is to always define duck-chained `[DuckCopy` `struct`s as nullable:
+The solution to this problem is to always define duck-chained `[DuckCopy]` `struct`s as nullable:
 
 ```csharp
 [DuckCopy]
@@ -750,7 +750,7 @@ internal static CallTargetState OnMethodBegin<TTarget, TMyProxy>(TTarget instanc
     if(DuckType.HasValue(proxy))
     {
         // safe to use proxy - underlying type is not null
-        if(DuckType.HasValue(proxy.Configuration)
+        if(DuckType.HasValue(proxy.Configuration))
         {
             // safe to use duck-chained property - underlying type is not null
         }
@@ -758,8 +758,7 @@ internal static CallTargetState OnMethodBegin<TTarget, TMyProxy>(TTarget instanc
 }
 ```
 
-Yes this may be a bit annoying, but the consistently will make it easier to be confident about the nullability of the type, it helps the compiler with its nullability flow analysis, and it makes it easier to review. Note that it's important you also follow the other recommended steps however. 
-
+Yes, this may be a bit annoying, but the consistency will make it easier to be confident about the nullability of the type, it helps the compiler with its nullability flow analysis, and it makes it easier to review. Note that it's important you also follow the other recommended steps, however. 
 
 > Note, the above method has not yet been added. Added to the documentation for now to get feedback, and will implement and test in a subsequent PR if people agree. I intend to implement it as the following: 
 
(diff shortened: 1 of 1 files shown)


tonyredondo (approved):
Nice! thanks!

docs/development/DuckTyping.md
```diff
@@ -4,6 +4,9 @@ The duck typing library allows us to get and set data from fields and properties
 
 The goal of the library is to have an unify code to access unknown types as fastest and with the minimum allocations as possible.
 
+> [!NOTE]  
+> Please ensure you check the [Best Practices](#best-practices) section for details on how to choose the type of
```
zacharycmontoya, line 8: > how to choose the type of

type of proxy?

docs/development/DuckTyping.md
```diff
@@ -491,6 +498,284 @@ AMD Ryzen 9 5950X, 1 CPU, 32 logical and 16 physical cores
...
+Yes this may be a bit annoying, but the consistently will make it easier to be confident about the nullability of the type, it helps the compiler with its nullability flow analysis, and it makes it easier to review. Note that it's important you also follow the other recommended steps however. 
+
+
+> Note, the above method has not yet been added. Added to the documentation for now to get feedback, and will implement and test in a subsequent PR if people agree. I intend to implement it as the following: 
+
+```csharp
+public class DuckType
+{
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    internal static bool IsNull<T>([NotNullWhen(false)] T proxy)
```
zacharycmontoya, line 770: Should we add a generic constraint that T implements IDuckType?

docs/development/DuckTyping.md
```diff
@@ -491,6 +498,284 @@ AMD Ryzen 9 5950X, 1 CPU, 32 logical and 16 physical cores
...
+
+
+```csharp
+internal static CallTargetState OnMethodBegin<TTarget, TMyProxy>(TTarget instance, TMyProxy proxy)
+    where TMyProxy : IMyProxy, IDuckType
+{
+    if(DuckType.HasValue(proxy))
+    {
+        // safe to use proxy - underlying type is not null
+        if(DuckType.HasValue(proxy.Configuration)
```
zacharycmontoya, line 753: nit

```suggestion
        if(DuckType.HasValue(proxy.Configuration))
```

docs/development/DuckTyping.md
```diff
@@ -491,6 +498,284 @@ AMD Ryzen 9 5950X, 1 CPU, 32 logical and 16 physical cores
...
+var chainedProxy = proxy.Configuration
+```
+
+As before, if `originalObject` is `null` you can test that directly, and similarly `proxy` will be `null` (because `DuckCast<T>` returns `T?`, i.e. a nullable). Where things are different is with a duck-chained `[DuckCopy]` struct.
+
+In the above example, the `ProxyMyHandlerStruct.Configuration` field is of type `ProxyMyHandlerConfiguration`, which is a `struct`. Therefore _it can never be `null`_. If the underlying `originalObject.Configuration` property was null, `proxy.Configuration` will be the `default` value, so `proxy.Configuration.MaxConnections == 0`! 
+
+What's more, the `[DuckCopy]` `struct` does _not_ implement `IDuckType`. So with the above duck-types, there's no way to detect when `originalObject.Configuration` is `null`! This is a problem, as it could mean accidentally using the `default` value for `MaxConnections` (i.e. `0`) when in practice the `Configuration` property was `null`. 
+
+The solution to this problem, is to always define duck-chained `[DuckCopy` `struct`s as nullable:
```
zacharycmontoya, line 689:
```suggestion
The solution to this problem, is to always define duck-chained `[DuckCopy]` `struct`s as nullable:
```

zacharycmontoya (approved):
Thanks for the thorough documentation! I've left a couple of small comments but LGTM

docs/development/DuckTyping.md
lucaspimentel: also remove the comma in "The solution to this problem is to..."

docs/development/DuckTyping.md
```diff
@@ -491,6 +498,284 @@ AMD Ryzen 9 5950X, 1 CPU, 32 logical and 16 physical cores
...
+        // safe to use proxy - underlying type is not null
+        if(DuckType.HasValue(proxy.Configuration)
+        {
+            // safe to use duck-chained property - underlying type is not null
+        }
+    }
+}
+```
+
+Yes this may be a bit annoying, but the consistently will make it easier to be confident about the nullability of the type, it helps the compiler with its nullability flow analysis, and it makes it easier to review. Note that it's important you also follow the other recommended steps however.
```
lucaspimentel, line 761:
```suggestion
Yes, this may be a bit annoying, but the consistency will make it easier to be confident about the nullability of the type, it helps the compiler with its nullability flow analysis, and it makes it easier to review. Note that it's important you also follow the other recommended steps, however. 
```

I have a friend who is a published fiction author and he doesn't use any commas at all. It drives me crazy 😅

lucaspimentel (approved):

docs/development/DuckTyping.md
andrewlock: No, because not all proxies _do_ implement `IDuckType` (e.g. `DuckCopy` structs) 😄  The goal with this was to have something that works for all cases.

I'm still on the fence as to whether this is worth it or not. I like the consistency, but it's a bit annoying if you know what you're doing 🤔

PR #4824, opened 2023-11-07
Vendor-in GitHub actions

andrewlock:
## Summary of changes

Vendors in the external GitHub Actions we currently use

## Reason for change

We're considering changing the allowed GitHub actions to be `Allow enterprise, and select non-enterprise, actions and reusable workflows`. As long as we also enable actions by GitHub and verified creators, then there are only 2 to vendor

## Implementation details

We currently use these: 
- `github` and `actions` (GitHub created)
- `octokit`, `advanced-security`, `docker` (verified creators)
- `DataDog/system-tests/lib-injection/runner@main` (would need to add exception)

Additionally, we have 

- `peter-evans/create-pull-request@v3.10.0`
- `softprops/action-gh-release@v1`

which are vendored in this PR

## Test coverage

Yeah... easier said than done. I guess I should try to test these, but it's a _real_ pain...

## Other details

Instead of vendoring these, we could just add them to the exceptions list?


bouwkast:
We can consider doing this still, but we have gotten approval for both of the external ones

peter-evans/create-pull-request
softprops/action-gh-release

https://datadoghq.atlassian.net/browse/LANGPLAT-865

PR #4810, opened 2023-11-03
Stop pushing to the reliability environment

andrewlock:
## Summary of changes

- Stop pushing to S3
- Stop pushing container images for the reliability environment

## Reason for change

The reliability environment is not reliable, so we aren't monitoring it any more, and have switched to testing in Azure instead

## Implementation details

Deleted the `upload_to_s3` stage in the pipeline, and the `create-rel-env-docker-base-images.yml` workflow

## Test coverage

N/A

## Other details

Saves me having to "fix" the workflow for next major version bump

PR #4862, opened 2023-11-14
Stop pushing system-test base images again

andrewlock:
## Summary of changes

Revert the revert of the delete + clean up some more stuff

## Reason for change

So, in #4810 I deleted a bunch of scripts and CI stuff that I thought were related to the reliability environment. I partially reverted that in #4858 because we thought we reused those images in the system-tests, but it appears we don't.

## Implementation details

Revert the revert and delete some more scripts I discovered

## Test coverage

Hah

## Other details

Who wants to bet there'll be a "revert the revert the revert" PR 🤔


andrewlock:
Well, that was quicker than I expected

PR #4892, opened 2023-11-18
Bump Sep to 0.3.0

nietras:
@andrewlock 

first of all thank you for choosing and using Sep. Been wondering why Sep 0.2.2 has seen quite a lot of downloads and I have a suspicion 😉 Perhaps it's time to upgrade to 0.3.0 if you'd like. There should be no breaking changes and only improvements, none of consequence to your usage though, I think.


nietras:
CI appears to fail to unrelated actions key issue, not sure if I could do anything. :)

andrewlock:
Thanks @nietras! Yeah, we really need to look into more caching of our packages for builds 🤣

nietras:
> Yeah, we really need to look into more caching of our packages for builds 🤣

Oh noo! 🤣 That will reduce my download numbers by 99% 😅

andrewlock (approved):
Thanks again for this 🙏

PR #4919, opened 2023-11-23
Increase timeout benchmarks

anna-git:
## Summary of changes

Revert to older encoder makes benchmarks much slower, or timeout. Increase timeout for now


andrewlock (approved):
Wow, now that's a timeout 😂

pierotibou (approved):

PR #4936, opened 2023-11-29
Tiny cleanup of static `Tracer` instance usages

andrewlock:
## Summary of changes

- Use the static `Tracer` instance associated with a span in `OtelHelpers`
- Create some "internal" versions of methods that we need for the initial implementation of v3

## Reason for change

While tracking down flakiness in a manual instrumentation sample I'm working on, I noticed that we're accessing the global tracer version. _Technically_ that may have different settings than the tracer used to create the span, and we should use that instead AFAIK.

The "internal" methods are just to make it easier to incrementally work on the v3-public-api changes. Moving those changes here will reduce the footprint of the v3 PRs when they land.

## Implementation details

We expose the tracer instance that created the span at `Span.Context.TracerContext.Tracer`.

The other changes are just creating "internal" versions of methods

## Test coverage

Should be covered by existing.

### DIFF
--- a/tracer/src/Datadog.Trace/Activity/OtlpHelpers.cs
+++ b/tracer/src/Datadog.Trace/Activity/OtlpHelpers.cs
@@ -61,8 +61,9 @@ namespace Datadog.Trace.Activity
             }
 
             // Fixup "version" tag
-            if (Tracer.Instance.Settings.ServiceVersionInternal is null
-                && span.GetTag("service.version") is { Length: > 1 } otelServiceVersion)
+            var tracer = span.Context.TraceContext.Tracer;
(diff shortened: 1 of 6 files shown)


bouwkast (approved):
Looks good to me!

tracer/src/Datadog.Trace/Activity/OtlpHelpers.cs
```diff
@@ -61,8 +59,10 @@ private static void AgentConvertSpan<TInner>(TInner activity, Span span)
             }
 
             // Fixup "version" tag
-            if (Tracer.Instance.Settings.ServiceVersionInternal is null
```
lucaspimentel, line 64: Good catch.

PR #4962, opened 2023-12-08
Stop caching exceptions in lazy

robertpi:
## Summary of changes

The create the `Obfuscator` could result in an exception. As the `Obfuscator` is wrapped in an `Lazy<T>`, if the creation fails, the exception will be cached and all clients trying to retrieve the `Obfuscator` will receive an exception instead.

## Implementation details

In the case where the creation of the `Obfuscator` we create a simpler obfuscator that just redacts everything.

## Test coverage

Covered by existing test.

### DIFF
--- a/tracer/src/Datadog.Trace/Util/Http/QueryStringObfuscation/ObfuscatorFactory.cs
+++ b/tracer/src/Datadog.Trace/Util/Http/QueryStringObfuscation/ObfuscatorFactory.cs
@@ -14,9 +14,14 @@ namespace Datadog.Trace.Util.Http.QueryStringObfuscation
 
         internal static ObfuscatorBase GetObfuscator(double timeoutInMs, string pattern, IDatadogLogger logger, bool reportQueryString = true)
         {
-            if (string.IsNullOrEmpty(pattern) || !reportQueryString)
+            if (string.IsNullOrEmpty(pattern))
             {
                 return new NullObfuscator();
+            }
+
+            if (!reportQueryString)
+            {
+                return new RedactAllObfuscator();
             }
 
             try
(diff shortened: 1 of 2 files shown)


lucaspimentel (approved):

tracer/src/Datadog.Trace/Util/Http/QueryStringObfuscation/ObfuscatorFactory.cs
```diff
@@ -10,9 +10,24 @@ namespace Datadog.Trace.Util.Http.QueryStringObfuscation
...
     internal class ObfuscatorFactory
     {
+        private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor<ObfuscatorFactory>();
+
         internal static ObfuscatorBase GetObfuscator(double timeoutInMs, string pattern, IDatadogLogger logger, bool reportQueryString = true)
-            => string.IsNullOrEmpty(pattern) || !reportQueryString
-                   ? new NullObfuscator()
-                   : new Obfuscator(pattern, TimeSpan.FromMilliseconds(timeoutInMs), logger ?? DatadogLogging.GetLoggerFor(typeof(QueryStringManager)));
+        {
+            if (string.IsNullOrEmpty(pattern) || !reportQueryString)
```
anna-git, line 17: it was a mistake from before, but `if !reportQueryString`, then maybe we want to return the new `RedactAllObfuscator` instead? 
this bug shouldn't happen as in the `QueryStringManager` there's a guard for `!reportQueryString` but it's still weird that !reportQueryString leads to a non obfuscated query string..

andrewlock (approved):
LGTM, though I agree with Anna's comment about the reportQueryString flag

tracer/src/Datadog.Trace/Util/Http/QueryStringObfuscation/ObfuscatorFactory.cs
robertpi: Fixed in cc3961402

robertpi: In fact it's in f380181ae

anna-git (approved):

PR #5003, opened 2023-12-26
speedup CI step verify_files_without_nullability

vandonr:
I'm pretty sure building the whole project is not necessary when we just want to check that the nullability csv is built properly.
We can skip directly to regenerating the file, which reduces the time this step takes from ~14 min to a couple seconds


robertpi (approved):
🎉

kevingosse (approved):
I don't think the restore step is needed either

vandonr:
> I don't think the restore step is needed either

Oh you think so ? There is a dependency on Nuke because we manipulate Nuke objects, but also, this is running inside nuke, so the stuff that's needed might already be there. I'll try.

andrewlock:
> There is a dependency on Nuke

FWIW, Nuke is built and restored separately, which is why this works without the restore. Thanks!

PR #5177, opened 2024-02-12
Default to .net core runtime on Linux in dd-dotnet

kevingosse:
## Summary of changes

When inspecting an app published as a single file, we currently can't detect the runtime (that's because `libcoreclr.so` is embedded in the file, so we don't see it when listing the imported libraries). We can somewhat mitigate the issue by defaulting on .NET Core (instead of .NET Framework) when running on Linux.

## Reason for change

The subsequent environment variable checks depend on the detected runtime, so they're all wrong when running against a single-file app on Linux.

## Test coverage

Added a test for the fallback.

## Other details

Noticed that `ProcessBasicCheck.TracingWithBundle` was calling `Process.GetProcessById` despite already having all the information it needed.


andrewlock (approved):
LGTM

zacharycmontoya (approved):
LGTM, thanks for the fix!

PR #5200, opened 2024-02-15
Tweak pool metrics

andrewlock:
Potentially easiest to review commit-by-commit

## Summary of changes

- Rename non-pooling things to be "Allocator"
- Consolidate metrics into single gauge with multiple tags (has same cardinality)
  - Means we can easily extend to use in other locations later if we want
- Moved the instance-tracking into the pools themselves, incase we create a pool in some other part of dd-trace. Inlined the factory at that point, as it has very little logic and is ASM-specific

TODO: Once we are happy, we need to update dd-go with the dotnet_metrics.json changes


daniel-romano-DD (approved):
Thanks a lot

PR #5249, opened 2024-02-29
Force update to latest Octokit version

andrewlock:
## Summary of changes

Forces Nuke to use the latest version of Octokit which fixes a critical bug

## Reason for change

[There's a critical bug in Octokit](https://github.com/octokit/octokit.net/issues/2889) that prevents it working. [The bug is in Octokit.GraqhQL too](https://github.com/octokit/octokit.graphql.net/issues/311), but that actually doesn't appear to impact us.

## Implementation details

Manually reference the package with the fix

## Test coverage

Tested locally, it works :phew:

### DIFF
--- a/tracer/build/_build/_build.csproj
+++ b/tracer/build/_build/_build.csproj
@@ -35,6 +35,9 @@
     <PackageReference Include="Nuke.Common" Version="6.3.0" ExcludeAssets="build" />
     <PackageReference Include="Colorful.Console" Version="1.2.15" />
     <PackageReference Include="Octokit.GraphQL" Version="0.1.8-beta" />
+
+    <!-- Explicitly overide the version of octokit used by Nuke -->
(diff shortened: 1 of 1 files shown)


kevingosse (approved):

tracer/build/_build/_build.csproj
```diff
@@ -35,6 +35,9 @@
     <PackageReference Include="Nuke.Common" Version="6.3.0" ExcludeAssets="build" />
     <PackageReference Include="Colorful.Console" Version="1.2.15" />
     <PackageReference Include="Octokit.GraphQL" Version="0.1.8-beta" />
+
+    <!-- Explicitly overide the version of octokit used by Nuke -->
```
kevingosse, line 39:
```suggestion
    <!-- Explicitly override the version of octokit used by Nuke -->
```

PR #5383, opened 2024-04-02
K8s Lib Injection: Migration

robertomonteromiguel:
## Summary of changes
Migrate the system-tests lib injection test to the new approach
## Reason for change
Lib Injection tests were build using bash scripting, new tests use kubernetes python client

### DIFF
--- a/.github/workflows/lib-injection-test.yml
+++ b/.github/workflows/lib-injection-test.yml
@@ -37,12 +37,12 @@ jobs:
       - name: Install runner
         uses: ./.github/actions/install_runner 
 
-      - name: Build weblog latest base images
+      - name: Build weblog base images
         env:
-          APP_DOCKER_IMAGE_REPO: ghcr.io/datadog/system-tests/${{ matrix.variant.weblog-variant }}
+          APP_DOCKER_IMAGE_REPO: ghcr.io/datadog/system-tests/${{ env.WEBLOG_VARIANT }}
         run: |
           cd lib-injection/build/docker/$TEST_LIBRARY/$WEBLOG_VARIANT 
-          LIBRARY_INJECTION_TEST_APP_IMAGE=$APP_DOCKER_IMAGE_REPO:latest ./build.sh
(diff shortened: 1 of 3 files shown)


.github/workflows/lib-injection-test.yml
```diff
@@ -31,20 +23,58 @@ jobs:
...
       DOCKER_REGISTRY_IMAGES_PATH: ghcr.io/datadog
       DOCKER_IMAGE_TAG: ${{ inputs.commit_id }}${{ inputs.runtime == 'alpine' && '-musl' || '' }}
       DOCKER_IMAGE_WEBLOG_TAG: ${{ inputs.commit_id }}-${{ inputs.runtime }}
       RUNTIME: ${{ inputs.runtime }}
       BUILDX_PLATFORMS: linux/amd64
-      MODE: manual
     steps:
-      - name: lib-injection test runner
-        id: lib-injection-test-runner
-        uses: DataDog/system-tests/lib-injection/runner@main
```
andrewlock, line 45: I wonder, is there any way to encapsulate all this in a github action like was being done before? 🤔 Seems like a maintenance headache to have to go through and update this across all repos if something changes?

andrewlock (approved):
:blindfold::approval:

I guess we'll have to wait till it's merged to find out if it actually works?

.github/workflows/lib-injection-test.yml
robertomonteromiguel: > I wonder, is there any way to encapsulate all this in a github action like was being done before? 🤔 Seems like a maintenance headache to have to go through and update this across all repos if something changes?

You are right. Same case for system-tests and parametric. I have decided to discard this option for the moment, since the composite we were using for lib-injection, had some references to the main branch and this also caused us problems.
But yes, in the future it should be the way to go, to build common and reusable actions from all repositories (dd-trace-XYZ)...

robertomonteromiguel:
> :blindfold::approval:
> 
> I guess we'll have to wait till it's merged to find out if it actually works?

I tested using an Azure build Id from another Job.
https://github.com/DataDog/dd-trace-dotnet/actions/runs/8534442104

PR #5401, opened 2024-04-04
Use docker mirror image in GitLab instead of dockerhub

andrewlock:
## Summary of changes

Use the mirrored image in registry.ddbuild.io instead of from dockerhub

## Reason for change

We're not supposed to pull directly from dockerhub, as we get rate limited

## Implementation details

Set up mirroring [here](https://github.com/DataDog/images/pull/5063).

## Test coverage

[Tested here](https://gitlab.ddbuild.io/DataDog/dd-trace-dotnet/-/jobs/478182263). 

## Other details

I hoped this would be faster as it's pulling from the internal repository. It does not appear to be.


lucaspimentel (approved):
Nice

PR #5481, opened 2024-04-22
test: Lint produced deb and rpm packages

spencergilbert:
## Summary of changes

The latest stable version of the packaging template now includes a tool `dd-pkg`. This PR uses this tool to lint the produced DEB and RPM packages.

## Reason for change

This lint has been run at the time of package promotion, this change moves this step to the left and lets teams identify issues _before_ they're trying to release to production.

## Other details

[BARX-245](https://datadoghq.atlassian.net/browse/BARX-245)

[BARX-245]: https://datadoghq.atlassian.net/browse/BARX-245?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -166,6 +166,7 @@ package:
       allow_failure: false
   script:
     - ../.gitlab/build-deb-rpm.sh
+    - find . -iregex '.*\.\(deb\|rpm\)' -printf '%f\0' | xargs -0 dd-pkg lint
   variables:
     ARCH: amd64
 
@@ -179,6 +180,7 @@ package-arm:
       allow_failure: false
   script:
     - ../.gitlab/build-deb-rpm.sh
+    - find . -iregex '.*\.\(deb\|rpm\)' -printf '%f\0' | xargs -0 dd-pkg lint
   variables:
     ARCH: arm64
 


spencergilbert:
> Does this _only_ apply to the lib injection deb/rpm? Can/should it be applied to the deb/rpm that we distribute as part of our normal release? If so, how should we obtain the `dd-pkg` tool in those cases, given we build them in Azure Devops rather than GitLab? Thanks!

Hey! I wasn't aware of the non-injection packages, there's no reason we couldn't lint additional packages. The `dd-pkg` is publicly accessible from S3, so assuming there's nothing blocking network requests from the Azure Devops side - it should be straightforward.

[This commit](https://github.com/DataDog/observability-pipelines-worker/commit/0fb5e5de2b548047cd1d23d49276b3dc3994a7a5) has the install and linting changes for a job that isn't using the template from `auto_inject`.

spencergilbert:
It seems like the two failures are due to my PR coming in from a fork - not clear to me why the rest of those required jobs don't have a status.

robertpi:
superseded by: https://github.com/DataDog/dd-trace-dotnet/pull/5516

PR #5494, opened 2024-04-24
[CI] Check build log for errors in a single, separate stage

andrewlock:
## Summary of changes

Moves the checking of the build logs for errors to a separate stage

## Reason for change

Occasionally we see flake in the Check build logs stage. If this happens due to flake in a required stage (integration_tests_windows for example), then you need to re-run the whole job, which takes a long time. If the error is due to flake (i.e. the tests passed, the error in the logs doesn't matter), then this adds a big delay

## Implementation details

Remove the `CheckBuildLogsForErrors` step at the end of all the integration test stages. Instead, add an extra stage that runs after they've all finished which checks _all_ the logs. As this stage is not a _required_ stage, it doesn't block merging

## Test coverage

Did a manual test to confirm it works as expected. Otherwise, this is the tst.

## Other details

I'm not sure whether we actually _want_ to do this or not 🤔 
- For rare but "expected" messages, we want to add them to the exceptions list explicitly. We will still do that, this change just means you don't get blocked by a "new" error, before it gets added to the list.
- If the issue is retries due to network flake, and we recover, maybe this is an indication the log _shouldn't_ be an error/warning?
- There is a risk people merge _before_ the check logs stage has run, and therefore miss a real issue. This is my biggest concern about the approach.

### DIFF
--- a/.azure-pipelines/ultimate-pipeline.yml
+++ b/.azure-pipelines/ultimate-pipeline.yml
@@ -1343,9 +1343,6 @@ stages:
       condition: succeededOrFailed()
       continueOnError: true
 
-    - script: tracer\build.cmd CheckBuildLogsForErrors
-      displayName: Check logs for errors
-
 - stage: integration_tests_windows_debugger
   condition: >
     and(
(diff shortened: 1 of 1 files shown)


andrewlock:
Meh, we probably don't actually want to do this

PR #5516, opened 2024-04-30
test: Lint produced deb and rpm packages

robertpi:
## Summary of changes

The latest stable version of the packaging template now includes a tool `dd-pkg`. This PR uses this tool to lint the produced DEB and RPM packages.

## Reason for change

This lint has been run at the time of package promotion, this change moves this step to the left and lets teams identify issues _before_ they're trying to release to production.

## Other details

[BARX-245](https://datadoghq.atlassian.net/browse/BARX-245)

[BARX-245]: https://datadoghq.atlassian.net/browse/BARX-245?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Originally opened here:
- https://github.com/DataDog/dd-trace-dotnet/pull/5481

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -67,7 +67,7 @@ publish:
           try {
               # The grants option at the end is used to allow public access on the files we upload as the acls only aren't enough.
               aws s3 cp artifacts/ s3://dd-windowsfilter/builds/tracer/${CI_COMMIT_SHA} --recursive --region us-east-1 --exclude "*" --include "*.zip" --include "*.msi" --grants read=uri=http://acs.amazonaws.com/groups/global/AllUsers full=id=3a6e02b08553fd157ae3fb918945dd1eaae5a1aa818940381ef07a430cf25732
-              If ($LASTEXITCODE -eq 0) { 
+              If ($LASTEXITCODE -eq 0) {
                 return
               }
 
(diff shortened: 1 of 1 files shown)


andrewlock (dismissed):

andrewlock (commented):
Just to confirm, have you done a test run of this in GitLab? Because AFAICT, those are currently only produced when we actually do a release, which is a really dangerous time to be doing the linting...

Could we tweak this so we always produce those packages (and lint them) on merges to master at least, as that way we know the release isn't going to be blocked?

That will require some significant CI changes I think however, something @robertomonteromiguel is looking at I believe

spencergilbert:
@andrewlock I think that's a fair concern, but technically that's the state of things today - the failure just happens in `agent-release-management` when the repo requests a promotion of the packages.

I think building (and not publishing) the packages regularly is a better path in the long term.

robertpi:
have agreed more infrastructure work is required before this can be built correctly.

spencergilbert:
@robertpi will the team re-add this once they've made adjustments to how the packages are built?

robertpi:
@spencergilbert for sure! just trying to keep the number of open PRs to a minimum (even if we have some way to go on this)

PR #5728, opened 2024-06-24
Add explicit permissions to all workflows

andrewlock:
## Summary of changes

- Adds explicit permissions to all workflows

## Reason for change

It's a best practice to explicitly specify required permissions for all workflows, and soon we'll have to

## Implementation details

Worked through each workflow, trying to work out what it does and what permissions it needs

## Test coverage

I wish

## Other details

Please don't just 🙈 approve this, it needs 👀 because it has the potential to break at critical times otherwise! The workflows that run as part of standard PRs should be fine, but anything that's rare/release specific needs properly understanding

Fixes VULN-8204

### DIFF
--- a/.github/workflows/code_freeze_end.yml
+++ b/.github/workflows/code_freeze_end.yml
@@ -11,6 +11,11 @@ jobs:
       github.event_name == 'workflow_dispatch' 
       || (github.event.milestone.title == 'Code Freeze' && github.event.milestone.state == 'closed')
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read # Fetches PRs
+      issues: write # Opens milestones
+      statuses: write # add a commit status check
     env:
       GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
 
--- a/.github/workflows/override_version_bump_pr_checks.yml
+++ b/.github/workflows/override_version_bump_pr_checks.yml
@@ -6,6 +6,9 @@ on:
 jobs:
   override_checks:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      statuses: write # Update status checks
 
     steps:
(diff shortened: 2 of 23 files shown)


.github/workflows/code_freeze_end.yml
```diff
@@ -51,6 +56,13 @@ jobs:
         github_token: ${{ secrets.GITHUB_TOKEN }}
         end_freeze: "true"
 
+    - uses: ./.github/actions/code-freeze
```
andrewlock, line 59: Not related to permissions, but you know, we have a lot of PRs now 🙈

.github/workflows/override_version_bump_pr_checks.yml
```diff
@@ -6,12 +6,15 @@ on:
...
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      statuses: write # Update status checks
 
     steps:
-      - name: Fail if branch is not version-bump PR
-        if: ${{ !startsWith(github.ref, 'refs/heads/version-bump-') }}
+      - name: Fail if branch is not version-bump PR or bot PR
+        if: ${{ !startsWith(github.ref, 'refs/heads/version-bump-') && !startsWith(github.ref, 'refs/heads/bot/') }}
```
andrewlock, line 15: Fixes a minor annoyance that the version-bump PRs can't be merged unless you rebase/force-merge, because the checks don't automatically run

lucaspimentel (approved):

.github/workflows/code_freeze_end.yml
andrewlock: Moved to https://github.com/DataDog/dd-trace-dotnet/pull/5734

andrewlock:
Converted to draft, as we should check to see what the monitor comes up with before committing to this

daniel-romano-DD (approved):
Nothing obviously wrong, but I'm no expert here

PR #5805, opened 2024-07-24
[Samples] Update IIS sample Dockerfile

lucaspimentel:
## Summary of changes

Improve the `Dockerfile` used to install `dd-trace-dotnet` in the containerized IIS app sample in `tracer/samples/IISInDocker`.

## Reason for change

A user reported having issues with the Dockerfile, and I was able to reproduce.

## Implementation details

- move install script to a separate `ps1` file
- use absolute paths for msi download and log file (this is the main reason for this PR, the relative paths fail in common scenarios)
- use PowerShell's built-in `Invoke-WebRequest` instead of .NET's `System.Net.WebClient.DownloadFile()`
(which is [obsolete](https://learn.microsoft.com/en-us/dotnet/api/system.net.webclient#remarks))
- get the github release flagged as `latest` instead of the most recent (not the same thing! this would've downloaded preview versions of v3.x, for example)
- add more verbosity in console messages (show url, download path, log path, etc) to ease troubleshooting
- add basic `.dockerignore` file

## Test coverage

Tested manually.

### DIFF
--- a/tracer/samples/IISInDocker/Dockerfile
+++ b/tracer/samples/IISInDocker/Dockerfile
@@ -22,9 +22,7 @@ COPY install-dd-trace-dotnet.ps1 startup.ps1 ./
 # Install the Datadog SDK for .NET using the msi installer
 RUN .\install-dd-trace-dotnet.ps1
 
-# Remove the default ENTRYPOINT "C:\ServiceMonitor.exe w3svc"
-ENTRYPOINT []
-
 # Note: Normally an IIS reset is also required but 'C:\ServiceMonitor.exe w3svc'
 # will restart IIS and update the environment variables received by the child w3wp processes
-CMD .\startup.ps1
+# Replace the default ENTRYPOINT "C:\ServiceMonitor.exe w3svc"
+ENTRYPOINT .\startup.ps1
--- a/tracer/samples/IISInDocker/Dockerfile
+++ b/tracer/samples/IISInDocker/Dockerfile
@@ -3,7 +3,7 @@ FROM mcr.microsoft.com/dotnet/framework/sdk:4.8 AS build
 WORKDIR /app
 
 # copy csproj and restore as distinct layers
-COPY *.sln .
+COPY *.sln ./
 COPY src/*.csproj ./src/
 COPY src/*.config ./src/
 RUN nuget restore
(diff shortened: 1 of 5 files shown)


tracer/samples/IISInDocker/Dockerfile
```diff
@@ -15,24 +15,16 @@ RUN msbuild /p:Configuration=Release
...
+# Install the Datadog SDK for .NET using the msi installer
+RUN .\install-dd-trace-dotnet.ps1
+
+# Remove the default ENTRYPOINT "C:\ServiceMonitor.exe w3svc"
+ENTRYPOINT []
 
-# Install the .NET Tracer MSI
 # Note: Normally an IIS reset is also required but 'C:\ServiceMonitor.exe w3svc'
 # will restart IIS and update the environment variables received by the child w3wp processes
-RUN Write-Host 'Determining latest release' ;\
```
andrewlock, line 27: Is there a reason this can't be done here and instead has to be moved to a file? 🤔 I don't know of a good reason it needs to be?

tracer/samples/IISInDocker/Dockerfile
```diff
@@ -15,24 +15,16 @@ RUN msbuild /p:Configuration=Release
...
 WORKDIR /inetpub/wwwroot
 COPY --from=build /app/src/. ./
-COPY startup.ps1 .
+COPY install-dd-trace-dotnet.ps1 startup.ps1 ./
+
+# Install the Datadog SDK for .NET using the msi installer
+RUN .\install-dd-trace-dotnet.ps1
+
+# Remove the default ENTRYPOINT "C:\ServiceMonitor.exe w3svc"
+ENTRYPOINT []
```
andrewlock, line 26: Why remove the entrypoint and use cmd instead? 🤔 We would basically never want to change the cmd for this dockerfile otherwise it won't be configured correctly, so why make that easier to do accidentally?

andrewlock (approved):
I take your word for it, but I don't really understand _why_ we need to make some of these changes 🤷‍♂️

tracer/samples/IISInDocker/Dockerfile
lucaspimentel: The `ENTRYPOINT` added by the base image is `C:\ServiceMonitor.exe w3svc`, but we need to call `startup.ps1` instead to add some last-minute env vars (which cannot be added earlier). That script then calls `C:\ServiceMonitor.exe w3svc` as the last step.

The previous version of this was also overriding the `ENTRYPOINT` for the same reason, but I think it makes more sense to call `startup.ps1` from `CMD` than from `ENTRYPOINT`, so I had to clear out the `ENTRYPOINT`.

lucaspimentel: > I think it makes more sense to call `startup.ps1` from `CMD` than from `ENTRYPOINT`

The reasoning here is if you want to run a custom command, you override `CMD`. If you have an `ENTRYPOINT` defined, then what you pass to `CMD` just becomes _arguments_ to `ENTRYPOINT`, which would not work here.

I'll test this again to be sure, though.

lucaspimentel: Although, now that I think about it, the base image was using `ENTRYPOINT`, so why mess with that? I'll change this, thanks.

lucaspimentel: Done!

lucaspimentel: This is so confusing: https://docs.docker.com/reference/dockerfile/#understand-how-cmd-and-entrypoint-interact

PR #5851, opened 2024-08-06
[Profiler] Fix null named thread

chrisnas:
## Summary of changes
Check against unnamed thread case

## Reason for change
Avoid null reference crash when listing threads CPU consumption

## Test coverage
The corresponding test has been updated


profiler/src/ProfilerEngine/Datadog.Profiler.Native.Windows/WindowsThreadsCpuManager.cpp
```diff
@@ -94,8 +94,16 @@ void ThreadsCpuManager::LogCpuTimes()
                 builder << std::setw(6) << threadEntry.th32ThreadID << " | ";
                 builder << std::setw(8) << ThreadMs << " ms [";
                 builder << std::setw(5) << std::setprecision(2) << percent << " %]  ";
-                auto name = *element->second->GetName();
-                builder << shared::ToString(name);
+                auto name = element->second->GetName();
```
gleocadie, line 97: nit: can you use `auto*` . Just to ease the reading.
Thx :)

gleocadie (approved):

profiler/src/ProfilerEngine/Datadog.Profiler.Native.Windows/WindowsThreadsCpuManager.cpp
gleocadie, line 97:
```suggestion
                auto* name = element->second->GetName();
```

chrisnas: ✅ done

PR #5857, opened 2024-08-07
Update Datadog.Trace README to reference v3 migration guide

andrewlock:
## Summary of changes

Updates the Datadog.Trace README to reference the v3 migration guide

## Reason for change

The v3 package should reference the v3 migration, not the v2 migration

## Implementation details

Copied the summary from the migration guide into the package README. Added warning that we don't support custom-only any more


daniel-romano-DD (approved):
👍

PR #5881, opened 2024-08-12
Fix `NullReferenceException` in `HttpClientResponse.GetCharsetEncoding`

andrewlock:
## Summary of changes

Fixes a `NullReferenceException` that occurred on some responses in .NET 6+

## Reason for change

We were getting null reference exceptions in some cases

## Implementation details

Fix the NRE, and add some more `#nullable enable`

## Test coverage

Should probably have some :awkward-monkey:

PR #5887, opened 2024-08-13
Fix `NullReferenceException` in `HttpClientResponse.GetCharsetEncoding` (#5881 -> v2)

andrewlock:
## Summary of changes

Fixes a `NullReferenceException` that occurred on some responses in .NET 6+

## Reason for change

We were getting null reference exceptions in some cases

## Implementation details

Fix the NRE, and add some more `#nullable enable`

## Test coverage

Should probably have some :awkward-monkey:


bouwkast (approved):
👍

PR #5899, opened 2024-08-14
Fix musl tags for dd-lib-dotnet-init image

randomanderson:
## Summary of changes
Musl tags generated from `generate-lib-init-pinned-tag-values-musl` have an `_musl` suffix. This PR ensures that the pinning job uses the correct variable.


andrewlock (approved):
Thanks!

PR #5906, opened 2024-08-15
Ignore `StyleCop.Analyzers` in dependabot

bouwkast:
## Summary of changes

Adds `StyleCope.Analyzers` to the ignored dependencies within Dependabot.

## Reason for change

I've been looking into why it seemed that Dependabot wasn't creating PRs and found several issues in the logs:

```
`.../NuGet.targets(169,5): error : '[1.2.0-beta.*]' is not a valid version string. [...]`
```

This was repeated throughout all of the Dependabot logs and doesn't _appear_ to cause the run to loudly fail, but appears to cause it to silently fail. This version comes from here: https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/Directory.Build.props#L20

I attempted to replicate this issue and have had no luck. But I think it is fine to just try to ignore this package as it is a floating version anyway and will automatically update (and only applies to the editor).

## Implementation details

Added the package to the ignore list.

## Test coverage

- Attempted to debug this locally with MSBuild tasks to parse versions (which seemed fine) and with dependabot

## Other details

I don't think there is a way to manually run this file until it is merged.

There is another issue that may or may not be related with `InlineIL.Fody` and the `SourceGenerators` project, but want to fix this one first as it is in all of them.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,6 +10,7 @@ updates:
       - "area:dependabot"
     ignore:
       - dependency-name: "MessagePack" # Locked at a version that supports our net452 build target
+      - dependency-name: "StyleCop.Analyzers" # floating version is being used and it seemed to break dependabot (for some reason)
       - dependency-name: "*" # Ignore patches for all integrations
         update-types: ["version-update:semver-patch"]
   # Src libraries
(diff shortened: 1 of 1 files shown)


andrewlock (approved):
Tbh, I thought StyleCop.Analyzers was deprecated in favour of other analyzers anyway 🤔 We should look into that at some point 😄

lucaspimentel (approved):

bouwkast:
> Tbh, I thought StyleCop.Analyzers was deprecated in favour of other analyzers anyway 🤔 We should look into that at some point 😄

I think it provides additional analyzers compared to the official ones, if it still poses an issue I can take a peek at what the current status is and whether it can be removed in place of the other ones.

PR #5922, opened 2024-08-21
Stop testing .NET Core 2.1 on PRs

andrewlock:
## Summary of changes

- .NET Core 2.1 is EOL so stop testing it on PRs
- Refactor the ARM64 tests to use the same strategy-generation as other integration tests

## Reason for change

.NET Core 2.1 is EOL, so stop testing it by default. We may also remove it from the full/master runs too, but we can't quite agree the best option for that. Everyone agrees on _this_ step though.

Also switch the arm64 integration tests to use the same generation strategy as all the other integration tests because it was inconsistent, and will make some other things I'm working on easier.

## Implementation details

- Include `IsArm64` as a decision point in the `TestFrameworks` property
- Tweak the existing arm64 integration tests yaml to be the in line with the x64 integration tests

## Test coverage

This is the test, but we have less coverage in PRs now

### DIFF
--- a/.azure-pipelines/ultimate-pipeline.yml
+++ b/.azure-pipelines/ultimate-pipeline.yml
@@ -2694,7 +2694,6 @@ stages:
     targetBranch: $[ stageDependencies.merge_commit_id.fetch.outputs['set_sha.branch']]
     TestAllPackageVersions: true
     IncludeMinorPackageVersions: $[eq(variables.perform_comprehensive_testing, 'true')]
-    baseImage: debian
 
   jobs:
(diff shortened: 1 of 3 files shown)


daniel-romano-DD (approved):
Thanks a lot

andrewlock:

- [This PR (5922)](https://github.com/DataDog/dd-trace-dotnet/tree/3b3fef70a76f11d12b2c6a640cc15e874e8093bb)
- [master](https://github.com/DataDog/dd-trace-dotnet/tree/9a9d06b0bac95b33b37a59694353d24e4ed47698)

PR #5942, opened 2024-08-23
Stop unloading the native loader on error

kevingosse:
## Summary of changes

When something goes wrong and we collect the telemetry, don't unload the native loader. Instead, just disable the profiler callbacks.

## Reason for change

We discovered a potential crash because `SingleStepGuardRails` spawns a thread to send telemetry. If in the meantime we return `E_FAIL` in `ICorProfilerInfo::Initialize`, .NET will unload the module while the background thread is still running, causing a segfault.

This is a stopgap measure until we figure out a safe way to unload the module. In any case, the native tracer and the native profiler were not unloaded even before this change (because we never realized that the CLR doesn't know about them, so it can't unload them), so keeping the native loader in memory doesn't make a big difference.

## Implementation details

Added a flag that is checked in the `RunInAllProfilers` macro, to stop forwarding the callbacks to the other profilers.

## Test coverage

`OnEolFrameworkInSsi_WhenForwarderPathExists_CallsForwarderWithExpectedTelemetry` fails on 3.0 without that change. Unfortunately it currently doesn't run in the CI.


andrewlock (approved):

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -245,14 +251,16 @@ namespace datadog::shared::nativeloader
             Log::Warn("CorProfiler::Initialize: Failed to attach profiler, interface ICorProfilerInfo4 not found.");
             // we're not recording the exact version here, we just know that at this point it's not enough
             single_step_guard_rails.RecordBootstrapError(SingleStepGuardRails::NetFrameworkRuntime, inferredVersion, "incompatible_runtime");
-            return E_FAIL;
+            m_disabled = true;
+            return S_OK;
```
andrewlock, line 255: We only spawn the thread _if_ we're in an SSI scenario. Do we want to return `E_FAIL` in the situation where we are _not_ in SSI? 

i.e. should `single_step_guard_rails.RecordBootstrapError()` return a boolean (indicating whether we should return `S_OK/E_FAIL`)?

andrewlock:
Closed in favor of https://github.com/DataDog/dd-trace-dotnet/pull/5944

PR #5956, opened 2024-08-28
Protect the connection string tags extractor from an invalid connection string

tonyredondo:
## Summary of changes

This PR fixes an scenario when the connection string could not be parsed by avoiding the exception.

## Reason for change

Seen in Error tracking:

image

## Implementation details

Wrap the method in a try catch block, and return default on that invalid connection string.
The caller method already returns `default` in other cases.

## Test coverage

This is an small fix, I didn't write a test case for this... yet.

### DIFF
--- a/tracer/src/Datadog.Trace/Util/DbCommandCache.cs
+++ b/tracer/src/Datadog.Trace/Util/DbCommandCache.cs
@@ -97,13 +97,23 @@ namespace Datadog.Trace.Util
 
         private static TagsCacheItem ExtractTagsFromConnectionString(string connectionString)
         {
-            // Parse the connection string
-            var builder = new DbConnectionStringBuilder { ConnectionString = connectionString };
+            try
+            {
+                // Parse the connection string
+                var builder = new DbConnectionStringBuilder { ConnectionString = connectionString };
 
-            return new TagsCacheItem(
-                dbName: GetConnectionStringValue(builder, "Database", "Initial Catalog", "InitialCatalog"),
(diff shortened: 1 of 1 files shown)


andrewlock:

- [This PR (5956)](https://github.com/DataDog/dd-trace-dotnet/tree/71b44207b43d921639990a25d6089e2f8c320217)
- [master](https://github.com/DataDog/dd-trace-dotnet/tree/009c689dff5bce6014c5044ec1cc045abb90435d)

tracer/src/Datadog.Trace/Util/DbCommandCache.cs
```diff
@@ -97,13 +97,23 @@ public static TagsCacheItem GetTagsFromDbCommand(IDbCommand command)
...
+
+                // Extract the tags
+                return new TagsCacheItem(
+                    dbName: GetConnectionStringValue(builder, "Database", "Initial Catalog", "InitialCatalog"),
+                    dbUser: GetConnectionStringValue(builder, "User ID", "UserID"),
+                    outHost: GetConnectionStringValue(builder, "Server", "Data Source", "DataSource", "Network Address", "NetworkAddress", "Address", "Addr", "Host"));
+            }
+            catch (Exception)
+            {
+                // DbConnectionStringBuilder can throw exceptions if the connection string is invalid (as seen in Error Tracking)
```
andrewlock, line 113: nit: I would just say this
```suggestion
                // DbConnectionStringBuilder can throw exceptions if the connection string is invalid
```

andrewlock (approved):
Awesome, thanks!

PR #5979, opened 2024-09-03
Publish musl tags for images as well

randomanderson:
## Summary of changes
Publish `-musl` pinned tags for releases

## Reason for change

Pinned tags like `latest-musl` we not being applied. 3.2.0 was the first non-prerelease release of the 3.x branch


randomanderson:
decided not to add -musl tags anymore

PR #6027, opened 2024-09-12
Fix gitlab build (#6025 -> v2)

andrewlock:
## Summary of changes

Enables longpath support on the gitlab Windows jobs

## Reason for change

We moved the location of the repo in GitLab, which increased the path length, and we hit long path issues

## Implementation details

Make the runner set `git config --system core.longpaths true` - it's [a crazy olde time restriction](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=powershell), so realistically everyone should have this anyway

## Test coverage

The build works, so it's fixed!

## Other details

We could shorten the paths to work around it, but it's just another accident waiting to happen. This is not the first time 🙈 

Backport of #6025

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -25,6 +25,9 @@ build:
       - $CI_COMMIT_TAG # We don't need to build/publish when building a release tag
   stage: build
   tags: ["runner:windows-docker", "windowsversion:1809"]
+  hooks:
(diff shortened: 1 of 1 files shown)


daniel-romano-DD (approved):
TY

robertpi (approved):

PR #6048, opened 2024-09-18
Fix dlsym issue

gleocadie:
## Summary of changes

This PR addresses the issue https://github.com/DataDog/dd-trace-dotnet/issues/6045

## Reason for change

When using the `dlsym` function, the compiler adds in the import symbols table that we need the `dlsym` symbol.
Before being a universal binary (same binary used for glibc-based linux and musl-libc-based linux) and the compiler added in a `DT_NEEDED` section the library `libdl.so` (the library containing `dlsym`). When the wrapper is loaded, it will look through all the `DT_NEEDED` sections to find a library that contains the `dlsym` symbol.
Since being a universal binary, the `DT_NEEDED` sections are removed (part of being universal) and we have to resolve by hand needed symbols (`dlsym`, `pthread_once` ..).
If we use `dlsym` (or other symbol), we will hit this issue.

## Implementation details

- use `__dd_dlsym` instead

## Test coverage

Added a snapshot test using `nm` that verifies that the undefined symbols in the universal binary haven't changed. It's equivalent to running

```bash
nm -D Datadog.Linux.ApiWrapper.x64.so | grep ' U ' | awk '{print $2}' | sed 's/@.*//' | sort
```

but done using Nuke instead. It would probably make sense for this to be a "normal" test in the native tests, but given it has a dependency on `nm`, which is _definitely_ available in the universal build dockerfile it was quicker and easier to get this up and running directly. When it fails, it prints the diff and throws an exception, e.g.

```bash
System.Exception: Found differences in undefined symbols (dlsym) in the Native Wrapper library. Verify that these changes are expected, and will not cause problems. Removing symbols is generally a safe operation, but adding them could cause crashes. If the new symbols are safe to add, update the snapshot file at C:\repos\dd-trace-dotnet\tracer\test\snapshots\native-wrapper-symbols-x64.verified.txt with the new values
```

## Other details

This will be hotfixed onto 3.3.1 and 2.59.1

Issue #6045, opened 2024-09-17
The latest Tracer agent causes errors in our docker container


ryanrivest:
**Describe the bug**
We ran into an issue with some deployments on 9/16 after we automatically installed the latest tracer agent that was released  (3.3.0).

When this version is installed, it causes some of our shell commands to return an error. For example, when we run `cat`, it fails with the following error:

>cat: symbol lookup error: /opt/datadog/continuousprofiler/Datadog.Linux.ApiWrapper.x64.so: undefined symbol: dlsym

When we switch back to 3.2.0, the error goes away. It also goes away if the dotnet runtime is updated to 8.0.

We're not quite ready to change to .NET 8, so it would be nice if this issue could be resolved with the .NET 6 container.

I've created a reprod for you here:  https://github.com/ryanrivest/dd-tracer-reprod/tree/main

**To Reproduce**
1. Run `docker build --file .\Dockerfile --tag ddtracer:test .`
2. Run `docker run --rm -it --entrypoint /bin/bash ddtracer:test`
3. Try to run `cat` or `clear`, and it should fail with the error below.

Error:
>cat: symbol lookup error: /opt/datadog/continuousprofiler/Datadog.Linux.ApiWrapper.x64.so: undefined symbol: dlsym

**Expected behavior**
No symbol lookup error when running commands.

**Runtime environment (please complete the following information):**
- Tracer version: 3.3.0
- Env: Docker
- CLR: .NET 6

alvarovera-epam:
This also affecting our ECS Fargate containers, the health check fails due to this error. This is critical

bouwkast:
Hi thanks a bunch for the issue and the repository to reproduce this @ryanrivest much appreciated.

For the time being would it be acceptable to revert to `3.2.0` of the .NET Tracer _or_ to disable the Continuous Profiler by setting `DD_PROFILING_ENABLED` set to `false` while we investigate this issue?

> It also goes away if the dotnet runtime is updated to 8.0.

Thanks for this bit of info as well as it'll likely help speed up the investigation.

Also, if necessary please don't hesitate to open a support ticket via Datadog Support here: https://help.datadoghq.com/hc/requests/new?tf_1260824651490=pt_product_type:apm&tf_1900004146284=pt_apm_language:.net

However, we are looking into this and I've looped in the Continuous Profilers folks to help.

ryanrivest:
[...]

PR #6051, opened 2024-09-18
Fix dlsym issue (#6048 -> v2)

andrewlock:
## Summary of changes

This PR addresses the issue https://github.com/DataDog/dd-trace-dotnet/issues/6045

## Reason for change

When using the `dlsym` function, the compiler adds in the import symbols table that we need the `dlsym` symbol.
Before being a universal binary (same binary used for glibc-based linux and musl-libc-based linux) and the compiler added in a `DT_NEEDED` section the library `libdl.so` (the library containing `dlsym`). When the wrapper is loaded, it will look through all the `DT_NEEDED` sections to find a library that contains the `dlsym` symbol.
Since being a universal binary, the `DT_NEEDED` sections are removed (part of being universal) and we have to resolve by hand needed symbols (`dlsym`, `pthread_once` ..).
If we use `dlsym` (or other symbol), we will hit this issue.

## Implementation details

- use `__dd_dlsym` instead

## Test coverage

Added a snapshot test using `nm` that verifies that the undefined symbols in the universal binary haven't changed. It's equivalent to running

```bash
nm -D Datadog.Linux.ApiWrapper.x64.so | grep ' U ' | awk '{print $2}' | sed 's/@.*//' | sort
```

but done using Nuke instead. It would probably make sense for this to be a "normal" test in the native tests, but given it has a dependency on `nm`, which is _definitely_ available in the universal build dockerfile it was quicker and easier to get this up and running directly. When it fails, it prints the diff and throws an exception, e.g.

```bash
System.Exception: Found differences in undefined symbols (dlsym) in the Native Wrapper library. Verify that these changes are expected, and will not cause problems. Removing symbols is generally a safe operation, but adding them could cause crashes. If the new symbols are safe to add, update the snapshot file at C:\repos\dd-trace-dotnet\tracer\test\snapshots\native-wrapper-symbols-x64.verified.txt with the new values
```

## Other details

This is a hotfix of
- #6048


bouwkast (approved):
Awesome thanks!

PR #6056, opened 2024-09-19
Test enabling lib injection kubernetes tests

randomanderson:

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -7,7 +7,7 @@ stages:
   - benchmarks
 
 include:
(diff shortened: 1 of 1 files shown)


andrewlock (commented):
FYI, I manually triggered an Azure Devops build [here](https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=164648&view=results) so that the SSI artifacts can be created. 

Not building for edits to this made more sense once upon a time, but now it's a bit of a pain when you're doing work like this... Hopefully unblocks your test anyway

randomanderson:
@andrewlock it built fine before. This PR was unnecessarily left open

PR #6074, opened 2024-09-24
Switch .NET tracer to injecting both base64 & binary headers

veerbia:
## Summary of changes
- Switch .NET tracer to injecting both base64 & binary headers
- Binary headers can be disabled via a config variable: DD_DATA_STREAMS_LEGACY_HEADERS=false (defaults to true)
- .NET tracer, when extracting the context from headers looks at both base64 & binary headers

## Reason for change
Data Streams previously used binary encoding in Kafka headers. This was causing issues in cross language communication because of the difference in negative byte handling. That’s why we switched to base64 encoding.
Today, .NET is the only remaining tracer using binary encoding for Kafka, SQS & RabbitMQ.

This is causing 3 issues:
- Communication between .NET & Java breaks, since Java doesn’t support negative bytes in Kafka headers
- Manual instrumentation breaks between .NET and any other language, since manual instrumentation uses base64 encoding
- .NET can’t consume payloads from any other tracers, since the other tracers encode in base64

Also, byte headers are causing a crash of the [.NET application](https://datadog.zendesk.com/agent/tickets/1824351) (not reproduced yet).

## Implementation details
- Injects base64-encoded headers by default to ensure cross-language compatibility.
- Binary headers are injected only if `DD_DATA_STREAMS_LEGACY_HEADERS` (default: true) is enabled for backward compatibility.
- Prefers base64 headers for extraction. If base64 is unavailable or malformed, it falls back to binary headers when legacy headers are enabled.
- A new configuration setting DD_DATA_STREAMS_LEGACY_HEADERS controls whether binary headers are included.

## Test coverage
I've added tests for the following cases:
- Inject_WhenLegacyHeadersDisabled_DoesNotIncludeBinaryHeader 
    - Ensures that when legacy headers are disabled, the binary header is not included in the injected headers.

- Extract_WhenBothHeadersPresent_PrefersBase64Header
    - Confirms that when both Base64 and binary headers are present, the Base64 header is preferred for extraction.

- InjectedHeaders_HaveCorrectFormat
    - Validates that injected headers are in the correct format, especially checking if the Base64-encoded header is valid and correctly decoded.

- InjectHeaders_WhenLegacyHeadersDisabled_DoesNotIncludeLegacyHeader 
    - Ensures that legacy headers are excluded when legacy header support is disabled.

- Inject_WhenLegacyHeadersEnabled_IncludesBothHeaders 
    - Confirms that both Base64 and binary headers are injected when legacy header support is enabled.

- Extract_WhenBase64HeaderIsMalformed_ReturnsFallbackToBinary 
    - Verifies that if the Base64 header is malformed, the system falls back to using the binary header for extraction.


lucaspimentel:
Does this PR replace #6012?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/SQS/ContextPropagation.cs
```diff
@@ -26,8 +26,21 @@ private static void Inject<TMessageRequest>(SpanContext context, IDictionary mes
...
-            dataStreamsManager?.InjectPathwayContext(context.PathwayContext, AwsSqsHeadersAdapters.GetInjectionAdapter(sb));
+
+            if (dataStreamsManager != null && dataStreamsManager.IsEnabled && context.PathwayContext != null)
+            {
+                var base64EncodedContext = Convert.ToBase64String(PathwayContextEncoder.Encode(context.PathwayContext.Value));
+                sb.AppendFormat("\"{0}\":\"{1}\",", DataStreamsPropagationHeaders.PropagationKeyBase64, base64EncodedContext);
+
+                if (Tracer.Instance.Settings.IsDataStreamsLegacyHeadersEnabled)
+                {
+                    var binaryEncodedContext = Convert.ToBase64String(PathwayContextEncoder.Encode(context.PathwayContext.Value));
```
lucaspimentel, line 39: Isn't this injecting the same as above by using `Convert.ToBase64String`?

tracer/src/Datadog.Trace/Configuration/ImmutableDynamicSettings.cs
```diff
@@ -20,6 +20,8 @@ internal class ImmutableDynamicSettings : IEquatable<ImmutableDynamicSettings>
 
         public bool? DataStreamsMonitoringEnabled { get; init; }
 
+        public bool? DataStreamsLegacyHeadersEnabled { get; init; }
```
lucaspimentel, line 23: "Dynamic" settings are used for that is now called Remote Configuring. We don't need to add this here unless you're going to support changing this setting via remote config.

tracer/src/Datadog.Trace/Configuration/ImmutableTracerSettings.cs
```diff
@@ -528,6 +530,11 @@ internal ImmutableTracerSettings(TracerSettings settings, bool unusedParamNotToU
         /// </summary>
         internal bool IsDataStreamsMonitoringEnabled => DynamicSettings.DataStreamsMonitoringEnabled ?? _isDataStreamsMonitoringEnabled;
 
+        /// <summary>
+        /// Gets a value indicating whether data streams monitoring legacy headers are enabled or not.
+        /// </summary>
+        internal bool IsDataStreamsLegacyHeadersEnabled => DynamicSettings.DataStreamsLegacyHeadersEnabled ?? _isDataStreamsLegacyHeadersEnabled;
```
lucaspimentel, line 536: If you remove this from `DynamicSettings` (see my other comment), then you can remove the `_isDataStreamsLegacyHeadersEnabled` and convert `IsDataStreamsLegacyHeadersEnabled` into a get-only property (like the ones with `{ get; }` below).

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsContextPropagator.cs
```diff
@@ -31,7 +34,13 @@ public void Inject<TCarrier>(PathwayContext context, TCarrier headers)
     {
         if (headers is null) { ThrowHelper.ThrowArgumentNullException(nameof(headers)); }
 
-        headers.Add(DataStreamsPropagationHeaders.PropagationKey, PathwayContextEncoder.Encode(context));
+        var base64EncodedContext = Convert.ToBase64String(PathwayContextEncoder.Encode(context));
```
lucaspimentel, line 37: You could call `PathwayContextEncoder.Encode(context)` only once and re-use that value for both headers instead of doing the work twice and allocating an additional `byte[]`.

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
```diff
@@ -73,7 +73,9 @@ public async Task DisposeAsync()
     /// </summary>
     public PathwayContext? ExtractPathwayContext<TCarrier>(TCarrier headers)
         where TCarrier : IBinaryHeadersCollection
-        => IsEnabled ? DataStreamsContextPropagator.Instance.Extract(headers) : null;
+    {
+        return IsEnabled ? DataStreamsContextPropagator.Instance.Extract(headers) : null;
+    }
```
lucaspimentel, line 78: Any reason to revert back to the older syntax? If not, do you mind reverting this change?

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
```diff
@@ -83,16 +85,12 @@ public async Task DisposeAsync()
     public void InjectPathwayContext<TCarrier>(PathwayContext? context, TCarrier headers)
         where TCarrier : IBinaryHeadersCollection
     {
-        if (!IsEnabled)
+        if (!IsEnabled || context is null)
```
lucaspimentel, line 88: 👍

tracer/src/Datadog.Trace/TracerManager.cs
```diff
@@ -29,6 +29,8 @@
 using Datadog.Trace.Sampling;
 using Datadog.Trace.Telemetry;
 using Datadog.Trace.Util.Http;
+using Datadog.Trace.VendoredMicrosoftCode.System.Reflection.Metadata;
+using Datadog.Trace.Vendors.dnlib.DotNet.Writer;
```
lucaspimentel, line 33: Looks like these two lines were added by mistake. Mind reverting them?

veerbia: Thank you, removed them.

tracer/test/Datadog.Trace.Tests/DataStreamsMonitoring/DataStreamsContextPropagatorTests.cs
```diff
@@ -38,6 +38,43 @@ public void CanRoundTripPathwayContext()
...
+            DateTimeOffset.UtcNow.AddSeconds(-5).ToUnixTimeNanoseconds(),
+            DateTimeOffset.UtcNow.ToUnixTimeNanoseconds());
+
+        DataStreamsContextPropagator.Instance.Inject(context, headers);
+
+        headers.Values.Should().ContainKey(DataStreamsPropagationHeaders.PropagationKeyBase64);
+        headers.Values[DataStreamsPropagationHeaders.PropagationKeyBase64].Should().NotBeNullOrEmpty();
+
+        headers.Values.Should().ContainKey(DataStreamsPropagationHeaders.PropagationKey);
+        headers.Values[DataStreamsPropagationHeaders.PropagationKey].Should().NotBeNullOrEmpty();
```
lucaspimentel, line 56: You should also add a test that the legacy header is _not_ injected when disabled.

And you'll probably also want to add tests that assert the headers (both old and new) contain the data we expect and in the correct format. I don't see any tests for that here (the existing round-tripping is not enough to ensure the format is compatible across tracers).

tracer/src/Datadog.Trace/Configuration/ImmutableDynamicSettings.cs
piochelepiotr: 👍 I think we don't need to add it to remote configuration.

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
veerbia: Agreed, moved to newer syntax

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/SQS/ContextPropagation.cs
lucaspimentel: If that's intentional, then you can save the result of `Convert.ToBase64String(PathwayContextEncoder.Encode(context.PathwayContext.Value))` and re-use it to avoid the extra work and heap allocations.

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -760,6 +760,13 @@ internal static class DataStreamsMonitoring
             /// </summary>
             /// <see cref="TracerSettings.IsDataStreamsMonitoringEnabled"/>
             public const string Enabled = "DD_DATA_STREAMS_ENABLED";
+
+            /// <summary>
+            /// Configuration key for enabling legacy binary headers in Data Streams Monitoring.
+            /// Default is true.
+            /// </summary>
+             /// <see cref="TracerSettings.IsDataStreamsLegacyHeadersEnabled"/>
+            public const string LegacyHeadersEnabled = "DD_DATA_STREAMS_LEGACY_HEADERS";
```
bouwkast, line 769: Has this been added to https://github.com/DataDog/dd-go/blob/prod/trace/apps/tracer-telemetry-intake/telemetry-payload/static/config_norm_rules.json?

bouwkast (commented):
I think Lucas' comments still need to be resolved.

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -961,6 +965,11 @@ public bool DiagnosticSourceEnabled
         /// </summary>
         internal bool IsDataStreamsMonitoringEnabled { get; }
 
+        /// <summary>
+        /// Gets a value indicating whether to inject legacy binary headers for Data Streams.
+        /// </summary>
+        public bool IsDataStreamsLegacyHeadersEnabled { get; }
```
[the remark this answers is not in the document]
veerbia: Thanks, switched to internal

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/SQS/ContextPropagation.cs
veerbia: Fixed to inject different values (binary and base64) based on header

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsContextPropagator.cs
```diff
@@ -31,7 +34,14 @@ public void Inject<TCarrier>(PathwayContext context, TCarrier headers)
     {
         if (headers is null) { ThrowHelper.ThrowArgumentNullException(nameof(headers)); }
 
-        headers.Add(DataStreamsPropagationHeaders.PropagationKey, PathwayContextEncoder.Encode(context));
+        var encodedBytes = PathwayContextEncoder.Encode(context);
+        var base64EncodedContext = Convert.ToBase64String(encodedBytes);
+        headers.Add(DataStreamsPropagationHeaders.PropagationKeyBase64, Encoding.UTF8.GetBytes(base64EncodedContext));
```
lucaspimentel, line 39: In newer versions of .NET there's a method that converts from bytes into UTF-8 encoded base-64 bytes while skipping the intermediate string: `Base64.EncodeToUtf8` ([docs](https://learn.microsoft.com/en-us/dotnet/api/system.buffers.text.base64.encodetoutf8)). We have the code [in our repo](https://github.com/DataDog/dd-trace-dotnet/blob/bd582eb71638575c00fe904ee1b3a0a563de77d0/tracer/src/Datadog.Trace/Vendors/System.Memory/Buffers/Text/Base64.cs#L578) so we can use it from any version of .NET.

It will look something like:
```csharp
var encodedBytes = PathwayContextEncoder.Encode(context);

int base64Length = ((encodedBytes.Length + 2) / 3) * 4;
byte[] base64EncodedContextBytes = new byte[base64Length];
var status = Base64.EncodeToUtf8(encodedBytes , base64EncodedContextBytes, out _, out _);
// TODO: check that status == OperationStatus.Done

headers.Add(DataStreamsPropagationHeaders.PropagationKeyBase64, base64EncodedContextBytes);
```

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsContextPropagator.cs
```diff
@@ -45,9 +55,41 @@ public void Inject<TCarrier>(PathwayContext context, TCarrier headers)
...
 
-        var bytes = headers.TryGetLastBytes(DataStreamsPropagationHeaders.PropagationKey);
+        // Try to extract from the base64 header first
+        var base64Bytes = headers.TryGetLastBytes(DataStreamsPropagationHeaders.PropagationKeyBase64);
+        if (base64Bytes is { Length: > 0 })
+        {
+            try
+            {
+                var base64String = Encoding.UTF8.GetString(base64Bytes);
+                var decodedBytes = Convert.FromBase64String(base64String);
```
lucaspimentel, line 65: Similar to my comment above, but with `Base64.DecodeFromUtf8` ([docs](https://learn.microsoft.com/en-us/dotnet/api/system.buffers.text.base64.decodefromutf8)), you can skip the intermedia string.

veerbia: Thanks, moved this one over to the new direct byte conversion method as well to avoid intermediate string allocation.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/SQS/ContextPropagation.cs
```diff
@@ -26,8 +26,22 @@ private static void Inject<TMessageRequest>(SpanContext context, IDictionary mes
...
+
+            if (dataStreamsManager != null && dataStreamsManager.IsEnabled && context.PathwayContext != null)
+            {
+                var encodedBytes = PathwayContextEncoder.Encode(context.PathwayContext.Value);
+                var base64EncodedContext = Convert.ToBase64String(encodedBytes);
+                sb.AppendFormat("\"{0}\":\"{1}\",", DataStreamsPropagationHeaders.PropagationKeyBase64, base64EncodedContext);
+
+                if (Tracer.Instance.Settings.IsDataStreamsLegacyHeadersEnabled)
+                {
+                    var binaryContext = Encoding.UTF8.GetString(encodedBytes);
```
lucaspimentel, line 40: Are the bytes in `encodedBytes` guaranteed to be valid UTF-8 bytes for you to convert them directly into a string like this?

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsContextPropagator.cs
```diff
@@ -31,7 +36,24 @@ public void Inject<TCarrier>(PathwayContext context, TCarrier headers)
...
+        var encodedBytes = PathwayContextEncoder.Encode(context);
+
+        int base64Length = ((encodedBytes.Length + 2) / 3) * 4;
+        byte[] base64EncodedContextBytes = new byte[base64Length];
+        var status = Base64.EncodeToUtf8(encodedBytes, base64EncodedContextBytes, out _, out int bytesWritten);
+
+        if (status != OperationStatus.Done)
+        {
+            Log.Error("Failed to encode Data Streams context to Base64. OperationStatus: {Status}", status);
+            throw new InvalidOperationException($"Base64 encoding failed with status: {status}");
```
lucaspimentel, line 48: Are we handling this exception somewhere? If not, consider _not_ throwing and instead simply logging the error and not injecting the headers, otherwise we may break user's applications (big no-no).

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsContextPropagator.cs
```diff
@@ -31,7 +36,24 @@ public void Inject<TCarrier>(PathwayContext context, TCarrier headers)
...
+        }
+
+        headers.Add(DataStreamsPropagationHeaders.PropagationKeyBase64, base64EncodedContextBytes.AsSpan(0, bytesWritten).ToArray());
```
lucaspimentel, line 51: Oops. Now this `ToArray` will allocate a new byte array. I think in most cases `bytesWritten` will be equal to `base64EncodedContextBytes.Length`, so you can use `base64EncodedContextBytes` directly. Maybe add a check in case they are not equal.

Roughly:
```csharp
if (bytesWritten == base64EncodedContextBytes.Length)
{
    headers.Add(DataStreamsPropagationHeaders.PropagationKeyBase64, base64EncodedContextBytes);
}
else
{
    // allocates a new byte array
    headers.Add(DataStreamsPropagationHeaders.PropagationKeyBase64, base64EncodedContextBytes.AsSpan(0, bytesWritten).ToArray());
}
```

Sorry for being so fussy, but we take heap allocations (and their effect on user's apps) very seriously in the tracer 😅

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsContextPropagator.cs
```diff
@@ -45,9 +67,49 @@ public void Inject<TCarrier>(PathwayContext context, TCarrier headers)
...
+                var status = Base64.DecodeFromUtf8(base64Bytes, decodedBytes, out _, out int bytesWritten);
 
-        return bytes is { } ? PathwayContextEncoder.Decode(bytes) : null;
+                if (status != OperationStatus.Done)
+                {
+                    Log.Error("Failed to decode Base64 data streams context. OperationStatus: {Status}", status);
+                }
+                else
+                {
+                    return PathwayContextEncoder.Decode(decodedBytes.AsSpan(0, bytesWritten).ToArray());
```
lucaspimentel, line 87: Similar comment as above about using `Span<byte>.ToArray()` which always allocates a new byte array.

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsContextPropagator.cs
veerbia: Thanks, switched to logging error instead.

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsContextPropagator.cs
```diff
@@ -31,7 +36,23 @@ public void Inject<TCarrier>(PathwayContext context, TCarrier headers)
...
-        headers.Add(DataStreamsPropagationHeaders.PropagationKey, PathwayContextEncoder.Encode(context));
+        var encodedBytes = PathwayContextEncoder.Encode(context);
+
+        int base64Length = ((encodedBytes.Length + 2) / 3) * 4;
+        byte[] base64EncodedContextBytes = new byte[base64Length];
+        var status = Base64.EncodeToUtf8(encodedBytes, base64EncodedContextBytes, out _, out int bytesWritten);
+
+        if (status != OperationStatus.Done)
+        {
+            Log.Error("Failed to encode Data Streams context to Base64. OperationStatus: {Status}", status);
```
lucaspimentel, line 47: I think you want to bailed out here after logging the error. You were throwing an exception before, which I suggested you remove, but should you add a `return` now so the code below doesn't try to add the headers?

veerbia: Ah, good catch, thank you. Just added the return.

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsContextPropagator.cs
veerbia: Agreed, migrated to your suggestion.

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsContextPropagator.cs
veerbia: Agreed, migrated here as well.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/SQS/ContextPropagation.cs
```diff
@@ -26,8 +26,21 @@ private static void Inject<TMessageRequest>(SpanContext context, IDictionary mes
...
+
+            if (dataStreamsManager != null && dataStreamsManager.IsEnabled && context.PathwayContext != null)
+            {
+                var encodedBytes = PathwayContextEncoder.Encode(context.PathwayContext.Value);
+                var base64EncodedContext = Convert.ToBase64String(encodedBytes);
+                sb.AppendFormat("\"{0}\":\"{1}\",", DataStreamsPropagationHeaders.PropagationKeyBase64, base64EncodedContext);
+
+                if (Tracer.Instance.Settings.IsDataStreamsLegacyHeadersEnabled)
+                {
+                    sb.AppendFormat("\"{0}\":\"{1}\",", DataStreamsPropagationHeaders.PropagationKey, base64EncodedContext);
```
bouwkast, line 40: Is this supposed to be `base64EncodedContext` or `encodedBytes`?  Right now there isn't anything different except for the key name
ex)

```suggestion
                    sb.AppendFormat("\"{0}\":\"{1}\",", DataStreamsPropagationHeaders.PropagationKey, encodedBytes);
```

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsContextPropagator.cs
```diff
@@ -31,7 +36,30 @@ public void Inject<TCarrier>(PathwayContext context, TCarrier headers)
     {
         if (headers is null) { ThrowHelper.ThrowArgumentNullException(nameof(headers)); }
 
-        headers.Add(DataStreamsPropagationHeaders.PropagationKey, PathwayContextEncoder.Encode(context));
+        var encodedBytes = PathwayContextEncoder.Encode(context);
+
+        int base64Length = ((encodedBytes.Length + 2) / 3) * 4;
```
bouwkast, line 41: Could you add a comment here for why we are doing `+2 / 3 * 4`? As a reader I can't easily tell

 `((encodedBytes.Length + 2) / 3) * 4;`

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsContextPropagator.cs
bouwkast: Forget to push the `return` maybe as I'm not seeing it?

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsContextPropagator.cs
```diff
@@ -45,9 +73,57 @@ public void Inject<TCarrier>(PathwayContext context, TCarrier headers)
...
         if (headers is null) { ThrowHelper.ThrowArgumentNullException(nameof(headers)); }
 
-        var bytes = headers.TryGetLastBytes(DataStreamsPropagationHeaders.PropagationKey);
+        // Try to extract from the base64 header first
+        var base64Bytes = headers.TryGetLastBytes(DataStreamsPropagationHeaders.PropagationKeyBase64);
+        if (base64Bytes is { Length: > 0 })
+        {
+            try
+            {
+                int decodedLength = (base64Bytes.Length * 3) / 4;
```
bouwkast, line 82: Same thing as earlier if you could just document what the `* 3 / 4` is doing.

Also more curious as it doesn't have the `/2` as the encoding one did (not saying that is an issue, just hard to tell what the purpose is)

bouwkast (changes requested):
Couple of minor things, but looking good!

Feel free to ping on Slack for re-review :)

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/SQS/ContextPropagation.cs
veerbia: Great question! Yes, this is intended. I'd discussed this with Lucas briefly on slack but it is essentially because we want to support both the old and new key names for backwards compatibility.

bouwkast: Great!
Could we add a comment then here saying that :)

bouwkast (approved):
Thanks you!

piochelepiotr (approved):

lucaspimentel (approved):

veerbia:
There are currently some CI issues with forks for the dotnet tracer. I will close this PR and open an identical PR which uses a distinct branch as opposed to a fork. You can find the new PR [here](https://github.com/DataDog/dd-trace-dotnet/pull/6257)

PR #6128, opened 2024-10-08
[Profiler] Make sure Watcher thread is started before Sampler thread

gleocadie:
## Summary of changes

Ensure Watcher thread is started before Sampler thread.

## Reason for change

We got a customer report where the app was hanged. 
The Sampler thread interrupted the Main thread (because it might have executed managed code at some point or a bug in the CLR ¯\\\_(ツ)\_/¯) and that thread was holding at least 3 locks (the loadlibrary one, one shared between `RtlLookupFunctionEntry` and `RtlInvertedFunctionTable` and one between `LdrpXXX` functions).
The Sampler thread interrupted the Main thread while it was holding those locks. The Sampler thread called `RtlLookupFunctionEntry` function and was blocked on the lock (the one taken by `RtlInvertedFunctionTable`).
We noticed that the Watcher thread was blocked on `LdrpDrainWorkQueue` (lock taken by a call to one the `LdrpXX`  called by the Main thread)

## Implementation details

Start the sampler thread iff the Watcher thread has started.

PR #6142, opened 2024-10-10
[Profiler] Make sure Watcher thread is started before Sampler thread (#6128 => hotfix)

gleocadie:
## Summary of changes

Ensure Watcher thread is started before Sampler thread.

## Reason for change

We got a customer report where the app was hanged. 
The Sampler thread interrupted the Main thread (because it might have executed managed code at some point or a bug in the CLR ¯\\\_(ツ)\_/¯) and that thread was holding at least 3 locks (the loadlibrary one, one shared between `RtlLookupFunctionEntry` and `RtlInvertedFunctionTable` and one between `LdrpXXX` functions).
The Sampler thread interrupted the Main thread while it was holding those locks. The Sampler thread called `RtlLookupFunctionEntry` function and was blocked on the lock (the one taken by `RtlInvertedFunctionTable`).
We noticed that the Watcher thread was blocked on `LdrpDrainWorkQueue` (lock taken by a call to one the `LdrpXX`  called by the Main thread)

## Implementation details

Start the sampler thread iff the Watcher thread has started.


chrisnas (approved):
LGTM

gleocadie:
(╯°□°)╯︵ ┻━┻

PR #6180, opened 2024-10-22
Support Hot Chocolate v14

tobias-tengler:
Hot Chocolate v14 was released last week: [NuGet](https://www.nuget.org/packages/HotChocolate.AspNetCore/14.0.0)

We attempted to update today, but noticed that the DataDog integration no longer worked.

This PR updates the Tracer auto-instrumentation to support Hot Chocolate v14.


vandonr:
Hello, thank you for the contribution ! We're gonna have a look at the code, but just to answer your questions about MacOS development:

> the solution won't properly load in any of my IDEs

did you try to open the full `sln` ? There is an `slnf` that's dedicated to macs, it's `Datadog.Trace.OSX.slnf`, that's the one you should use.

> I can't run the integration tests on my ARM-based MacBook

ARM-based macbook shouldn't be an issue. We have a little bit of doc [here](https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/README.MD#macos) about that, you should use `.[/tracer/]build.sh BuildAndRunOsxIntegrationTests`. Click the link for an example with useful parameters.

tracer/missing-nullability-files.csv
```diff
@@ -655,12 +655,15 @@ src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Elasticsearch/V7/Elasticsearch
...
 src/Datadog.Trace/ClrProfiler/AutoInstrumentation/GraphQL/HotChocolate/ExecuteAsyncIntegrationExtraV13.cs
+src/Datadog.Trace/ClrProfiler/AutoInstrumentation/GraphQL/HotChocolate/ExecuteAsyncIntegrationExtraV14.cs
 src/Datadog.Trace/ClrProfiler/AutoInstrumentation/GraphQL/HotChocolate/ExecuteAsyncIntegrationV13.cs
+src/Datadog.Trace/ClrProfiler/AutoInstrumentation/GraphQL/HotChocolate/ExecuteAsyncIntegrationV14.cs
 src/Datadog.Trace/ClrProfiler/AutoInstrumentation/GraphQL/HotChocolate/HotChocolateCommon.cs
 src/Datadog.Trace/ClrProfiler/AutoInstrumentation/GraphQL/HotChocolate/IError.cs
 src/Datadog.Trace/ClrProfiler/AutoInstrumentation/GraphQL/HotChocolate/IExecutionResult.cs
 src/Datadog.Trace/ClrProfiler/AutoInstrumentation/GraphQL/HotChocolate/IOperationContext.cs
 src/Datadog.Trace/ClrProfiler/AutoInstrumentation/GraphQL/HotChocolate/IOperationContextV13.cs
+src/Datadog.Trace/ClrProfiler/AutoInstrumentation/GraphQL/HotChocolate/IOperationContextV14.cs
```
vandonr, line 666: I imagine you added those entries here because you got an error from our tools, but the better solution is to add `#nullable enable` at the top of those files, which will enable null-state analysis for safer code.

This csv files is supposed to only decrease in size as we add null-state analysis to more and more files.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/GraphQL/HotChocolate/ExecuteAsyncIntegrationExtraV14.cs
```diff
@@ -0,0 +1,47 @@
...
+    public class ExecuteAsyncIntegrationExtraV14
+    {
+        /// <summary>
+        /// OnMethodBegin callback
+        /// </summary>
+        /// <typeparam name="TTarget">Type of the target</typeparam>
+        /// <typeparam name="TOperationContext">Type of the first parameter</typeparam>
+        /// <param name="instance">Instance value, aka `this` of the instrumented method.</param>
+        /// <param name="operationContext">Operation context</param>
+        /// <returns>Calltarget state value</returns>
```
vandonr, line 35: those kind of "auto-generated" comments don't add value, and just make the codebase more verbose for no added benefit, it's better to add no comment in this case.
(same remark for other instances of this below)

tobias-tengler: Agreed. I just copied what was there in the old code, since I thought that was already according to your guidelines :) Removed it again.

tobias-tengler:
@vandonr Thanks for the feedback :) I was able to get the tests to run.
From my POV this PR should now be ready for proper review.

tobias-tengler:
@vandonr bump :)

bouwkast:
Hi @tobias-tengler 

Thanks for the contribution! Overall it looks really good
There are a couple of things to note:

- We can't merge PRs that are from forks due to some security/CI requirements that we have
- We also can't allow push access to `dd-trace-dotnet` outside of Datadog

So, what we can do though is cherry-pick your commits onto one of our own branches and make a PR on your behalf.
I've created a task for someone to pick this up to do that, so we'll follow up with that.

Sorry for these limitations, it is something that we are aware of and are attempting to improve.

tobias-tengler:
That's fine for me :) Can I assist in any way? For example by squashing my commits?

tobias-tengler:
@bouwkast Squashed everything into 26f35155e9ed567a7c060cdf49608be33b4b8950 :)

vandonr:
ok thanks, sorry to have kept you waiting like that, we don't get external contributions very often, we had to sync internally :)
I'll take care of cherry picking your commit on a branch of mine tomorrow !

PR #6194, opened 2024-10-24
Fix typo in create_draft_release.yml

andrewlock:
## Summary of changes

Fixes a bug in the release process

## Reason for change

The release process is broken due to a typo in the tar command

## Implementation details

Fix the typo

## Test coverage

Can't really test it without running the release again

## Other details

Will do a manual upload of them using the other action


gleocadie (approved):
🤞

PR #6262, opened 2024-11-08
Add first documentation on how to update the deny list

pierotibou:
## Summary of changes

Adds a quick documentation on how to add an item to the deny list. 
The document is located at the root of the `docs` folder as I'd like to use the same location across repo. But I can also maybe add it to a `prod` or `operations` folder)? 

## Reason for change
Not all tracers use the same method, so documenting per repo is important. 

## Implementation details
Just a md.

## Test coverage
N/A


pierotibou:
I'll have those centralized for now.

PR #6312, opened 2024-11-19
Fix the debug symbol publish step in release process

andrewlock:
## Summary of changes

Fixes the debug symbol publish step in release process.

## Reason for change

It's still broken.

## Implementation details

Use the correct path

## Test coverage

Nope

## Other details

If at first you don't succeed, fail, and fail again...


gleocadie (approved):
🤞

PR #6324, opened 2024-11-20
Use agent telemetry proxy instead of the standalone telemetry

andrewlock:
## Summary of changes

Use the agent proxy instead of the standalone telemetry in the HttpMessageHandlerTests

## Reason for change

Currently (after updating the VM images) the non-socket handler tests fail due to a lack of telemetry. I really don't understand how or why - the logs show that we are _successfully_ sending telemetry _somewhere_ 😬 Will investigate that in parallel.

Note also, this _only_ impacts alpine with .NET Core 2.1, 3.0, and 3.1. Debian or .NET 5+ are unaffected 🤷‍♂️ 

## Implementation details

Switched to using the agent proxy for telemetry (and added some missing extension helpers to the helper)

## Test coverage

The same really, there's no reason we need to use the external telemetry proxy for these tests.

## Other details

Will follow up once we have more details. Hopefully this will finally unblock the builds


daniel-romano-DD (approved):
Great :)

kevingosse (approved):

PR #6331, opened 2024-11-21
[CI Visibility] Catch `UploadRepositoryChanges` exceptions

tonyredondo:
## Summary of changes

This PR adds some missing try/catch for the UploadRepositoryChanges calls.

## Reason for change

This method is being used in the `dd-trace` runner, currently if the method fails (backend error) then `dd-trace` will throw an exception.

## Implementation details

We create a function method to wrap the call with a try/catch block so we don't throw and fail the customer CI.


andrewlock (approved):
LGTM

PR #6445, opened 2024-12-16
Fix codeql to work on newer version of Ubuntu

vandonr:
since Github [updated their runners](https://github.blog/changelog/2024-11-05-notice-of-breaking-changes-for-github-actions/#ubuntu-latest-upcoming-breaking-changes) to use a more recent version of ubuntu, it broke the codeQL job because it couldn't find LLVM using the new version name (`noble`).
The error we're seeing is `The repository 'http://apt.llvm.org/noble llvm-toolchain-noble-16 Release' does not have a Release file.`
It turns out Clang-16 is available by default in ubuntu Noble (24.04), so we don't need the script at all and we can just install from apt.

`DEBIAN_FRONTEND=noninteractive` is necessary to prevent a prompt from tzdata asking us to choose our timezone.


vandonr:
it says 
```clang-16 is already the newest version (1:16.0.6-23ubuntu4).
0 upgraded, 0 newly installed, 0 to remove and 112 not upgraded.```
so idk maybe we can remove this step entirely ? (until 2026 and the next LTS release :trollface:)

kevingosse (approved):

bouwkast (approved):
Thanks!

PR #6451, opened 2024-12-17
K8s new tests scenarios

robertomonteromiguel:
## Summary of changes

New K8s scenarios have been added to system-tests. These scenarios are managed by the one pipeline. 
Change the k8s lib injection tests jobs on gitlab

## Reason for change
New K8s scenarios have been added to system-tests.  We need to adapt the tracer pipeline
## Implementation details
Only added a test scenarios matrix for k8s lib injection tests. gitlab.


andrewlock (approved):
LGTM if it all passes 🙂

PR #6626, opened 2025-02-04
ci: pin github actions by hash and update via dependabot

xopham:
## Summary of changes

- **Add dependabot for github actions**
- **Pin actions by hash**

## Reason for change

Pinning 3rd-party GitHub Actions by commit SHA makes them less vulnerable to compromise of the 3rd party. To avoid outdating and non-verbosity, versions are commented after the SHA and updating via dependabot is introduced that will automatically update the commented version tag as well.

In case of a false commit SHA, this change could break the corresponding workflow. Typically, this does not cause major interruptions, but it can for example affect a release pipeline and require restart causing delays.


andrewlock (approved):
Thanks!

PR #6633, opened 2025-02-05
Bump Sep to v0.7.0

nietras:
@andrewlock time for another update :)

### DIFF
--- a/tracer/build/_build/_build.csproj
+++ b/tracer/build/_build/_build.csproj
@@ -42,7 +42,7 @@
     <!-- Explicitly override the version of octokit used by Nuke -->
     <PackageReference Include="Octokit" Version="10.0.0" />
     <PackageReference Include="Perfolizer" Version="0.2.1" />
-    <PackageReference Include="Sep" Version="0.5.2" />
+    <PackageReference Include="Sep" Version="0.9.0" />
     <PackageReference Include="ByteSize" Version="2.1.0" />
     <!-- Enable binary serialization for .NET 9 -->
     <PackageReference Include="System.Runtime.Serialization.Formatters" Version="9.0.0" Condition="$(USE_NATIVE_SDK_VERSION) != 'true'" />
--- a/tracer/build/_build/_build.csproj
+++ b/tracer/build/_build/_build.csproj
@@ -42,7 +42,7 @@
     <!-- Explicitly override the version of octokit used by Nuke -->
     <PackageReference Include="Octokit" Version="10.0.0" />
     <PackageReference Include="Perfolizer" Version="0.2.1" />
-    <PackageReference Include="Sep" Version="0.9.0" />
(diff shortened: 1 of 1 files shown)


andrewlock (approved):
You know - I saw Sep mentioned on yesterday's [asp.net community standup](https://www.youtube.com/watch?v=xs_V3VqkxHQ&list=PLdo4fOcmZ0oX-DBuRG4u58ZTAJgBAeQ-t&index=0) and I thought the same thing 😅 Thanks! 😄

nietras:
> I saw Sep mentioned on yesterday's [asp.net community standup](https://www.youtube.com/watch?v=xs_V3VqkxHQ&list=PLdo4fOcmZ0oX-DBuRG4u58ZTAJgBAeQ-t&index=0) 

Really? Hope they were kind 😅 I'll take a look, thanks for the heads up.

nietras:
> net7. 0

ah okay, yeah I dropped support for that due to needing UnsafeAccessor, at least I can change to 0.7.0 then for now.

You mentioned this issue with centos7 in your blog I think, weird they dropped support for it. Any technical reasons why?

nietras:
do you multi target? Should I switch on tfm?

nietras:
https://github.com/dotnet/runtime/issues/83428

old glibc I guess being an issue and of course centos7 being eol.

for now I've changed to 0.7.0, can add multi targetting if need be. if net7.0 etc.

andrewlock:
> yeah I dropped support for that due to needing UnsafeAccessor

Totally reasonable, .NET 7 is EOL anyway 🤷‍♂️ 

> old glibc I guess being an issue and of course centos7 being eol.

Yeah, exactly. It's reasonable enough for them to do, but it really sucks for us 😅 

> for now I've changed to 0.7.0

perfect, thanks. I don't think multi-targeting is worth it at this point.

nietras:
Given 
```xml
    <TargetFramework>net9.0</TargetFramework>
    <TargetFramework Condition="$(USE_NATIVE_SDK_VERSION) == 'true'">net7.0</TargetFramework>
    <TargetFramework Condition="$(OS.StartsWith('Windows')) AND '$(NUKE_NOTIFY)' != ''">$(TargetFramework)-windows10.0.19041</TargetFramework>
    <DefineConstants Condition="$(OS.StartsWith('Windows')) AND '$(NUKE_NOTIFY)' != ''">$(DefineConstants);NUKE_NOTIFY</DefineConstants>
```
I think it would be easy enough to add a `Condition="$(TargetFramework.StartsWith("net7.0"))"` that uses 0.7.0 instead of 0.9.0 :)

andrewlock:
Absolutely it would, but sep is only used as part of Nuke, in a minor way, so I don't think it's worth the extra complexity (it's complex enough as-is 😅)

PR #6665, opened 2025-02-13
Tack on apm-idm-dotnet reviewers to auto-bump PRs

bouwkast:
## Summary of changes

Adds `apm-idm-dotnet` as a reviewer to auto PRs. 

## Reason for change

I get a summary of open PRs, but these are missing

## Implementation details

Added the teams
Updated to `team-reviewers` as that seems to be more correct according to documentation

## Other details

Note: this shouldn't make `apm-idm-dotnet` a _required_ reviewer. If it does then revert this change.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,6 +8,8 @@ updates:
     labels:
       - "dependencies"
       - "area:dependabot"
+    reviewers:
+      - "DataDog/apm-idm-dotnet"
     ignore:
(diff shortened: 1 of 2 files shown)


andrewlock (approved):
This just changes the _suggested_ reviewers right, it doesn't change anything about codeowners?

bouwkast:
> This just changes the _suggested_ reviewers right, it doesn't change anything about codeowners?

Correct, this _should_ be just suggested reviewers, if it actually _requires_ them which I don't think it will I'll revert it

PR #6666, opened 2025-02-13
fix(telemetry): only report SSI configuration if set

dmehala:
### What does this PR do?
This PR ensures that telemetry configuration is only reported when it has valid values. Since the `value` field is mandatory for configuration entries (as specified in the [schema](https://github.com/DataDog/instrumentation-telemetry-api-docs/blob/main/GeneratedDocumentation/ApiDocs/v2/SchemaDocumentation/Schemas/conf_key_value.md)), reporting entries without values results in invalid payloads that our intake do not or entirely process.

### Motivation
To reduce the number of invalid payloads and ensure the tracer conforms to the expected schema.

## Other details
This is a bandaid. The telemetry API currently accepts `null` for values while this should be forbidden but I lack the time and knowledge on the tracer to offer a better PR at the moment.


andrewlock (commented):
This change won't really solve the issue. At it's _core_ the telemetry config allows nulls, and that has _always_ been the case 🤔 

IMO, `null` _is_ a valid value to set. It is different to "value missing", and personally I think we should address this in the intake rather than here...

dmehala:
Discussed on Slack. .NET team will implement a long solution.

PR #6681, opened 2025-02-18
Add workflows for scheduled code freeze and AAS deployment

bouwkast:
## Summary of changes

This adds three new workflows to support a scheduled code freeze and test app deployment every _even_ week number on Friday afternoon EST time.

## Reason for change

Think it'd be better to more toward a more proactive approach for code freezing and releasing.

## Implementation details

- `scheduled_code_freeze.yml`
  - Runs every Friday at 4PM EST
  - Odd week -> skipped freeze
  - `Skip Scheduled Code Freeze` is `open` -> skipped freeze
  - Even week -> freeze
    - Update PR status
- `scheduled_aas_deploy.yml`
  - Runs every Friday at 4:30PM EST
    - Unsure how to trigger it to be ready when the AzDo stage has built the necessary artifacts, so this is hacky
    - Odd week -> skipped deploy
    - `Skip Scheduled Code Freeze` is `open` -> skipped deploy
      - close the `Skip Scheduled Code Freeze`
    - Even week -> deploy
- `create_skip_code_freeze.yml`
  - Opens `Skip Scheduled Code Freeze`
- `pr-status-updater/action.yml`
  - Copy / modification of `code-freeze/action.yml` to be configurable, but haven't tested this with large number of PRs
- Skip Scheduled Code Freeze Milestone: https://github.com/DataDog/dd-trace-dotnet/milestone/183

## Test coverage

I testing this out in a different repository as that was easier / safer.

Let me know if you want access to that :)

The `pr-status-updater/action.yml` is a version of `code-freeze/action.yml` but more configurable, only thing is that I needed to use `curl` instead of the GitHub action and I didn't test on the same number of PRs. **Let me know if it makes more sense to swap to `code-freeze/action.yml`**

I left in the workflow dispatch stuff as it is easier to just leave them in for testing purposes


.github/actions/pr-status-updater/action.yml
```diff
@@ -0,0 +1,68 @@
...
+    description: 'Name of the workflow file for the target URL'
+    required: true
+  page_size:
+    description: 'Number of PRs to process per page'
+    required: false
+    default: '25'
+  max_pages:
+    description: 'Maximum number of pages to process'
+    required: false
+    default: '5'
```
andrewlock, line 23: We should probably bump this, seeing as we already have 127 PRs 😅 
```suggestion
    default: '6'
```

.github/actions/pr-status-updater/action.yml
```diff
@@ -0,0 +1,68 @@
+# Copied / Extracted from `code-freeze/action.yml`:
```
andrewlock, line 1: I didn't quite follow _why_ you needed to duplicate this? 🤔 Is it because you've extended it? We could update the other places to use this one if it has more features etc?

.github/workflows/scheduled_code_freeze.yml
```diff
@@ -0,0 +1,105 @@
+name: Scheduled Code Freeze
+
+on:
+  schedule:
+    - cron: '0 21 * * FRI'  # Every Friday at 9 PM UTC
```
andrewlock, line 5: nit: seeing as we don't care about the _exact_ time, we can offset the minutes, to reduce the chance of github throttling it (not that we care that much, as long as it's not skipped)

```suggestion
    - cron: '3 21 * * FRI'  # Every Friday at 9:03 PM UTC
```

.github/workflows/scheduled_code_freeze.yml
```diff
@@ -0,0 +1,105 @@
...
+        if: steps.check_skip.outputs.should_proceed == 'true' && success()
+        uses: actions/checkout@ee0669bd1cc54295c223e0bb666b733df41de1c5 # v2.7.0
+
+      # Update PR statuses
+      - uses: ./.github/actions/pr-status-updater
+        if: steps.check_skip.outputs.should_proceed == 'true' && success()
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          state: failure
+          workflow_name: scheduled_code_freeze.yml
```
andrewlock, line 105: ah, ok, so this is used to provide the clickable link in the UI on the code-freeze, right?

.github/workflows/scheduled_aas_deploy.yml
```diff
@@ -0,0 +1,101 @@
+name: Scheduled AAS Deployment
+
+on:
+  schedule:
+    - cron: '30 21 * * FRI'  # Every Friday at 9:30 PM UTC
```
andrewlock, line 5: Theoretically, someone could merge _just_ before the code freeze. Given we need to wait for all the release artifacts to be ready, which can take ~1hr in many cases, it may be safest to push this back an hour? It probably doesn't make much difference overall, and may reduce number of failures?
```suggestion
    - cron: '30 22 * * FRI'  # Every Friday at 10:30 PM UTC
```

.github/workflows/scheduled_aas_deploy.yml
```diff
@@ -0,0 +1,101 @@
...
+          owner: ${{ github.repository_owner }}
+          repo: ${{ github.event.repository.name }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Process skip state
+        if: steps.should_run.outputs.should_proceed == 'true'
+        id: check_skip
+        run: |
+          state=$(echo '${{ steps.skip_milestone.outputs.data }}' | jq -r '.state')
```
andrewlock, line 70: I assume the skip milestone is a safety check way of skipping the freeze and deploy? Like, if we _know_ we don't want to freeze for some reason? 

Presumably we wouldn't expect to need to use this often, and worst case you could run the end code freeze workflow too, right? We would manually open the milestone to enable it right?

.github/actions/pr-status-updater/action.yml
bouwkast: Okay so initially I thought the `targetUrl` needed to be mapped to the workflow file(s) so I made it configurable and then learned that it didn't need to be.

It just makes the link in the `[code_freeze](https://github.com/DataDog/dd-trace-dotnet/actions/workflows/auto_code_freeze_block_pr.yml) — No code freeze is in place` clickable

I wasn't entirely sure on how to loop the current one but I can just use that one if it makes more sense.

.github/workflows/scheduled_code_freeze.yml
bouwkast: Yeah :) 

I initially thought it was what controlled whether or not that status was working and not just making it clickable.

.github/workflows/scheduled_aas_deploy.yml
bouwkast: Ahh yeah

That seems fine to me

.github/workflows/scheduled_aas_deploy.yml
bouwkast: Correct!

The idea that I have is:

- Thursday we have an automated slack message about upcoming freeze
- If we don't want to freeze we run that to skip it without needing to do a code change on them
- Code freeze would be skipped
- Deploy would be skipped and then the deploy closes the skip milestone

I'm working on a doc / diagram to show the flow of everything

.github/actions/pr-status-updater/action.yml
andrewlock: I mean, it would be _best_ to not have the duplication if we can avoid it, but it's not a blocker IMO 🙂

bouwkast: Okay, my plan is to see how it works on Friday

If it fails then I'll run the freeze manually and then PR this to the old (current) action or fix if the fix is obvious.

If it runs fine then I'll swap the current way to this new way

PR #6690, opened 2025-02-20
Add NuGet API verification check

andrewlock:
## Summary of changes

Adds an early stage in the create_draft_release workflow to check that the Nuget API key is valid

## Reason for change

We recently did a release, and the "upload NuGet packages step failed". Unfortunately, this happens _after_ the draft release is created which is essentially a one-way trip (it triggers downstream things).

## Implementation details

Unfortunately, due to 
- https://github.com/NuGet/Home/issues/13475
- https://github.com/NuGet/NuGetGallery/issues/8283

There's no good way to test the key _Without_ pushing a package.

So this takes a hacky approach:
- Download one of our packages
- Try to push the package (and set `--skip-duplicate` so that it becomes a no-op)
- If it succeeds, it's a no-op, if the key is invalid, it fails, and we bail

## Test coverage

Tested the steps locally, and they work so 🤞

### DIFF
--- a/.github/workflows/create_draft_release.yml
+++ b/.github/workflows/create_draft_release.yml
@@ -56,7 +56,7 @@ jobs:
           fi
 
           echo "Pushing package to test NuGet API key validity"
-          if ! dotnet nuget push "Datadog.Trace.Annotations.1.0.0.nupkg" --api-key "meh" --skip-duplicate --source https://api.nuget.org/v3/index.json; then
+          if ! dotnet nuget push "Datadog.Trace.Annotations.1.0.0.nupkg" --api-key ${{ secrets.NUGET_API_KEY }} --skip-duplicate --source https://api.nuget.org/v3/index.json; then
             echo "Push failed - the API key is likely not valid" >&2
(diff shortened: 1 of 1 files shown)


.github/workflows/create_draft_release.yml
```diff
@@ -47,6 +47,26 @@ jobs:
...
+      - name: "Check NuGet token validity"
+        run: |
+          echo "Downloading test package"
+          if ! curl -fL -o Datadog.Trace.Annotations.1.0.0.nupkg https://www.nuget.org/api/v2/package/Datadog.Trace.Annotations/1.0.0; then
+            echo "Download failed!" >&2
+            exit 1
+          fi
+
+          echo "Pushing package to test NuGet API key validity"
+          if ! dotnet nuget push "Datadog.Trace.Annotations.1.0.0.nupkg" --api-key "meh" --skip-duplicate --source https://api.nuget.org/v3/index.json; then
```
bouwkast, line 59: 😅 is `--api-key "meh"` correct?

bouwkast (approved):

PR #6696, opened 2025-02-21
Fix dependabot configuration for gh-actions-packages group

bouwkast:
## Summary of changes

Gave up, just fixed the dependabot file

## Reason for change

It was broken and nothing was running

`The property '#/updates/5/groups/gh-actions-packages/update-types/0' value "version-update:semver-patch" did not match one of the following values: major, minor, patch`

## Implementation details

Deleted `version-update:semver-` not available in group


andrewlock (approved):
Why not, let's try it!

PR #6717, opened 2025-02-25
Add notification channel for build pipeline

randomanderson:
## Summary of changes

This PR adds a channel for notifications for pipeline failures

## Reason for change

The shared pipeline has added notifications for failures during a release pipeline. More notifications may be added in the future.

## Implementation details
Of the channels available, I chose `#apm-dotnet-bots`

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -13,6 +13,7 @@ include:
 variables:
   DOTNET_PACKAGE_VERSION:
     description: "Used by the package stage when triggered manually"
+  REPO_NOTIFICATION_CHANNEL: "#apm-dotnet-error-bots"
 
 build:
   except:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -13,7 +13,7 @@ include:
 variables:
   DOTNET_PACKAGE_VERSION:
     description: "Used by the package stage when triggered manually"
-  REPO_NOTIFICATION_CHANNEL: "#apm-dotnet-error-bots"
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -13,6 +13,7 @@ include:
 variables:
   DOTNET_PACKAGE_VERSION:
     description: "Used by the package stage when triggered manually"
+  REPO_NOTIFICATION_CHANNEL: "#apm-dotnet-error-bots"
```
andrewlock, line 16: I think `#apm-dotnet-bots` _may_ be the better channel (`#apm-dotnet-error-bots` is for prod errors)
```suggestion
  REPO_NOTIFICATION_CHANNEL: "#apm-dotnet-bots"
```

andrewlock (approved):
LGTM

.gitlab-ci.yml
bouwkast: 😆 

I was literally going to comment that I thought `#apm-dotnet-error-bots` may be a better one even though that yes it is supposed to be for prod errors.

My only concern is that `#apm-dotnet-bots` can be a tad noisy so it may slip through unnoticed there, but maybe not

I say we can try it and if they fail and we don't notice them then we can re-route.

bouwkast (approved):
My only concern is that we will now get a new notification when the build fails (which is a great change), but I am unsure how to act on that notification.

Granted I haven't seen it yet, so maybe it is just go to the GitLab build and click re-run?

PR #6726, opened 2025-02-27
[WIP] use built-in .NET runtime types instead of vendored types when possible

lucaspimentel:
## Summary of changes

When targeting .NET Core 3.1 and .NET 6, do not include the files copied ("vendored") from the .NET runtime. Instead, use the built-in types.

TODO update table

| Runtime         | Before  | After   | After (`global using`) |
|-----------------|---------|---------|------------------------|
| `netcoreapp3.1` | 7.59 MB | 7.29 MB | 7.25 MB
| `net6.0`        | 7.62 MB | 7.32 MB | 7.28 MB

## Reason for change

- smaller artifacts for .NET Core 3.1 and .NET 6.0 (important for Serverless)
- (bonus) consistency: we currently have different ways of choosing built-in types vs vendored types

## Implementation details

- `#if NETCOREAPP3_1_OR_GREATER` and `global using` in `GlobalUsings.cs` to choose the namespace based on target framework (TFM)
- When targeting `netcoreapp3.1` or `net6.0`, exclude most of the files in:
```
Vendors/System.Collections.Immutable;
Vendors/System.Memory;
Vendors/System.Private.CoreLib;
Vendors/System.Runtime.CompilerServices.Unsafe;
```

We're keeping `Vendors/System.Reflection.Metadata` for now because we access non-public members from those types. Also keep these files which are used elsewhere:
```
Vendors/System.Private.CoreLib/src/System/Runtime/InteropServices/LibraryImportAttribute.cs
Vendors/System.Private.CoreLib/src/System/Diagnostics/CodeAnalysis/7NullableAttributes.cs
Vendors/System.Private.CoreLib/src/System/Runtime/InteropServices/StringMarshalling.cs
Vendors/System.Private.CoreLib/SR.cs
```

## Test coverage

No behavior changes expected. All passing tests should keep passing.

## Other details
n/a


andrewlock:

- [This PR (6726)](https://github.com/DataDog/dd-trace-dotnet/tree/3acc23d61f788bba1bafb39c0c03dae6b1a6fbb2)
- [master](https://github.com/DataDog/dd-trace-dotnet/tree/73a4d8f5cc835eafbd6b9d454e84da9d448a4f2f)

```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Bailout
    This PR (6726) - mean (72ms)  : 71, 73
     .   : milestone, 72,
    master - mean (72ms)  : 71, 73
     .   : milestone, 72,

    section Baseline
    This PR (6726) - mean (68ms)  : 66, 70
     .   : milestone, 68,
    master - mean (68ms)  : 65, 72
     .   : milestone, 68,

    This PR (6726) - mean (1,054ms)  : 999, 1109
     .   : milestone, 1054,
    master - mean (1,047ms)  : 998, 1096
     .   : milestone, 1047,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Bailout
    This PR (6726) - mean (106ms)  : 105, 107
     .   : milestone, 106,
    master - mean (106ms)  : 105, 107
     .   : milestone, 106,

    section Baseline
    This PR (6726) - mean (106ms)  : 104, 108
     .   : milestone, 106,
    master - mean (106ms)  : 104, 108
     .   : milestone, 106,

    This PR (6726) - mean (730ms)  : 710, 749
     .   : milestone, 730,
    master - mean (744ms)  : 721, 766
     .   : milestone, 744,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Bailout
    This PR (6726) - mean (100ms)  : 99, 101
     .   : milestone, 100,
    master - mean (100ms)  : 99, 101
     .   : milestone, 100,

    section Baseline
    This PR (6726) - mean (99ms)  : 97, 102
     .   : milestone, 99,
    master - mean (100ms)  : 97, 102
     .   : milestone, 100,

    This PR (6726) - mean (766ms)  : 723, 809
     .   : milestone, 766,
    master - mean (772ms)  : 719, 824
     .   : milestone, 772,

```
```mermaid
gantt
    title Execution time (ms) FakeDbCommand (.NET 8) 
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Bailout
    This PR (6726) - mean (93ms)  : 92, 94
     .   : milestone, 93,
    master - mean (93ms)  : 92, 94
     .   : milestone, 93,

    section Baseline
    This PR (6726) - mean (92ms)  : 90, 94
     .   : milestone, 92,
    master - mean (92ms)  : 90, 94
     .   : milestone, 92,

    This PR (6726) - mean (649ms)  : 634, 664
     .   : milestone, 649,
    master - mean (665ms)  : 652, 679
     .   : milestone, 665,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Bailout
    This PR (6726) - mean (207ms)  : 203, 211
     .   : milestone, 207,
    master - mean (204ms)  : 200, 208
     .   : milestone, 204,

    section Baseline
    This PR (6726) - mean (202ms)  : 198, 206
     .   : milestone, 202,
    master - mean (201ms)  : 195, 207
     .   : milestone, 201,

    This PR (6726) - mean (1,211ms)  : 1128, 1293
     .   : milestone, 1211,
    master - mean (1,190ms)  : 1136, 1244
     .   : milestone, 1190,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Bailout
    This PR (6726) - mean (291ms)  : 284, 299
     .   : milestone, 291,
    master - mean (288ms)  : 282, 294
     .   : milestone, 288,

    section Baseline
    This PR (6726) - mean (292ms)  : 286, 298
     .   : milestone, 292,
    master - mean (288ms)  : 283, 293
     .   : milestone, 288,

    This PR (6726) - mean (966ms)  : 913, 1018
     .   : milestone, 966,
    master - mean (965ms)  : 923, 1006
     .   : milestone, 965,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Bailout
    This PR (6726) - mean (295ms)  : 285, 305
     .   : milestone, 295,
    master - mean (293ms)  : 286, 299
     .   : milestone, 293,

    section Baseline
    This PR (6726) - mean (296ms)  : 285, 306
     .   : milestone, 296,
    master - mean (292ms)  : 286, 298
     .   : milestone, 292,

    This PR (6726) - mean (1,009ms)  : 969, 1048
     .   : milestone, 1009,
    master - mean (1,016ms)  : 955, 1076
     .   : milestone, 1016,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Bailout
    This PR (6726) - mean (279ms)  : 272, 286
     .   : milestone, 279,
    master - mean (283ms)  : 274, 292
     .   : milestone, 283,

    section Baseline
    This PR (6726) - mean (282ms)  : 271, 294
     .   : milestone, 282,
    master - mean (282ms)  : 274, 289
     .   : milestone, 282,

    This PR (6726) - mean (862ms)  : 840, 884
     .   : milestone, 862,
    master - mean (879ms)  : 856, 901
     .   : milestone, 879,

```

tracer/src/Datadog.Trace/Ci/Coverage/Util/FileBitmap.cs
```diff
@@ -227,7 +227,7 @@ public static FileBitmap Or(FileBitmap fileBitmapA, FileBitmap fileBitmapB, bool
         }
 #endif
 
-#if NETCOREAPP3_0_OR_GREATER
+#if NETCOREAPP3_1_OR_GREATER
```
andrewlock, line 230: Why the 3_0 => 3_1 change? 🤔 I know it doesn't matter as it has the same result currently, just curious

tracer/src/Datadog.Trace/Debugger/Helpers/MemoryExtensions.cs
```diff
@@ -16,7 +14,7 @@ internal static class MemoryExtensions
         /// </summary>
         internal static IMemoryOwner<T> EnlargeBuffer<T>(this IMemoryOwner<T> memoryOwner, int currentSize)
         {
-            var newMemory = ArrayMemoryPool<T>.Shared.Rent(currentSize * 2);
+            var newMemory = MemoryPool<T>.Shared.Rent(currentSize * 2);
```
andrewlock, line 17: For reviewers, this looks like it might be a change to worry about given that `ArrayMemoryPool` _derives_ from `MemoryPool`. But the `MemoryPool<T>.Shared` instance is actually an instance of `ArrayMemoryPool<T>` so this whole code is effectively identical

tracer/src/Datadog.Trace/PDBs/DatadogMetadataReader.cs
```diff
@@ -657,6 +658,13 @@ internal bool IsCompilerGeneratedAttributeDefinedOnType(int typeToken)
             return IsCompilerGeneratedAttributeDefine(attributes);
         }
 
+        internal bool IsCompilerGeneratedAttributeDefinedOnType(TypeDefinitionHandle typeHandle)
+        {
+            var nestedType = MetadataReader.GetTypeDefinition(typeHandle);
+            var attributes = nestedType.GetCustomAttributes();
+            return IsCompilerGeneratedAttributeDefine(attributes);
+        }
+
```
andrewlock: AFAICT, this is not used? 🤔 
```suggestion
```

tracer/src/Datadog.Trace/Util/BitConverterShim.cs
```diff
@@ -5,12 +5,10 @@
 
 #nullable enable
 
-#if !NETCOREAPP
+#if !NETCOREAPP3_1_OR_GREATER
```
andrewlock, line 8: There's a potential future where we don't ship a netstandard dll, and instead ship a .NET Core 2.x dll. In that future, this code path won't be hit, whereas it would otherwise, so setting the "real" minimum level is generally preferable. Is this _really_ .NET Core 3.1+, or _really_ .NET Core 2.x+?

tracer/src/Datadog.Trace/Vendors/System.Reflection.Metadata/src/System/Reflection/Internal/Utilities/EnumerableExtensions.cs
```diff
@@ -22,24 +21,6 @@ namespace Datadog.Trace.VendoredMicrosoftCode.System.Reflection.Internal
     /// </summary>
     internal static class EnumerableExtensions
   {
-    public static T? FirstOrDefault<T>(this ImmutableArray<T> collection, Func<T, bool> predicate)
```
andrewlock, line 25: For reviewers: These two are deleted for the reasons described in the comment above:

```csharp
/// <summary>
/// Replacements for System.Linq to avoid an unnecessary dependency.
/// Parameter and return types strengthened to actual internal usage as an optimization.
/// </summary>
```

Seeing as we _do_ reference System.Linq, they're both unecessary, and cause ambiguity. Also, they're trivial implementations, so doesn't seem like we're really losing anything there

tracer/src/Datadog.Trace/Datadog.Trace.csproj
```diff
@@ -100,6 +100,45 @@
...
+  <PropertyGroup>
+    <DotNetRuntimeFiles>
+      Vendors/System.Collections.Immutable/**;
+      Vendors/System.Memory/**;
+      Vendors/System.Private.CoreLib/**;
+      Vendors/System.Runtime.CompilerServices.Unsafe/**;
+    </DotNetRuntimeFiles>
+  </PropertyGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp3.1' OR '$(TargetFramework)' == 'net6.0' ">
```
andrewlock: We can make this a bit more robust to future TFM changes using this:
```suggestion
    <ItemGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible($(TargetFramework), 'netcoreapp3.1'))">
```

tracer/src/Datadog.Trace/Datadog.Trace.csproj
```diff
@@ -100,6 +100,45 @@
...
+    <None Include="$(DotNetRuntimeFiles)" />
+
+    <!-- do not exclude these files from the folders listed above -->
+    <Compile Include="Vendors/System.Private.CoreLib/src/System/Diagnostics/CodeAnalysis/7NullableAttributes.cs" />
+    <Compile Include="Vendors/System.Private.CoreLib/src/System/Runtime/InteropServices/LibraryImportAttribute.cs" />
+    <Compile Include="Vendors/System.Private.CoreLib/src/System/Runtime/InteropServices/StringMarshalling.cs" />
+    <Compile Include="Vendors/System.Private.CoreLib/SR.cs" />
+
+    <!-- We use Unsafe.IsNullRef() which was added in .NET 5 -->
+    <Compile Include="Vendors/System.Runtime.CompilerServices.Unsafe/Unsafe.cs" />
```
andrewlock: If we have it in .NET 5, we can exclude it from there up?
```suggestion
    <Compile Include="Vendors/System.Runtime.CompilerServices.Unsafe/Unsafe.cs"
        Condition="!$([MSBuild]::IsTargetFrameworkCompatible($(TargetFramework), 'net5.0'))" />
```

tracer/src/Datadog.Trace/Datadog.Trace.csproj
```diff
@@ -100,6 +100,45 @@
...
+    </DotNetRuntimeFiles>
+  </PropertyGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp3.1' OR '$(TargetFramework)' == 'net6.0' ">
+    <Compile Remove="$(DotNetRuntimeFiles)" />
+    <EmbeddedResource Remove="$(DotNetRuntimeFiles)" />
+    <None Include="$(DotNetRuntimeFiles)" />
+
+    <!-- do not exclude these files from the folders listed above -->
+    <Compile Include="Vendors/System.Private.CoreLib/src/System/Diagnostics/CodeAnalysis/7NullableAttributes.cs" />
```
andrewlock: This one almost certainly _shouldn't_ be included, because it's basically used "incorrectly" in that case 🤔 We instead should fix the usages to use the "real" attributes and/or add our own "proper" version if required (I'll push a commit to do that subsequently)

In _fact_, we shouldn't be vendoring those attributes _at all_, doing so is "incorrect", because they're not recognized by the compiler.

tracer/src/Datadog.Trace/GlobalUsings.cs
```diff
@@ -5,10 +5,27 @@
...
+global using Datadog.Trace.ExtensionMethods;
 global using ThrowHelper = Datadog.Trace.Util.ThrowHelper;
 
-#if NETFRAMEWORK || NETSTANDARD2_0
-global using Datadog.Trace.VendoredMicrosoftCode.System;
+#if NET5_0_OR_GREATER
+global using Unsafe = System.Runtime.CompilerServices.Unsafe;
+#else
+// we use some methods from Unsafe that were added in .NET 5
+global using Unsafe = Datadog.Trace.VendoredMicrosoftCode.System.Runtime.CompilerServices.Unsafe.Unsafe;
```
andrewlock, line 15: Given we literally use _one_ method from .NET 5 (`IsNullRef()`) - I wonder if we should just extract that out specifically (i.e. put it in its own separate file that only targets `NETCOREAPP_3_1_OR_GREATER && !NET_5_OR_GREATER` ), and use the built-in `Unsafe` types for everything else. Seems like that might simplify how we need to think about this in general, WDYT? 🤔

tracer/src/Datadog.Trace/Datadog.Trace.csproj
andrewlock: - The nullable attributes won't actually work, they need to be in the correct namespace, and we already polyfill those
- The LibraryImport attribute drives a source generator, but that generator is only present in .NET 7+, so it does nothing here.
- This also simplifies the vendoring in general

tracer/src/Datadog.Trace/Ci/Coverage/Util/FileBitmap.cs
lucaspimentel: Just consistency. We use `NETCOREAPP3_1_OR_GREATER` a lot more than `NETCOREAPP3_0_OR_GREATER`. And I'm secretly hoping it will make things easier later if we ever have to search/replace some of these (edit: which is silly because we can just search for both).

Same with `NET6_0_OR_GREATER` vs `NET5_0_OR_GREATER`, for example.

tracer/src/Datadog.Trace/Debugger/Helpers/MemoryExtensions.cs
lucaspimentel: The static `Shared` property is actually on the `MemoryPool<T>` type and you can't override static members, so we were always using that one. This would compile to `MemoryPool<T>.Shared`.

tracer/src/Datadog.Trace/Datadog.Trace.csproj
lucaspimentel: nice, thanks!

tracer/src/Datadog.Trace/Datadog.Trace.csproj
lucaspimentel: Yes, thanks.

tracer/src/Datadog.Trace/Datadog.Trace.csproj
lucaspimentel: > they need to be in the correct namespace, and we already polyfill those

Oops, good catch

tracer/src/Datadog.Trace/Ci/Coverage/Util/FileBitmap.cs
andrewlock: I see. I'm honestly torn. The reason being, having the "wrong" values for these was a pain when I was trying to change our target TFMs a while back. So it's targeting a different pain point. So meh 😄

tracer/src/Datadog.Trace/GlobalUsings.cs
lucaspimentel: Hah, I actually did that a couple of days ago then changed my mind. I was trying something like this:

```csharp
static Unsafe
{
    IsNullRef()
    {
#if net5+
        // defer to the built-in version
        return System.Unsafe.IsNullRef()
#else
        // copied from the current code
        Ldarg(nameof(source));
        Ldc_I4_0();
        Conv_U();
        Ceq();
        return IL.Return<bool>();
#endif
    }
}
```

I was having naming conflicts (tried `UnsafeHelper` for a while) and it was ugly, which is why I removed it. But yeah, go for it.

tracer/src/Datadog.Trace/Util/BitConverterShim.cs
lucaspimentel: Just like the `NETCOREAPP3_0_OR_GREATER` comment elsewhere, I was just trying to be consistent. Also, we're talking about .NET Core 2.x, so... _meh?_

lucaspimentel: fwiw it looks like we don't need this `#if` at all anymore? it looks like `GlobalUsings.cs` is handling this correctly.

lucaspimentel: ...ish. It's still using the vendored `Unsafe.ReadUnaligned` in .NET Core 3.1 even though that method was added in .NET Core 3.0. It might be a similar issue to `UnSafe.IsNullRef()` which was added in .NET 5.

andrewlock: > Also, we're talking about .NET Core 2.x,

Yeah, I mean... that was the original 3.x plan 😅 Because it's actually a big improvement over netstandard, but you know, reasons...

> It's still using the vendored Unsafe.ReadUnaligned in .NET Core 3.1

I mean, we're using the vendored `Unsafe` _everywhere_ for 3.1 now 😉

tracer/src/Datadog.Trace/GlobalUsings.cs
andrewlock: Yeah, I was torn honestly. I was just going to move that one method to a completely different file, exclude the "main" one, and then `#if` the namespace, but that has issues too, which might be even more confusing tbh (seeing as we can't extend `Unsafe` unless we use [extension members](https://andrewlock.net/exploring-dotnet-10-preview-features-3-csharp-14-extensions-members/) which I don't know if we should get into 😅

tracer/src/Datadog.Trace/Util/BitConverterShim.cs
lucaspimentel: oh right

tracer/test/Datadog.Trace.Tests/VendoredCodeTests.cs
```diff
@@ -35,7 +35,7 @@ public void UnsafeTypeVerifyIlTest()
         {
 #if DEBUG
             throw new Xunit.SkipException("This test requires RELEASE mode and will fail in DEBUG mode on some target frameworks");
-#else
+#elif !NETCOREAPP3_1_OR_GREATER
             var originalType = typeof(System.Runtime.CompilerServices.Unsafe);
             var vendoredType = typeof(VendoredMicrosoftCode.System.Runtime.CompilerServices.Unsafe.Unsafe);
```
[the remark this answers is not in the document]
lucaspimentel: makes sense

tracer/src/Datadog.Trace/Datadog.Trace.csproj
```diff
@@ -100,6 +100,39 @@
     <None Remove="AppSec\Waf\ConfigFiles\apisecurity-config.json" />
   </ItemGroup>
 
+  <!-- we don't need the vendored .NET runtime code when targetting .NET Core 3.1+ -->
```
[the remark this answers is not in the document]
lucaspimentel: TIL some new English ~rules~ suggestions (American vs British): https://english.stackexchange.com/questions/205815/is-targetted-a-standard-british-english-spelling

PR #6730, opened 2025-02-28
Fix: Rabbitmq7 header injection overwrites user supplied basic properties

### LINKED ISSUE #6723: [BUG]: RabbitMQ Basic Properties overwritten to default values
### Tracer Version(s)

3.11.1

### Operating system and platform

Linux

### Instrumentation Mode

manual with nuget package

### TFM

net9.0

### Bug Report

If `BasicPublishAsync` is called  with a writable `BasicProperties` then `PopulateBasicPropertiesHeaders` in rabbitmq-dotnet-client will return null as no new object is created. In turn the instrumentation code will get a return value of null and create a new basic properties object resulting in all values having default values instead of the ones set by the caller.

Implementation in rabbitmq-dotnet-client: https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/85fd42fc6e245dc68901802967acbccc216b5980/projects/RabbitMQ.Client/Impl/Channel.BasicPublish.cs#L190

Implementation in dd-trace: https://github.com/DataDog/dd-trace-dotnet/blob/f5e1c396ee1a4b78fbb015943d27012c362fe779/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/RabbitMQ/PopulateBasicPropertiesHeadersIntegration.cs#L48

### Reproduction Code

_No response_

johang88:
Note: Created as draft as I have not actually run or tested the changes 😶

## Summary of changes
Fixes the header injection so that basic properties are preserved from the supplied argument list in case `PopulateBasicPropertiesHeaders` returns null which happens if the basic properties argument is writable or if no changes had to be made to it.

## Reason for change
Fixes a bug that would remove all user supplied basic properties.

## Implementation details
Stores the basicproperties argument in the active scope and retrieves it a null return value is encountered. Changed `CachedBasicPropertiesHelper` to call the copy constructor that takes a `IReadOnlyBasicProperties` as argument.

## Test coverage
Added basic properties to one of the rabbitmq7 test methods as well as assertions.

## Other details
Fixes #6723

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/RabbitMQ/CachedBasicPropertiesHelper.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/RabbitMQ/CachedBasicPropertiesHelper.cs
@@ -27,12 +27,13 @@ internal static class CachedBasicPropertiesHelper<TBasicProperties>
         var createBasicPropertiesMethod = new DynamicMethod(
             $"TypeActivator_{targetType.Name}_{parameterType.Name}",
             targetType,
-            [parameterType],
+            [typeof(object)],
             typeof(DuckType).Module,
             true);
 
         var il = createBasicPropertiesMethod.GetILGenerator();
-        il.Emit(OpCodes.Ldarg_0);
+        il.Emit(OpCodes.Ldarg_0); // Load first argument (object).
+        il.Emit(OpCodes.Castclass, parameterType); // Cast to IReadOnlyBasicProperties
         il.Emit(OpCodes.Newobj, constructor);
         il.Emit(OpCodes.Ret);
         Activator = (Func<object, object>)createBasicPropertiesMethod.CreateDelegate(typeof(Func<object, object>), targetType);
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/RabbitMQ/CachedBasicPropertiesHelper.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/RabbitMQ/CachedBasicPropertiesHelper.cs
@@ -1,9 +1,13 @@
-﻿// <copyright file="CachedBasicPropertiesHelper.cs" company="Datadog">
+// <copyright file="CachedBasicPropertiesHelper.cs" company="Datadog">
 // Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2 License.
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
 // </copyright>
 
 #nullable enable
+using System;
+using System.Reflection.Emit;
+using System.Runtime.InteropServices;
+using Datadog.Trace.DuckTyping;
 using Datadog.Trace.Util;
 
 namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.RabbitMQ;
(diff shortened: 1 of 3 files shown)


lucaspimentel:
@johang88: Thanks for this! We'll take a look, run all tests, and merge it if all goes well.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/RabbitMQ/CachedBasicPropertiesHelper.cs
```diff
@@ -1,23 +1,43 @@
...
+            targetType,
+            [parameterType],
+            typeof(DuckType).Module,
+            true);
+
+        var il = createBasicPropertiesMethod.GetILGenerator();
+        il.Emit(OpCodes.Ldarg_0);
+        il.Emit(OpCodes.Newobj, constructor);
+        il.Emit(OpCodes.Ret);
+        Activator = (Func<object, object>)createBasicPropertiesMethod.CreateDelegate(typeof(Func<object, object>), targetType);
```
bouwkast, line 38: Unfortunately it looks like this is throwing an exception when attempting to run the tests / sample

```
[ERR] Exception occurred when calling the CallTarget integration continuation. System.TypeInitializationException: The type initializer for 'Datadog.Trace.ClrProfiler.AutoInstrumentation.RabbitMQ.CachedBasicPropertiesHelper`1' threw an exception.
 ---> System.ArgumentException: Cannot bind to the target method because its signature is not compatible with that of the delegate type.
   at System.Delegate.CreateDelegateNoSecurityCheck(Type type, Object target, RuntimeMethodHandle method)
   at System.Reflection.Emit.DynamicMethod.CreateDelegate(Type delegateType, Object target)
   at Datadog.Trace.ClrProfiler.AutoInstrumentation.RabbitMQ.CachedBasicPropertiesHelper`1..cctor() 
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/RabbitMQ/PopulateBasicPropertiesHeadersIntegration.cs
```diff
@@ -45,7 +51,29 @@ public class PopulateBasicPropertiesHeadersIntegration
...
+        {
+            // Use the existing basic properties if possible,
+            // if not create new instance using the copy constructor on BasicProperties.
+            if (state.State.DuckIs<IBasicProperties>())
+            {
+                returnValue = (TReturn)state.State!;
+            }
+            else if (state.State.DuckIs<IReadOnlyBasicProperties>())
+            {
+                returnValue = CachedBasicPropertiesHelper<TReturn>.CreateHeaders(state.State!);
```
bouwkast, line 68: I wonder if we need to preserve / copy headers here?
Haven't tested it out if that is necessary though

bouwkast (changes requested):
Hey thanks a lot @johang88 

I ran the CI against this and am seeing test failures for 7.1.0 related to the new `Activator` code

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/RabbitMQ/PopulateBasicPropertiesHeadersIntegration.cs
johang88: I'll take a look and see if I can sort it out. We definitely need to copy the basic properties and headers as they will be dropped completely from the message otherwise. For example the current implementation completely breaks our rabbitmq message sender if instrumentation is enabled as the messageid, timestame, persistance settings will be removed.

bouwkast: Thank you!
And let us know if need help on this as we definitely want to fix this

johang88: I have added a commit that should fix the activator, not super happy about the cast but also not sure how to get rid of it. But I suppose optimizations can come after it's working properly :)

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/RabbitMQ/CachedBasicPropertiesHelper.cs
```diff
@@ -1,23 +1,44 @@
...
+            [typeof(object)],
+            typeof(DuckType).Module,
+            true);
+
+        var il = createBasicPropertiesMethod.GetILGenerator();
+        il.Emit(OpCodes.Ldarg_0); // Load first argument (object).
+        il.Emit(OpCodes.Castclass, parameterType); // Cast to IReadOnlyBasicProperties
+        il.Emit(OpCodes.Newobj, constructor);
+        il.Emit(OpCodes.Ret);
+        Activator = (Func<object, object>)createBasicPropertiesMethod.CreateDelegate(typeof(Func<object, object>), targetType);
```
lucaspimentel, line 39: @johang88: After some debugging, we figured out that the call to `CreateDelegate()` here was throwing an exception about the method signatures not matching. I fixed the issue by removing the `targetType` parameter, since the dynamic method will not be bound to a type (i.e. it acts like a static method, not an instance method).

We had to copy your commits into a new branch in the repository to run all our tests: see #6753. Once we get that new PR reviewed and approved, and all tests pass, we will merge that one in favor of this one. Your commits are still attributed to you as the author, though. Thanks for your help with this issue.

andrewlock:
Hi @johang88, thanks for the effort on this - it took longer than we hoped to resolve some issues in the tests, and we had to create a separate PR #6753 which has just been merged. Thanks again for raising it and proposing the fix, it's much appreciated!

PR #6774, opened 2025-03-17
Pin github actions to sha

andrewlock:
## Summary of changes

Pins the action used inside our custom actions to sha

## Reason for change

We pinned all the actions that are used in workflows, but we should pin the ones used in actions too

## Implementation details

- Check what the current version is, grab the sha, use it
- Didn't update the version at all, just used whatever we're _currently_ pinned to

## Test coverage

Nope

### DIFF
--- a/.github/actions/code-freeze/action.yml
+++ b/.github/actions/code-freeze/action.yml
@@ -15,7 +15,7 @@ inputs:
 runs:
   using: "composite"
   steps:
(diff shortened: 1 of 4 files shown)


bouwkast (approved):
Thanks!

ganeshnj (approved):

lucaspimentel (approved):

PR #6753, opened 2025-03-10
[Tracer] Fix Rabbitmq7 header injection overwrites user supplied basic properties [from external PR #6730]

lucaspimentel:
Based on #6730 by @johang88.

## Summary of changes
Fixes the header injection so that basic properties are preserved from the supplied argument list in case `PopulateBasicPropertiesHeaders` returns null which happens if the basic properties argument is writable or if no changes had to be made to it.

## Reason for change
Fixes a bug that would remove all user supplied basic properties.

## Implementation details
Stores the basicproperties argument in the active scope and retrieves it a null return value is encountered. Changed `CachedBasicPropertiesHelper` to call the copy constructor that takes a `IReadOnlyBasicProperties` as argument.

## Test coverage
Added basic properties to one of the rabbitmq7 test methods as well as assertions.

## Other details
Fixes #6723

PR #6778, opened 2025-03-19
Refactor RabbitMQ test sample

andrewlock:
## Summary of changes

- Refactor RabbitMQ sample to reduce duplication
- Update manual spans to make hierarchy clearer and deterministic in snapshots

## Reason for change

Was trying to debug a failure in #6753 and the snapshots were making it too confusing

## Implementation details

- Created a RabbitMQ "Helper" that wraps the differences in APIs between version
  - The critical thing is that the v7 APIs are all async and the previous version are all sync
  - Previously we duplicated large amounts of code - this creates a helper to encapsulate the differences
  - ✅ Ran the snapshots at this point for all supported versions, and confirmed that they still passed
- Updated the manual spans to make sure each one is unique
- Updated all the snapshots

## Test coverage

The coverage is the same, the samples are just refactored and the snapshots updated.

## Other details

Previously, the top-level spans were all identical, and therefor the sorting in the snapshots was non-deterministic:

image

After this PR, the manual root spans are all unique

image

Prerequisite for 
- #6753

### DIFF
--- a/tracer/test/snapshots/RabbitMQTests.3_x.SchemaV0.verified.txt
+++ b/tracer/test/snapshots/RabbitMQTests.3_x.SchemaV0.verified.txt
@@ -1,4 +1,4 @@
-[
+﻿[
   {
     TraceId: Id_1,
     SpanId: Id_2,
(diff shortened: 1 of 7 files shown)


lucaspimentel (approved):

bouwkast (approved):
Thank you!

PR #6784, opened 2025-03-20
Update dependabot to scan composite actions

TonyCTHsu:
## Summary of changes

Update dependabot to scan composite actions


andrewlock (approved):
Thanks!

PR #6879, opened 2025-04-22
Add "all-green" job to enfore green CI policy

cbeauchesne:
## Summary of changes

* [ ] add a GHA job that check xxx workflow
* [ ] check on azure
* [ ] check on gitlab
* [ ] skipped checks :


andrewlock:
Before you go to far with this @cbeauchesne, I'd suggest that this is currently 100% not feasible to enforce on .NET. We have a 0% pass rate, and adding this will purely add frustration with little to no real benefit.

We definetely _should_ do this at some point, but IMO now is not the right time...

cbeauchesne:
Hi Andrew! 

Don't worry, I really don't want to kill the guild's agility.

The idea is that if we wait for all jobs/stages to be healthy to enable green CI policy, we'll probably never achieved it, because some jobs may become unhealthy while we're fixing other. So the strategy is it list healthy jobs, and make them a requirement (or to be more precise list unhelathy job, and do not include them in required jobs, that's my 4th point in the PR description : _skipped checks :_). 

And as soon as a job become healthy, we remove it from the exclusion list.

Right now, the unhealthy jobs (ratio of PR are merged with those jobs being in success < 90%) are : 

*    89% (131) system_tests
*    81% (129) dd-gitlab/default-pipeline
*    79% (131) benchmarks
*    77% (26) throughput_profiler
*    73% (131) profiler_execution_benchmarks
*    14% (131) throughput
*    1% (125) consolidated-pipeline
*    0% (2) devflow/merge
*    0% (2) dd-gitlab/user-triggered-pipeline
*    0% (1) dd-gitlab/notify-pipeline-failure
*    0% (1) dd-gitlab/promote-oci-to-prod
*    0% (1) dd-gitlab/publish-lib-init-ghcr-tags

cbeauchesne:
Adding a all-green jobs in azure is not trivial, I'll think of a universal way of tackling this  : the difference between GHA and azure is that azure use a legacy API to report status, and this API is not covered by `wechuli/allcheckspassed`. Turns out that gitlab also use this legacy API, so it's better to have a way to adress this.

Right now, GHA jobs are covered (and all of them are covered), it's a first step.

andrewlock:
> Adding a all-green jobs in azure is not trivial, I'll think of a universal way of tackling this : the difference between GHA and azure is that azure use a legacy API to report status, and this API is not covered by wechuli/allcheckspassed. Turns out that gitlab also use this legacy API, so it's better to have a way to adress this.

FWIW, I don't even think it's necessary. If/when we _want_ to enforce this, we can add a step at the end of the azure pipeline and mark it as required in GitHub. We have already marked the stages we care about the most as required anyway. 

Same goes for gitlab - if/when we want to, we can mark the `dd-gitlab/default-pipeline` step as required I believe. But we need to do some major rearchitecting of the gitlab integration before we can do that.

> Right now, GHA jobs are covered (and all of them are covered), it's a first step.

👍 Most of these are marked required already anyway, so this sounds fine to me

.github/workflows/all-green.yml
```diff
@@ -0,0 +1,28 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      checks: read
+      contents: read
+    steps:
+      - uses: wechuli/allcheckspassed@e22f45a4f25f4cf821d1273705ac233355400db1
+        with:
+          delay: 10  # wait for this delay before starting
+          polling_interval: 1 # after a failure, how long do it wait before checking again
+          retries: 60  # how many retries before stopping
```
andrewlock, line 23: What happens if this fails because of flake in the github actions. If we come back 2 hours later, retry the failed action and it passes, will this then pass?

andrewlock (approved):

cbeauchesne:
> we can add a step at the end of the azure pipeline

 I tried this, and I didn't succeed, as I didn't find a way to run a stage at the end of the pipeline WITHOUT having to list all jobs (a key point of this policy is not to list all jobs, as it would open a door to one day forget some of them). But I may miss some knowledge on azure, do you know if it's possible ?

PR #6931, opened 2025-05-07
Use DataDog/ensure-ci-success instead of wechuli/allcheckspassed

cbeauchesne:
## Summary of changes

Use DataDog/ensure-ci-success instead of wechuli/allcheckspassed for `all-green` job

## Reason for change

This action checks the commit statuses API, so it get results from azure

### DIFF
--- a/.github/workflows/all-green.yml
+++ b/.github/workflows/all-green.yml
@@ -33,7 +33,6 @@ jobs:
             profiler_execution_benchmarks
             dd-gitlab/download-serverless-artifacts
             dd-gitlab/download-single-step-artifacts
-            integration_tests_arm64_debugger
 
 # Why some checks are excluded? Success ratio are too low :
 # - consolidated-pipeline : 10%
--- a/.github/workflows/all-green.yml
+++ b/.github/workflows/all-green.yml
@@ -1,6 +1,10 @@
 name: All Green
 on:
   pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
   push:
     branches:
       - master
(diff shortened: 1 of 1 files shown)


.github/workflows/all-green.yml
```diff
@@ -14,15 +18,34 @@ jobs:
...
+      - uses: DataDog/ensure-ci-success@f40e6ffd8e60280d478b9b92209aaa30d3d56895
         with:
-          delay: 10  # wait for this delay before starting
-          polling_interval: 1 # after a failure, how long do it wait before checking again
-          retries: 60  # how many retries before stopping 
-          # checks_exclude: ''  # comma separated list of checks to exclude, nothing right now
+          initial-delay-seconds: 10  # wait for this delay before starting
+          max-retries: 60  # how many retries before stopping 
+          ignored-name-patterns: |
+            consolidated-pipeline
```
andrewlock, line 28: Ironically, this is an overall success status for the whole azure devops pipeline. So we could simply require that 😅

.github/workflows/all-green.yml
```diff
@@ -14,15 +18,34 @@ jobs:
...
+            throughput
+            dd-gitlab/run-benchmarks
+            .github/dependabot.yml
+            dd-gitlab/default-pipeline
```
andrewlock, line 32: Similarly, this is a success status for the whole shared pipeline I believe

.github/workflows/all-green.yml
```diff
@@ -14,15 +18,34 @@ jobs:
...
+            benchmarks
+            profiler_execution_benchmarks
+            dd-gitlab/download-serverless-artifacts
+            dd-gitlab/download-single-step-artifacts
+            throughput_profiler
```
andrewlock, line 37: Throughput have been deleted.

```suggestion
            dd-gitlab/run-benchmarks
            .github/dependabot.yml
            dd-gitlab/default-pipeline
            benchmarks
            profiler_execution_benchmarks
            dd-gitlab/download-serverless-artifacts
            dd-gitlab/download-single-step-artifacts
```

.github/workflows/all-green.yml
```diff
@@ -14,15 +18,34 @@ jobs:
...
+            integration_tests_arm64_debugger
```
andrewlock, line 38: This one almost certainly _shouldn't_ be skipped. 
```suggestion
```

andrewlock (approved):
I wonder, what happens if there's a build configuration issue 🤔  No checks run, so no _failed_ statuses are reported, and this... passes? 🤔

.github/workflows/all-green.yml
cbeauchesne: One pain point would be flaky management : if at some point one of your job is flaky, and it takes more than few hours to fix, for PR merge, you need to : 

* remove the requirement for all gitlab jobs
* or harass repo admin to merge PRs

With this, you can temporary add the failing job in this list, and keep your green CI policy for all other jobs.

----

One another pitfall is new workflows. On Github PR engine, jobs are optional by default (what a shame), unless you add them in required list.
With that solution, it changes this logic : all jobs are required by default, **unless** you add it in the exclusion list.

cbeauchesne:
> I wonder, what happens if there's a build configuration issue 🤔 No checks run, so no failed statuses are reported, and this... passes? 🤔

Could you precise this? What is `build configuration` ? it's a job in azure ?

andrewlock:
> > I wonder, what happens if there's a build configuration issue 🤔 No checks run, so no failed statuses are reported, and this... passes? 🤔
> 
> Could you precise this? What is `build configuration` ? it's a job in azure ?

I mean, if you screw up the yaml for example, so that no jobs run. The way GitHub works with "required checks" means if those jobs aren't reported, it fails. How does this all-green work? If it doesn't know what's expected to run, does it know when something doesn't run but should have?

cbeauchesne:
Ok, I understand now : your point is about CI completeness.

The current job mission is **not** to ensure CI completeness. If you want to ensure that nobody screws up yaml files in a way that some job are not executed, you must think about the possibility of screwing any job, removing it or making it a success whatever happen, including this one :). It's something we'll need to tackle at some point, but it can't be solved (I mean in a reliable way) by some side job in a PR.

Playing with `needs` parameter (or the equivalent in azure), and required jobs in github is a good start. We can also think of a new parameter for this action, called `required-jobs`, but it'll be actually the same as github options, but prone to human errors. And a rock solid solution is a monitoring of what has been executed during PR/main/releases. But it's a whole project by itself.

The present mission of this job is actually very limited, and can be expressed by : _if a job runs on a PR, it must be green to a allows the PR to be merged_. Nothing more :)

PR #6948, opened 2025-05-09
Fix arm64 SQL server not found flaky tests

NachoEchevarria:
## Summary of changes

In some alpine, arm64 tests, some SqlServer tests are failing because the server is not found.

This PR adds the skip (13) return code to the samples, so we can skip the test and log it.


andrewlock (approved):
LGTM, thanks!

bouwkast (approved):
Great thanks!

NachoEchevarria:
Thanks for your reviews!

PR #6983, opened 2025-05-16
Add `net8.0` and `net9.0` to trimming nuspec

bouwkast:
## Summary of changes

Adds `net8.0` and `net9.0` to trimming `nuspec` file.

## Reason for change

Seems strange to have the others declared and was asked if we support those later ones which we do.

## Other details

I'm actually unsure why we have a `nuspec` for this

### DIFF
--- a/tracer/src/Datadog.Trace.Trimming/Datadog.Trace.Trimming.nuspec
+++ b/tracer/src/Datadog.Trace.Trimming/Datadog.Trace.Trimming.nuspec
@@ -19,6 +19,8 @@
     <tags>Datadog APM tracing profiling instrumentation trimming</tags>
     <repository type="git" url="https://github.com/DataDog/dd-trace-dotnet.git" />
     <dependencies>
+      <group targetFramework="net9.0" />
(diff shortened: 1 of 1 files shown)


andrewlock (approved):
Huh. Interesting. This definitely looks like something we need to keep an eye out for in the future... I wonder if it just means trimming wasn't working for .NET 8/9 🤔  but we have tests for that so.... 🤷‍♂️

tracer/src/Datadog.Trace.Trimming/Datadog.Trace.Trimming.nuspec
```diff
@@ -19,6 +19,8 @@
     <tags>Datadog APM tracing profiling instrumentation trimming</tags>
     <repository type="git" url="https://github.com/DataDog/dd-trace-dotnet.git" />
     <dependencies>
```
lucaspimentel, line 21: We probably don't need the `<dependencies>` element at all, since we have no dependencies (they are all empty).

PR #6995, opened 2025-05-21
[dd-trace] Improved GAC commands

tonyredondo:
## Summary of changes

This PR improves the support in the GAC command, by allowing us to install, uninstall and get info for multiple versions of an assembly.

## Reason for change

The current support only checks if an assembly is installed in the GAC no matter version, causing an issue when a previous version of dd-trace installed an old version of Datadog.Trace in the GAC meaning that we bailout and don't install the new version, causing a crash when the application loads.

image

## Implementation details

Now we include more unmanaged Fusion apis to list the version of the assemblies installed in the GAC.

## Test coverage

Added a new `GacWithMultipleVersions` test.

PR #7003, opened 2025-05-23
Fix exit code from `GacUninstallCommand` when the assembly is not in the GAC

andrewlock:
## Summary of changes

Fixes the return code for the `GacUninstallCommand` when the assembly is not in the GAC

## Reason for change

If the assembly isn't in the GAC, we shouldn't return error when uninstalling it, because the desired state is correct. This was changed in #6995 but is an unintended behaviour change

## Implementation details

- Change the exit code `1` -> `0`
- Write a success message instead of a warning
- Fix typo 

## Test coverage

🙈 

## Other details

Currently some of the other "failure" messages are _effectively_ the same thing, so it might make change to change those too. However, their behaviour wasn't changed in #6995 so is non-critical


tonyredondo (approved):
Thank you!

PR #7014, opened 2025-05-27
[Tracer] Service discovery integration

anna-git:
## Summary of changes

Process or service discovery dotnet implementation using libdatadog
Will only work on linux as it's creating an memfd behind the scenes.

## Reason for change
https://docs.google.com/document/d/1WCOAcu0S-pQ4lt8jVtkkOKJt-pUJNdI8V60lARXo77Y/edit?tab=t.0#heading=h.eq6l6vmm63m2

## Implementation details

Calling libdatadog via interop to create the memfd.

## Test coverage
System test is passing in CI

### DIFF
--- a/tracer/src/Datadog.Trace/LibDatadog/NativeStringHelpers.cs
+++ b/tracer/src/Datadog.Trace/LibDatadog/NativeStringHelpers.cs
@@ -2,6 +2,7 @@
 // Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2 License.
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
 // </copyright>
+#nullable enable
 
 using System;
 using System.Text;


tracer/src/Datadog.Trace/Datadog.Trace.csproj
```diff
@@ -119,6 +119,10 @@
     <DefineConstants>$(DefineConstants);ASYNCLOCAL;HASHTABLE;FEATURE_DEFAULT_INTERFACE;FEATURE_SPAN</DefineConstants>
   </PropertyGroup>
 
+  <PropertyGroup Condition=" '$(OS)' == 'Linux' ">
+    <DefineConstants>$(DefineConstants);LINUX</DefineConstants>
+  </PropertyGroup>
```
andrewlock, line 124: You can't do this - we run the same dlls across all platforms, so you'll need to have a runtime check instead

andrewlock:

- [This PR (7014)](https://github.com/DataDog/dd-trace-dotnet/tree/29fcbd9fc41012e7f96b1892a2d2b4c474fe87fd)
- [master](https://github.com/DataDog/dd-trace-dotnet/tree/c05ddfb1191eb57e7281f6eac462b82d76cbf74c)

```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Baseline
    This PR (7014) - mean (68ms)  : 65, 71
     .   : milestone, 68,
    master - mean (69ms)  : 63, 74
     .   : milestone, 69,

    This PR (7014) - mean (1,020ms)  : 1006, 1035
     .   : milestone, 1020,
    master - mean (1,015ms)  : 1001, 1029
     .   : milestone, 1015,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Baseline
    This PR (7014) - mean (103ms)  : 101, 105
     .   : milestone, 103,
    master - mean (102ms)  : 100, 104
     .   : milestone, 102,

    This PR (7014) - mean (699ms)  : 689, 710
     .   : milestone, 699,
    master - mean (699ms)  : 690, 708
     .   : milestone, 699,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Baseline
    This PR (7014) - mean (89ms)  : 87, 92
     .   : milestone, 89,
    master - mean (89ms)  : 87, 91
     .   : milestone, 89,

    This PR (7014) - mean (666ms)  : 654, 679
     .   : milestone, 666,
    master - mean (673ms)  : 655, 691
     .   : milestone, 673,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Baseline
    This PR (7014) - mean (191ms)  : 187, 194
     .   : milestone, 191,
    master - mean (191ms)  : 188, 194
     .   : milestone, 191,

    This PR (7014) - mean (1,168ms)  : 1151, 1185
     .   : milestone, 1168,
    master - mean (1,168ms)  : 1151, 1186
     .   : milestone, 1168,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Baseline
    This PR (7014) - mean (270ms)  : 266, 275
     .   : milestone, 270,
    master - mean (276ms)  : 252, 300
     .   : milestone, 276,

    This PR (7014) - mean (929ms)  : 915, 944
     .   : milestone, 929,
    master - mean (938ms)  : 923, 953
     .   : milestone, 938,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Baseline
    This PR (7014) - mean (263ms)  : 259, 267
     .   : milestone, 263,
    master - mean (265ms)  : 262, 269
     .   : milestone, 265,

    This PR (7014) - mean (925ms)  : 912, 939
     .   : milestone, 925,
    master - mean (930ms)  : 913, 948
     .   : milestone, 930,

```

tracer/build/_build/Projects.cs
```diff
@@ -46,7 +46,8 @@ public static class Projects
         AppSecUnitTests,
         ClrProfilerManagedTests,
         TraceIntegrationTests,
-        TraceTests
+        TraceTests,
+        DdTraceIntegrationTests
```
lucaspimentel, line 50: This list is going to include everything soon except `Datadog.Trace.Tests` (basic unit tests), right? 😅 

(no action required, just a "grumble" comment)

tracer/missing-nullability-files.csv
```diff
@@ -340,6 +340,7 @@ src/Datadog.Trace/Iast/Analyzers/UserStringInterop.cs
 src/Datadog.Trace/Iast/Aspects/DebugAspects.cs
 src/Datadog.Trace/Iast/Dataflow/AspectFilter.cs
 src/Datadog.Trace/Iast/Dataflow/AspectType.cs
+src/Datadog.Trace/LibDatadog/ServiceDiscovery/ResultTag.cs
```
lucaspimentel: Can you add `#nullable enable` to `ResultTag.cs`? We never want to _add_ files to this list, only remove them.

(it's silly, I know, because it's just an `enum`)

tracer/src/Datadog.Trace/LibDatadog/ServiceDiscovery/TracerMemfdHandleResult.cs
```diff
@@ -0,0 +1,59 @@
...
+namespace Datadog.Trace.LibDatadog.ServiceDiscovery;
+
+// Do not use this type in x86 to map with libdatadog
+[StructLayout(LayoutKind.Explicit)]
+internal struct TracerMemfdHandleResult
+{
+    [FieldOffset(0)]
+    public ResultTag Tag;
+
+    // beware that offset 8 is only valid on x64 and would cause a crash if read on x86.
```
lucaspimentel, line 24: > would cause a crash if read on x86.

How are we making sure this crash doesn't happen? Do we have another type for x86?

tracer/src/Datadog.Trace/LibDatadog/ServiceDiscovery/TracerMemfdHandleResult.cs
```diff
@@ -0,0 +1,59 @@
...
+        var errorMessage = Encoding.UTF8.GetString(buffer);
+        NativeInterop.Common.DropError(ref resultErr);
+        return errorMessage;
+    }
+}
+
+[StructLayout(LayoutKind.Sequential)]
+internal struct VecU8
+{
+    public IntPtr Ptr; // const uint8_t*
```
lucaspimentel: [coding style nit-picking] Either use the type name (`IntPtr`/`UIntPtr`) for all fields, or use the C# keywords (`nint`/`nuint`), instead of mixing styles.

I would suggest the keywords (`nint`/`nuint`) since that's what we do for other types (`string`, `int`, etc) and is enforced by StyleCop.

tracer/build/_build/Projects.cs
andrewlock: Wait, _why_ does this include `DdTraceIntegrationTests` 🤔

anna-git: because otherwise we get an error in logs and `CheckBuildLogForErrors` isn't happy :
image
as this happens at startup..

tracer/src/Datadog.Trace/LibDatadog/ServiceDiscovery/TracerMemfdHandleResult.cs
anna-git: we don't have another type as the feature isn't necessary for x86, and it's guarded here: https://github.com/DataDog/dd-trace-dotnet/blob/4f9a15ba7541e2b62187f037e9b6a24a79208013/tracer/src/Datadog.Trace/TracerManager.cs#L693 
```
  if (FrameworkDescription.Instance.OSPlatform == OSPlatformName.Linux && Environment.Is64BitProcess)
            {
                try
                {
                    var result = Utils.StoreTracerMetadata(1, Tracer.RuntimeId, TracerConstants.Language, TracerConstants.ThreePartVersion, Environment.MachineName, tracerSettings.ServiceName, tracerSettings.Environment, tracerSettings.ServiceVersion);
                    ...
            }
```

tracer/src/Datadog.Trace/LibDatadog/ServiceDiscovery/TracerMemfdHandleResult.cs
anna-git: yes fixed in https://github.com/DataDog/dd-trace-dotnet/pull/7014/commits/89d1fc49b4dcf11799a838fd8ca079ac1846788d#diff-92c8490d903af7a243bd16f38944bf7feace807c58d0cec3343348df9b01f7b1R14

tracer/missing-nullability-files.csv
anna-git: fixed in https://github.com/DataDog/dd-trace-dotnet/pull/7014/commits/89d1fc49b4dcf11799a838fd8ca079ac1846788d#diff-093e7ac26ece59c12cdd808c3bde66ca6fd1ae6937c8a270eb310037a05946f1

tracer/build/_build/Projects.cs
anna-git: I could make it a warning, and leave this untested for integration tests?

tracer/src/Datadog.Trace/LibDatadog/ServiceDiscovery/TracerMemfdHandleResult.cs
anna-git: I changed the comment to make it [more visible as well](https://github.com/DataDog/dd-trace-dotnet/pull/7014/files#diff-d0791cbc0ed7a961c2c5091e0cd21f6623da194c3b6b0ccb67a0c73fdc6eeb9bR12) 😅 
image

lucaspimentel: Oh, because we're only supporting it on Linux, and we don't support x86 on Linux. Got it, thanks.

tracer/src/Datadog.Trace/LibDatadog/ResultTag.cs
```diff
@@ -0,0 +1,12 @@
...
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2 License.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
+// </copyright>
+#nullable enable
+namespace Datadog.Trace.LibDatadog;
+
+internal enum ResultTag
+{
+    Ok,
+    Err
```
lucaspimentel: If the implied values here (`Ok:0` and `Err:1`) must be consistent with native code, I recommend specifying them explicitly:
```
    Ok = 0,
    Err = 1
```

Otherwise, we run the risk of, for example, someone adding a value in between these two, and then `Err` will be `2` instead of `1`
```
    Ok,    (0)
    Maybe, (1)
    Err    (2)
```

lucaspimentel: Also maybe add a comment saying "don't change these values because...", maybe even a link to the values in `libdatadog`.

tracer/src/Datadog.Trace/LibDatadog/Error.cs
```diff
@@ -0,0 +1,32 @@
...
+// </copyright>
+
+#nullable enable
+using System.Runtime.InteropServices;
+using System.Text;
+
+namespace Datadog.Trace.LibDatadog;
+
+[StructLayout(LayoutKind.Sequential)]
+internal struct Error
```
lucaspimentel, line 19: If this matches a struct (or equivalent) in `libdatadog`, should we add a comment with a link to that struct?

lucaspimentel: not a blocker for this PR, and not asking for this specific file, but in general. is this something we should be doing for all `libdatadog` structs and method calls (p/invokes)?

tracer/src/Datadog.Trace/LibDatadog/ServiceDiscovery/TracerMemfdHandle.cs
```diff
@@ -0,0 +1,14 @@
...
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2 License.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
+// </copyright>
+#nullable enable
+using System.Runtime.InteropServices;
+
+namespace Datadog.Trace.LibDatadog.ServiceDiscovery;
+
+[StructLayout(LayoutKind.Sequential)]
+internal struct TracerMemfdHandle
```
lucaspimentel: Do we need this type? I think you can replace the one field of type `TracerMemfdHandle` with `int`, no?

tracer/src/Datadog.Trace/LibDatadog/ServiceDiscovery/TracerMemfdHandleResult.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    [FieldOffset(0)]
+    public ResultTag Tag;
+
+    // beware that offset 8 is only valid on x64 and would cause a crash if read on x86.
+    [FieldOffset(8)]
+    public TracerMemfdHandle Ok;
+
+    [FieldOffset(8)]
+    public Error Err;
```
lucaspimentel: Not super important, but I generally prefer to use C# naming conventions in C# types (and abbreviation like `Err` hurt my eyes), so maybe consider calling this field `Error`?
```csharp
public Error Error;
```

lucaspimentel (approved):
LGTM aside from a few non-blocking comments or questions.

tracer/src/Datadog.Trace/LibDatadog/ResultTag.cs
lucaspimentel: Similar to my other comment, can we call this `Error` instead of abbreviating it to `Err`? The names don't need to match the native code, and I'd prefer we follow C# naming conventions in C# code. Not a blocker, though.

anna-git: yes !  fixed in [c41522a](https://github.com/DataDog/dd-trace-dotnet/pull/7014/commits/c41522a736b8ff90aab4c85e2d9a6fbd641dc900)

tracer/src/Datadog.Trace/LibDatadog/ServiceDiscovery/TracerMemfdHandleResult.cs
anna-git: yes ! fixed in [c41522a](https://github.com/DataDog/dd-trace-dotnet/pull/7014/commits/c41522a736b8ff90aab4c85e2d9a6fbd641dc900)

tracer/src/Datadog.Trace/LibDatadog/ServiceDiscovery/TracerMemfdHandle.cs
anna-git: good point we dont need it, got rid of in yes ! fixed in [c41522a](https://github.com/DataDog/dd-trace-dotnet/pull/7014/commits/c41522a736b8ff90aab4c85e2d9a6fbd641dc900)

tracer/src/Datadog.Trace/LibDatadog/Error.cs
anna-git: added in [c41522a](https://github.com/DataDog/dd-trace-dotnet/pull/7014/commits/c41522a736b8ff90aab4c85e2d9a6fbd641dc900)
yes in general it'd be good to add a link to the rust class or mention the c header , it's a good idea, wondering if we should have a linter or something like that to enforce it

tracer/src/Datadog.Trace/LibDatadog/ResultTag.cs
anna-git: yes good points changed in  [c41522a](https://github.com/DataDog/dd-trace-dotnet/pull/7014/commits/c41522a736b8ff90aab4c85e2d9a6fbd641dc900)

tracer/src/Datadog.Trace/LibDatadog/Error.cs
```diff
@@ -0,0 +1,36 @@
...
+        var message = resultErr.ErrorMessage;
+        if (message.Length == 0)
+        {
+            return string.Empty;
+        }
+
+        var buffer = new byte[(int)resultErr.ErrorMessage.Length];
+        Marshal.Copy(message.Ptr, buffer, 0, (int)message.Length);
+
+        var errorMessage = Encoding.UTF8.GetString(buffer);
```
andrewlock: I think we can avoid the allocation here:

```suggestion
        unsafe
        {
#if NETCOREAPP
            var messageBytes = new ReadOnlySpan<byte>(message.Ptr.ToPointer(), (int)resultErr.ErrorMessage.Length);
            var errorMessage = Encoding.UTF8.GetString(messageBytes);
#else
            return Encoding.UTF8.GetString((byte*)message.Ptr.ToPointer(), (int)Length);
#endif
         }
```

tracer/src/Datadog.Trace/LibDatadog/Error.cs
```diff
@@ -0,0 +1,36 @@
...
+        NativeInterop.Common.DropError(ref resultErr);
```
andrewlock: This can throw, so we may want to wrap it in a try-catch 🤔  But _then_ you need to log the error and you don't want to include a logger reference in this type... so I would suggest splitting the Read and the Drop so that you can read (and log) the error, but then _also_ log any errors from the drop? 🤔

tracer/src/Datadog.Trace/TracerManager.cs
```diff
@@ -682,13 +682,29 @@ private static TracerManager CreateInitializedTracer(TracerSettings settings, Tr
...
-        private static void OneTimeSetup()
+        private static void OneTimeSetup(TracerSettings tracerSettings)
         {
             // Register callbacks to make sure we flush the traces before exiting
             LifetimeManager.Instance.AddAsyncShutdownTask(RunShutdownTasksAsync);
 
             // start the heartbeat loop
             _heartbeatTimer = new Timer(HeartbeatCallback, state: null, dueTime: TimeSpan.Zero, period: TimeSpan.FromMinutes(1));
+
+            if (FrameworkDescription.Instance.OSPlatform == OSPlatformName.Linux && Environment.Is64BitProcess && !Util.EnvironmentHelpers.IsServerlessEnvironment())
```
andrewlock, line 694: We should potentially think about how this is rolling out 🤔 For data pipeline, we very explicitly _don't_ want to start using libdatadog everywhere at once, because we don't know for sure what the impact will be. Maybe it's not too bad as it's linux only,  specifically not serverless, and only a single call. It might act as a decent test essentially 🤔

tracer/src/Datadog.Trace/TracerManager.cs
```diff
@@ -682,13 +682,29 @@ private static TracerManager CreateInitializedTracer(TracerSettings settings, Tr
...
+            {
+                try
+                {
+                    var result = Utils.StoreTracerMetadata(1, Tracer.RuntimeId, TracerConstants.Language, TracerConstants.ThreePartVersion, Environment.MachineName, tracerSettings.ServiceName, tracerSettings.Environment, tracerSettings.ServiceVersion);
```
andrewlock, line 698: These values will be incorrect if the customer re-configures the tracer in code. Do we care about that? Is the team that needs this data aware of that?

andrewlock (approved):
LGTM in general, just a couple of comments 👍

tracer/src/Datadog.Trace/TracerManager.cs
anna-git: Yes they are aware of it and accepted it

tracer/src/Datadog.Trace/LibDatadog/NativeStringHelpers.cs
```diff
@@ -0,0 +1,28 @@
+// <copyright file="NativeStringHelpers.cs" company="Datadog">
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2 License.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
+// </copyright>
+
```
andrewlock, line 6:
```suggestion
#nullable enable
```

tracer/src/Datadog.Trace/LibDatadog/Error.cs
anna-git: good point, so I split them and just call the drop in the originl try catch, so that if it throws, it goes in the main catch and still logs in [0a3d791](https://github.com/DataDog/dd-trace-dotnet/pull/7014/commits/0a3d791cfd505a68edd90b2818af16b5018feb2c)

tracer/src/Datadog.Trace/LibDatadog/Error.cs
anna-git: thanks, added a helper with this for future needs in [0a3d791](https://github.com/DataDog/dd-trace-dotnet/pull/7014/commits/0a3d791cfd505a68edd90b2818af16b5018feb2c)

tracer/src/Datadog.Trace/TracerManager.cs
anna-git: Yes.. and only x64.. I hope this isn't a too large panel for a first rollout 🤔

andrewlock: I mean, we explicitly _didn't_ want to roll this out to linux first, given that it will impact everyone using SSI, but I guess that's what we're doing now 😬

PR #7017, opened 2025-05-27
Activate runtime metrics by default on CI

anna-git:
## Summary of changes

Activate runtime metrics by default on CI for net 9 or netfx. Easier to review per commit
- Change the way we dispose, to dispose more things in .net core, except for the `EventListener` itself because [of this runtime issue.
](https://github.com/dotnet/runtime/issues/103480)
- Use scoped tracer wherever possible in CI (most modified files)
- add macro benchmarks scenarios

## Reason for change

to prepare for activation by default of runtime metrics

## Test coverage

Change tests accordingly

### DIFF
--- a/.gitlab/benchmarks/macrobenchmarks.yml
+++ b/.gitlab/benchmarks/macrobenchmarks.yml
@@ -249,16 +249,6 @@ profiler_cpu_timer_create:
     DD_PROFILING_CPU_ENABLED: 1
     COMPlus_EnableDiagnostics: 1
     ENDPOINT: "hello"
-    
-runtime-metrics:
-  extends: .benchmarks
-  variables:
-    NATIVE_PROFILER_PATH: "dd-trace-dotnet/tracer/tracer-home-linux/linux-x64/Datadog.Trace.ClrProfiler.Native.so"
-    TRACER_HOME_PATH: "dd-trace-dotnet/tracer/tracer-home-linux"
-    COR_ENABLE_PROFILING: 1
-    CORECLR_ENABLE_PROFILING: 1
-    DD_RUNTIME_METRICS_ENABLED: 1
-    ENDPOINT: "hello"
 
 .benchmarks-arm64:
   stage: benchmarks
@@ -455,18 +445,6 @@ profiler_cpu_timer_create-arm64:
     COMPlus_EnableDiagnostics: 1
     ENDPOINT: "hello"
 
-runtime-metrics-arm64:
-  tags: ["runner:apm-k8s-arm-metal"]
-  extends: .benchmarks-arm64
-  variables:
-    NATIVE_PROFILER_PATH: "dd-trace-dotnet/tracer/tracer-home-linux-arm64/linux-arm64/Datadog.Trace.ClrProfiler.Native.so"
-    TRACER_HOME_PATH: "dd-trace-dotnet/tracer/tracer-home-linux-arm64"
-    COR_ENABLE_PROFILING: 1
-    CORECLR_ENABLE_PROFILING: 1
-    DD_RUNTIME_METRICS_ENABLED: 1
-    ENDPOINT: "hello"
-
-
 .benchmarks-win:
   stage: benchmarks-win
   needs: ["check_azure_pipeline"]
@@ -630,12 +608,4 @@ profiler_cpu_timer_create-win:
     DD_PROFILING_WALLTIME_ENABLED: 0
     DD_PROFILING_CPU_ENABLED: 1
     COMPlus_EnableDiagnostics: 1
-    ENDPOINT: "hello"
-    
-runtime-metrics-win:
-  extends: .benchmarks-win
-  variables:
-    COR_ENABLE_PROFILING: 1
-    CORECLR_ENABLE_PROFILING: 1
-    DD_RUNTIME_METRICS_ENABLED: 1
     ENDPOINT: "hello"
--- a/.gitlab/benchmarks/macrobenchmarks.yml
+++ b/.gitlab/benchmarks/macrobenchmarks.yml
@@ -25,7 +25,7 @@ check_azure_pipeline:
     name: "artifacts"
     when: always
     paths:
-      - artifacts/
+      - platform/artifacts/
       - build-id.txt
     expire_in: 3 months
   tags: ["arch:amd64"]
@@ -90,11 +90,13 @@ update-bp-infra:
     - git clone --branch dd-trace-dotnet/macro https://gitlab-ci-token:${CI_JOB_TOKEN}@gitlab.ddbuild.io/DataDog/benchmarking-platform platform && cd platform
     - ./steps/setup-sut.sh
     - bp-runner bp-runner.yml --debug
+  after_script:
+    - mv /var/log/datadog/dotnet/* platform/artifacts
   artifacts:
     name: "artifacts"
     when: always
     paths:
-      - artifacts/
+      - platform/artifacts/
     expire_in: 3 months
   variables:
     K6_OPTIONS_WARMUP_RATE: 14000
@@ -123,7 +125,7 @@ baseline:
     COR_ENABLE_PROFILING: 0
     CORECLR_ENABLE_PROFILING: 0
     ENDPOINT: "hello"
-
+    
 calltarget_ngen:
   extends: .benchmarks
   variables:
@@ -247,6 +249,16 @@ profiler_cpu_timer_create:
     DD_PROFILING_CPU_ENABLED: 1
     COMPlus_EnableDiagnostics: 1
     ENDPOINT: "hello"
+    
+runtime-metrics:
+  extends: .benchmarks
+  variables:
+    NATIVE_PROFILER_PATH: "dd-trace-dotnet/tracer/tracer-home-linux/linux-x64/Datadog.Trace.ClrProfiler.Native.so"
+    TRACER_HOME_PATH: "dd-trace-dotnet/tracer/tracer-home-linux"
+    COR_ENABLE_PROFILING: 1
+    CORECLR_ENABLE_PROFILING: 1
+    DD_RUNTIME_METRICS_ENABLED: 1
+    ENDPOINT: "hello"
 
 .benchmarks-arm64:
   stage: benchmarks
@@ -272,11 +284,12 @@ profiler_cpu_timer_create:
     - git clone --branch dd-trace-dotnet/macro https://gitlab-ci-token:${CI_JOB_TOKEN}@gitlab.ddbuild.io/DataDog/benchmarking-platform platform && cd platform
     - ./steps/setup-sut.sh
     - bp-runner bp-runner.arm.yml --debug
+  after_script: !reference [.benchmarks, after_script]
   artifacts:
     name: "artifacts"
     when: always
     paths:
-      - artifacts/
+      - platform/artifacts/
     expire_in: 3 months
   variables:
     K6_OPTIONS_WARMUP_RATE: 14000
@@ -442,6 +455,18 @@ profiler_cpu_timer_create-arm64:
     COMPlus_EnableDiagnostics: 1
     ENDPOINT: "hello"
 
+runtime-metrics-arm64:
+  tags: ["runner:apm-k8s-arm-metal"]
+  extends: .benchmarks-arm64
+  variables:
+    NATIVE_PROFILER_PATH: "dd-trace-dotnet/tracer/tracer-home-linux-arm64/linux-arm64/Datadog.Trace.ClrProfiler.Native.so"
+    TRACER_HOME_PATH: "dd-trace-dotnet/tracer/tracer-home-linux-arm64"
+    COR_ENABLE_PROFILING: 1
+    CORECLR_ENABLE_PROFILING: 1
+    DD_RUNTIME_METRICS_ENABLED: 1
+    ENDPOINT: "hello"
+
+
 .benchmarks-win:
   stage: benchmarks-win
   needs: ["check_azure_pipeline"]
@@ -464,7 +489,7 @@ profiler_cpu_timer_create-arm64:
     name: "artifacts"
     when: always
     paths:
-      - artifacts/
+      - platform/artifacts/
     expire_in: 3 months
   variables:
     # Whether to cleanup ephemeral instances after benchmarks are run
@@ -605,4 +630,12 @@ profiler_cpu_timer_create-win:
     DD_PROFILING_WALLTIME_ENABLED: 0
     DD_PROFILING_CPU_ENABLED: 1
     COMPlus_EnableDiagnostics: 1
+    ENDPOINT: "hello"
+    
+runtime-metrics-win:
+  extends: .benchmarks-win
+  variables:
+    COR_ENABLE_PROFILING: 1
+    CORECLR_ENABLE_PROFILING: 1
+    DD_RUNTIME_METRICS_ENABLED: 1
     ENDPOINT: "hello"
--- a/tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
+++ b/tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
@@ -136,8 +136,19 @@ namespace Datadog.Trace.RuntimeMetrics
                 return;
             }
 
-            Log.Debug("Disposing Runtime Metrics");
-            _timer.Dispose();
+            Log.Debug("Disposing Runtime Metrics timer");
+            using (var manualResetEvent = new ManualResetEvent(false))
+            {
+                if (_timer.Dispose(manualResetEvent))
+                {
+                    if (!manualResetEvent.WaitOne(30000))
+                    {
+                        Log.Warning("Failed to dispose Runtime Metrics timer after 30 seconds");
+                    }
+                }
+            }
+
+            Log.Debug("Disposing other resources for Runtime Metrics");
             AppDomain.CurrentDomain.FirstChanceException -= FirstChanceException;
             // We don't dispose runtime metrics on .NET Core because of https://github.com/dotnet/runtime/issues/103480
 #if NETFRAMEWORK
@@ -260,7 +271,6 @@ namespace Datadog.Trace.RuntimeMetrics
 #if NETCOREAPP
             return new RuntimeEventListener(statsd, delay);
 #elif NETFRAMEWORK
-
             try
             {
                 return new MemoryMappedCounters(statsd);
--- a/tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
+++ b/tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
@@ -59,7 +59,7 @@ namespace Datadog.Trace.RuntimeMetrics
 
         private TimeSpan _previousUserCpu;
         private TimeSpan _previousSystemCpu;
-        private bool _disposed;
+        private int _disposed;
 
         public RuntimeMetricsWriter(IDogStatsd statsd, TimeSpan delay, bool inAzureAppServiceContext)
             : this(statsd, delay, inAzureAppServiceContext, InitializeListenerFunc)
@@ -130,13 +130,13 @@ namespace Datadog.Trace.RuntimeMetrics
 
         public void Dispose()
         {
-            if (_disposed)
+            if (Interlocked.CompareExchange(ref _disposed, 1, 0) == 1)
             {
+                Log.Debug("Disposing Runtime Metrics but it was already disposed");
                 return;
             }
 
             Log.Debug("Disposing Runtime Metrics");
-            _disposed = true;
             _timer.Dispose();
             AppDomain.CurrentDomain.FirstChanceException -= FirstChanceException;
             // We don't dispose runtime metrics on .NET Core because of https://github.com/dotnet/runtime/issues/103480
@@ -149,7 +149,7 @@ namespace Datadog.Trace.RuntimeMetrics
 
         internal void PushEvents()
         {
-            if (_disposed)
+            if (Volatile.Read(ref _disposed) == 1)
             {
                 Log.Debug("Runtime metrics is disposed and can't push new events");
                 return;
--- a/tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
+++ b/tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
@@ -137,14 +137,15 @@ namespace Datadog.Trace.RuntimeMetrics
             }
 
             Log.Debug("Disposing Runtime Metrics timer");
+            // Callbacks can occur after the Dispose() method overload has been called,
+            // because the timer queues callbacks for execution by thread pool threads.
+            // Using the Dispose(WaitHandle) method overload to waits until all callbacks have completed.
+            // ManualResetEventSlim doesn't do well with Timer so need to use ManualResetEvent
             using (var manualResetEvent = new ManualResetEvent(false))
             {
-                if (_timer.Dispose(manualResetEvent))
+                if (_timer.Dispose(manualResetEvent) && !manualResetEvent.WaitOne(5_000))
                 {
-                    if (!manualResetEvent.WaitOne(30000))
-                    {
-                        Log.Warning("Failed to dispose Runtime Metrics timer after 30 seconds");
-                    }
+                    Log.Warning("Failed to dispose Runtime Metrics timer after 5 seconds");
                 }
             }
 
--- a/tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
+++ b/tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
@@ -59,6 +59,7 @@ namespace Datadog.Trace.RuntimeMetrics
 
         private TimeSpan _previousUserCpu;
         private TimeSpan _previousSystemCpu;
+        private bool _disposed;
 
         public RuntimeMetricsWriter(IDogStatsd statsd, TimeSpan delay, bool inAzureAppServiceContext)
             : this(statsd, delay, inAzureAppServiceContext, InitializeListenerFunc)
@@ -129,14 +130,31 @@ namespace Datadog.Trace.RuntimeMetrics
 
         public void Dispose()
         {
-            AppDomain.CurrentDomain.FirstChanceException -= FirstChanceException;
+            if (_disposed)
+            {
+                return;
+            }
+
+            Log.Debug("Disposing Runtime Metrics");
+            _disposed = true;
             _timer.Dispose();
+            AppDomain.CurrentDomain.FirstChanceException -= FirstChanceException;
+            // We don't dispose runtime metrics on .NET Core because of https://github.com/dotnet/runtime/issues/103480
+#if NETFRAMEWORK
             _listener?.Dispose();
+#endif
             _exceptionCounts.Clear();
+            _statsd.Dispose();
         }
 
         internal void PushEvents()
         {
+            if (_disposed)
+            {
+                Log.Debug("Runtime metrics is disposed and can't push new events");
+                return;
+            }
+
             var now = DateTime.UtcNow;
 
             try
--- a/tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
+++ b/tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
@@ -144,7 +144,6 @@ namespace Datadog.Trace.RuntimeMetrics
             _listener?.Dispose();
 #endif
             _exceptionCounts.Clear();
-            _statsd.Dispose();
         }
 
         internal void PushEvents()
--- a/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AgentMalfunctionTests.cs
+++ b/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AgentMalfunctionTests.cs
@@ -69,10 +69,16 @@ namespace Datadog.Trace.ClrProfiler.IntegrationTests
         {
             SkipOn.Platform(SkipOn.PlatformValue.MacOs);
 
+            var canUseStatsD = EnvironmentHelper.CanUseStatsD(transportType);
+            if (!canUseStatsD)
+            {
+                SetEnvironmentVariable(ConfigurationKeys.RuntimeMetricsEnabled, "0");
+            }
+
             EnvironmentHelper.EnableTransport(transportType);
             SetEnvironmentVariable(ConfigurationKeys.TraceDataPipelineEnabled, dataPipelineEnabled.ToString());
 
-            using var agent = EnvironmentHelper.GetMockAgent();
+            using var agent = EnvironmentHelper.GetMockAgent(useStatsD: canUseStatsD);
             var customResponse = behaviour switch
             {
                 AgentBehaviour.Return404 => new MockTracerResponse { StatusCode = 404 },
--- a/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AgentMalfunctionTests.cs
+++ b/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AgentMalfunctionTests.cs
@@ -68,17 +68,10 @@ namespace Datadog.Trace.ClrProfiler.IntegrationTests
         private async Task SubmitsTraces(AgentBehaviour behaviour, TestTransports transportType, string metadataSchemaVersion, bool dataPipelineEnabled)
         {
             SkipOn.Platform(SkipOn.PlatformValue.MacOs);
-
-            var canUseStatsD = EnvironmentHelper.CanUseStatsD(transportType);
-            if (!canUseStatsD)
-            {
-                SetEnvironmentVariable(ConfigurationKeys.RuntimeMetricsEnabled, "0");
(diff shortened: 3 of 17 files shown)


andrewlock:

- [This PR (7017)](https://github.com/DataDog/dd-trace-dotnet/tree/3c64b7ffa92dd402a78e81ead32cfbfe4442686a)
- [master](https://github.com/DataDog/dd-trace-dotnet/tree/c490e2cabb35ad333a93cf41723a32132be47e7d)

```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Bailout
    This PR (7017) - mean (81ms)  : 77, 84
     .   : milestone, 81,
    master - mean (80ms)  : 76, 84
     .   : milestone, 80,

    section Baseline
    This PR (7017) - mean (76ms)  : 71, 81
     .   : milestone, 76,
    master - mean (76ms)  : 71, 82
     .   : milestone, 76,

    This PR (7017) - mean (1,078ms)  : 1051, 1106
     .   : milestone, 1078,
    master - mean (1,078ms)  : 1049, 1108
     .   : milestone, 1078,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Bailout
    This PR (7017) - mean (120ms)  : 115, 125
     .   : milestone, 120,
    master - mean (119ms)  : 115, 123
     .   : milestone, 119,

    section Baseline
    This PR (7017) - mean (119ms)  : 114, 124
     .   : milestone, 119,
    master - mean (118ms)  : 114, 122
     .   : milestone, 118,

    This PR (7017) - mean (753ms)  : 732, 775
     .   : milestone, 753,
    master - mean (753ms)  : 731, 775
     .   : milestone, 753,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Bailout
    This PR (7017) - mean (107ms)  : 102, 111
     .   : milestone, 107,
    master - mean (107ms)  : 102, 112
     .   : milestone, 107,

    section Baseline
    This PR (7017) - mean (106ms)  : 100, 111
     .   : milestone, 106,
    master - mean (105ms)  : 100, 111
     .   : milestone, 105,

    This PR (7017) - mean (711ms)  : 684, 737
     .   : milestone, 711,
    master - mean (719ms)  : 690, 748
     .   : milestone, 719,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Bailout
    This PR (7017) - mean (217ms)  : 211, 223
     .   : milestone, 217,
    master - mean (217ms)  : 209, 224
     .   : milestone, 217,

    section Baseline
    This PR (7017) - mean (214ms)  : 206, 223
     .   : milestone, 214,
    master - mean (212ms)  : 203, 221
     .   : milestone, 212,

    This PR (7017) - mean (1,215ms)  : 1179, 1252
     .   : milestone, 1215,
    master - mean (1,219ms)  : 1188, 1249
     .   : milestone, 1219,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Bailout
    This PR (7017) - mean (306ms)  : 299, 313
     .   : milestone, 306,
    master - mean (308ms)  : 299, 316
     .   : milestone, 308,

    section Baseline
    This PR (7017) - mean (303ms)  : 294, 313
     .   : milestone, 303,
    master - mean (308ms)  : 299, 316
     .   : milestone, 308,

    This PR (7017) - mean (976ms)  : 939, 1013
     .   : milestone, 976,
    master - mean (980ms)  : 951, 1009
     .   : milestone, 980,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Bailout
    This PR (7017) - mean (300ms)  : 292, 309
     .   : milestone, 300,
    master - mean (301ms)  : 290, 311
     .   : milestone, 301,

    section Baseline
    This PR (7017) - mean (299ms)  : 291, 307
     .   : milestone, 299,
    master - mean (299ms)  : 290, 307
     .   : milestone, 299,

    This PR (7017) - mean (964ms)  : 931, 998
     .   : milestone, 964,
    master - mean (970ms)  : 939, 1002
     .   : milestone, 970,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Bailout
    This PR (7017) - mean (298ms)  : 286, 310
     .   : milestone, 298,
    master - mean (300ms)  : 292, 308
     .   : milestone, 300,

    section Baseline
    This PR (7017) - mean (297ms)  : 288, 306
     .   : milestone, 297,
    master - mean (298ms)  : 287, 308
     .   : milestone, 298,

    This PR (7017) - mean (870ms)  : 815, 924
     .   : milestone, 870,
    master - mean (872ms)  : 812, 932
     .   : milestone, 872,

```

tracer/src/Datadog.Trace/LifetimeManager.cs
```diff
@@ -60,7 +60,7 @@ public static LifetimeManager Instance
             }
         }
 
-        public TimeSpan TaskTimeout { get; set; } = TimeSpan.FromSeconds(30);
+        public TimeSpan TaskTimeout { get; set; } = TimeSpan.FromSeconds(60);
```
lucaspimentel: I had to look at the code to know what this is used for 😅 . Do you mind renaming `TaskTimeout` to clarify, maybe `ShutdownTaskTimeout`? It looks like there's only one usage in line 126 below, anyway. (Not sure why it has a public setter)

lucaspimentel: Why are we extending this to a full minute? Are we ok with delaying shutdown that long? I think after some amount of time we should give up instead of just waiting longer. Maybe add a short comment explaining why we need 60 seconds.

(We may have to make this configurable in the future so that in serverless we can default to something _very_ short. Extending the time a AWS Lambda function's execution time has billing implications for users $$$.)

tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
```diff
@@ -129,14 +130,31 @@ internal RuntimeMetricsWriter(IDogStatsd statsd, TimeSpan delay, bool inAzureApp
...
+            if (_disposed)
+            {
+                return;
+            }
+
+            Log.Debug("Disposing Runtime Metrics");
+            _disposed = true;
             _timer.Dispose();
+            AppDomain.CurrentDomain.FirstChanceException -= FirstChanceException;
+            // We don't dispose runtime metrics on .NET Core because of https://github.com/dotnet/runtime/issues/103480
```
lucaspimentel, line 154: Isn't this fixed in .NET 9? Should we add a runtime check (since we can't use `#if NET_9_0_OR_GREATER`) and dispose if .NET version is >= 9?

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AzureFunctionsTests.cs
```diff
@@ -177,7 +177,9 @@ public InProcessRuntimeV4(ITestOutputHelper output)
         [Trait("RunOnWindows", "True")]
         public async Task SubmitsTraces()
         {
-            using var agent = EnvironmentHelper.GetMockAgent(useTelemetry: true);
+            using var agent = EnvironmentHelper.GetMockAgent(useTelemetry: true, useStatsD: true);
+            // in this case and version, azure registers an event listener, as runtime metrics does, we can't have 2 listeners, it will crash
```
lucaspimentel: Does this only crash in tests, or should we forcibly disable runtime metrics when the tracer is running in Azure Functions? (not in this PR, just checking)

anna-git: It technically works in Azure but if Azure registers its logforwarder event listener, which seems to happen in the InProcess test in .net 6, because of the runtime bug, it will provoke a fatal clr crash when application exits, on shutdown.. but in this PR, we don't activate runtime metrics yet, the idea for later is to instrument the EventListener.Dispose method up to net8.0 to bail out before it does anything...

tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
anna-git: from what I understood from Andrew:
> We don't build for .NET 7+, so you'd have to do a runtime check, so prob not worth the hassle IMO

tracer/src/Datadog.Trace/LifetimeManager.cs
anna-git: good to know! I reverted to 30 seconds in [905b103](https://github.com/DataDog/dd-trace-dotnet/pull/7017/commits/905b1036add96b7e2e23d901eb3713ebc00580f4). And after running the CI in debug mode and adding several logs I realized that it's not runtime metrics timeout on .net 3.1 on windows at shutdown but it's [timing out before](https://github.com/DataDog/dd-trace-dotnet/blob/06ef37a42f32ed3862d5b91f1e2ba5751c902ff2/tracer/src/Datadog.Trace/TracerManager.cs#L734) as the other log messages after that aren't printed, so I'll revert for this PR

anna-git: good point fixed in [905b103](https://github.com/DataDog/dd-trace-dotnet/pull/7017/commits/905b1036add96b7e2e23d901eb3713ebc00580f4)

tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
lucaspimentel: Yeah, that's fine 😅

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AzureFunctionsTests.cs
lucaspimentel: Thanks for clarifying.

tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
```diff
@@ -129,14 +130,31 @@ internal RuntimeMetricsWriter(IDogStatsd statsd, TimeSpan delay, bool inAzureApp
 
         public void Dispose()
         {
-            AppDomain.CurrentDomain.FirstChanceException -= FirstChanceException;
+            if (_disposed)
```
andrewlock: Is there an actual risk of double-disposal here? I ask because this check isn't thread safe, so if there _is_ a risk, we probably need a lock or at least an _interlocked_

tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
```diff
@@ -129,14 +130,31 @@ internal RuntimeMetricsWriter(IDogStatsd statsd, TimeSpan delay, bool inAzureApp
...
+#if NETFRAMEWORK
             _listener?.Dispose();
+#endif
             _exceptionCounts.Clear();
+            _statsd.Dispose();
         }
 
         internal void PushEvents()
         {
+            if (_disposed)
```
andrewlock: What happens if `PushEvents()` is called on a disposed instance of `statsd`? I ask because this isn't thread safe, and it could happen

tracer/test/Datadog.Trace.TestHelpers/TestTracer/ScopedTracer.cs
```diff
@@ -0,0 +1,28 @@
...
+    TracerSettings settings = null,
+    IAgentWriter agentWriter = null,
+    ITraceSampler sampler = null,
+    IScopeManager scopeManager = null,
+    IDogStatsd statsd = null,
+    ITelemetryController telemetryController = null,
+    IDiscoveryService discoveryService = null)
+    : Tracer(settings, agentWriter, sampler, scopeManager, statsd, telemetry: telemetryController, discoveryService: discoveryService), IAsyncDisposable
+{
+    public ValueTask DisposeAsync() => new(TracerManager.ShutdownAsync());
```
andrewlock, line 27: I think we should actually call the await here, it's safer 🤔 
```suggestion
    public async ValueTask DisposeAsync() => await TracerManager.ShutdownAsync();
```

tracer/test/Datadog.Trace.TestHelpers/TestTracer/ScopedTracerFixture.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Datadog.Trace.Agent;
+using Datadog.Trace.Configuration;
+using Datadog.Trace.Sampling;
+using Datadog.Trace.Vendors.StatsdClient;
+
+namespace Datadog.Trace.TestHelpers.TestTracer;
+
+public class ScopedTracerFixture : IAsyncDisposable
```
andrewlock: I don't think `ScopedTracerFixture` should exist. Having multiple `Tracer` instances kept alive for a long time seems like a recipe for issues 😟 

I also don't think xunit actually supports `IAsyncDisposable` in v2, which could cause a bunch of issues 😅

tracer/test/Datadog.Trace.IntegrationTests/OriginTagSendTraces.cs
```diff
@@ -8,22 +8,23 @@
 using Datadog.Trace.Agent;
 using Datadog.Trace.Configuration;
 using Datadog.Trace.TestHelpers;
+using Datadog.Trace.TestHelpers.TestTracer;
 using FluentAssertions;
 using Xunit;
 
 namespace Datadog.Trace.IntegrationTests
 {
-    public class OriginTagSendTraces
```
andrewlock: I don't think we should use `IClassFixture<ScopedTracerFixture>` here, we should have one tracer created per test, and it should be disposed at the end. The easiest and safest thing to do would be to make the existing ctor just be a helper method that returns a `Tracer,` make the `_tracer` a variable in the tests, and use `await using` on the returned value like you have elsewhere 🙂

tracer/test/Datadog.Trace.IntegrationTests/SpanTagTests.cs
```diff
@@ -9,24 +9,25 @@
...
 using Datadog.Trace.Sampling;
 using Datadog.Trace.TestHelpers;
+using Datadog.Trace.TestHelpers.TestTracer;
 using FluentAssertions;
 using Xunit;
 
 namespace Datadog.Trace.IntegrationTests
 {
-    public class SpanTagTests
+    public class SpanTagTests : IClassFixture<ScopedTracerFixture>
```
andrewlock: As above, we should remove the fixture

tracer/test/Datadog.Trace.IntegrationTests/Tagging/SamplingPriorityTests_MultipleChunksWithUpstreamService.cs
```diff
@@ -8,27 +8,28 @@
...
 using Datadog.Trace.Configuration;
 using Datadog.Trace.TestHelpers;
+using Datadog.Trace.TestHelpers.TestTracer;
 using FluentAssertions;
 using Xunit;
 
 namespace Datadog.Trace.IntegrationTests.Tagging;
 
-public class SamplingPriorityTests_MultipleChunksWithUpstreamService
+public class SamplingPriorityTests_MultipleChunksWithUpstreamService : IClassFixture<ScopedTracerFixture>
```
andrewlock: go away fixture!

tracer/test/Datadog.Trace.Tests/ContinuousProfiler/ContextTrackerTests.cs
```diff
@@ -29,12 +31,12 @@ public void SetEndpointOnRootWebSpans()
 
             contextTracker.Setup(c => c.SetEndpoint(It.IsAny<ulong>(), It.IsAny<string>()))
                           .Callback<ulong, string>((i, e) => invocations.Add((i, e)));
-
+            ScopedTracer tracer = null;
```
andrewlock: I think you can just use the standard `await using` pattern here 🙂

tracer/test/Datadog.Trace.Tests/PlatformHelpers/AzureAppServicesTests.cs
```diff
@@ -192,47 +194,49 @@ public async Task DoNotTagSpans()
             // AAS Tags are handled at serialization now. So no tags should be set on spans
             var vars = AzureAppServiceHelper.GetRequiredAasConfigurationValues(SubscriptionId, DeploymentId, PlanResourceGroup, SiteResourceGroup);
             var settings = new TracerSettings(vars);
-            await using var tracer = TracerHelper.CreateWithFakeAgent(settings);
-            var spans = new List<ISpan>();
-            var iterations = 5;
-            var remaining = iterations;
-
-            while (remaining-- > 0)
+            await using (var tracer = TracerHelper.CreateWithFakeAgent(settings))
```
andrewlock: nit: we can keep using a declaration here, no? 

```suggestion
            await using var tracer = TracerHelper.CreateWithFakeAgent(settings);	
```

tracer/test/Datadog.Trace.TestHelpers/EnvironmentHelper.cs
```diff
@@ -92,6 +92,8 @@ public static bool IsCoreClr()
             return RuntimeFrameworkDescription.Contains("core") || IsNet5();
         }
 
+        public static bool CanUseStatsD(TestTransports transport) => transport != TestTransports.Uds || !EnvironmentTools.IsWindows();
```
andrewlock: nit: I think this is maybe easier to parse? 🤔 i.e. it's available unless you're using UDS on Windows?

```suggestion
        public static bool CanUseStatsD(TestTransports transport) => !(EnvironmentTools.IsWindows() && transport == TestTransports.Uds);
```

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AgentMalfunctionTests.cs
```diff
@@ -69,10 +69,16 @@ private async Task SubmitsTraces(AgentBehaviour behaviour, TestTransports transp
         {
             SkipOn.Platform(SkipOn.PlatformValue.MacOs);
 
+            var canUseStatsD = EnvironmentHelper.CanUseStatsD(transportType);
+            if (!canUseStatsD)
+            {
+                SetEnvironmentVariable(ConfigurationKeys.RuntimeMetricsEnabled, "0");
+            }
```
andrewlock: I wonder if this should be moved out somewhere else 🤔 e.g. should we set the default value for `useStatsD` based on the TFM, and potentially set the `RuntimeMetricsEnabled` flag based on the variable? 🤔 I'm not sure, WDYT?

.gitlab/benchmarks/macrobenchmarks.yml
```diff
@@ -25,7 +25,7 @@ check_azure_pipeline:
     name: "artifacts"
     when: always
     paths:
-      - artifacts/
+      - platform/artifacts/
```
andrewlock, line 28: I'm surprised this is necessary? It seems like our benchmarks would be broken right now if so? 🤔

.gitlab/benchmarks/macrobenchmarks.yml
```diff
@@ -250,6 +252,16 @@ profiler_cpu_timer_create:
...
+    
+runtime-metrics:
+  extends: .benchmarks
+  variables:
+    NATIVE_PROFILER_PATH: "dd-trace-dotnet/tracer/tracer-home-linux/linux-x64/Datadog.Trace.ClrProfiler.Native.so"
+    TRACER_HOME_PATH: "dd-trace-dotnet/tracer/tracer-home-linux"
+    COR_ENABLE_PROFILING: 1
+    CORECLR_ENABLE_PROFILING: 1
+    DD_RUNTIME_METRICS_ENABLED: 1
+    ENDPOINT: "hello"
```
andrewlock: What's the impact on execution time with this change? We're running a lot of different jobs now (and none of them seem particularly useful for us 😅). I think it's a good idea, just wish we still had throughput tests 🙁

andrewlock (changes requested):
Thanks, just requesting changes for the fixture part which I think could potentially cause issues!

andrewlock:
The big change in execution-time benchmarks is surprising, and I can't explain it 😕 There's nothing in this PR that should have impacted them right?

.gitlab/benchmarks/macrobenchmarks.yml
anna-git: it's necessary to get the log files (before adding this I got an error and now we get the log files as artifacts with it), otherwise I think here we're one folder too high cf https://dd.slack.com/archives/C045F97HYJD/p1751033134170879?thread_ts=1750944398.786819&cid=C045F97HYJD

andrewlock: oh I see, it's an unrelated fix, in which case, thanks! 😅

.gitlab/benchmarks/macrobenchmarks.yml
anna-git: yes :/
do the macro benchmarks dont execute unless manually triggered? so far when I ran CI runs, they didn't unless I explicitely pressed the button on gitlab..   apart from this it didn't seem to take long, but happy to remove as the numbers don't help so much anyway

tracer/test/Datadog.Trace.TestHelpers/TestTracer/ScopedTracer.cs
anna-git: sure! for my understanding how is it safer? is it safer because of the implicit conversion?

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AgentMalfunctionTests.cs
anna-git: yes! it's not needed actually as it's done further down the line in https://github.com/DataDog/dd-trace-dotnet/blob/ceff518b394d035e3c5c576c555b1f76ff7d1c58/tracer/test/Datadog.Trace.TestHelpers.AutoInstrumentation/TestHelper.cs#L145

tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
anna-git: I'm not sure if it could be double disposed, I added a `Volatile` and Interlocked around the variable in 20fd8a909d4c23b073ff67a65a4c0e8165ddf13a

tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
anna-git: good point thanks, added in 20fd8a909d4c23b073ff67a65a4c0e8165ddf13a

tracer/test/Datadog.Trace.TestHelpers/TestTracer/ScopedTracer.cs
anna-git: I changed it in b182bd3e04ec65979a59948e5e51b2316f9dcfab

tracer/test/Datadog.Trace.TestHelpers/EnvironmentHelper.cs
anna-git: I wrote this at first and then got a suggestion to write an or, changed in 357a71d34cd63f28b7fb317d2046424532942d64

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AgentMalfunctionTests.cs
anna-git: removed in 56954850ea8d79f4113cb3cf437f239f1c994fed

tracer/test/Datadog.Trace.IntegrationTests/OriginTagSendTraces.cs
anna-git: Removed it in b182bd3e04ec65979a59948e5e51b2316f9dcfab ! I had to realize that before, every test was reinstantiating the class of course.. so thanks I changed it back to a tracer instance per class 🤦

anna-git:
> The big change in execution-time benchmarks is surprising, and I can't explain it 😕 There's nothing in this PR that should have impacted them right?

I'm not sure why... I havent changed the StringAspects in this PR 😅

tracer/test/Datadog.Trace.TestHelpers/TestTracer/ScopedTracer.cs
andrewlock: > how is it safer

My reason about "safer" was entirely due to it being a `ValueTask`. The thing with `ValueTask` is that there are extra semantics like "if you call it twice, it's an error". But that still doesn't make sense, because we're returning a `ValueTask` either way, so yeah, I think that was "friday brain" taking over 😂 So yeah, ignore me 🙈

.gitlab/benchmarks/macrobenchmarks.yml
andrewlock: The macro benchmarks run on master or when manually triggered

tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
```diff
@@ -129,14 +130,31 @@ internal RuntimeMetricsWriter(IDogStatsd statsd, TimeSpan delay, bool inAzureApp
...
+
+            Log.Debug("Disposing Runtime Metrics");
             _timer.Dispose();
+            AppDomain.CurrentDomain.FirstChanceException -= FirstChanceException;
+            // We don't dispose runtime metrics on .NET Core because of https://github.com/dotnet/runtime/issues/103480
+#if NETFRAMEWORK
             _listener?.Dispose();
+#endif
             _exceptionCounts.Clear();
+            _statsd.Dispose();
```
andrewlock: I'm not sure that this is safe to call unfortunately 😟 The `IDogStatsd` instance is used in _multiple_ components, and disposing it here could cause failures in other components... it is probably better to move responsibility for disposing the `IDogStatsd` instance _out_ of `RuntimeMetricsWriter` entirely... That way you can ensure all the dependencies have been disposed _before_ disposing the `IDogStatsd` instance.

tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
```diff
@@ -129,14 +130,31 @@ internal RuntimeMetricsWriter(IDogStatsd statsd, TimeSpan delay, bool inAzureApp
...
         }
 
         internal void PushEvents()
         {
+            if (Volatile.Read(ref _disposed) == 1)
+            {
+                Log.Debug("Runtime metrics is disposed and can't push new events");
+                return;
+            }
```
andrewlock, line 167: FYI, this doesn't make this thread-safe with the dispose, so there's still a race condition there 😄 Really the only way around it is a lock, but if we move the `IDogStatsd` instance disposal out of this type, then it's probably not a problem anyway 🙂

tracer/test/Datadog.Trace.IntegrationTests/SpanTagTests.cs
```diff
@@ -26,182 +28,193 @@ public SpanTagTests()
             var matchAllRule = "[{\"service\":\"*\", \"name\":\"*\", \"sample_rate\":1.0, \"max_per_second\":1000.0}]";
             var settings = TracerSettings.Create(new() { { ConfigurationKeys.SpanSamplingRules, matchAllRule } });
             _writer = new AgentWriter(_testApi, statsAggregator: null, statsd: null);
-            _tracer = new Tracer(settings, _writer, sampler: null, scopeManager: null, statsd: null);
+            _tracer = ScopedTracerHelper.BuildScopedTracer(settings, _writer, sampler: null, scopeManager: null, statsd: null);
```
andrewlock: nit: this would all be cleaner if we stopped having "field" tracer instances and used local variables instead, so you can do `await using var tracer = ScopedTracerHelper.BuildScopedTracer()` like you do elsewhere 😄

andrewlock: In fact, I just extracted all your `ScopedTracer` work into a separate PR here #7220 in the hopes we can get that merged sooner rather than later 🙂

andrewlock:
I extracted all the scoped tracer changes into a separate PR, to make this one more focused, and rebased 🙂  https://github.com/DataDog/dd-trace-dotnet/pull/7220

.gitlab/benchmarks/macrobenchmarks.yml
anna-git: So on the one hand q3 okr being to enable by default runtime metrics  for .net seem to make it necessary, on the other hand... numbers aren't really meaningful, so happy to remove it if you want :)

tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
anna-git: good point didn't realize it was shared across other components!!! moved the dispose in main shutdown task in [a0900c3](https://github.com/DataDog/dd-trace-dotnet/pull/7017/commits/a0900c335b65876260ddd76b1c6eab12dd5287e4)

.gitlab/benchmarks/macrobenchmarks.yml
anna-git: going to run them and see how long they take actually

anna-git: so arm64, x64, windows take around 30/40 min each to run 😬

tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
anna-git: hopefully like you said [a0900c3](https://github.com/DataDog/dd-trace-dotnet/pull/7017/commits/a0900c335b65876260ddd76b1c6eab12dd5287e4) is enough 😅

.gitlab/benchmarks/macrobenchmarks.yml
anna-git: given the numbers dont help and the time they can take in CI on master, removed in [7206df1](https://github.com/DataDog/dd-trace-dotnet/pull/7017/commits/7206df1db34ac7b8b314d63efc675fd74d7f154d)

andrewlock (approved):
LGTM, thanks!

tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
```diff
@@ -129,14 +130,41 @@ internal RuntimeMetricsWriter(IDogStatsd statsd, TimeSpan delay, bool inAzureApp
...
         public void Dispose()
         {
+            if (Interlocked.CompareExchange(ref _disposed, 1, 0) == 1)
+            {
+                Log.Debug("Disposing Runtime Metrics but it was already disposed");
+                return;
+            }
+
+            Log.Debug("Disposing Runtime Metrics timer");
+            using (var manualResetEvent = new ManualResetEvent(false))
```
andrewlock, line 144: nit: we probably should tell people why 😅 
```suggestion
            // Callbacks can occur after the Dispose() method overload has been called, 
            // because the timer queues callbacks for execution by thread pool threads. 
            // Using the Dispose(WaitHandle) method overload to waits until all callbacks have completed.
            using (var manualResetEvent = new ManualResetEvent(false))
```

tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
```diff
@@ -129,14 +130,41 @@ internal RuntimeMetricsWriter(IDogStatsd statsd, TimeSpan delay, bool inAzureApp
...
+            {
+                if (_timer.Dispose(manualResetEvent))
+                {
+                    if (!manualResetEvent.WaitOne(30000))
```
tonyredondo: I think 30s it's too much, 5s it's enough to push the events.

tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
```diff
@@ -129,14 +130,41 @@ internal RuntimeMetricsWriter(IDogStatsd statsd, TimeSpan delay, bool inAzureApp
...
+                    {
+                        Log.Warning("Failed to dispose Runtime Metrics timer after 30 seconds");
+                    }
+                }
```
andrewlock: nit: 30s is too long - we'll kill the process rather than wait that long, so I think 5s should be plenty... theoretically it's only the time it takes for a scheduled `PushEvents` loop to run one final time on the thread pool I think.. Also tiny nit for reducing nesting

```suggestion
                if (_timer.Dispose(manualResetEvent) &&!manualResetEvent.WaitOne(5_000))
                {
                    Log.Warning("Failed to dispose Runtime Metrics timer after 5 seconds");
                }
```

tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
anna-git: changed to 5s thanks!

PR #7019, opened 2025-05-27
Fix a `NullReferenceException` in AWS SDK `RuntimePipelineInvokeAsyncIntegration`

bouwkast:
## Summary of changes

Fixes a `NullReferenceException` in `RuntimePipelineInvokeAsyncIntegration` for the AWS SDK instrumentation. May be related to V4 as I don't see prior errors from it.

## Reason for change

Error Tracking alerted on our dogfood test application

```
System.NullReferenceException
   at Datadog.Trace.Util.UriHelpers.Combine(String baseUri, String relativePath)
   at Datadog.Trace.ClrProfiler.AutoInstrumentation.AWS.SDK.RuntimePipelineInvokeAsyncIntegration.OnAsyncMethodEnd[TTarget,TResponse](TTarget instance, TResponse response, Exception exception, CallTargetState& state)
   at REDACTED
   at Datadog.Trace.ClrProfiler.CallTarget.Handlers.Continuations.TaskContinuationGenerator`4.SyncCallbackHandler.ContinuationAction(Task`1 previousTask, TTarget target, CallTargetState state)
```

## Implementation details

Just adding a check for when `absolutePath is not null` to call prior to `UriHelpers.Combine`.

## Test coverageg

😬

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/SDK/RuntimePipelineInvokeAsyncIntegration.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/SDK/RuntimePipelineInvokeAsyncIntegration.cs
@@ -81,9 +81,8 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.AWS.SDK
                         var path = request.ResourcePath switch
                         {
                             null => absolutePath,
-                            string resourcePath when absolutePath == "/" => resourcePath,
-                            string resourcePath when absolutePath is not null => UriHelpers.Combine(absolutePath, resourcePath),
-                            string resourcePath => resourcePath, // absolutePath is null, use resourcePath directly
+                            string resourcePath when absolutePath is null or "/" => resourcePath,
+                            string resourcePath => UriHelpers.Combine(absolutePath, resourcePath),
                         };
 
                         // The request object is populated later by the Marshaller,
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/SDK/RuntimePipelineInvokeAsyncIntegration.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/SDK/RuntimePipelineInvokeAsyncIntegration.cs
@@ -82,7 +82,8 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.AWS.SDK
                         {
                             null => absolutePath,
                             string resourcePath when absolutePath == "/" => resourcePath,
-                            string resourcePath => UriHelpers.Combine(absolutePath, resourcePath),
+                            string resourcePath when absolutePath is not null => UriHelpers.Combine(absolutePath, resourcePath),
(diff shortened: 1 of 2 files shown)


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/SDK/RuntimePipelineInvokeAsyncIntegration.cs
```diff
@@ -82,7 +82,8 @@ internal static TResponse OnAsyncMethodEnd<TTarget, TResponse>(TTarget instance,
                         {
                             null => absolutePath,
                             string resourcePath when absolutePath == "/" => resourcePath,
-                            string resourcePath => UriHelpers.Combine(absolutePath, resourcePath),
+                            string resourcePath when absolutePath is not null => UriHelpers.Combine(absolutePath, resourcePath),
+                            string resourcePath => resourcePath, // absolutePath is null, use resourcePath directly
```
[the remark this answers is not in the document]
bouwkast: > Applying suggestions on deleted lines is currently not supported.

Don't think I've seen this before, but yeah I'll merge them and I think it must be something from V4 as well as there weren't any changes on our end

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/SDK/RuntimePipelineInvokeAsyncIntegration.cs
```diff
@@ -81,7 +81,7 @@ internal static TResponse OnAsyncMethodEnd<TTarget, TResponse>(TTarget instance,
                         var path = request.ResourcePath switch
                         {
                             null => absolutePath,
-                            string resourcePath when absolutePath == "/" => resourcePath,
+                            string resourcePath when absolutePath is null or "/" => resourcePath,
                             string resourcePath => UriHelpers.Combine(absolutePath, resourcePath),
                         };
```
lucaspimentel, line 86: Not directly related to your changes, so feel free to ignore, I find this particular usage of a `switch` expression hard to read. It _starts_ like a `switch` over different values of `request.ResourcePath`, but it ends up checking both `request.ResourcePath` and `absolutePath` and returns either one or the other, or a combination of both.

I think this is much easier to follow:
```csharp
var absolutePath = uri?.AbsolutePath;
var resourcePath = request.ResourcePath;
string path;

if (resourcePath is null)
{
    // resourcePath not available
    path = absolutePath; 
}
else if (absolutePath is null or "/")
{
    // absolutePath not available
    path = resourcePath;
}
else
{
    // both are available, combine them
    path = UriHelpers.Combine(absolutePath, resourcePath);
}
```

lucaspimentel (approved):

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/SDK/RuntimePipelineInvokeAsyncIntegration.cs
lucaspimentel: Using `switch` here feels like something Resharper/Rider suggested but we should've ignored 😅

PR #7020, opened 2025-05-27
feat(baggage): adding baggage to span tags

rachelyangdog:
## Summary of changes

This PR introduces automatic propagation of baggage key-value pairs as span tags, prefixed with baggage. The new configuration option `DD_TRACE_BAGGAGE_TAG_KEYS` allows customization of this feature.

## Reason for change

See the related [RFC](https://docs.google.com/document/d/1ZsgAcQHKjaBxDmhHup0zqFz07WYiOrbBksHPoXJe6u4/edit?tab=t.0#heading=h.q69qzmye8dfb) for more details.


tracer/src/Datadog.Trace/Propagators/SpanContextPropagator.cs
```diff
@@ -419,5 +419,42 @@ public void Set(TCarrier carrier, string key, string value)
...
+                // add all baggage items as tags
+                foreach (var item in baggage)
+                {
+                    span.SetTag(item.Key, item.Value);
+                }
+                return;
+            }
+
+            // add only specified baggage items as tags
+            var allowedKeys = baggageTagKeys.Split(',', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);
```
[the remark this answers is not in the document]
lucaspimentel: I looked into this, and even if we used `baggageTagKeys.SplitIntoSpans()`, we would only save on the allocation for the array itself, but we would still have to force the allocation of the individual strings to use `baggage.TryGetValue()` and especially `span.SetTag()`, so never mind 😅.

tracer/src/Datadog.Trace/PlatformHelpers/AspNetCoreHttpRequestHandler.cs
```diff
@@ -125,6 +140,7 @@ public Scope StartAspNetCorePipelineScope(Tracer tracer, Security security, Http
             var scope = tracer.StartActiveInternal(_requestInOperationName, extractedContext.SpanContext, tags: tags, links: extractedContext.Links);
             scope.Span.DecorateWebServerSpan(resourceName, httpMethod, host, url, userAgent, tags);
             AddHeaderTagsToSpan(scope.Span, request, tracer);
+            AddBaggageTagsToSpan(scope.Span, extractedContext.Baggage, tracer);
```
lucaspimentel, line 143: You'll want to do this for other web server spans as well.

tracer/src/Datadog.Trace/Propagators/SpanContextPropagator.cs
```diff
@@ -419,5 +419,42 @@ public void Set(TCarrier carrier, string key, string value)
...
+            foreach (var key in allowedKeys)
+            {
+                if (baggage.TryGetValue(key, out var value))
+                {
+                    span.SetTag(key, value);
+                }
+            }
```
lucaspimentel, line 457: `StringSplitOptions.TrimEntries` was added in .NET 5, so this code won't compile for earlier versions of .NET. Instead, you can remove `StringSplitOptions.TrimEntries` here and call `Trim()` manually below:

```suggestion
            var allowedKeys = baggageTagKeys.Split(',', StringSplitOptions.RemoveEmptyEntries);

            foreach (var key in allowedKeys)
            {
                var trimmedKey = key.Trim();

                if (baggage.TryGetValue(trimmedKey, out var value))
                {
                    span.SetTag(trimmedKey, value);
                }
            }
```

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -590,6 +590,10 @@ _ when x.ToBoolean() is { } boolean => boolean,
                                  .WithKeys(ConfigurationKeys.BaggageMaximumBytes)
                                  .AsInt32(defaultValue: W3CBaggagePropagator.DefaultMaximumBaggageBytes);
 
+            BaggageTagKeys = config
+                            .WithKeys(ConfigurationKeys.BaggageTagKeys)
+                            .AsString(defaultValue: "user.id,session.id,account.id");
```
lucaspimentel: Try making `BaggageTagKeys` a `string[]` and splitting here so we only do it once and not when we try to apply this setting in each incoming http request. See `disabledIntegrationNames` around line 331 above as an example. Probably just need to add `Split()`, like:
```suggestion
            BaggageTagKeys = config
                            .WithKeys(ConfigurationKeys.BaggageTagKeys)
                            .AsString(defaultValue: "user.id,session.id,account.id")
                            .Split([','], StringSplitOptions.RemoveEmptyEntries);
```

You'll have to change the property in line 1067 below, too:
```diff
- internal string BaggageTagKeys { get; }
+ internal string[] BaggageTagKeys { get; }
```

tracer/src/Datadog.Trace/Propagators/SpanContextPropagator.cs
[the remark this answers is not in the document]
lucaspimentel: If we change `settings.BaggageTagKeys` to a `string[]` and do the split when we read the config (see my other comment in `TracerSettings.cs`), then we don't need to split here anymore.

You can remove this line and change the `foreach` below to something like:
```
foreach (var key in baggageTagKeys)
```

tracer/test/Datadog.Trace.Tests/Telemetry/config_norm_rules.json
```diff
@@ -676,6 +676,7 @@
   "DD_TRACE_DATA_PIPELINE_ENABLED": "trace_data_pipeline_enabled",
   "DD_TRACE_INJECT_CONTEXT_INTO_STORED_PROCEDURES_ENABLED": "trace_inject_context_into_stored_procedures_enabled",
   "DD_TRACE_GRAPHQL_ERROR_EXTENSIONS": "trace_graphql_error_extensions",
+  "DD_TRACE_BAGGAGE_TAG_KEYS": "trace_baggage_tag_keys"
```
zacharycmontoya: Looks like you're missing a comma, which is causing the corresponding unit test to fail

```suggestion
  "DD_TRACE_BAGGAGE_TAG_KEYS": "trace_baggage_tag_keys",
```

rachelyangdog: Thank you!!

tracer/src/Datadog.Trace/Propagators/SpanContextPropagator.cs
```diff
@@ -310,6 +310,43 @@ internal void ExtractHeaderTags<THeaders, TProcessor>(ref TProcessor processor,
             }
         }
 
+        internal void AddBaggageToSpanAsTags(Tracer tracer, ISpan span, Baggage? baggage)
```
zacharycmontoya: nit: I recommend getting the set of keys from `tracer.Settings.BaggageTagKeys` and passing have that list as an argument to this method instead of the tracer object. I think that will make the design cleaner

tracer/test/Datadog.Trace.Tests/SpanContextPropagatorTests_ExtractBaggageHeaderTags.cs
```diff
@@ -0,0 +1,142 @@
...
+using Datadog.Trace.TestHelpers.TestTracer;
+using FluentAssertions;
+using Xunit;
+
+namespace Datadog.Trace.Tests
+{
+    [Collection(nameof(WebRequestCollection))]
+    public class SpanContextPropagatorTests_ExtractBaggageHeaderTags
+    {
+        public enum HeaderCollectionType
```
zacharycmontoya: It looks to me like you can delete all of the code prior to the first test `AddBaggageTags`

tracer/test/Datadog.Trace.Tests/SpanContextPropagatorTests_ExtractBaggageHeaderTags.cs
```diff
@@ -0,0 +1,142 @@
...
+using Datadog.Trace.Headers;
+using Datadog.Trace.Propagators;
+using Datadog.Trace.TestHelpers;
+using Datadog.Trace.TestHelpers.TestTracer;
+using FluentAssertions;
+using Xunit;
+
+namespace Datadog.Trace.Tests
+{
+    [Collection(nameof(WebRequestCollection))]
```
zacharycmontoya: You can also delete this [Collection] attribute entirely based on my below comment since you're not interacting with WebRequest objects

tracer/test/Datadog.Trace.Tests/SpanContextPropagatorTests_ExtractBaggageHeaderTags.cs
```diff
@@ -0,0 +1,142 @@
...
+
+        [Fact]
+        internal async Task AddBaggageTags()
+        {
+            // Add Baggage items using a key that's in the default configuration
+            var baggageItems = new Dictionary<string, string>();
+            baggageItems.Add("user.id", "doggo");
+            var baggage = new Baggage(baggageItems);
+
+            var expectedResult = "doggo";
```
zacharycmontoya: super small nit: I think you can remove this and just set the expectedResult below to `"doggo"`

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNet/AspNetMvc5Tests.cs
```diff
@@ -173,6 +174,110 @@ public AspNetMvc5TestsInferredProxySpansDisabled(IisFixture iisFixture, ITestOut
...
+            SetEnvironmentVariable(ConfigurationKeys.ExpandRouteTemplatesEnabled, false.ToString());
+            SetEnvironmentVariable(ConfigurationKeys.FeatureFlags.TraceId128BitGenerationEnabled, false.ToString());
+            SetEnvironmentVariable(ConfigurationKeys.FeatureFlags.InferredProxySpansEnabled, false.ToString());
+
+            _classicMode = false;
+            _enableInferredProxySpans = false;
+            _iisFixture = iisFixture;
+            _iisFixture.ShutdownPath = "/home/shutdown";
+
+            _testName = $"{nameof(AspNetMvc5Tests)}.WithBaggage";
```
zacharycmontoya: To make it easier for me to review the snapshots, can you update the _testName so that it has the same naming convention as the other tests, following this logic?

https://github.com/DataDog/dd-trace-dotnet/blob/98f5d1c35d36edb53b6bc477916788592354c152/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNet/AspNetMvc5Tests.cs#L211-L217

I'm happy to update the snapshots after you make the change

zacharycmontoya (approved):
LGTM, thanks! Please get a second approval before merging, especially since my name is in some of the commits for build fixes/snapshot commits

tracer/src/Datadog.Trace/PlatformHelpers/AspNetCoreHttpRequestHandler.cs
```diff
@@ -98,6 +98,21 @@ private void AddHeaderTagsToSpan(ISpan span, HttpRequest request, Tracer tracer)
             }
         }
 
+        private void AddBaggageTagsToSpan(ISpan span, Baggage baggage, Tracer tracer)
```
lucaspimentel, line 101: This exact method is copied 4 times, which makes it a good candidate for refactoring to reduce duplication. Not only that, but pretty much all they do is
- check if `baggage` is `null`
- call `SpanContextPropagator.AddBaggageToSpanAsTags()`
- log any exception

`SpanContextPropagator.AddBaggageToSpanAsTags()` already checks if `baggage` is `null`, so we could move the `try`/`catch` into that method and call it directly from all 4 web server integrations.

tracer/src/Datadog.Trace/Propagators/SpanContextPropagator.cs
```diff
@@ -310,6 +310,40 @@ internal void ExtractHeaderTags<THeaders, TProcessor>(ref TProcessor processor,
...
+            }
+
+            // add only specified baggage items as tags
+            foreach (var key in baggageTagKeys)
+            {
+                if (baggage.TryGetValue(key, out var value))
+                {
+                    span.SetTag("baggage." + key, value);
+                }
+            }
```
lucaspimentel, line 344: You can consolidate these two loops into one to make the code smaller, which increases the chances of the method getting inlined (which is good for performance since this is on the hot path).

Something like
```csharp
var addAllItems = baggageTagKeys.Length == 1 && baggageTagKeys[0] == "*";

foreach (var key in baggageTagKeys)
{
    if (addAllItems || baggage.TryGetValue(key, out var value))
    {
        span.SetTag("baggage." + key, value);
    }
}
```

If `addAllItems` is `true`, it will short-circuit the `if` so `baggage.TryGetValue()` won't be called.

tracer/src/Datadog.Trace/Propagators/SpanContextPropagator.cs
```diff
@@ -310,6 +310,40 @@ internal void ExtractHeaderTags<THeaders, TProcessor>(ref TProcessor processor,
...
+                return;
+            }
+
+            if (baggageTagKeys.Length == 0)
+            {
+                // feature disabled
+                return;
+            }
+
+            if (baggageTagKeys.Length == 1 && baggageTagKeys[0] == "*")
```
lucaspimentel, line 326: What is the expected behavior if I set `DD_TRACE_BAGGAGE_TAG_KEYS` to multiple values and one of them is `*`? For example:
```
DD_TRACE_BAGGAGE_TAG_KEYS=foo,*,bar
```

It that valid config? I expect it's either invalid, or it should add all baggage just like a simple `"*"` (since `*` includes `foo` and `bar).

tracer/test/snapshots/OwinWebApi2Tests.NoFF.__path=_api2_delay_0_statusCode=200.verified.txt
```diff
@@ -10,6 +10,7 @@
       aspnet.action: delay,
       aspnet.controller: conventions,
       aspnet.route: api2/{action}/{value},
+      baggage.user.id: doggo,
```
lucaspimentel, line 13: Is it expected that all these `OwinWebApi2Tests.*` snapshots now have a `baggage.user.id: doggo` tag?

tracer/src/Datadog.Trace/Propagators/SpanContextPropagator.cs
rachelyangdog: It would technically be 'invalid'. But in actuality, we would look for a baggage key equal to "*". foo and bar would still be added as tags if found.

tracer/test/snapshots/OwinWebApi2Tests.NoFF.__path=_api2_delay_0_statusCode=200.verified.txt
rachelyangdog: Yes, instead of creating new snapshots. We decided to add the tag to the existing snapshot tests.

tracer/src/Datadog.Trace/Propagators/SpanContextPropagator.cs
lucaspimentel: > It would technically be 'invalid'

If it's invalid, then we should reject the entire setting and not add `foo` and `bar` as tags, right? We can validate this in `TracerSettings`.

tracer/test/snapshots/OwinWebApi2Tests.NoFF.__path=_api2_delay_0_statusCode=200.verified.txt
lucaspimentel: Ok, thanks for clarifying. Resolving this comment.

lucaspimentel (approved):
Thanks!

tracer/src/Datadog.Trace/PlatformHelpers/AspNetCoreHttpRequestHandler.cs
lucaspimentel: Not a hard blocker for this PR. Feel free to do in a separate PR later.

tracer/src/Datadog.Trace/Propagators/SpanContextPropagator.cs
lucaspimentel: Not a hard blocker for this PR. Feel free to do in a separate PR later.

tracer/src/Datadog.Trace/Propagators/SpanContextPropagator.cs
rachelyangdog: Sorry. I meant that it would technically be 'valid' since a [baggage key could theoretically be just "*". ](https://docs.google.com/document/d/1cX4bWYW-EPqKc64oynXpeHOI8FCW3XI-h0uH1tnzV3Q/edit?tab=t.0).

lucaspimentel: Oh! That didn't even occur to me. Thanks! (resolving)

tracer/src/Datadog.Trace/PlatformHelpers/AspNetCoreHttpRequestHandler.cs
rachelyangdog: Ok! I'll create a cleanup PR right after this one. Thank youu!!

tracer/src/Datadog.Trace/Propagators/SpanContextPropagator.cs
rachelyangdog: Ok! I'll create a cleanup PR right after this one. Thank youu!!

PR #7027, opened 2025-05-28
Use ReaderWriterLock instead of AsyncManualResetEvent

kr-igor:

### DIFF
--- a/tracer/src/Datadog.Trace/DataStreamsMonitoring/Aggregation/DataStreamsAggregator.cs
+++ b/tracer/src/Datadog.Trace/DataStreamsMonitoring/Aggregation/DataStreamsAggregator.cs
@@ -8,7 +8,9 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Threading;
 using Datadog.Trace.DataStreamsMonitoring.Utils;
+using Datadog.Trace.Vendors.Serilog;
 
 namespace Datadog.Trace.DataStreamsMonitoring.Aggregation;
 
@@ -18,6 +20,9 @@ namespace Datadog.Trace.DataStreamsMonitoring.Aggregation;
 /// </summary>
 internal class DataStreamsAggregator
 {
+    private readonly ReaderWriterLock _lock = new();
+    private readonly int _lockTimeoutMs = 500;
+
     // The inner dictionary is constrained in size by the number of unique hashes seen by the app
     // Unique hashes are unique paths from origin to here, which could be unbounded if there are loops
     // The outer dictionary is constrained by FlushInterval/BucketDuration + 1
@@ -34,6 +39,7 @@ internal class DataStreamsAggregator
     private readonly long _bucketDurationInNs;
     private List<SerializableStatsBucket>? _statsToWrite;
     private List<SerializableBacklogBucket>? _backlogsToWrite;
+    private long _pointsDropped;
 
     public DataStreamsAggregator(DataStreamsMessagePackFormatter formatter, int bucketDurationMs)
     {
@@ -46,26 +52,56 @@ internal class DataStreamsAggregator
     /// </summary>
     public void Add(in StatsPoint point)
     {
-        var currentBucketStartTime = BucketStartTimeForTimestamp(point.TimestampNs);
-        AddToBuckets(point, currentBucketStartTime, _currentBuckets);
+        try
+        {
+            _lock.AcquireReaderLock(_lockTimeoutMs);
+            try
+            {
+                var currentBucketStartTime = BucketStartTimeForTimestamp(point.TimestampNs);
+                AddToBuckets(point, currentBucketStartTime, _currentBuckets);
 
-        var originTimestamp = point.TimestampNs - point.PathwayLatencyNs;
-        var originBucketStartTime = BucketStartTimeForTimestamp(originTimestamp);
-        AddToBuckets(point, originBucketStartTime, _originBuckets);
+                var originTimestamp = point.TimestampNs - point.PathwayLatencyNs;
+                var originBucketStartTime = BucketStartTimeForTimestamp(originTimestamp);
+                AddToBuckets(point, originBucketStartTime, _originBuckets);
+            }
+            finally
+            {
+                _lock.ReleaseReaderLock();
+            }
+        }
+        catch
+        {
+            Interlocked.Increment(ref _pointsDropped);
+        }
     }
 
     public void AddBacklog(in BacklogPoint point)
     {
-        var currentBucketStartTime = BucketStartTimeForTimestamp(point.TimestampNs);
-        if (!_backlogBuckets.TryGetValue(currentBucketStartTime, out var bucket))
+        try
         {
-            bucket = new Dictionary<string, BacklogBucket>();
-            _backlogBuckets[currentBucketStartTime] = bucket;
-        }
+            _lock.AcquireReaderLock(_lockTimeoutMs);
+            try
+            {
+                var currentBucketStartTime = BucketStartTimeForTimestamp(point.TimestampNs);
+                if (!_backlogBuckets.TryGetValue(currentBucketStartTime, out var bucket))
+                {
+                    bucket = new Dictionary<string, BacklogBucket>();
+                    _backlogBuckets[currentBucketStartTime] = bucket;
+                }
 
-        if (!bucket.TryGetValue(point.Tags, out var group) || group.Value < point.Value)
+                if (!bucket.TryGetValue(point.Tags, out var group) || group.Value < point.Value)
+                {
+                    bucket[point.Tags] = new BacklogBucket(point.Tags, point.Value);
+                }
+            }
+            finally
+            {
+                _lock.ReleaseReaderLock();
+            }
+        }
+        catch
         {
-            bucket[point.Tags] = new BacklogBucket(point.Tags, point.Value);
+            Interlocked.Increment(ref _pointsDropped);
         }
     }
 
@@ -77,17 +113,40 @@ internal class DataStreamsAggregator
     /// <returns>True if data was serialized to the stream</returns>
     public bool Serialize(Stream stream, long maxBucketFlushTimeNs)
     {
-        var statsToAdd = Export(maxBucketFlushTimeNs) ?? new();
-        var backlogsToAdd = ExportBacklogs(maxBucketFlushTimeNs) ?? new();
-        if (statsToAdd.Count > 0 || backlogsToAdd.Count > 0)
+        Interlocked.Exchange(ref _pointsDropped, 0);
+        try
         {
-            _formatter.Serialize(stream, _bucketDurationInNs, statsToAdd, backlogsToAdd);
-            Clear(statsToAdd, backlogsToAdd);
+            _lock.AcquireWriterLock(_lockTimeoutMs);
+            try
+            {
+                var statsToAdd = Export(maxBucketFlushTimeNs) ?? new();
+                var backlogsToAdd = ExportBacklogs(maxBucketFlushTimeNs) ?? new();
+                if (statsToAdd.Count > 0 || backlogsToAdd.Count > 0)
+                {
+                    _formatter.Serialize(stream, _bucketDurationInNs, statsToAdd, backlogsToAdd);
+                    Clear(statsToAdd, backlogsToAdd);
 
-            return true;
+                    return true;
+                }
+
+                return false;
+            }
+            finally
+            {
+                var dropped = Interlocked.Read(ref _pointsDropped);
+                _lock.ReleaseWriterLock();
+
+                if (dropped > 0)
+                {
+                    Log.Debug("Dropped {Dropped} data points during serialization", dropped);
+                }
+            }
         }
-
-        return false;
+        catch
+        {
+            Log.Debug("Failed to serialize data streams data.");
+            return false;
+        }
     }
 
     internal List<SerializableBacklogBucket> ExportBacklogs(long maxBucketFlushTimeNs)
--- a/tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
+++ b/tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
@@ -25,7 +25,6 @@ internal class DataStreamsWriter : IDataStreamsWriter
     private readonly BoundedConcurrentQueue<StatsPoint> _buffer = new(queueLimit: 10_000);
     private readonly BoundedConcurrentQueue<BacklogPoint> _backlogBuffer = new(queueLimit: 10_000);
     private readonly Task _processTask;
-    private readonly AsyncManualResetEvent _processingMutex = new(set: false);
     private readonly TaskCompletionSource<bool> _processExit = new(TaskCreationOptions.RunContinuationsAsynchronously);
     private readonly DataStreamsAggregator _aggregator;
     private readonly IDiscoveryService _discoveryService;
@@ -82,35 +81,28 @@ internal class DataStreamsWriter : IDataStreamsWriter
 
     public void Add(in StatsPoint point)
     {
-        if ((Volatile.Read(ref _isSupported) != SupportState.Unsupported)
-         && _buffer.TryEnqueue(point))
+        if (Volatile.Read(ref _isSupported) != SupportState.Unsupported)
         {
-            if (!_processingMutex.IsSet)
+            if (_buffer.TryEnqueue(point))
             {
-                _processingMutex.Set();
+                return;
             }
         }
-        else
-        {
-            // TODO: Monitor with telemetry
-            Interlocked.Increment(ref _pointsDropped);
-        }
+
+        Interlocked.Increment(ref _pointsDropped);
     }
 
     public void AddBacklog(in BacklogPoint point)
     {
-        if ((Volatile.Read(ref _isSupported) != SupportState.Unsupported)
-         && _backlogBuffer.TryEnqueue(point))
+        if (Volatile.Read(ref _isSupported) != SupportState.Unsupported)
         {
-            if (!_processingMutex.IsSet)
+            if (_backlogBuffer.TryEnqueue(point))
             {
-                _processingMutex.Set();
+                return;
             }
         }
-        else
-        {
-            Interlocked.Increment(ref _pointsDropped);
-        }
+
+        Interlocked.Increment(ref _pointsDropped);
     }
 
     public async Task DisposeAsync()
@@ -151,10 +143,6 @@ internal class DataStreamsWriter : IDataStreamsWriter
     private void RequestFlush()
     {
         Interlocked.Exchange(ref _flushRequested, 1);
-        if (!_processingMutex.IsSet)
-        {
-            _processingMutex.Set();
-        }
     }
 
     private async Task WriteToApiAsync()
@@ -234,11 +222,7 @@ internal class DataStreamsWriter : IDataStreamsWriter
                 // do one more loop to make sure everything is flushed
                 RequestFlush();
                 isFinalFlush = true;
-                continue;
             }
-
-            await _processingMutex.WaitAsync().ConfigureAwait(false);
-            _processingMutex.Reset();
         }
     }
 
--- a/tracer/src/Datadog.Trace.Trimming/build/Datadog.Trace.Trimming.xml
+++ b/tracer/src/Datadog.Trace.Trimming/build/Datadog.Trace.Trimming.xml
@@ -1033,6 +1033,7 @@
       <type fullname="System.Threading.ManualResetEventSlim" />
       <type fullname="System.Threading.Monitor" />
       <type fullname="System.Threading.Mutex" />
+      <type fullname="System.Threading.ReaderWriterLock" />
       <type fullname="System.Threading.ReaderWriterLockSlim" />
       <type fullname="System.Threading.SemaphoreSlim" />
       <type fullname="System.Threading.SpinLock" />


andrewlock:

- [This PR (7027)](https://github.com/DataDog/dd-trace-dotnet/tree/59d8577d36c82b923896ede5fd42d97b61836d9a)
- [master](https://github.com/DataDog/dd-trace-dotnet/tree/c05ddfb1191eb57e7281f6eac462b82d76cbf74c)

```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Baseline
    This PR (7027) - mean (68ms)  : 65, 71
     .   : milestone, 68,
    master - mean (68ms)  : 65, 72
     .   : milestone, 68,

    This PR (7027) - mean (1,017ms)  : 1001, 1032
     .   : milestone, 1017,
    master - mean (1,019ms)  : 1003, 1034
     .   : milestone, 1019,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Baseline
    This PR (7027) - mean (102ms)  : 100, 105
     .   : milestone, 102,
    master - mean (102ms)  : 100, 104
     .   : milestone, 102,

    This PR (7027) - mean (699ms)  : 689, 708
     .   : milestone, 699,
    master - mean (701ms)  : 691, 711
     .   : milestone, 701,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Baseline
    This PR (7027) - mean (89ms)  : 86, 92
     .   : milestone, 89,
    master - mean (89ms)  : 87, 91
     .   : milestone, 89,

    This PR (7027) - mean (670ms)  : 651, 689
     .   : milestone, 670,
    master - mean (673ms)  : 658, 688
     .   : milestone, 673,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Baseline
    This PR (7027) - mean (191ms)  : 187, 195
     .   : milestone, 191,
    master - mean (191ms)  : 187, 195
     .   : milestone, 191,

    This PR (7027) - mean (1,168ms)  : 1150, 1186
     .   : milestone, 1168,
    master - mean (1,160ms)  : 1141, 1178
     .   : milestone, 1160,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Baseline
    This PR (7027) - mean (270ms)  : 266, 273
     .   : milestone, 270,
    master - mean (270ms)  : 265, 275
     .   : milestone, 270,

    This PR (7027) - mean (932ms)  : 918, 947
     .   : milestone, 932,
    master - mean (932ms)  : 919, 945
     .   : milestone, 932,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Baseline
    This PR (7027) - mean (263ms)  : 260, 266
     .   : milestone, 263,
    master - mean (263ms)  : 260, 266
     .   : milestone, 263,

    This PR (7027) - mean (928ms)  : 908, 948
     .   : milestone, 928,
    master - mean (929ms)  : 911, 947
     .   : milestone, 929,

```

tracer/src/Datadog.Trace/DataStreamsMonitoring/Aggregation/DataStreamsAggregator.cs
```diff
@@ -77,17 +113,40 @@ public void AddBacklog(in BacklogPoint point)
...
+                return false;
+            }
+            finally
+            {
+                var dropped = Interlocked.Read(ref _pointsDropped);
+                _lock.ReleaseWriterLock();
+
+                if (dropped > 0)
+                {
+                    Log.Debug("Dropped {Dropped} data points during serialization", dropped);
```
lucaspimentel: This (and line 147 below) is using `Datadog.Trace.Vendors.Serilog.Log`, which is probably not what you want. Instead, the usual logging pattern is adding a static `IDatadogLogger Log` field.

```diff
-using Datadog.Trace.Vendors.Serilog;

namespace Datadog.Trace.DataStreamsMonitoring.Aggregation;

/// <summary>
/// Aggregates multiple <see cref="StatsPoint"/>s into their correct buckets
/// Note that this class is not thread safe
/// </summary>
internal class DataStreamsAggregator
{
+    private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor<DataStreamsAggregator>();
```

You can find example usages by searching for `DatadogLogging.GetLoggerFor`.

tracer/src/Datadog.Trace/DataStreamsMonitoring/Aggregation/DataStreamsAggregator.cs
```diff
@@ -18,6 +20,9 @@ namespace Datadog.Trace.DataStreamsMonitoring.Aggregation;
 /// </summary>
 internal class DataStreamsAggregator
 {
+    private readonly ReaderWriterLock _lock = new();
```
lucaspimentel: Can you use `ReaderWriterLockSlim` instead of `ReaderWriterLock`? [docs](https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-threading-readerwriterlockslim)
It's meant to be faster and more efficient.

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
```diff
@@ -237,8 +227,10 @@ private async Task ProcessQueueLoopAsync()
                 continue;
             }
 
-            await _processingMutex.WaitAsync().ConfigureAwait(false);
-            _processingMutex.Reset();
+            if (!_completionSource.Task.IsCompleted)
+            {
+                _completionSource.Task.Wait(_waitTimeSpan);
```
lucaspimentel: We need to be very careful here. Using a synchronous `Task.Wait()` in async code is generally not recommended. It can cause deadlocks in some runtimes (e.g. ASPNET) and by blocking the thread, it can starve the thread pool.

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
```diff
@@ -22,15 +22,16 @@ internal class DataStreamsWriter : IDataStreamsWriter
 {
     private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor<DataStreamsWriter>();
 
-    private readonly BoundedConcurrentQueue<StatsPoint> _buffer = new(queueLimit: 10_000);
-    private readonly BoundedConcurrentQueue<BacklogPoint> _backlogBuffer = new(queueLimit: 10_000);
```
lucaspimentel: [coding style] Can you revert this two lines? 😅  We like having the parameter name (`queueLimit`) in cases like this with no variable names so it's obvious what the `10_000` means when reading the code.

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
andrewlock: Yeah, this is basically a "never ever do this" pattern in our code. It _will_ break someone, practically guaranteed

kr-igor: This is something I forgot to cleanup from my experiments. The recent version of the code is using proper wait async.

tracer/src/Datadog.Trace/DataStreamsMonitoring/Aggregation/DataStreamsAggregator.cs
```diff
@@ -46,26 +53,56 @@ public DataStreamsAggregator(DataStreamsMessagePackFormatter formatter, int buck
     /// </summary>
     public void Add(in StatsPoint point)
     {
-        var currentBucketStartTime = BucketStartTimeForTimestamp(point.TimestampNs);
-        AddToBuckets(point, currentBucketStartTime, _currentBuckets);
+        try
+        {
+            _lock.TryEnterReadLock(_lockTimeoutMs);
```
lucaspimentel: `TryEnterReadLock()` returns "`true` if the calling thread entered read mode, otherwise, `false`." ([docs](https://learn.microsoft.com/en-us/dotnet/api/system.threading.readerwriterlockslim.tryenterreadlock#system-threading-readerwriterlockslim-tryenterreadlock(system-int32)))

Shouldn't we check the return value?

tonyredondo: Yes, you have to check the returned value, otherwise this synchronization will not work because you could be modifying the Dictionaries while serializing (write lock), and also `ExitReadLock` will throw `SynchronizationLockException` if you don't own that lock (`TryEnterReadLock == false`)

Also I guess you are interested in incrementing `_pointsDropped` when `false`.

tracer/src/Datadog.Trace/DataStreamsMonitoring/Aggregation/DataStreamsAggregator.cs
```diff
@@ -46,26 +53,56 @@ public DataStreamsAggregator(DataStreamsMessagePackFormatter formatter, int buck
...
+            {
+                var currentBucketStartTime = BucketStartTimeForTimestamp(point.TimestampNs);
+                AddToBuckets(point, currentBucketStartTime, _currentBuckets);
 
-        var originTimestamp = point.TimestampNs - point.PathwayLatencyNs;
-        var originBucketStartTime = BucketStartTimeForTimestamp(originTimestamp);
-        AddToBuckets(point, originBucketStartTime, _originBuckets);
+                var originTimestamp = point.TimestampNs - point.PathwayLatencyNs;
+                var originBucketStartTime = BucketStartTimeForTimestamp(originTimestamp);
+                AddToBuckets(point, originBucketStartTime, _originBuckets);
```
tonyredondo: As you know this is not thread safe (multiple threads can enter in read mode/lock) and you have code to modify the bucket dictionary in AddToBuckets, even this may not throw an exception you could be dropping points without noticing it or incrementing the counter. Are you ok with that?

Modifying dictionaries on a read lock makes this code confusing to read.
I guess the DDSketch object is thread safe, consider enabling a write lock when modifying the dictionary.

tracer/src/Datadog.Trace/DataStreamsMonitoring/Aggregation/DataStreamsAggregator.cs
```diff
@@ -46,26 +53,56 @@ public DataStreamsAggregator(DataStreamsMessagePackFormatter formatter, int buck
...
+                if (!_backlogBuckets.TryGetValue(currentBucketStartTime, out var bucket))
+                {
+                    bucket = new Dictionary<string, BacklogBucket>();
+                    _backlogBuckets[currentBucketStartTime] = bucket;
+                }
+
+                if (!bucket.TryGetValue(point.Tags, out var group) || group.Value < point.Value)
+                {
+                    bucket[point.Tags] = new BacklogBucket(point.Tags, point.Value);
+                }
```
tonyredondo: same as my previous comment this is not thread safe, so you may loose data here if this doesn't throw.

tracer/src/Datadog.Trace/DataStreamsMonitoring/Aggregation/DataStreamsAggregator.cs
```diff
@@ -46,26 +53,56 @@ public DataStreamsAggregator(DataStreamsMessagePackFormatter formatter, int buck
...
 
     public void AddBacklog(in BacklogPoint point)
     {
-        var currentBucketStartTime = BucketStartTimeForTimestamp(point.TimestampNs);
-        if (!_backlogBuckets.TryGetValue(currentBucketStartTime, out var bucket))
+        try
         {
-            bucket = new Dictionary<string, BacklogBucket>();
-            _backlogBuckets[currentBucketStartTime] = bucket;
+            _lock.TryEnterReadLock(_lockTimeoutMs);
```
tonyredondo: you need to handle the return value

tracer/src/Datadog.Trace/DataStreamsMonitoring/Aggregation/DataStreamsAggregator.cs
```diff
@@ -77,17 +114,40 @@ public void AddBacklog(in BacklogPoint point)
...
     {
-        var statsToAdd = Export(maxBucketFlushTimeNs) ?? new();
-        var backlogsToAdd = ExportBacklogs(maxBucketFlushTimeNs) ?? new();
-        if (statsToAdd.Count > 0 || backlogsToAdd.Count > 0)
+        Interlocked.Exchange(ref _pointsDropped, 0);
+        try
         {
-            _formatter.Serialize(stream, _bucketDurationInNs, statsToAdd, backlogsToAdd);
-            Clear(statsToAdd, backlogsToAdd);
+            _lock.TryEnterWriteLock(_lockTimeoutMs);
```
tonyredondo: You need to handle the return value here to check if you own the lock or not. Or change this to `EnterWriteLock`

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
```diff
@@ -237,8 +227,17 @@ private async Task ProcessQueueLoopAsync()
...
+            {
+                try
+                {
+                    await _completionSource.Task.WaitAsync(_waitTimeSpan).ConfigureAwait(false);
+                }
+                catch
+                {
+                    // ignore the exception, it's expected
+                }
+            }
```
tonyredondo: I know you are trying to maximize throughput. The only effective async operation in this loop is the `WriteToApiAsync()` call, but you are not handling any return value from that call. So you could change the WriteToApiAsync code to be mostly sync and just wrap the final `api.Send... + success logic + log + event triggering` in a different async method that you just call it in a fire and forget manner with a completion source to wait in FlushAndClose() 

With that you could modify this complete async method with the loop with a sync one. And this Wait with a `ManualResetEventSlim.Wait`.

Also, very important you will need to change the `Task.Run` with a LongLiving one to avoid any lock in a threadpool thread.

I think by having this closed loop in sync mode will perform better than with all the async. But is just a feeling, you would need to test that hypothesis.

tonyredondo (commented):
I've left a couple of comments.

Have you considered using `System.Threading.Channels` approach for the latest runtime versions (.netcoreapp3.1 and >= .NET 5)?

tracer/src/Datadog.Trace/DataStreamsMonitoring/Aggregation/DataStreamsAggregator.cs
andrewlock: > so you may loose data here if this doesn't throw

Or to rephrase, this _could_ throw, so you probably shouldn't do it 😄

kr-igor:
> Have you considered using `System.Threading.Channels` approach for the latest runtime versions (.netcoreapp3.1 and >= .NET 5)?

I didn't try Channels because they are not a part of the framework in older versions, which will require adding some defines and complicate the code. I can try how they perform tho, maybe the added code complexity will be justified by the performance differences.

tracer/src/Datadog.Trace/DataStreamsMonitoring/Aggregation/DataStreamsAggregator.cs
kr-igor: My bad, i missed the return value and thought it will raise an exception when failed to acquire. I'll rework this part. Thanks!

tracer/src/Datadog.Trace/DataStreamsMonitoring/Aggregation/DataStreamsAggregator.cs
kr-igor: This code is only called from the flushing background task in `DataStreamsWriter->ProcessQueueLoopAsync`. The concurrency on add is handled before that using a `ConcurrentQueue`. The read / write look here is useful because we need to block processing `only` when flushing the data, i.e. the only place where multiple tasks have access to the dictionary.

kr-igor: I agree it's not immediately obvious from the code, but it's the way it was originally implemented and I didn't want to refactor everything :D

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
```diff
@@ -227,17 +229,7 @@ private async Task ProcessQueueLoopAsync()
...
-                try
-                {
-                    await _completionSource.Task.WaitAsync(_waitTimeSpan).ConfigureAwait(false);
-                }
-                catch
-                {
-                    // ignore the exception, it's expected
-                }
-            }
+            Thread.Sleep(_waitTimeSpan);
```
tonyredondo: `Thread.Sleep` do a kernel call to sleep the thread (https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Threading/Thread.Windows.cs#L22), the implementation in `ManualResetEventSlim.Wait(10)` uses a SpinWait that should be more performant for tiny timeouts. https://github.com/dotnet/runtime/blob/1d1bf92fcf43aa6981804dc53c5174445069c9e4/src/libraries/System.Private.CoreLib/src/System/Threading/ManualResetEventSlim.cs#L515-L528

I did not benchmark it, but because you are running tests, maybe it worth to try it out.

tonyredondo: You can use SpinWait directly 🤦🏻 : https://learn.microsoft.com/en-us/dotnet/api/system.threading.spinwait?view=net-9.0

kr-igor: Surprisingly it is a bit slower than using the WaitAsync.... I've copied the logic from the framework and removed the exception raising / handling.

tracer/src/Datadog.Trace/DataStreamsMonitoring/Aggregation/DataStreamsAggregator.cs
kr-igor: I realized that his class is always called from one method sequentially (add -> add -> flush), so no need to make it thread safe.

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
kr-igor: Async performs better, tested last Friday.

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
tonyredondo: This is surprising, just out of curiosity, how are you testing this? what's your setup?

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
```diff
@@ -237,8 +227,16 @@ private async Task ProcessQueueLoopAsync()
                 continue;
             }
 
-            await _processingMutex.WaitAsync().ConfigureAwait(false);
-            _processingMutex.Reset();
+            if (!_completionSource.Task.IsCompleted)
+            {
+                // The logic is copied from https://github.com/dotnet/runtime/blob/main/src/libraries/Common/tests/System/Threading/Tasks/TaskTimeoutExtensions.cs#L26
```
andrewlock: nit: we could use the built-in implementation directly in .NET 6+ (so we transparently get any improvements they may introduce in .NET 6+)

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
andrewlock: Yeah, this seems surprising to me too 😃 Even if it is faster, this is way more allocate-y right? all the extra objects + await machinery 🤔 Is it worth the tradeoff?

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
kr-igor: We can refactor later, the built-in implementation throws an exception on timeout, which requires adding a try/catch block. It's an extra allocation + profiler shows 1000s of exceptions :)

kr-igor:
> I can see that this improves the throughput in high-message per second scenarios, but what's the impact on resource usage in low-message per second scenarios 🤔 It _looks_ like we're consistently constantly allocating in this loop, every 10ms? Even if there are no messages _ever_ sent in the app, i.e. enabling DSM by default would have a negative impact on perf even if there's no messaging in the app? Can we measure that impact?

Good point. I manually tested an application which was producing ~200msg/sec with the tracer from master vs the tracer from this branch. This branch uses ~20% less CPU for the same throughput with no noticeable overhead in allocations. You can see trace comparison [here](https://ddstaging.datadoghq.com/profiling/comparison?compare_end_A=1749836024088.1345&compare_end_B=1749837530742&compare_query_A=env%3Astaging%20service%3Akafka-dotnet-producer&compare_query_B=env%3Astaging%20service%3Akafka-dotnet-producer&compare_start_A=1749834224088.1345&compare_start_B=1749836779175.4634&compareValuesMode=absolute&comparisonViz=flamegraph&my_code=disabled&profile_type=alloc-samples) (A is from the branch, B from master)

andrewlock (approved):

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
```diff
@@ -126,7 +120,7 @@ public async Task DisposeAsync()
 
     private async Task FlushAndCloseAsync()
     {
-        if (!_processExit.TrySetResult(true))
+        if (!_processExit.TrySetResult(true) || !_completionSource.TrySetResult(true))
```
andrewlock: Why do we need the new `_completionSource` if it's just set at the same time as `_processExit` 🤔 Can't we just use that?

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
```diff
@@ -237,8 +227,16 @@ private async Task ProcessQueueLoopAsync()
...
             }
 
-            await _processingMutex.WaitAsync().ConfigureAwait(false);
-            _processingMutex.Reset();
+            if (!_completionSource.Task.IsCompleted)
+            {
+                // The logic is copied from https://github.com/dotnet/runtime/blob/main/src/libraries/Common/tests/System/Threading/Tasks/TaskTimeoutExtensions.cs#L26
+                // and modified to avoid dealing with exceptions
+                var tcs = new TaskCompletionSource<bool>();
+                using (new Timer(s => ((TaskCompletionSource<bool>)s!).SetResult(true), tcs, _waitTimeSpan, Timeout.InfiniteTimeSpan))
```
andrewlock: We maybe already discussed this, but all this complexity seems to be to "just" make sure we break out of the 10ms if the process is exiting. I wonder if that's really necessary 🤔 It's only going to save us 5ms on average and we could replace all this logic with a 

```csharp
await Task.Delay(_waitTimeSpan);
```
or something similar, no? 🤔

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
kr-igor: I'm using [this](https://github.com/DataDog/data-streams-dev/tree/main/k8s-dev-env/dotnet/kafka-worker) test app and run it with two different setups - single and multi core. My approach is definitely a bit more allocate-y, but the CPU overhead with a dedicated thread + lock wait times were 10% higher (and throughput ~10% lower).

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
kr-igor: You are right, no need for a separate _completionSource. I'll update the code.

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
kr-igor: Task.Delay will also allocate a new task each time, right? I remember testing it in the beginning and it was slower due to increased CPU usage. I'll test one more time to make sure the difference is not marginal and it's worth keeping this custom implementation. I'll attach the traces here.

andrewlock: Yeah, `Task.Delay` would allocate a task too + there's the await machinery you're always going to get unless you use `Thread.Sleep(_waitTimeSpan);` which then has the kernel call 😅 But still seems a lot more light-weight than everything else here - fundamentally you're trying to do _more_ here (to make the wait cancellable) so would be v. surprised if it ends up being more efficient. And ultimately just questioning if that extra work (perf wise + complexity) is worth the 10ms (5ms) average delay on shutdown 🤷‍♂️

kr-igor:
Allocations for a simple application with Console.WriteLine in a loop and DSM enabled look the same in master and in this branch. [Here's](https://ddstaging.datadoghq.com/profiling/comparison?compare_end_A=1750090222949.9456&compare_end_B=1750091249382&compare_query_A=env%3Astaging%20service%3Akafka-dotnet-producer&compare_query_B=env%3Astaging%20service%3Akafka-dotnet-producer&compare_start_A=1750088422949.9456&compare_start_B=1750090534883.4548&compareValuesMode=absolute&comparisonViz=flamegraph&my_code=disabled&profile_type=alloc-samples) profile comparison (A = this branch, B = master).

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
kr-igor: This implementation is ~7% more efficient in my throughput test with 24.3k msg/sec vs 22.7k msg/sec with Task.Delay. You can check [here](https://ddstaging.datadoghq.com/apm/entity/service%3Akafka-dotnet-producer?deployments=qson%3A%28data%3A%28hits%3A%28selected%3Aversion_rate%29%2Cerrors%3A%28selected%3Aversion_count%29%2Clatency%3A%28selected%3Ap95%29%2CtopN%3A%215%29%2Cversion%3A%210%29&env=staging&fromUser=true&graphType=flamegraph&groupMapByOperation=null&operationName=KafkaProduce&panels=qson%3A%28data%3A%28%29%2Cversion%3A%210%29&resources=qson%3A%28data%3A%28visible%3A%21t%2Chits%3A%28selected%3Arate%29%2Cerrors%3A%28selected%3Atotal%29%2Clatency%3A%28selected%3Ap50%29%2CtopN%3A%215%29%2Cversion%3A%211%29&shouldShowLegend=true&sort=time&spanKind=internal&summary=qson%3A%28data%3A%28visible%3A%21t%2Cchanges%3A%28%29%2Cerrors%3A%28selected%3Acount%29%2Chits%3A%28selected%3Aversion_rate%29%2Clatency%3A%28selected%3Alatency%2Cslot%3A%28agg%3A95%29%2Cdistribution%3A%28isLogScale%3A%21f%29%2CshowTraceOutliers%3A%21t%2Coperation%3Aukv_graph_store%29%2Csublayer%3A%28slot%3A%28layers%3AserviceAndInferred%29%2Cselected%3Apercentage%29%2ClagMetrics%3A%28selectedMetric%3A%21s%2CselectedGroupBy%3A%21s%29%29%2Cversion%3A%211%29&start=1750098073881&end=1750100402006&paused=true#infrastructure)

PR #7075, opened 2025-06-06
feat(libdatadog-logger): integrate libdatadog logging

ganeshnj:
## Summary of changes

Fixes https://datadoghq.atlassian.net/browse/APMSP-1835

This PR adds enabling the file based logging for libdatadog, ie we start to write the logs to the configured file as it happens.

## Reason for change

This is missing feature in the native exporter and must have for troubleshooting. 

## Implementation details

This PR https://github.com/DataDog/libdatadog/pull/1018 does the bulk of the logging setup. In this, we are only calling the exposed APIs.

## Test coverage

Since all the coverage is done in libdatadog there nothing to test in CI, however I verified manually that log file is created and written.

## Other details

We will need https://github.com/DataDog/libdatadog/pull/1088 for fully support the file based logging practices.

Also, we will need a release of libdatadog to pass the CI.


andrewlock:

- [This PR (7075)](https://github.com/DataDog/dd-trace-dotnet/tree/860ab8795c6c02f3ba5fe1de33b62074bf118642)
- [master](https://github.com/DataDog/dd-trace-dotnet/tree/524bc48ac21c4bcdbb6ae1a160dae0d0cd265ed2)

```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Bailout
    This PR (7075) - mean (74ms)  : 72, 76
     .   : milestone, 74,
    master - mean (74ms)  : 73, 76
     .   : milestone, 74,

    section Baseline
    This PR (7075) - mean (71ms)  : 67, 74
     .   : milestone, 71,
    master - mean (71ms)  : 67, 75
     .   : milestone, 71,

    This PR (7075) - mean (1,032ms)  : 1002, 1062
     .   : milestone, 1032,
    master - mean (1,034ms)  : 993, 1075
     .   : milestone, 1034,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Bailout
    This PR (7075) - mean (109ms)  : 105, 112
     .   : milestone, 109,
    master - mean (108ms)  : 106, 109
     .   : milestone, 108,

    section Baseline
    This PR (7075) - mean (107ms)  : 103, 111
     .   : milestone, 107,
    master - mean (106ms)  : 104, 109
     .   : milestone, 106,

    This PR (7075) - mean (714ms)  : 693, 734
     .   : milestone, 714,
    master - mean (712ms)  : 692, 732
     .   : milestone, 712,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Bailout
    This PR (7075) - mean (94ms)  : 93, 96
     .   : milestone, 94,
    master - mean (95ms)  : 93, 97
     .   : milestone, 95,

    section Baseline
    This PR (7075) - mean (95ms)  : 92, 97
     .   : milestone, 95,
    master - mean (94ms)  : 92, 97
     .   : milestone, 94,

    This PR (7075) - mean (671ms)  : 653, 690
     .   : milestone, 671,
    master - mean (674ms)  : 654, 693
     .   : milestone, 674,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Bailout
    This PR (7075) - mean (192ms)  : 191, 194
     .   : milestone, 192,
    master - mean (192ms)  : 190, 194
     .   : milestone, 192,

    section Baseline
    This PR (7075) - mean (189ms)  : 184, 195
     .   : milestone, 189,
    master - mean (188ms)  : 185, 192
     .   : milestone, 188,

    This PR (7075) - mean (1,123ms)  : 1097, 1149
     .   : milestone, 1123,
    master - mean (1,121ms)  : 1090, 1151
     .   : milestone, 1121,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Bailout
    This PR (7075) - mean (270ms)  : 266, 273
     .   : milestone, 270,
    master - mean (270ms)  : 267, 273
     .   : milestone, 270,

    section Baseline
    This PR (7075) - mean (268ms)  : 265, 272
     .   : milestone, 268,
    master - mean (268ms)  : 264, 272
     .   : milestone, 268,

    This PR (7075) - mean (887ms)  : 859, 914
     .   : milestone, 887,
    master - mean (883ms)  : 858, 907
     .   : milestone, 883,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Bailout
    This PR (7075) - mean (263ms)  : 260, 265
     .   : milestone, 263,
    master - mean (262ms)  : 259, 265
     .   : milestone, 262,

    section Baseline
    This PR (7075) - mean (263ms)  : 259, 267
     .   : milestone, 263,
    master - mean (262ms)  : 259, 266
     .   : milestone, 262,

    This PR (7075) - mean (873ms)  : 835, 911
     .   : milestone, 873,
    master - mean (878ms)  : 839, 916
     .   : milestone, 878,

```
```mermaid
gantt
    dateFormat  X
    axisFormat %s
    todayMarker off
    section Bailout
    This PR (7075) - mean (261ms)  : 258, 264
     .   : milestone, 261,
    master - mean (261ms)  : 258, 263
     .   : milestone, 261,

    section Baseline
    This PR (7075) - mean (261ms)  : 257, 265
     .   : milestone, 261,
    master - mean (261ms)  : 258, 264
     .   : milestone, 261,

    This PR (7075) - mean (779ms)  : 758, 800
     .   : milestone, 779,
    master - mean (782ms)  : 760, 804
     .   : milestone, 782,

```

build/cmake/FindLibdatadog.cmake
```diff
@@ -62,27 +62,38 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
 else()
     if(CMAKE_SYSTEM_PROCESSOR STREQUAL "aarch64" OR CMAKE_SYSTEM_PROCESSOR STREQUAL "arm64")
         if(DEFINED ENV{IsAlpine} AND "$ENV{IsAlpine}" MATCHES "true")
-            set(SHA256_LIBDATADOG "4b64b58162d215a4f16b6ced4d602667565ebe20015341219daa998e3cf4e0a8" CACHE STRING "libdatadog sha256")
-            set(FILE_TO_DOWNLOAD libdatadog-aarch64-alpine-linux-musl.tar.gz)
+            FetchContent_Declare(libdatadog-install
```
andrewlock: What are these changes about 🤔

tracer/src/Datadog.Trace/LibDatadog/Error.cs
```diff
@@ -0,0 +1,27 @@
...
+[StructLayout(LayoutKind.Sequential)]
+internal readonly struct Error
+{
+    internal readonly FFIVec Message;
+
+    public LibDatadogException ToException()
+    {
+        var messageBytes = Message.ToByteArray();
+        var message = Encoding.UTF8.GetString(messageBytes);
+        return new LibDatadogException(message);
```
andrewlock: We could likely do this much more efficiently using stackalloc for small messages in .NET Core 3.1+ 🤔

tracer/src/Datadog.Trace/LibDatadog/LogEventLevelExtension.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        return level switch
+        {
+            Vendors.Serilog.Events.LogEventLevel.Verbose => LogEventLevel.Trace,
+            Vendors.Serilog.Events.LogEventLevel.Debug => LogEventLevel.Debug,
+            Vendors.Serilog.Events.LogEventLevel.Information => LogEventLevel.Info,
+            Vendors.Serilog.Events.LogEventLevel.Warning => LogEventLevel.Warn,
+            // We don't have a "Fatal" level in libdatadog, so we map it to Error since this is closed mapping
+            Vendors.Serilog.Events.LogEventLevel.Error or Vendors.Serilog.Events.LogEventLevel.Fatal => LogEventLevel.Error,
+            _ => throw new ArgumentOutOfRangeException(nameof(level), level, null)
```
andrewlock: In general I'm a fan of this pattern normally, but also we must never crash customers, so we maybe should just fallback to Error instead?

tracer/src/Datadog.Trace/LibDatadog/SpanExtensions.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Datadog.Trace.LibDatadog;
+
+internal static class SpanExtensions
+{
+    public static T[] ToArray<T>(this Span<T> span)
+    {
+        return span.ToArray();
+    }
+}
```
andrewlock: :do-not-want: This is very ambiguous and looks recursive to me 😕? Also not specific to `Libdatadog`

tracer/src/Datadog.Trace/LibDatadog/Logger.cs
```diff
@@ -0,0 +1,103 @@
...
+{
+    private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor<Logger>();
+    private static readonly Lazy<Logger> _instance = new(() => new Logger());
+
+    private readonly TracerSettings _tracerSettings;
+
+    private Logger()
+    {
+        _tracerSettings = TracerSettings.FromDefaultSourcesInternal();
+    }
```
andrewlock: I think we have to rework this in general. I think we need to take an approach similar to the telemetry logger, where it's lazy initialized 🤔

tracer/src/Datadog.Trace/LibDatadog/StdConfig.cs
```diff
@@ -0,0 +1,19 @@
...
+#nullable enable
+
+using System.Runtime.InteropServices;
+
+namespace Datadog.Trace.LibDatadog;
+
+/// <summary>
+/// Represents a configuration for a standard output/error logger.
+/// </summary>
+[StructLayout(LayoutKind.Sequential)]
```
andrewlock, line 15: Could/should we move this to a `Logging` sub-folder in the `LibDatadog` folder? It's already getting a bit confusing with the number of supporting types we have, would be nice to keep things a bit more organized IMO

tracer/src/Datadog.Trace/LibDatadog/TraceExporterErrorHandle.cs
```diff
@@ -0,0 +1,57 @@
...
+
+#nullable enable
+
+using System;
+using System.Runtime.InteropServices;
+using Datadog.Trace.Logging;
+
+namespace Datadog.Trace.LibDatadog;
+
+internal class TraceExporterErrorHandle : SafeHandle
```
andrewlock, line 14: Should this derive from `ErrorHandle`? Or should `ErrorHandle` actually be `LoggerErrorHandle`?

tracer/src/Datadog.Trace/Logging/Internal/DatadogLogging.cs
```diff
@@ -44,6 +44,8 @@ static DatadogLogging()
 
                 var domainMetadata = DomainMetadata.Instance;
                 SharedLogger = DatadogLoggingFactory.CreateFromConfiguration(in config, domainMetadata) ?? SharedLogger;
+
+                LibDatadog.Logger.Instance.Enable(config, domainMetadata);
```
andrewlock: I don't think this belongs here 🤔 It's fundamentally got nothing to do with _our_ logging system really, so it should probably be configured in the `TracerManagerFactory` just _after_ we configure the datapipeline (so that, for example, if we fail to configure it, we don't call this method, or if we _reconfigure_, we update as required)...

tracer/src/Datadog.Trace/Logging/Internal/DatadogLogging.cs
```diff
@@ -90,6 +94,7 @@ internal static void CleanLogFiles(int deleteAfter, string logsDirectory)
               {
                 "dotnet-tracer-*.log",
                 "dotnet-native-loader-*.log",
+                "dotnet-libdatadog-*.log",
```
andrewlock: Are the logs going to be strictly this, or should they be `dotnet-tracer-libdatadog-*` (seeing as we have `dotnet-tracer-managed` and `dotnet-tracer-native` for example

tracer/src/Datadog.Trace/Logging/Internal/DatadogLogging.cs
```diff
@@ -71,11 +73,13 @@ internal static void CloseAndFlush()
         internal static void Reset()
         {
             LoggingLevelSwitch.MinimumLevel = DefaultLogLevel;
+            LibDatadog.Logger.Instance.SetLogLevel(DefaultLogLevel);
```
andrewlock: I think we should _probably_ move these calls to the _callsites_ for these methods, e.g. in `GlobalSettings.SetDebugEnabledInternal()` and `GlobalSettings.Reset()` 🤔  We may run into testing issues otherwise

build/cmake/FindLibdatadog.cmake
ganeshnj: only for testing, since we don't have new libdatadog version available yet

pending release, I want to test it end to end and not stay blocked.

since freeze we can't do a release now now.

tracer/src/Datadog.Trace/LibDatadog/LogEventLevelExtension.cs
ganeshnj: totally, we can fallback to a sensible default

tracer/src/Datadog.Trace/LibDatadog/StdConfig.cs
ganeshnj: in my head, my plan is to create a dir structure as following
- Libdatadog
  - Common
  - DataPipeline
  - Logger

tracer/src/Datadog.Trace/LibDatadog/TraceExporterErrorHandle.cs
ganeshnj: Just `ErrorHandle` since we using common error https://github.com/DataDog/libdatadog/blob/main/ddcommon-ffi/src/error.rs#L14

tracer/src/Datadog.Trace/Logging/Internal/DatadogLogging.cs
ganeshnj: I'm flexible as we were talking offline, this could be a statement just before init of trace exporter.

we only need to respect the log level which will be outside the exporter AFAIK.

tracer/src/Datadog.Trace/Logging/Internal/DatadogLogging.cs
ganeshnj: I'm totally flexible on this, this is the whole point that tracer can tell what they want.

build/cmake/FindLibdatadog.cmake
ganeshnj: These must be reverted with newer version of libdatadog.

tracer/src/Datadog.Trace/LibDatadog/Error.cs
andrewlock: Done in https://github.com/DataDog/dd-trace-dotnet/pull/7084

tracer/src/Datadog.Trace/LibDatadog/LogEventLevelExtension.cs
andrewlock: Done in https://github.com/DataDog/dd-trace-dotnet/pull/7084

tracer/src/Datadog.Trace/LibDatadog/SpanExtensions.cs
andrewlock: Removed in https://github.com/DataDog/dd-trace-dotnet/pull/7084

tracer/src/Datadog.Trace/LibDatadog/Logger.cs
andrewlock: Changed in https://github.com/DataDog/dd-trace-dotnet/pull/7084

tracer/src/Datadog.Trace/LibDatadog/StdConfig.cs
andrewlock: Sounds good, not a blocker tho

tracer/src/Datadog.Trace/Logging/Internal/DatadogLogging.cs
andrewlock: Updated in https://github.com/DataDog/dd-trace-dotnet/pull/7084

tracer/src/Datadog.Trace/Logging/Internal/DatadogLogging.cs
andrewlock: Done in https://github.com/DataDog/dd-trace-dotnet/pull/7084

tracer/src/Datadog.Trace/Logging/Internal/DatadogLogging.cs
andrewlock: Done in https://github.com/DataDog/dd-trace-dotnet/pull/7084

andrewlock:
Also, just realised we should add some additional assertions to the data pipeline integration tests, to make sure the logging is working as expected

tracer/src/Datadog.Trace/LibDatadog/FFIVec.cs
```diff
@@ -0,0 +1,33 @@
...
+#nullable enable
+
+using System;
+using System.Runtime.InteropServices;
+using System.Text;
+
+namespace Datadog.Trace.LibDatadog;
+
+[StructLayout(LayoutKind.Sequential)]
+internal readonly struct FFIVec
```
anna-git: fwiw, class already exists [here ](https://github.com/DataDog/dd-trace-dotnet/blob/1a08eca00de790799989d1e3bc51f765198ec47a/tracer/src/Datadog.Trace/LibDatadog/VecU8.cs#L16)

tracer/src/Datadog.Trace/LibDatadog/FFIVec.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    internal readonly IntPtr Data;
+    internal readonly nuint Length;
+    internal readonly nuint Capacity;
+
+    public string ToUtf8String()
+    {
+        unsafe
+        {
+#if NETCOREAPP
```
anna-git: there's a helper [here ](https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/src/Datadog.Trace/LibDatadog/NativeStringHelpers.cs) we might want to mutualize this

tracer/src/Datadog.Trace/LibDatadog/FFIVec.cs
andrewlock: Thanks, unified them!

tracer/src/Datadog.Trace/LibDatadog/FFIVec.cs
andrewlock: Thanks, unified them!

tracer/src/Datadog.Trace/LibDatadog/Error.cs
```diff
@@ -12,23 +13,19 @@
 namespace Datadog.Trace.LibDatadog;
 
 /// <summary>
+/// Represents a generic error with message returned by the libdatadog library.
 /// Do not change the values of this enum unless you really need to update the interop mapping.
 /// Libdatadog interop mapping of https://github.com/DataDog/libdatadog/blob/60583218a8de6768f67d04fcd5bc6443f67f516b/ddcommon-ffi/src/error.rs#L14
 /// </summary>
+/// <summary>
+/// </summary>
```
bouwkast:
```suggestion
```

nit

tracer/src/Datadog.Trace/LibDatadog/ErrorHandle.cs
```diff
@@ -11,47 +11,40 @@
...
-
-    protected override bool ReleaseHandle()
+    public void ThrowIfError()
     {
-        try
-        {
-            NativeInterop.Exporter.FreeError(handle);
-        }
-        catch (Exception ex)
+        if (!IsInvalid)
```
bouwkast, line 22: Should this `!` be flipped?
```suggestion
        if (IsInvalid)
```

bouwkast: Oh wait, maybe this is saying "if the handle is valid we can then get the exception"

Wording here is a bit confusing IMO so maybe it'd be better off checking if the `handle != IntPtr.Zero` to avoid a double negative

bouwkast: Yeah it looks like that is the case, I think using `!IsInvalid` is harder to read

tracer/src/Datadog.Trace/LibDatadog/ErrorHandle.cs
```diff
@@ -11,47 +11,40 @@
...
-    {
-    }
+    public override bool IsInvalid => handle == IntPtr.Zero;
 
-    public ErrorHandle(IntPtr handle)
-        : base(handle, true)
+    public Exception ToException()
     {
-        SetHandle(handle);
+        var error = Marshal.PtrToStructure<Error>(handle);
```
bouwkast: Do we need to check if `handle` is valid before this?

tracer/src/Datadog.Trace/LibDatadog/Logger.cs
```diff
@@ -0,0 +1,101 @@
...
+        try
+        {
+            Log.Debug("Enabling libdatadog logger");
+            using var error = NativeInterop.Logger.ConfigureFile(cfg);
+            error.ThrowIfError();
+            _loggingEnabled = true;
+        }
+        catch (Exception ex)
+        {
+            Log.Error(ex, "Failed to configure libdatadog logger");
```
bouwkast, line 51:
```suggestion
            _loggingEnabled = false;
            Log.Error(ex, "Failed to configure libdatadog logger");
```

Do we need to set this back to false?
If it fails for some reason and then we call Disable it will attempt to call `Disable`

tracer/src/Datadog.Trace/LibDatadog/TraceExporterErrorHandle.cs
```diff
@@ -0,0 +1,57 @@
...
+    {
+    }
+
+    public TraceExporterErrorHandle(IntPtr handle)
+        : base(handle, true)
+    {
+        SetHandle(handle);
+    }
+
+    public override bool IsInvalid => handle == IntPtr.Zero;
```
bouwkast, line 29: Just commenting here also that going with a `IsValid` check instead of `IsInvalid` would read better if we are then going to use `!IsInvalid` in places.

tracer/src/Datadog.Trace/LibDatadog/Logger.cs
bouwkast: I see we have this comment later, so maybe a non-issue

`We assume that we can't go from pipeline enabled -> disabled, so we should never need to call logger.Disable()`

tracer/src/Datadog.Trace/LibDatadog/ErrorHandle.cs
andrewlock: I 100% completely agree, this completely confused me before too - I basically wrote the same series of comments on the original PR 😂 

However, `SafeHandle` is a Microsoft type, and `IsInvalid` is the property you have to override 🙈 So our hands are somewhat tied :sigh:

tracer/src/Datadog.Trace/LibDatadog/ErrorHandle.cs
andrewlock: Good question 🤔 I assume we should only ever call this if `!IsInvalid`, otherwise it's a programming error. Going to see if we can avoid exposing this in "raw" form

tracer/src/Datadog.Trace/LibDatadog/Logger.cs
andrewlock: Yeah, I think it's something that shouldn't occur in practice, but I think your suggestion is overall safer

tracer/src/Datadog.Trace/LibDatadog/TraceExporterErrorHandle.cs
andrewlock: _shakes fist at Microsoft's `SafeHandle`_

tracer/src/Datadog.Trace/LibDatadog/ErrorHandle.cs
andrewlock: Fixed in https://github.com/DataDog/dd-trace-dotnet/pull/7075/commits/a9577e75583d3859abff9f4c3023d0f279e0e374

tracer/src/Datadog.Trace/LibDatadog/VecU8.cs
```diff
@@ -13,11 +15,22 @@ namespace Datadog.Trace.LibDatadog;
...
-    public nuint Length; // size_t
-
-    public nuint Capacity; // size_t
+    public string ToUtf8String()
+    {
+        unsafe
+        {
+#if NETCOREAPP
+            var messageBytes = new ReadOnlySpan<byte>((void*)Ptr, (int)Length);
+            return StringEncoding.UTF8.GetString(messageBytes);
```
tonyredondo: if we are already in unsafe territory, I don't see any advantages to use `ReadOnlySpan` here. We can use the code below for both runtimes.

andrewlock: done in https://github.com/DataDog/dd-trace-dotnet/pull/7075/commits/860ab8795c6c02f3ba5fe1de33b62074bf118642

tonyredondo: cool, this method is not being called multiple times for the same struct instance right? right? because this will allocate the same string multiple times :) . If that is the case we should think on having a cache due the fact strings are inmutables.

andrewlock: _Currently_, no, it's not - it's only used for error messages etc. If that changes and it becomes an issue we could definitely consider it though 🙂

PR #7091, opened 2025-06-11
Replace fine-grained token usage with app token

andrewlock:
## Summary of changes

Replaces our current usage of a fine-grained token, with a token based on a GitHub App

## Reason for change

With recent security changes, we would have to use fine-grained tokens, and rotate them every 30 days. By using a [trusted GitHub app](https://github.com/organizations/DataDog/settings/apps/dd-trace-dotnet-ci-bot) to generate the tokens, we don't have to. Instead, we store a private key and app ID, and use _that_ to generate a short-lived token.

## Implementation details

In GitHub it's easy enough, as [there's a GitHub action](https://github.com/actions/create-github-app-token) that handles it. In Azure DevOps it's more of a pain. You need to:
- [Generate a JWT](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-json-web-token-jwt-for-a-github-app#about-json-web-tokens-jwts) using the private key and IDs associated with the app
- Make a request with the JWT to retrieve [an installation access token](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/authenticating-as-a-github-app-installation#generating-an-installation-access-token)
- Use that installation token as the bearer token for subsequent API requests

To encapsulate that flow, created an Azure DevOps step: `steps/create-github-app-token` which does all of the above, and writes the token as a secret, which can be reused in subsequent steps.

One complexity is that this retrieves private key (which contains line breaks) from an Azure Variable. tl;dr; I had a bunch of issues with this, so I stored the file with the line breaks replaced with explicit `\n` characters, which are replaced in the script to generate the "real" pem file.

## Test coverage

I did some manual testing of the GihHb actions, and I'm pretty sure they're ok. Eventually got there with Azure DevOps too so 🤞 We'll need to keep an eye out though

## Other details

Stored the secrets in 1Password, and will update the Secrets doc in confluence

### DIFF
--- a/.azure-pipelines/steps/create-github-app-token.yml
+++ b/.azure-pipelines/steps/create-github-app-token.yml
@@ -12,7 +12,8 @@ steps:
       set -o pipefail
 
       client_id="$GITHUB_APP_INSTALLATION_ID"
-      pem="$GITHUB_APP_PRIVATE_KEY"
+      pem_file=$(mktemp)
+      echo "$GITHUB_APP_PRIVATE_KEY" > "$pem_file"
 
       now=$(date +%s)
       iat=$((${now} - 60)) # Issues 60 seconds in the past
@@ -41,11 +42,12 @@ steps:
       header_payload="${header}"."${payload}"
       echo "Signing payload"
       signature=$(
-          openssl dgst -sha256 -sign <(echo -n "${pem}") \
+          openssl dgst -sha256 -sign "$pem_file" \
           <(echo -n "${header_payload}") | b64enc
       )
 
       # Create JWT
+      rm -f "$pem_file"
       JWT="${header_payload}"."${signature}"
       
       # Curl to get an installation token
--- a/.azure-pipelines/steps/create-github-app-token.yml
+++ b/.azure-pipelines/steps/create-github-app-token.yml
@@ -74,7 +74,7 @@ steps:
       echo "##vso[task.setvariable variable=GITHUB_APP_TOKEN;issecret=true]${token}"
 
     displayName: Retrieve Installation Token
-    #retryCountOnTaskFailure: 3
+    retryCountOnTaskFailure: 3
     env:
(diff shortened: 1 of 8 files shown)


.azure-pipelines/steps/create-github-app-token.yml
```diff
@@ -0,0 +1,80 @@
...
+
+      set -o pipefail
+
+      client_id="$GITHUB_APP_CLIENT_ID"
+      pem_file=$(mktemp)
+      
+      # The private key value stored contains _explicit_ '\n' characters instead of line breaks
+      # so fix those
+      pem=$(echo "$GITHUB_APP_PRIVATE_KEY" | sed 's/\\n/\n/g')
+      printf "%s" "$pem" > "$pem_file"
```
NachoEchevarria, line 15: NIT: Adding this could delete the file even if the script fails before reaching rm -f "$pem_file"

trap 'rm -f "$pem_file"' EXIT

.github/actions/deploy-aas-dev-apps/action.yml
```diff
@@ -2,18 +2,29 @@ name: 'Deploy AAS dev app'
 description: 'Deploy AAS dev app'
 
 inputs:
-  aas_github_token:
-    description: 'aas_github_token'
+  github_app_id:
+    description: 'App ID for retrieving a token from a GitHub App'
+    required: true
+  github_app_private_key:
+    description: 'App ID for retrieving a token from a GitHub App'
```
NachoEchevarria: NIT: change the description to something like: "Private key for the GitHub App"

.azure-pipelines/steps/create-github-app-token.yml
andrewlock: ooh nice, thank you!

.github/actions/deploy-aas-dev-apps/action.yml
andrewlock: Good catch 👍

PR #7094, opened 2025-06-11
[AAP] Reduce waf config diagnostics log levels

daniel-romano-DD:
## Summary of changes
Waf config diagnostics log levels were being logged as Warnings, but should be Debug level

## Reason for change
Specified in this [RFC](https://docs.google.com/document/d/1t6U7WXko_QChhoNIApn0-CRNe6SAKuiiAQIyCRPUXP4/edit?tab=t.0)

> Note that aside from through telemetry logs, none of the diagnostics presented here should be logged to a regular logger using a level above debug, as it will inevitably result in customer complaints.

### DIFF
--- a/tracer/src/Datadog.Trace/AppSec/Waf/Initialization/WafConfigurator.cs
+++ b/tracer/src/Datadog.Trace/AppSec/Waf/Initialization/WafConfigurator.cs
@@ -186,29 +186,44 @@ namespace Datadog.Trace.AppSec.Waf.Initialization
                 result.ReportedDiagnostics.Rest.Warnings is { Count: > 0 })
             {
                 var diags = result.ReportedDiagnostics;
-                var sb = StringBuilderCache.Acquire();
                 DumpStatsMessages(ref diags.Rules);
                 DumpStatsMessages(ref diags.Rest);
 
-                void DumpStatsMessages(ref WafStats stats)
+                if (diags.HasErrors)
                 {
-                    DumpMessages(stats.Errors, "ERR: ");
-                    DumpMessages(stats.Warnings, "WRN: ");
+                    // This message should go to telemetry logs only, skipping the regular logs
+                    Log.Debug("Some errors were found while applying waf configuration (RulesFile: {RulesFile})", rulesFile);
+                }
+                else
+                {
+                    Log.Debug("Some warnings were found while applying waf configuration (RulesFile: {RulesFile})", rulesFile);
                 }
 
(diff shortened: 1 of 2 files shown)


tracer/src/Datadog.Trace/AppSec/Waf/Initialization/WafConfigurator.cs
```diff
@@ -186,29 +186,44 @@ internal UpdateResult Update(IntPtr wafBuilderHandle, ConfigurationState configu
                 result.ReportedDiagnostics.Rest.Warnings is { Count: > 0 })
             {
                 var diags = result.ReportedDiagnostics;
-                var sb = StringBuilderCache.Acquire();
                 DumpStatsMessages(ref diags.Rules);
                 DumpStatsMessages(ref diags.Rest);
 
+                if (diags.HasErrors)
+                {
+                    // This message should go to telemetry logs only, skipping the regular logs
```
andrewlock: Just to make it clear that's _not_ what's happening 😅 
```suggestion
                    // TODO: This message should go to telemetry logs only, skipping the regular logs
```

tracer/src/Datadog.Trace/AppSec/Waf/Initialization/WafConfigurator.cs
```diff
@@ -186,29 +186,44 @@ internal UpdateResult Update(IntPtr wafBuilderHandle, ConfigurationState configu
...
                 {
                     if (messages is { Count: > 0 })
                     {
                         foreach (var item in messages)
                         {
-                            sb.Append($"${prefix}{item.Key}: [{string.Join(", ", item.Value)}] ");
+                            var message = $"{item.Key}: [{string.Join(", ", item.Value)}]";
+                            if (isError)
+                            {
+                                // This message should go to telemetry logs only, skipping the regular logs
```
andrewlock:
```suggestion
                                // TODO: This message should go to telemetry logs only, skipping the regular logs
```

andrewlock (approved):
Thanks!

tracer/src/Datadog.Trace/AppSec/Waf/Initialization/WafConfigurator.cs
dromanol: According to that RFC, you know...

andrewlock: Yeah, I just want to make it super clear that this is saying what _should_ happen, not what _is_ happening

PR #7102, opened 2025-06-13
[Test Optimization] Failed Test Replay

tonyredondo:
## Summary of changes

This PR enables Exception Replay on the Test Optimization product.

## Reason for change

This is a Test Optimization new feature called Failed Test Replay. When enabled all automatic test retries for a tests are executed with the Exception replay feature turned on, meaning that if the same exception happens again we will retrieve debug information of the exception stackframes.

## Implementation details

- [x] Added `DynamicInstrumentationEnabled` into test optimization settings and added logic to enable related environment variables when the feature is active. (tracer/src/Datadog.Trace.Tools.Runner/CiUtils.cs)
- [x] Added a new property `DynamicInstrumentationEnabled` and its corresponding setter method to manage the feature's configuration. (tracer/src/Datadog.Trace/Ci/Configuration/TestOptimizationSettings.cs)
- [x] Added a new interface for the Dynamic Instrumentation feature, including a method for waiting on exception instrumentation (Exception replay event). (tracer/src/Datadog.Trace/Ci/ITestOptimizationDynamicInstrumentationFeature.cs)
- [x] Updated the xUnit instrumentation: waiting to the exception instrumentation event after a test failure.
- [x] Updated the NUnit instrumentation: waiting to the exception instrumentation event after a test failure.
- [x] Updated the MsTest instrumentation: waiting to the exception instrumentation event after a test failure.   

## Test coverage

A new serie of test suite were added by every case to assert if exception replay add the tags to the span.

## Other details

DI team (@GreenMatan) it's investigating why the feature is not working on xUnit v3. That's why the test is currently disabled on this framework version.


tracer/src/Datadog.Trace/Ci/TestOptimizationDynamicInstrumentationFeature.cs
```diff
@@ -0,0 +1,74 @@
...
+    private readonly Task<ExceptionIdentifier?> _doneTask;
+    private TaskCompletionSource<ExceptionIdentifier?> _doneTaskSource = new(TaskCreationOptions.RunContinuationsAsynchronously);
+
+    public TestOptimizationDynamicInstrumentationFeature(TestOptimizationSettings settings, TestOptimizationClient.SettingsResponse clientSettingsResponse)
+    {
+        _doneTask = Task.FromResult<ExceptionIdentifier?>(null);
+        if (settings.DynamicInstrumentationEnabled == null && clientSettingsResponse.DynamicInstrumentationEnabled.HasValue)
+        {
+            Log.Information("TestOptimizationDynamicInstrumentationFeature: Dynamic instrumentation has been changed to {Value} by the settings api.", clientSettingsResponse.DynamicInstrumentationEnabled.Value);
+            settings.SetFlakyRetryEnabled(clientSettingsResponse.DynamicInstrumentationEnabled.Value);
```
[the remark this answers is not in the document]
tonyredondo: Not true, on line [31, 40] we check again the settings property to set the DynamicInstrumentationEnabled property.

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -707,6 +707,11 @@ public static class CIVisibility
             /// </summary>
             public const string TotalFlakyRetryCount = "DD_CIVISIBILITY_TOTAL_FLAKY_RETRY_COUNT";
 
+            /// <summary>
+            /// Configuration key for a kill-switch that allows to explicitly disable dynamic instrumentation even if the remote setting is enabled.
+            /// </summary>
+            public const string DynamicInstrumentationEnabled = "DD_CIVISIBILITY_DI_ENABLED";
```
andrewlock, line 713: Do we need a new key for this? Shouldn't we use the existing DI enabled keys? 🤔

tracer/src/Datadog.Trace/Ci/TestOptimization.cs
```diff
@@ -194,6 +195,20 @@ public ITestOptimizationFlakyRetryFeature? FlakyRetryFeature
         private set => _flakyRetryFeature = value;
     }
 
+    public ITestOptimizationDynamicInstrumentationFeature? DynamicInstrumentationFeature
+    {
+        get
+        {
+            if (_dynamicInstrumentationFeature is null)
+            {
+                _additionalFeaturesTask?.SafeWait();
```
andrewlock, line 204: Anything called `Safe` makes me nervous 😛

tracer/src/Datadog.Trace/Ci/TestOptimization.cs
```diff
@@ -227,7 +242,7 @@ public void Initialize()
             getDiscoveryServiceFunc: static s => DiscoveryService.Create(
                 s.TracerSettings.Exporter,
                 tcpTimeout: TimeSpan.FromSeconds(5),
-                initialRetryDelayMs: 10,
+                initialRetryDelayMs: 100,
```
andrewlock, line 245: Is this an intentional fix? I assume so, 10 ms is very low 😄

tracer/src/Datadog.Trace/Ci/TestOptimizationDynamicInstrumentationFeature.cs
```diff
@@ -0,0 +1,74 @@
...
+            ExceptionTrackManager.ExceptionCaseInstrumented += exceptionIdentifier =>
+            {
+                Log.Debug("TestOptimizationDynamicInstrumentationFeature: Exception instrumentation completed for {ExceptionIdentifier}", exceptionIdentifier);
+                var tcs = Interlocked.Exchange(ref _doneTaskSource, new(TaskCreationOptions.RunContinuationsAsynchronously));
+                tcs.TrySetResult(exceptionIdentifier);
+            };
+        }
+        else
+        {
+            settings.SetDynamicInstrumentationEnabled(false);
```
andrewlock, line 56: If this can change, do we need to potentially remove the event handler from `ExceptionTrackManager.ExceptionCaseInstrumented`?

tracer/src/Datadog.Trace/Ci/TestOptimizationTracerManagement.cs
```diff
@@ -38,6 +38,11 @@ public TestOptimizationTracerManagement(
                 if (Enum.TryParse<EventPlatformProxySupport>(settings.ForceAgentsEvpProxy, out var parsedValue))
                 {
                     EventPlatformProxySupport = parsedValue;
+                    if (settings.DynamicInstrumentationEnabled == true)
+                    {
+                        DiscoveryService = getDiscoveryServiceFunc?.Invoke(settings) ?? NullDiscoveryService.Instance;
+                        Log.Debug("TestOptimizationTracerManagement: Discovery service set to {DiscoveryServiceType}.", DiscoveryService.GetType().Name);
```
andrewlock: nit, avoid the `GetType()` if we don't need it

```suggestion
                        if (Log.IsEnabled(LogEventLevel.Debug))
                        {
                            Log.Debug("TestOptimizationTracerManagement: Discovery service set to {DiscoveryServiceType}.", DiscoveryService.GetType().Name);
                        }
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Testing/MsTestV2/TestMethodAttributeExecuteIntegration.cs
```diff
@@ -155,6 +155,15 @@ internal static CallTargetState OnMethodBegin<TTarget, TTestMethod>(TTarget inst
             }
             else if (testOptimization.FlakyRetryFeature?.Enabled == true && resultStatus == TestStatus.Fail)
             {
+                // check if is the first execution and the dynamic instrumentation feature is enabled
+                if (testOptimization.DynamicInstrumentationFeature?.Enabled == true)
+                {
+                    // let's wait for the instrumentation of an exception has been done
+                    Common.Log.Debug("TestMethodAttributeExecuteIntegration: First execution with an exception detected. Waiting for the exception instrumentation.");
+                    testOptimization.DynamicInstrumentationFeature.WaitForExceptionInstrumentation(2_000).SafeWait();
```
andrewlock: Should this `2_000` be a configurable parameter somewhere? Or at least an explicit constant`?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Testing/XUnit/XUnitIntegration.cs
```diff
@@ -209,12 +209,19 @@ internal static void FinishTest(Test test, IExceptionAggregator? exceptionAggreg
...
+                        if (testCaseMetadata.IsAttemptToFix)
+                        {
+                            testCaseMetadata.AllAttemptsPassed = false;
+                        }
                     }
 
                     WriteFinalTagsFromMetadata(test, testCaseMetadata);
+                    var span = Tracer.Instance.ActiveScope?.Span;
+                    Common.Log.Debug("XUnitIntegration: Reporting exception {ExceptionType} for test {TestName}", exception.GetType().FullName, test.Name);
+                    Common.Log.Debug("XUnitIntegration: Tracer.ActiveScope: TraceId: {TraceId}, SpanId: {SpanId}, ResourceName: {ResourceName}", span?.TraceId, span?.SpanId, span?.ResourceName);
```
andrewlock: I think you can skip all of this is debug isn't enabled.
```suggestion
                    if (Log.IsEnabled(LogEventLevel.Debug)
                    {
                        var span = Tracer.Instance.ActiveScope?.Span;
                        Common.Log.Debug("XUnitIntegration: Reporting exception {ExceptionType} for test {TestName}", exception.GetType().FullName, test.Name);
                        Common.Log.Debug("XUnitIntegration: Tracer.ActiveScope: TraceId: {TraceId}, SpanId: {SpanId}, ResourceName: {ResourceName}", span?.TraceId, span?.SpanId, span?.ResourceName);
                    }
```

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/CI/TestingFrameworkRetriesTests.cs
```diff
@@ -132,7 +133,105 @@ public virtual async Task FlakyRetries(string packageVersion)
             throw;
         }
 
-        await Task.CompletedTask;
+        return tests;
+    }
+
+    public virtual async Task FlakyRetriesWithExceptionReplay(string packageVersion)
+    {
+        SetEnvironmentVariable(ConfigurationKeys.DebugEnabled, "1");
```
andrewlock: I don't think we should enable this by default - we have a dedicated nightly run that explicitly enables this to catch related issues
```suggestion
```

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/CI/TestingFrameworkRetriesTests.cs
```diff
@@ -132,7 +133,105 @@ public virtual async Task FlakyRetries(string packageVersion)
...
+        SetEnvironmentVariable(ConfigurationKeys.CIVisibility.DynamicInstrumentationEnabled, "1");
+        SetEnvironmentVariable(ConfigurationKeys.Debugger.ExceptionReplayEnabled, "1");
```
andrewlock, line 142: Do you have to enable both of these? 🤔 It seems kind of redundant 🤷‍♂️

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/CI/TestingFrameworkRetriesTests.cs
```diff
@@ -132,7 +133,105 @@ public virtual async Task FlakyRetries(string packageVersion)
...
+        SetEnvironmentVariable(ConfigurationKeys.Debugger.RateLimitSeconds, "0");
+        SetEnvironmentVariable(ConfigurationKeys.Debugger.UploadFlushInterval, "1000");
+        SetEnvironmentVariable("DD_CLR_ENABLE_INLINING", "0");
```
andrewlock: I feel like we shouldn't need to set this? 😬 Is this just for debugging something?
```suggestion
```

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/CI/TestingFrameworkRetriesTests.cs
```diff
@@ -132,7 +133,105 @@ public virtual async Task FlakyRetries(string packageVersion)
...
+
+        var tests = await FlakyRetries(packageVersion);
```
andrewlock, line 146: I know you love your explicit checks, but should we really just have snapshots here? It's hard to spot things that should or shouldn't be there 🤷‍♂️

tracer/test/Datadog.Trace.Tests/Telemetry/config_norm_rules.json
```diff
@@ -398,6 +398,7 @@
   "DD_CIVISIBILITY_FLAKY_RETRY_COUNT": "ci_visibility_flaky_retry_count",
   "DD_CIVISIBILITY_TOTAL_FLAKY_RETRY_COUNT": "ci_visibility_total_flaky_retry_count",
   "DD_CIVISIBILITY_IMPACTED_TESTS_DETECTION_ENABLED": "dd_civisibility_impacted_tests_detection_enabled",
+  "DD_CIVISIBILITY_DI_ENABLED": "ci_visibility_dynamic_instrumentation_enabled",
```
andrewlock, line 401: Don't forget to add this into dd-go

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
tonyredondo: For dynamic instrumentation they have 4 or 5 environment variables with defaults that we might want to change on test optimization mode. Similar to the logs implementation, I want to control everything with a single environment variable to set the new defaults to those 4 or 5 environment variables.

tracer/src/Datadog.Trace/Ci/TestOptimization.cs
tonyredondo: hahaha, yeah I know, this is just a `.GetAwaiter().GetResult()` inside a block to remove the synchronization context. Maybe not the best name tbh. 

```csharp
    public static void SafeWait(this Task task)
    {
        if (task.IsCompleted)
        {
            return;
        }

        var originalContext = SynchronizationContext.Current;
        try
        {
            // Set the synchronization context to null to avoid deadlocks.
            SynchronizationContext.SetSynchronizationContext(null);

            // Wait synchronously for the task to complete.
            task.GetAwaiter().GetResult();
        }
        finally
        {
            // Restore the original synchronization context.
            SynchronizationContext.SetSynchronizationContext(originalContext);
        }
    }
```

tracer/src/Datadog.Trace/Ci/TestOptimization.cs
tonyredondo: yeah exactly that, I saw 10ms and I thought this must be an error, 10ms is just too low.

tracer/src/Datadog.Trace/Ci/TestOptimizationDynamicInstrumentationFeature.cs
tonyredondo: Yes, but the handler is only set when we set the settings to `true` we do `settings.SetDynamicInstrumentationEnabled(false)` just to make sure that the settings has the right value, because is a nullable<bool>. I hate it and all the settings needs a refactor from my side.

tracer/src/Datadog.Trace/Ci/TestOptimizationTracerManagement.cs
tonyredondo: done here [3e9d938](https://github.com/DataDog/dd-trace-dotnet/pull/7102/commits/3e9d938b5ef47d381bd3ff9c2f93bc8313d48136)

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Testing/MsTestV2/TestMethodAttributeExecuteIntegration.cs
tonyredondo: changed in [3e9d938](https://github.com/DataDog/dd-trace-dotnet/pull/7102/commits/3e9d938b5ef47d381bd3ff9c2f93bc8313d48136)

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Testing/XUnit/XUnitIntegration.cs
tonyredondo: done in [3e9d938](https://github.com/DataDog/dd-trace-dotnet/pull/7102/commits/3e9d938b5ef47d381bd3ff9c2f93bc8313d48136)

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/CI/TestingFrameworkRetriesTests.cs
tonyredondo: done in [3e9d938](https://github.com/DataDog/dd-trace-dotnet/pull/7102/commits/3e9d938b5ef47d381bd3ff9c2f93bc8313d48136)

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/CI/TestingFrameworkRetriesTests.cs
tonyredondo: done in [3e9d938](https://github.com/DataDog/dd-trace-dotnet/pull/7102/commits/3e9d938b5ef47d381bd3ff9c2f93bc8313d48136)

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/CI/TestingFrameworkRetriesTests.cs
tonyredondo: done in [3e9d938](https://github.com/DataDog/dd-trace-dotnet/pull/7102/commits/3e9d938b5ef47d381bd3ff9c2f93bc8313d48136)

tracer/src/Datadog.Trace/Ci/TestOptimization.cs
dudikeleti: `WaitWithoutContext` or `WaitIgnoreContext` 🤔

tracer/src/Datadog.Trace/Ci/Net/TestOptimizationClient.GetSettingsAsync.cs
```diff
@@ -198,11 +200,14 @@ public readonly struct SettingsResponse
         [JsonProperty("default_branch")]
         public readonly string? DefaultBranch;
 
+        [JsonProperty("di_enabled")]
+        public readonly bool? DynamicInstrumentationEnabled;
```
dudikeleti, line 204: Is it intentionally nullable instead of giving it a default value (false)?

tracer/src/Datadog.Trace/Ci/TestOptimization.cs
dudikeleti: WaitWithoutContext or WaitIgnoreContext 🤔

tracer/src/Datadog.Trace/Ci/TestOptimizationDynamicInstrumentationFeature.cs
```diff
@@ -0,0 +1,76 @@
...
+    public Task WaitForExceptionInstrumentation(int timeout)
+    {
+        if (!Enabled)
+        {
+            return _doneTask;
+        }
+
+        var dts = _doneTaskSource;
+        dts = Interlocked.CompareExchange(ref _doneTaskSource, dts, dts);
+        return Task.WhenAny(dts.Task, Task.Delay(timeout));
```
dudikeleti, line 74: @GreenMatan timeout (2000 ms) is enough?

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
dudikeleti, line 713: Maybe I'm wrong but it seems that what you want here is `ExceptionReplayEnabled` and not DI.

tracer/src/Datadog.Trace/Debugger/Upload/SnapshotUploadApi.cs
```diff
@@ -49,10 +49,12 @@ public override async Task<bool> SendBatchAsync(ArraySegment<byte> data)
 
             var request = _apiRequestFactory.Create(new Uri(uri));
 
+            Log.Debug("SnapshotUploadApi: Sending snapshots to {Uri}", uri);
```
dudikeleti, line 52: Are you using the same endpoint as the debugger? It may not matter, but from the next version of agent\tracer we have a new endpoint for snapshots. https://github.com/DataDog/dd-trace-dotnet/pull/7388

dudikeleti (approved):
LGTM. I left minor comments \ questions.

tracer/src/Datadog.Trace/Ci/Net/TestOptimizationClient.GetSettingsAsync.cs
tonyredondo: Yes currently the TestOptimization product uses nullable everywhere to represent that it needs to gather information from the backend to make the final decision (true or false). Values can be rewritten using the environment variables (no nullable) and that prevents the backend request.

I actually don't like this and I'll do a refactor of all this mechanism in the future

tracer/src/Datadog.Trace/Ci/TestOptimization.cs
tonyredondo: oh I like those names, I'll handle the renaming in a different PR though.

tracer/src/Datadog.Trace/Debugger/Upload/SnapshotUploadApi.cs
tonyredondo: Currently I'm just using what exception replay is using, I'm just adding some logs in debug mode. I need to talk with you about the new endpoint though, let's talk offline.

PR #7104, opened 2025-06-16
Fix log4net tests to account for new process span to check for Android

bouwkast:
## Summary of changes

Fixes the log4net tests for v3.1.0 which caused us to create a new process span when log4net is checking if the OS is Android.

## Reason for change

Tests were broken due to changes in https://github.com/apache/logging-log4net/releases/tag/rel%2F3.1.0 with how they detect Android.

## Implementation details

Just ignored the process span - the call is only made on non-Windows for log4net and only in v3.1.0 and above.

## Test coverage

Updated tests to exclude the process span.


tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/Log4NetTests.cs
```diff
@@ -88,8 +89,26 @@ public async Task InjectsLogsWhenEnabled(string packageVersion, bool enableLogSh
             using (var agent = EnvironmentHelper.GetMockAgent())
             using (var processResult = await RunSampleAndWaitForExit(agent, packageVersion: packageVersion))
             {
-                var spans = agent.WaitForSpans(1, 2500);
+                IImmutableList<MockSpan> spans = null;
+                if (EnvironmentTools.IsWindows())
```
andrewlock, line 93: nit: I wonder if we should extract all this bit out seeing as it's in both methods, but very low level of care

andrewlock (approved):
Excellent, thanks for handling this!

PR #7108, opened 2025-06-16
Fix protobuf instrumentation error on versions < 3.13 (before Aug 14 2020)

vandonr:
## Summary of changes

We got [an error on telemetry](https://dd.slack.com/archives/C07TBLYDQHX/p1749057343498999) that prompted me to look into the instrumentation once again. It was a ducktype error.
Turns out the error occurs on versions older that 3.13 because MapField(s) go through the instrumentation, and they are actually `System.Object`, and as such cannot be duckcasted to our IMessageProxy.
This was not directly an issue, but looking into it, I discovered an other distantly related issue impacting the same version range where instead of extracting the main message, we extracted the deepest sub-message.
This is because on older versions, the instrumented method is called recursively, and we instrument on method exit for performance reasons I'm not going to get into here. Thus, we extracted the message of the first method call to return, which was a sub-message and not the main one.

I fixed the issue by writing an alternative instrumentation for versions [3.0, 3.12.*] that instruments on method enter, thus capturing correctly the top-most message, and the other calls are caught by the "loop prevention" code:
https://github.com/DataDog/dd-trace-dotnet/blob/addcbd29e3bbee738847ebd2e5bf2c8f3f9d5bcc/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Protobuf/SchemaExtractor.cs#L50-L54

## Reason for change

broken feature for users of protobuf < 3.13 using sub-messages

## Implementation details

This older instrumentation _could_ be applied to all versions, though it has slightly worse perf because we cannot use type constraints for it. I don't know where that stands on the code complexity vs perf balance 🤷

## Test coverage

added package versions to tests and made sure we run them on v3.12

### DIFF
--- a/.azure-pipelines/ultimate-pipeline.yml
+++ b/.azure-pipelines/ultimate-pipeline.yml
@@ -119,7 +119,7 @@ variables:
   DD_LOGGER_DD_TAGS: test.configuration.job:$(System.JobDisplayName)
   DD_LOGGER_ENABLED: true
   DD_COLLECTOR_CPU_USAGE: true
-  ToolVersion: 3.19.0
+  ToolVersion: 3.20.0
   # .NET SDK performance optimization variables
   DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1
   DOTNET_CLI_TELEMETRY_OPTOUT: 1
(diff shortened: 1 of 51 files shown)


andrewlock (approved):
LGTM, thanks for this!

PR #7110, opened 2025-06-16
Add global tracing command to FleetInstaller.exe

andrewlock:
## Summary of changes

Adds the ability to globally enable the tracer with the FleetInstaller.exe file

## Reason for change

For Windows SSI, we are going to start instrumenting all .NET Core processes by default. As part of this work, we add a new command to the FleetInstaller.exe installer (which is called by the Go fleet installer). 

## Implementation details

- The actual implementation is mostly the addition of 3 new classes
  - `EnableGlobalInstrumentationCommand`
  - `RemoveGlobalInstrumentation`
  - `GlobalEnvVariableHelper`
- Most of the PR is additional smoke tests

## Test coverage

- Unit/integration tests for the `GlobalEnvVariableHelper`
- Diversified the existing IIS smoke tests to work with both the existing instrumentation command and the new one
- Added new non-iis smoke tests
- Handle issue caused by running the fleetinstaller with servicemonitor.exe.
  - Created an issue to see if we can solve this directly in servicemonitor https://github.com/microsoft/IIS.ServiceMonitor/issues/94
  - As a workaround, running servicemonitor _before_ installing the fleetinstaller
  - In practice ServiceMonitor.exe is designed for containers, not hosts, so we won't support it for now

## Other details

The actual implementation of the command is pretty simple.. it's the testing that's taken the time!


tracer/build/_build/docker/smoke.windows.fleet-installer.dockerfile
```diff
@@ -53,26 +52,7 @@ ENV DD_INTERNAL_WORKAROUND_77973_ENABLED=1
...
-RUN echo '$completedFile=\"C:\logs\completed.txt\"; if (Test-Path $completedFile) { Remove-Item $completedFile;};'  > C:\app\entrypoint.ps1; \
-    echo 'Write-Host \"Running servicemonitor to copy variables\"; Start-Process -NoNewWindow -PassThru -FilePath \"c:/ServiceMonitor.exe\" -ArgumentList @(\"w3svc\", \"AspNetCorePool\");' >> C:\app\entrypoint.ps1; \
-    echo 'Write-Host \"Waiting 10s before starting app pool\"; Start-Sleep -Seconds 10;' >> C:\app\entrypoint.ps1; \
-    echo 'Write-Host \"Starting AspNetCorePool app pool\"; Start-WebAppPool -Name \"AspNetCorePool\" -PassThru;' >> C:\app\entrypoint.ps1; \
-    echo 'Write-Host \"Making 404 request\"; curl http://localhost:5000;' >> C:\app\entrypoint.ps1; \
-    echo 'while (-not (Test-Path "C:\logs\completed.txt")) { Write-Host \"Waiting for app shutdown\"; Start-Sleep -Seconds 1; }; ' >> C:\app\entrypoint.ps1; \
-    echo 'Write-Host \"Stopping pool\";Stop-WebAppPool \"AspNetCorePool\" -PassThru;' >> C:\app\entrypoint.ps1;  \
-    echo 'Write-Host \"Shutting down\"' >> C:\app\entrypoint.ps1;
+# Copy the app across
+COPY --from=builder /src/publish /app/.
```
bouwkast: I think this is just duped

```suggestion
```

tracer/build/_build/docker/smoke.windows.iis.fleet-installer.dockerfile
bouwkast, line 1: I think this is just the old contents of that `fleet-installer.dockerfile` right? Didn't really review this much more than just looking at it

tracer/src/Datadog.FleetInstaller/Commands/EnableGlobalInstrumentationCommand.cs
```diff
@@ -0,0 +1,95 @@
...
+            log.WriteError("Failed to set global environment variables");
+            return ReturnCode.ErrorSettingGlobalEnvironmentVariables;
+        }
+
+        // We can't enable iis instrumentation if IIS is not available or it's too low of a version
+        if (!HasValidIIsVersion(out var errorMessage))
+        {
+            // nothing more to do in this case
+            log.WriteInfo("Skipping IIS instrumentation. " + errorMessage);
+            log.WriteInfo("Instrumentation complete");
```
bouwkast, line 64: Should these be errors or is this "expected" behavior?

tracer/src/Datadog.FleetInstaller/Commands/RemoveGlobalInstrumentation.cs
```diff
@@ -0,0 +1,67 @@
...
+using System.CommandLine.Invocation;
+using System.CommandLine.Parsing;
+using System.Threading.Tasks;
+
+namespace Datadog.FleetInstaller.Commands;
+
+/// <summary>
+/// Remove IIS instrumentation completely
+/// </summary>
+internal class RemoveGlobalInstrumentation : CommandBase
```
bouwkast: the most nitpicky of nitpicks but:
```suggestion
internal class RemoveGlobalInstrumentationCommand: CommandBase
```

bouwkast: Would have to also update the XML docs and ctor if accepted so not really important

tracer/src/Datadog.FleetInstaller/GlobalEnvVariableHelper.cs
```diff
@@ -0,0 +1,89 @@
...
+                log.WriteInfo($"Reverting global environment variable {kvp.Key}");
+                Environment.SetEnvironmentVariable(kvp.Key, kvp.Value, EnvironmentVariableTarget.Machine);
+            }
+
+            log.WriteInfo("All global environment variables reverted");
+            return true;
+        }
+        catch (Exception e)
+        {
+            log.WriteError("Failed revert global environment variables: " + e);
```
bouwkast:
```suggestion
            log.WriteError("Failed to revert global environment variables: " + e);
```

tracer/src/Datadog.FleetInstaller/GlobalEnvVariableHelper.cs
```diff
@@ -0,0 +1,89 @@
...
+        try
+        {
+            log.WriteInfo("Installing global environment variables");
+
+            foreach (var kvp in values.GlobalRequiredEnvVariables)
+            {
+                var previousValue = Environment.GetEnvironmentVariable(kvp.Key, EnvironmentVariableTarget.Machine);
+                if (!string.IsNullOrWhiteSpace(previousValue))
+                {
+                    previousValues[kvp.Key] = previousValue;
```
bouwkast: What happens when there is no previous value and we set it, and then we revert it?

Since the revert is just looking at the previous would the new values not be reverted? 
Unsure if that case can even really happen

bouwkast: Maybe they get removed in `RemoveMachineEnvironmentVariables`

tracer/test/Datadog.Trace.IntegrationTests/FleetInstaller/GlobalEnvVariableHelperTests.cs
```diff
@@ -0,0 +1,124 @@
...
+
+        AssertAllEnvVarValuesAreSetAtMachineLevel(values);
+
+        // Confirm that we can clear them all again
+        GlobalEnvVariableHelper.RemoveMachineEnvironmentVariables(Log);
+        AssertAllKeysAreEmpty(values);
+    }
+
+    [SkippableFact]
+    public void FullInstallAndRevert_WhenExistingVariablesPresent_RestoresPreviousValues()
```
bouwkast: Should we add a test where only some of the variables are present and then attempts to restore them?

bouwkast (approved):
Only real questions are (1) I don't really know Docker very well, but I think the docker stuff is mostly not new so I didn't do a close review and (2) I was wondering if we need to test out case where only some variables were present while others weren't when reverting.

tracer/src/Datadog.FleetInstaller/Commands/RemoveGlobalInstrumentation.cs
bouwkast: However I see other classes missing Command so 🤷

tracer/build/_build/docker/smoke.windows.iis.fleet-installer.dockerfile
andrewlock: Almost, the changes are shown in [this commit](https://github.com/DataDog/dd-trace-dotnet/pull/7110/commits/d4dba89cce811cc888e9c44a29c0253f7dfc0c8d). The main change is the addition of the `INSTALL_COMMAND` argument, and the move to running that _after_ service monitor due to https://github.com/microsoft/IIS.ServiceMonitor/issues/94

tracer/src/Datadog.FleetInstaller/Commands/EnableGlobalInstrumentationCommand.cs
andrewlock: No this is fine. It's one of the scenarios we see with customers today and is perfectly valid 🙂

tracer/src/Datadog.FleetInstaller/Commands/RemoveGlobalInstrumentation.cs
andrewlock: I'll do it in a follow up cleanup PR for clarity, thanks 🙂

tracer/src/Datadog.FleetInstaller/GlobalEnvVariableHelper.cs
andrewlock: No, good catch, I think this is a bug. I'll write a test for it, thanks 👀

tracer/src/Datadog.FleetInstaller/Commands/RemoveGlobalInstrumentation.cs
andrewlock: Actually, will just fix in this PR instead

tracer/test/Datadog.Trace.IntegrationTests/FleetInstaller/GlobalEnvVariableHelperTests.cs
andrewlock: Done in https://github.com/DataDog/dd-trace-dotnet/pull/7110/commits/546d38fa10212f962f58c2866247f8ee86a7a43a

tracer/src/Datadog.FleetInstaller/GlobalEnvVariableHelper.cs
andrewlock: Fixed in https://github.com/DataDog/dd-trace-dotnet/pull/7110/commits/546d38fa10212f962f58c2866247f8ee86a7a43a

tracer/src/Datadog.FleetInstaller/Commands/RemoveGlobalInstrumentation.cs
andrewlock: Done in https://github.com/DataDog/dd-trace-dotnet/pull/7110/commits/ea1a542d5ed94382dcd794b397125763763091ae

PR #6923, opened 2025-05-02
remove unused nuget files on publish

lucaspimentel:
## Summary of changes

Add a target to the `Datadog.Trace.Bundle` nuget package that runs after `Publish` which removes unused files.

## Implementation details

Before this PR, when using the `Datadog.Trace.Bundle` files, we add about TODO MB of files to you application:

```
> dust .\bin\Release\net8.0\win-x64\publish\datadog\ -r -b

TODO UPDATE WITH FULL BUNDLE
 64M └─┬ datadog
 13M   ├─┬ win-x64
5.9M   │ ├── datadog_profiling_ffi.dll
3.5M   │ ├── ddwaf.dll
2.7M   │ ├── Datadog.Tracer.Native.dll
1.1M   │ └── Datadog.Trace.ClrProfiler.Native.dll
 10M   ├─┬ win-x86
4.4M   │ ├── datadog_profiling_ffi.dll
2.9M   │ ├── ddwaf.dll
2.3M   │ ├── Datadog.Tracer.Native.dll
966K   │ └── Datadog.Trace.ClrProfiler.Native.dll
 10M   ├─┬ net461
7.7M   │ ├── Datadog.Trace.dll
2.7M   │ └── Datadog.Trace.pdb
 10M   ├─┬ netstandard2.0
7.5M   │ ├── Datadog.Trace.dll
2.6M   │ ├── Datadog.Trace.pdb
 48K   │ ├── System.Threading.dll
 42K   │ └── System.Diagnostics.DiagnosticSource.dll
 10M   ├─┬ net6.0
7.5M   │ ├── Datadog.Trace.dll
2.6M   │ └── Datadog.Trace.pdb
 10M   └─┬ netcoreapp3.1
7.5M     ├── Datadog.Trace.dll
2.6M     └── Datadog.Trace.pdb
```

This PR adds two files to the `Datadog.Trace.Bundle` nuget package:
- `Datadog.Trace.Bundle.targets`: adds a `RemoveDatadogNuGetFiles` target which runs after `Publish` which determines which nuget package files are not needed and deletes them from the publishes `datadog/` directory.
- `Datadog.Trace.Bundle.props`: adds a `RemoveUnusedDatadogFiles` build property (default to `true`) so users can opt-out of this feature.

For example, executing:
```
dotnet publish -f net8.0 -r win-x64
```
will output:
```
[Datadog.Trace.Bundle] TargetFramework: net8.0 => net6.0
[...]

PR #7111, opened 2025-06-16
Add new `Datadog.AzureFunctions` nuget package

lucaspimentel:
## Summary of changes

Add a new nuget package use to instrument .NET Azure Functions: `Datadog.AzureFunctions`.

## Reason for change

This package will be used when the Azure App Services extensions cannot be used.

During preview, we've been using `Datadog.Trace.Bundle`, but this new package provides some advantages:
- much smaller size footprint of published (197 MB vs 15-21 MB)
- references the required `Datadog.Serverless.Compat` nuget package which contains the agent executable
- references `Datadog.Trace` and `Datadog.Trace.Annotations` for custom instrumentation using code or annotations
  - `Datadog.Trace` also contains the trimmer descriptor xml files (`TrimmerRootDescriptor`)
- in the future, this package may contain different binaries than `Datadog.Trace.Bundle`, for example we may include ReadyToRun images to reduce cold-start times or a leaner build of the tracer that removes unused features (`dd-trace-dotnet-lite` anyone? 😆 )

## Implementation details

Add a new no-code project to pack the `nupkg` file. Use the same artifacts layout as `Datadog.Trace.Bundle`, but excluding files not used in Azure Functions.

Package layout:
```
.
└── contentFiles
    └── any
        ├── any (21 MB)
        │   └── datadog
        │       ├── linux-x64
        │       │   ├── Datadog.Trace.ClrProfiler.Native.so
        │       │   ├── Datadog.Tracer.Native.so
        │       │   └── loader.conf
        │       ├── net6.0
        │       │   └── Datadog.Trace.dll
        │       ├── win-x64
        │       │   ├── Datadog.Trace.ClrProfiler.Native.dll
        │       │   ├── Datadog.Tracer.Native.dll
        │       │   └── loader.conf
        │       └── win-x86
        │           ├── Datadog.Trace.ClrProfiler.Native.dll
        │           ├── Datadog.Tracer.Native.dll
        │           └── loader.conf
        └── net461 (15 MB)
            └── datadog
                ├── net461
                │   └── Datadog.Trace.dll
                ├── win-x64
                │   ├── Datadog.Trace.ClrProfiler.Native.dll
                │   ├── Datadog.Tracer.Native.dll
                │   └── loader.conf
                └── win-x86
                    ├── Datadog.Trace.ClrProfiler.Native.dll
                    ├── Datadog.Tracer.Native.dll
                    └── loader.conf
```

[Project type](https://learn.microsoft.com/en-us/nuget/resources/check-project-format)  | [Reference type](http://learn.microsoft.com/en-us/nuget/consume-packages/migrate-packages-config-to-package-reference) | Target runtime | [Uses files from](https://learn.microsoft.com/en-us/nuget/reference/msbuild-targets#including-content-in-a-package) | Size
--------------|--------------------------------------------|-----------------------|---------------------------|-
SDK-style     | `PackageReference`                         | .NET 6+               | `contentFiles/any/any`    | 21 MB
SDK-style     | `PackageReference`                         | .NET Framework 4.6.1+ | `contentFiles/any/net461` | 15 MB
Non-SDK-style | `PackageReference` (aka "migrated")        | .NET Framework 4.6.1+ | ❌ Not supported         |
Non-SDK-style | `packages.config`                          | .NET Framework 4.6.1+ | ❌ Not supported         |

## Test coverage

Tested manually. Will add smoke tests in separate PR.

## Other details

APMSVLS-56


tracer/build/_build/Build.cs
```diff
@@ -408,6 +408,21 @@ void DeleteReparsePoints(string path)
...
+        .Executes(() =>
+        {
+            DotNetPack(x => x
+                .SetProject(Solution.GetProject(Projects.DatadogAzureFunctions))
+                .EnableNoRestore()
+                .EnableNoDependencies()
+                .SetConfiguration(BuildConfiguration)
+                .SetNoWarnDotNetCore3()
+                .SetProperty("PackageOutputPath", ArtifactsDirectory / "nuget" / "azure-functions")
+                .SetDDEnvironmentVariables("dd-trace-dotnet-runner-tool"));
```
andrewlock: 🤔

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
```diff
@@ -0,0 +1,90 @@
...
+    <NoBuild>true</NoBuild>
+    <SourcePath Condition="$(SourcePath) == ''">$(MSBuildThisFileDirectory)../Datadog.Trace.Bundle/home</SourcePath>
+
+    <!-- nuget build properties -->
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+    <NoDefaultExcludes>true</NoDefaultExcludes>
+
+    <!-- nuget version: move $Version to $VersionPrefix so we can add a prerelease suffix -->
+    <VersionPrefix>$(Version)</VersionPrefix>
+    <VersionSuffix>preview4</VersionSuffix>
```
andrewlock: preview _4_? 🤔 
```suggestion
    <VersionSuffix>preview</VersionSuffix>
```

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
```diff
@@ -0,0 +1,90 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFrameworks>net6.0;net461</TargetFrameworks>
+    <NoBuild>true</NoBuild>
+    <SourcePath Condition="$(SourcePath) == ''">$(MSBuildThisFileDirectory)../Datadog.Trace.Bundle/home</SourcePath>
```
andrewlock, line 5: Is this the correct path? 🤔

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
```diff
@@ -0,0 +1,90 @@
...
+    <!-- reenable warnings that are disabled in tracer/src/Directory.Build.props -->
+    <NugetAudit>true</NugetAudit>
+    <CheckEolTargetFramework>true</CheckEolTargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="$(MSBuildThisFileDirectory)..\Datadog.Trace.Manual\Datadog.Trace.Manual.csproj" />
+    <ProjectReference Include="$(MSBuildThisFileDirectory)..\Datadog.Trace.Annotations\Datadog.Trace.Annotations.csproj" />
+
+    <PackageReference Include="Datadog.Serverless.Compat" Version="0.1.0" PrivateAssets="none" />
```
andrewlock: You don't want `PrivateAssets` do you? 🤔

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
```diff
@@ -0,0 +1,90 @@
...
+    <None Include="$(SourcePath)/net6.0/Datadog.Trace.dll" Pack="true" Visible="false" PackagePath="contentFiles/any/any/datadog/net6.0/">
+      <PackageCopyToOutput>true</PackageCopyToOutput>
+    </None>
+    <None Include="$(SourcePath)/net461/Datadog.Trace.dll" Pack="true" Visible="false" PackagePath="contentFiles/any/net461/datadog/net461/">
+      <Link>home/net461/Datadog.Trace.dll</Link>
+      <PackageCopyToOutput>true</PackageCopyToOutput>
+    </None>
+
+    <!-- native libraries -->
+    <None Include="$(SourcePath)/win-x64/*" Pack="true" Visible="false" PackagePath="contentFiles/any/any/datadog/win-x64/;contentFiles/any/net461/datadog/win-x64/">
```
andrewlock, line 51: You're packaging it to _both_ any _and_ `net461`? Is that necessary? 🤔

andrewlock, line 51: What about x86, Functions apps can run on x86 right?

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
```diff
@@ -0,0 +1,90 @@
...
+      <PackageCopyToOutput>true</PackageCopyToOutput>
+    </None>
+    <None Include="$(SourcePath)/linux-x64/*" Pack="true" Visible="false" PackagePath="contentFiles/any/any/datadog/linux-x64/">
```
andrewlock, line 57: No arm64 support?

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
```diff
@@ -0,0 +1,90 @@
...
+      <PackageCopyToOutput>true</PackageCopyToOutput>
+    </None>
+
+    <!-- extras -->
+    <None Include="README.md" Pack="true" Visible="true" PackagePath="/">
```
andrewlock, line 62: Do you need to pack the logo in too, or is that already packed, I can't remember 😅

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
```diff
@@ -0,0 +1,90 @@
...
+      <PackageCopyToOutput>false</PackageCopyToOutput>
+    </None>
+
+    <!-- win-x64 exclusions -->
+    <None Remove="$(SourcePath)/win-x64/Datadog.Profiler.Native.dll" />       <!-- 1.3 MB -->
+    <None Remove="$(SourcePath)/win-x64/datadog_profiling_ffi.dll" />         <!-- 5.0 MB -->
```
andrewlock: nooooooo

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
```diff
@@ -0,0 +1,90 @@
...
+      <PackageCopyToOutput>true</PackageCopyToOutput>
+    </None>
+
+    <!-- extras -->
+    <None Include="README.md" Pack="true" Visible="true" PackagePath="/">
+      <PackageCopyToOutput>false</PackageCopyToOutput>
+    </None>
+
+    <!-- win-x64 exclusions -->
+    <None Remove="$(SourcePath)/win-x64/Datadog.Profiler.Native.dll" />       <!-- 1.3 MB -->
```
andrewlock: Do we actually _not_ support profiling or ASM 🤔 Is there a reason?

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
```diff
@@ -0,0 +1,90 @@
...
+    <!-- linux-x64 exclusions -->
+    <None Remove="$(SourcePath)/linux-x64/Datadog.Linux.ApiWrapper.x64.so" /> <!-- 0.02 MB -->
+    <None Remove="$(SourcePath)/linux-x64/Datadog.Profiler.Native.so" />      <!-- 2.7  MB -->
+    <None Remove="$(SourcePath)/linux-x64/dd-dotnet" />                       <!-- 3.8  MB -->
+    <None Remove="$(SourcePath)/linux-x64/libdatadog_profiling.so" />         <!-- 8.4  MB -->
+    <None Remove="$(SourcePath)/linux-x64/libddwaf.so" />                     <!-- 2.5  MB -->
+
+    <!-- include trimming files -->
+    <None Include="$(MSBuildThisFileDirectory)../Datadog.Trace.Trimming/build/Datadog.Trace.Trimming.props" Pack="true" PackagePath="buildTransitive/net6.0/$(PackageId).props" />
+    <None Include="$(MSBuildThisFileDirectory)../Datadog.Trace.Trimming/build/Datadog.Trace.Trimming.xml" Pack="true" PackagePath="buildTransitive/net6.0/" />
```
andrewlock: These are included in Datadog.Trace so are not required I think

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
```diff
@@ -0,0 +1,90 @@
...
+
+    <!-- include empty file to resolve NU5127 -->
+    <None Include="_._" Pack="true" PackagePath="lib/net6.0/;lib/net461/" />
```
andrewlock: A better option is probably just to do no warn for `NU5127` IMO

tracer/src/Directory.Build.props
```diff
@@ -20,19 +20,23 @@
...
     <PackageTags>Datadog;APM;tracing;profiling;instrumentation</PackageTags>
-    <RepositoryType>git</RepositoryType>
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <!-- override the RepositoryUrl because PublishRepositoryUrl generates different URLs locally and on CI which makes test PublicApiHasNotChanged fail -->
     <RepositoryUrl>https://github.com/DataDog/dd-trace-dotnet.git</RepositoryUrl>
     <Copyright>Copyright 2017 Datadog, Inc.</Copyright>
     <Company>Datadog</Company>
     <Authors>Datadog</Authors>
+    <Owners>Datadog</Owners>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
```
andrewlock, line 30: I guess we can stop setting this everywhere in the sub-projects in that case?

tracer/src/Datadog.AzureFunctions/README.md
```diff
@@ -0,0 +1,12 @@
+# Datadog Instrumentation Library for .NET Azure Functions
+## `Datadog.AzureFunctions`
+
+**Note: This is an unsupported pre-release package.**
+
+**Note: If your Azure Functions are running on an App Service plan (aka Dedicated plan) on Windows, use the [Datadog Azure App Services Site Extension](https://docs.datadoghq.com/serverless/azure_app_services/azure_app_services_windows/?tab=net) instead.**
+
+Add this package to your Azure Functions project to enable Datadog APM tracing.
```
andrewlock, line 6: I think we should be very explicit about exactly which scenarios this nuget is for, and which it's _not_ for.

e.g. 
- App Service plan on Windows
- App service plan on Linux
- App service plan on Linux container
- Consumption plan
etc etc

ideally a markdown table with cross/tick or something😄

tracer/src/Datadog.AzureFunctions/remove-from-nuget-cache.ps1
andrewlock, line 1: Do we need this file? 🤨 Make it a Nuke target if you really want IMO, and purge _all_ our Datadog NuGets.

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
lucaspimentel: Oops, I've been testing. Thanks.

Do you mind if we start with `preview1` in case I need to publish a few previews?

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
lucaspimentel: We do. The default value is
```
PrivateAssets=contentfiles;analyzers;build
```
so content files are not transitive by default. We need `contentfiles` to _not_ be private assets to include the content files from `Datadog.Serverless.Compat` in apps that reference this `Datadog.AzureFunctions`.

We could use
```
PrivateAssets=analyzers;build
```
instead if you prefer, but we know `Datadog.Serverless.Compat` doesn't have those anyway, so we get the same end result.

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
lucaspimentel: When publishing the app that references this package, nuget picks one or the other, not both. If the project targets `net461`, it will copy the content files from `contentFiles/any/net461`. For any other target it will fallback to `contentFiles/any/any`.

This way, `contentFiles/any/net461` only needs to include the files for .NET Framework and Windows, keeping the published size smaller.

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
lucaspimentel: Not supported by Azure Functions yet.

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
lucaspimentel: It's already included in one of the shared props files: https://github.com/DataDog/dd-trace-dotnet/blob/d1bcc1e4b242e28a65631b6eaf5d47defd302145/tracer/src/Directory.Build.props#L35

(nuget recommends an 128x128 image, but I didn't want to add more noise to this PR)

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
andrewlock: > Do you mind if we start with preview1 in case I need to publish a few previews?

I thought we agreed that we were going to be doing the same thing as we have done for other preview packages, i.e. we have the `preview` suffix, and if we need a new package, we push a new hotfix release instead? 🤔 Generally just reduces the number of moving pieces 🤷‍♂️

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
lucaspimentel: There are several technical and product reasons. For example, these products haven't been tested extensively on serverless platforms (would they even work correctly if we enabled them?). There are also billing issues that need to be resolved.

Another issue is that none of the serverless agents support the Profiling endpoint yet. Profiling _could_ work using their agentless mode, but the team is not ready to support that yet, either.

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
lucaspimentel: In the `Datadog.Trace` package, they live in `build`, not in `buildTransitive`, so they are not transitive. (We should fix that someday.)

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
andrewlock: _oh_ I see, it's because this is describing the layout in a sort of perpendicular way 😅 I was thinking that you could not copy to `contentFiles/any/net461` because the any would _already_ include these `win-*` assets. But the point is you _can't_ (currently) target via <rid>. But you're _going to_ do that in this PR, right? 😉 🥺 🙏

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
andrewlock: 😉 https://github.com/DataDog/dd-trace-dotnet/pull/7140

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
lucaspimentel: Using empty `_._` files is the recommended solution: https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5127

The `_._` file doesn't actually get deployed, but the directory's presence allows the package to show the supported TFMs.

For example, [Datadog.Trace.Bundle](https://www.nuget.org/packages/Datadog.Trace.Bundle#supportedframeworks-body-tab) doesn't show supported frameworks, but [Datadog.Trace](https://www.nuget.org/packages/Datadog.Trace#supportedframeworks-body-tab) does because it has `lib/{tfm}` directories.

image

image

tracer/src/Directory.Build.props
lucaspimentel: Yes. I may do a clean-up of all nuget packages in a separate PR to implement some of the things I learned with this one, like the `lib/{tfm}/_._` files for the bundle and the separate `contentFiles/any/any` vs `contentFiles/any/net461` folders (or maybe finish #6923 instead of that last one).

tracer/src/Datadog.AzureFunctions/remove-from-nuget-cache.ps1
lucaspimentel: Oh no, this was for local testing only. I moved it to my local bin folder and meant to remove it from here. Thanks.

> Make it a Nuke target if you really want

yeah, that could be a good idea

tracer/build/_build/Build.cs
lucaspimentel: I just copied from `BuildBundleNuget` just above 🙈 
Should I remove `SetDDEnvironmentVariables("dd-trace-dotnet-runner-tool")`?

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
lucaspimentel: Oh yeah, this were just test versions of the nuget package I was building locally (building the package using the released 3.19 binaries, not building everything from scratch). They are preview versions of only the "package" itself (the file structure and metadata).

I'll just remove this. Realizing now this was only for local testing.

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
lucaspimentel: > But you're going to do that in this PR, right? 😉 🥺 🙏

No.

tracer/src/Datadog.AzureFunctions/README.md
lucaspimentel: Agreed.

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
lucaspimentel: Yeah...? If I understand our pipeline correctly, the `CreateBundleHome` nuke target creates this `tracer/src/Datadog.Trace.Bundle/home` folder, and there are several other targets that run after it:
```
BuildBundleNuget
BuildAzureFunctionsNuget (new in this PR)
BuildRunnerTool -> PackRunnerToolNuget
BuildStandaloneTool
```

For example, `Datadog.Trace.Tools.Runner.csproj` also includes several files from `..\Datadog.Trace.Bundle\home` when packed:
```xml
<Content Include="..\Datadog.Trace.Bundle\home\**\*.*" LinkBase="home">
```

lucaspimentel: Open to any suggestions here! I'm not an expert when it comes to our build pipelines, just trying to follow existing code for the other nugets 😅

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
lucaspimentel: I can't find any official documentation on this. Some ..uhh.. "generative" sources insist that it is not supported, but I can't find any confirmation.

lucaspimentel: Well, they do provide both x64 and x86 installers for Azure Functions Core Tools, but that doesn't  necessarily mean they support x86 at runtime?
https://github.com/Azure/azure-functions-core-tools/releases/tag/4.0.7317

lucaspimentel: I found one mention of `x86` in this section about ReadyToRun:
https://learn.microsoft.com/en-us/azure/azure-functions/dotnet-isolated-process-guide?tabs=hostbuilder%2Cwindows#readytorun

but yeah, that implies it is supported.

lucaspimentel: We're not going to support `x86` in the first release. We can add it later if needed.

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
andrewlock: > The _._ file doesn't actually get deployed, but the directory's presence allows the package to show the supported TFMs.

😮 TIL, we should probably add it to the bundle too then?

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
andrewlock: Ohhhh, right🤔

Does this _actually_ flow through in the NuGet package though? I'm kind of surprised it does anything to the final output tbh 🤔

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
lucaspimentel: Yup.

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
lucaspimentel: > Does this actually flow through in the NuGet package?

It does. In the `nuspec` generated by the SDK and included in the `nupkg`, the `dependency` element has `include="All"`:
```xml
<dependency id="Datadog.Trace" version="3.19.0" exclude="Build,Analyzers" />
<dependency id="Datadog.Trace.Annotations" version="1.0.0" exclude="Build,Analyzers" />
<dependency id="Datadog.Serverless.Compat" version="0.1.0" include="All" />
```

And I tested this manually with an app referencing only the `Datadog.AzureFunctions` package. Without changing the default `PrivateAssets` value, publishing the app copies the assembly from the `lib/` directory (`Datadog.Serverless.Compat.dll`), but not the content files (the agent executable) from `contentFiles/`.

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
lucaspimentel: I'd like to do that in a separate PR. There's enough going on here already.

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
lucaspimentel: Thanks. Since that change is released now 3.20.0, I removed these lines from `Datadog.AzureFunctions`.

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
lucaspimentel: I added a comment to explain this.

tracer/build/_build/Build.cs
andrewlock: > Should I remove SetDDEnvironmentVariables("dd-trace-dotnet-runner-tool")?

Yeah, I'm honestly not sure why that's there for either of them, seems like copy pasta 😅

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
andrewlock: Ah, ok, I see. Meh, we should probably build the bundle home to somewhere in `artifacts` but very much a problem for another day

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
```diff
@@ -0,0 +1,90 @@
...
+    <!-- nuget build properties -->
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+    <NoDefaultExcludes>true</NoDefaultExcludes>
+
+    <!-- nuget version: move $Version to $VersionPrefix so we can add a prerelease suffix.
+         This version will be published by default when we publish a new tracer release and is
+         temporary until we go GA. Using "preview1" to allow publishing multiple preview versions
+         for a given tracer release. -->
+    <VersionPrefix>$(Version)</VersionPrefix>
+    <VersionSuffix>preview1</VersionSuffix>
```
andrewlock: Using anything other than `preview` goes against all the conventions we've had previously, and how we agreed we were going to publish updates 😉 

```suggestion
    <VersionSuffix>preview</VersionSuffix>
```

andrewlock (approved):
LGTM apart from a comment on the preview suffix

tracer/src/Datadog.AzureFunctions/README.md
lucaspimentel: I started a table (WIP).

tracer/build/_build/Build.cs
lucaspimentel: `SetDDEnvironmentVariables()` sets the `DD_SERVICE_NAME` env var. I'm guessing for the CI Vis tracing?

I removed it from here, but don't want to risk breaking anything in the other targets, so I left those alone.

tracer/src/Datadog.AzureFunctions/README.md
lucaspimentel: Finished the table, and we'll do more work around documentation in both the README and the docs web site for GA. We should also update other READMEs like the bundle's to list Azure Functions under the "Do not install this package if..." list, for example.

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
lucaspimentel: This is temporary before GA. During "preview", we may need to publish multiple preview versions of the package for the _same_ tracer version. For example,
```
3.20.0-preview1 <-- tracer 3.20.0 released
3.20.0-preview2 <-- same tracer, package changes only, manual nuget publish
3.20.0-preview3 <-- same tracer, package changes only, manual nuget publish

3.21.0-preview1 <-- tracer 3.21.0 released
3.21.0-preview2 <-- same tracer, package changes only, manual nuget publish

3.21.0 (GA!)    <-- same tracer, package changes only, manual nuget publish
```

These preview releases will be handle by hand, not requiring full tracer releases for each one (that would be too slow).  For GA I'll remove these lines for `<Version>` etc and rely on that the shared version.

I'd rather have `preview1`/`preview2` than `preview`/`preview2`.

tracer/build/_build/Build.cs
andrewlock: > I'm guessing for the CI Vis tracing?

Yeah, but this is just building something, so I really don't understand 😅

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
andrewlock: > These preview releases will be handle by hand, not requiring full tracer releases for each one (that would be too slow).

This is the bit I explicitly disagree with, and therefore the conclusion. Creating a hotfix is just the click of a button, is consistent with our existing processes, and avoids a big aspect of the confusion. I'm very hesitant for us to take a different approach for this _single_ package 😉

lucaspimentel: I really don't want to create a hotfix in the _entire_ repository and all the artifacts it generates just to test a change in a single nuget package's file structure (not even the code it contains). Also, hotfix != preview?

lucaspimentel: Done!

PR #7112, opened 2025-06-16
common props

lucaspimentel:
## Reason for change

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version

> ⚠️ Warning
>
> Setting the LangVersion element to latest is discouraged. The latest setting means the installed compiler uses its latest version. The value of latest can change from machine to machine, making builds unreliable. In addition, it enables language features that might require runtime or library features not included in the current SDK.


andrewlock (commented):
Probably a silly question, but why 🤔 You can already reference Directory.Build.props outside of directory hierarchy, and we can just override any new values you don't need? 🤷‍♂️ I'm not opposed in general, just seems like it's going be Yet Another File without reducing complexity?

lucaspimentel:
> Probably a silly question, but why

Not a silly question! I was just experimenting because the default `Directory.Build.props` have lots of stuff I wasn't interested in. Pretty much the only thing I wanted to keep was `<Version>` since you consolidated it recently and that was nice.

Anyway, that was yesterday, and earlier today we discussed this on zoom, so I'm closing this. Thanks!

lucaspimentel:
(Although I do think we can make some of those things easier to opt-in/opt-out rather than applying blindly everywhere.)

PR #7084, opened 2025-06-10
Updates to libdatadog logging integration

andrewlock:
## Summary of changes

Address my comments in https://github.com/DataDog/dd-trace-dotnet/pull/7075

PR #7115, opened 2025-06-17
Minor updates for libdatadog data pipeline changes

andrewlock:
## Summary of changes

- Simplify the copying of libdatadog for tests
- Minor changes to the serilog logger to allow subsequent usage by libdatadog

## Reason for change

Prerequistes for #7075

## Implementation details

- Copy the libdatadog file to all the test output directories (maintaining a seperate list of projects to copy is redundant, and ever-increasing.
- Expose `FileLoggingConfiguration` on `DatadogSerilogLogger` (we already exposed _one_ of the properties, may as well expose it all

## Test coverage

If it builds, we're good

## Other details

Required for 
- https://github.com/DataDog/dd-trace-dotnet/pull/7075
- https://github.com/DataDog/dd-trace-dotnet/pull/7084

### DIFF
--- a/tracer/build/_build/Build.Steps.cs
+++ b/tracer/build/_build/Build.Steps.cs
@@ -657,8 +657,9 @@ partial class Build
         .Executes(() =>
         {
             // Copy the native files to all the test projects for simplicity.
-            var testProjects = Solution.GetProjects("*.Tests")
-                                       .Where(p => p.Path.ToString().EndsWith(".csproj")); // exclude native test projects
+            var testProjects = Solution.GetProjects("*Tests")
+                                       .Where(p => p.SolutionFolder.Name == "test"
+                                               &&  p.Path.ToString().EndsWith(".csproj")); // exclude native test projects
             foreach(var projectName in testProjects)
             {
                 Logger.Information("Copying native files for project {ProjectName}", projectName);


tracer/build/_build/Build.Steps.cs
```diff
@@ -652,44 +652,39 @@ async Task DownloadWafVersion(string libddwafVersion = null, string uncompressFo
...
+
+                foreach (var framework in frameworks)
                 {
-                    var source = MonitoringHomeDirectory/ "osx" / $"libdatadog_profiling.dylib";
-                    foreach (var framework in frameworks)
-                    {
-                        var dest = testBinFolder / framework / $"LibDatadog.dylib";
-                        CopyFile(source, dest, FileExistsPolicy.Overwrite);
-                    }
+                    var dest = testBinFolder / framework / $"LibDatadog.{ext}";
```
bricefriha, line 687: I like the approach but I have a question.

As we are looping through `frameworks` twice (once `frameworks.Where(x=> x == Framework)` and another time with this foreach) and we aren't using it between L673 and L684,
Wouldn't it be simpler to move the `where` predicate condition (`x == Framework` that is) and inverse it:
```suggestion
                    if  (framework != Framework) 
                    {
                         continue;
                    }
                    var dest = testBinFolder / framework / $"LibDatadog.{ext}";
```

now maybe there is something I'm missing  😄

andrewlock: Yeah, we could, but we would also need to add the `Framework is not null` case in there, so I think it's prob simpler as is tbh.

bricefriha: true I missed that sorry. 
Well in this case maybe we could simply do that 😁:
```suggestion
                    if  (Framework is not null && framework != Framework) 
                    {
                         continue;
                    }
                    var dest = testBinFolder / framework / $"LibDatadog.{ext}";
```
therefore if `Framework` is null, we use the element as is. otherwise we compare it with `Framework`'s value and skip if they are different.

I don't think it's more complicated.

*I know I don't always suggest the prettiest code* 😄

bouwkast: One other thing is that the current code reads as something such as "get the frameworks that we care about, then process them" whereas the more optimized one reads as "process all of the frameworks, then determine whether we care about the framework"

It's really just a readability thing IMO, it may be more performant but I don't think that matters too much for this code in this case.

bouwkast (approved):
👍

tracer/build/_build/Build.Steps.cs
bricefriha: I do see your point, here. 
I personally read the latter as "Go through all the frameworks and only process the one we care about". 
But if you all find it more difficult to read, go for the former as It's for the build so I think readability is more important still

PR #7116, opened 2025-06-17
Try update `ProcessHelper` to fix flakiness

andrewlock:
## Summary of changes

Updates the `ProcessHelper` used in tests to try to avoid deadlocks, and remove `[Flaky]` attribute to see if it works

## Reason for change

We often see deadlocks invoking the .NET SDK. Raised an issue about it [here](https://github.com/NuGet/Home/issues/14343), and they suggested our `ProcessHelper` implementation could cause deadlocks, so updating to the suggested version. It looks reasonable to me.

## Implementation details

- Replaced our existing version with the suggested version
- Removed the flaky attribute (to see if we see any deadlocks)

## Test coverage

Covered by existing tests


bouwkast (approved):
LGTM had to look some stuff up here as I wasn't really aware of what the new stuff was doing compared to the old way😅

PR #7117, opened 2025-06-17
Add `-ApiVersion` Nuke parameter for specific NuGet version testing

bouwkast:
## Summary of changes

Adds an `-ApiVersion` Nuke parameter so that when running integration tests locally we can just do something like `-ApiVersion "1.2.3"` to build the sample project against the "1.2.3" NuGet. This is instead of having to edit the `.csproj` directly or having to run `-TestAllPackageVersions` which takes longer.

## Reason for change

Small improvement to local developer experience so that we don't need to change NuGet versions in the `.csproj`. Unfortunately, Windows only as I couldn't figure out how to get `dotnet` commands to work with it and MSBuild worked right away.

## Implementation details

Added new parameter `ApiVersion`
Added logic to the `CompileSamples` target that will set that version when specifying a specific Sample to test.

Example usage:

` tracer/build.ps1 BuildAndRunWindowsIntegrationTests -BuildConfiguration Debug -Filter "Datadog.Trace.ClrProfiler.IntegrationTests.CosmosTests" -SampleName "Samples.CosmosDb" -Framework net8.0 -ApiVersion "3.52.0"`

This will build the `Samples.CosmosDb` project with `3.52.0` as the ApiVersion, which in turn will cause 3.52.0 to be restored instead of the default 3.18.0

`<ApiVersion Condition="'$(ApiVersion)' == ''">3.18.0</ApiVersion>`
`<PackageReference Include="Microsoft.Azure.Cosmos" Version="$(ApiVersion)" />` 

## Test coverage

Ran it locally with a few versions seems to work (I think)

## Other details

Note that `ApiVersion` doesn't necessarily equal _one_ NuGet or the NuGet that we want to target in the csproj, but that is usually the case.

Additionally, there are typically constraints around `ApiVersion` (e.g., specific .NET Framework versions). This doesn't account for those it's just a simple property set.

### DIFF
--- a/tracer/build/_build/Build.Steps.cs
+++ b/tracer/build/_build/Build.Steps.cs
@@ -1522,13 +1522,30 @@ partial class Build
               var samples = GetSamplesToBuild();
               Logger.Information("Building {SampleName}", samples);
 
-              // TODO: set Samples.Trimming as don't build, as we have to explicitly build that on every platform anyway
-              DotNetBuild(config => config
-                                   .SetConfiguration(BuildConfiguration)
-                                   .SetProperty("BuildInParallel", "true")
-                                   .SetProcessArgumentConfigurator(arg => arg.Add("/nowarn:NU1701"))
-                                   .When(Framework is not null, x => x.SetFramework(Framework))
-                                   .SetProjectFile(samples));
+              // If we are building a specific sample, with a specific NuGet version, we can target just that one with MSBuild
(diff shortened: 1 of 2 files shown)


andrewlock (approved):
Thanks, it sucks that dotnet build doesn't work as expected with the properites 🙁

link04 (approved):
TY

PR #7118, opened 2025-06-17
Fix AWS S3 DeleteBucket integration

andrewlock:
## Summary of changes

Fixes the broken integration

## Reason for change

4.0.1.7 broke our integration, this fixes it 

## Implementation details

The backing field name changed, which broke the integration. Use the public property instead of the field, given it's basically just an auto-property anyway

## Test coverage

Did a full test (all specific versions [here](https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=179662&view=results)) and all expected tests passed


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/S3/BucketManagement/IDeleteBucketRequest.cs
```diff
@@ -17,6 +17,6 @@ internal interface IDeleteBucketRequest : IDuckType
     /// <summary>
     /// Gets the S3 bucket name.
     /// </summary>
-    [DuckField(Name = "bucketName")]
```
andrewlock, line 20: This is the fix

bouwkast (approved):
Thanks

PR #7119, opened 2025-06-18
Create a `StringUtil` helper for correct nullable annotations on `string.IsNullOrEmpty`

andrewlock:
## Summary of changes

Instead of using `string.IsNullOrEmpty()` of `string.IsNullOrWhitespace()` use `StringUtil.IsNullOrEmpty()` etc.

## Reason for change

`string.IsNullOrEmpty()` _doesn't_ have the correctly nullable annotations for .NET FX and .NET Standard. That means that the flow analysis often says a string variable _could_ be null, when it really  couldn't. That in turn means you have to use `!` all over the place, which in turn makes it harder to spot when we're _actually_ making nullability assumptions vs. the cases where we're just trying to shut the compiler up. In other places, we actively avoid the method to avoid the noise. `StringUtil.IsNullOrEmpty` is simply a correctly annotated version which delegates to the real implementation.

## Implementation details

`StringUtil` is a thin wrapper around the `string` utilities, adding the correct annotations. I "fixed" a few places that had this problem as an example, but theoretically we should just be able to replace all usages.

## Test coverage

Added unit tests to confirm behaviour is unchanged, and that the warning is gone


bouwkast (approved):
Awesome thanks!

PR #7122, opened 2025-06-19
Update service discovery check

andrewlock:
## Summary of changes

- Refactor service discovery check to make it testable
- Add some additional debug logs
- Bail out when libdatadog is not available or PInvoking is not supported

## Reason for change

We recently [noticed](https://app.datadoghq.com/error-tracking/unified?query=source%3Adotnet+issue.age%3A%3C%3D3600000+-System.OutOfMemoryException+service%3Ainstrumentation-telemetry-data+version%3A%3E%3D3.0.0.0&issueId=c610d57a-4ab5-11f0-84cd-da7ad0900002&from_ts=1749995650000&to_ts=1750082050000&live=false&monitor_id=156515650&monitor_sub_type=.new%28%29&link_source=monitor_notif) that the service discovery supported introduced in #7014 was failing when running `dd-trace ci run` in CI Visbility scenarios.

This is because the `dd-trace` tool creates a `Tracer` instance, so that it can create a `TestSession` span, and propagate the context to the testhost. With the recent discovery service work, the `dd-trace` process therefore tries to load the libdatadog library, but the call fails, as we haven't rewritten the PInvokes (because this isn't an _instrumented_ process, we're effectively doing "Manual" instrumentation). 

## Implementation details

As a workaround, make sure we check to see if we're in an auto-instrumented process by checking that the native tracer dll is loaded. This should give a reliable determination for this scenario. we can also cache the check for use anywhere else we need to use libdatadog.

I put the `IsLibDatadogAvailable` property on the `Libdatadog.NativeInterop` type, to try to highlight that it needs checking any time you're assuming libdatadog is there, but I'm open to moving it.

Otherwise, mostly refactored to make the code easier to test.

## Test coverage

Added a couple of tests to make sure we _skip_ the instrumentation (instead of erroring, as we would have done before the final fix commit). 

The positive case (i.e. that the feature works) is already tested in system tests.

## Other details

We don't have tests for the `dd-trace ci run` commands yet - we should look to add them, but that's out of the scope for this PR. We should do a manual test instead


tracer/src/Datadog.Trace/LibDatadog/ServiceDiscovery/Utils.cs
andrewlock, line 1: This file moved to `ServiceDiscoveryHelpers` - when the namespaces are removed `Utils` was too generic for my liking 😄

tracer/src/Datadog.Trace/TracerManager.cs
```diff
@@ -691,22 +691,8 @@ private static void OneTimeSetup(TracerSettings tracerSettings)
             // start the heartbeat loop
             _heartbeatTimer = new Timer(HeartbeatCallback, state: null, dueTime: TimeSpan.Zero, period: TimeSpan.FromMinutes(1));
 
-            if (FrameworkDescription.Instance.OSPlatform == OSPlatformName.Linux && Environment.Is64BitProcess && !Util.EnvironmentHelpers.IsServerlessEnvironment())
```
andrewlock, line 694: This code was moved as-is into `ServiceDiscoveryHelpers` and then refactored to make it easier to test

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/ServiceDiscovery/ServiceDiscoveryHelperTests.cs
```diff
@@ -0,0 +1,41 @@
...
+using Datadog.Trace.LibDatadog.ServiceDiscovery;
+using Datadog.Trace.TestHelpers;
+using FluentAssertions;
+using Xunit;
+
+namespace Datadog.Trace.ClrProfiler.IntegrationTests.ServiceDiscovery;
+
+public class ServiceDiscoveryHelperTests
+{
+    [Fact]
```
andrewlock:
```suggestion
    [SkippableFact]
```

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/ServiceDiscovery/ServiceDiscoveryHelperTests.cs
```diff
@@ -0,0 +1,41 @@
...
+        // We do store metadata for linux - so this test doesn't apply
+        SkipOn.Platform(SkipOn.PlatformValue.Linux);
+
+        var tracerSettings = TracerSettings.Create([]);
+        var result = ServiceDiscoveryHelper.StoreTracerMetadata(tracerSettings);
+
+        result.Should().Be(ServiceDiscoveryHelper.StoreMetadataResult.Skipped);
+    }
+
+    [Fact]
```
andrewlock:
```suggestion
    [SkippableFact]
```

anna-git (approved):

tracer/src/Datadog.Trace/LibDatadog/ServiceDiscovery/ServiceDiscoveryHelper.cs
andrewlock, line 1: Most of this code is just refactoring, moved from elsewhere to make it easier to test

tracer/src/Datadog.Trace/LibDatadog/ServiceDiscovery/ServiceDiscoveryHelper.cs
```diff
@@ -0,0 +1,87 @@
...
+        Success,
+        Skipped,
+        Error,
+        Exception,
+    }
+
+    internal static StoreMetadataResult StoreTracerMetadata(TracerSettings tracerSettings)
+    {
+        var platformIsSupported = FrameworkDescription.Instance.OSPlatform == OSPlatformName.Linux && Environment.Is64BitProcess;
+        var deploymentIsSupported = LibDatadog.NativeInterop.IsLibDatadogAvailable;
```
andrewlock, line 28: This is the real fix, checking `LibDatadog.NativeInterop.IsLibDatadogAvailable` instead of purely for serverless env vars

tonyredondo (approved):
LGTM, just a nit not for this PR, `CharSlice` can be optimized, currently is allocating in both managed head and unmanaged heap just for interop, we can use `stackalloc` when is supported or an array buffer in old frameworks.

PR #7125, opened 2025-06-20
Avoid dlclose on glibc 2.34-2.36

andrewlock:

### DIFF
--- a/shared/src/native-src/dynamic_library_base.cpp
+++ b/shared/src/native-src/dynamic_library_base.cpp
@@ -56,6 +56,83 @@ void* DynamicLibraryBase::GetFunction(const std::string& funcName)
 #endif
 }
 
+#if LINUX
+/// \brief Checks if the current system has a buggy implementation of dlclose (glibc 2.34-2.36).
+/// \return True if the system is affected by the buggy dlclose, false otherwise.
+bool DynamicLibraryBase::HasBuggyDlclose() const
+{
+    // In certain versions of glibc, there is a TLS-reuse bug that can cause crashes when unloading shared libraries.
+    // The bug was introduced in 2.34, fixed in 2.36 on x86-64, and fixed in 2.37 on aarch64.
+    // See https://sourceware.org/git/gitweb.cgi?p=glibc.git;h=3921c5b40f293c57cb326f58713c924b0662ef59
+    // Explanation in Fedora where we spotted it: https://bugzilla.redhat.com/show_bug.cgi?id=2251557
+    //
+    // 2.34 shipped with a regression: after a dlclose() of a library that carried dynamic-TLS, the loader could reuse
+    // the same “module-ID” for a different library without first clearing the associated DTV (Dynamic Thread Vector)
+    // entry. The next time any code accessed that TLS slot it could read or write an unmapped address → SIGSEGV.
+    //
+    // This manifested as a crash in the WAF when we called `ddwaf_context_info` on arm64. It explicitly happens
+    // on arm64 when we unload the continuous profiler (because it's not supported).
+    //
+    // It manifests in this scenario, because ddwaf_context_init starts like this:
+    //   +128  bl   __tls_get_addr     ; ask glibc for the TLS slot for libddwaf
+    //   +136  mrs  x11, TPIDR_EL0     ; TLS base for this thread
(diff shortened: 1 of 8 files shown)


shared/src/native-src/dynamic_library_base.cpp
```diff
@@ -104,6 +104,18 @@ bool DynamicLibraryBase::Unload()
...
 #else
+#if LINUX
+    auto [is_buggy, glibc_version] = ::shared::HasBuggyDlclose();
+    if (is_buggy)
+    {
+        _logger->Warn("Unload: Skipping unload of dynamic library '", _filePath,
+                      "' due to buggy dlclose implementation on this system.",
+                      "GLIBC version 2.34-2.36 has a TLS-reuse bug that can cause crashes when unloading"
+                      " shared libraries. Consider updating the installed version of glibc. Found GLIBC version: ", glibc_version);
+        return false;
```
tonyredondo, line 115: what's the behaviour of the loader when we return false (failure) here?

shared/src/native-src/util.cpp
```diff
@@ -303,4 +306,80 @@ namespace shared
...
+            const auto is_buggy = (version == "2.34" || version == "2.35" || version == "2.36");
+#else
+            const auto is_buggy = (version == "2.34" || version == "2.35");
+#endif
+
+            if (!is_buggy) {
+                // Not buggy, so we can close the handle
+                dlclose(handle);
+                return std::make_tuple(false, ::shared::ToWSTRING(version));
+            }
```
tonyredondo, line 378: hahaha

tonyredondo (approved):
Great, it looks correct. One thing missing is reporting the `DD_INTERNAL_GLIBC_VERSION` value to the telemetry or similar.

Another thing is, I guess we behave just fine in cases where this is used: https://github.com/Stantheman/gcompat 
If not I guess it doesn't matter because we actually releases something for musl, we can just say we don't support that kind of stuff and use the musl version of the library.

shared/src/native-src/dynamic_library_base.cpp
andrewlock: AFAICT, we currently only call `Unload()` here: https://github.com/DataDog/dd-trace-dotnet/blob/e26708c1bc226bc20eccb0c4532db0442cb99f1c/shared/src/Datadog.Trace.ClrProfiler.Native/dynamic_instance.cpp#L30

And we ignore the return value there so it's fine 🤷‍♂️

tonyredondo: should we at least start logging this value, or we just don't care?

andrewlock:
> One thing missing is reporting the `DD_INTERNAL_GLIBC_VERSION` value to the telemetry or similar.

Currently, we're only doing the work to try to grab the glibc_version _if_ we need to call `dlclose`. That means the variable _might_ only be set at an arbitrary point, and will make reporting it in telemetry unreliable.

Do you think it's worth us doing the work upfront as part of the native loader's load path to make sure this variable is set, so that we get reliable reporting? Is the overhead worth it? 🤔 


> Another thing is, I guess we behave just fine in cases where this is used: https://github.com/Stantheman/gcompat If not I guess it doesn't matter because we actually releases something for musl, we can just say we don't support that kind of stuff and use the musl version of the library.

Yeah, we _essentially_ just don't support this at the moment. If you're running on musl, we'll load the musl binaries, but then we'll _also_ hit the glibc path in this PR, which means we will incorrectly not close some libraries. I don't _think_ that's a bigger enough thing to worry about, and if we want to change the behaviour, we _could_ potentially move the glibc detection to be part of the `IsAlpine()` call, but I don't think we need to worry about it for now

shared/src/native-src/dynamic_library_base.cpp
andrewlock: 🤷‍♂️ I don't know tbh, _I_ don't care 😂 As it's an existing issue, I'm inclined to punt on that question for a separate PR

gleocadie (approved):
Overall looks good.
I'll look into the profiler for dlclose
The ApiWrapper does not call `dlclose` just wrap it

daniel-romano-DD (approved):
Great

PR #7127, opened 2025-06-23
Improve CharSlice allocations

tonyredondo:
## Summary of changes

This PR reduces the number of allocation of the CharSlice type

## Reason for change

The current code allocates in both Managed and Unmanaged heap for each instance, we can skip the managed allocation.

## Test coverage

The struct is used in several parts of the code, so current tests covers the change.


tracer/src/Datadog.Trace/LibDatadog/CharSlice.cs
```diff
@@ -42,11 +43,16 @@ internal CharSlice(string? str)
         }
         else
         {
-            // copy over str to unmanaged memory
-            var bytes = System.Text.Encoding.UTF8.GetBytes(str);
-            Ptr = Marshal.AllocHGlobal(bytes.Length);
-            Marshal.Copy(bytes, 0, Ptr, bytes.Length);
-            Len = (nuint)bytes.Length;
+            var encoding = Encoding.UTF8;
```
andrewlock, line 47: I wonder, should we be using `StringEncoding.UTF8` here so that we don't emit the BOM 🤔

andrewlock (approved):
Awesome, thanks!

tonyredondo:
## Benchmarks 

```
// * Summary *

BenchmarkDotNet=v0.13.5, OS=macOS 15.5 (24F74) [Darwin 24.5.0]
Apple M2 Max, 1 CPU, 12 logical and 12 physical cores
.NET SDK=9.0.203
  [Host]     : .NET 9.0.4 (9.0.425.16305), Arm64 RyuJIT AdvSIMD
  DefaultJob : .NET 9.0.4 (9.0.425.16305), Arm64 RyuJIT AdvSIMD


|                     Method |     Mean |   Error |  StdDev |   Gen0 | Allocated |
|--------------------------- |---------:|--------:|--------:|-------:|----------:|
|          OriginalCharSlice | 491.2 us | 6.72 us | 5.96 us | 4.8828 |  640000 B |
|         OptimizedCharSlice | 236.8 us | 1.88 us | 1.66 us |      - |         - |
| OptimizedCharSliceWithPool | 109.9 us | 0.45 us | 0.37 us |      - |         - |

// * Hints *
Outliers
  CharSliceBenchmark.OriginalCharSlice: Default          -> 1 outlier  was  removed (666.39 us)
  CharSliceBenchmark.OptimizedCharSlice: Default         -> 1 outlier  was  removed (248.48 us)
  CharSliceBenchmark.OptimizedCharSliceWithPool: Default -> 2 outliers were removed (111.17 us, 114.81 us)

// * Legends *
  Mean      : Arithmetic mean of all measurements
  Error     : Half of 99.9% confidence interval
  StdDev    : Standard deviation of all measurements
  Gen0      : GC Generation 0 collects per 1000 operations
  Allocated : Allocated memory per single operation (managed only, inclusive, 1KB = 1024B)
  1 us      : 1 Microsecond (0.000001 sec)

```

This PR introduces `OptimizedCharSlice` as the default implementation (`OptimizedCharSliceWithPool` will be enabled later).

andrewlock (approved):
Thanks!

PR #7128, opened 2025-06-23
Disable data pipeline if libdatadog is not available

andrewlock:
## Summary of changes

Disabled data pipeline if libdatadog is not available

## Reason for change

Libdatadog is not available in all scenarios e.g. serverless scenarios, `dd-trace ci run` ([as in this recent issue](https://github.com/DataDog/dd-trace-dotnet/pull/7122)). We should not enable data pipeline in these scenarios

## Implementation details

Do not allow enabling data pipeline if libdatadog isn't available

## Test coverage

We have a unit test for IsLibDatadogAvailable, I don't think we need tests for the value here

## Other details
Fixes some other log messages to remove structured logging variables as not really necessary.

### DIFF
--- a/tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
+++ b/tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
@@ -427,10 +427,7 @@ namespace Datadog.Trace.Configuration
                 {
                     DataPipelineEnabled = false;
                     Log.Warning(
-                        "{ConfigurationKey} is enabled, but {StatsComputationEnabled} is enabled. Disabling {TraceDataPipelineEnabled}.",
-                        ConfigurationKeys.TraceDataPipelineEnabled,
-                        ConfigurationKeys.StatsComputationEnabled,
-                        ConfigurationKeys.TraceDataPipelineEnabled);
+                        $"{ConfigurationKeys.TraceDataPipelineEnabled} is enabled, but {ConfigurationKeys.StatsComputationEnabled} is enabled. Disabling data pipeline.");
                     _telemetry.Record(ConfigurationKeys.TraceDataPipelineEnabled, false, ConfigurationOrigins.Calculated);
                 }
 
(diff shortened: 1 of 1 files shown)


ganeshnj (approved):

ganeshnj:
Not related to this PR but libdatadog available check isn't really checking the libdatadog presence rather checking features that controls the presence of libdatadog. Theoretically it is possible that we may not have it on a supported feature as well like AAS.

No blocking though.

andrewlock:
> Not related to this PR but libdatadog available check isn't really checking the libdatadog presence rather checking features that controls the presence of libdatadog. Theoretically it is possible that we may not have it on a supported feature as well like AAS.

The intention is that the description of `IsLibdatadogAvailable` being available is correct. If we _know_ it isn't available (as in some serverless environments, or when auto-instrumentation is not enabled) then we short-circuit. To be certain, we could/should PInvoke into libdatadog itself. and I'm not averse to that in general.

The main reason to _not_ do that is currently there's no other scenario we think it _should_ happen, and we want to find out about it if it does! Down the line we could def switch this logic up though 🙂

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -440,9 +437,15 @@ _ when x.ToBoolean() is { } boolean => boolean,
...
-                        ConfigurationKeys.TraceDataPipelineEnabled);
+                        $"{ConfigurationKeys.TraceDataPipelineEnabled} is enabled, but TracesTransport is set to UnixDomainSocket which is not supported on Windows. Disabling data pipeline.");
+                    _telemetry.Record(ConfigurationKeys.TraceDataPipelineEnabled, false, ConfigurationOrigins.Calculated);
+                }
+
+                if (!LibDatadog.NativeInterop.IsLibDatadogAvailable)
+                {
+                    DataPipelineEnabled = false;
+                    Log.Warning(
+                        $"{ConfigurationKeys.TraceDataPipelineEnabled} is enabled, but Libddatadog is not available. Disabling data pipeline.");
```
lucaspimentel: Typo in `Libddatadog` (double `d`). Also, do we want to capitalize as `LibDatadog` or `libdatadog`? (level of care low)

andrewlock:
```suggestion
                        $"{ConfigurationKeys.TraceDataPipelineEnabled} is enabled, but libdatadog is not available. Disabling data pipeline.");
```

PR #7130, opened 2025-06-23
Add force meta struct config value

NachoEchevarria:

### DIFF
--- a/tracer/src/Datadog.Trace/AppSec/Security.cs
+++ b/tracer/src/Datadog.Trace/AppSec/Security.cs
@@ -629,6 +629,11 @@ namespace Datadog.Trace.AppSec
 
         internal bool IsMetaStructSupported()
         {
+            if (Tracer.Instance.Settings.ForceMetaStruct)
+            {
+                return true;
+            }
+
             if (_discoveryService is null)
             {
                 _discoveryService = Tracer.Instance.TracerManager.DiscoveryService;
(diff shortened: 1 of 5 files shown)


tracer/test/Datadog.Trace.TestHelpers/MockTracerAgent.cs
```diff
@@ -120,6 +122,23 @@ public static TcpUdpAgent Create(ITestOutputHelper output, int? port = null, int
 
         public static NamedPipeAgent Create(ITestOutputHelper output, WindowsPipesConfig config, AgentConfiguration agentConfiguration = null) => new NamedPipeAgent(config) { Output = output, Configuration = agentConfiguration ?? new() };
 
+        public bool WaitForConfigSent(int timeoutInMilliseconds = 10000)
```
bricefriha: I'm just wandering, why is `MockTracerAgent.WaitForConfigSent` not an asynchronous method? 
Because it would be handy to be one when calling it in `SendRequestsAsync`, right? 🙂

NachoEchevarria: The thing is that SendRequestsAsync is not actually async even though the name suggest that. Of course, this is weird. I have updated the code. Thanks for your observation.

NachoEchevarria:
Thanks for your feedback!

PR #7131, opened 2025-06-23
Add support for NLog 6.0

andrewlock:
## Summary of changes

Adds support for NLog 6.x.x

## Reason for change

There was a major version, and we need to update our integrations

## Implementation details

There's been some significant refactoring, which meant that we needed new instrumentation points. 
Luckily the refactoring simplifies some things:
- `LogFactory.ActivateLoggingConfiguration()` is called with a non-null config, when config is activated. 
- `LoggingConfigurationFileLoader.Load()` is called to load the config. This can return `null`, and if it does, we must intercept the null and return a new non-null instance when logs injection or direct logs injection is enabled. 

## Test coverage

- Expanded testing to include 6.x.x. 
- Updated default tested version on Windows to 6.x.x
- A slight tweak to the regex used for validating injection was required, because the json written no longer contains spaces.

## Other details

I've started an "all tfms, all package versions" run here: https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=180796&view=results (serverless tests failed in that one, but that's fine, they were expected to, I didn't build the sample)

### DIFF
--- a/tracer/test/test-applications/integrations/LogsInjection.NLog/LogsInjection.NLog.csproj
+++ b/tracer/test/test-applications/integrations/LogsInjection.NLog/LogsInjection.NLog.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <ApiVersion Condition="'$(ApiVersion)' == '' AND $(TargetFramework.StartsWith('net4')) ">2.1.0</ApiVersion>
-    <ApiVersion Condition="'$(ApiVersion)' == '' AND !$(TargetFramework.StartsWith('net4')) ">5.0.0</ApiVersion>
+    <ApiVersion Condition="'$(ApiVersion)' == '' AND !$(TargetFramework.StartsWith('net4')) ">6.0.0</ApiVersion>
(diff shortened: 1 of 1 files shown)


bouwkast (commented):
Running this locally it appears that the Regex for validating the logs injection needs to be updated to 

```csharp
RegexFormat = @"""{0}"":\s*{1}"
```

from

```csharp
RegexFormat = @"""{0}"": {1}"
```

as the dots in the injected JSON have changed, which seems strange, unsure if it is the regex validation logic here or an actual change.

Then the final tests that are failing after that all have to do with the `None` configuration type when logs injection is enabled but no configuration object within NLog was created, but haven't dug too deep into that

snakefoot:
> @bouwkast the dots in the injected JSON have changed, which seems strange.

This could be caused by the breaking change [NLog v6 - JsonLayout SuppressSpaces default true](https://nlog-project.org/2025/04/29/nlog-6-0-major-changes.html#nlog-jsonlayout-suppressspaces-default-true).

But normal message-formatting should continue to emit spaces:
```csharp
logger.Info("{@user}", new { Name = "Bob", UserId = 42 });
```

zacharycmontoya (approved):
LGTM

link04 (approved):

PR #7133, opened 2025-06-23
Adding Throughput Test for Kafka & DSM

robcarlan-datadog:
## Summary of changes
We want to enable Data Streams Monitoring by default. To ensure that the DSM code path has a minimal impact on customers, we've added throughput testing to ensure that the overhead of DSM is acceptable in various scenarios.

This PR adds throughput testing for publishing and consuming N Kafka messages. We compare the runtime of DSM with APM and APM without DSM enabled to ensure we have an acceptable performance overhead.

If the overhead of DSM is >= the given threshold (currently 4%), we fail the CI check.

See [here](https://gitlab.ddbuild.io/DataDog/apm-reliability/dd-trace-dotnet/-/jobs/1017470789) for a successful run.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -222,7 +222,6 @@ microbenchmarks:
 
 dsm_throughput:
   stage: benchmarks
-  needs: [ ]
   trigger:
     include: .gitlab/benchmarks/dsm-throughput.yml
   allow_failure: true
--- a/.gitlab/benchmarks/dsm-throughput.yml
+++ b/.gitlab/benchmarks/dsm-throughput.yml
@@ -1,14 +1,10 @@
-.setup:
-  script:
-    - mkdir -p ~/.aws
-    - /app/infra/tools/fetch-ssm-parameter.sh $AWS_EPHEMERAL_INFRA_PROFILE_SSM_PARAMETER > ~/.aws/config || exit $?
-    - export AWS_PROFILE=ephemeral-infra-ci
-    - export BP_INFRA_KEY_PAIR_NAME=$(cat ~/.aws/key-pair-name.txt)
-    - export BP_INFRA_KEY_PAIR_PRIVATE_KEY_PATH=~/.aws/key-pair-private-key.pem
-
 variables:
   DSM_THROUGHPUT_CI_IMAGE: 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/benchmarking-platform:dd-trace-dotnet-dsm
   MACROBENCHMARKS_CI_IMAGE: 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/benchmarking-platform:dotnet-throughput-3
+
+stages:
+  - check-azure-pipeline
+  - benchmarks
 
 check_azure_pipeline:
   stage: check-azure-pipeline
@@ -30,7 +26,6 @@ check_azure_pipeline:
 
 dsm_throughput:
   stage: benchmarks
-  when: on_success
   tags: ["runner:apm-k8s-same-cpu"]
   needs: ["check_azure_pipeline"]
   image: $DSM_THROUGHPUT_CI_IMAGE


.gitlab/benchmarks/macrobenchmarks.yml
```diff
@@ -83,7 +112,11 @@ update-bp-infra:
       - api_failure
       - script_failure
   rules:
-    - when: on_success
+    - if: '$CI_PIPELINE_SOURCE == "push" && $CI_COMMIT_BRANCH == "master"'
```
andrewlock: Do we need to change these? We I believe we want to run these on all PRs too etc, and only on sucess? 🤔

.gitlab/benchmarks/macrobenchmarks.yml
```diff
@@ -33,6 +34,34 @@ check_azure_pipeline:
   rules:
     - when: on_success
 
+throughput:
```
andrewlock: nit: These are DSM specific, so we should probably rename them to this?
```suggestion
dsm_throughput:
```

Or something similar?

.gitlab-ci.yml
```diff
@@ -207,11 +207,7 @@ macrobenchmarks:
     include: .gitlab/benchmarks/macrobenchmarks.yml
   allow_failure: true
   rules:
-    - if: '$CI_PIPELINE_SOURCE == "push" && $CI_COMMIT_BRANCH == "master"'
-      when: always
-    - if: '$CI_PIPELINE_SOURCE == "schedule" && $BENCHMARK_RUN == "true"'
-      when: always
-    - when: manual
+    - when: always
```
andrewlock: If this is temporary for testing that's all good, but I don't know if the benchmarking platform will be ok with this change in general, as it may cause too much load 🤔

.gitlab/benchmarks/macrobenchmarks.yml
```diff
@@ -266,7 +294,11 @@ profiler_cpu_timer_create:
       - api_failure
       - script_failure
   rules:
-    - when: on_success
+    - if: '$CI_PIPELINE_SOURCE == "push" && $CI_COMMIT_BRANCH == "master"'
+      when: always
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $BENCHMARK_RUN == "true"'
+      when: always
+    - when: manual
```
andrewlock: 😅 I'm confused, these two _stop_ running on every push?

tracer/test/test-applications/throughput/Samples.KafkaBenchmark/Program.cs
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        private static void RunBenchmark()
+        {
+            var producerConfig = new ProducerConfig
+            {
+                BootstrapServers = BootstrapServers,
+                EnableDeliveryReports = true,
+                Acks = Acks.All,
+                MessageSendMaxRetries = 3,
```
andrewlock: I wonder, do we want retries for this test? 🤔 I don't know if it really matters or not, just thinking out loud 😄

tracer/test/test-applications/throughput/Samples.KafkaBenchmark/Program.cs
```diff
@@ -0,0 +1,171 @@
...
+            try
+            {
+                Console.WriteLine("Starting Kafka benchmark...");
+                
+                var config = new ClientConfig
+                {
+                    BootstrapServers = BootstrapServers
+                };
+
+                CreateTopicIfNotExists(Topic, config).GetAwaiter().GetResult();
```
andrewlock: Should use async await here 
```suggestion
                await CreateTopicIfNotExists(Topic, config);
```

tracer/test/test-applications/throughput/Samples.KafkaBenchmark/Program.cs
```diff
@@ -0,0 +1,171 @@
...
+{
+    class Program
+    {
+        private const string BootstrapServers = "localhost:9092";
+        private const string ConsumerGroupId = "benchmark-consumer-group";
+        private const int MessageSize = 32;
+        private static string Topic;
+        private static int MessageCount;
+
+        static void Main(string[] args)
```
andrewlock: If you're using async await (you should!) you'll need to make this async: 

```suggestion
        static async Task Main(string[] args)
```

tracer/test/test-applications/throughput/Samples.KafkaBenchmark/Program.cs
```diff
@@ -0,0 +1,171 @@
...
+                        NumPartitions = 1,
+                        ReplicationFactor = 1
+                    }
+                });
+            }
+            catch (CreateTopicsException ex)
+            {
+                if (ex.Results[0].Error.Code == ErrorCode.TopicAlreadyExists)
+                {
+                    Console.WriteLine($"Topic {topicName} already exists, skipping creation");
```
andrewlock: This is a one-time operation, so it will increase the duration for the first run, which may not be a bit problem as long as we have warm-up work? Would it be better to always be creating a new topic for the run? If not, how persistent is this data? Do we need to _clear_ the topic before we start producing/consuming to/from it?

andrewlock (approved):
Nice work! This looks great as a base, I much prefer that the sample is in here, thank you 🙂  🙂

A couple of other things I'm wondering about, looking [at the benchmarking-platform branch](https://github.com/DataDog/benchmarking-platform/tree/dd-trace-dotnet/data-streams-monitoring):
- Even though it's 1000 messages, this is still pretty low pressure for the app, because we're only using a single thread, and producing/consuming separately. I wonder if we should set like 5 different consumer threads listening, and then have 5 threads producing 1000 messages each, and then wait for all that to finish. It will increase variation, but the scenario in which the app is actually under load is the one we're most interested in.
- 3.37 GB is a _lot_ of memory for this app to have at it's disposal. That will likely hide any overhead issues introduced by DSM. I think we should experiment with different heap values [by setting `DOTNET_GCHeapHardLimit=250`](https://learn.microsoft.com/en-us/dotnet/core/runtime-config/garbage-collector#heap-hard-limit). Basically, we want the app to be really struggling to keep up, and _then_ to add DSM on top 😅 As that's the scenario we have to worry about, and if we can survive that, we can have a lot of confidence in general.
- Startup overhead is always significant, and I'm wondering if it makes sense to try to exclude it from the comparison and to use the `process.internal_duration_ms` instead to compare between runs. I'm not sure, just a thought, as it will likely make us more sensitive to DSM impact.

These are all just tweaks though to get the most out of the test, overall it looks great to me 🙂

tracer/test/test-applications/throughput/Samples.KafkaBenchmark/Program.cs
robcarlan-datadog: Added!

tracer/test/test-applications/throughput/Samples.KafkaBenchmark/Program.cs
robcarlan-datadog: I moved the topic creation out of the code and into the start script. We had a warm up run before which would create this, and I don't think we _need_ to clear the topic as each run waits for the X messages to all be consumed before finishing

tracer/test/test-applications/throughput/Samples.KafkaBenchmark/Program.cs
robcarlan-datadog: I changed it to 1, hopefully we don't need them so may as well be explicit about it!

tracer/test/test-applications/throughput/Samples.KafkaBenchmark/Program.cs
andrewlock: > I moved the topic creation out of the code and into the start script. 
Awesome 👍 

>  I don't think we need to clear the topic as each run waits for the X messages to all be consumed before finishing

Yeah, I was thinking, for example, what happens on the second run - if the first run creates 100 messages and consumes 100, does the second run create 100 messages, and then the consume _200_ images? I don't think so because we exit the loop after 100, so it might consume _different_ messages I guess? 🤔 Does that matter? Probably not? 🤷‍♂️  😄

robcarlan-datadog: We acknowledge each message as they are consumed so each run will be practically independent of each other, and subsequent runs will consume from the end of the first run

.gitlab/benchmarks/macrobenchmarks.yml
robcarlan-datadog: Removed this

.gitlab-ci.yml
robcarlan-datadog: (Removed)

PR #7134, opened 2025-06-23
Deduce StreamName from StreamARN for Kinesis when StreamName not provided

robcarlan-datadog:
## Summary of changes
DSM does not submit data when requests are sent with StreamARN rather than StreamName. The cilent allows providing either, but recommends sending the full StreamARN instead ([docs](https://docs.aws.amazon.com/sdkfornet/v3/apidocs/items/Kinesis/TPutRecordRequest.html)). So if people follow the AWS docs, they will not see DSM data.

We fix this by using a deduced StreamName from the StreamARN if provided to supply the `streamname` tag.

This is pretty much the same logic as we use in the Java tracer: [here](https://github.com/DataDog/dd-trace-java/blob/16284fa2f1726e3038006252088d6a15df74367b/dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/AwsSdkClientDecorator.java#L161-L178).

## Reason for change
People using .NET & Kinesis won't see DSM data if they use use the StreamARN parameter, leading to a bad user experience.

## Implementation details

This only affects the PutRecord/PutRecords (sync & async) requests. GetRecords always uses the stream ARN, and we use the same logic of deducing a stream name from the StreamARN to create the context on the receiving side. [AWS Docs](https://docs.aws.amazon.com/sdkfornet/v3/apidocs/items/Kinesis/TGetRecordsRequest.html)

We have to split the Kinesis instrumentation into two, as the new field is only introduced in v3.7 and above. Most of the size of the PR is just duplicating the interface for the new version, and the autogenerated code.

## Test coverage
Added unit tests for the name deduction logic, existing integration tests cover versions below 3.7 and above, so test both instrumentations. The test doesn't use the StreamARN parameter, which I tested manually.


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Kinesis/AwsKinesisCommon.cs
```diff
@@ -38,6 +38,17 @@ internal static class AwsKinesisCommon
             return arnComponents[1];
         }
 
+        public static string? GetStreamName(IAmazonKinesisRequestWithStreamNameAndStreamArn request)
+        {
+            string? streamName = StreamNameFromARN(request.StreamARN);
+            if (string.IsNullOrEmpty(streamName))
```
bouwkast:
```suggestion
            if (StringUtil.IsNullOrEmpty(streamName))
```

Small nitpick, we added this recently (used in `StreamNameFromARN`) so it may not be on your branch unless you've rebased, but this `StringUtil.IsNullOrEmpty` allows for `string.IsNullOrEmpty` to work with `#nullable enable` as otherwise for older versions of .NET if you assert that a `string.IsNullOrEmpty` is `false` (e.g., it has a value) it will still show up as "possibly null".

This just gets around that, not very important though.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Kinesis/AwsKinesisCommon.cs
```diff
@@ -38,6 +38,17 @@ internal static class AwsKinesisCommon
...
 
+        public static string? GetStreamName(IAmazonKinesisRequestWithStreamNameAndStreamArn request)
+        {
+            string? streamName = StreamNameFromARN(request.StreamARN);
+            if (StringUtil.IsNullOrEmpty(streamName))
+            {
+                streamName = request.StreamName;
+            }
+
+            return streamName;
```
bricefriha, line 49: We could simplify this:
```suggestion
            string? streamName = StreamNameFromARN(request.StreamARN);
            return StringUtil.IsNullOrEmpty(streamName)? request.StreamName : streamName ;
```

At least I find it a bit less confusing this way, personally

bricefriha: I would simplify this like so:
```suggestion
             string? streamName = StreamNameFromARN(request.StreamARN);
             return StringUtil.IsNullOrEmpty(streamName)? request.StreamName : streamName ;
```
I find it a bit less confusing this way, personally

robcarlan-datadog: Done! I updated the variable name too to make it clearer.

PR #7137, opened 2025-06-24
Lower Case DSM RabbitMQ has_routing_key

robcarlan-datadog:

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/RabbitMQ/RabbitMQIntegration.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/RabbitMQ/RabbitMQIntegration.cs
@@ -141,7 +141,7 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.RabbitMQ
                 var edgeTags = string.IsNullOrEmpty(tags.Exchange) ?
                                    // exchange can be empty for "direct"
                                    new[] { "direction:out", $"topic:{tags.Queue ?? tags.RoutingKey}", "type:rabbitmq" } :
-                                   new[] { "direction:out", $"exchange:{tags.Exchange}", $"has_routing_key:{!string.IsNullOrEmpty(tags.RoutingKey)}", "type:rabbitmq" };
+                                   new[] { "direction:out", $"exchange:{tags.Exchange}", $"has_routing_key:{(!string.IsNullOrEmpty(tags.RoutingKey)).ToString().ToLowerInvariant()}", "type:rabbitmq" };
                 span.SetDataStreamsCheckpoint(dataStreamsManager, CheckpointKind.Produce, edgeTags, GetHeadersSize(headers) + messageSize, 0);
                 dataStreamsManager.InjectPathwayContext(span.Context.PathwayContext, headersAdapter);
             }


tracer/test/snapshots/DataStreamsMonitoringRabbitMQTests.HandleProduceAndConsume.verified.txt
```diff
@@ -1,38 +1,12 @@
 [
```
robcarlan-datadog, line 1: Diff looks worse because we order by hash, which has changed. But same results

robcarlan-datadog:
Failing integration test is unrelated:
```15:06:19 [DBG]   Error Message:
15:06:19 [DBG]    Docker.DotNet.DockerImageNotFoundException : Docker API responded with status code=NotFound, response={"message":"No such image: aerospike/aerospike-server:6.2.0.6"}
```

piochelepiotr (approved):
Thanks for doing the fix!

tracer/test/snapshots/DataStreamsMonitoringRabbitMQTests.HandleProduceAndConsume.verified.txt
```diff
@@ -287,10 +261,10 @@
     EdgeTags: [
       direction:out,
       exchange:DirectExchange,
-      has_routing_key:True,
+      has_routing_key:true,
```
bouwkast, line 264: Do customers use these EdgeTags? Just wondering if someone could be expecting a `True` or `False` exactly and then this breaks something?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/RabbitMQ/RabbitMQIntegration.cs
```diff
@@ -141,7 +141,7 @@ internal static void SetDataStreamsCheckpointOnProduce(Tracer tracer, Span span,
                 var edgeTags = string.IsNullOrEmpty(tags.Exchange) ?
                                    // exchange can be empty for "direct"
                                    new[] { "direction:out", $"topic:{tags.Queue ?? tags.RoutingKey}", "type:rabbitmq" } :
-                                   new[] { "direction:out", $"exchange:{tags.Exchange}", $"has_routing_key:{!string.IsNullOrEmpty(tags.RoutingKey)}", "type:rabbitmq" };
+                                   new[] { "direction:out", $"exchange:{tags.Exchange}", $"has_routing_key:{(!string.IsNullOrEmpty(tags.RoutingKey)).ToString().ToLowerInvariant()}", "type:rabbitmq" };
```
andrewlock: I think it might be preferable to just forgo the interpolation and use constants here:

```suggestion
                                   new[] { "direction:out", $"exchange:{tags.Exchange}", string.IsNullOrEmpty(tags.RoutingKey) ? "has_routing_key:false" : ""has_routing_key:true", "type:rabbitmq" };
```

tracer/test/snapshots/DataStreamsMonitoringRabbitMQTests.HandleProduceAndConsume.verified.txt
robcarlan-datadog: They shouldn't use this directly

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/RabbitMQ/RabbitMQIntegration.cs
robcarlan-datadog: Done!

tracer/test/snapshots/DataStreamsMonitoringRabbitMQTests.HandleProduceAndConsume.verified.txt
andrewlock: Just to confirm what Steven asked, this isn't like tracing tags, right? i.e. we don't _really_ expose these tags to customers in the DSM product?

andrewlock (approved):
Thanks!

tracer/test/snapshots/DataStreamsMonitoringRabbitMQTests.HandleProduceAndConsume.verified.txt
robcarlan-datadog: Correct, I looked into it a bit more, we don't show these to users at all. We drop these tags internally and just use it to collect metrics. Our internal code doesn't care about casing

PR #7138, opened 2025-06-24
[OTEL] Metrics API Support

link04:
## Summary of changes
Experimental OpenTelemetry Metrics support:
- Added configuration keys for enabling OTLP metrics (`DD_METRICS_OTEL_ENABLED`) and specifying enabled meters (`DD_METRICS_OTEL_METER_NAMES`).
- Introduced `OtlpMetricsExporter `(with duck-typed interfaces) for initializing OTLP metrics export.
- Updated tracer/test helpers to handle, deserialize, and verify OTLP metric requests.
- Added metrics instrumentation to the sample OpenTelemetrySdk test app.

This feature is tested across the following runtime + SDK combinations to ensure the correct set of metrics are exported for each:

| .NET Runtime | OTel SDK Version | Supported Metrics                                                                 |
|--------------|------------------|------------------------------------------------------------------------------------|
| `.NET 6`     | `1.3.2`          | `Counter`, `Histogram`, `ObservableCounter`, `ObservableGauge`                    |
| `.NET 7 / 8` | `1.5.1`          | All of the above + `UpDownCounter`, `ObservableUpDownCounter`                     |
| `.NET 9`     | `1.12.0`         | All of the above + `Gauge<T>` (requires both .NET 9 and OTel SDK ≥ 1.10.0)        |

✅ These specific versions were selected to represent the earliest SDK version that supports the full metric surface available on each runtime.

We test one representative OpenTelemetry version per runtime to avoid duplication and snapshot churn.

To enable OTLP metrics export via this integration, the application must:
Use OpenTelemetry SDK version 1.3.2 or higher
Configure an OTLP metrics exporter (e.g. to http://localhost:port)
Set the following environment variables:
```
DD_METRICS_OTEL_ENABLED=true — turns on metrics export
DD_TRACE_ENABLED_METERS=MyAppMeter — limits export to specific meters
OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf — required format
OTEL_EXPORTER_OTLP_ENDPOINT=http://<agent>:<port> — where metrics are sen
```
## Reason for change
POC to enable exporting metrics using the OpenTelemetry Metrics API and OTLP protocol for improved observability and interoperability. [APMAPI-1321](https://datadoghq.atlassian.net/browse/APMAPI-1321)

## Implementation details
- Uses duck-typing to integrate with OpenTelemetry SDK and OTLP exporter without direct dependency.
- Configuration-driven activation of metrics exporting and meter selection.
- Test helpers extended to parse protobuf OTLP metrics for validation.
- Sample application now emits various metric types (counter, histogram, up/down counter, gauge).
## Test coverage
- Added end-to-end integration tests to verify OTLP metrics are emitted and received.
- MockAgent updated to deserialize OTLP metric payloads and support snapshot verification, using the proto Otel files I generated locally and the adding the `Google.Protobuf` package.
- Sample app exercises new metrics instrumentation for test scenarios.

## Other details
The work on this PR is part of a experimental POC to demonstrate the tracer been able to setup and send the meters the customer has to the Agent using the OTLP endpoint it has, work has been scheduled to provide support and update some of the implementation details here before going GA.

[APMAPI-1321]: https://datadoghq.atlassian.net/browse/APMAPI-1321?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tracer/test/test-applications/integrations/Samples.OpenTelemetrySdk/Samples.OpenTelemetrySdk.csproj
```diff
@@ -3,7 +3,7 @@
     <!-- System.Runtime.CompilerServices.Unsafe doesn't support netcoreapp2.1-->
     <!-- https://andrewlock.net/stop-lying-about-netstandard-2-support/-->
     <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings>
-    <ApiVersion Condition="'$(ApiVersion)' == ''">1.7.0</ApiVersion>
+    <ApiVersion Condition="'$(ApiVersion)' == ''">1.10.0</ApiVersion>
```
link04: Just realized I left this in which I changed to match the system-tests, unless others tests do not pass I see no reason to not leave it as it is, otherwise I can downgrade it.

tracer/test/Datadog.Trace.TestHelpers/MockTracerAgent.cs
```diff
@@ -948,6 +959,54 @@ private void HandleTracerFlarePayload(MockHttpRequest request)
...
+            if (request.PathAndQuery.StartsWith("/v1/metrics") && contentType.Contains("protobuf"))
+            {
+                try
+                {
+                    var metricsData = MetricsData.Parser.ParseFrom(body);
+                    var resource = metricsData.ResourceMetrics[0];
+
+                    if (resource.ScopeMetrics[0] is not null)
+                    {
+                        deserializedData = resource.ScopeMetrics[0];
```
zacharycmontoya: How come you're only extracting the scope metrics? I think it's also important that we extract and test against the enclosing ResourceMetrics, since it contains the ScopeMetrics and Resources, which will become important later on

link04: To get something simple and make sure the values, unit, attributes and name match I decided to only do the scope but just changed it to send the full ResourceMetrics: 
- https://github.com/DataDog/dd-trace-dotnet/pull/7138/commits/011b913691f67646155f56aa3cded696919e2d61

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -805,6 +805,16 @@ internal static class FeatureFlags
             /// </summary>
             public const string OpenTelemetryEnabled = "DD_TRACE_OTEL_ENABLED";
 
+            /// <summary>
+            /// Enables beta support for exporting OTLP metrics generated by the OpenTelemetry Metrics API.
```
zacharycmontoya: Two small suggestions here:
- Since we said "experimental" support below, maybe we can change `beta` to `experimental`?
- Can you document that this is only supported for .NET 6 and later? Internally, the reason is that the BCL class `MeterListener` that makes this all possible is shipped in-box starting with [.NET 6](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics.meterlistener)

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -805,6 +805,16 @@ internal static class FeatureFlags
...
+            /// </summary>
+            public const string OpenTelemetryMetricsEnabled = "DD_METRICS_OTEL_ENABLED";
+
+            /// <summary>
+            /// List of meters to add to the metrics exporter for the experimental OpenTelemetry Metrics API support.
+            /// </summary>
+            public const string EnabledMeters = "DD_TRACE_ENABLED_METERS";
```
zacharycmontoya: Now that we're going with `DD_METRICS_OTEL_ENABLED`, what do you think of renaming this to something like `DD_METRICS_OTEL_METER_NAMES`?

tracer/src/Datadog.Trace/OTelMetrics/DuckTypes/IMeterProviderBuilder.cs
```diff
@@ -0,0 +1,22 @@
...
+#nullable enable
+
+using Datadog.Trace.DuckTyping;
+
+namespace Datadog.Trace.OTelMetrics.DuckTypes
+{
+    internal interface IMeterProviderBuilder : IDuckType
+    {
+        [Duck(ParameterTypeNames = new[] { "System.String[]" })]
+        IMeterProviderBuilder AddMeter(params string[] names);
```
zacharycmontoya: I think this can be simplified to the following suggestion. At runtime the target method AddMeter accepts a `string[]` so in our proxy let's just make it explicit, and I expect the DuckType matching to find the overload just fine.

```suggestion
        IMeterProviderBuilder AddMeter(string[] names);
```

tracer/src/Datadog.Trace/OTelMetrics/OtlpMetricsExporter.cs
```diff
@@ -0,0 +1,77 @@
...
+                    var otelSdkProxy = (IOtelSdk)otelSdkProxyResult.CreateInstance(null!);
+                    var meterProviderBuilder = otelSdkProxy.CreateMeterProviderBuilder();
+                    var builderProxy = meterProviderBuilder.DuckCast<IMeterProviderBuilder>();
+                    builderProxy.AddMeter(Tracer.Instance.Settings.EnabledMeters);
+
+                    var otlpExporterType = Type.GetType("OpenTelemetry.Metrics.OtlpMetricExporterExtensions, OpenTelemetry.Exporter.OpenTelemetryProtocol", throwOnError: false);
+                    if (otlpExporterType is null)
+                    {
+                        ThrowHelper.ThrowNullReferenceException($"The OpenTelemetry Protocol Exporter type is null, make sure the nuget is installed to collect metrics.");
+                        return;
```
zacharycmontoya: We can remove this line, the ThrowHelper will throw an exception and the control will switch to the catch block below.
```suggestion
```

tracer/src/Datadog.Trace/OTelMetrics/OtlpMetricsExporter.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        public static void Initialize()
+        {
+            try
+            {
+                var otelSdkType = Type.GetType("OpenTelemetry.Sdk, OpenTelemetry", throwOnError: false);
+                if (otelSdkType is null)
+                {
+                    ThrowHelper.ThrowNullReferenceException($"The OpenTelemetry SDK type is null, make sure the nuget installed to collect metrics.");
+                    return;
```
zacharycmontoya: Same comment as below, we don't need the return statement since it will never be hit.
```suggestion
```

tracer/src/Datadog.Trace/OTelMetrics/OtlpMetricsExporter.cs
```diff
@@ -0,0 +1,77 @@
...
+                    ThrowHelper.ThrowNullReferenceException($"Resulting proxy type after Ducktyping attempt {otelSdkProxyResultType} is null");
+                }
+                else if (otelSdkProxyResult.Success)
+                {
+                    var otelSdkProxy = (IOtelSdk)otelSdkProxyResult.CreateInstance(null!);
+                    var meterProviderBuilder = otelSdkProxy.CreateMeterProviderBuilder();
+                    var builderProxy = meterProviderBuilder.DuckCast<IMeterProviderBuilder>();
+                    builderProxy.AddMeter(Tracer.Instance.Settings.EnabledMeters);
+
+                    var otlpExporterType = Type.GetType("OpenTelemetry.Metrics.OtlpMetricExporterExtensions, OpenTelemetry.Exporter.OpenTelemetryProtocol", throwOnError: false);
```
zacharycmontoya: Can you rename this `otlpExporterType` variable? The actual type is the static class `OtlpMetricExporterExtensions` so the current name is misleading. The same comment applies below to the `otlpExporterProxyResult`

tracer/src/Datadog.Trace/OTelMetrics/OtlpMetricsExporter.cs
```diff
@@ -0,0 +1,77 @@
...
+                    }
+
+                    var otlpExporterProxyResult = DuckType.GetOrCreateProxyType(typeof(IOtlpMetricExporterExtensions), otlpExporterType);
+                    if (otlpExporterProxyResult.Success)
+                    {
+                        var otlpExporterProxy = (IOtlpMetricExporterExtensions)otlpExporterProxyResult.CreateInstance(null!);
+                        otlpExporterProxy.AddOtlpExporter(builderProxy);
+
+                        var meterProvider = builderProxy.Build();
+                        AppDomain.CurrentDomain.ProcessExit += (_, _) => meterProvider.Dispose();
```
zacharycmontoya: Small nit: In `tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs` we register a couple of actions on shutdown by calling `LifetimeManager.Instance.AddShutdownTask`. Can you add your Action to dipose the meterprovider there instead of directly calling ProcessExit?

tracer/test/Datadog.Trace.TestHelpers/MockTracerAgent.cs
zacharycmontoya: Thanks for updating it! This will be good to have moving forward

zacharycmontoya (approved):
I left small nits, but overall I think this design looks good for a POC. LGTM after the new integration tests pass

zacharycmontoya:
Note: My name is on some of the earlier commits, but that work has largely been excluded from this POC so I think it's ok that I approve it.

tracer/test/test-applications/integrations/Samples.OpenTelemetrySdk/OpenTelemetryMetricsMeter.cs
```diff
@@ -0,0 +1,24 @@
...
+    private static readonly Meter _meter = new Meter(MeterName);
+    public static Counter<long> LongCounter => _meter.CreateCounter<long>("example.counter");
+    public static Histogram<double> DoubleHistogram => _meter.CreateHistogram<double>("example.histogram");
+    public static ObservableCounter<long> AsyncLongCounter = _meter.CreateObservableCounter<long>("example.async.counter", () => 22L);
+    public static ObservableGauge<double> AsyncGauge = _meter.CreateObservableGauge<double>("example.async.gauge", () => 88L);
+#if NET7_0_OR_GREATER
+    public static UpDownCounter<long> LongUpDownCounter = _meter.CreateUpDownCounter<long>("example.upDownCounter");
+    public static ObservableUpDownCounter<long> AsyncLongUpDownCounter = _meter.CreateObservableUpDownCounter<long>("example.async.upDownCounter", () => 66L);
+#endif
+#if NET8_0_OR_GREATER
```
zacharycmontoya: According to these [docs](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics.gauge-1?view=net-9.0), this type is only built-in starting with .NET 9 (notice how it says `8 (package-provided),` whereas it says `9,`
```suggestion
#if NET9_0_OR_GREATER
```

link04: Addressed on my last commit: https://github.com/DataDog/dd-trace-dotnet/pull/7138/commits/aea1ab86d8d095e7351fe93be88e1f5eed97a8c0

tracer/src/Datadog.Trace/OTelMetrics/OtlpMetricsExporter.cs
link04: Addressed on my last commit: https://github.com/DataDog/dd-trace-dotnet/pull/7138/commits/aea1ab86d8d095e7351fe93be88e1f5eed97a8c0

tracer/src/Datadog.Trace/OTelMetrics/OtlpMetricsExporter.cs
link04: Addressed on my last commit: https://github.com/DataDog/dd-trace-dotnet/pull/7138/commits/aea1ab86d8d095e7351fe93be88e1f5eed97a8c0

tracer/src/Datadog.Trace/OTelMetrics/DuckTypes/IMeterProviderBuilder.cs
link04: Addressed on my last commit: https://github.com/DataDog/dd-trace-dotnet/pull/7138/commits/aea1ab86d8d095e7351fe93be88e1f5eed97a8c0

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/OpenTelemetrySdkTests.cs
```diff
@@ -197,6 +198,81 @@ public async Task IntegrationDisabled(string packageVersion)
...
+
+            if ((runtimeMajor == 7 || runtimeMajor == 8) && parsedVersion < new Version(1, 4, 0))
+            {
+                throw new SkipException("UpDownCounter requires OTel 1.4.0+.");
+            }
+
+            if (runtimeMajor == 9 && parsedVersion < new Version(1, 10, 0))
+            {
+                throw new SkipException("Gauge<T> in .NET 9 requires OTel 1.10.0+.");
+            }
```
link04: This is the part I am unsure about and believe could be improved, check the PR summary to understand the support OTEL provides and why I decided to add the tests like this. Appreciate all feedback 🙇 .

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -805,6 +805,18 @@ internal static class FeatureFlags
...
+            /// Enables experimental support for exporting OTLP metrics generated by the OpenTelemetry Metrics API.
+            /// This feature is only available starting with .NET 6.0, as it relies on the BCL class MeterListener
+            /// which is shipped in-box starting with .NET 6.
+            /// </summary>
+            public const string OpenTelemetryMetricsEnabled = "DD_METRICS_OTEL_ENABLED";
+
+            /// <summary>
+            /// List of meters to add to the metrics exporter for the experimental OpenTelemetry Metrics API support.
+            /// </summary>
+            public const string OpenTelemetryMeterNames = "DD_METRICS_OTEL_METER_NAMES";
```
NachoEchevarria, line 819: I did not see this one in https://github.com/DataDog/dd-go/blob/prod/trace/apps/tracer-telemetry-intake/telemetry-payload/static/config_norm_rules.json

I guess it should be added at some point before merging.

tracer/src/Datadog.Trace/OTelMetrics/OtlpMetricsExporter.cs
```diff
@@ -0,0 +1,75 @@
...
+#if NET6_0_OR_GREATER
+
+using System;
+using Datadog.Trace.DuckTyping;
+using Datadog.Trace.Logging;
+using Datadog.Trace.OTelMetrics.DuckTypes;
+
+#nullable enable
+
+namespace Datadog.Trace.OTelMetrics
```
NachoEchevarria, line 15: NIT: As a matter of personal preference, scoped namespaces seem cleaner.

tracer/src/Datadog.Trace/OTelMetrics/OtlpMetricsExporter.cs
```diff
@@ -0,0 +1,75 @@
...
+                if (otelSdkProxyResultType is null)
+                {
+                    ThrowHelper.ThrowNullReferenceException($"Resulting proxy type after Ducktyping attempt of {typeof(IOtelSdk)} is null.");
+                }
+                else if (otelSdkProxyResult.Success)
+                {
+                    var otelSdkProxy = (IOtelSdk)otelSdkProxyResult.CreateInstance(null!);
+                    var meterProviderBuilder = otelSdkProxy.CreateMeterProviderBuilder();
+                    var meterProviderProxy = meterProviderBuilder.DuckCast<IMeterProviderBuilder>();
+                    meterProviderProxy.AddMeter(Tracer.Instance.Settings.OpenTelemetryMeterNames);
```
NachoEchevarria: NIT: Maybe we should be avoid this call for performance reasons if OpenTelemetryMeterNames is empty. WDYT?

NachoEchevarria (approved):
While I am not used to work with OTEL and I have limited knowledge of that area, the code looks fine to me. Nice job!

tracer/src/Datadog.Trace/OTelMetrics/OtlpMetricsExporter.cs
```diff
@@ -0,0 +1,75 @@
...
+                        LifetimeManager.Instance.AddShutdownTask(_ => meterProvider.Dispose());
+                        Log.Information("Successfully Ducktyped and configured OTLP Metrics Exporter.");
+                    }
+                }
+            }
+            catch (Exception e)
+            {
+                Log.Error("Exception when initializing OTLP Metrics Exporter: {E}", e.ToString());
+                throw;
+            }
```
bouwkast: A nitpick, but it looks like `Initialize` is already covered by a try/catch so this isn't entirely necessary (and we'd be double logging the error here)

Maybe remove the `Log.Error` here 
Or just remove the try/catch entirely

Or do nothing as it isn't big, but would be nice to remove the `Log.Error` as if we do hit this we'll generate duplicate log messages essentially.
```suggestion
            catch (Exception e)
            {
                throw;
            }
```

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/OpenTelemetrySdkTests.cs
bouwkast: It looks fine to me, one thing though is that by default `packageVersion` when running locally will always be empty, so it'd hit the first `SkipException`

What we do in other cases is just assume a package version (usually what is installed in the sample application) and use that as the version if it is empty. Not great, but allows for at least running the test suite without code modification.

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/OpenTelemetrySdkTests.cs
```diff
@@ -197,6 +198,81 @@ public async Task IntegrationDisabled(string packageVersion)
...
+
+            var snapshotName = (runtimeMajor, packageVersion) switch
+            {
+                (6, "1.3.2") => "NET_6",
+                (7 or 8, "1.5.1") => "NET_7_8",
+                (>= 9, "1.12.0") => string.Empty,
+                _ => throw new SkipException($"Unnecessary test: runtime {runtimeMajor}, OTel {packageVersion}")
+            };
```
bouwkast: I think this will cause SkipExceptions to be thrown if we update the latest version? 

Probably want to make this consider ranges something like this (haven't run to see if it actually works 😄 ):



```suggestion
            var snapshotName = runtimeMajor switch
            {
                6 when parsedVersion >= new Version("1.3.2") && parsedVersion < new Version("1.5.0") => "NET_6",
                7 or 8 when parsedVersion >= new Version("1.5.1") && parsedVersion < new Version("1.10.0") => "NET_7_8",
                >= 9 when parsedVersion >= new Version("1.12.0") => string.Empty,
                _ => string.Empty
            };
```

tracer/test/Datadog.Trace.TestHelpers/Datadog.Trace.TestHelpers.csproj
```diff
@@ -8,6 +8,7 @@
     <PackageReference Include="HttpMultipartParser" Version="5.1.0" />
     <PackageReference Include="PublicApiGenerator" Version="10.2.0" />
     <PackageReference Include="DiffPlex" Version="1.7.1" />
+    <PackageReference Include="Google.Protobuf" Version="3.25.1" />
```
bouwkast, line 17: Could you add a comment in this file or in the sample app for how the proto files were generated?

tracer/src/Datadog.Trace/OTelMetrics/OtlpMetricsExporter.cs
```diff
@@ -0,0 +1,75 @@
...
+        private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(OtlpMetricsExporter));
+
+        public static void Initialize()
+        {
+            try
+            {
+                var otelSdkType = Type.GetType("OpenTelemetry.Sdk, OpenTelemetry", throwOnError: false);
+                if (otelSdkType is null)
+                {
+                    ThrowHelper.ThrowNullReferenceException($"The OpenTelemetry SDK type is null, make sure the nuget installed to collect metrics.");
```
bouwkast:
```suggestion
                    ThrowHelper.ThrowNullReferenceException($"The OpenTelemetry SDK type is null, make sure the OpenTelemetry NuGet package is installed to collect metrics.");
```

tracer/src/Datadog.Trace/OTelMetrics/OtlpMetricsExporter.cs
```diff
@@ -0,0 +1,75 @@
...
+                {
+                    var otelSdkProxy = (IOtelSdk)otelSdkProxyResult.CreateInstance(null!);
+                    var meterProviderBuilder = otelSdkProxy.CreateMeterProviderBuilder();
+                    var meterProviderProxy = meterProviderBuilder.DuckCast<IMeterProviderBuilder>();
+                    meterProviderProxy.AddMeter(Tracer.Instance.Settings.OpenTelemetryMeterNames);
+
+                    var otlpMetricExporterExtensionsType = Type.GetType("OpenTelemetry.Metrics.OtlpMetricExporterExtensions, OpenTelemetry.Exporter.OpenTelemetryProtocol", throwOnError: false);
+                    if (otlpMetricExporterExtensionsType is null)
+                    {
+                        ThrowHelper.ThrowNullReferenceException($"The OpenTelemetry Metrics Exporter Extensions type is null, make sure the nuget is installed to collect metrics.");
```
bouwkast:
```suggestion
                        ThrowHelper.ThrowNullReferenceException($"The OpenTelemetry Metrics Exporter Extensions type is null, make sure the OpenTelemetry NuGet package is installed to collect metrics.");
```

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
link04: So for this one we are not sure it will be used outside of the POC yet(not be intended for production) so will follow up on this if needed.

tracer/src/Datadog.Trace/OTelMetrics/OtlpMetricsExporter.cs
link04: I did a change so that if the config is not set we don't even try to initialize as it would be useless. Thank you.

PR #7141, opened 2025-06-25
[AAP] Report ASM RC status to Telemetry Config

daniel-romano-DD:
## Summary of changes
Report ASM enabled status from RC to Telemetry Config

## Reason for change
When ASM was enabled or disabled by RC (`ASM_FEATURES`) the new status was not reported to Telemetry Config
[Jira](https://datadoghq.atlassian.net/browse/APPSEC-57894)

### DIFF
--- a/tracer/src/Datadog.Trace/AppSec/Waf/ConfigurationState.cs
+++ b/tracer/src/Datadog.Trace/AppSec/Waf/ConfigurationState.cs
@@ -37,7 +37,7 @@ internal record ConfigurationState
 
     private readonly IReadOnlyDictionary<string, IAsmConfigUpdater> _productConfigUpdaters;
 
-    private readonly IConfigurationTelemetry? _telemetry;
+    private readonly IConfigurationTelemetry _telemetry;
 
     private readonly string? _rulesPath;
     private readonly bool _canBeToggled;
@@ -45,7 +45,7 @@ internal record ConfigurationState
     private readonly Dictionary<string, List<RemoteConfigurationPath>> _fileRemoves = new();
     private bool _defaultRulesetApplied = false;
 
-    public ConfigurationState(SecuritySettings settings, IConfigurationTelemetry? telemetry, bool wafIsNull)
+    public ConfigurationState(SecuritySettings settings, IConfigurationTelemetry telemetry, bool wafIsNull)
     {
         _telemetry = telemetry;
         _productConfigUpdaters = new Dictionary<string, IAsmConfigUpdater>
@@ -63,8 +63,8 @@ internal record ConfigurationState
         }
     }
 
-    public ConfigurationState(SecuritySettings settings, bool wafIsNull, Dictionary<string, RuleSet>? rulesetConfigs, Dictionary<string, Models.Asm.Payload>? asmConfigs, Dictionary<string, Models.AsmData.Payload>? asmDataConfigs)
-        : this(settings, null, wafIsNull)
+    public ConfigurationState(SecuritySettings settings, IConfigurationTelemetry telemetry, bool wafIsNull, Dictionary<string, RuleSet>? rulesetConfigs, Dictionary<string, Models.Asm.Payload>? asmConfigs, Dictionary<string, Models.AsmData.Payload>? asmDataConfigs)
+        : this(settings, telemetry, wafIsNull)
     {
         if (rulesetConfigs is not null)
         {
@@ -326,16 +326,13 @@ internal record ConfigurationState
             }
 
             // Update AppsecEnabled state based on the RCM data on Telemetry Config
-            if (_telemetry is not null)
+            if (IncomingUpdateState.ShouldInitAppsec)
             {
-                if (IncomingUpdateState.ShouldInitAppsec)
-                {
-                    _telemetry.Record(ConfigurationKeys.AppSec.Enabled, true, ConfigurationOrigins.RemoteConfig);
-                }
-                else if (IncomingUpdateState.ShouldDisableAppsec)
-                {
-                    _telemetry.Record(ConfigurationKeys.AppSec.Enabled, false, ConfigurationOrigins.RemoteConfig);
-                }
+                _telemetry.Record(ConfigurationKeys.AppSec.Enabled, true, ConfigurationOrigins.RemoteConfig);
+            }
+            else if (IncomingUpdateState.ShouldDisableAppsec)
+            {
+                _telemetry.Record(ConfigurationKeys.AppSec.Enabled, false, ConfigurationOrigins.RemoteConfig);
             }
         }
 
--- a/tracer/test/Datadog.Trace.Security.IntegrationTests/AsmTelemetryTests.cs
+++ b/tracer/test/Datadog.Trace.Security.IntegrationTests/AsmTelemetryTests.cs
@@ -1,4 +1,4 @@
-// <copyright file="AspNetCore5AsmTelemetry.cs" company="Datadog">
+// <copyright file="AsmTelemetryTests.cs" company="Datadog">
 // Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2 License.
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
 // </copyright>
@@ -23,10 +23,10 @@ using Xunit.Abstractions;
 
 namespace Datadog.Trace.Security.IntegrationTests
 {
-    public class AspNetCore5AsmTelemetry : AspNetBase, IClassFixture<AspNetCoreTestFixture>
+    public class AsmTelemetryTests : AspNetBase, IClassFixture<AspNetCoreTestFixture>
     {
-        public AspNetCore5AsmTelemetry(ITestOutputHelper outputHelper)
-            : base("AspNetCore5", outputHelper, "/shutdown", testName: nameof(AspNetCore5AsmTelemetry))
+        public AsmTelemetryTests(ITestOutputHelper outputHelper)
+            : base("AspNetCore5", outputHelper, "/shutdown", testName: nameof(AsmTelemetryTests))
         {
         }
 
@@ -44,7 +44,7 @@ namespace Datadog.Trace.Security.IntegrationTests
                 Output.WriteLine($"Assigned port {telemetry.Port} for the telemetry port.");
                 EnableAgentlessTelemetry(telemetry.Port);
 
-                await fixture.TryStartApp(this, enableSecurity: enableSecurity, sendHealthCheck: false, useTelemetry: true);
+                await fixture.TryStartApp(this, enableSecurity: enableSecurity, sendHealthCheck: false);
                 SetHttpPort(fixture.HttpPort);
 
                 var agent = fixture.Agent;
@@ -60,13 +60,11 @@ namespace Datadog.Trace.Security.IntegrationTests
             using (var fixture = new AspNetCoreTestFixture())
             {
                 fixture.SetOutput(Output);
-                // telemetry metric events under test are sent only when using managed trace exporter
-                SetEnvironmentVariable(ConfigurationKeys.TraceDataPipelineEnabled, "false");
 
                 Output.WriteLine($"Assigned port {telemetry.Port} for the telemetry port.");
                 EnableAgentlessTelemetry(telemetry.Port);
 
-                await fixture.TryStartApp(this, enableSecurity: null, sendHealthCheck: false, useTelemetry: true);
+                await fixture.TryStartApp(this, enableSecurity: null, sendHealthCheck: false);
                 SetHttpPort(fixture.HttpPort);
                 var configValues0 = (await telemetry.GetConfigurationValuesAsync(ConfigurationKeys.AppSec.Enabled)).ToList();
                 configValues0.Should().HaveCount(1);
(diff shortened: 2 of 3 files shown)


tracer/src/Datadog.Trace/AppSec/SecuritySettings.cs
```diff
@@ -29,8 +29,9 @@ internal class SecuritySettings
 
         public SecuritySettings(IConfigurationSource? source, IConfigurationTelemetry telemetry)
         {
+            Telemetry = telemetry ?? TelemetryFactory.Config;
```
andrewlock: `telemetry` is not allowed to be null based on nullable attributes, so you don't need `TelemetryFactory.Config`. Also, as this is passed in the constructor, you shouldn't need to expose this as a property at all? 🤔

tracer/test/Datadog.Trace.Security.IntegrationTests/ApiSecurity/AspNetCore2ApiSecurity.cs
```diff
@@ -19,15 +19,15 @@ namespace Datadog.Trace.Security.IntegrationTests.ApiSecurity
     public class AspNetCore2ApiSecurityEnabled : AspNetCoreApiSecurity
     {
         public AspNetCore2ApiSecurityEnabled(AspNetCoreTestFixture fixture, ITestOutputHelper outputHelper)
-            : base(fixture, outputHelper, true, "AspNetCore2")
+            : base(fixture, outputHelper, true, true, "AspNetCore2")
```
andrewlock: nit: bool parameters should use named parameters so it's easier to read what they're doing :)

tracer/test/Datadog.Trace.Security.IntegrationTests/AspNetCore5AsmTelemetry.cs
```diff
@@ -0,0 +1,105 @@
...
+
+namespace Datadog.Trace.Security.IntegrationTests
+{
+    public class AspNetCore5AsmTelemetry : AspNetBase, IClassFixture<AspNetCoreTestFixture>
+    {
+        public AspNetCore5AsmTelemetry(ITestOutputHelper outputHelper)
+            : base("AspNetCore5", outputHelper, "/shutdown", testName: nameof(AspNetCore5AsmTelemetry))
+        {
+            // telemetry metric events under test are sent only when using managed trace exporter
+            SetEnvironmentVariable(ConfigurationKeys.TraceDataPipelineEnabled, "false");
```
andrewlock: That's not true 😄 You're reading configuration values, and those are sent regardless. It's only "trace metrics" e.g. "number of traces sent" etc which aren't emitted with the data pipeline
```suggestion
```

tracer/test/Datadog.Trace.Security.IntegrationTests/AspNetCore5AsmTelemetry.cs
```diff
@@ -0,0 +1,105 @@
...
+        [Trait("RunOnWindows", "True")]
+        [InlineData(true)]
+        [InlineData(false)]
+        public async Task TestSecurityInitialization(bool enableSecurity)
+        {
+            using (var fixture = new AspNetCoreTestFixture())
+            {
+                fixture.SetOutput(Output);
+                // telemetry metric events under test are sent only when using managed trace exporter
+                SetEnvironmentVariable(ConfigurationKeys.TraceDataPipelineEnabled, "false");
```
andrewlock:
```suggestion
```

tracer/test/Datadog.Trace.Security.IntegrationTests/AspNetCore5AsmTelemetry.cs
```diff
@@ -0,0 +1,105 @@
...
+            SetEnvironmentVariable(ConfigurationKeys.TraceDataPipelineEnabled, "false");
+        }
+
+        [SkippableTheory]
+        [Trait("RunOnWindows", "True")]
+        [InlineData(true)]
+        [InlineData(false)]
+        public async Task TestSecurityInitialization(bool enableSecurity)
+        {
+            using (var fixture = new AspNetCoreTestFixture())
```
andrewlock: This is all a little strange, which is partly just how the ASM stuff works tbh 🤔 You're using an existing fixture in the app (because of this: `IClassFixture<AspNetCoreTestFixture>`) but then creating _another_ instance here... That's "fine", just more work and more complex than it needs to be I think 🤔 

Basically, there's no need to use the more complex "aspnetcore" samples for these tests AFAICT, you could just "normal" samples like we do in the "tracer" (i.e. not security) tests

tracer/test/Datadog.Trace.Security.IntegrationTests/AspNetCore5AsmTelemetry.cs
```diff
@@ -0,0 +1,105 @@
...
+            {
+                fixture.SetOutput(Output);
+                // telemetry metric events under test are sent only when using managed trace exporter
+                SetEnvironmentVariable(ConfigurationKeys.TraceDataPipelineEnabled, "false");
+
+                using var telemetry = new MockTelemetryAgent();
+                Output.WriteLine($"Assigned port {telemetry.Port} for the telemetry port.");
+                EnableAgentlessTelemetry(telemetry.Port);
```
andrewlock: I don't think you need these, because you set `useTelemetry`  when starting the app, which sets up the telemetry agent proxy...

tracer/test/Datadog.Trace.Security.IntegrationTests/AspNetCore5AsmTelemetry.cs
```diff
@@ -0,0 +1,105 @@
...
+
+                Output.WriteLine($"Assigned port {telemetry.Port} for the telemetry port.");
+                EnableAgentlessTelemetry(telemetry.Port);
+
+                await fixture.TryStartApp(this, enableSecurity: null, sendHealthCheck: false, useTelemetry: true);
+                SetHttpPort(fixture.HttpPort);
+                var configValues0 = telemetry.GetConfigurationValues(ConfigurationKeys.AppSec.Enabled).ToList();
+                configValues0.Should().HaveCount(1);
+                configValues0[0].Value.Should().Be(false);
+                configValues0[0].Origin.Should().Be("default");
```
andrewlock: I need to do some refactoring soon, and all this is going to break 😅 Not your problem, but still 😿

tracer/test/Datadog.Trace.Security.IntegrationTests/AspNetCore5AsmTelemetry.cs
```diff
@@ -0,0 +1,105 @@
...
+
+                var agent = fixture.Agent;
+                var request = await agent.SetupRcmAndWait(Output, [(new AsmFeatures { Asm = new AsmFeature { Enabled = true } }, "ASM_FEATURES", "TestSecurityInitializationRC-1")]);
+
+                telemetry.WaitForLatestTelemetry(x => x.IsRequestType(TelemetryRequestTypes.AppClientConfigurationChanged));
```
andrewlock: This needs to be async after tony's changes 🙂

tracer/test/Datadog.Trace.Security.IntegrationTests/Datadog.Trace.Security.IntegrationTests.csproj
```diff
@@ -110,4 +111,8 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
   </ItemGroup>
+
+  <ItemGroup>
+    <Folder Include="Helpers\" />
+  </ItemGroup>
```
andrewlock:
```suggestion
```

tracer/test/Datadog.Trace.Security.IntegrationTests/AspNetCore5AsmTelemetry.cs
dromanol: That's true, we could use a simpler example. I'll do it.
The reason for not reusing the fixture is to reuse the class and code and make sure the app and agents are recreated always

tracer/test/Datadog.Trace.Security.IntegrationTests/AspNetCore5AsmTelemetry.cs
dromanol: This is copy pasta from `TelemetryTests`

tracer/test/Datadog.Trace.Security.IntegrationTests/AspNetCore5AsmTelemetry.cs
dromanol: I've taken the logic from `TelemetryTests`. I tried to do it with the `useTelemetry` flag with little success..

tracer/src/Datadog.Trace/AppSec/Waf/ConfigurationState.cs
```diff
@@ -58,7 +64,7 @@ public ConfigurationState(SecuritySettings settings, bool wafIsNull)
     }
 
     public ConfigurationState(SecuritySettings settings, bool wafIsNull, Dictionary<string, RuleSet>? rulesetConfigs, Dictionary<string, Models.Asm.Payload>? asmConfigs, Dictionary<string, Models.AsmData.Payload>? asmDataConfigs)
-        : this(settings, wafIsNull)
+        : this(settings, null, wafIsNull)
```
andrewlock: Shouldn't we require a telemetry instance here? 🤔

tracer/test/Datadog.Trace.Security.IntegrationTests/AspNetCore5AsmTelemetry.cs
andrewlock: > I tried to do it with the useTelemetry flag with little success..

In which case, you should probably just set `useTelemetry: false` to avoid confusion 🙂

tracer/test/Datadog.Trace.Security.IntegrationTests/AspNetCore5AsmTelemetry.cs
```diff
@@ -0,0 +1,101 @@
...
+using Datadog.Trace.Configuration;
+using Datadog.Trace.Telemetry;
+using Datadog.Trace.TestHelpers;
+using FluentAssertions;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Datadog.Trace.Security.IntegrationTests
+{
+    public class AspNetCore5AsmTelemetry : AspNetBase, IClassFixture<AspNetCoreTestFixture>
```
andrewlock: nit: I would rename this to `AsmTelemetryTests` - the aspnetcore is an implementation detail, what you're really testing is ASM telemetry. But definitely your choice 😄

tracer/test/Datadog.Trace.Security.IntegrationTests/AspNetCore5AsmTelemetry.cs
```diff
@@ -0,0 +1,101 @@
...
+        [SkippableFact]
+        [Trait("RunOnWindows", "True")]
+        public async Task TestSecurityInitializationRC()
+        {
+            using var telemetry = new MockTelemetryAgent();
+            using (var fixture = new AspNetCoreTestFixture())
+            {
+                fixture.SetOutput(Output);
+                // telemetry metric events under test are sent only when using managed trace exporter
+                SetEnvironmentVariable(ConfigurationKeys.TraceDataPipelineEnabled, "false");
```
andrewlock: This isn't true for what you're testing
```suggestion
```

tracer/test/Datadog.Trace.Security.IntegrationTests/Datadog.Trace.Security.IntegrationTests.csproj
```diff
@@ -10,6 +10,7 @@
   <ItemGroup>
     <None Remove="applicationHost.config" />
     <None Remove="xunit.runner.json" />
+    <Compile Include="..\Datadog.Trace.ClrProfiler.IntegrationTests\Helpers\TelemetryHelper.cs" Link="Helpers\TelemetryHelper.cs" />
```
andrewlock, line 13: Can we move this file into `Datadog.Trace.TestHelpers.AutoInstrumentation` or  `Datadog.Trace.TestHelpers` instead?

tracer/test/Datadog.Trace.Security.Unit.Tests/RASP/RASPRcmTests.cs
```diff
@@ -54,7 +54,7 @@ private static AppSec.Security CreateSecurity()
     {
         var source = CreateConfigurationSource([(ConfigurationKeys.AppSec.Enabled, "true")]);
         var settings = new SecuritySettings(source, NullConfigurationTelemetry.Instance);
-        var confState = new ConfigurationState(settings, true);
+        var confState = new ConfigurationState(settings, null, true);
```
andrewlock, line 57: Instead of using null (and _allowing_) null in `ConfigurationState` (it's probably better to use `NullTelemetryConfig.Instance`

robertpi (approved):
LGTM

tracer/src/Datadog.Trace/AppSec/Waf/ConfigurationState.cs
dromanol: This method is meant for tests, basically, but I can enforce a telemetry instance

tracer/test/Datadog.Trace.Security.IntegrationTests/AspNetCore5AsmTelemetry.cs
dromanol: Agree

PR #7142, opened 2025-06-25
[AAP] Send WAF config errors to Telemetry Logs

daniel-romano-DD:
## Summary of changes
Upload WAF config errors to Telemetry Logs

## Reason for change
WAF errors were only dumped to local logs

[Jira](https://datadoghq.atlassian.net/browse/APPSEC-57959)


tracer/src/Datadog.Trace/AppSec/Waf/Initialization/WafConfigurator.cs
```diff
@@ -212,15 +216,10 @@ void DumpMessages(IReadOnlyDictionary<string, object>? messages, bool isError)
...
-                            {
-                                // TODO: This message should go to telemetry logs only, skipping the regular logs
-                                Log.Debug("rc::asm_dd::diagnostic Error: {Err}", message);
-                            }
-                            else
-                            {
-                                Log.Debug("rc::asm_dd::diagnostic Warning: {Err}", message);
-                            }
+                            var severity = isError ? "Error" : "Warning";
+#pragma warning disable DDLOG004 // Message templates should be constant
```
andrewlock, line 220:
```suggestion
                            var severity = isError ? "Error" : "Warning";
                            // We explicitly disable this here only because we know that the values we're sending
                            // are not going to contain PII, and that we will only send the value very rarely
#pragma warning disable DDLOG004 // Message templates should be constant
```

tracer/src/Datadog.Trace/AppSec/Waf/Initialization/WafConfigurator.cs
```diff
@@ -191,8 +194,9 @@ internal UpdateResult Update(IntPtr wafBuilderHandle, ConfigurationState configu
 
                 if (diags.HasErrors)
                 {
-                    // TODO: This message should go to telemetry logs only, skipping the regular logs
-                    Log.Debug("Some errors were found while applying waf configuration (RulesFile: {RulesFile})", rulesFile);
+#pragma warning disable DDLOG004 // Message templates should be constant
```
andrewlock, line 197:
```suggestion
               // We explicitly disable this here only because we know that the values we're sending
                // are not going to contain PII, and that we will only send the value very rarely
#pragma warning disable DDLOG004 // Message templates should be constant
```

tracer/src/Datadog.Trace/AppSec/Waf/ReturnTypes.Managed/DiagnosticResultUtils.cs
```diff
@@ -17,7 +17,7 @@ namespace Datadog.Trace.AppSec.Waf.ReturnTypes.Managed;
 
 internal static class DiagnosticResultUtils
 {
-    internal static ReportedDiagnostics ExtractReportedDiagnostics(DdwafObjectStruct diagObject, bool noRuleDiagnoticsIsError)
+    internal static ReportedDiagnostics ExtractReportedDiagnostics(ref DdwafObjectStruct diagObject, bool noRuleDiagnoticsIsError)
```
andrewlock: This can be `in` I think?
```suggestion
    internal static ReportedDiagnostics ExtractReportedDiagnostics(in DdwafObjectStruct diagObject, bool noRuleDiagnoticsIsError)
```

tracer/src/Datadog.Trace/AppSec/Waf/ReturnTypes.Managed/DiagnosticResultUtils.cs
```diff
@@ -143,6 +143,8 @@ public WafStats()
     }
 
     public int Total => (Failed + Loaded + Skipped);
+
+    public bool HasErrors => Errors is { Count: > 0 } || Warnings is { Count: > 0 };
```
andrewlock, line 147: So this is a behaviour change from the above, because it now includes warnings, is that correct?

tracer/src/Datadog.Trace/AppSec/Waf/ReturnTypes.Managed/UpdateResult.cs
```diff
@@ -14,16 +14,16 @@ namespace Datadog.Trace.AppSec.Waf.ReturnTypes.Managed
 {
     internal class UpdateResult
     {
-        private UpdateResult(DdwafObjectStruct? diagObject, IntPtr builderHandle = default(IntPtr), IntPtr wafHandle = default(IntPtr), WafLibraryInvoker? invoker = null, IEncoder? encoder = null)
+        private UpdateResult(ref DdwafObjectStruct diagObject, IntPtr builderHandle = default(IntPtr), IntPtr wafHandle = default(IntPtr), WafLibraryInvoker? invoker = null, IEncoder? encoder = null)
```
andrewlock: Prefer `in` over `ref` where possible

tracer/test/Datadog.Trace.Security.Unit.Tests/RASP/RASPRcmTests.cs
```diff
@@ -21,7 +21,7 @@ public class RaspRcmTests : SettingsTestsBase
...
...t\t\t\"category\": \"attack_attempt\",\r\n\t\t\t\t\"cwe\": \"77\",\r\n\t\t\t\t\"capec\": \"1000/152/248/88\",\r\n\t\t\t\t\"confidence\": \"1\"\r\n\t\t\t},\r\n\t\t\t\"conditions\": [\r\n\t\t\t\t{\r\n\t\t\t\t\t\"parameters\": {\r\n\t\t\t\t\t\t\"inputs\": [\r\n\t\t\t\t\t\t\t{\r\n\t\t\t\t\t\t\t\t\"address\": \"server.request.query\"\r\n\t\t\t\t\t\t\t},\r\n\t\t\t\t\t\t\t{\r\n\t\t\t\t\t\t\t\t\"address\": \"server.request.body\"\r\n\t\t\t\t\t\t\t},\r\n\t\t\t\t\t\t\t{\r\n\t\t\t\t\t\t\t\t\"address\": \"server.request.path_params\"\r\n\t\t\t\t\t\t\t},\r\n\t\t\t\t\t\t\t{\r\n\t\t\t\t\t\t\t\t\"address\": \"grpc.server.request.message\"\r\n\t\t\t\t\t\t\t},\r\n\t\t\t\t\t\t\t{\r\n\t\t\t\t\t\t\t\t\"address\": \"graphql.server.all_resolvers\"\r\n\t\t\t\t\t\t\t},\r\n\t\t\t\t\t\t\t{\r\n\t\t\t\t\t\t\t\t\"address\": \"graphql.server.resolver\"\r\n\t\t\t\t\t\t\t}\r\n\t\t\t\t\t\t],\r\n\t\t\t\t\t\t\"list\": [\r\n\t\t\t\t\t\t\t\"dev/zero\"\r\n\t\t\t\t\t\t]\r\n\t\t\t\t\t},\r\n\t\t\t\t\t\"operator\": \"phrase_match\"\r\n\t\t\t\t}\r\n\t\t\t],\r\n\t\t\t\"transformers\": [\r\n\t\t\t\t\"lowercase\",\r\n\t\t\t\t\"cmdLine\"\r\n\t\t\t],\r\n\t\t\t\"on_match\": [\r\n\t\t\t\t\"block\"\r\n\t\t\t]\r\n\t\t}\r\n\t]\r\n}")]
```
andrewlock, line 24: Did you expect this to change? 😅

tracer/build/_build/Build.Steps.cs
```diff
@@ -2418,6 +2418,11 @@ string NormalizedPath(AbsolutePath ap)
                new(@".*Exception occurred when calling the CallTarget integration continuation. Datadog.Trace.ClrProfiler.CallTarget.CallTargetInvokerException: Throwing a call target invoker exception.*"),
                // error thrown to check error handling in RC tests
                new(@".*haha, you weren't expect this!*", RegexOptions.Compiled),
+               // known errors in waf config
+               new(@".*rc::asm_dd::diagnostic Error: missing key.*", RegexOptions.Compiled),
```
andrewlock, line 2439: It would be nice if we could scope these to specific test runs/files, as if we get these errors in _other_ tests that in it indicates a problem 🤔

andrewlock (approved):

tracer/src/Datadog.Trace/AppSec/Waf/Initialization/WafConfigurator.cs
dromanol: Fine enough

tracer/src/Datadog.Trace/AppSec/Waf/ReturnTypes.Managed/DiagnosticResultUtils.cs
dromanol: That's it

tracer/test/Datadog.Trace.Security.Unit.Tests/RASP/RASPRcmTests.cs
dromanol: Yes, the previous payload had config errors

tracer/build/_build/Build.Steps.cs
dromanol: I considered that option, as we talked, but then realized that if this happened elsewhere that test would fail on its own, and preferred to keep complexity low

PR #6748, opened 2025-03-06
[tracer] fix NRE when `Span.ServiceName` is null

lucaspimentel:
## Summary of changes

If `Span.ServiceName` is null, a `NullReferenceException` is thrown when closing the span.

## Reason for change

Prevent the exception.

## Implementation details

Add two different null checks to be safe:
- `TraceContext.CloseSpan()`
- `ExtraServicesProvider.AddService()`

## Test coverage

Added test for both `TraceContext` and `ExtraServicesProvider`. Confirmed they failed before, and that each one is resolved separately with each added null check.

## Other details
n/a

PR #7144, opened 2025-06-26
Check for incorrect values in Manual assembly

NachoEchevarria:
## Summary of changes

We are getting the following exception:

```
System.ArgumentException
at Datadog.Trace.Util.ThrowHelper.ThrowArgumentException(String message, String paramName)
at Datadog.Trace.UserDetails..ctor(String id)
at Datadog.Trace.ClrProfiler.AutoInstrumentation.ManualInstrumentation.Extensions.SpanExtensionsSetUserIntegration.OnMethodBegin[TTarget,TSpan](TSpan& span, String email, String name, String id, Boolean propagateId, String sessionId, String role, String scope)
at REDACTED
at Datadog.Trace.ClrProfiler.CallTarget.Handlers.BeginMethodHandler`10.Invoke(TTarget instance, TArg1& arg1, TArg2& arg2, TArg3& arg3, TArg4& arg4, TArg5& arg5, TArg6& arg6, TArg7& arg7, TArg8& arg8)
at Datadog.Trace.ClrProfiler.CallTarget.CallTargetInvoker.BeginMethod[TIntegration,TTarget,TArg1,TArg2,TArg3,TArg4,TArg5,TArg6,TArg7,TArg8](TTarget instance, TArg1& arg1, TArg2& arg2, TArg3& arg3, TArg4& arg4, TArg5& arg5, TArg6& arg6, TArg7& arg7, TArg8& arg8)
at Datadog.Trace.SpanExtensions.SetUser(ISpan span, String email, String name, String id, Boolean propagateId, String sessionId, String role, String scope)
```

This happens because the constructor of UserDetails does not allow empty or null ids. This exception can be thrown if an external user send null values but they are logged as an internal tracer error. In order to avoid that, we are checking for incorrect values in Trace.Manual methods.


tracer/src/Datadog.Trace.Manual/SpanExtensions.cs
```diff
@@ -30,7 +31,14 @@ public static class SpanExtensions
     /// <param name="span">The span to be tagged</param>
     /// <param name="userDetails">The details of the current logged on user</param>
     public static void SetUser(this ISpan span, UserDetails userDetails)
-        => SetUser(span, userDetails.Email, userDetails.Name, userDetails.Id, userDetails.PropagateId, userDetails.SessionId, userDetails.Role, userDetails.Scope);
+    {
+        if (string.IsNullOrEmpty(userDetails.Id))
```
lucaspimentel, line 35: This will now throw an exception in cases where it didn't before. Is that a concern?

lucaspimentel: For comparison, we had a similar issue if users set `Span.ServiceName` to `null`, so we recently added a check in #6748 to silently ignore it to avoid the `NullReferenceException` while also avoiding a breaking change.

NachoEchevarria: We are not going to throw new exceptions but to launch the same one earlier, at Datadog.Trace.Manual instead.
Previously, it was thrown at UserDetails.cs inside the tracer. By checking that in the Manual assembly, we make sure that we don't log the error.

lucaspimentel: I see, thanks! Resolving and approving.

lucaspimentel (approved):

NachoEchevarria:

PR #7147, opened 2025-06-26
Don't write native loader logs when bailing out in Windows SSI

andrewlock:
## Summary of changes

- When we're running in Windows SSI, buffer logs for native loader until we know whether we're bailing out. If we bail out, drop the logs, otherwise, write them and enable-auto flushing
- Refactor `InstrumentationTests`

## Reason for change

In Windows SSI, we inject into every process by default, and then bail out. Today, this results in creating a native loader file for every injection, which causes a lot of noise in the log files. With the updated implementation, we buffer those logs until we know we're continuing with instrumentation instead, so we don't create loads of log files.

## Implementation details

- Use spdlog's "backtrace" support, that allows buffering logs without writing them.
- Introduce an `enable_buffering()` method to the `TLoggerPolicy` abstraction, to know whether logs should be buffered initially, or just auto-flushed
  - This returns `false` for Tracer and Profiler, because by the time they're running, we know we want to instrument.
  - For the native loader, we look for the SSI variable. If it's set, we check if buffering is explicitly enabled/disabled (default is enabled). This ensures that if we have a crash (for example) that we have a way of forcing logs to be written.
- If buffering is **not enabled**, the existing behavior (pre this PR) is used.
- If buffering is **enabled**:
  - We disable automatic flushing
  - Set the log level to be "don't write any logs"
  - Enable back trace
  - Use a `lazy_rolling_file_sink`. This is necessary because the built-in `rolling_file_sink` creates and opens the target file as soon as it's created. The updated version delays this until we write to the sink
  - Later, if we determine that we _do_ want to instrument the app, we call `FlushAndDisableBuffering` to restore the original behaviour

## Test coverage

I refactored the existing instrumentation tests 
- Don't not rely on `dotnet run`, because is a bit messy because we instrument it _and_ the target app (we should look into this seperately)
- Reduce duplication
- Add expanded assertions that we don't have native tracer or profiler in "normal" bailout scenarios
- Add additional tests and assertions for the new behavaiour in this PR
- Updated execution benchmarks scenario to emulate bailout in SSI (i.e. no logs scenario)

### DIFF
--- a/shared/src/Datadog.Trace.ClrProfiler.Native/log.h
+++ b/shared/src/Datadog.Trace.ClrProfiler.Native/log.h
@@ -23,6 +23,34 @@ public:
         inline static const shared::WSTRING folder_path = WStr(R"(Datadog .NET Tracer\logs)");
 #endif
         inline static const std::string pattern = "[%Y-%m-%d %H:%M:%S.%e | %l | PId: %P | TId: %t] %v";
+
+        static bool enable_buffering() {
+#ifdef _WIN32
+            // For Windows SSI, we are injected into all .NET Core processes, so we can't enable flushing of logs
+            // until we know that the process is going to be instrumented, otherwise we will create way
+            // too many log files. Note that we don't currently distinguish between .NET FX and .NET Core, because
+            // we need to make the decision _before_ we have that information.
+            const auto isSingleStepVariable = shared::GetEnvironmentValue(EnvironmentVariables::SingleStepInstrumentationEnabled);
+            if (isSingleStepVariable.empty())
+            {
+                // not in Windows SSI, so we don't need buffering
+                return false;
+            }
+
+            const auto bufferEnvVar = shared::GetEnvironmentValue(EnvironmentVariables::LogBufferingEnabled);
+
+            // enable buffering _unless_ the variable is present and set to false
+            // if the variable is not set, or it's set to false
+            bool enable_buffering;
+            return bufferEnvVar.empty()
+                || !shared::TryParseBooleanEnvironmentValue(bufferEnvVar, enable_buffering)
+                || enable_buffering == true;
+#else
+            // On linux/mac, we are only injected where we know we are needed, so we don't buffer
+            return false;
+#endif
+        }
+
         struct logging_environment
         {
             inline static const shared::WSTRING log_path = EnvironmentVariables::LogPath;
@@ -31,6 +59,11 @@ public:
     };
 
     inline static ds::Logger* const Instance = ds::LogManager::Get<Log::NativeLoaderLoggerPolicy>();
+
+    static void EnableAutoFlush()
+    {
+        Instance->FlushAndDisableBuffering();
+    }
 
     static bool IsDebugEnabled()
     {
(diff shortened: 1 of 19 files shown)


shared/src/native-src/lazy_rotating_file_sink-inl.h
andrewlock, line 1: This file is basically a copy of the `spdlog/sinks/rotating_file_sink-inl`. The only difference (other than naming) is that opening the file is delayed until we first write a log.

image

shared/src/native-src/lazy_rotating_file_sink.h
andrewlock, line 1: This file is basically a copy of the `spdlog/sinks/rotating_file_sink`. The only difference (other than naming) is the addition of the `file_opened_` flag

image

shared/src/native-src/lazy_rotating_file_sink-inl.h
```diff
@@ -0,0 +1,158 @@
...
+    std::lock_guard<Mutex> lock(base_sink<Mutex>::mutex_);
+    return file_helper_.filename();
+}
+
+template<typename Mutex>
+SPDLOG_INLINE void lazy_rotating_file_sink<Mutex>::sink_it_(const details::log_msg &msg)
+{
+     if (!file_opened_)
+    {
+        // I _think_ this is called under a mutex, so we don't need to lock here.
```
tonyredondo, line 74: is that true? is it a spdlog thing?

tonyredondo (commented):
Looks good. I think we should also add an execution benchmark test simulating SSI in the envvars.

shared/src/native-src/lazy_rotating_file_sink-inl.h
andrewlock: > is that true? is it a spdlog thing?

Yes, I believe so. What I know for sure is that if I move this to a separate function, and add a mutex, it deadlocks 😅

tonyredondo: yeah now I see this function signature and it seems to be pretty clear 😅  `template<typename Mutex>...`

shared/src/Datadog.Trace.ClrProfiler.Native/log.h
```diff
@@ -23,6 +23,34 @@ class Log final
...
+            }
+
+            const auto bufferEnvVar = shared::GetEnvironmentValue(EnvironmentVariables::LogBufferingEnabled);
+
+            // enable buffering _unless_ the variable is present and set to false
+            // if the variable is not set, or it's set to false
+            bool enable_buffering;
+            return bufferEnvVar.empty()
+                || !shared::TryParseBooleanEnvironmentValue(bufferEnvVar, enable_buffering)
+                || enable_buffering == true;
```
gleocadie, line 47: Just to be sure I understand. We will enable buffering if
- `EnvironmentVariables::LogBufferingEnabled` is not present (default behavior)
- `EnvironmentVariables::LogBufferingEnabled` is set but not parseable
- `EnvironmentVariables::LogBufferingEnabled` is set to true

andrewlock: Correct!

Put another way, the _only_ way will _not_ enable buffering in SSI on Windows is if `EnvironmentVariables::LogBufferingEnabled` is set to `false` (or equivalent)

PR #7148, opened 2025-06-26
[Docs] Fix Readme BuildAndRunOsxIntegrationTests sample

robcarlan-datadog:
## Summary of changes
The sample integration test command also matches `DataStreamsMonitoringRabbitMQTests`, which depends on `Samples.DataStreams.RabbitMQ` which causes the example command to fail. Updating the readme to match the exact test name instead.

## Reason for change
New users following this guide will likely be confused why this example test fails.

### DIFF
--- a/tracer/README.MD
+++ b/tracer/README.MD
@@ -161,7 +161,7 @@ docker-compose up StartDependencies.OSXARM64
 ./build.sh BuildAndRunOsxIntegrationTests
 
 # Build and run integration tests filtering on one framework, one set of tests and a sample app.
-./build.sh BuildAndRunOsxIntegrationTests --framework "net6.0" --filter "rabbit" --SampleName "Samples.Rabbit"
+./build.sh BuildAndRunOsxIntegrationTests --framework "net6.0" --filter "FullyQualifiedName=Datadog.Trace.ClrProfiler.IntegrationTests.RabbitMQTests" --SampleName "Samples.Rabbit"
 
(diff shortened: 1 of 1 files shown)


tracer/README.MD
```diff
@@ -161,7 +161,7 @@ docker-compose up StartDependencies.OSXARM64
 ./build.sh BuildAndRunOsxIntegrationTests
 
 # Build and run integration tests filtering on one framework, one set of tests and a sample app.
-./build.sh BuildAndRunOsxIntegrationTests --framework "net6.0" --filter "rabbit" --SampleName "Samples.Rabbit"
+./build.sh BuildAndRunOsxIntegrationTests --framework "net6.0" --filter "FullyQualifiedName=Datadog.Trace.ClrProfiler.IntegrationTests.RabbitMQTests" --SampleName "Samples.Rabbit"
```
bouwkast:
```suggestion
./build.sh BuildAndRunOsxIntegrationTests --framework "net6.0" --filter "Datadog.Trace.ClrProfiler.IntegrationTests.RabbitMQTests" --SampleName "Samples.Rabbit"
```

Is `FullyQualifiedName=` required on Mac for this?
I typically do exactly what you've outlined just omitting that part (and on Windows)

robcarlan-datadog: It works without, I'll use that as it's simpler 👍

bouwkast: Awesome thanks!

PR #5005, opened 2023-12-27
Add support for writing managed logs to stdout

andrewlock:
## Summary of changes

Adds ~support~ experimental setting for writing managed logs to stdout/console

## Reason for change

For testing in particular, this simplifies various things

## Implementation details

Added a new sink

## Test coverage

Basic coverage

## Other details

Currently writes logs in the same format as the file logs. This means it likely isn't massively useful for customers (who will generally want JSON/custom templates) etc, but that isn't the target for now

APMSVLS-57 (@lucaspimentel)

PR #7160, opened 2025-06-30
Add support for writing managed logs to stdout

lucaspimentel:
## Summary of changes

Based on #5005. Adds _experimental_ setting for writing managed logs to stdout/console.

## Reason for change

Writing logs to stdout instead of a file is more convenient or required in certain situations:
- Testing
- Serverless or other environments without file system access

## Implementation details

- Add a new `AsyncConsoleSink : ILogEventSink` which buffers messages and writes them from a background thread.
- Add a new `SingleLineTextFormatter : ITextFormatter` which writes messages into a single line (including stack traces) and uses UTC timestamps.
- Add a new value for `DD_TRACE_LOG_SINKS`: `console-experimental` which uses the new console sink and text formatter.

## Test coverage

Add unit tests for `AsyncConsoleSink` and `SingleLineTextFormatter`.

## Other details

APMSVLS-57


tracer/src/Datadog.Trace/Logging/Internal/DatadogLoggingFactory.cs
```diff
@@ -30,57 +30,67 @@ internal static class DatadogLoggingFactory
...
             {
-                return false;
-            }
-
-            foreach (var value in array)
-            {
-                if (string.IsNullOrWhiteSpace(value))
-                {
-                    continue;
-                }
```
andrewlock, line 67: nit: You removed the fast path, seems good to keep it to me, as it avoids the `Trim()` allocation if it's not needed? 🤷‍♂️

tracer/src/Datadog.Trace/Logging/Internal/Sinks/AsyncConsoleSink.cs
```diff
@@ -0,0 +1,117 @@
...
+
+    private readonly ITextFormatter _textFormatter;
+    private readonly BlockingCollection<LogEvent> _writeQueue;
+    private readonly TextWriter _consoleWriter;
+    private readonly Task _writeTask;
+
+    public AsyncConsoleSink(ITextFormatter formatter, TextWriter consoleWriter, int queueLimit)
+    {
+        // in-memory buffer used only from the background thread
+        _buffer = new StringBuilder(capacity: 512);
```
andrewlock: IS 512 a reasonable number? The `DATADOG TRACER CONFIGURATION` log (for example) is currently 2654 characters long. Given we will _always_ log that, a `512` starting size seems to ensure that we will pretty much _always_ end up doing _at least_ 3 doubling -> 512 -> 1024 -> 2048 -> 4096?

May as well start with at least the contiguous 4096, right? Because as soon as you add in stack traces, this is going to balloon (and not ever be re-claimed)

Pretty much the _shortest_ log we write is about ~300 characters, so basically this currently does no buffering 😅

tracer/src/Datadog.Trace/Logging/Internal/Sinks/AsyncConsoleSink.cs
```diff
@@ -0,0 +1,117 @@
...
+            WriteToConsoleStream,
+            CancellationToken.None,
+            TaskCreationOptions.LongRunning | TaskCreationOptions.DenyChildAttach,
+            TaskScheduler.Default);
+    }
+
+    public void Emit(LogEvent? logEvent)
+    {
+        // TODO: metrics about dropped logs?
+
```
andrewlock: meh
```suggestion
```

tracer/src/Datadog.Trace/Logging/Internal/Sinks/AsyncConsoleSink.cs
```diff
@@ -0,0 +1,117 @@
...
+            // The foreach loop will exit when the collection is marked as complete.
+            foreach (var logEvent in _writeQueue.GetConsumingEnumerable())
+            {
+                // clear in-memory buffer and format event into buffer
+                _buffer.Clear();
+                _textFormatter.Format(logEvent, _bufferWriter);
+
+                // In more recent runtimes, this uses the new WriteLine(StringBuilder) overload which
+                // uses StringBuilder.GetChunks() without allocating a new string.
+                // In earlier runtimes, the WriteLine(object) overload calls StringBuilder.ToString() instead.
```
andrewlock: This comment doesn't apply now, does it? 🤔

tracer/src/Datadog.Trace/Logging/Internal/Sinks/AsyncConsoleSink.cs
```diff
@@ -0,0 +1,117 @@
...
+        try
+        {
+            // GetConsumingEnumerable() blocks until an item
+            // is available or the collection is marked as complete.
+            // The foreach loop will exit when the collection is marked as complete.
+            foreach (var logEvent in _writeQueue.GetConsumingEnumerable())
+            {
+                // clear in-memory buffer and format event into buffer
+                _buffer.Clear();
+                _textFormatter.Format(logEvent, _bufferWriter);
```
andrewlock: If we have _really_ big logs (which we do get, esp in Debug), what happens to the stringbuilder... it grows indefinitely, right? Should we worry about that? :/

tracer/src/Datadog.Trace/Logging/Internal/Sinks/AsyncConsoleSink.cs
```diff
@@ -0,0 +1,117 @@
...
+        }
+    }
+
+    public async Task FlushAsync()
+    {
+        _writeQueue.CompleteAdding();
+        await _writeTask.ConfigureAwait(false);
+    }
+
+    public void Dispose()
```
andrewlock: Hmmm, can we have an `AsyncDispose` where possible? 🤔 I'm guessing Serilog doesn't support that in our vendored version?

tracer/src/Datadog.Trace/Logging/Internal/TextFormatters/SingleLineTextFormatter.cs
```diff
@@ -0,0 +1,50 @@
...
+using Datadog.Trace.Vendors.Serilog.Formatting.Display;
+
+namespace Datadog.Trace.Logging.Internal.TextFormatters;
+
+internal class SingleLineTextFormatter : ITextFormatter
+{
+    public void Format(LogEvent logEvent, TextWriter output)
+    {
+        if (logEvent == null!) { ThrowHelper.ThrowArgumentNullException(nameof(logEvent)); }
+        if (output == null!) { ThrowHelper.ThrowArgumentNullException(nameof(output)); }
```
andrewlock: Meh, we can probably omit these and rely on nullable annotations given it's all our own internal code?

tracer/src/Datadog.Trace/Logging/Internal/TextFormatters/SingleLineTextFormatter.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        // buffer the entire log event into a single line before writing it to the output
+        var buffer = StringBuilderCache.Acquire();
```
andrewlock: So we have two layers of `StringBuilder` buffering? 🤔 Is that actually useful?

tracer/src/Datadog.Trace/Logging/Internal/TextFormatters/SingleLineTextFormatter.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        // if there is an exception, write it to the output
+        if (logEvent.Exception != null)
+        {
+            bufferWriter.Write($" | {logEvent.Exception}");
+        }
+
+        // replace any newlines with literal "\n" to ensure the output is a single line
+        buffer.Replace("\r\n", "\\n")
+              .Replace("\n", "\\n");
```
andrewlock: eww 😅 That should make it nice and hard to read 😛

tracer/src/Datadog.Trace/Logging/Internal/Sinks/AsyncConsoleSink.cs
```diff
@@ -0,0 +1,117 @@
...
+    {
+        FlushAsync().GetAwaiter().GetResult();
```
tonyredondo: 😢 😿 😭 🩸

tonyredondo: **using Datadog.Trace.ExtensionMethods;**

```suggestion
        FlushAsync().SafeWait();
```

Doesn't change too much but handles the synchronization context, let's try to use it:
https://github.com/DataDog/dd-trace-dotnet/blob/524bc48ac21c4bcdbb6ae1a160dae0d0cd265ed2/tracer/src/Datadog.Trace/ExtensionMethods/TaskExtensions.cs#L16

tracer/src/Datadog.Trace/Logging/Internal/Sinks/AsyncConsoleSink.cs
```diff
@@ -0,0 +1,117 @@
...
+            {
+                // clear in-memory buffer and format event into buffer
+                _buffer.Clear();
+                _textFormatter.Format(logEvent, _bufferWriter);
+
+                // In more recent runtimes, this uses the new WriteLine(StringBuilder) overload which
+                // uses StringBuilder.GetChunks() without allocating a new string.
+                // In earlier runtimes, the WriteLine(object) overload calls StringBuilder.ToString() instead.
+                _consoleWriter.Write(_buffer);
+                _consoleWriter.Flush();
```
tonyredondo: I think this is not required for a console stdout textwriter.

https://source.dot.net/#System.Console/System/ConsolePal.Unix.ConsoleStream.cs,52

https://github.com/microsoft/referencesource/blob/f7df9e2399ecd273e90908ac11caf1433e142448/mscorlib/system/io/__consolestream.cs#L109

tracer/src/Datadog.Trace/Logging/Internal/TextFormatters/SingleLineTextFormatter.cs
tonyredondo: .NET >= 6.0 we can use: https://learn.microsoft.com/en-us/dotnet/api/system.string.replacelineendings?view=net-6.0

tracer/src/Datadog.Trace/Logging/Internal/Sinks/AsyncConsoleSink.cs
```diff
@@ -0,0 +1,117 @@
...
+
+        _textFormatter = formatter;
+        _writeQueue = new BlockingCollection<LogEvent>(queueLimit);
+        _consoleWriter = consoleWriter;
+
+        _writeTask = Task.Factory.StartNew(
+            WriteToConsoleStream,
+            CancellationToken.None,
+            TaskCreationOptions.LongRunning | TaskCreationOptions.DenyChildAttach,
+            TaskScheduler.Default);
```
tonyredondo: just a thought but we could use a normal Thread.Start here and set the priority to `BelowNormal` to reduce the impact of the `Console.Write` contention (interop calls) on other threads in high traffic scenarios. WDYT @andrewlock

tracer/src/Datadog.Trace/Logging/Internal/DatadogLoggingFactory.cs
lucaspimentel: I think this is better for the common case. Simpler code and less redundant checks.

- Neither `item` and `value` can ever be null or empty strings (I have updated the nullability annotations, thanks), so checking for that is redundant.
- `Trim()` returns the same string without allocating if a string doesn't need trimming.
- I suspect a whitespace-only `item` will be very rare. If it does happen, `string.IsNullOrWhiteSpace()` will iterate over all chars. `Trim()` will also iterate over all chars, will return `string.Empty` (no allocation), and `string.Equals()` will fail fast when it compares string lengths. I think it's a fair trade.
- If `item` needs trimming, then ~the `Trim()` allocation is unavoidable anyway and~ the call to `string.IsNullOrWhiteSpace()` was redundant. edit: I switched to `ReadOnlySpan<char>` to avoid the allocation from `string.Trim()`.

tracer/src/Datadog.Trace/Logging/Internal/Sinks/AsyncConsoleSink.cs
lucaspimentel: Makes sense. Bumped to 4096.

tracer/src/Datadog.Trace/Logging/Internal/Sinks/AsyncConsoleSink.cs
lucaspimentel: I couldn't sworn I had removed that. Thanks.

tracer/src/Datadog.Trace/Logging/Internal/Sinks/AsyncConsoleSink.cs
lucaspimentel: It does apply. I tweaked the comment for clarity. What I meant is that the `TextWriter.Write()` overloads already to that internally, so we don't need to do it ourselves like we used to.

tracer/src/Datadog.Trace/Logging/Internal/TextFormatters/SingleLineTextFormatter.cs
lucaspimentel: Sure. Removed.

tracer/src/Datadog.Trace/Logging/Internal/TextFormatters/SingleLineTextFormatter.cs
lucaspimentel: Yeah, I was wondering about that myself. The reason why I did this is because `SingleLineTextFormatter` and `AsyncConsoleSink` could in _theory_ be used separately (they are decoupled), and I want both to buffer an entire line before writing it, for different reasons:

- `SingleLineTextFormatter` buffers each line so it can replace all newlines. Although, we could replace newlines in each part separately (message, exception, etc) and not buffer the entire line? Yeah, I'll try this.
- `AsyncConsoleSink` buffers so it can write entire lines to the output in an attempt to avoid other threads writing output _inside_ our log line.

tracer/src/Datadog.Trace/Logging/Internal/TextFormatters/SingleLineTextFormatter.cs
lucaspimentel: `ReplaceLineEndings()` allocates a new string. I was doing these `Replace()` calls on a `StringBuilder`. Do we care?

tracer/src/Datadog.Trace/Logging/Internal/Sinks/AsyncConsoleSink.cs
lucaspimentel: Done, thanks!

tracer/src/Datadog.Trace/Logging/Internal/Sinks/AsyncConsoleSink.cs
lucaspimentel: I _think_ we were safe here because the parent task was created with `TaskScheduler.Default` (despite it's name, the "default" scheduler is `TaskScheduler.Current`), but it doesn't hurt to be extra safe. Thanks.

tracer/src/Datadog.Trace/Logging/Internal/TextFormatters/SingleLineTextFormatter.cs
lucaspimentel: > we could replace newlines in each part separately (message, exception, etc) and not buffer the entire line?

This would allocate new strings for each `string.Replace()`, though. Currently the code is using `StringBuilder.Replace()`.

tracer/src/Datadog.Trace/Logging/Internal/Sinks/AsyncConsoleSink.cs
lucaspimentel: We're only buffering single log lines here, so I wouldn't say it grows _indefinitely_ 😅

We could reset the buffer's size after each use
```csharp
_buffer.Capacity = 4096
```
but if log lines are frequently longer than 4096, the `StringBuilder` will be constantly shrinking and growing, and both operations allocate a new `char[]`.

tracer/src/Datadog.Trace/Logging/Internal/Sinks/AsyncConsoleSink.cs
lucaspimentel: Replaced with `FlushAsync().SafeWait()` (although this task was already using `TaskScheduler.Default`, not `TaskScheduler.Current`)

tracer/src/Datadog.Trace/Logging/Internal/Sinks/AsyncConsoleSink.cs
lucaspimentel: They are no-ops! Removed, thanks.

tracer/src/Datadog.Trace/Logging/Internal/TextFormatters/SingleLineTextFormatter.cs
lucaspimentel: Turns out `StringBuilder.Replace()` also allocates internal `int` arrays, so I don't think either one is much worse that the other for this usage.

lucaspimentel: Switched to `ReplaceLineEndings()` in .NET 6+, thanks.

tracer/src/Datadog.Trace/Logging/Internal/TextFormatters/SingleLineTextFormatter.cs
lucaspimentel: Ok, I removed the `StringBuilder` from `SingleLineTextFormatter` so we're not double-buffering, we only replace newlines in the exception part of the message (if there is one).

tracer/src/Datadog.Trace/Logging/Internal/Sinks/AsyncConsoleSink.cs
andrewlock: > so I wouldn't say it grows indefinitely

Yeah, indefinitely was the wrong phrase, I meant permanently 😄 Would be kinda nice to have an array-pool backed `StringBuilder` implementation to handle this 🤔 Anyway, thinking out loud 😄

tracer/src/Datadog.Trace/Logging/Internal/TextFormatters/SingleLineTextFormatter.cs
andrewlock: > Turns out StringBuilder.Replace() also allocates internal int arrays, so I don't think either one is much worse that the other for this usage.

I only see a `stackalloc` int array 🤔 And that's only if you're replacing whole strings. If you could switch it to being a `char` for `char` replacement, you can completely remove the allocation. e.g. you could (for example) replace `\n` with `\t` 

Also, I just realised, your existing code doesn't work as you expect 🤔 

```csharp
// replace any newlines with literal "\n" to ensure the output is a single line
    buffer.Replace("\r\n", "\\n")
          .Replace("\n", "\\n");
```

It replaces  `\r\n` with `\\\n` (too many `\`!) 😅

tracer/src/Datadog.Trace/Logging/Internal/DatadogLoggingFactory.cs
```diff
@@ -30,43 +30,41 @@ internal static class DatadogLoggingFactory
...
         {
             fileConfig = GetFileLoggingConfiguration(source, telemetry);
         }
 
+        ConsoleLoggingConfiguration? consoleConfig = null;
+        if (logSinkOptions is not null && Contains(logSinkOptions, LogSinkOptions.Console))
+        {
+            consoleConfig = GetConsoleLoggingConfiguration(source);
+        }
+
```
bricefriha: I wonder if it wouldn't be better to simplify this:

```suggestion
FileLoggingConfiguration? fileConfig = logSinkOptions is null || Contains(logSinkOptions, LogSinkOptions.File) ? GetFileLoggingConfiguration(source, telemetry) : null;
ConsoleLoggingConfiguration? consoleConfig = logSinkOptions is not null && Contains(logSinkOptions, LogSinkOptions.Console) ? GetConsoleLoggingConfiguration(source) : null ;

```
Maybe it's not as readable though. 
But it saves lines and the fact that we declare `fileConfig` and `consoleConfig` and set them to `null` just to set them again 2 lines later triggers something in me 😅

andrewlock: Personally not a fan of the one-liners, I think they're really hard to follow while the separate `if` is clear 😄

bricefriha: I totally get this point!
I guess in other contexts I would have been more for separate `if` statements, I just feel like it's kind of dragging the code.

But again, just a personal feeling

tracer/src/Datadog.Trace/Logging/Internal/TextFormatters/SingleLineTextFormatter.cs
lucaspimentel: > I only see a `stackalloc` int array

Ahh, I missed that. I did switch to `ReplaceLineEndings()` in .NET 6+ anyway (before seeing this comment).

> It replaces `\r\n` with `\\\n` (too many `\`!)

Are you sure?

First `Replace()` replaces `[CR][LF]` with literal `@"\n"` (two characters).
Second one replaces any stand-alone `[LF]` also with literal `@"\n"` (two characters).

tracer/src/Datadog.Trace/Logging/Internal/DatadogLoggingFactory.cs
lucaspimentel: Related, we could also set the setting's default value to `"file"`. Then `logSinkOptions` would never be null and we can remove some of the null checks:
```csharp
FileLoggingConfiguration? fileConfig = null;
if (Contains(logSinkOptions, LogSinkOptions.File))
{
    fileConfig = GetFileLoggingConfiguration(source, telemetry);
}

ConsoleLoggingConfiguration? consoleConfig = null;
if (Contains(logSinkOptions, LogSinkOptions.Console))
{
    consoleConfig = GetConsoleLoggingConfiguration(source);
}
```

That may make the ternary operator look a little nicer, but meh.

@andrewlock would defaulting to `"file"` be better for config telemetry?

tracer/src/Datadog.Trace/Logging/Internal/TextFormatters/SingleLineTextFormatter.cs
andrewlock: Ah, yeah, you're right 👍  but you know what would clear that up for sure, a unit test 😉

tracer/src/Datadog.Trace/Logging/Internal/DatadogLoggingFactory.cs
andrewlock: > Related, we could also set the setting's default value to "file". Then logSinkOptions would never be null and we can remove some of the null checks:

Yeah, that makes a lot of sense actually, as that's the _actual_ default.

lucaspimentel: Done, thank you.

tracer/src/Datadog.Trace/Logging/Internal/TextFormatters/SingleLineTextFormatter.cs
lucaspimentel: added a unit test

tracer/src/Datadog.Trace/Logging/Internal/Sinks/AsyncTextWriterSink.cs
```diff
@@ -0,0 +1,129 @@
...
+            // if the queue is full, the log event is dropped.
+            _writeQueue.TryAdd(logEvent);
+        }
+        catch (InvalidOperationException)
+        {
+            // queue is marked as complete, drop log event.
+            // this won't usually happen as we check IsAddingCompleted above,
+            // but it can theoretically happen if the queue is completed from another thread
+            // after we checked IsAddingCompleted but before we called TryAdd().
+        }
```
andrewlock, line 76: I hope we _don't_ hit this path, as that's exactly the sort of Exception on shutdown that [this PR is hoping to avoid](https://github.com/DataDog/dd-trace-dotnet/pull/7200) 😅

tracer/src/Datadog.Trace/Logging/Internal/Sinks/AsyncTextWriterSink.cs
```diff
@@ -0,0 +1,129 @@
...
+
+        _textFormatter = formatter;
+        _textWriter = textWriter;
+        _writeQueue = new BlockingCollection<LogEvent>(queueLimit);
+
+        _writeTask = Task.Factory.StartNew(
+            WriteBufferedLogEvents,
+            CancellationToken.None,
+            TaskCreationOptions.LongRunning | TaskCreationOptions.DenyChildAttach,
+            TaskScheduler.Default);
```
andrewlock, line 52: @tonyredondo suggested switching away from a `Task` to a `Thread` here... I'm undecided if it's worth the hassle or not tbh, We already get a dedicated thread here by default due to `LongRunning` so it's really only about changing the schedule priority. That might be good for startup times in Serverless. That said, I don't think it's critical, given that we only intended to enable this in debugging scenarios now, and we can always refactor this later.

tracer/src/Datadog.Trace/Logging/Internal/TextFormatters/SingleLineTextFormatter.cs
```diff
@@ -0,0 +1,48 @@
...
+namespace Datadog.Trace.Logging.Internal.TextFormatters;
+
+internal class SingleLineTextFormatter : ITextFormatter
+{
+    public void Format(LogEvent logEvent, TextWriter output)
+    {
+        // write the timestamp, source, and log level to the output
+        var utcTimestamp = logEvent.Timestamp.ToUniversalTime();
+        var logLevel = LevelOutputFormat.GetLevelMoniker(logEvent.Level, "u3");
+        output.Write($"[{utcTimestamp:yyyy-MM-dd HH:mm:ss.fff zzz} | DD_TRACE_DOTNET {TracerConstants.ThreePartVersion} | {logLevel}] ");
```
andrewlock: nit: using `$` allocates the string, and _then_ writes it to the output. Ultimately you're writing into a `StringBuilder` so I think it's better to split this in to multiple `Write` calls and avoid the extra allocation?

tracer/src/Datadog.Trace/Logging/Internal/TextFormatters/SingleLineTextFormatter.cs
```diff
@@ -0,0 +1,48 @@
...
+    internal static string ToSingleLineString(Exception exception)
+    {
+        var exceptionString = exception.ToString();
+
+#if NET6_0_OR_GREATER
+        return exceptionString.ReplaceLineEndings("\\n");
+#else
+        return exceptionString.Replace("\r\n", "\\n")
+                              .Replace("\n", "\\n");
+#endif
```
andrewlock, line 54: It's kind of annoying that both of these allocate a new string. I feel like there should be a way to do this with `ReadOnlySpan<T>` and stuff and then write that to the `TextWriter`, but meh 😅

andrewlock (approved):
Nice work, thanks!

tracer/src/Datadog.Trace/Logging/Internal/Sinks/AsyncTextWriterSink.cs
lucaspimentel: `:noooo:`

tracer/src/Datadog.Trace/Logging/Internal/TextFormatters/SingleLineTextFormatter.cs
lucaspimentel: Yeah, I agree. But if you have an exception, you're already allocating all sorts of thing like stack traces etc, so I don't feel _too_ bad about it.

tracer/src/Datadog.Trace/Logging/Internal/TextFormatters/SingleLineTextFormatter.cs
lucaspimentel: Good call. Done.

PR #7163, opened 2025-07-01
Fix test InitializesTracerWhenTracingIsDisabled Flakiness

NachoEchevarria:
## Summary of changes

Test InitializesTracerWhenTracingIsDisabled has been [detected as flaky](https://app.datadoghq.com/ci/test/runs?query=test_level%3Atest%20%40test.service%3Add-trace-dotnet%20%40test.status%3Afail%20%40git.branch%3Amaster%20%40test.name%3AInitializesTracerWhenTracingIsDisabled&agg_m=count&agg_m_source=base&agg_t=count&currentTab=overview&eventStack=&fromUser=true&index=citest&start=1750763474473&end=1751368274473&paused=false).

The issue could be reproduced locally and seems to be related with a timeout (it happens in the debug runs, with more log lines)

The current message is misleading because it suggests that the reason of the failure is a missing file. This message was also shown in timeouts.

The test code has been updated with the following changes:
* Increase timeout
* Add an optional output helper to increase log info.
* Better exception message according to the failure reason.


tracer/test/Datadog.Trace.TestHelpers/LogEntryWatcher.cs
```diff
@@ -87,8 +103,10 @@ public async Task<string[]> WaitForLogEntries(string[] logEntries, TimeSpan? tim
 
             if (line != null)
             {
+                logsRead.Add(line);
```
andrewlock: I'm kind of unsure if we _should_ add that one, as it might slow things down too much, but worth a try!

andrewlock (approved):
Great, thanks!

tracer/test/Datadog.Trace.TestHelpers/LogEntryWatcher.cs
NachoEchevarria: Maybe we could potentially replace it by "last log line read"?

PR #7164, opened 2025-07-01
[IAST] Added compatibility for stable config

daniel-romano-DD:


e-n-0 (approved):

tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
```diff
@@ -16,18 +16,15 @@ using namespace std::chrono;
...
-static const WSTRING _fixedAppDomainExcludeFilters[] = {
-    WStr("DD*"), WStr("DataDog*"),
-    LastEntry, // Can't have an empty array. This must be the last element
+static std::vector<WSTRING> _domainExcludeFilters = {
+    WStr("DD*"), 
+    WStr("DataDog*"),
 };
-static const WSTRING _fixedAssemblyIncludeFilters[] = {
-    LastEntry, // Can't have an empty array. This must be the last element
+static std::vector<WSTRING> _assemblyIncludeFilters = {
```
anna-git, line 25: is it ok now to have an empty array? referring to previous comment 
```
// Can't have an empty array. This must be the last element
```

dromanol: Yes, because it's not an array anymore, it's a vector, so it can be empty

PR #7169, opened 2025-07-02
Don't load the profiler on arm64

andrewlock:
## Summary of changes

- Don't try to load the profiler on Linux arm64 at all
- Cherry pick #7153 to see if it resolves the issues

## Reason for change

#7153 ran into what we think is glibc `dlclose` issue (i.e. #7125). Given the profiler always bails out on arm64 as it's not yet supported, this is the easiest way to ensure that we don't load (and therefore, more importantly, we don't _unload_ the profiler on arm64).

## Implementation details

Comment out the profiler from the loader.conf for arm64. That ensures we don't _try_ to load it on arm64, while making it easy to reenable for testing etc.

## Test coverage

Cherry picked the #7153 smoke tests, to confirm it resolves the issues

## Other details

Supersedes #7153 given that it enables the tests to verify the fix


gleocadie (approved):
👍

PR #756, opened 2020-06-16
Remove the heap allocation from Stopwatch

kevingosse:
Remove the heap allocation from `Stopwatch` (one per trace).
See comments below for benchmarks.

PR #7170, opened 2025-07-02
Update build and testing to use .NET 10 preview 7

NachoEchevarria:
## Summary of changes

Updates build and testing to use .NET 10 preview 7.

## Reason for change

We want to test our changes ahead of the .NET 10 GA release.

## Implementation details

We had to address a bunch of issues as part of the update, in addition to the usual "bump version numbers", some of which are listed below. Much of the work was already extracted to separate PRs, but not all of it.

The standard changes were
- Bump version numbers
- Update VMs 
- Update docker build images
- Add additional .NET 10 smoke tests
- Update integrations to instrument .NET 10 APIs

The additional issues addressed in this PR

- .NET 10 [has bumped](https://github.com/dotnet/runtime/issues/109939) the minimum supported version of Alpine from 3.13 ([.NET 9<](https://github.com/dotnet/core/blob/main/release-notes/9.0/supported-os.md#linux-compatibility)) to 3.17 ([.NET 10+](https://github.com/dotnet/core/blob/main/release-notes/10.0/supported-os.md#linux-compatibility)). My _previous_ understanding was that musl is essentially unversioned, so they're essentially also forward compatible. Unfortunately, that is apparently not true. Attempting to run .NET 10 on Alpine 3.13 does not work, and it specifically fails with a `symbol not found` error.
  - We extracted the work to fix it to a separate PR, but as that requires updating the VMs, it's included in this branch, and won't be explicitly merged separately.
  - Basically we bump our build image from `alpine:3.14` to `alpine:3.17` and make sure we don't depend on any additional symbols that aren't available.
  - https://github.com/DataDog/dd-trace-dotnet/pull/7228
- As a consequence of bumping the alpine version to 3.17, The SQLite tests started failing on .NET Core 3.1 and .NET 5. It was basically a version of [this issue](https://github.com/dotnet/runtime/issues/65152); those versions of .NET don't "recognise" alpine, so fail to load the native library.
  - As a workaround, we explicitly set `DOTNET_RUNTIME_ID=linux-musl-x64` in the tests that use SQLite.
  - We _could_ set this globally, but I would rather have the "default" resolution behaviour (except where we explicitly need the fix), to make sure we don't mask any resolution issues in our own code.
- Ignore warning `NU1510` "PackageReference SomePackage will not be pruned. Consider removing this package from your dependencies, as it is likely unnecessary". Super noisy, we really don't care for tests.
- Had to change the exclusions for `CallTargetNativeTests` because the stack trace changed. It looks like the runtime is more aggressively inlining methods, including our `OnMethodBegin` calltarget methods.
- The `debian:buster` repos are EOL 💀 But bumping what we build on will currently mean bumping the min version of GLIBC, which is a breaking change. For now we have switched to using the archive repos instead
- Azure Functions behavior was different, in that we start seeing spans for `Timer ExitApp`. As a workaround for now, we just filter that out of the snapshots:
  - https://github.com/DataDog/dd-trace-dotnet/pull/7391
- `InstrumentationTests` (where we test SSI instrumentation behaviour) were added because the expected behaviour instrumenting preview versions is different in SSI. In SSI, we _won't_ instrument preview .NET versions by default, so we updated the tests to allow for that
  - There's an additional fix pending in `ProcessBasicChecksTests.cs` which should be removed after the .NET 10 GA release
- We ran into issues with XUnit and Verify.XUnit. We ultimately "fixed" it by updating from preview 5 to preview 6, but our initial work invovled updating xunit version, so we stuck with that. This revealed a bunch of sub-issues we had to fix too:
  - https://github.com/DataDog/dd-trace-dotnet/pull/7280
  - https://github.com/DataDog/dd-trace-dotnet/pull/7293
  - https://github.com/DataDog/dd-trace-dotnet/pull/7281

Other minor fixes that are part of this PR just for convenience (to avoid rebuilding the VMs again)

- Add explicit `DOTNET_ROOT` variables, both on the Windows VMs and in the docker images
- Fix `'as' has wrong casing` in docker images
- The .NET Framework 3.5 installer (required by WIX) was broken (link no longer works). Switched to using a `dism` based approach.

## Test coverage

All tests should be running in .NET 10 by default. We should make sure we do a run that tests e.g. profiler too, as that doesn't seem to be running in the PR branch runs currently.

## Other details

https://datadoghq.atlassian.net/browse/LANGPLAT-632

### DIFF
--- a/.azure-pipelines/ultimate-pipeline.yml
+++ b/.azure-pipelines/ultimate-pipeline.yml
@@ -120,7 +120,7 @@ variables:
   DD_LOGGER_DD_TAGS: test.configuration.job:$(System.JobDisplayName)
   DD_LOGGER_ENABLED: true
   DD_COLLECTOR_CPU_USAGE: true
-  ToolVersion: 3.25.0
+  ToolVersion: 3.26.0
   # .NET SDK performance optimization variables
   DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1
   DOTNET_CLI_TELEMETRY_OPTOUT: 1
@@ -3275,10 +3275,6 @@ stages:
     pool:
       name: azure-managed-windows-x64-1
 
-    # Enable the Datadog Agent service for this job
-    services:
-      dd_agent: dd_agent_no_pull
-
     steps:
     - template: steps/clone-repo.yml
       parameters:
@@ -3289,6 +3285,28 @@ stages:
 
     - script: tracer\build.cmd SetupExplorationTests --ExplorationTestUseCase $(explorationTestUseCase) --ExplorationTestName $(explorationTestName)
       displayName: SetupExplorationTest $(explorationTestUseCase) $(explorationTestName)
+
+    - powershell: |
+        dotnet build -c Release $(System.DefaultWorkingDirectory)\tracer\tools\MockAgent
+        $serviceName = "MockAgent"
+        $exePath = "$(System.DefaultWorkingDirectory)\tracer\tools\MockAgent\bin\release\net9.0\MockAgent.exe"
+
+        # Check if service already exists
+        if (Get-Service -Name $serviceName -ErrorAction SilentlyContinue) {
+            Write-Host "Service $serviceName already exists. Stopping and removing..."
+            Stop-Service -Name $serviceName -Force -ErrorAction SilentlyContinue
+            sc.exe delete $serviceName | Out-Null
+        }
+
+        # Install the service
+        sc.exe create $serviceName binPath= "`"$exePath`"" start= auto DisplayName= "Datadog Mock Agent"
+
+        # Start the service
+        Start-Service -Name $serviceName
+
+        # Verify
+        Get-Service -Name $serviceName
+      displayName: Start Mock Agent
 
     - script: tracer\build.cmd RunExplorationTests --ExplorationTestUseCase $(explorationTestUseCase) --ExplorationTestName $(explorationTestName)
       displayName: RunExplorationTest $(explorationTestUseCase) $(explorationTestName)
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -27,6 +27,84 @@
 
 
 
+
+## [Release 3.25.0](https://github.com/DataDog/dd-trace-dotnet/releases/tag/v3.25.0)
+
+## Summary
+
+- Various bug fixes and optimisations
+
+## Changes
+
+### Tracer
+* Implement "enahanced telemetry reporting" (#7360)
+* Don't call `ToUpperInvariant` if we don't need to (#7393)
+* Handle TypeLoadException when retrieving HostingEnvironment (#7417)
+* Add missing `GetDictionary()` method to `ConfigurationBuilder` (#7421)
+* More `ConfigurationBuilder` refactoring to avoid incorrect usages (#7422)
+* [Tracer] fix: Re-use runtime metrics writer resources to limit memory growth (#7434)
+
+### Continuous Profiler
+* [Profiler] Make `timer_create`-based CPU profiler default (#7322)
+* Revert "[Profiler] Make `timer_create`-based CPU profiler default (#7322)" (#7427)
+
+### Debugger
+* Disable debugger initialization if it's not enabled initially (#7398)
+
+### Serverless
+* [Azure Functions] Remove all Serverless Mini-Agent (SMA) code (#7389)
+
+### ASM
+* [IAST] Fix NRE in Native IAST ModuleInfo->GetFullName() (#7404)
+
+### Fixes
+* fix: Clean up small C# and C++ build warnings (#7363)
+* chore: update malformed telemetry name (#7402)
+* Get process start info variables safely (#7423)
+* Fix null reference exception in proxy spans (#7428)
+
+### Build / Test
+* [CI] Creates PR Label Based Docker Image (#7337)
+* [Test Package Versions Bump] Updating package versions (#7315)
+* Include the target framework in our `InstrumentationTests` (#7347)
+* Ignore rejit timeout errors in smoke tests and report a metric (#7370)
+* fix: Make OTLP metrics snapshot testing more stable (#7374)
+* Add final job for blocking merges of PRs (#7376)
+* Fix debug symbol upload in release job (#7377)
+* Fix Windows SSI release (#7379)
+* Fix null ref type warning in Nuke (#7382)
+* Fix creation of label-based artifacts for system tests (#7385)
+* Filter our exit span from Azure Functions tests (#7391)
+* Fix variable generation for `linux-arm64` unit tests (#7394)
+* Add more alpine checks for native library (#7395)
+* [Test Package Versions Bump] Updating package versions (#7396)
+* Fix ASM WAF unit tests (#7397)
+* [SINT-3848] 🔒 Pin GitHub Actions to specific SHA versions for enhanced security (#7400)
+* Improve PR labelling and release note generation (#7401)
+* [Native] Add native string unit tests (#7407)
+* Add `Os`, `OsVersion`, and `Tfm` to CI flaky-smoke-test metrics (#7409)
+* Scrape native tracer stats from logs in smoke tests and track in Datadog (#7410)
+* Update `tracer_home_smoke_tests` stage to use MS hosted windows images (#7412)
+* Enable runtime metrics for macrobenchmarks (#7414)
+* Try to fix dynamic instrumentation tests (#7415)
+* Make the `Instrumentation.cs` file owned by everyone (#7416)
+* Try to fix more flake in fleet installer IIS tests (#7418)
+* Bump the gh-actions-packages group across 1 directory with 5 updates (#7429)
+* [Test Package Versions Bump] Updating package versions (#7430)
+
+### Miscellaneous
+* Fix minor warnings and issues and improvements (#7390)
+* refactor(libdatadog): namespaces and file organization, error code and typo (#7405)
+* Don't write debug logs to console for hands off config (#7392)
+* chore(trace-exporter): disable data pipeline when SSI is enabled (#7386)
+* [Service Disc] Change type layout to make it x86 safe (#7387)
+* [HandsOff Config] Read sources from libdatadog (#7282)
+* Refactor telemetry in `ConfigurationBuilder` (#7354)
+* [Native] Added DBG macro that checks for IsDebugEnabled before logging a debug trace (#7406)
+* refactor(libdatadog): error handling and trace exporter response management (#7411)
+
+[Changes since 3.24.1](https://github.com/DataDog/dd-trace-dotnet/compare/v3.24.1...v3.25.0)
+
 ## [Release 3.24.0](https://github.com/DataDog/dd-trace-dotnet/releases/tag/v3.24.0)
 
 ## Summary
--- a/profiler/src/ProfilerEngine/Datadog.Linux.ApiWrapper/CMakeLists.txt
+++ b/profiler/src/ProfilerEngine/Datadog.Linux.ApiWrapper/CMakeLists.txt
@@ -2,7 +2,7 @@
 # Project definition
 # ******************************************************
 
-project("Datadog.Linux.ApiWrapper" VERSION 3.25.0)
+project("Datadog.Linux.ApiWrapper" VERSION 3.26.0)
 
 # ******************************************************
 # Compiler options
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/CMakeLists.txt
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/CMakeLists.txt
@@ -2,7 +2,7 @@
 # Project definition
 # ******************************************************
 
-project("Datadog.Profiler.Native.Linux" VERSION 3.25.0)
+project("Datadog.Profiler.Native.Linux" VERSION 3.26.0)
 
 option(RUN_ASAN "Build with Clang Undefined-Behavior Sanitizer" OFF)
 option(RUN_UBSAN "Build with Clang Undefined-Behavior Sanitizer" OFF)
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native.Windows/Resource.rc
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native.Windows/Resource.rc
@@ -62,8 +62,8 @@ END
 
 // ------- version info -------------------------------------------------------
 VS_VERSION_INFO VERSIONINFO
-FILEVERSION             3,25,0,0
-PRODUCTVERSION          3,25,0,0
+FILEVERSION             3,26,0,0
+PRODUCTVERSION          3,26,0,0
 FILEFLAGSMASK           VS_FF_PRERELEASE
 FILEOS                  VOS_NT
 FILETYPE                VFT_DLL
@@ -74,12 +74,12 @@ BEGIN
         BEGIN
             VALUE "CompanyName", "Datadog"
             VALUE "FileDescription", "Continuous Profiler for .NET Applications"
-            VALUE "FileVersion", "3.25.0.0"
+            VALUE "FileVersion", "3.26.0.0"
             VALUE "InternalName", "Native Profiler Engine"
             VALUE "LegalCopyright", "(c) Datadog 2020-2022"
             VALUE "OriginalFilename", "Datadog.Profiler.Native.dll"
             VALUE "ProductName", "Continuous Profiler for .NET Applications"
-            VALUE "ProductVersion", "3.25.0.0"
+            VALUE "ProductVersion", "3.26.0.0"
         END
     END
     BLOCK "VarFileInfo"
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native/dd_profiler_version.h
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native/dd_profiler_version.h
@@ -3,4 +3,4 @@
 
 #pragma once
 
-constexpr auto PROFILER_VERSION = "3.25.0";
+constexpr auto PROFILER_VERSION = "3.26.0";
--- a/profiler/src/ProfilerEngine/ProductVersion.props
+++ b/profiler/src/ProfilerEngine/ProductVersion.props
@@ -5,7 +5,7 @@
 
   <!-- * * * * * * * * * * * INPUTS. Update this section EVERY time the component is shipped/released! * * * * * * * * * * *                -->
   <PropertyGroup>
-    <ProductVersion>3.25.0</ProductVersion>
+    <ProductVersion>3.26.0</ProductVersion>
   </PropertyGroup>
   <!-- * * * * * * * * * * * END OF INPUTS.  * * * * * * * * * * *                                                                          -->
 
--- a/shared/src/Datadog.Trace.ClrProfiler.Native/CMakeLists.txt
+++ b/shared/src/Datadog.Trace.ClrProfiler.Native/CMakeLists.txt
@@ -12,7 +12,7 @@ endif()
 # Project definition
 # ******************************************************
 
-project("Datadog.Trace.ClrProfiler.Native" VERSION 3.25.0)
+project("Datadog.Trace.ClrProfiler.Native" VERSION 3.26.0)
 
 if (UNIVERSAL)
    find_package(GlibcCompat REQUIRED)
--- a/shared/src/Datadog.Trace.ClrProfiler.Native/Resource.rc
+++ b/shared/src/Datadog.Trace.ClrProfiler.Native/Resource.rc
@@ -57,8 +57,8 @@ VS_VERSION_INFO VERSIONINFO
 #else
  FILEFLAGS 0x0L
 #endif
- FILEVERSION             3,25,0,0
- PRODUCTVERSION          3,25,0,0
+ FILEVERSION             3,26,0,0
+ PRODUCTVERSION          3,26,0,0
  FILEOS                  VOS_NT
  FILETYPE                VFT_DLL
 BEGIN
@@ -68,12 +68,12 @@ BEGIN
         BEGIN
             VALUE "CompanyName", "Datadog"
             VALUE "FileDescription", "Native loader for Datadog .NET APM"
-            VALUE "FileVersion", "3.25.0.0"
+            VALUE "FileVersion", "3.26.0.0"
             VALUE "InternalName", "Native loader"
             VALUE "LegalCopyright", "(c) Datadog 2020-2022"
             VALUE "OriginalFilename", "Datadog.Trace.ClrProfiler.Native.dll"
             VALUE "ProductName", "Native loader for Datadog .NET APM"
-            VALUE "ProductVersion", "3.25.0.0"
+            VALUE "ProductVersion", "3.26.0.0"
         END
     END
     BLOCK "VarFileInfo"
--- a/shared/src/msi-installer/WindowsInstaller.wixproj
+++ b/shared/src/msi-installer/WindowsInstaller.wixproj
@@ -17,9 +17,9 @@
     <IntermediateOutputPath>obj\$(Configuration)\$(Platform)\</IntermediateOutputPath>
     <SuppressPdbOutput>True</SuppressPdbOutput>
     <DefineSolutionProperties>false</DefineSolutionProperties>
-    <OutputName>datadog-dotnet-apm-3.25.0-$(Platform)</OutputName> <!-- -The regex recognizes this line -->
+    <OutputName>datadog-dotnet-apm-3.26.0-$(Platform)</OutputName> <!-- -The regex recognizes this line -->
     <MonitoringHomeDirectory Condition="'$(MonitoringHomeDirectory)' == ''">$(MSBuildThisFileDirectory)..\..\bin\monitoring-home</MonitoringHomeDirectory>
-    <DefineConstants>InstallerVersion=3.25.0;MonitoringHomeDirectory=$(MonitoringHomeDirectory);</DefineConstants>
+    <DefineConstants>InstallerVersion=3.26.0;MonitoringHomeDirectory=$(MonitoringHomeDirectory);</DefineConstants>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|x86' ">
     <DefineConstants>$(DefineConstants);Debug</DefineConstants>
--- a/shared/src/native-src/version.h
+++ b/shared/src/native-src/version.h
@@ -1,3 +1,3 @@
 #pragma once
 
-constexpr auto PROFILER_VERSION = "3.25.0";
+constexpr auto PROFILER_VERSION = "3.26.0";
--- a/tracer/build/_build/Build.cs
+++ b/tracer/build/_build/Build.cs
@@ -65,7 +65,7 @@ partial class Build : NukeBuild
     const int LatestMajorVersion = 3;
 
     [Parameter("The current version of the source and build")]
-    readonly string Version = "3.25.0";
+    readonly string Version = "3.26.0";
 
     [Parameter("Whether the current build version is a prerelease(for packaging purposes)")]
     readonly bool IsPrerelease = false;
--- a/tracer/build/artifacts/dd-dotnet.sh
+++ b/tracer/build/artifacts/dd-dotnet.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-TRACER_VERSION="3.25.0"
+TRACER_VERSION="3.26.0"
 
 # Get the directory of the script
 DIR=$(dirname "$(readlink -f "$0")")
--- a/tracer/samples/AutomaticTraceIdInjection/Log4NetExample/Log4NetExample.csproj
+++ b/tracer/samples/AutomaticTraceIdInjection/Log4NetExample/Log4NetExample.csproj
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Datadog.Trace.Bundle" Version="3.24.0" />
+    <PackageReference Include="Datadog.Trace.Bundle" Version="3.25.0" />
     <PackageReference Include="log4net" Version="2.0.12" />
     <PackageReference Include="log4net.Ext.Json" Version="2.0.8.3" />
   </ItemGroup>
--- a/tracer/samples/AutomaticTraceIdInjection/MicrosoftExtensionsExample/MicrosoftExtensionsExample.csproj
+++ b/tracer/samples/AutomaticTraceIdInjection/MicrosoftExtensionsExample/MicrosoftExtensionsExample.csproj
@@ -6,7 +6,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Datadog.Trace.Bundle" Version="3.24.0" />
+    <PackageReference Include="Datadog.Trace.Bundle" Version="3.25.0" />
     <PackageReference Include="Microsoft.Extensions.Hosting" Version="7.0.1" />
     <PackageReference Include="Serilog.Extensions.Logging" Version="3.1.0" />
     <PackageReference Include="Serilog.Sinks.File" Version="5.0.0" />
--- a/tracer/samples/AutomaticTraceIdInjection/NLog40Example/NLog40Example.csproj
+++ b/tracer/samples/AutomaticTraceIdInjection/NLog40Example/NLog40Example.csproj
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Datadog.Trace.Bundle" Version="3.24.0" />
+    <PackageReference Include="Datadog.Trace.Bundle" Version="3.25.0" />
     <PackageReference Include="NLog" Version="4.0.0" />
   </ItemGroup>
 
--- a/tracer/samples/AutomaticTraceIdInjection/NLog45Example/NLog45Example.csproj
+++ b/tracer/samples/AutomaticTraceIdInjection/NLog45Example/NLog45Example.csproj
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Datadog.Trace.Bundle" Version="3.24.0" />
+    <PackageReference Include="Datadog.Trace.Bundle" Version="3.25.0" />
     <PackageReference Include="NLog" Version="4.5.11" />
   </ItemGroup>
 
--- a/tracer/samples/AutomaticTraceIdInjection/NLog46Example/NLog46Example.csproj
+++ b/tracer/samples/AutomaticTraceIdInjection/NLog46Example/NLog46Example.csproj
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Datadog.Trace.Bundle" Version="3.24.0" />
+    <PackageReference Include="Datadog.Trace.Bundle" Version="3.25.0" />
     <PackageReference Include="NLog" Version="4.6.7" />
   </ItemGroup>
 
--- a/tracer/samples/AutomaticTraceIdInjection/SerilogExample/SerilogExample.csproj
+++ b/tracer/samples/AutomaticTraceIdInjection/SerilogExample/SerilogExample.csproj
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Datadog.Trace.Bundle" Version="3.24.0" />
+    <PackageReference Include="Datadog.Trace.Bundle" Version="3.25.0" />
     <PackageReference Include="Serilog" Version="2.9.0" />
     <PackageReference Include="Serilog.Formatting.Compact" Version="1.1.0" />
     <PackageReference Include="Serilog.Sinks.File" Version="4.1.0" />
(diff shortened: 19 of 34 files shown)


.azure-pipelines/ultimate-pipeline.yml
```diff
@@ -56,10 +56,10 @@ schedules:
 # Global variables
 variables:
   buildConfiguration: Release
-  dotnetCoreSdkLatestVersion: 9.0.203
+  dotnetCoreSdkLatestVersion: 10.0.100-preview.5.25277.114
   # This is required until we're out of rc.
   # After GA we can do a find and replace and get rid of this
-  dotnetCoreSdkLatestVersionShort: 9.0.203
+  dotnetCoreSdkLatestVersionShort: 10.0.100-preview.5.25277.114
```
andrewlock: I think this needs to be `10.0.100-preview.5` or maybe `10.0.100-preview` otherwise you'll get docker pull errors

tracer/Directory.Build.props
```diff
@@ -4,6 +4,9 @@
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <GenerateDocumentationFile>false</GenerateDocumentationFile>
     <LangVersion>latest</LangVersion>
+    <!-- Strong name signature -->
+    <WarningsNotAsErrors>NU1510</WarningsNotAsErrors>
+    <NoWarn>NU1510</NoWarn>
```
andrewlock: 🤔 Where is this error being raised? I'm guessing it's only in the test projects? Either way we probably can and should fix it (though doesn't _have_ to be in this PR)

tracer/build/_build/docker/gitlab/gitlab.windows.dockerfile
```diff
@@ -31,8 +31,8 @@ RUN powershell -Command .\install_wix.ps1 -Version $ENV:WIX_VERSION -Sha256 $ENV
 
 # Install .NET 9
 # To find these links, visit https://dotnet.microsoft.com/en-us/download, click the Windows, x64 installer, and grab the download url + SHA512 hash
-ENV DOTNET_VERSION="9.0.203" \
-    DOTNET_DOWNLOAD_URL="https://builds.dotnet.microsoft.com/dotnet/Sdk/9.0.203/dotnet-sdk-9.0.203-win-x64.exe" \
+ENV DOTNET_VERSION="10.0.100-preview.5.25277.114" \
+    DOTNET_DOWNLOAD_URL="https://builds.dotnet.microsoft.com/dotnet/Sdk/10.0.100-preview.5.25277.114/dotnet-sdk-10.0.100-preview.5.25277.114-win-x64.exe" \
```
andrewlock: You need to update the `DOTNET_SHA512` variable below too - it's displayed on the "thank you for your download" page 😄

tracer/src/Datadog.Trace.Coverage.collector/AssemblyProcessor.cs
```diff
@@ -824,6 +824,12 @@ private string CopyRequiredAssemblies(AssemblyDefinition assemblyDefinition, Tra
         {
             try
             {
+                if (_settings.TracerHome is null)
+                {
+                    _logger.Error("CopyRequiredAssemblies: TracerHome is not set. Cannot copy Datadog.Trace assembly.");
+                    return string.Empty;
+                }
```
andrewlock, line 831: This looks like something that shouldn't be possible, otherwise there's other issues 🤔 Maybe there's refactoring to do here @tonyredondo?

tracer/src/Datadog.Trace.Coverage.collector/FiltersHelper.cs
```diff
@@ -87,15 +87,15 @@ public static bool FilteredBySourceFile(string sourcePath, IReadOnlyList<string>
                             }
                         }
 
-                        instance.AddInclude(Path.IsPathRooted(filter) ? filter.Substring(Path.GetPathRoot(filter).Length) : filter);
+                        instance.AddInclude(Path.IsPathRooted(filter) ? filter.Substring(Path.GetPathRoot(filter)?.Length ?? 0) : filter);
```
andrewlock: I think _technically_ these aren't necessary because `Path.GetPathRoot` will be non null if `Path.IsPathRooted` returns true.

In fact, `Path.GetPathRoot()` internally calls `Path.IsPathRooted` so it's probably better to rewrite these as:

```suggestion
                        instance.AddInclude(Path.GetPathRoot(filter) is { } rooted ? filter.Substring(rooted.Length) : filter);
```

tracer/src/Datadog.Trace.Coverage.collector/FiltersHelper.cs
```diff
@@ -87,15 +87,15 @@ public static bool FilteredBySourceFile(string sourcePath, IReadOnlyList<string>
...
                     }
 
                     return Tuple.Create(instance, (IReadOnlyList<Regex>)lstRegex);
                 });
 
             var matcher = value.Item1;
             // https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.filesystemglobbing.matcher?view=dotnet-plat-ext-6.0
-            var globbingResult = matcher.Match(Path.IsPathRooted(sourcePath) ? sourcePath.Substring(Path.GetPathRoot(sourcePath).Length) : sourcePath).HasMatches;
+            var globbingResult = matcher.Match(Path.IsPathRooted(sourcePath) ? sourcePath.Substring(Path.GetPathRoot(sourcePath)?.Length ?? 0) : sourcePath).HasMatches;
```
andrewlock:
```suggestion
            var globbingResult = matcher.Match(Path.GetPathRoot(sourcePath) is { } rooted ? sourcePath.Substring(rooted.Length) : sourcePath).HasMatches;
```

tracer/test/Datadog.Trace.TestHelpers/MockTracerAgent.cs
```diff
@@ -1403,7 +1403,9 @@ public void Start()
                     {
                         _log("Starting PipeServer " + _pipeName);
                         using var mutex = new ManualResetEventSlim();
+#pragma warning disable CA2025
```
andrewlock: What is this warning? 🤔 It's best to add a comment next to it to make it clear what we're skipping

tracer/test/Datadog.Trace.Tests/Datadog.Trace.Tests.csproj
```diff
@@ -35,7 +35,10 @@
 
   <ItemGroup Condition=" $(TargetFramework.StartsWith('net4')) ">
     <Reference Include="System.Net.Http" />
-    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.6.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.7.0" />
```
andrewlock, line 56: Do we _need_ to update this? 🤔 Just wondering 🤷‍♂️

tracer/test/Datadog.Trace.Tests/HttpOverStreams/TraceAgentHttpHeaderHelperTests.cs
```diff
@@ -41,7 +41,7 @@ public void WriteLeadingHeaders()
                          + fxVersion + "\r\nDatadog-Client-Computed-Top-Level: 1\r\n";
 
             var sb = new StringBuilder();
-            using var textWriter = new StringWriter(sb);
+            var textWriter = new StringWriter(sb);
```
andrewlock: Hmmm, is this flagged by an analyzer now? Because I can see Rider etc telling you to add that back, because it implements `IDisposable` 😅

tracer/test/test-applications/integrations/Samples.HotChocolate/Samples.HotChocolate.csproj
```diff
@@ -24,4 +24,8 @@
     <PackageReference Include="HotChocolate.AspNetCore" Version="$(ApiVersion)" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="3.1.4" />
+  </ItemGroup>
```
andrewlock, line 29: Curious, why didn't we need that before? 🤔

tracer/test/test-applications/integrations/Samples.InstrumentedTests/Samples.InstrumentedTests.csproj
```diff
@@ -52,7 +52,7 @@
     <PackageReference Include="NHibernate" Version="5.5.0" />
   </ItemGroup>
   
-  <ItemGroup Condition="'$(TargetFramework)'=='net9.0' OR '$(TargetFramework)'=='net8.0' OR '$(TargetFramework)'=='net7.0' OR '$(TargetFramework)'=='net6.0'">
+  <ItemGroup Condition="'$(TargetFramework)'=='net10.0' OR '$(TargetFramework)'=='net9.0' OR '$(TargetFramework)'=='net8.0' OR '$(TargetFramework)'=='net7.0' OR '$(TargetFramework)'=='net6.0'">
```
andrewlock: I recently learned about this, which means we won't have to make all these fixes in the future!

```suggestion
  <ItemGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net6.0'))" >
```

tracer/test/test-applications/integrations/Samples.InstrumentedTests/Samples.InstrumentedTests.csproj
```diff
@@ -88,19 +88,19 @@
     <PackageReference Include="RestSharp" Version="108.0.0" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)'=='net9.0' OR '$(TargetFramework)'=='net8.0' OR '$(TargetFramework)'=='net7.0' OR '$(TargetFramework)'=='net6.0'">
+  <ItemGroup Condition="'$(TargetFramework)'=='net10.0' OR '$(TargetFramework)'=='net9.0' OR '$(TargetFramework)'=='net8.0' OR '$(TargetFramework)'=='net7.0' OR '$(TargetFramework)'=='net6.0'">
```
andrewlock:
```suggestion
  <ItemGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net6.0'))" >
```

tracer/test/test-applications/integrations/Samples.InstrumentedTests/Samples.InstrumentedTests.csproj
```diff
@@ -88,19 +88,19 @@
...
     <PackageReference Include="RestSharp" Version="110.2.0" />
   </ItemGroup>  
   
-  <ItemGroup Condition="'$(TargetFramework)'=='net9.0' OR '$(TargetFramework)'=='net8.0' OR '$(TargetFramework)'=='net7.0' OR '$(TargetFramework)'=='net6.0' OR '$(TargetFramework)'=='net5.0'">
+  <ItemGroup Condition="'$(TargetFramework)'=='net10.0' OR '$(TargetFramework)'=='net9.0' OR '$(TargetFramework)'=='net8.0' OR '$(TargetFramework)'=='net7.0' OR '$(TargetFramework)'=='net6.0' OR '$(TargetFramework)'=='net5.0'">
```
andrewlock:
```suggestion
  <ItemGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net5.0'))" >
```

tracer/test/test-applications/integrations/Samples.InstrumentedTests/Samples.InstrumentedTests.csproj
```diff
@@ -88,19 +88,19 @@
...
     <PackageReference Include="Microsoft.EntityFrameworkCore" Version="5.0.8" />
     <PackageReference Include="Microsoft.EntityFrameworkCore.SQLite" Version="5.0.8" />
     <PackageReference Include="Microsoft.EntityFrameworkCore.SqlServer" Version="5.0.8" />
     <PackageReference Include="Dapper.StrongName" Version="2.0.78" />
     <PackageReference Include="Npgsql" Version="7.0.4" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)'=='net9.0' OR '$(TargetFramework)'=='net8.0' OR '$(TargetFramework)'=='net7.0' OR '$(TargetFramework)'=='net6.0' OR $(TargetFramework.StartsWith('net4'))">
+  <ItemGroup Condition="'$(TargetFramework)'=='net10.0' OR '$(TargetFramework)'=='net9.0' OR '$(TargetFramework)'=='net8.0' OR '$(TargetFramework)'=='net7.0' OR '$(TargetFramework)'=='net6.0' OR $(TargetFramework.StartsWith('net4'))">
```
andrewlock:
```suggestion
  <ItemGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net6.0')) OR $(TargetFramework.StartsWith('net4'))">
```

tracer/test/test-applications/integrations/Samples.InstrumentedTests/Vulnerabilities/SqlInjection/EFCoreBaseTests.cs
```diff
@@ -255,10 +255,14 @@ public void GivenACoreDatabase_WhenCallingExecuteScalarWithTainted_Vulnerability
...
 
+    //In .NET 10, method overload resolution is stricter and more sensitive to ambiguous extension methods, particularly when both 
+    //System.Linq.AsyncEnumerable and System.Linq.Queryable are in scope. DbSet<T> implements both IQueryable<T> and is compatible with IAsyncEnumerable<T> 
+    //via EF Core async support. So the compiler can't decide whether you're using IQueryable.Where() or AsyncEnumerable.Where()
+
     [Fact]
     public void GivenACoreDatabase_WhenCallingToListSafe_VulnerabilityIsNotReported()
     {
-        dbContext.Books.Where(x => x.Title == taintedTitle).ToList();
+        _ = ((IQueryable<Book>)dbContext.Books).Where(x => x.Title == taintedTitle).ToList();
```
andrewlock, line 265: _wow_ that seems like a terrible breaking change 😬

tracer/test/test-applications/security/Samples.Security.AspNetCore5/Samples.Security.AspNetCore5.csproj
```diff
@@ -84,6 +84,14 @@
     </Content>
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'net10.0'">
+    <PackageReference Include="Microsoft.AspNetCore.Identity.EntityFrameworkCore" Version="9.0.0" />
```
andrewlock: Aren't there 10.0.0 versions of these packages?

tracer/test/test-applications/security/Samples.Security.AspNetCore5/Samples.Security.AspNetCore5.csproj
```diff
@@ -111,7 +119,7 @@
     <PackageReference Include="System.DirectoryServices.AccountManagement" Version="5.0.0" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)'=='net9.0' OR '$(TargetFramework)'=='net8.0' OR'$(TargetFramework)'=='net7.0' OR '$(TargetFramework)'=='net6.0'">
+  <ItemGroup Condition="'$(TargetFramework)' == 'net10.0' OR '$(TargetFramework)'=='net9.0' OR '$(TargetFramework)'=='net8.0' OR'$(TargetFramework)'=='net7.0' OR '$(TargetFramework)'=='net6.0'">
```
andrewlock:
```suggestion
  <ItemGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net6.0'))" >
```

NachoEchevarria: Cool! Thanks!

tracer/test/test-applications/security/Samples.Security.AspNetCore5/Samples.Security.AspNetCore5.csproj
NachoEchevarria: Yes, you are right, I have updated them. Thanks!

tracer/test/test-applications/integrations/Samples.InstrumentedTests/Vulnerabilities/SqlInjection/EFCoreBaseTests.cs
NachoEchevarria: I was getting this error:
EFCoreBaseTests.cs(262,25): error CS0121: The call is ambiguous between the following methods or properties: 'System.Linq.AsyncEnumerable.Where<TSource>(System.Collections.Generic.IAsyncEnumerable<TSource>, System.Func<TSource, bool>)' and 'System.Linq.Queryable.Where<TSource>(System.Linq.IQueryable<TSource>, System.Linq.Expressions.Expression<System.Func<TSource, bool>>)'

This seems to be the explanation:
This is a breaking change caused by .NET 10’s enhanced overload resolution rules. In .NET 10, method overload resolution is stricter and more sensitive to ambiguous extension methods, particularly when both System.Linq.AsyncEnumerable and System.Linq.Queryable are in scope. DbSet<T> implements both IQueryable<T> and is compatible with IAsyncEnumerable<T> via EF Core async support. So the compiler can't decide whether you're using IQueryable.Where() or AsyncEnumerable.Where().

In previous versions, it was resolving it automatically as Queryable so I made that explicit to avoid the error as a quick fix.

tracer/Directory.Build.props
NachoEchevarria: error NU1510: Warning As Error: PackageReference Microsoft.NETCore.App will not be pruned. Consider removing this package from your dependencies, as it is likely unnecessary.

Basically, it's raised everywhere with different assemblies. We literally have hundreds of errors like this. I added this as a quick solution to compile it because it was treated as an error by default, but, yes, we should, at some point review them.

I will remove the <NoWarn>NU1510</NoWarn> though. I added it to avoid get flooded by these warnings, but we should probably keep them so we don't forget to check them.

https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/10.0/nu1510-pruned-references

tracer/build/_build/docker/gitlab/gitlab.windows.dockerfile
NachoEchevarria: Done. Thanks!!!

tracer/src/Datadog.Trace.Coverage.collector/AssemblyProcessor.cs
NachoEchevarria: Yes, it seems that .NET 10 adds some nullability checks by default that previously were not checked. In this case, the code will not compile without this check since TracerHome is defined as public string? TracerHome

Possible null reference argument for parameter 'path1' in 'string Path.Combine(string path1, string path2, string path3)'.

tracer/test/test-applications/integrations/Samples.InstrumentedTests/Vulnerabilities/SqlInjection/EFCoreBaseTests.cs
andrewlock: Ah, I think it's this problem : https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/10.0/asyncenumerable#recommended-action

I suggest we use the recommended action described there instead of this approach 🙂

tracer/Directory.Build.props
andrewlock: > I added it to avoid get flooded by these warnings, but we should probably keep them so we don't forget to check them.

Yeah, I think we should def fix them, doesn't have to be in this PR though. The link you added has a nice example of how to do it: https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/10.0/nu1510-pruned-references#recommended-action

Though it seems weird tbh... AFAICT, we don't add `Microsoft.NETCore.App` _anywhere_ as a PackageReference 😕 which is... annoying 😅

tracer/src/Datadog.Trace.Coverage.collector/AssemblyProcessor.cs
andrewlock: Yeah, totally agree, the check is required, but it suggests that refactoring would be useful to avoid the _need_ for this check (because basically this is an error scenario that _shouldn't_ happen 😅) But again, not required for this PR, just something for Tony to consider later 🙂

tracer/test/test-applications/integrations/Samples.InstrumentedTests/Vulnerabilities/SqlInjection/EFCoreBaseTests.cs
NachoEchevarria: I tried, but did not work, it looks like both System.Linq.Queryable and System.Linq.AsyncEnumerable are part of the platform assemblies. Adding a reference to System.Linq.AsyncEnumerable and/or excluding System.Linq.Async (as suggested in the link) did not work either.

.azure-pipelines/ultimate-pipeline.yml
NachoEchevarria: Updated. Thanks!

tracer/test/Datadog.Trace.Tests/HttpOverStreams/TraceAgentHttpHeaderHelperTests.cs
NachoEchevarria: I was getting a "Ensure tasks using 'IDisposable' instances complete before the instances are disposed" error. I have updated the code with a better solution. Thanks!

tracer/test/test-applications/integrations/Samples.HotChocolate/Samples.HotChocolate.csproj
NachoEchevarria: 'HotChocolate.AspNetCore' with identity 'HotChocolate.AspNetCore, Version=12.18.0.0, Culture=neutral, PublicKeyToken=null' uses 'Microsoft.Extensions.DependencyInjection.Abstractions, Version=3.1.4.0, Culture=neutral, PublicKeyToken=adb9793829ddae60' which has a higher version than referenced assembly 'Microsoft.Extensions.DependencyInjection.Abstractions' with identity 'Microsoft.Extensions.DependencyInjection.Abstractions, Version=3.1.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60' 

According to chatGPT:
🧠 Root Cause: SDK Resolution Behavior Changed in .NET 10 Builds
While you're still targeting .NET Core 3.1 in one of your target frameworks, the upgrade to the .NET 10 SDK (preview) changed how project builds are evaluated and restored across multi-targeted builds.
⚠️ In .NET SDKs 6–9
Each target framework (e.g., netcoreapp3.1, net5.0, etc.) was mostly restored and built in isolation.
Version mismatches like between 3.1.0.0 and 3.1.4.0 were often silently resolved or obscured if the runtime assembly (from the shared framework) provided the newer one.
NuGet would often resolve a "best match" during restore, depending on implicit dependencies.

🔥 In .NET SDK 10 (preview), Behavior Tightened:
Overload resolution and assembly version matching are stricter.
NuGet package resolution behavior in multi-targeted projects is more consistent but less forgiving of mismatches.
Build logic more rigorously detects when a package (like HotChocolate.AspNetCore) references a higher version of a shared dependency (Microsoft.Extensions.DependencyInjection.Abstractions, v3.1.4) than your project declares (v3.1.0).
The compiler now refuses to compile in that case (CS1705), whereas earlier SDKs might silently accept it at build time but risk runtime failures.

tracer/test/Datadog.Trace.TestHelpers/MockTracerAgent.cs
NachoEchevarria: Ensure tasks using 'IDisposable' instances complete before the instances are disposed. Looks like previously, this would not trigger an error but now they are more strict.

Probably we would need to find a better solution for this. I just was trying it to compile, a better solution is pending.
Added a comment.

tracer/test/Datadog.Trace.Tests/Datadog.Trace.Tests.csproj
NachoEchevarria: NU1605: Warning As Error: Detected package downgrade: System.Diagnostics.DiagnosticSource from 4.7.0 to 4.6.0. Reference the package directly from the project to select a different version. 
7>C:\CommonFolder\shared\repos\dd-trace-6\tracer\test\Datadog.Trace.Tests\Datadog.Trace.Tests.csproj : error NU1605:  Datadog.Trace.Tests -> System.Data.SqlClient 4.8.2 -> System.Diagnostics.DiagnosticSource (>= 4.7.0) 
7>C:\CommonFolder\shared\repos\dd-trace-6\tracer\test\Datadog.Trace.Tests\Datadog.Trace.Tests.csproj : error NU1605:  Datadog.Trace.Tests -> System.Diagnostics.DiagnosticSource (>= 4.6.0)
7>Done building project "Datadog.Trace.Tests.csproj" -- FAILED.

Similar to the HotChocolate.AspNetCore issue below.

.azure-pipelines/ultimate-pipeline.yml
```diff
@@ -1667,7 +1669,7 @@ stages:
       displayName: 'Initialize LocalDB'
       workingDirectory: $(Build.Repository.LocalPath)
 
-    - script: tracer\build.cmd CompileTrimmingSamples BuildWindowsIntegrationTests BuildWindowsRegressionTests -Framework $(framework) --code-coverage-enabled $(CodeCoverageEnabled)
+    - script: tracer\build.cmd CompileTrimmingSamples BuildWindowsIntegrationTests BuildWindowsRegressionTests -Framework $(framework) --code-coverage-enabled $(CodeCoverageEnabled) -PrintDriveSpace true
```
andrewlock:
```suggestion
    - script: tracer\build.cmd CompileTrimmingSamples BuildWindowsIntegrationTests BuildWindowsRegressionTests -Framework $(framework) --code-coverage-enabled $(CodeCoverageEnabled)
```

.gitlab-ci.yml
```diff
@@ -36,7 +36,7 @@ build:
         -e AWS_NETWORKING=true `
         -e SIGN_WINDOWS=true `
         -e NUGET_CERT_REVOCATION_MODE=offline `
-        registry.ddbuild.io/images/mirror/datadog/dd-trace-dotnet-docker-build:latest `
+        datadog/dd-trace-dotnet-docker-build:dotnet10 `
```
andrewlock: We need to create a new mirrored version for this, I've created a PR here: https://github.com/DataDog/images/pull/7514

tracer/Directory.Build.props
```diff
@@ -4,6 +4,8 @@
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <GenerateDocumentationFile>false</GenerateDocumentationFile>
     <LangVersion>latest</LangVersion>
+    <!-- Strong name signature -->
```
andrewlock: My gut feeling is that we should probably move this into the `tracer/test/Directory.Build.props` directory, so that it doesn't apply to `tracer/src` projects, because we probably _do_ want to address the warnings there, it's just in the tests that we don't care 🤔 We can do that later though if it's preferable

```suggestion
    <!-- PackageReference X will not be pruned. Consider removing this package from your dependencies, as it is likely unnecessary.
 -->
```

tracer/build/_build/TargetFrameworkExtensions.cs
```diff
@@ -1,25 +1,40 @@
...
-            ("net9.0" or "net8.0", "net7.0") => true,
-            ("net9.0", "net8.0") => true,
-            _ => false,
-        };
+    {
+        var source = (string)instance;
+        var compareTo = (string)target;
+
+        if (IgnoredFrameworks.Contains(source) || IgnoredFrameworks.Contains(compareTo))
+            return false;
```
andrewlock: This is a much better approach, thanks! 😄 

nit:
```suggestion
        if (IgnoredFrameworks.Contains(source) || IgnoredFrameworks.Contains(compareTo))
        {
            return false;
        }
```

tracer/src/Datadog.Trace.Tools.Runner/Datadog.Trace.Tools.Runner.csproj
```diff
@@ -157,6 +157,7 @@
     <PackageReference Include="NETStandard.Library" Version="2.0.3" />
     <PackageReference Include="System.Runtime.InteropServices" Version="4.3.0" />
     <PackageReference Include="System.Management" Version="5.0.0" />
+    <PackageReference Include="System.Security.Principal.Windows" Version="5.0.0" />
```
andrewlock: Is this one really necessary? I wonder if it was added accidentally? 🤔

tracer/test/Datadog.Trace.Tools.Runner.Tests/Datadog.Trace.Tools.Runner.Tests.csproj
```diff
@@ -1,9 +1,10 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.1;netcoreapp3.0;netcoreapp3.1;net5.0;net6.0;net7.0;net8.0;net9.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.1;netcoreapp3.0;netcoreapp3.1;net5.0;net6.0;net7.0;net8.0;net9.0;net10.0</TargetFrameworks>
 
     <RunAnalyzersDuringBuild>true</RunAnalyzersDuringBuild>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <NoWarn>NU1608;NU1510</NoWarn>
```
andrewlock, line 7: I think we actually _do_ want to fix the warnings in this project (except for .NET Core 2.1, where we _can't_)

tracer/test/test-applications/azure-functions/Samples.AzureFunctions.V4Isolated.AspNetCore/Samples.AzureFunctions.V4Isolated.AspNetCore.csproj
```diff
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net6.0;net7.0;net8.0;net9.0</TargetFrameworks>
+    <!-- TODO: Failing compilation in CI: error : Invalid combination of TargetFramework and AzureFunctionsVersion is set.-->
```
andrewlock: Is this still the case? 🤔

andrewlock (approved):
We're so close I think!

tracer/test/Datadog.Trace.Tools.Runner.Tests/Datadog.Trace.Tools.Runner.Tests.csproj
andrewlock: Oh wait, this is the test, ignore me

tracer/src/Datadog.Trace.Tools.Runner/Datadog.Trace.Tools.Runner.csproj
NachoEchevarria: It is not required. I have updated the code. Thanks!

tracer/Directory.Build.props
andrewlock: Fixed in a subsequent commit (moved to ignore it in the `test/` folder, and keep it in the `src/` folder

tracer/test/test-applications/azure-functions/Samples.AzureFunctions.V4Isolated.AspNetCore/Samples.AzureFunctions.V4Isolated.AspNetCore.csproj
NachoEchevarria: Removed!

.azure-pipelines/steps/install-dotnet.yml
```diff
@@ -53,6 +53,13 @@ steps:
     version: 8.0.x
   retryCountOnTaskFailure: 5
 
+- task: UseDotNet@2
+  displayName: install dotnet core ${{ parameters.packageType }} 8.0
```
lucaspimentel:
```suggestion
  displayName: install dotnet core ${{ parameters.packageType }} 9.0
```

tracer/build/_build/docker/gitlab/install_net35.ps1
```diff
@@ -3,98 +3,13 @@
...
-        # Trying to get the patch with HTTPS yields:
-        # 2022-02-24T15:53:44.6586746+00:00: curl: (60) schannel: SNI or certificate check failed: SEC_E_WRONG_PRINCIPAL (0x80090322) - The target principal name is incorrect.
-        patch = "http://download.windowsupdate.com/c/msdownload/update/software/updt/2020/01/windows10.0-kb4534119-x64_a2dce2c83c58ea57145e9069f403d4a5d4f98713.msu";
-        patchsha256 = "0287ba4106ba5462ba542e0124c4211f50bbce1123167375f125ea27b8f48632"
-        expandedpatch = "windows10.0-kb4534119-x64.cab"
-    };
-    1909 = @{
-    ## Taken from the mcr.microsoft.com/dotnet/framework/runtime:3.5 Dockerfile:
-    ## https://github.com/microsoft/dotnet-framework-docker/blob/abc2ca65b28058f7c71ec8cd8763a8fbf2a9c03f/3.5/runtime/windowsservercore-1909/Dockerfile
+# Install .NET Framework 3.5
```
lucaspimentel, line 10: .NET Framework 3.5 ?? 😅

tracer/test/test-applications/integrations/LogsInjection.ILogger/LogsInjection.ILogger.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFrameworks>net48;netcoreapp2.1;netcoreapp3.0;netcoreapp3.1;net5.0;net6.0;net7.0;net8.0;net9.0</TargetFrameworks>
+    <TargetFrameworks>net48;netcoreapp2.1;netcoreapp3.0;netcoreapp3.1;net5.0;net6.0;net7.0;net8.0;net9.0;net10.0</TargetFrameworks>
```
lucaspimentel: There are so many of these. Can't we just remove this line (in most test applications, some have custom TFMs) so it picks up the shared one in `tracer/test/test-applications/Directory.Build.props`?

tracer/test/Datadog.Trace.Tests/Util/StopwatchTests.cs
```diff
@@ -23,10 +23,15 @@ public void ComputesCorrectTimespan()
             sw.Stop();
 
             // Extract the internal ticks
+#if NET6_0_OR_GREATER
+            // Since at least .NET 6, Stopwatch.GetRawElapsedTicks is directly exposed as ElapsedTicks
+            // and in .NET 10+ the GetRawElapsedTicks method doesn't exist
+            var stopwatchTicks = sw.ElapsedTicks;
```
lucaspimentel: Why can't we use `Stopwatch.ElapsedTicks` in all TFMs?

lucaspimentel (approved):

tracer/build/_build/docker/gitlab/install_net35.ps1
andrewlock: WIX 🙁

tracer/test/Datadog.Trace.Tests/Util/StopwatchTests.cs
andrewlock: That's a good question, seeing as it's been that way [for 5 years](https://github.com/DataDog/dd-trace-dotnet/pull/756) 😅 But as far as I can tell, in all cases it calls `GetRawElapsedTicks` so yeah, I agree 🤔

tracer/test/test-applications/integrations/LogsInjection.ILogger/LogsInjection.ILogger.csproj
andrewlock: To be fair, there was only 5, but I've removed them anyway 😄

tracer/test/Datadog.Trace.Tests/Util/StopwatchTests.cs
andrewlock: (and will be addressed in https://github.com/DataDog/dd-trace-dotnet/pull/7390)

tracer/test/test-applications/integrations/LogsInjection.ILogger/LogsInjection.ILogger.csproj
lucaspimentel: Oops, I expected a lot more than 5!

tracer/build/_build/docker/gitlab/install_net35.ps1
lucaspimentel: Oof. Let's add "update WIX" to our list of R&D week projects 😄

gleocadie (approved):

tracer/src/Datadog.Trace.Coverage.collector/AssemblyProcessor.cs
tonyredondo: True, I need to refactor this code.... +1 to the backlog

PR #7173, opened 2025-07-03
Remove all-green jobs

cbeauchesne:
## Summary of changes

Remove all-green jobs

## Reason for change

Right now, regarding the current state of CI, it's not a good fit : as many jobs are retried during PR, and as github action has no way to know that a job has been retried, the all-green jobs must be manually retried., causing friction.

In practice, the all-green jobs is not retried, making the all-green job the main source of "red" pull request. Which is the opposite of what it's meant.

The guild will discuss about implementing the same pattern, but inside the azure pipeline (so it'll be automatically retried).


andrewlock (approved):
Thanks!

PR #7175, opened 2025-07-03
Only enable Schema extraction when DSM explicitly enabled

robcarlan-datadog:
## Summary of changes
DSM enabled by default will soon happen, but we don't want the schema extraction logic to be enabled by default due to performance concerns and the addition of potentially large span tags. This should only be enabled when DSM is explicitly enabled, so that this behaviour works as is for existing DSM clients.


tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/GoogleProtobufTests.cs
```diff
@@ -94,7 +94,7 @@ public async Task NoInstrumentationForGoogleTypes()
     [Trait("Category", "EndToEnd")]
     public async Task OnlyEnabledWithDsm()
     {
-        SetEnvironmentVariable(ConfigurationKeys.DataStreamsMonitoring.Enabled, "0");
+        SetEnvironmentVariable(ConfigurationKeys.DataStreamsMonitoring.Enabled, null);
```
robcarlan-datadog, line 97: Both this test & below should still pass when we introduce the change that the DSM enabled flag defaults to true

PR #7178, opened 2025-07-04
Explicitly disable debug mode as required

andrewlock:
## Summary of changes

Explicitly sets `DD_TRACE_DEBUG=0` if `DebugModeEnabled==false`

## Reason for change

Some tests have different behaviour between debug and non-debug mode.

## Implementation details

Explicitly set `DD_TRACE_DEBUG=0`, instead of just leaving the ambient value

## Test coverage

This is the test, but will also do a "forced debug" run to confirm

## Other details

Removed some explicit setting of `DD_TRACE_DEBUG` from tests, as we generally shouldn't do this (it changes test behaviour, and we have the debug run explicitly to verify both paths).

### DIFF
--- a/tracer/test/Datadog.Trace.TestHelpers/EnvironmentHelper.cs
+++ b/tracer/test/Datadog.Trace.TestHelpers/EnvironmentHelper.cs
@@ -244,7 +244,7 @@ namespace Datadog.Trace.TestHelpers
                 environmentVariables["COR_PROFILER_PATH"] = nativeLoaderPath;
             }
 
-            environmentVariables["DD_TRACE_DEBUG"] = DebugModeEnabled ? "1" "0";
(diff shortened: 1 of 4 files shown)


tracer/test/Datadog.Trace.TestHelpers/EnvironmentHelper.cs
```diff
@@ -244,10 +244,7 @@ public void SetEnvironmentVariables(
                 environmentVariables["COR_PROFILER_PATH"] = nativeLoaderPath;
             }
 
-            if (DebugModeEnabled)
-            {
-                environmentVariables["DD_TRACE_DEBUG"] = "1";
-            }
+            environmentVariables["DD_TRACE_DEBUG"] = DebugModeEnabled ? "1" "0";
```
andrewlock:
```suggestion
            environmentVariables["DD_TRACE_DEBUG"] = (DebugModeEnabled ? "1" "0");
```

PR #7184, opened 2025-07-04
Make sure we can override the `DD_PROFILING_ENABLED` value

andrewlock:
## Summary of changes

- Make sure we can set `DD_PROFILING_ENABLED=0` for the crashtracking tests

## Reason for change

There's a known interaction with the crash-tracking tests and the profiler which can cause the crash tracker to hang. We have set `DD_PROFILING_ENABLED=0` as a workaround for this for some time, but we were still seeing issues with `dd-dotnet`. It turns out that we were accidentally forcing it back on in those tests, so this should fix it

## Implementation details

Instead of explicitly calling `--set-env` in the dd-dotnet entrypoint, set the variable _globally_, so that it can be disabled as required in the crash tracking test here: 

https://github.com/DataDog/dd-trace-dotnet/blob/283c79a11b324511db1dc30cb16b4146d3e93fbd/.azure-pipelines/steps/run-snapshot-test.yml#L157

## Test coverage

This is the test - hopefully all the installer tests pass!


tracer/build/_build/docker/smoke.chiseled.dockerfile
```diff
@@ -74,8 +74,8 @@ ENTRYPOINT ["dotnet", "AspNetCoreSmokeTest.dll"]
 # Because there _is_ no shell in the chiseled containers
 # Also, we can't use env vars/args in the entrypoint, hence the need for separate targets
 FROM installer-base as dd-dotnet-final-linux-x64
-ENTRYPOINT ["/opt/datadog/linux-x64/dd-dotnet", "run", "--set-env", "DD_PROFILING_ENABLED=1","--set-env", "DD_APPSEC_ENABLED=1","--set-env", "DD_TRACE_DEBUG=1", "--", "dotnet", "/app/AspNetCoreSmokeTest.dll"]
+ENTRYPOINT ["/opt/datadog/linux-x64/dd-dotnet", "run", "--", "dotnet", "/app/AspNetCoreSmokeTest.dll"]
```
andrewlock: I'm not sure why, but this seems to break, so reverting it for now, and can revisit. I don't think this one fails so much anyway

```suggestion
ENTRYPOINT ["/opt/datadog/linux-x64/dd-dotnet", "run", "--set-env", "DD_PROFILING_ENABLED=1","--set-env", "DD_APPSEC_ENABLED=1","--set-env", "DD_TRACE_DEBUG=1", "--", "dotnet", "/app/AspNetCoreSmokeTest.dll"]
```

tracer/build/_build/docker/smoke.chiseled.dockerfile
```diff
@@ -74,8 +74,8 @@ ENTRYPOINT ["dotnet", "AspNetCoreSmokeTest.dll"]
...
 
 ###########################################################
 FROM installer-base as dd-dotnet-final-linux-arm64
-ENTRYPOINT ["/opt/datadog/linux-arm64/dd-dotnet", "run", "--set-env", "DD_PROFILING_ENABLED=1","--set-env", "DD_APPSEC_ENABLED=1","--set-env", "DD_TRACE_DEBUG=1", "--", "dotnet", "/app/AspNetCoreSmokeTest.dll"]
+ENTRYPOINT ["/opt/datadog/linux-arm64/dd-dotnet", "run", "--", "dotnet", "/app/AspNetCoreSmokeTest.dll"]
```
andrewlock:
```suggestion
ENTRYPOINT ["/opt/datadog/linux-arm64/dd-dotnet", "run", "--set-env", "DD_PROFILING_ENABLED=1","--set-env", "DD_APPSEC_ENABLED=1","--set-env", "DD_TRACE_DEBUG=1", "--", "dotnet", "/app/AspNetCoreSmokeTest.dll"]
```

PR #7187, opened 2025-07-07
[Profiler] Protect against thread abort

chrisnas:
## Summary of changes
Check if a thread is aborted before trying to access thread local storage

## Reason for change
Crash detected in memory dump

## Implementation details
AbortThread is asynchronous and it is possible to check against `ThreadState` before accessing the data that will rethrow the `ThreadAbortException`.
Even if there is still a tiny race condition (i.e. `Thread.Abort` is called between the test and the call to `IsValueCreated`), calling `ResetAbort` in a dedicated `ThreadAbortException` catch block might have unexpected results for hosts that might use `ThreadAbort` to avoid "timeouted" threads.


andrewlock (approved):
Thanks! Like you say, won't solve it completely, but hopefully that's sufficient 😅

PR #7192, opened 2025-07-07
[tracing] add OTEL quartz autoinstrumentation

chojomok:
## Summary of changes
* Added a new `QuartzDiagnosticObserver` to enable OTEL instrumentation from Quartz.
  * v3 uses a Diagnostic Observer to make Activities. 
  * v4 uses ActivitySource. 
    * v4 snapshot generated from version `4.0.0-preview-20250718-1741` of the library 
* Integration is supported for `v3.1` of Quartz.  
* Requires `DD_TRACE_OTEL_ENABLED=true` for tracing to work.
* Currently not available for .NET framework due to our DiagnosticObserver infrastructure not being compatible with .NET Framework. Will look into expanding support in the future. 
* Changes to Activity5
  * Added setters to the following properties
    * `DisplayName` 
    * `Kind` - this is private but with bytecode we are able to call the setter. 

## Reason for change

This change enables Datadog APM to trace Quartz.NET job scheduling and execution events, providing observability for background job processing. It expands instrumentation coverage for .NET applications using Quartz, addressing a gap in our supported integrations.

## Implementation details

* We create an Diagnostic Observer to listen on "Quartz" so the built in OTEL Instrumentation can be used and transformed into a Datadog Span. 

## Test coverage
* Tests from v3.1 of Quartz to the latest version
* Tests for upcoming v4 from the prerelease is available and tested on locally. 

### Activity Kind Setter
* We used the same implementation as the `opentelemetry-dotnet-contrib` repo which can be found in this [code snippet](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/d22fdd94a54d64b311620c4c8e8f49e652f790b2/src/Shared/ActivityInstrumentationHelper.cs#L10)
* since we don't have access to the classes as the contrib repo, we used bytecode to avoid casting issues.


tracer/build/PackageVersionsGeneratorDefinitions.json
```diff
@@ -1,4 +1,18 @@
...
+    "SampleProjectName": "Samples.Quartz",
+    "NugetPackageSearchName": "Quartz",
+    "MinVersion": "3.1.0",
+    "MaxVersionExclusive": "4.0.0",
+    "SpecificVersions": [
+      "3.*.*"
+    ],
+    "VersionConditions": [{
+      "MinVersion": "4.0.0",
+      "IncludeOnlyTargetFrameworks": ["net8.0","net9.0"]
```
andrewlock: This is for the upcoming (unreleased) version 4, right? Given we don't test that here anyway, maybe we should just remove this to avoid confusion? 🤔 
```suggestion
```

tracer/test/test-applications/integrations/Samples.Quartz/Program.cs
```diff
@@ -0,0 +1,112 @@
...
+        IJobDetail exceptionJob = JobBuilder.Create<ExceptionJob>()
+                                            .WithIdentity("exceptionJob", "group2")
+                                            .Build();
+
+        ITrigger exceptionTrigger = TriggerBuilder.Create()
+                                                  .WithIdentity("exceptionTrigger", "group2")
+                                                  .StartNow()
+                                                  .Build();
+
+        await SchedulerHolder.Scheduler.ScheduleJob(exceptionJob, exceptionTrigger);
```
andrewlock: It seems weird to me to be scheduling a job from inside another job... is that normal? I never did it, but maybe it is 😄

tracer/test/test-applications/integrations/Samples.Quartz/Program.cs
```diff
@@ -0,0 +1,112 @@
...
+                                        .Build();
+
+        ITrigger helloTrigger = TriggerBuilder.Create()
+                                              .WithIdentity("helloTrigger", "group1")
+                                              .StartNow()
+                                              .Build();
+        
+        await scheduler.ScheduleJob(helloJob, helloTrigger);
+        
+        await Task.Delay(TimeSpan.FromSeconds(10));
```
andrewlock: We should try to remove this delay - any time there's a delay in a test/sample because you need to wait for something to happen, we risk introducing flakiness. The best approach is to have something that waits deterministically. 

e.g. create a `TaskCompletionSource<bool> tcs` somewhere. When the final job runs, it can call `tcs.TrySetResult(true)`, and then here you would have 

```csharp
        await TaskCompletionSource.Task;
```

that way you don't have any timing issues 🙂

tracer/test/test-applications/integrations/Samples.Quartz/Program.cs
```diff
@@ -0,0 +1,112 @@
...
+        await scheduler.Start();
+
+        // HelloJob: logs a greeting
+        IJobDetail helloJob = JobBuilder.Create<HelloJob>()
+                                        .WithIdentity("helloJob", "group1")
+                                        .Build();
+
+        ITrigger helloTrigger = TriggerBuilder.Create()
+                                              .WithIdentity("helloTrigger", "group1")
+                                              .StartNow()
```
andrewlock: Just so I understand, with this trigger, this job will only ever run _once_, right?

tracer/test/test-applications/integrations/Samples.Quartz/Program.cs
```diff
@@ -0,0 +1,112 @@
...
+{
+    public static IScheduler Scheduler { get; set; }
+}
+// The original HelloJob
+public class HelloJob : IJob
+{
+#if QUARTZ_4_0
+    async ValueTask IJob.Execute(IJobExecutionContext context)
+    {
+        await Console.Out.WriteLineAsync("Greetings from HelloJob!");
```
andrewlock: I think it would be good for these jobs to create child spans, so that we know that span parenting flows correctly

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/QuartzTests.cs
```diff
@@ -0,0 +1,109 @@
...
+                              .UseFileName(filename);
+
+            // this isn't real
+            await telemetry.AssertIntegrationEnabledAsync(IntegrationId.OpenTelemetry);
+        }
+    }
+
+    private static string GetSuffix(string packageVersion)
+    {
+#if NETCOREAPP3_1
```
andrewlock: As far as I can tell, all the v3 spans are identical, except that `span.kind: internal` is missing from the `V315plusNETCOREAPP31` spans (and I can't see a reason that they _should_ be missing? 🤔 )

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/QuartzTests.cs
```diff
@@ -0,0 +1,109 @@
...
+            settings.AddRegexScrubber(traceIdRegexHigh, "TraceIdHigh: LinkIdHigh");
+            settings.AddRegexScrubber(traceIdRegexLow, "TraceIdLow: LinkIdLow");
+            settings.AddRegexScrubber(_timeUnixNanoRegex, @"time_unix_nano"":<DateTimeOffset.Now>");
+            settings.AddRegexScrubber(_versionRegex, "telemetry.sdk.version: sdk-version");
+            settings.AddRegexScrubber(fireInstanceId, "fire.instance.id: <fire.instance.id>");
+
+            await VerifyHelper.VerifySpans(spans, settings)
+                              .UseFileName(filename);
+
+            // this isn't real
```
andrewlock: What does that mean? 😂

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/QuartzTests.cs
```diff
@@ -0,0 +1,109 @@
...
+            Log.Information(filename);
+            const int expectedSpanCount = 2;
+            var spans = await agent.WaitForSpansAsync(expectedSpanCount);
+
+            using var s = new AssertionScope();
+            spans.Count.Should().Be(expectedSpanCount);
+
+            var myServiceNameSpans = spans.Where(s => s.Service == "samples.quartz.consoleapp");
+
+            ValidateIntegrationSpans(myServiceNameSpans, metadataSchemaVersion: "v0", expectedServiceName: "samples.quartz.consoleapp", isExternalSpan: false);
```
andrewlock: 🤔 _None_ of the spans have that service name. I think this is just a bug from the `launchSettings.json` - we don't use the launchSettings file in CI (and if we are, we shouldn't 😅 ) 
```suggestion
            ValidateIntegrationSpans(spans, metadataSchemaVersion: "v0", isExternalSpan: false);
```

tracer/test/test-applications/integrations/Samples.Quartz/Program.cs
chojomok: I only did this to keep the order of spans in the output consistent, otherwise it can be a 50/50 on which one appears first. 

I did hear there's a config to make the spans show in the snapshot in alphabetical order by span name - so I could use that if this is too weird.

tracer/test/test-applications/integrations/Samples.Quartz/Program.cs
chojomok: yes! I removed the repeat option. In my testing this has been consistent.

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/QuartzTests.cs
chojomok: The code used to transform OTEL Activities to Datadog Spans has some piece of code that is only available for NET 5.0+, which leads to the span tag missing.

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/QuartzTests.cs
chojomok: 😢 - leftovers from a previous iteration. 
I used to have a Quartz Integration ID for the telemetry metric but since we're defaulting to using OTEL's instrumentation, I made a comment pointing out that the Quartz Integration ID isn't reeal.

tracer/test/test-applications/integrations/Samples.Quartz/Program.cs
chojomok: I'll add some via OTEL APIs on the jobs

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/QuartzTests.cs
chojomok: oh and yeah v315plus isn't needed, at first incorrectly thought the different in the span.kind missing was for all for 3.15+, but it's just for netcoreapp 3.1

tracer/test/test-applications/integrations/Samples.Quartz/Program.cs
chojomok: @andrewlock - i tested with adding a parent span and child span and it worked fine. 
The issue is that I need to change the validation code for the spans since the new OTEL spans don't have the expected Quartz tags. 

```
14:39:14 [DBG]   Failed Datadog.Trace.ClrProfiler.IntegrationTests.QuartzTests.SubmitsTraces(packageVersion: "") [10 s]
14:39:14 [DBG]   Error Message:
14:39:14 [DBG]    Result: False
14:39:14 [DBG] Span: {TraceId: 13766735146284015308, SpanId: 10956003715056285648, Name: internal, Resource: OtelChild1, Service: Samples.Quartz}
14:39:14 [DBG] Errors:
14:39:14 [DBG] - tag "component" was expected to have value "Quartz", but the value is "null"
14:39:14 [DBG] - tag "job.group" was expected to be present
14:39:14 [DBG] - tag "job.name" was expected to be present
14:39:14 [DBG] - tag "trigger.group" was expected to be present
14:39:14 [DBG] - tag "trigger.name" was expected to be present
14:39:14 [DBG] - Expected to have no remaining tags, but the following tags were found: ["otel.trace_id"="08e17b1d41b5d902bf0d44ad1f24decc","otel.library.name"="AdditionalActivitySource","otel.status_code"="STATUS_CODE_UNSET"]
```

IMO, since this is using OTEL's instrumentation, the tests made for OTEL should cover the parent/child context propagation part of the test and may not be needed for integration test? 
I could be missing the better way of doing this and if so please let me know!

chojomok: image

chojomok: Also I think this avoids additional complexity of adding customs spans via otel in different TFM

tracer/test/test-applications/integrations/Samples.Quartz/Program.cs
chojomok: :) TIL

tracer/test/Datadog.Trace.TestHelpers/SpanMetadataV0Rules.cs
```diff
@@ -760,5 +760,21 @@ public static Result IsWebRequestV0(this MockSpan span) => Result.FromSpan(span)
                 .IsOptional("_dd.base_service")
                 .MatchesOneOf("component", "HttpMessageHandler", "WebRequest")
                 .Matches("span.kind", "client"));
+
+        public static Result IsQuartzV0(this MockSpan span) => Result.FromSpan(span)
+          .Properties(s => s
+                         .MatchesOneOf(Name,  "internal"))
+          .Tags(s => s
+                    .Matches("span.kind", "internal")
```
chojomok: span.kind doesn't exist for v3.15 on netcoreapp3.1 - need to update v0 and v1

tracer/test/test-applications/integrations/Samples.Quartz/Program.cs
chojomok: jk - just need to update the filter code. Testing with child spans is fine. 
Is there any issues with using Systems.Diagnostic depending on the TFM?
* probably some line of code in the csproj for the dependency?

tracer/test/test-applications/integrations/Samples.Quartz/Properties/launchSettings.json
```diff
@@ -0,0 +1,42 @@
...
+        "COR_PROFILER_PATH": "$(SolutionDir)shared/bin/monitoring-home/osx/Datadog.Trace.ClrProfiler.Native.dylib",
+
+        "CORECLR_ENABLE_PROFILING": "1",
+        "CORECLR_PROFILER": "{846F5F1C-F9AE-4B07-969E-05C26BC060D8}",
+        "CORECLR_PROFILER_PATH": "$(SolutionDir)shared/bin/monitoring-home/osx/Datadog.Trace.ClrProfiler.Native.dylib",
+
+        "DD_DOTNET_TRACER_HOME": "$(SolutionDir)shared/bin/monitoring-home",
+        "DD_TRACE_AGENT_PORT": "8136",
+        "DD_TRACE_DEBUG": "true",
+        "DD_ENV": "Mohammad",
```
bouwkast: Just a nitpick but we usually don't set this (mainly because we have it set globally usually) and then goes to the environment that we expect if/when we run locally

tracer/test/test-applications/integrations/Samples.Quartz/Samples.Quartz.csproj
```diff
@@ -0,0 +1,20 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <!-- System.Runtime.CompilerServices.Unsafe doesn't support netcoreapp2.1-->
+    <!-- https://andrewlock.net/stop-lying-about-netstandard-2-support/-->
+    <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings>
+    <ApiVersion Condition="'$(ApiVersion)' == ''">3.15.0</ApiVersion>
+    <DefineConstants Condition="'$(ApiVersion)'&gt;='4.0.0'">$(DefineConstants);QUARTZ_4_0</DefineConstants>
```
bouwkast, line 8: Nice thanks! Will help when this new version comes out

tracer/test/test-applications/integrations/Samples.Quartz/Jobs/ExceptionJob.cs
```diff
@@ -0,0 +1,25 @@
...
+#if QUARTZ_4_0
+    async ValueTask IJob.Execute(IJobExecutionContext context)
+#else
+    async Task IJob.Execute(IJobExecutionContext context)
+#endif
+    {
+        try
+        {
+            await Console.Out.WriteLineAsync("Doing work...");
+            throw new InvalidOperationException("Something went wrong");
```
bouwkast:
```suggestion
            throw new InvalidOperationException("Expected InvalidOperationException thrown");
```

Beneficial to specify that exception is "expected" as it is easy to get caught up investigating exceptions when they aren't immediately obvious that they are the expected flow.

tracer/test/snapshots/QuartzTestsV4.verified.txt
```diff
@@ -0,0 +1,77 @@
...
...":<DateTimeOffset.Now>,"attributes":{"exception.message":"Something went wrong","exception.stacktrace":"Parameters: refire = False, unscheduleFiringTrigger = False, unscheduleAllTriggers = False \n Quartz.JobExecutionException: Something went wrong\n ---> System.InvalidOperationException: Something went wrong\n   at QuartzSampleApp.ExceptionJob.Quartz.IJob.Execute(IJobExecutionContext context) in /Users/mohammad.islam/DDRepos/dd-trace-dotnet/tracer/test/test-applications/integrations/Samples.Quartz/Program.cs:line 100\n   --- End of inner exception stack trace ---\n   at QuartzSampleApp.ExceptionJob.Quartz.IJob.Execute(IJobExecutionContext context) in /Users/mohammad.islam/DDRepos/dd-trace-dotnet/tracer/test/test-applications/integrations/Samples.Quartz/Program.cs:line 105\n   at Quartz.Core.JobRunShell.Run(CancellationToken cancellationToken) [See nested exception: System.InvalidOperationException: Something went wrong\n   at QuartzSampleApp.ExceptionJob.Quartz.IJob.Execute(IJobExecutionContext context) in /Users/mohammad.islam/DDRepos/dd-trace-dotnet/tracer/test/test-applications/integrations/Samples.Quartz/Program.cs:line 100]","exception.type":"Quartz.JobExecutionException"}}],
```
bouwkast: Will need to do some filtering here in Verify to make it repeatable (note the `/Users/mohammad.islam/DDRepos/` path)

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
```diff
@@ -456,6 +456,7 @@ private static void StartDiagnosticManager()
             else
             {
                 observers.Add(new AspNetCoreDiagnosticObserver());
+                observers.Add(new QuartzDiagnosticObserver());
```
bouwkast, line 488: I think my only main question is whether there is another way than use this so that we can support .NET Framework (e.g., normal instrumentation but that probably has the same caveats as Hangfire?)

I think this was already discussed though, if so probably good to put that conversation somewhere in the code (`QuartzDiagnosticObserver`) or in the PR

tracer/test/test-applications/integrations/Samples.Quartz/Program.cs
andrewlock: Yeah, I would tend towards that personally, as this seems non-standard.

The `VerifySpans()` function takes an orderSpans function so you should be able to use that, and you can pass in a custom order

https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/test/Datadog.Trace.TestHelpers.SharedSource/VerifyHelper.cs#L133-L146

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/QuartzTests.cs
andrewlock: > and I can't see a reason that they should be missing
> which leads to the span tag missing

Interesting, I wonder if this is actually a bug in the OTEL code then, because I _think_ `span.kind` is a required tag 🤔

tracer/test/snapshots/QuartzTestsV4.verified.txt
chojomok: I totally missed this. I can add some filtering to remove starting with my file path. 
```
---> System.InvalidOperationException: Something went wrong
at QuartzSampleApp.ExceptionJob.Quartz.IJob.Execute(IJobExecutionContext context),
      error.type: Quartz.JobExecutionException,
      events: [{"name":"exception","time_unix_nano":<DateTimeOffset.Now>,"attributes":{"exception.message":"Something went wrong","exception.stacktrace":"Parameters: refire = False, unscheduleFiringTrigger = False, unscheduleAllTriggers = False \n Quartz.JobExecutionException: Something went wrong\n ---> System.InvalidOperationException: Something went wrong\n   at QuartzSampleApp.ExceptionJob.Quartz.IJob.Execute(IJobExecutionContext context)
```
should be good

tracer/build/PackageVersionsGeneratorDefinitions.json
chojomok: need to change MaxVersionExclusive to 5
and another entry to specific versions 

   "SpecificVersions": [
      "3.*.*"
"4.*.*"
    ],

chojomok: risk is that it doesn't work and it causes crashes instead of not having instrumentation 
personal - pref- create a separate branch for v4.

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/QuartzTests.cs
chojomok: @zacharycmontoya @bouwkast - possible bug to discuss for activities + < .net 5

tracer/test/Datadog.Trace.TestHelpers/SpanMetadataV0Rules.cs
chojomok: might need to mark as optional

tracer/test/snapshots/QuartzTestsV4.verified.txt
chojomok: actually lets cut up to error.type
```
---> System.InvalidOperationException: Something went wrong
at QuartzSampleApp.ExceptionJob.Quartz.IJob.Execute(IJobExecutionContext context),
      error.type: Quartz.JobExecutionException,

```

chojomok: oh wait this is for v4- Andrew and I discussed that maybe we should not be future proofing for v4 in case something breaks. I'll make a thread to discuss some of the points with the whole team to get a consensus

tracer/test/test-applications/integrations/Samples.Quartz/Program.cs
chojomok: moving forward with not testing for parent/child relationships for the integration test.

tracer/test/test-applications/integrations/Samples.Quartz/Program.cs
chojomok: I had a problem before with the order of the spans - but it looks like now the default order is working fine.
I wonder if it's because I changed from using a delay to a deterministic method of waiting until the spans are ready that may have fixed the result? 
I ran the tests locally a couple of times and it passed.

tracer/build/PackageVersionsGeneratorDefinitions.json
chojomok: actually I think I caused a misunderstanding - we don't really trace anything we're just converting the OTEL made spans into Datadog spans so there's no risk other than our tests failing. 
Customers can enable `DD_TRACE_OTEL_ENABLED` to get spans for v4 and get spans. This is just creating tests for it so our CI would fail if anything, but we just need to update the snapshots.

tracer/test/test-applications/integrations/Samples.Quartz/Program.cs
chojomok: default does not work with fine with v4 - I added a custom order :)

tracer/test/test-applications/integrations/Samples.Quartz/Program.cs
chojomok: added listener to make sure the job executed before closing the app.

tracer/build/PackageVersionsGeneratorDefinitions.json
chojomok: @zacharycmontoya @andrewlock - I updated the snapshot for v4 and our sample app code to work with v4, but should I update the `PackageVersionsGeneratorDefinitions.json`  to include v4? feels like it doesn't do anything so we should just wait. Let me know your thoughts on how to handle this situation.

tracer/test/test-applications/integrations/Samples.Quartz/Properties/launchSettings.json
chojomok: removed!

tracer/test/snapshots/QuartzTestsV4.verified.txt
chojomok: updated! the snapshots and test code to do the filtering

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
chojomok: @bouwkast - it's possible doing a reverse proxy of the DiagnosticSource class since the Quartz library uses it, we know it's available during the runtime. 
Similar to how I created a a hangfire jobfilter (server and client). 
Zach made a card to track the dev work on this -> probably my next project.

bouwkast: 👍 

Sounds good thanks
(and for V4 of Quartz it sounds like we'll have built in Activity support so a non-issue for that major)

tracer/build/PackageVersionsGeneratorDefinitions.json
andrewlock: Oh true, that's a good point. I would say that if you're going to do everything to work with v4, then we may as well update this too 🙂

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/QuartzTests.cs
```diff
@@ -0,0 +1,111 @@
...
+                                        orderSpans: s => s
+                                                       .OrderBy(x => x.Name)
+                                                       .ThenBy(x => x.Resource))
+                              .UseFileName(filename);
+
+            await telemetry.AssertIntegrationEnabledAsync(IntegrationId.OpenTelemetry);
+        }
+    }
+
+    private static string GetSuffix(string packageVersion)
```
andrewlock: nit: totally optional, but "switch expressions" are awesome, and make things much easier to read IMO 😄


```csharp
    private static string GetSuffix(string packageVersion)
    {
        if (string.IsNullOrEmpty(packageVersion))
        {
            return "V4";
        }

        return new Version(packageVersion) switch
        {
            { } v when v >= new Version("4.0.0") => "V4",
    #if NETCOREAPP3_1
            { } v when v >= new Version("3.15.0") => "V315plusNETCOREAPP31",
    #endif
            _ => "V3",
        }
    }
```

andrewlock (approved):
Nice job, once the tests pass (and you've verified the "all packages" test run) then I think we're good! 🎉

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/QuartzTests.cs
chojomok: decision made to add `span.kind: internal` manually for these spans!

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/QuartzTests.cs
chojomok: yeah I should use this more - thanks for the snippet :)

lucaspimentel:
@codex review

tracer/src/Datadog.Trace/Activity/ActivityListener.cs
```diff
@@ -286,5 +309,43 @@ private static void CreateDiagnosticSourceListenerInstance(Type diagnosticListen
...
+                typeof(DuckType).Module,
+                skipVisibility: true);
+
+            var il = dynMethod.GetILGenerator();
+            il.Emit(OpCodes.Ldarg_0);                  // Load the first argument (object)
+            il.Emit(OpCodes.Castclass, runtimeActivityType); // Cast this value to runtime System.Diagnostics.Activity type
+            il.Emit(OpCodes.Ldarg_1);                  // Load the second argument (int)
+            // For value types (enums), we don't use Castclass - int can be directly used as enum value
+            il.Emit(OpCodes.Call, runtimeActivityType.GetProperty("Kind")!.SetMethod!); // Call the setter method
+            il.Emit(OpCodes.Ret); // return
```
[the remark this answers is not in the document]
lucaspimentel: `Activity.Kind` is not a read-only property. It has a private setter:
https://github.com/dotnet/runtime/blob/df1b77dbf71747d87f0fca8014625c63b83f351d/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.cs#L170

tracer/test/test-applications/integrations/Samples.Quartz/Properties/launchSettings.json
```diff
@@ -0,0 +1,40 @@
...
+        "DD_TRACE_AGENT_PORT": "8136",
+        "DD_SERVICE": "Samples.Quartz",
+        "DD_TRACE_DEBUG": "false"
+      },
+      "nativeDebugging": false
+    },
+    "NoDatadog": {
+      "commandName": "Project",
+      "environmentVariables": {
+        "DD_ENV": "mohammad",
```
chojomok: remove this

tracer/test/snapshots/QuartzTestsV315plusNETCOREAPP31.verified.txt
```diff
@@ -0,0 +1,72 @@
+[
+  {
```
chojomok, line 2: This is identical to the v3 snapshot now with the changes to the instrumentation of manually adding a span.kind.
Delete this

tracer/src/Datadog.Trace/Activity/DuckTypes/IActivity5.cs
```diff
@@ -12,11 +12,11 @@ namespace Datadog.Trace.Activity.DuckTypes
 {
     internal interface IActivity5 : IW3CActivity
     {
-        string DisplayName { get; }
+        string DisplayName { get; set; }
```
chojomok, line 15: add these additions to the PR description

tracer/src/Datadog.Trace/Activity/ActivityListener.cs
```diff
@@ -286,5 +309,43 @@ private static void CreateDiagnosticSourceListenerInstance(Type diagnosticListen
 
             return typeBuilder.CreateTypeInfo()?.AsType();
         }
+
+        private static void CreateActivityKindSetter(Type activityType, Type activityKindType)
+        {
+            // Create dynamic method
```
chojomok: simplify the comments here and add this to the PR description

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/QuartzTests.cs
```diff
@@ -0,0 +1,122 @@
...
+        if (string.IsNullOrEmpty(packageVersion))
+        {
+            return Set(out expectedSpanCount, 2, "V3");
+        }
+
+        return new Version(packageVersion) switch
+        {
+            { } v when v >= new Version("4.0.0") => Set(out expectedSpanCount, 3, "V4"),
+#if NETCOREAPP3_1
+            { } v when v >= new Version("3.15.0") => Set(out expectedSpanCount, 2, "V315plusNETCOREAPP31"),
```
chojomok: not needed now that the v3 and v3coreapp snapshots are the same

tracer/test/Datadog.Trace.TestHelpers/SpanMetadataV0Rules.cs
```diff
@@ -760,5 +760,22 @@ public static Result IsWebRequestV0(this MockSpan span) => Result.FromSpan(span)
...
+                .IsPresent("job.group")
+                .IsPresent("job.name")
+                .IsOptional("job.type")
+                .IsOptional("otel.library.name")
+                .IsOptional("otel.library.version")
+                .IsPresent("otel.trace_id")
+                .IsPresent("otel.status_code")
+                .IsOptional("scheduler.id")
+                .IsOptional("scheduler.name")
+                .IsOptional("span.kind")
```
chojomok, line 832: `IsOptional("span.kind")`

this should be always present now that we have code to add this in

tracer/test/Datadog.Trace.TestHelpers/SpanMetadataV1Rules.cs
```diff
@@ -1095,5 +1095,22 @@ public static Result IsWebRequestV1(this MockSpan span) => Result.FromSpan(span)
...
+                .IsPresent("job.group")
+                .IsPresent("job.name")
+                .IsOptional("job.type")
+                .IsOptional("otel.library.name")
+                .IsOptional("otel.library.version")
+                .IsPresent("otel.trace_id")
+                .IsPresent("otel.status_code")
+                .IsOptional("scheduler.id")
+                .IsOptional("scheduler.name")
+                .IsOptional("span.kind")
```
chojomok, line 1173: same as before - this is no longer optional

tracer/test/test-applications/integrations/Samples.Quartz/Properties/launchSettings.json
chojomok: the DD_ENV:mohammad

tracer/src/Datadog.Trace/Activity/ActivityListener.cs
chojomok: just need to simplify the comments for IL stuff now

tracer/build/PackageVersionsGeneratorDefinitions.json
zacharycmontoya: I'd say let's keep it as-is so we only pull 3.x packages and when 4.x comes out, we can use Dependabot to let us know and we can include it then. Can you create a new `Datadog.Dependabot.Quartz.csproj` project file in the `tracer/dependabot/integrations` folder so we can get notified when it's released?

tracer/src/Datadog.Trace/Activity/Handlers/QuartzActivityHandler.cs
```diff
@@ -0,0 +1,100 @@
...
+using static Datadog.Trace.ClrProfiler.AutoInstrumentation.Quartz.QuartzCommon;
+
+namespace Datadog.Trace.Activity.Handlers
+{
+    /// <summary>
+    /// Handles Quartz.NET activities for job scheduling and execution.
+    /// This handler captures Quartz diagnostic events to trace job execution,
+    /// scheduling, and other Quartz-related operations.
+    /// </summary>
+    internal class QuartzActivityHandler : IActivityHandler
```
zacharycmontoya, line 28: Let's also add a note in the `<summary>` that this is responsible for handling Quartz v4.x, whereas earlier Quartz libraries are handled by `tracer/src/Datadog.Trace/DiagnosticListeners/QuartzDiagnosticObserver.cs` and `tracer/src/Datadog.Trace/Activity/Handlers/DefaultActivityHandler.cs`

tracer/src/Datadog.Trace/Activity/Handlers/QuartzActivityHandler.cs
```diff
@@ -0,0 +1,100 @@
...
+            // Find the span and update it before the common handler processes it
+            string key;
+            if (activity is IW3CActivity w3cActivity)
+            {
+                key = w3cActivity.TraceId + w3cActivity.SpanId;
+            }
+            else
+            {
+                key = activity.Id;
+            }
```
zacharycmontoya: I love switch expressions now, so I think this could be improved by writing the statements like this. This also conveys that it will _always_ have a value and can't be null:

```suggestion
            string key = activity switch
            {
                IW3CActivity w3cActivity => w3cActivity.TraceId + w3cActivity.SpanId,
                _ => activity.Id,
            };
```

tracer/src/Datadog.Trace/Activity/DuckTypes/IActivity5.cs
```diff
@@ -12,11 +12,11 @@ namespace Datadog.Trace.Activity.DuckTypes
...
 
         bool IsAllDataRequested { get; set; }
 
-        ActivityKind Kind { get; }
+        ActivityKind Kind { get; set; }
```
zacharycmontoya: Let's remove this. The public interface only has a `get`

tracer/src/Datadog.Trace/Activity/Handlers/QuartzActivityHandler.cs
```diff
@@ -0,0 +1,100 @@
...
+                {
+                    // Update the span's resource name by appending the job name
+                    var originalResourceName = span.ResourceName;
+                    var newResourceName = CreateResourceName(activity.OperationName, jobNameTag.Value);
+                    span.ResourceName = newResourceName;
+                    Log.Debug("Updated span resource name from '{OriginalResourceName}' to '{NewResourceName}' for job '{JobName}'", originalResourceName, newResourceName, jobNameTag.Value);
+                }
+                else
+                {
+                    Log.Debug("No job.name tag found in Activity.Tags");
```
zacharycmontoya: In this log statement and the next one, I suggest adding to the message to indicate why this is important, like: `"Unable to update Quartz resource name: No job.name tag found in Activity.Tags"`

tracer/src/Datadog.Trace/Activity/ActivityListener.cs
```diff
@@ -286,5 +309,43 @@ private static void CreateDiagnosticSourceListenerInstance(Type diagnosticListen
...
+            // - Cast this value to System.Diagnostics.ActivityKind type
+            // - Call the Setter method for System.Diagnostics.ActivityKind.Set(System.Diagnostics.Activity, System.Diagnostics.ActivityKind)
+            // - Ret
+
+            // Fields: on an object
+            // Properties: Kinda like fields, they give you a _method_ under the hood for set/get
+            // activityType.GetProperty("Kind")!.SetMethod => actual (private) method that sets the value for the Kind property
+
+            // Get the actual runtime types from the System.Diagnostics.DiagnosticSource assembly
+            var runtimeActivityType = Type.GetType("System.Diagnostics.Activity, System.Diagnostics.DiagnosticSource", throwOnError: false) ?? activityType;
```
zacharycmontoya: nit: You already have the types since they're the arguments to this method. You should be able to use those rather than call `Type.GetType` again

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/QuartzTests.cs
```diff
@@ -0,0 +1,122 @@
...
+                                                        .OrderBy(x => x.Name)
+                                                        .ThenBy(x => x.Resource)
+                                                        .ThenBy(x => x.Error))
+                              .UseFileName(filename);
+
+            await telemetry.AssertIntegrationEnabledAsync(IntegrationId.OpenTelemetry);
+        }
+    }
+
+    private static string GetSuffix(string packageVersion, out int expectedSpanCount)
```
zacharycmontoya: Since you're trying to return two values, you might want to return a `Tuple<string, int>` rather than return a `string` as one output plus an `out int` parameter. This will allow you to get rid of the `Set` call and just return a Tuple. This would turn the first return call into:

```csharp
return new("V3", 2);
```

or

```csharp
return Tuple.Create("V3", 2);
```

tracer/test/Datadog.Trace.TestHelpers/SpanMetadataV0Rules.cs
zacharycmontoya: If this is no longer optional, then refactoring it to `IsPresent` will assert that it's there (and throw when it's not)

tracer/test/snapshots/QuartzTestsV4.verified.txt
```diff
@@ -0,0 +1,110 @@
+﻿[
+  {
+    TraceId: Id_1,
+    SpanId: Id_2,
+    Name: internal,
```
zacharycmontoya: In v3 the Name was `quartz.job.execute` but now in v4 it's `internal`. Was that intentional?

tracer/test/snapshots/QuartzTestsV4.verified.txt
```diff
@@ -0,0 +1,110 @@
...
+Parameters: refire = False, unscheduleFiringTrigger = False, unscheduleAllTriggers = False 
+Quartz.JobExecutionException: Expected InvalidOperationException thrown
+---> System.InvalidOperationException: Expected InvalidOperationException thrown
+at QuartzSampleApp.Jobs.ExceptionJob.Quartz.IJob.Execute(IJobExecutionContext context),
+      error.type: Quartz.JobExecutionException,
+      events: <error_stack>","exception.type":"Quartz.JobExecutionException"}}],
+      fire.instance.id: <fire.instance.id>,
+      job.group: group2,
+      job.name: exceptionJob,
+      job.type: QuartzSampleApp.Jobs.ExceptionJob, Samples.Quartz,
```
zacharycmontoya, line 23: It looks like v4 has the full name of the type as the job.type (`<Namespace-qualified Type>, <Type's Aseembly>`) whereas v3 had only the Namespace-qualified Type. Was this intentional?

tracer/test/snapshots/QuartzTestsV4.verified.txt
```diff
@@ -0,0 +1,110 @@
...
+    Tags: {
+      env: integration_tests,
+      error.msg: Expected InvalidOperationException thrown,
+      error.stack:
+Parameters: refire = False, unscheduleFiringTrigger = False, unscheduleAllTriggers = False 
+Quartz.JobExecutionException: Expected InvalidOperationException thrown
+---> System.InvalidOperationException: Expected InvalidOperationException thrown
+at QuartzSampleApp.Jobs.ExceptionJob.Quartz.IJob.Execute(IJobExecutionContext context),
+      error.type: Quartz.JobExecutionException,
+      events: <error_stack>","exception.type":"Quartz.JobExecutionException"}}],
```
zacharycmontoya: This scrubbing looks off. Since we're already copying the events into tags, should we just scrub the entire "events" tag for now?

zacharycmontoya (commented):
I left a number of small comments for improvements, but one area I want to clarify is the expected outputs in our snapshots. There are several changes from v3 to v4 so I want to validate that those are all intended before merging.

tracer/test/Datadog.Trace.TestHelpers/SpanMetadataV0Rules.cs
chojomok: I think that's fine right? I only had it as optional before because coreapp 3.1 didn't add it in, but now with the changes all the versions have the snap tags for the snapshots.

tracer/src/Datadog.Trace/Activity/Handlers/QuartzActivityHandler.cs
chojomok: 👍🏾

tracer/src/Datadog.Trace/Activity/DuckTypes/IActivity5.cs
chojomok: ah I see with the IL stuff we don't need this anymore - i thought we did. Thank you!

tracer/src/Datadog.Trace/Activity/Handlers/QuartzActivityHandler.cs
chojomok: No longer using this method, using the following from quartz.common. 
```
        internal static void EnhanceActivityMetadata(IActivity5 activity)
        {
            var displayName = CreateResourceName(activity.DisplayName, activity.Tags.FirstOrDefault(kv => kv.Key == "job.name").Value ?? string.Empty);
            activity.AddTag("operation.name", activity.DisplayName);
            activity.DisplayName = displayName;
        }
```

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/QuartzTests.cs
chojomok: ooooo - good idea

tracer/test/snapshots/QuartzTestsV4.verified.txt
chojomok: yea - v4 is doing this on its own.

tracer/test/snapshots/QuartzTestsV4.verified.txt
chojomok: I need to figure this out. I'm open to scrubbing it all since we don't need to know it, we just need to know we're getting the tag there.

tracer/test/snapshots/QuartzTestsV4.verified.txt
chojomok: - resolved the issue. This was not supposed to happen.

tracer/test/snapshots/QuartzTestsV4.verified.txt
chojomok: looking into this deeper. Something I changed is causing the following to be missing in the new snapshot:
```
      error.stack:
Parameters: refire = False, unscheduleFiringTrigger = False, unscheduleAllTriggers = False 
Quartz.JobExecutionException: Expected InvalidOperationException thrown
---> System.InvalidOperationException: Expected InvalidOperationException thrown
at QuartzSampleApp.Jobs.ExceptionJob.Quartz.IJob.Execute(IJobExecutionContext context),
      error.type: Quartz.JobExecutionException,
      events: <error_stack>","exception.type":"Quartz.JobExecutionException"}}],\
```

Looking into this now

chojomok: I used the wrong 4.0 preview version... 
AHHHHH - ok things are changing too much from the pre-release of v4.

chojomok: ah ok now i see what you're saying. The events tag looks off
```
events: <error_stack>","exception.type":"Quartz.JobExecutionException"}}],
```
I'll scrub it :)

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Quartz/QuartzCommon.cs
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        internal static void SetActivityKind(IActivity5 activity)
+        {
+            ActivityListener.SetActivityKind(activity, GetActivityKind(activity));
+        }
+
+        internal static void EnhanceActivityMetadata(IActivity5 activity)
+        {
+            var displayName = CreateResourceName(activity.DisplayName, activity.Tags.FirstOrDefault(kv => kv.Key == "job.name").Value ?? string.Empty);
```
chojomok: add a debug log here? 
maybe some null handling?

tracer/src/Datadog.Trace/Activity/Handlers/QuartzActivityHandler.cs
```diff
@@ -0,0 +1,73 @@
...
+namespace Datadog.Trace.Activity.Handlers
+{
+    /// <summary>
+    /// Handles Quartz.NET activities for job scheduling and execution.
+    /// This handler captures Quartz diagnostic events to trace job execution,
+    /// scheduling, and other Quartz-related operations.
+    /// This handler is responsible for Quartz v4.x.
+    /// Earlier Quartz library versions are handled by:
+    /// - tracer/src/Datadog.Trace/DiagnosticListeners/QuartzDiagnosticObserver.cs
+    /// - tracer/src/Datadog.Trace/Activity/Handlers/DefaultActivityHandler.cs
```
chojomok, line 26: remove this

tracer/test/test-applications/integrations/Samples.Quartz/Samples.Quartz.csproj
```diff
@@ -0,0 +1,24 @@
...
+    <Nullable>enable</Nullable>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <GenerateTargetFrameworkAttribute>false</GenerateTargetFrameworkAttribute>
+  </PropertyGroup>
+  
+  <ItemGroup>
+    <PackageReference Include="Quartz" Version="$(ApiVersion)" />
+    <PackageReference Include="OpenTelemetry.Exporter.Console" Version="1.12.0" />
```
chojomok: comment out otel package references

tracer/build/PackageVersionsGeneratorDefinitions.json
chojomok: zach and I looked through adding a Dependabot but it gets wiped by the CI. 
We won't look at add it now, but in the future we should look into a way to add libraries for instrumentations we support via OTEL

tracer/src/Datadog.Trace/Activity/ActivityListener.cs
```diff
@@ -286,5 +309,24 @@ private static void CreateDiagnosticSourceListenerInstance(Type diagnosticListen
...
+                typeof(DuckType).Module,
+                skipVisibility: true);
+
+            var il = dynMethod.GetILGenerator();
+            il.Emit(OpCodes.Ldarg_0);                  // Load activity object
+            il.Emit(OpCodes.Castclass, activityType);  // Cast to Activity type
+            il.Emit(OpCodes.Ldarg_1);                  // Load kind int
+            il.Emit(OpCodes.Call, activityType.GetProperty("Kind")!.SetMethod!); // Call Kind property setter
+            il.Emit(OpCodes.Ret);
+            _setKindProperty = (Action<object, int>)dynMethod.CreateDelegate(typeof(Action<object, int>));
```
bouwkast, line 342:
```suggestion
            var activityKindProperty = activityType.GetProperty("Kind");
            if (activityKindProperty is null)
            {
                Log.Warning("Activity Kind property was not found.");
                return;
            }

            if (activityKindProperty?.SetMethod == null)
            {
                Log.Warning("ActivityKind did not have a setter.");
                return;
            }

            // Create dynamic method: (object activity, int kind) => activity.Kind = kind
            var dynMethod = new DynamicMethod(
                "ActivityKindSetter",
                null,
                [typeof(object), typeof(int)],
                typeof(DuckType).Module,
                skipVisibility: true);

            var il = dynMethod.GetILGenerator();
            il.Emit(OpCodes.Ldarg_0);                  // Load activity object
            il.Emit(OpCodes.Castclass, activityType);  // Cast to Activity type
            il.Emit(OpCodes.Ldarg_1);                  // Load kind int
            il.Emit(OpCodes.Call, activityKindProperty.SetMethod); // Call Kind property setter
            il.Emit(OpCodes.Ret);
            _setKindProperty = (Action<object, int>)dynMethod.CreateDelegate(typeof(Action<object, int>));
```

My one concern here is that we have two null-forgiving operators assuming that we have both the `Kind` property and that we can call the setter.

If that isn't the case we will throw.

It _may_ be true in how the function is currently used (as you have it gated behind that functionality) but I think it would be slightly better to just check for those if they are `null` and handle that if that is the case.
From what I can tell the usage of it in `SetActivityKind` already handles that as it is a conditional

tracer/src/Datadog.Trace/Activity/Handlers/QuartzActivityHandler.cs
```diff
@@ -0,0 +1,73 @@
...
+                _ => activity.Id
+            };
+
+            if (key != null && ActivityHandlerCommon.ActivityMappingById.TryRemove(key, out var activityMapping) && activityMapping.Scope.Span is Span span)
+            {
+                Log.Debug("ActivityStopped: Processing span for activity '{ActivityId}'", activity.Id);
+
+                // Finish the span manually
+                EnhanceActivityMetadata((IActivity5)activity);
+                OtlpHelpers.UpdateSpanFromActivity(activity, span);
```
bouwkast:
```suggestion
                // Finish the span manually
                if (activity is IActivity5 activity5)
                {
                    EnhanceActivityMetadata(activity5);
                }

                OtlpHelpers.UpdateSpanFromActivity(activity, span);
```

Just to be safe could we do some pattern matching here instead of a direct cast? IF for some reason the `activity` isn't an `IActivity5` we want to avoid the possible InvalidCastException

tracer/test/test-applications/integrations/Samples.Quartz/Program.cs
```diff
@@ -0,0 +1,101 @@
+﻿using Quartz;
+using Quartz.Impl;
+using Quartz.Impl.Matchers;
+using QuartzSampleApp.Infrastructure;
+using QuartzSampleApp.Jobs;
+using Microsoft.Extensions.DependencyInjection;
+using OpenTelemetry.Resources;
+using OpenTelemetry.Trace;
+using OpenTelemetry;
+using OpenTelemetry.Exporter;
```
bouwkast:
```suggestion
```

I think since they are commented out we don't need these?

bouwkast (changes requested):
Nice!

I have a couple of suggestions to just add some more protections. Requesting changes for that, but feel free to re-request review afterwards

I vaguely remember the tests being slow for this (or was that Hangfire?) did we resolve that?

tracer/build/PackageVersionsGeneratorDefinitions.json
bouwkast:
```suggestion
      "IncludeOnlyTargetFrameworks": ["net8.0","net9.0", "net10.0"]
```

tracer/test/test-applications/integrations/Samples.Quartz/Program.cs
chojomok: yeah looks like it fails some tests unless I remove them.

tracer/src/Datadog.Trace/Activity/Handlers/QuartzActivityHandler.cs
chojomok: sgtm! quartz v4 is only supported for .net 8 and 9 (and 10?) so I don't think we'll run into any issues.

chojomok:
> I vaguely remember the tests being slow for this (or was that Hangfire?) did we resolve that?


@bouwkast 
yeah that was for hangfire - quartz is working fine with the tests. The sample app stops once all the scheduled jobs are finished running.

tracer/test/test-applications/integrations/Samples.Quartz/Program.cs
```diff
@@ -0,0 +1,97 @@
+﻿using Quartz;
+using Quartz.Impl;
+using Quartz.Impl.Matchers;
+using QuartzSampleApp.Infrastructure;
+using QuartzSampleApp.Jobs;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
```
bouwkast:
```suggestion
```
I think these are causing the sample build to fail

bouwkast (approved):
Nice!

I think just one small thing to fix the sample to build again - I think it is actually left over stuff maybe for testing with OpenTelemetry? Not 100% sure though.

But after that gets fixed and the tests pass I think we are good.

chojomok:
there were some additional issues with netcoreapp snapshots.
Since we're only updating IActivity5, netcoreapp 3.1 spans are very lack luster but that's fine for now until we get complaints to look into updating the feature.

PR #7199, opened 2025-07-10
Show correct `agent_uri` path in startup log

andrewlock:
## Summary of changes

Shows the correct `agent_uri` path in the startup logs

## Reason for change

Currently, `agent_uri` _always_ shows the output of `TracerSettings.AgentUri`. However, this is set to a default/dummy value for windows named pipes. This PR ensures we report the correct value in the `DATADOG TRACER CONFIGURATION` log when using named pipes

## Implementation details

Exposed another property to write the "real" `TraceAgentUriBase` value. We can't easily change `AgentUri` to return the correct value because 

1. The named pipe name isn't a valid Uri
2. AgentUri is exposed publicly, so we can't change it without introducing potentially breaking behaviour

If, in the future, we consider introducing a `npipe://` prefix for _setting_ the named pipe, then we may be able to reconsider this, but for now, this is the least risky change while still exposing the correct info.

## Test coverage

Added some unit tests

## Other details

Did some minor cleanup in `NamedPipeClientStreamFactory`

### DIFF
--- a/tracer/src/Datadog.Trace/Agent/StreamFactories/NamedPipeClientStreamFactory.cs
+++ b/tracer/src/Datadog.Trace/Agent/StreamFactories/NamedPipeClientStreamFactory.cs
@@ -14,34 +14,27 @@ namespace Datadog.Trace.Agent.StreamFactories
 {
     internal class NamedPipeClientStreamFactory : IStreamFactory
     {
+        private const string ServerName = ".";
+        private const PipeOptions PipeOptions = System.IO.Pipes.PipeOptions.Asynchronous;
         private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(NamedPipeClientStreamFactory));
 
         private readonly string _pipeName;
-        private readonly string _serverName;
-        private readonly PipeOptions _pipeOptions;
         private readonly int _timeoutMs;
 
         public NamedPipeClientStreamFactory(string pipeName, int timeoutMs)
-            : this(pipeName, ".", PipeOptions.Asynchronous, timeoutMs)
-        {
-        }
-
-        public NamedPipeClientStreamFactory(string pipeName, string serverName, PipeOptions pipeOptions, int timeoutMs)
         {
             _pipeName = pipeName;
-            _serverName = serverName;
-            _pipeOptions = pipeOptions;
             _timeoutMs = timeoutMs;
         }
 
         public string Info()
         {
-            return $@"\\{_serverName}\pipe\{_pipeName}";
+            return $@"\\{ServerName}\pipe\{_pipeName}";
         }
 
         public Stream GetBidirectionalStream()
         {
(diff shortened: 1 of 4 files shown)


tracer/src/Datadog.Trace/Agent/StreamFactories/NamedPipeClientStreamFactory.cs
andrewlock, line 1: The changes in this file aren't required, just some minor cleanup to simplify things by extracting out constants

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -146,7 +146,22 @@ internal ExporterSettings(IConfigurationSource? source, Func<string, bool> fileE
         /// <seealso cref="ConfigurationKeys.AgentPort"/>
         public Uri AgentUri { get; }
 
-#pragma warning disable SA1624 // Documentation summary should begin with "Gets" - the documentation is primarily for public property
+        /// <summary>
+        /// Gets the base Uri where traces will be sent, taking the transport into account. It may be
+        /// different for statsd.
+        /// </summary>
+        /// <seealso cref="AgentUri"/>
+        public string TraceAgentUriBase => TracesTransport switch
```
bouwkast, line 154: Do we want this to be `public`?

bouwkast (approved):
👍

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
andrewlock: Meh, it's not _really_ public, because nothing in _Datadog.Trace_ is directly referenced 😄

andrewlock: We should really clean all that up one day 🙈

lucaspimentel: > We should really clean all that up one day

I have a branch where I started doing that, `lpimentel/all-public-to-internal`, but there are "vendored" Microsoft types that cannot be public or they conflict with the built-in ones. Also there are some files shared with `Datadog.Trace.Manual` which need to be public. Those might need `#if` to switch between public/internal.

I have _another_ branch, #6726, which should fix the vendoring issue by removing most of those Microsoft vendored types when targeting .NET Core 3.1 and above and using the built-in types instead.

lucaspimentel (approved):

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
andrewlock: > I have another branch, https://github.com/DataDog/dd-trace-dotnet/pull/6726, which should fix the vendoring issue by removing most of those Microsoft vendored types when targeting .NET Core 3.1 and above and using the built-in types instead.

I'd like to see this one, plus it's good for serverless 🎉 😉

PR #7200, opened 2025-07-10
Try removing exceptions from the shutdown path

andrewlock:
## Summary of changes

Replace `CancellationTokenSource` with `TaskCompletionSource`

## Reason for change

There's a bug in the runtime, [which we have seen primarily on .NET 9 on arm64](https://github.com/dotnet/runtime/issues/112565), but which we have occasionally seen both on x64 and also .NET FX. The bug appears to be triggered by some combination of:

- The app shutting down
- Rejit ongoing
- An exception is thrown

It's very likely a race condition in the runtime somewhere, but we have started seeing this more commonly in CI since making changes to how we delay loading the native tracer/profiler. It's particularly noticable for short programs.

The crashing cases we have seen recently were all triggered by the `DiscoveryService` cancelling its `FetchConfigurationLoopAsync` by setting `CancellationTokenSource.Cancel()`. This triggers a `TaskCancelledException` to be thrown (as is normal for task cancellation), which is then caught, and the loop exits gracefully.

However, the bug means that _the act of throwing the exception_ triggers a crash 😅 So this PR is about trying to avoid throwing an exception in non-exceptional circumstances (i.e. graceful cancellation). 

## Implementation details

tl;dr; 

- `CancellationTokenSource` => `TaskCompletionSource<bool>`
- `_processExit.IsCancellationRequested` => `_processExit.Task.IsCompleted`
- `Task.Delay(duration, _processExit.Token)` => `Task.WhenAny(_processExit.Task, Task.Delay(duration))`

The semantics are _almost_ the same, and as we're always "cancelling" a _delay_ it's pretty safe; the delay will technically continue in the background but that's not a problem really.

I did a sweep for all the places using a `CancellationTokenSource` in _Datadog.Trace_ (we don't care about the `dd-trace` runner etc, because the profiler isn't attached).

- `DiscoveryService` ✅ Fixed
- `XunitTestMethodRunnerContextCtorV3Integration` ✅ Required for instrumentation, unused
- `Debugger*` ⚠️ _dozens of places_... we may need to look into this in more detail
- `TaskExtensions.SafeWait` ⚠️ Only used in `TestOptimization.Flush`, I'll let @tonyredondo evaluate
- `HardcodedSecretsAnalyzer` ✅ Fixed
- `RemoteConfigurationManager` ✅ Fixed
- `AsyncManualResetEvent.WaitAsync` ⚠️ Only used in `CIVisibilityProtocolWriter`, I'll let @tonyredondo evaluate
- `AsyncUtil.WaitAsync` ⚠️ Used in `AsyncUtil.RunSync()` which we use in the shutdown path, but _shoudln't_ be a massive problem, as we only hit it in a timeout scenario. We probably _could_ change it to be the same though
- `TraceClock` ✅ Fixed

## Test coverage

In general, we're covered by the existing tests. But I hooked up a quick app to check on the first chance exceptions

```csharp
using System;
using System.Threading;

int _firstChanceExceptionCount = 0;

AppDomain.CurrentDomain.FirstChanceException += (sender, e) =>
{
    Interlocked.Increment(ref _firstChanceExceptionCount);
};

AppDomain.CurrentDomain.ProcessExit += (sender, e) =>
{
    Console.WriteLine($"First chance exceptions: {_firstChanceExceptionCount}");
};
```

The app just starts, does nothing, and exits.

Without instrumentation:
```
First chance exceptions: 0
```

With instrumentation (with 3.20.1 release)
```
First chance exceptions: 2
```

With instrumentation (this PR)
```
First chance exceptions: 0
```

## Other details

This almost certainly won't completely solve the problem. For example the [original issue](https://github.com/dotnet/runtime/issues/112565) happened with an `ObjectDisposedException` being thrown from a socket. That could still happen, and isn't resolved by this PR.

There's also a whole separate avenue we could explore related to a Rejit happening at the same time. It's very plausible that's part of the issue. We could look into using `PrepareMethod` to side step that [RuntimeHelpers.PrepareMethod](https://github.com/dotnet/runtime/issues/1711), but that would be a next step. Hopefully this PR will reduce the impact anyway

### DIFF
--- a/tracer/src/Datadog.Trace/Agent/DiscoveryService/DiscoveryService.cs
+++ b/tracer/src/Datadog.Trace/Agent/DiscoveryService/DiscoveryService.cs
@@ -35,7 +35,7 @@ namespace Datadog.Trace.Agent.DiscoveryService
         private readonly int _initialRetryDelayMs;
         private readonly int _maxRetryDelayMs;
         private readonly int _recheckIntervalMs;
-        private readonly CancellationTokenSource _processExit = new();
+        private readonly TaskCompletionSource<bool> _processExit = new();
         private readonly List<Action<AgentConfiguration>> _agentChangeCallbacks = new();
         private readonly object _lock = new();
         private readonly Task _discoveryTask;
@@ -168,7 +168,7 @@ namespace Datadog.Trace.Agent.DiscoveryService
 
             int? sleepDuration = null;
 
-            while (!_processExit.IsCancellationRequested)
+            while (!_processExit.Task.IsCompleted)
             {
                 try
                 {
@@ -194,7 +194,7 @@ namespace Datadog.Trace.Agent.DiscoveryService
 
                 try
                 {
-                    await Task.Delay(sleepDuration ?? _recheckIntervalMs, _processExit.Token).ConfigureAwait(false);
+                    await Task.WhenAny(_processExit.Task, Task.Delay(sleepDuration ?? _recheckIntervalMs)).ConfigureAwait(false);
                 }
                 catch (OperationCanceledException)
                 {
@@ -312,7 +312,7 @@ namespace Datadog.Trace.Agent.DiscoveryService
 
         public Task DisposeAsync()
         {
-            _processExit.Cancel();
+            _processExit.SetResult(true);
             return _discoveryTask;
         }
     }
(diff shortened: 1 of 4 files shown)


zacharycmontoya (approved):
LGTM, I appreciate the detailed PR summary 🙏🏼

lucaspimentel (approved):

PR #7201, opened 2025-07-10
Add agent response handling to native interop

hoolioh:
## Summary of changes

Adds proper handling of the agent response using the libdatadog trace exporter.

## Reason for change

We have observed memory leaks in testing when the data-pipeline tracer exporter is used. Analysis by @hoolioh and @gleocadie identified one potential source as being that memory for the agent response was being allocated, but the managed side was not inspecting this and freeing the memory.

Additionally, we noticed that updates to sampling rules, which are returned by the agent, were not being passed back from libdatadog and used to update the sampler.

## Implementation details

Update the implementation to call `ddog_trace_exporter_response_get_body` and `ddog_trace_exporter_response_free` to retrieve the body and free the response.

The currently exposed `ddog_trace_exporter_response_get_body` is sub-optimal, in that it doesn't provide the length of the body, so we have to enumerate the data to find the length. That will be addressed subsequently by adding a new API, once we have verified this resolves the memory leak issue.

## Test coverage

https://github.com/DataDog/dd-trace-dotnet/pull/7165, which enables data pipeline has been rebased on top of this PR to verify that this passes all our existing tests, but it's possible that we don't have integration tests that cover updating sampling rules...

If everything looks good after #7165 is rebased, we will merge this and deploy to AAS reliability to evaluate if it solves the memory leak issue


tracer/src/Datadog.Trace/LibDatadog/TraceExporter.cs
```diff
@@ -59,6 +64,21 @@ public Task<bool> SendTracesAsync(ArraySegment<byte> traces, int numberOfTraces,
...
+                        {
+                            string? response = Marshal.PtrToStringAnsi(NativeInterop.ExporterResponse.GetBody(responsePtr));
+                            if (!string.IsNullOrEmpty(response))
+                            {
+                                var apiResponse = JsonConvert.DeserializeObject<ApiResponse>(response);
+
+                                _updateSampleRates(apiResponse.RateByService);
+                            }
+
+                            NativeInterop.ExporterResponse.Free(responsePtr);
```
andrewlock: We probably need to wrap this in a try-catch to make sure we always call free?
```suggestion
                            try
                            {
                                string? response = Marshal.PtrToStringAnsi(NativeInterop.ExporterResponse.GetBody(responsePtr));
                                if (!string.IsNullOrEmpty(response))
                                {
                                    var apiResponse = JsonConvert.DeserializeObject<ApiResponse>(response);
    
                                    _updateSampleRates(apiResponse.RateByService);
                                }
                            }
                            catch(Exception ex)
                            {
                                _log.Error(ex, "An error occurred deserializing the response.");
                            }
                            finally
                            {
                                NativeInterop.ExporterResponse.Free(responsePtr);
                            }
```

tracer/src/Datadog.Trace/LibDatadog/TraceExporter.cs
```diff
@@ -59,6 +64,21 @@ public Task<bool> SendTracesAsync(ArraySegment<byte> traces, int numberOfTraces,
 #pragma warning restore DDLOG004
                         throw ex;
                     }
+                    else
+                    {
+                        if (responsePtr != IntPtr.Zero)
+                        {
+                            string? response = Marshal.PtrToStringAnsi(NativeInterop.ExporterResponse.GetBody(responsePtr));
```
andrewlock: A couple of thoughts
- Is the response _actually_ definitely ANSI? Isn't it more likely to be UTF-8? If so then I think we need to change the FFI to also return the length. That way we can use UTF8 APIs (because `Marshal.PtrToStringUTF8` is not available in all TFMs 🙁)
- This allocates a string, and _then_ we deserialize that into our JSON response. If we returned the length of the response too then we could skip the string allocation entirely, and go straight to the deserialization. Seeing as you are already checking for duplicate values in libdatadog, I don't think there's any value to use materializing the string directly.

hoolioh: You're right, I wanted to use `Marshal.PtrToStringUTF8` but, as you said, I noticed that it was only available from 6 onwards. When you say UFT8 APIs you mean using something like [Utf8JsonReader](https://learn.microsoft.com/es-es/dotnet/api/system.text.json.utf8jsonreader.-ctor?view=net-9.0#system-text-json-utf8jsonreader-ctor(system-readonlyspan((system-byte))-system-boolean-system-text-json-jsonreaderstate))?

andrewlock: So in .NET Core, there are a bunch of APIs for adding a managed layer directly over native code, i.e. `Span<T>` and `ReadOnlySpan<T>`. These are designed for high perf, and they just require a pointer to memory, and a length. 

_Right now_, we can't deserialize directly from JSON, because we use the old Json.NET serialiser, but there's been a bunch of discussion about using the System.Text.Json serializer because 1. it's much more performant, 2. It's built in. This _would_ allow us to deserialize directly from the bytes without an inermediate allocation, something like  this:

```csharp
var bodyPointer = NativeInterop.ExporterResponse.GetBody(responsePtr, out var length);

var readOnlySpan = new ReadOnlySpan<byte>(bodyPointer, length)
var json = Json.Deserialize<ApiResponse(readonlyspan)
```

However, in terms of solving the Ansi issue, the best thing would be if you returned a `vec`, as we can map that to our existing `VecU8`: https://github.com/DataDog/dd-trace-dotnet/blob/b430c4c70b458d790cbd37d83c65eb3de0f1d4f9/tracer/src/Datadog.Trace/LibDatadog/VecU8.cs#L18

and then we can get the UTF8 string from that.

Does that make sense? I went on a few tangents 😅

andrewlock: Related, I assume `GetBody` _could_ fail, in which case, should it actually return an `Error`, and return the `VecU8` as an output parameter? 🤔

hoolioh: I've refactored it to avoid the extra allocation. I made a workaround to get the length, I will add a new function to the native side to get the body length but I think we could go with this to test that we got rid of the mem leak and then once the new API is in libdd just remove the managed function.
Regarding `GetBody`, I does not return an error if the response is null or there is no body present it will return a Null pointer so as long as that situation is handled in the try-catch clause it

tracer/src/Datadog.Trace/LibDatadog/TraceExporter.cs
```diff
@@ -59,6 +65,25 @@ public Task<bool> SendTracesAsync(ArraySegment<byte> traces, int numberOfTraces,
...
                     }
+
+                    try
+                    {
+                        // TODO: replace GetBodyLen with a native function in order to avoid iterating over the response to get its length.
+                        int len = GetBodyLen(responsePtr);
+                        byte* body = (byte*)NativeInterop.ExporterResponse.GetBody(responsePtr);
+                        ReadOnlySpan<byte> span = new ReadOnlySpan<byte>(body, len);
+
+                        var apiResponse = System.Text.Json.JsonSerializer.Deserialize<ApiResponse>(span);
```
andrewlock: We can't use System.Text.Json yet, so this is about the best we can do I think:
```suggestion
                        var response = StringEncoding.UTF8.GetString((byte*)Ptr, (int)Length);
                        var apiResponse = JsonConvert.DeserializeObject<ApiResponse>(response);
```

tracer/src/Datadog.Trace/LibDatadog/TraceExporter.cs
```diff
@@ -92,4 +116,21 @@ protected override bool ReleaseHandle()
 
         return true;
     }
+
+    private unsafe int GetBodyLen(IntPtr body)
+    {
+        if (body == IntPtr.Zero)
+        {
+            throw new ArgumentNullException(nameof(body));
+        }
```
andrewlock: In hindsight, throwing is expensive, we should maybe just do a fast return here (and maybe add a debug log given that this is unexpected), and then in the calling method check that len > 0, and if not, exit gracefully. My concern is that this is called a lot, so if there _is_ any issue here, we're likely going to start writing a _lot_ of logs...

tonyredondo: this will also affect inlining. Let's avoid throwing in an hot path, if this is absolutely necessary then let's create a different NonInlined function and throw from there.

andrewlock:
It looks like we _don't_ have integration tests for the sampling rate update which would cover these changes (which is going to be a repeated problem I think 🙁). [A bunch of tests are failing in the data pipeline enabled](https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=181529&view=logs&j=cc578895-adfa-5051-1356-3b220e12921d&t=1c63e335-ed52-58c7-827e-80e5f4113871) branch, indicating the deserialization is failing... So this isn't behaving like the managed exporter one way or another...

```
 Datadog.Trace.Telemetry.DTOs.LogMessageData
13:24:36 [DBG]     {
13:24:36 [DBG]         Count = <null>, 
13:24:36 [DBG]         Level = TelemetryLogLevel.ERROR {value: 0}, 
13:24:36 [DBG]         Message = "An error ocurred deserializing the response.", 
13:24:36 [DBG]         StackTrace = "Datadog.Trace.Vendors.Newtonsoft.Json.JsonReaderException
13:24:36 [DBG]    at Datadog.Trace.Vendors.Newtonsoft.Json.JsonTextReader.Read()
13:24:36 [DBG]    at Datadog.Trace.Vendors.Newtonsoft.Json.Serialization.JsonSerializerInternalReader.Deserialize(JsonReader reader, Type objectType, Boolean checkAdditionalContent)
13:24:36 [DBG]    at Datadog.Trace.Vendors.Newtonsoft.Json.JsonSerializer.DeserializeInternal(JsonReader reader, Type objectType)
13:24:36 [DBG]    at Datadog.Trace.Vendors.Newtonsoft.Json.JsonConvert.DeserializeObject(String value, Type type, JsonSerializerSettings settings)
13:24:36 [DBG]    at Datadog.Trace.Vendors.Newtonsoft.Json.JsonConvert.DeserializeObject[T](String value, JsonSerializerSettings settings)
13:24:36 [DBG]    at Datadog.Trace.LibDatadog.TraceExporter.SendTracesAsync(ArraySegment`1 traces, Int32 numberOfTraces, Boolean statsComputationEnabled, Int64 numberOfDroppedP0Traces, Int64 numberOfDroppedP0Spans, Boolean appsecStandaloneEnabled)
13:24:36 [DBG] ", 
```

My bigger concern is that the fact this is only caught incidentally by one unrelated telemetry-related integration test. I think we need to port all the existing Api tests to run against the exporter, which is easier said than done unfortunately...

tracer/test/Datadog.Trace.IntegrationTests/LibDatadog/TraceExporterTests.cs
```diff
@@ -87,11 +107,20 @@ public async Task SendsTracesUsingDataPipeline(TestTransports transport)
...
+        sampleRateResponses.Should()
+                           .NotBeEmpty()
+                           .And.AllSatisfy(rates => rates.Should().BeEquivalentTo(expectedRates));
+
         Dictionary<string, object> GetSettings()
         {
             var settingsMap = new Dictionary<string, object>
             {
-                { ConfigurationKeys.StatsComputationEnabled, true },
+                { ConfigurationKeys.StatsComputationEnabled, false },
```
andrewlock, line 123: Data pipeline isn't enabled if `StatsComputationEnabled=true` 🤦

andrewlock (approved):
Thanks!

PR #7204, opened 2025-07-10
Use non-BOM version of `Encoding.UTF8` in Libdatadog interop code

andrewlock:
## Summary of changes

Use the non-BOM version of `Encoding.UTF8` in libdatadog interop-code

## Reason for change

libdatadog assumes no BOM, so we should too.

## Implementation details

Use the existing non-BOM instance we are using in various places:

https://github.com/DataDog/dd-trace-dotnet/blob/cdf2c3b9d2dfe72c885c1e58e7c99def0769699e/tracer/src/Datadog.Trace/Vendors/MessagePack/StringEncoding.cs#L12

## Test coverage

Meh, covered by existing (kinda 😅)

## Other details

### DIFF
--- a/tracer/src/Datadog.Trace/LibDatadog/CharSlice.cs
+++ b/tracer/src/Datadog.Trace/LibDatadog/CharSlice.cs
@@ -9,6 +9,7 @@ using System;
 using System.Runtime.InteropServices;
 using System.Text;
 using Datadog.Trace.Util;
+using Datadog.Trace.Vendors.MessagePack;
 
 namespace Datadog.Trace.LibDatadog;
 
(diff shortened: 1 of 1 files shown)


tracer/src/Datadog.Trace/LibDatadog/CharSlice.cs
```diff
@@ -44,7 +45,7 @@ internal CharSlice(string? str)
         }
         else
         {
-            var encoding = Encoding.UTF8;
+            var encoding = StringEncoding.UTF8;
```
bouwkast, line 48: A nitpick, but this is not terribly obvious to me that it is using this constructor -> `UTF8Encoding(false)` to disable the byte order mark.

I could see someone swapping back / changing this by accident

andrewlock: Yeah, good point. You know, that gives me an idea, and excuse for a project I've wanted for a while, watch this space 😉

PR #7205, opened 2025-07-10
fix: Handle both SQS StringValue and SNS BinaryValue in context extraction

robcarlan-datadog:
## Summary of changes
SNS -> SQS context propagation was broken, as we encode the value of the `_datadog` header into base64. On the consume side, we assume it is written as a string, which is only true if we send a message directly through SQS.

The fix is to decode the `_datadog` value based on the type, just as in [this](https://github.com/DataDog/datadog-lambda-js/pull/269) PR.

This soles issues where the DSM map doesn't show links between SNS and SQS, as the checkpoint is still created on the consumer, just without a link to the parent.

## Test coverage
I tested this via system-tests, with a description of my test methodology in [this PR](https://github.com/DataDog/system-tests/pull/4897).
The SNS system test above pass when I ran locally with this tracer build.

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Shared/AwsMessageAttributesHeadersAdapters.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Shared/AwsMessageAttributesHeadersAdapters.cs
@@ -11,6 +11,7 @@ using System.Collections.Generic;
 using System.Text;
 using Datadog.Trace.DuckTyping;
 using Datadog.Trace.Headers;
+using Datadog.Trace.Vendors.MessagePack;
 using Datadog.Trace.Vendors.Newtonsoft.Json;
 
 namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.AWS.Shared;
@@ -76,7 +77,20 @@ internal static class AwsMessageAttributesHeadersAdapters
                 else if (json.BinaryValue != null)
                 {
                     // SNS encodes the json string in base64
-                    jsonString = Encoding.UTF8.GetString(json.BinaryValue.ToArray());
+                    if (json.BinaryValue.TryGetBuffer(out var bytes))
+                    {
+                        jsonString = StringEncoding.UTF8.GetString(bytes.Array!, bytes.Offset, bytes.Count);
+                    }
+                    else
+                    {
+                        using var reader = new System.IO.StreamReader(
+                            json.BinaryValue,
+                            Encoding.UTF8,
+                            detectEncodingFromByteOrderMarks: false,
+                            bufferSize: 1024,
+                            leaveOpen: true);
+                        jsonString = reader.ReadToEnd();
+                    }
                 }
 
                 if (jsonString != null)


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Shared/AwsMessageAttributesHeadersAdapters.cs
```diff
@@ -66,9 +66,23 @@ public MessageAttributesAdapter(IDictionary? messageAttributes)
...
+                string? jsonString = null;
+                if (json.StringValue != null)
+                {
+                    jsonString = json.StringValue;
+                }
+                else if (json.BinaryValue != null)
+                {
+                    // SNS encodes the json string in base64
+                    jsonString = Encoding.UTF8.GetString(json.BinaryValue.ToArray());
+                }
```
andrewlock: The downside of calling `ToArray()` on the `MemoryStream` is that it allocates a whole new (potentially large) array before then converting it to a `string`. An alternative, (that uses a `StringBuilder` under the hood) would be:

```csharp
using var reader = new System.IO.StreamReader(json.BinaryValue, Encoding.UTF8, detectEncodingFromByteOrderMarks: false, bufferSize: 1024, leaveOpen: true);
jsonString = reader.ReadToEnd();
```

_Another_ alternative, which is likely even better, given that we've already got the bytes buffered in memory in the `MemoryStream`, is to try to get the underlying buffer, and use that directly.

```csharp
if (json.BinaryValue.TryGetBuffer(out var bytes))
{
    jsonString = Encoding.UTF8.GetString(bytes.Array!, bytes.Offset, bytes.Count);
}
```

And we could combine them to be safe like this:
```suggestion
                else if (json.BinaryValue != null)
                {
                    // SNS encodes the json string in base64
                    if (json.BinaryValue.TryGetBuffer(out var bytes))
                    {
                        jsonString = Encoding.UTF8.GetString(bytes.Array!, bytes.Offset, bytes.Count);
                    }
                    else
                    {
                        using var reader = new System.IO.StreamReader(json.BinaryValue, Encoding.UTF8, detectEncodingFromByteOrderMarks: false, bufferSize: 1024, leaveOpen: true);
                        jsonString = reader.ReadToEnd();
                    }
                }
```

(note that the reader assumes that the `json.BinaryValue.Position == 0`, if not, you would need to reset it in the `else` block before doing the read)

andrewlock (approved):
LGTM, just one perf-related suggestion, thaanks!

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Shared/AwsMessageAttributesHeadersAdapters.cs
```diff
@@ -66,9 +66,23 @@ public MessageAttributesAdapter(IDictionary? messageAttributes)
...
-                _ddAttributes = JsonConvert.DeserializeObject<Dictionary<string, string>>(json.StringValue);
+                string? jsonString = null;
+                if (json.StringValue != null)
+                {
+                    jsonString = json.StringValue;
+                }
+                else if (json.BinaryValue != null)
+                {
+                    // SNS encodes the json string in base64
+                    jsonString = Encoding.UTF8.GetString(json.BinaryValue.ToArray());
```
andrewlock: minor point: are you encoding this value with or without the BOM? Are you _expecting_ to include a BOM? If the answer is "without" and "no, we shouldn't include one", you might want to use the vendored `StringEncoding.UTF8` instead of `Encoding.UTF8`, as that does not include a BOM. This may well apply to _all_ your DSM integrations, so might be worth verifying if they're all doing the right thing 😄

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Shared/AwsMessageAttributesHeadersAdapters.cs
robcarlan-datadog: Done! Makes sense, thanks for explaining 😄

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Shared/AwsMessageAttributesHeadersAdapters.cs
robcarlan-datadog: I confirmed we don't use a BOM so moved to the vendored library. Good to know going forwards!

robcarlan-datadog:
Confirmed system-tests passed locally when using the agent after my most recent commit

PR #7222, opened 2025-07-14
remove redundant dependency check for dynamic assemblies

lucaspimentel:
## Summary of changes

Reverts one of the changes from #7212 which added a redundant check for dynamic assemblies when reporting dependencies to telemetry.

## Reason for change

We already ignore dynamic assemblies here:
https://github.com/DataDog/dd-trace-dotnet/blob/3487421367be7188312151a33a5028bfb29f69ab/tracer/src/Datadog.Trace/Telemetry/Collectors/DependencyTelemetryCollector.cs#L27

## Implementation details

Remove the code and its test case.

## Test coverage

_Removed_ a test.

## Other details

n/a


bouwkast (approved):
Thanks!

PR #7224, opened 2025-07-15
Add snapshot tests to document the symbols we depend on in alpine and native loader

andrewlock:
## Summary of changes

Adds more "native symbol" snapshot tests for the symbols we depend on in our native libraries

## Background

As [part of the initial work on "universal binaries"](https://github.com/DataDog/dd-trace-dotnet/pull/6048) we had to [fix an issue](https://github.com/DataDog/dd-trace-dotnet/issues/6045) where we accidentally started depending on a new symbol, which was not always guaranteed to be present and could break customers. As a consequence, we subsequently set-up snapshot tests that [use `nm`](https://en.wikipedia.org/wiki/Nm_(Unix)) to dump the symbols required by our "native wrapper library" AKA `Datadog.Linux.ApiWrapper.x64.so`. We also added verification checks that we were building against the correct glibc version so we didn't accidentally change our supported distros.

## Reason for change

.NET 10 [has bumped](https://github.com/dotnet/runtime/issues/109939) the minimum supported version of Alpine from 3.13 ([.NET 9<](https://github.com/dotnet/core/blob/main/release-notes/9.0/supported-os.md#linux-compatibility)) to 3.17 ([.NET 10+](https://github.com/dotnet/core/blob/main/release-notes/10.0/supported-os.md#linux-compatibility)). My _previous_ understanding was that musl is essentially unversioned, so they're essentially also forward compatible.

Unfortunately, that is apparently not true. Attempting to run .NET 10 on Alpine 3.13 does not work, and it specifically fails with a `symbol not found` error:

```
Failed to load /usr/share/dotnet/host/fxr/10.0.0-preview.5.25277.114/libhostfxr.so, error: 
Error relocating /usr/share/dotnet/host/fxr/10.0.0-preview.5.25277.114/libhostfxr.so: 
_ZSt28__throw_bad_array_new_lengthv: symbol not found
```

So this leaves us with a dilemma; we need to update to alpine 3.17 so we can build and test with .NET 10, but doing so _might_ end up introducing breaking dependencies on new symbols.

This PR is about _documenting_ more fully what we actually depend on. Currently we have 

- Symbol tests for `Datadog.Linux.ApiWrapper.x64.so` on `linux-x64` and `linux-arm64`
- Maximum GLibc usages verification for `linux-x64` and `linux-arm64` for
  - `Datadog.Tracer.Native.so`
  - `Datadog.Profiler.Native.so`
  - `libdatadog_profiling.so`

This PR adds additional tests for the missing libraries and targets

- Symbol tests for `Datadog.Trace.ClrProfiler.Native.so` on `linux-x64` and `linux-arm64`
- Symbol tests  for `linux-musl-x64` and `linux-musl-arm64` for
  - `Datadog.Tracer.Native.so`
  - `Datadog.Profiler.Native.so`
  - `libdatadog_profiling.so`

> Note that `Datadog.Linux.ApiWrapper.x64.so` and `Datadog.Trace.ClrProfiler.Native.so` are "universal" binaries, so the same binary is used for both Glibc and musl.

## Implementation details

- Extracted the existing symbol test to a reusable function (1st commit, feel free to skip).
- Call the reusable function in the missing scenarios.
- Run the tests and save the snapshots.

## Test coverage

More now! 

Note that I'm not really sure how much value there is in documenting the `libdatadog_profiling.so` symbols. It's not something we control in this repository and it means needing to update the snapshot when something changes. I've included it now to be safe, but we can always address later.

## Other details

Investigating the alpine:3.17 changes is out of scope for this PR, I'll investigate that separately.

The one notably not-documented native library here is ddwaf. Should we document the symbols in that one too?


tracer/build/_build/Build.Profiler.Steps.cs
```diff
@@ -158,73 +153,8 @@ partial class Build
...
-
-            PrintDiff(diff);
-
-            throw new Exception($"Found differences in undefined symbols ({string.Join(",", changedSymbols)}) in the Native Wrapper library. " +
-                                "Verify that these changes are expected, and will not cause problems. " +
-                                "Removing symbols is generally a safe operation, but adding them could cause crashes. " +
-                                $"If the new symbols are safe to add, update the snapshot file at {verifiedPath} with the " +
-                                "new values");
+            var snapshotName = $"native-wrapper-symbols-{UnixArchitectureIdentifier}";
+            CompareNativeSymbolsSnapshot(libraryPath, snapshotName);
```
andrewlock, line 157: This change is purely a copy-paste move to a helper function in `Build.Shared.Steps`

tracer/build/_build/Build.Profiler.Steps.cs
```diff
@@ -904,16 +834,21 @@ void RunCppCheck(string projectName, MSBuildTargetPlatform platform)
             // Since the CLR requires them too, it seems safe to accept them.
             // For information, those symbols comes from libgcc and are exposed for compatibility
             var libdatadogAllowedSymbols = IsArm64 && IsAlpine ? new[] { "__register_frame_info@GLIBC_2.0", "__deregister_frame_info@GLIBC_2.0" } : null;
-            var filesAndVersion = new List<Tuple<string, Version, IEnumerable<string>>>
+            var filesAndVersion = new []
```
andrewlock, line 837: tiny refactor to use value tuples and inference for brevity

tracer/build/_build/Build.Shared.Steps.cs
```diff
@@ -337,4 +362,80 @@ Version FindMaxGlibcVersion(AbsolutePath libraryPath, IEnumerable<string> allowe
               .Select(x => System.Version.Parse(x.Substring(x.IndexOf("@GLIBC_") + 7)))
               .Max();
     }
+
+    void CompareNativeSymbolsSnapshot(AbsolutePath libraryPath, string snapshotNamePrefix)
```
andrewlock, line 361: This is where the code was copy-pasted to, from `Build.Profiler.Steps.cs` so doesn't really need reviewing (unless you really want to!)

tracer/build/_build/Build.Shared.Steps.cs
```diff
@@ -337,4 +362,80 @@ Version FindMaxGlibcVersion(AbsolutePath libraryPath, IEnumerable<string> allowe
...
+                            .Select(x => x.text.Trim())
+                            .ToList();
+
+        if (changedSymbols.Count == 0)
+        {
+            Logger.Information("No changes found in Undefined symbols in {LibraryName}", libraryName);
+            return;
+        }
+
+        // Print the expected values, so it's easier to copy-paste them into the snapshot file as required
```
andrewlock, line 423: This was the only change, I tweaked the logging

tracer/test/snapshots/native-libdatadog-symbols-alpine-arm64.verified.txt
andrewlock, line 1: These snapshots are what we depend on today. We don't necessarily care too much what they contain at this point (unless we are investigating escalations later). For our purposes, we're mostly interested in _changes_, particularly additions of new symbols.

tracer/test/snapshots/native-profiler-symbols-alpine-arm64.verified.txt
```diff
@@ -0,0 +1,249 @@
...
+bind
+blaze_read_elf_build_id
+btowc
+calloc
+ceil
+ceill
+clock_gettime
+close
+closedir
+ddog_crasht_CrashInfo_upload_to_endpoint
```
andrewlock, line 20: I considered removing all these `ddog_` symbols, as they're all provided by libdatadog, but that would have been a bit of a pain so, meh?

tonyredondo (approved):
LGTM

PR #7228, opened 2025-07-16
Update to alpine 3.17

andrewlock:
## Summary of changes

Updates the `alpine:3.14` base image we use for `linux-musl-x64` to `alpine:3.17`

## Reason for change

.NET 10 [has bumped](https://github.com/dotnet/runtime/issues/109939) the minimum supported version of Alpine from 3.13 ([.NET 9<](https://github.com/dotnet/core/blob/main/release-notes/9.0/supported-os.md#linux-compatibility)) to 3.17 ([.NET 10+](https://github.com/dotnet/core/blob/main/release-notes/10.0/supported-os.md#linux-compatibility)). My _previous_ understanding was that musl is essentially unversioned, so they're essentially also forward compatible. Unfortunately, that is apparently not true. Attempting to run .NET 10 on Alpine 3.13 does not work, and it specifically fails with a `symbol not found` error.

This PR bumps the base image to 3.17 and updates our symbol snapshots

## Implementation details

We added snapshots for which symbols we depend on here:
- https://github.com/DataDog/dd-trace-dotnet/pull/7224

That means we can see exactly what the impact of this change should be. 

For the native tracer, the symbol changes moving from `3.14` -> `3.17`
```diff
-floorl
+getentropy
+strncmp
```

For the continuous profiler, the symbol changes moving from `3.14` -> `3.17`
```diff
-ceill
-floorl
+getentropy
```

We're only really worried about the extra symbols, i.e. `strncmp` and `getentropy`.

- `strncmp` [we already depended on this in the native loader](https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/test/snapshots/native-loader-symbols-x64.verified.txt#L108), so overall this symbol is not _new_, so we should be ok. It's also part of the base ISO C standard, and is there from the first release, so we're fine.
- `getentropy` was [added in 2018](https://www.openwall.com/lists/musl/2018/01/06/2?utm_source=chatgpt.com), and was introduced [in musl 1.1.20](https://www.openwall.com/lists/musl/2018/09/04/17). Our previous `3.14` alpine image was built with musl `1.2.2`, so again we're fine 

## Test coverage

For the most part, this is the test. The smoke tests should catch if I've got something wrong here, but 🤞 

## Other details

This is a prerequisite for the .NET 10 work:
- https://github.com/DataDog/dd-trace-dotnet/pull/7170

However, merging it will require that we update the VMs, so I suggest we don't merge it _until_ we're ready to merge the .NET 10 PR shortly afterwards (as that will require another update)

https://datadoghq.atlassian.net/browse/LANGPLAT-632

### DIFF
--- a/tracer/build/_build/docker/alpine.build.arm64.dockerfile
+++ b/tracer/build/_build/docker/alpine.build.arm64.dockerfile
@@ -1,5 +1,6 @@
 # syntax=docker/dockerfile:1.6
 
+# See alpine.build.dockerfile for more information on how to build this image
 FROM alpine:3.18 as base
 
 RUN apk update \
--- a/tracer/build/_build/docker/alpine.build.dockerfile
+++ b/tracer/build/_build/docker/alpine.build.dockerfile
@@ -1,11 +1,23 @@
 # syntax=docker/dockerfile:1.6
 
-FROM alpine:3.14 as base
+# Build this image as described in https://andrewlock.net/combining-multiple-docker-images-into-a-multi-arch-image
+# Ultimately the process will look something like this, assuming that you need to build for both amd64 and arm64
+# (if you're only changing one of the dockerfiles, you can use the existing manifest already on docker hub)
+#
+# cd tracer/build/_build/docker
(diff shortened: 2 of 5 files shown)


andrewlock:
Subsumed as part of the .NET 10 update https://github.com/DataDog/dd-trace-dotnet/pull/7170

PR #7234, opened 2025-07-17
Update smoke tests for `debian:trixie`

andrewlock:
## Summary of changes

- Adds (and removes) some more smoke tests
- Fixes an issue with recent glibc version (in `debian:trixie`) that causes native tracer to fail to load

## Reason for change

Initial .NET 10 tests show that `debian:trixie` causes the .NET tracer to fail to load with:

```
/opt/datadog/./linux-x64/Datadog.Tracer.Native.so: cannot enable executable stack as shared object requires: Invalid argument
```

This is due to the fact that debian 13 uses glibc 2.41 which stopped automatically enabling executable stacks when loading shared libraries. This PR addresses the root cause, and adds smoke tests to demonstrate that.

## Implementation details

- Added smoke tests for `debian:trixie`
- Removed some smoke tests for older runtime/distro coverage - ensures we still cover all the TFMs and distros, but removes some redundancy
- Fixes the glibc issue (TBC)

## Test coverage

Adds some more smoke tests

## Other details

We discovered this in the .NET 10 upgrade PR, but Microsoft have [since _stopped_ publishing](https://github.com/dotnet/dotnet-docker/pull/6541) `debian` images 😅 So I recreated the images and .NET 8/ .NET 9 versions so we can test it right now.

### DIFF
--- a/profiler/src/ProfilerEngine/Datadog.Linux.ApiWrapper/CMakeLists.txt
+++ b/profiler/src/ProfilerEngine/Datadog.Linux.ApiWrapper/CMakeLists.txt
@@ -16,6 +16,22 @@ if (DEFINED ENV{IsAlpine} AND "$ENV{IsAlpine}" MATCHES "true")
     add_compile_options(-DDD_ALPINE)
 endif()
 
+if (ISLINUX)
+    # ------------------------------------------------------
+    # Hardening: make sure no target in this project ever
+    # requests an executable stack.  Without this, glibc ≥2.41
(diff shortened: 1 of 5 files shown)


andrewlock (commented):
Awesome, thanks for the fixes @tonyredondo! 🎉

tonyredondo (approved):
LGTM 😄

zacharycmontoya (approved):
Nice work!

PR #7237, opened 2025-07-17
feat: add injection metadata fields to telemetry forwarder

sydney-tung:
## Summary of changes
This change adds additional fields to the telemetry forwarder used during injection. These fields will be used to add additional context that will be surfaced to the end user in the following Figma:

Screenshot 2025-07-03 at 10 39 35 AM

See [Injection Metadata](https://docs.google.com/document/d/1wydXj2fW0V0jaWU6W9pIbipT_AXD15tmbTtOPXFBEy4/edit?tab=t.f3vpy615c2dp) for more details.

## Reason for change
Adding addition information for telemetry forwarder during injection.

## Implementation details
Added result, result class, and result reason fields. 

## Test coverage
Used unit tests to check for metadata

## Other details

Jira ticket: [INPLAT-614](https://datadoghq.atlassian.net/browse/INPLAT-614)

[INPLAT-614]: https://datadoghq.atlassian.net/browse/INPLAT-614?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shared/src/Datadog.Trace.ClrProfiler.Native/single_step_guard_rails.cpp
```diff
@@ -171,17 +181,19 @@ bool SingleStepGuardRails::ShouldForceInstrumentationOverride(const std::string&
             "SingleStepGuardRails::ShouldForceInstrumentationOverride: ",
             EnvironmentVariables::ForceEolInstrumentation,
             " enabled, allowing unsupported runtimes and continuing");
+        SetInjectResult("success", "Force instrumentation enabled", "success_forced");
```
andrewlock: Should we mention in the message that you're forcing instrumentation with an incompatible runtime? Or is it enough that it's implied?

shared/src/Datadog.Trace.ClrProfiler.Native/single_step_guard_rails.cpp
```diff
@@ -280,7 +300,10 @@ void SingleStepGuardRails::SendTelemetry(const std::string& runtimeName, const s
         + "\",\"language_name\": \"dotnet\",\"language_version\": \"" + runtimeVersion
         + "\",\"tracer_version\": \"" + PROFILER_VERSION
         + "\",\"pid\":" + std::to_string(GetPID())
-        + "},\"points\": " + points + "}";
+        + ",\"inject_result\": \"" + m_injectResult
+        + "\",\"inject_result_reason\": \"" + m_injectResultReason
+        + "\",\"inject_result_class\": \"" + m_injectResultClass
```
andrewlock: Given that there's no guarantee `SetInjectResult` has been called, should we check if the values have been set, and if not give them a default value? 🤔

shared/src/Datadog.Trace.ClrProfiler.Native/single_step_guard_rails.cpp
```diff
@@ -84,6 +88,7 @@ HRESULT SingleStepGuardRails::CheckRuntime(const RuntimeInformation& runtimeInfo
             {
                 // supported
                 Log::Debug("SingleStepGuardRails::CheckRuntime: Supported .NET runtime version detected, continuing with single step instrumentation");
+                SetInjectResult("success", "Successfully configured ddtrace package", "success");
```
andrewlock: nit: `ddtrace` isn't a .NET thing 😄 

```suggestion
                SetInjectResult("success", "Successfully configured automatic instrumentation", "success");
```

shared/src/Datadog.Trace.ClrProfiler.Native/single_step_guard_rails.h
```diff
@@ -25,8 +28,9 @@ class SingleStepGuardRails
     ~SingleStepGuardRails();
     HRESULT CheckRuntime(const RuntimeInformation& runtimeInformation, IUnknown* pICorProfilerInfoUnk);
     void RecordBootstrapError(const std::string& runtimeName, const std::string& runtimeVersion, const std::string& errorType) const;
-    void RecordBootstrapError(const RuntimeInformation& runtimeInformation, const std::string& errorType) const;
+    void RecordBootstrapError(const RuntimeInformation& runtimeInformation, const std::string& errorType);
     void RecordBootstrapSuccess(const RuntimeInformation& runtimeInformation) const;
```
andrewlock, line 32: I don't think all of these methods are calling `SetInjectResult`? 🤔

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/InstrumentationTests.cs
```diff
@@ -492,7 +492,7 @@ public async Task OnSupportedFrameworkInSsi_CallsForwarderWithExpectedTelemetry(
                                "tags": ["injection_forced:false"]
                              }]
                              """;
-            AssertHasExpectedTelemetry(logFileName, processResult, pointsJson);
+            AssertHasExpectedTelemetry(logFileName, processResult, pointsJson, "success", "Successfully configured ddtrace package", "success");
```
andrewlock: If we update the description above we need to remember to update these!

```suggestion
            AssertHasExpectedTelemetry(logFileName, processResult, pointsJson, "success", "Successfully configured automatic instrumentation", "success");
```

andrewlock (commented):
LGTM in general, thanks! Just a few small questions/suggestions 🙂

shared/src/Datadog.Trace.ClrProfiler.Native/single_step_guard_rails.cpp
andrewlock: I think we should still explicitly check if these values are null before we use them, as there's no _guarantee_ in the code that we must always call `SetInjectResult()` before this method is called (even if it _is_ always called currently - code changes in the future could change that, and then we would crash)

sydney-tung: I think that the values are being set to unknown in the constructor so they will never be null? WDYT

andrewlock: Oh yes, you're right, that's fine, thanks!

andrewlock (approved):
LGTM, thanks!

PR #7239, opened 2025-07-18
Integrations: Disable AWS Kinesis injection when the integration is disabled

zacharycmontoya:
## Summary of changes
Updates the AWS Kinesis injection logic to leave the outgoing message unmodified when the integration is disabled, aka `scope == null`

## Reason for change
We've observed that even when the AWS Kinesis integration is disabled, attaching the .NET tracer modifies the message by adding `_datadog: {}` (when enabled the dictionary contains key-value pairs for trace context information and other properties). Instead, I think we should omit injecting this entirely when the integration is disabled.

## Implementation details
Adds a branch to `Datadog.Trace.ClrProfiler.AutoInstrumentation.AWS.Kinesis.ContextPropagation.InjectTraceIntoData` to return immediately if the scope is `null`

## Test coverage
Adds two unit tests that assert that the Kinesis message is not modified when the AWS Kinesis integration is disabled via `DD_TRACE_AwsSdk_ENABLED=false` and `DD_TRACE_AwsKinesis_ENABLED=false`


andrewlock (approved):
LGTM, thanks! I wonder if there are other AWS integrations that have similar issues, e.g. SNS/SQS - should we check those too?

bouwkast (approved):
Thanks!

kr-igor (approved):

PR #7244, opened 2025-07-20
DSM default state reversal

kr-igor:
## Summary of changes
This PR enables DSM by default in non-serverless environments. Additionally:
* DSM background writer will not start if the application doesn't produce DSM checkpoints or backlogs
* Added multiple safeguards

## Reason for change
https://docs.google.com/document/d/111QQyLxz5q5G3wccj5Q2Zu7e3NemlXDSt6HcchA2LKU/edit?tab=t.0

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Kinesis/ContextPropagation.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Kinesis/ContextPropagation.cs
@@ -21,6 +21,7 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.AWS.Kinesis
     {
         private const string KinesisKey = "_datadog";
         private const int MaxKinesisDataSize = 1024 * 1024; // 1MB
+        private const int MaxDSMHeaderSize = 34;
         private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(ContextPropagation));
 
         public static void InjectTraceIntoRecords<TRecordsRequest>(TRecordsRequest request, Scope? scope, string? streamName)
@@ -46,21 +47,7 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.AWS.Kinesis
                 return;
             }
 
-            Dictionary<string, object> propagatedContext = new Dictionary<string, object>();
-            if (scope.Span.Context != null && !string.IsNullOrEmpty(streamName))
-            {
-                var dataStreamsManager = Tracer.Instance.TracerManager.DataStreamsManager;
-                if (dataStreamsManager != null && dataStreamsManager.IsEnabled)
-                {
-                    var edgeTags = new[] { "direction:out", $"topic:{streamName}", "type:kinesis" };
-                    scope.Span.SetDataStreamsCheckpoint(dataStreamsManager, CheckpointKind.Produce, edgeTags, payloadSizeBytes: 0, timeInQueueMs: 0);
-                    var adapter = new KinesisContextAdapter();
-                    dataStreamsManager.InjectPathwayContext(scope.Span.Context.PathwayContext, adapter);
-                    propagatedContext = adapter.GetDictionary();
-                }
-            }
-
-            var context = new PropagationContext(scope?.Span.Context, Baggage.Current);
+            var context = new PropagationContext(scope.Span.Context, Baggage.Current);
             if (record.Data is null)
             {
                 return;
@@ -70,6 +57,32 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.AWS.Kinesis
             if (jsonData is null || jsonData.Count == 0)
             {
                 return;
+            }
+
+            var propagatedContext = new Dictionary<string, object>();
+            if (scope.Span.Context != null && !string.IsNullOrEmpty(streamName))
+            {
+                var dataStreamsManager = Tracer.Instance.TracerManager.DataStreamsManager;
+                if (dataStreamsManager != null && dataStreamsManager.IsEnabled)
+                {
+                    var payloadSize = record.Data?.Length ?? 0;
+                    var edgeTags = new[] { "direction:out", $"topic:{streamName}", "type:kinesis" };
+                    scope.Span.SetDataStreamsCheckpoint(
+                        dataStreamsManager,
+                        CheckpointKind.Produce,
+                        edgeTags,
+                        payloadSizeBytes: payloadSize,
+                        timeInQueueMs: 0);
+
+                    var adapter = new KinesisContextAdapter();
+                    // we should not inject context if its size is comparable to the message size itself
+                    if (payloadSize > MaxDSMHeaderSize)
+                    {
+                        dataStreamsManager.InjectPathwayContext(scope.Span.Context.PathwayContext, adapter);
+                    }
+
+                    propagatedContext = adapter.GetDictionary();
+                }
             }
 
             try
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
@@ -23,8 +23,8 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Kafka
         internal const string EnableDeliveryReportsField = "dotnet.producer.enable.delivery.reports";
         private const string MessagingType = "kafka";
         private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(KafkaHelper));
+        private static readonly string[] DefaultProduceEdgeTags = ["direction:out", "type:kafka"];
         private static bool _headersInjectionEnabled = true;
-        private static string[] defaultProduceEdgeTags = new[] { "direction:out", "type:kafka" };
 
         internal static Scope? CreateProducerScope(
             Tracer tracer,
@@ -280,7 +280,7 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Kafka
             try
             {
                 if (!tracer.Settings.IsIntegrationEnabled(KafkaConstants.IntegrationId)
-                    || !tracer.Settings.KafkaCreateConsumerScopeEnabled)
+                 || !tracer.Settings.KafkaCreateConsumerScopeEnabled)
                 {
                     // integration disabled, skip this trace
                     return;
@@ -339,11 +339,18 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Kafka
                 if (dataStreamsManager.IsEnabled)
                 {
                     var edgeTags = string.IsNullOrEmpty(topic)
-                        ? defaultProduceEdgeTags
-                        : new[] { "direction:out", $"topic:{topic}", "type:kafka" };
+                                       ? DefaultProduceEdgeTags
+                                       : ["direction:out", $"topic:{topic}", "type:kafka"];
                     var msgSize = dataStreamsManager.IsInDefaultState ? 0 : GetMessageSize(message);
                     // produce is always the start of the edge, so defaultEdgeStartMs is always 0
                     span.SetDataStreamsCheckpoint(dataStreamsManager, CheckpointKind.Produce, edgeTags, msgSize, 0);
+                    // DSM context should NOT be injected state if the message value size is <= DSM header size (~34 bytes).
+                    // This is needed to avoid situations when DSM context injection may cause issues on the Kafka side.
+                    if (dataStreamsManager.IsInDefaultState && MessageSizeHelper.TryGetSize(message.Value) <= 34)
+                    {
+                        return;
+                    }
+
                     dataStreamsManager.InjectPathwayContext(span.Context.PathwayContext, adapter);
                 }
             }
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/RabbitMQ/RabbitMQIntegration.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/RabbitMQ/RabbitMQIntegration.cs
@@ -26,6 +26,7 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.RabbitMQ
     internal static class RabbitMQIntegration
     {
         internal const string IntegrationName = nameof(Configuration.IntegrationId.RabbitMQ);
+        internal const int DefaultMaxMessageSize = 128 * 1024;
 
         private const string MessagingType = "rabbitmq";
         private const string MessagingSystem = "amqp";
@@ -138,11 +139,19 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.RabbitMQ
             try
             {
                 var headersAdapter = new RabbitMQHeadersCollectionAdapter(headers);
-                var edgeTags = string.IsNullOrEmpty(tags.Exchange) ?
+                var edgeTags = string.IsNullOrEmpty(tags.Exchange)
+                                   ?
                                    // exchange can be empty for "direct"
-                                   new[] { "direction:out", $"topic:{tags.Queue ?? tags.RoutingKey}", "type:rabbitmq" } :
-                                   new[] { "direction:out", $"exchange:{tags.Exchange}", string.IsNullOrEmpty(tags.RoutingKey) ? "has_routing_key:false" : "has_routing_key:true", "type:rabbitmq" };
-                span.SetDataStreamsCheckpoint(dataStreamsManager, CheckpointKind.Produce, edgeTags, GetHeadersSize(headers) + messageSize, 0);
+                                   new[] { "direction:out", $"topic:{tags.Queue ?? tags.RoutingKey}", "type:rabbitmq" }
+                                   : new[] { "direction:out", $"exchange:{tags.Exchange}", string.IsNullOrEmpty(tags.RoutingKey) ? "has_routing_key:false" : "has_routing_key:true", "type:rabbitmq" };
+                var size = dataStreamsManager.IsInDefaultState ? messageSize : GetHeadersSize(headers) + messageSize;
+                span.SetDataStreamsCheckpoint(dataStreamsManager, CheckpointKind.Produce, edgeTags, size, 0);
+                // DSM context will not be injected in default state if its size exceeds 128kb
+                if (dataStreamsManager.IsInDefaultState && size > DefaultMaxMessageSize)
+                {
+                    return;
+                }
+
                 dataStreamsManager.InjectPathwayContext(span.Context.PathwayContext, headersAdapter);
             }
             catch (Exception ex)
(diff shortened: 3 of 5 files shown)


kr-igor:
Fix for system tests https://github.com/DataDog/system-tests/pull/4981/files

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Kinesis/ContextPropagation.cs
```diff
@@ -21,6 +21,7 @@ internal static class ContextPropagation
     {
         private const string KinesisKey = "_datadog";
         private const int MaxKinesisDataSize = 1024 * 1024; // 1MB
+        private const int MaxDSMHeaderSize = 34;
```
lucaspimentel: nit-picking variable names: C# convention is `MaxDsmHeaderSize` (capitalize only the `D` in `Dsm`).

Level of care: low

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -339,11 +339,18 @@ internal static void TryInjectHeaders<TTopicPartitionMarker, TMessage>(
                 if (dataStreamsManager.IsEnabled)
                 {
                     var edgeTags = string.IsNullOrEmpty(topic)
-                        ? defaultProduceEdgeTags
-                        : new[] { "direction:out", $"topic:{topic}", "type:kafka" };
+                                       ? DefaultProduceEdgeTags
+                                       : ["direction:out", $"topic:{topic}", "type:kafka"];
```
lucaspimentel, line 343: Not a new issue in this PR, but this line is allocating a new `string[]` and a new string for `$"topic:{topic}"` on every call. Just something to maybe look into later.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/RabbitMQ/RabbitMQIntegration.cs
```diff
@@ -138,11 +139,19 @@ internal static void SetDataStreamsCheckpointOnProduce(Tracer tracer, Span span,
...
                 var headersAdapter = new RabbitMQHeadersCollectionAdapter(headers);
-                var edgeTags = string.IsNullOrEmpty(tags.Exchange) ?
+                var edgeTags = string.IsNullOrEmpty(tags.Exchange)
+                                   ?
                                    // exchange can be empty for "direct"
-                                   new[] { "direction:out", $"topic:{tags.Queue ?? tags.RoutingKey}", "type:rabbitmq" } :
-                                   new[] { "direction:out", $"exchange:{tags.Exchange}", string.IsNullOrEmpty(tags.RoutingKey) ? "has_routing_key:false" : "has_routing_key:true", "type:rabbitmq" };
-                span.SetDataStreamsCheckpoint(dataStreamsManager, CheckpointKind.Produce, edgeTags, GetHeadersSize(headers) + messageSize, 0);
+                                   new[] { "direction:out", $"topic:{tags.Queue ?? tags.RoutingKey}", "type:rabbitmq" }
+                                   : new[] { "direction:out", $"exchange:{tags.Exchange}", string.IsNullOrEmpty(tags.RoutingKey) ? "has_routing_key:false" : "has_routing_key:true", "type:rabbitmq" };
```
lucaspimentel, line 146: Similar to my other comment. Not a new issue in this PR, but these line are allocating a new `string[]` and a new string on every call. Maybe something to look into later.

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
```diff
@@ -252,8 +289,9 @@ private void HandleConfigUpdate(AgentConfiguration config)
             }
             else
             {
-                Log.Warning("Data streams monitoring was enabled but is not supported by the Agent. Disabling Data streams. " +
-                            "Consider upgrading your Datadog Agent to at least version 7.34.0+");
+                Log.Warning(
+                    "Data streams monitoring was enabled but is not supported by the Agent. Disabling Data streams. " +
+                    "Consider upgrading your Datadog Agent to at least version 7.34.0+");
```
lucaspimentel: > Consider upgrading your Datadog Agent to at least version 7.34.0+

Mentioning the agent version doesn't work in serverless where we use different agents, each one with its own versioning scheme.

tracer/test/Datadog.Trace.Tests/SpanContextInjectorExtractorTests.cs
```diff
@@ -102,7 +102,8 @@ public void InjectionWithDsm()
     [Fact]
     public void NoDsmInjectionIsDisabled()
     {
-        // DSM is disabled by default
+        // DSM is enabled by default
+        Tracer.Configure(TracerSettings.Create(new Dictionary<string, object> { { ConfigurationKeys.DataStreamsMonitoring.Enabled, false } }));
```
lucaspimentel: I think this uses the static `Tracer` which we shouldn't do in tests (static state is shared across tests and causes flake). cc @andrewlock for a suggested approach.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -339,11 +339,18 @@ internal static void TryInjectHeaders<TTopicPartitionMarker, TMessage>(
...
                     var msgSize = dataStreamsManager.IsInDefaultState ? 0 : GetMessageSize(message);
                     // produce is always the start of the edge, so defaultEdgeStartMs is always 0
                     span.SetDataStreamsCheckpoint(dataStreamsManager, CheckpointKind.Produce, edgeTags, msgSize, 0);
+                    // DSM context should NOT be injected state if the message value size is <= DSM header size (~34 bytes).
+                    // This is needed to avoid situations when DSM context injection may cause issues on the Kafka side.
```
andrewlock:
```suggestion
                    // This is needed to avoid situations when DSM context injection causes a significant 
                    // percentage increase in overall message size, leading to capacity issues on the kafka server.
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
andrewlock: Yeah, given that there's likely to be a limited number of `topic` in a given, there's potentially a _bunch_ of optimisation we can do 🤔 First of all to avoid allocating the string repeatedly, but also because this array is primarily used to calculate a `FnvHash64`... so that's something we can also potentially cache per topic? 🤔

tracer/test/Datadog.Trace.Tests/SpanContextInjectorExtractorTests.cs
andrewlock: Yup, good catch. In fact, the fundamental usage here (and `SpanContextInjector` etc) are problematic for use of static globals. I'll refactor separately and we can fix this after

tracer/build/_build/Build.Steps.cs
```diff
@@ -2498,6 +2498,8 @@ string NormalizedPath(AbsolutePath ap)
            // glibc TLS-reuse bug warnings
            knownPatterns.Add(new(@".*GLIBC version 2.34-2.36 has a TLS-reuse bug.*", RegexOptions.Compiled));
 
+           // not all agent versions support DSM
+           knownPatterns.Add(new(@".*Data streams monitoring was enabled but is not supported by the Agent.*", RegexOptions.Compiled));
```
andrewlock: All the versions _we_ use do and should be valid 🤔 Where did you see this?

tracer/test/Datadog.Trace.Tests/SpanContextInjectorExtractorTests.cs
andrewlock: I did the cleanup for all this in
- https://github.com/DataDog/dd-trace-dotnet/pull/7258

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/RabbitMQ/RabbitMQIntegration.cs
kr-igor: I will prepare a separate PR with DSM tags refactoring.

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
lucaspimentel: I think it's ok to just remove that last part, "Consider upgrading your Datadog Agent. ~to at least version 7.34.0+~".

tracer/build/_build/Build.Steps.cs
lucaspimentel: See also https://github.com/DataDog/dd-trace-dotnet/pull/7244#discussion_r2220536790

kr-igor: [Here's an example ](https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=184035&view=logs&j=eac47dc8-0dc1-56e2-29db-3882ca6a86a0&t=119e5eff-474b-5f74-b2b4-8efebec796ea&l=62)

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Kinesis/ContextPropagation.cs
```diff
@@ -46,21 +47,33 @@ public static void InjectTraceIntoData<TRecordRequest>(TRecordRequest record, Sc
...
+                        dataStreamsManager,
+                        CheckpointKind.Produce,
+                        edgeTags,
+                        payloadSizeBytes: payloadSize,
+                        timeInQueueMs: 0);
+
                     var adapter = new KinesisContextAdapter();
-                    dataStreamsManager.InjectPathwayContext(scope.Span.Context.PathwayContext, adapter);
+                    // we should not inject context if its size is comparable to the message size itself
+                    if (payloadSize > MaxDsmHeaderSize)
```
robcarlan-datadog: Is this meant to be regardless of whether DSM is in the default state?

andrewlock:
> Fix for system tests https://github.com/DataDog/system-tests/pull/4981

FYI I don't think that fix is sufficient, because it won't pass in system tests. Those system tests run against both library `main` branches _and_ releases, and they need to be written in such a way that the tests pass. I think you potentially need to create an entirely new test, so you can gate it based on the library version. You can then test that branch manually against this PR, but have the test disabled in system-tests (so that you can merge the change in system tests without breaking our main branch PR). Then, once this is merged, you can do a PR to system-tests to un-disable it. Yes, it's horrible 😅

tracer/build/_build/Build.Steps.cs
andrewlock: I can't see that example as the build's gone🙁 anyway, if it _is_ the case that the test agent doesn't support it, I think we probably want to update the test agent 🤔 Was it a specific platform (e.g. Windows) or anything that doesn't support it?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Kinesis/ContextPropagation.cs
```diff
@@ -46,21 +47,33 @@ public static void InjectTraceIntoData<TRecordRequest>(TRecordRequest record, Sc
...
             }
 
-            Dictionary<string, object> propagatedContext = new Dictionary<string, object>();
+            var propagatedContext = new Dictionary<string, object>();
             if (scope.Span.Context != null && !string.IsNullOrEmpty(streamName))
             {
                 var dataStreamsManager = Tracer.Instance.TracerManager.DataStreamsManager;
                 if (dataStreamsManager != null && dataStreamsManager.IsEnabled)
                 {
+                    var payloadSize = record.Data?.Length ?? 0;
```
andrewlock: What was the reason we weren't recording this before? 🤔 Is this `MemoryStream` guaranteed to be fully buffered in memory? i.e. is the network stream fully flushed to it? If not this could give a smaller than accurate size?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Kinesis/ContextPropagation.cs
andrewlock: Should we also be not injecting if the payload size is too small, similar to kafak?

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -665,9 +665,10 @@ _ when x.ToBoolean() is { } boolean => boolean,
                              .WithKeys(ConfigurationKeys.IpHeaderEnabled)
                              .AsBool(false);
 
+            // DSM is now enabled by default in non-serverless environments
             _isDataStreamsMonitoringEnabled = config
                                             .WithKeys(ConfigurationKeys.DataStreamsMonitoring.Enabled)
-                                            .AsBool(false);
+                                            .AsBool(!EnvironmentHelpers.IsServerlessEnvironment());
```
andrewlock: Note that this means it _will_ be enabled when using [the AAS extension](https://github.com/DataDog/datadog-aas-extension). This is pinned to using `trace-agent` `v7.60.1` and does _not_ include the full agent - does that include the required endpoints for DSM?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Kinesis/ContextPropagation.cs
kr-igor: You are right, this stream is fully read only when deserializing to JSON. I will restructure the code to capture this.

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
kr-igor: I didn't know about AAS, thanks for catching this! I'll what can be done

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Kinesis/ContextPropagation.cs
kr-igor: The decision to inject / not inject happens later, after the payload is deserialized from json (if possible).

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
lucaspimentel: To know if the app is running in AAS, you can check the value of `IsRunningInAzureAppService`, but only after it's set around line 137 above.

Something like this?
```csharp
.AsBool(!IsRunningInAzureAppService && !EnvironmentHelpers.IsServerlessEnvironment())
```

lucaspimentel: Related: I'm currently doing some refactoring around how we detect Azure App Service and Azure Functions in #7256 because we will soon support instrumenting without the AAS Site Extension.

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -668,7 +668,9 @@ _ when x.ToBoolean() is { } boolean => boolean,
             // DSM is now enabled by default in non-serverless environments
             _isDataStreamsMonitoringEnabled = config
                                             .WithKeys(ConfigurationKeys.DataStreamsMonitoring.Enabled)
-                                            .AsBool(!EnvironmentHelpers.IsServerlessEnvironment());
+                                            .AsBool(
+                                                  !EnvironmentHelpers.IsServerlessEnvironment() &&
```
johannbotha, line 672: Why do we filter out Azure envs? 

@robcarlan-datadog if we do this maybe we want to make it clear why in our user facing docs?

kr-igor: The [agent](https://github.com/DataDog/datadog-aas-extension) running in that env will not have the DSM checkpoint.

kr-igor:
> * unit_tests_linux

@andrewlock I've disabled that failing test for .NET tracer >=v3.22.0. Will enable it back right after merging this PR.

tracer/build/_build/Build.Steps.cs
```diff
@@ -2498,6 +2498,8 @@ string NormalizedPath(AbsolutePath ap)
            // glibc TLS-reuse bug warnings
            knownPatterns.Add(new(@".*GLIBC version 2.34-2.36 has a TLS-reuse bug.*", RegexOptions.Compiled));
 
+           // not all agent versions support DSM
+           // knownPatterns.Add(new(@".*Data streams monitoring was enabled but is not supported by the Agent.*", RegexOptions.Compiled));
```
andrewlock: OK, so this fails in _all_ the integration tests, because the test agent doesn't support that endpoint it seems (we should probably update it so that it _does_ so that we can actually smoke test this properly 🫤)

But for now, rather than ignore this warning, I think it would be better if we _only_ logged this warning if data streams is actually trying to be used i.e. it's either explicitly enabled, or it's initialized. That way we avoid confusing/worrying customers when it doesn't even apply to them. If we want to make sure we still log this, we could make it an info if it's only enabled by defatul, and a warning if they've explicitly enabled it?

andrewlock (commented):
I think we're looking good in general, just two things to do before merging:

- Disable the `tests/parametric/test_telemetry.py::Test_Defaults::test_library_settings` system test (in the system test repo), and merge it, so that the system tests pass.
- Address the failing smoke tests - I think we should downgrade the log level to log when DSM isn't explicitly enabled (which should also fix the tests)

tracer/build/_build/Build.Steps.cs
kr-igor: Good idea, I'll update the logging logic.

tracer/test/Datadog.Trace.Tests/SpanContextInjectorExtractorTests.cs
lucaspimentel: @andrewlock now that your PR is merged, what needs to be done here to fix this test? It's still using the static tracer instance.

andrewlock: Just need to merge/rebase master and remove this line 

```suggestion
```

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
```diff
@@ -252,8 +293,16 @@ private void HandleConfigUpdate(AgentConfiguration config)
             }
             else
             {
-                Log.Warning("Data streams monitoring was enabled but is not supported by the Agent. Disabling Data streams. " +
-                            "Consider upgrading your Datadog Agent to at least version 7.34.0+");
+                const string msg = "Data streams monitoring was enabled but is not supported by the Agent. Disabling Data streams. " +
```
andrewlock, line 296: nit: just noticed that I think the logging above is potentially incorrect too. e.g. DSM might be _supported_, but we won't actually enable _flushing_ at this point, so the above (existing) branch is kinda incorrect. I think we can probably just tidy up the logic in here in general, but definitely not a blocker

PR #7251, opened 2025-07-21
Add  rates payload version support.

hoolioh:

### DIFF
--- a/tracer/src/Datadog.Trace/LibDatadog/TraceExporter.cs
+++ b/tracer/src/Datadog.Trace/LibDatadog/TraceExporter.cs
@@ -22,6 +22,8 @@ internal class TraceExporter : SafeHandle, IApi
 {
     private readonly IDatadogLogger _log;
     private readonly Action<Dictionary<string, float>> _updateSampleRates;
+    private readonly bool _ratesVersioning;
+    private string _cachedResponse;
 
     public TraceExporter(
         TraceExporterConfiguration configuration,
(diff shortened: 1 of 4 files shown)


tracer/src/Datadog.Trace/TracerManagerFactory.cs
```diff
@@ -415,7 +415,8 @@ internal static IApi GetApi(TracerSettings settings, IDogStatsd statsd, Action<D
                         LanguageInterpreter = frameworkDescription.Name,
                         ComputeStats = settings.StatsComputationEnabled,
                         TelemetryClientConfiguration = telemetryClientConfiguration,
-                        ClientComputedStats = clientComputedStats
+                        ClientComputedStats = clientComputedStats,
+                        RatesPayloadVersion = false
```
hoolioh, line 419: I've checked that this was included in the agent version 7.27.0 which was released 4 years ago. I don't know if it could be possible that any customer could be using such version, that's why I decided to disable it and cache the response as a fallback to maintain parity with the managed trace manager. In case we would need to support that scenario I think that we would need to handle it internally in the trace exporter since at this point we don't have the agent version. @andrewlock WDYT?

andrewlock: > I don't know if it could be possible that any customer could be using such version,

Yeah, it definitely could be possible. Some customers still use the 6.x agent for example. 

Yeah, IMO this caching etc should be implemented in libdatadog otherwise we're paying the cost for the string creation unnecessarily 🤔 Given libdatadog is looking for that header, it knows if it's supported, so it seems like it should be relatively simple to implement?

tracer/src/Datadog.Trace/LibDatadog/TraceExporter.cs
```diff
@@ -149,8 +153,17 @@ private unsafe void ProcessResponse(IntPtr response)
...
+        {
+            var apiResponse = JsonConvert.DeserializeObject<ApiResponse>(json);
+            _updateSampleRates(apiResponse.RateByService);
+        }
+        else if (_cachedResponse != json)
+        {
+            _cachedResponse = json;
+            var apiResponse = JsonConvert.DeserializeObject<ApiResponse>(json);
+            _updateSampleRates(apiResponse.RateByService);
+        }
```
andrewlock, line 167: I understand the reasoning here, but I really think we should push this into libdatadog if possible

andrewlock (commented):
I think I agree with you that we should try to push this caching into libdatadog, otherwise essentially all libraries will have to implement this on their side

PR #7252, opened 2025-07-21
Set default timeout for Trace Exporter to 15s.

hoolioh:
## Summary of changes
Add new binding to set the trace exporter connection timeout.
## Reason for change
Trace exporter was setting a 3s connection timeout which probably was insufficient for some test cases. This version use the new bindings to increase the time up to 15 seconds.

### DIFF
--- a/tracer/src/Datadog.Trace/TracerManagerFactory.cs
+++ b/tracer/src/Datadog.Trace/TracerManagerFactory.cs
@@ -40,6 +40,7 @@ namespace Datadog.Trace
     internal class TracerManagerFactory
     {
         private const string UnknownServiceName = "UnknownService";
+        private const ulong DefaultTimeoutMs = 100_000;
         private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor<TracerManagerFactory>();
 
         public static readonly TracerManagerFactory Instance = new();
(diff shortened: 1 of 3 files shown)


tracer/src/Datadog.Trace/TracerManagerFactory.cs
```diff
@@ -40,6 +40,7 @@ namespace Datadog.Trace
     internal class TracerManagerFactory
     {
         private const string UnknownServiceName = "UnknownService";
+        private const ulong DefaultTimeoutMs = 100_000;
```
andrewlock: nit: I think we should inline this, as it's too general a name, and only used in one place. If/when we unify this across the managed code too, we can extract it out where it _should_ be (which probably isn't here anyway 😄)

```suggestion
```

tracer/src/Datadog.Trace/TracerManagerFactory.cs
```diff
@@ -415,7 +416,8 @@ internal static IApi GetApi(TracerSettings settings, IDogStatsd statsd, Action<D
                         LanguageInterpreter = frameworkDescription.Name,
                         ComputeStats = settings.StatsComputationEnabled,
                         TelemetryClientConfiguration = telemetryClientConfiguration,
-                        ClientComputedStats = clientComputedStats
+                        ClientComputedStats = clientComputedStats,
+                        ConnectionTimeoutMs = DefaultTimeoutMs
```
andrewlock: The 100s timeout is crazy long 😅 I think we should use a more sensible value here, this should still be more than enough for CI, and if it's not we prob need to dig in further

```suggestion
                        ConnectionTimeoutMs = 15_000,
```

andrewlock (commented):
Thanks, LGTM in general, I think we should lower the timeout from the crazy 100s that's the current default. I'll rebase the "enable by default" PR and see what happens 🙂

PR #7255, opened 2025-07-21
Refactor log directory resolution into smaller methods, add tests

lucaspimentel:
## Summary of changes

Refactor `DatadogLoggingFactory` log directory resolution logic into smaller, testable methods and add unit tests.

## Reason for change

The log directory resolution logic was a single large method mixing concerns:
- config reading
- default directory lookup
- directory creation 😦
- fallback logic

Breaking it apart makes each piece independently testable and easier to understand.

## Implementation details

- Extract `GetDefaultLogDirectory()` — determines the _default_ log directory based on platform and Azure environment
- Extract `GetProgramDataDirectory()` — resolves the `ProgramData` path on Windows (with Nano Server fallbacks)
- Extract `TryCreateLogDirectory()` — attempts to _create_ a log directory, returning success/failure
- Make extracted methods `internal` with `[TestingAndPrivateOnly]` for testability

Simplify `GetLogDirectory()` to orchestrate the above methods with clear fallback chain:
  1. `DD_TRACE_LOG_DIRECTORY` env var
  2. `DD_TRACE_LOG_PATH` (deprecated, directory extracted from file path)
  3. `GetDefaultLogDirectory()` (platform/Azure-aware default)
  4. `Path.GetTempPath()` (last resort)

## Test coverage

Flattened nested test classes (`FileLoggingConfiguration`, `RedactedLogConfiguration`, `SinkConfiguration`) into top-level methods with consistent `MethodUnderTest_Scenario_ExpectedBehavior` naming:

| Old (nested class.method) | New (flat method) |
|---|---|
| `FileLoggingConfiguration.UsesLogDirectoryWhenItExists` | `GetConfiguration_WithLogDirectory_UsesLogDirectory` |
| `FileLoggingConfiguration.UsesObsoleteLogDirectoryWhenAvailable` | `GetConfiguration_WithObsoleteLogPath_UsesObsoleteLogDirectory` |
| `FileLoggingConfiguration.UsesEnvironmentFallBackWhenBothNull` | `GetConfiguration_WithNoLogDirectorySettings_UsesEnvironmentFallback` |
| `FileLoggingConfiguration.CreatesLogDirectoryWhenItDoesntExist` | `GetConfiguration_WithNonExistentLogDirectory_CreatesDirectory` |
| `RedactedLogConfiguration.WhenNoOrInvalidConfiguration_TelemetryLogsEnabled` | `GetConfiguration_WithNoOrInvalidTelemetryLogsSetting_EnablesErrorLogging` |
| `RedactedLogConfiguration.WhenDisabled_TelemetryLogsDisabled` | `GetConfiguration_WithTelemetryLogsDisabled_DisablesErrorLogging` |
| `SinkConfiguration.WhenNoSinksProvided_UsesFileSink` | `GetConfiguration_WithNoSinksSetting_UsesFileSink` |
| `SinkConfiguration.WhenFileSinkIsIncluded_UsesFileSink` | `GetConfiguration_WithFileSinkIncluded_UsesFileSink` |
| `SinkConfiguration.WhenFileSinkIsNotIncluded_DoesNotUseFileSink` | `GetConfiguration_WithFileSinkNotIncluded_DoesNotUseFileSink` |
| `SinkConfiguration.WhenConsoleSinkIsIncluded_UsesConsoleSink` | `GetConfiguration_WithConsoleSinkIncluded_UsesConsoleSink` |
| `SinkConfiguration.WhenConsoleSinkIsNotIncluded_DoesNotUseConsoleSink` | `GetConfiguration_WithConsoleSinkNotIncluded_DoesNotUseConsoleSink` |

Additional tests for the newly extracted methods:
- `GetDefaultLogDirectory_*` — verifies default paths on Windows/Linux, and Azure App Services detection
- `GetProgramDataDirectory_*` — verifies `ProgramData` resolution on Windows
- `TryCreateLogDirectory_*` — verifies directory creation success, existing directory, invalid path, and nested path scenarios

## Other details

> *"I refactored a method into smaller methods. Now I have more methods to refactor into even smaller methods."* — Claude 🤖

### DIFF
--- a/tracer/test/Datadog.Trace.Tests/Logging/DatadogLoggingFactoryTests.cs
+++ b/tracer/test/Datadog.Trace.Tests/Logging/DatadogLoggingFactoryTests.cs
@@ -1,4 +1,4 @@
-﻿// <copyright file="DatadogLoggingFactoryTests.cs" company="Datadog">
+// <copyright file="DatadogLoggingFactoryTests.cs" company="Datadog">
 // Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2 License.
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
 // </copyright>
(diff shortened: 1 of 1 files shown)


tracer/src/Datadog.Trace/Logging/Internal/DatadogLoggingFactory.cs
```diff
@@ -165,94 +165,126 @@ private static ConsoleLoggingConfiguration GetConsoleLoggingConfiguration(IConfi
...
     {
-        // This entire block may throw a SecurityException if not granted the System.Security.Permissions.FileIOPermission
-        // because of the following API calls
-        //   - Directory.Exists
-        //   - Environment.GetFolderPath
-        //   - Path.GetTempPath
-        if (string.IsNullOrEmpty(logDirectory))
+        var isWindows = FrameworkDescription.Instance.IsWindows();
+
+        if (ImmutableAzureAppServiceSettings.IsRunningInAzureAppServices(source, telemetry))
```
andrewlock: Looks like you've removed the `|| ImmutableAzureAppServiceSettings.IsRunningInAzureFunctions()` - is that correct?

tracer/src/Datadog.Trace/Logging/Internal/DatadogLoggingFactory.cs
```diff
@@ -165,94 +165,126 @@ private static ConsoleLoggingConfiguration GetConsoleLoggingConfiguration(IConfi
...
-                // will have to create it on their own.
-                // Last effort at writing logs
-                logDirectory = Path.GetTempPath();
+                // fallback #2: hard-coded
+                programData = @"C:\ProgramData";
             }
         }
 
-        return logDirectory!;
+        return programData;
```
andrewlock: nit: I think early returns are neater in this case and reduces nesting
```suggestion
        if (!StringUtil.IsNullOrEmpty(programData))
        {
            return programData;
        }

        // fallback #1: try reading from the env var        
        // fallback #2: hard-coded
        var envProgramData = EnvironmentHelpersNoLogging.ProgramData();
        return StringUtil.IsNullOrEmpty(envProgramData)
            ? @"C:\ProgramData"
            : envProgramData;
```

tracer/test/Datadog.Trace.Tests/Logging/DatadogLoggingFactoryTests.cs
```diff
@@ -197,4 +198,160 @@ public void WhenConsoleSinkIsNotIncluded_DoesNotUseConsoleSink(string sinks)
             config.Console.Should().BeNull();
         }
     }
+
+    public class GetDefaultLogDirectoryTests
```
andrewlock: nit: I don't think these should be separate classes based on our typical code style

tracer/test/Datadog.Trace.Tests/Logging/DatadogLoggingFactoryTests.cs
```diff
@@ -197,4 +198,160 @@ public void WhenConsoleSinkIsNotIncluded_DoesNotUseConsoleSink(string sinks)
...
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                return;
+            }
+
+            var result = DatadogLoggingFactory.GetProgramDataDirectory();
+
+            result.Should().NotBeNullOrEmpty();
+        }
+
```
andrewlock: meh, covered by the next test
```suggestion
```

tracer/test/Datadog.Trace.Tests/Logging/DatadogLoggingFactoryTests.cs
```diff
@@ -197,4 +198,161 @@ public void WhenConsoleSinkIsNotIncluded_DoesNotUseConsoleSink(string sinks)
...
+            result.Should().BeTrue();
+        }
+
+        [Fact]
+        public void WithInvalidPath_ReturnsFalse()
+        {
+            // Use an invalid path that cannot be created
+            var logDirectory = RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
+                ? @"Z:\nonexistent\invalid\path\that\cannot\be\created"
+                : "/root/nonexistent/invalid/path/that/cannot/be/created";
```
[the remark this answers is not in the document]
andrewlock: AI is right I'm afraid, as shown by the failing tests 😅 Try this:
```suggestion
            var logDirectory = RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
                ? @"Z:\nonexistent\invalid\path\that\cannot\be\created"
                : "/dev/null/nonexistent/invalid/path/that/cannot/be/created";
```

tracer/test/Datadog.Trace.Tests/Logging/DatadogLoggingFactoryTests.cs
```diff
@@ -197,4 +198,160 @@ public void WhenConsoleSinkIsNotIncluded_DoesNotUseConsoleSink(string sinks)
...
+            // Cleanup
+            try
+            {
+                Directory.Delete(logDirectory);
+            }
+            catch
+            {
+                // Ignore cleanup errors
+            }
+        }
```
andrewlock: nit: I'd just combine this and the final tests. Plus the cleanup theoretically needs "fixing" to cleanup in case of failure:

```suggestion
        public void WithValidPath_CreatesDirectoryAndReturnsTrue()
        {
            var parentDir = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString());
            var logDirectory = Path.Combine(parentDir, "nested", "log", "directory");
            Directory.Exists(logDirectory).Should().BeFalse();
            Directory.Exists(parentDir).Should().BeFalse();

            try
            {
                var result = DatadogLoggingFactory.TryCreateLogDirectory(logDirectory);
    
                result.Should().BeTrue();
                Directory.Exists(logDirectory).Should().BeTrue();               
            }
            finally
            {
                // Cleanup
                try
                {
                    Directory.Delete(logDirectory, recursive: true);
                }
                catch
                {
                    // Ignore cleanup errors
                }
            }
        }
```

tracer/test/Datadog.Trace.Tests/Logging/DatadogLoggingFactoryTests.cs
```diff
@@ -197,4 +198,160 @@ public void WhenConsoleSinkIsNotIncluded_DoesNotUseConsoleSink(string sinks)
...
+            // Cleanup
+            try
+            {
+                Directory.Delete(parentDir, recursive: true);
+            }
+            catch
+            {
+                // Ignore cleanup errors
+            }
+        }
```
andrewlock: nit: covered by first test
```suggestion
```

tracer/src/Datadog.Trace/Logging/Internal/DatadogLoggingFactory.cs
lucaspimentel: "You're absolutely right!"

I was under the impression that _all_ Azure Functions run inside Azure App Services (and many do: some AAS apps are "web apps" and some are "function apps"), so I though one was always a superset/subset of the other.

BUT ... I did some quick research thanks to your comment and in some cases (e.g. "Azure Functions on Container Apps") Azure Functions do _not_ run inside AAS.

I'll restore this check, and now I have to go and check:
- for other similar checks we're probably getting wrong because of this wrong assumption 😓
- what's the right log path for container apps?

tracer/test/Datadog.Trace.Tests/Logging/DatadogLoggingFactoryTests.cs
lucaspimentel: 🤦🏽

tracer/test/Datadog.Trace.Tests/Logging/DatadogLoggingFactoryTests.cs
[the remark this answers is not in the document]
lucaspimentel: fixed

tracer/src/Datadog.Trace/Logging/Internal/DatadogLoggingFactory.cs
lucaspimentel: restore this check

tracer/src/Datadog.Trace/Logging/Internal/DatadogLoggingFactory.cs
lucaspimentel: fine 😝

tracer/test/Datadog.Trace.Tests/Logging/DatadogLoggingFactoryTests.cs
lucaspimentel: oops

tracer/test/Datadog.Trace.Tests/Logging/DatadogLoggingFactoryTests.cs
lucaspimentel: fixed

tracer/test/Datadog.Trace.Tests/Logging/DatadogLoggingFactoryTests.cs
lucaspimentel: thanks

tracer/test/Datadog.Trace.Tests/Logging/DatadogLoggingFactoryTests.cs
lucaspimentel: This file already had nested classes before this PR:
https://github.com/DataDog/dd-trace-dotnet/blob/5b423185857b3ec6d970e0294a608b62018cabb9/tracer/test/Datadog.Trace.Tests/Logging/DatadogLoggingFactoryTests.cs

I flattened the ones I added. Should I flatten the pre-existing ones, too?

```
DatadogLoggingFactoryTests
    FileLoggingConfiguration
    RedactedLogConfiguration
    SinkConfiguration
```

andrewlock: Hmmm, ok fine, whatever then 😅 

Those nested test classes are testing nested types, so are correct IMO, but creating a nested class for every method isn't a style we _tend_ to use. But it's tests, so meh, whatever 😉

lucaspimentel: > Those nested test classes are testing nested types, 

Ahh, I see. Thanks. Maybe I'll split the tests for nested and non-nested classes into separate files. Evidently I didn't look that deep into the existing structure 😅

lucaspimentel: I refactored this into a single flat class and renamed the methods to match. See PR description for mapping from old to new names.

PR #6093, opened 2024-09-27
[dynamic Instrumentation] DEBUG-2336 Cche redaction results

dudikeleti:
## Summary of changes
Redaction test could impact performance, this PR add the ability to cache results and return those results instead to re-test every time

PR #7256, opened 2025-07-21
[Azure Functions] Fixes, refactoring, and clean-up

lucaspimentel:
## Summary of changes

Fix issues that affect Azure Functions when not using the AAS Site extension. Refactor and clean up code used to detect serverless environments.

## Reason for change

Improve support for Azure Functions instrumented with the `Datadog.AzureFunctions` nuget package instead of the AAS Site Extension.

## Implementation details

- Distinguish between
  - running in Azure App Services
  - running in Azure Functions
  - using the AAS Site Extension
- Move Azure Functions configuration keys from `ConfigurationKeys.AzureAppService.cs` to new `ConfigurationKeys.AzureFunctions.cs` and add a few more
- In Azure Functions that _don't_ use the AAS Site extension:
  - Fix the default service name (use the Functions App name)
  - Fix the logging directory
- Fix missing `component` tag (old bug)
- More refactoring and clean-up and adding comments to make things easier for my future self 😅 

## Test coverage

- Fixed tests that relied on env vars added by the AAS Site Extension to determine if running in Azure App Services.
- Add new tests.

## Other details
Fixes APMSVLS-118

(AAS = Azure App Services)


tracer/src/Datadog.Trace/Tagging/AzureFunctionsTags.cs
```diff
@@ -45,7 +45,7 @@ internal static void SetRootSpanTags(
             {
                 aspNetTags.InstrumentationName = ComponentName;
             }
-            else if (tags.GetTag(Tags.InstrumentationName) is { })
```
lucaspimentel, line 48: Bug

tracer/src/Datadog.Trace/Logging/Internal/DatadogLoggingFactory.cs
```diff
@@ -195,25 +195,22 @@ private static string GetDefaultLogDirectory(IConfigurationSource source, IConfi
         //   - Path.GetTempPath
         if (string.IsNullOrEmpty(logDirectory))
         {
-#if NETFRAMEWORK
```
lucaspimentel, line 198: Bug

andrewlock: FWIW, these are all overwritten by the AAS extension anyway: https://github.com/DataDog/datadog-aas-extension/blob/master/dotnet/content/applicationHost.xdt#L65

Not saying this isn't a bug, just that it doesn't actually impact anyone using the AAS extension

lucaspimentel: Yeah, that's right. It's a bug that never manifested in apps using the AAS Site Extension because they always set `DD_TRACE_LOG_DIRECTORY`. Now that we will support Azure Functions _not_ using the extension, we need to handle this correctly.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -21,7 +21,7 @@ internal static class AzureFunctionsCommon
     {
         public const string IntegrationName = nameof(Configuration.IntegrationId.AzureFunctions);
 
-        public const string OperationName = "azure-functions.invoke";
+        public const string OperationName = "azure_functions.invoke";
```
bouwkast: I think this is a breaking change no?
Granted, I think `azure_functions` is more consistent with other names (and it appears that other traces opted for that too).

lucaspimentel: > I think this is a breaking change no?

The Agent is _already_ normalizing `azure-functions.invoke` to `azure_functions.invoke` before any trace stats are computed or the traces reach the backend because the hyphen is not allowed, so I don't think it's a breaking change.

This change fixes sampling rules that try to match by operation name, though, because the UI shows `azure_functions.invoke` to users, but if they try it, it won't match.

lucaspimentel: I reverted this change in this PR and moved it into it's own PR #7272 so we can discuss there.

tracer/src/Datadog.Trace/ClrProfiler/Helpers/HttpBypassHelper.cs
```diff
@@ -11,7 +13,7 @@ internal static class HttpBypassHelper
     {
         public static bool UriContainsAnyOf(Uri requestUri, string[] substrings)
         {
-            if (requestUri == null || substrings == null || substrings.Length == 0)
+            if (requestUri == null! || substrings == null! || substrings.Length == 0)
```
bouwkast: Are the `!` needed here?


```suggestion
            if (requestUri == null || substrings == null || substrings.Length == 0)
```

lucaspimentel: I added `#nullable enable` so now the compiler complains that we don't need to check for null because they are not nullable, but I'd thought we should keep checking for null just in case, so the `!` is to suppress the compiler warning.

But now that I think about it, I don't think either one of those can ever be null, so we can probably remove the null checks entirely.

lucaspimentel:
> there appear to be some unit test failures though for some of the settings

Yeah, I'm working my way through the unit tests, thanks 😅 

edit: unit tests pass, now on to `integration_tests_azure_functions`

tracer/test/snapshots/AzureFunctionsTests.InProcess.verified.txt
```diff
@@ -4,7 +4,7 @@
     SpanId: Id_2,
     Name: azure_functions.invoke,
     Resource: Timer TriggerAllTimer,
-    Service: AzureFunctionsAllTriggers,
+    Service: azurefunctionsalltriggers,
```
lucaspimentel: Not sure why these snapshots got updated, but the new values looks like what the agent would normalize the service name and operation name into (lowercase and replace spaces with underscores).

tracer/src/Datadog.Trace/ClrProfiler/Helpers/HttpBypassHelper.cs
zacharycmontoya: Hmm if you still think they may be null, then you should update the signature of `UriContainsAnyOf` to make each field a nullable reference type

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
```diff
@@ -439,11 +440,20 @@ private static void StartDiagnosticManager()
...
+            var functionsExtensionVersion = EnvironmentHelpers.GetEnvironmentVariable(ConfigurationKeys.AzureFunctions.FunctionsExtensionVersion);
+            var functionsWorkerRuntime = EnvironmentHelpers.GetEnvironmentVariable(ConfigurationKeys.AzureFunctions.FunctionsWorkerRuntime);
+
+            if (!string.IsNullOrEmpty(functionsExtensionVersion) && !string.IsNullOrEmpty(functionsWorkerRuntime))
+            {
+                // Not adding the `AspNetCoreDiagnosticObserver` is particularly important for in-process Azure Functions.
+                // The AspNetCoreDiagnosticObserver will be loaded in a separate Assembly Load Context, breaking the connection of AsyncLocal.
+                // This is because user code is loaded within the functions host in a separate context.
+                // Even in isolated functions, we don't want the AspNetCore spans to be created.
+                Log.Debug("Skipping AspNetCoreDiagnosticObserver in Azure Functions.");
```
zacharycmontoya, line 453: We plan to add at least one more Diagnostic Observer for instrumenting the Quartz library. Should that observer go inside this if/else block or outside?

tracer/src/Datadog.Trace/Util/EnvironmentHelpers.cs
```diff
@@ -98,8 +98,10 @@ public static IDictionary GetEnvironmentVariables()
 
         internal static bool IsServerlessEnvironment()
         {
+            // NOTE: this returns false in Azure App Services
```
zacharycmontoya: Should we rename this to IsFunctionAsAService or IsFaas? It's unintuitive to me that this returns false in `Azure App Services` so I think the better way to encode this is to change the method name rather than the code comment. If this is used all over the place, we can also do this refactor in another PR

tracer/src/Datadog.Trace/ClrProfiler/Helpers/HttpBypassHelper.cs
lucaspimentel: Good point 😅

lucaspimentel: I removed the null checks. Also went a step further and split this method so we can inline the fast path.

tracer/test/snapshots/AzureFunctionsTests.InProcess.verified.txt
lucaspimentel: Oh, I know. We are enabling `DD_TRACE_STATS_COMPUTATION_ENABLED` which does the same normalization as the agent before computing stats. I need to remove that since we won't use it anymore. It was only used during the preview with the "mini-agent."

lucaspimentel: I removed the code that enables `DD_TRACE_STATS_COMPUTATION_ENABLED` for Azure Functions and GCP and reverted the snapshot changes.

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
lucaspimentel: Honestly I don't know about this separate Assembly Load Context. The code and most of this comment was already there before. It sounds like it only applies to in-process functions.

However, even in isolated functions, skipping `AspNetCoreDiagnosticObserver` was convenient because we decided we don't want the `aspnetcore` spans, anyway.

My guess would be skip the Quartz diagnostic observer in in-process functions because of that Assembly Load Context, but feel free to include it in isolated functions.

tracer/src/Datadog.Trace/Util/EnvironmentHelpers.cs
lucaspimentel: Yeah, I had the same concerns when it was added in #6093, but it was decided at the time that it was a quick stop-gap solution because we knew I was going to come in and do some refactoring, anyway. I did a _little_ clean up in this PR, but I plan to do more refactoring, including this one method.

> _If this is used all over the place..._

It's only used in code introduced in that PR above, and more recently in `NativeInterop.LibDatadogAvailable`.

lucaspimentel: Ok, I broke it up into multiple methods and added comments to clarify that they check for env vars direct instead of using `TracerSettings`.

tracer/src/Datadog.Trace/ClrProfiler/Helpers/HttpBypassHelper.cs
```diff
@@ -3,20 +3,25 @@
...
+
 using System;
+using System.Runtime.CompilerServices;
 
 namespace Datadog.Trace.ClrProfiler.Helpers
 {
     internal static class HttpBypassHelper
     {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static bool UriContainsAnyOf(Uri requestUri, string[] substrings)
```
andrewlock: Are we sure that this Uri _isn't_ ever null? Just because we have the `#nullable enable` doesn't mean anything if it comes from a user or an integration. You _may_ want to play it safe regardless given the alternative is a potentially a crash, unless you're sure it's only called internally.

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
andrewlock: My guess is that Quartz isn't even applicable to Azure Functions, so we can probably luck out and exclude it 😅

tracer/src/Datadog.Trace/ClrProfiler/ScopeFactory.cs
```diff
@@ -76,11 +75,22 @@ internal static Scope CreateOutboundHttpScope(Tracer tracer, string httpMethod,
         /// <param name="startTime">The start time that should be applied to the span</param>
         /// <param name="addToTraceContext">Set to false if the span is meant to be discarded. In that case, the span won't be added to the TraceContext.</param>
         /// <returns>A new pre-populated scope.</returns>
-        internal static Span CreateInactiveOutboundHttpSpan(Tracer tracer, string httpMethod, Uri requestUri, IntegrationId integrationId, out HttpTags tags, TraceId traceId, ulong spanId, DateTimeOffset? startTime, bool addToTraceContext)
+        internal static Span CreateInactiveOutboundHttpSpan(
+            Tracer tracer,
+            string httpMethod,
+            Uri requestUri,
```
andrewlock, line 81: Here's a good example of my concern - `requestUri` _could_ be `null` in the caller of this method, because we don't have `null` ref types here. I would definitely be more cautious.

tracer/src/Datadog.Trace/Configuration/ImmutableAzureAppServiceSettings.cs
```diff
@@ -52,57 +52,43 @@ public ImmutableAzureAppServiceSettings(IConfigurationSource? source, IConfigura
...
-                    break;
+                IsFunctionsApp = false;
+                SiteKind = "app";
+                SiteType = "app";
             }
 
-            Runtime = FrameworkDescription.Instance.Name;
-
-            DebugModeEnabled = config.WithKeys(Configuration.ConfigurationKeys.DebugEnabled).AsBool(false);
+            DebugModeEnabled = config.WithKeys(ConfigurationKeys.DebugEnabled).AsBool(false);
```
andrewlock, line 98: Kinda wondering why this is here tbh 🤔 I know it was there before, but.... why?

tracer/src/Datadog.Trace/Configuration/ImmutableAzureAppServiceSettings.cs
```diff
@@ -206,24 +180,60 @@ private static bool ShouldSkipClientSpanWithinFunctions(Scope? scope)
             return null;
         }
 
-        public static bool GetIsFunctionsAppUsingMiniAgent(IConfigurationSource source, IConfigurationTelemetry telemetry)
+        public static bool GetIsFunctionsAppUsingMiniAgent(bool isFunctionsApp, string? websiteSku)
```
andrewlock: Seems a bit strange to have something else have to determine whether something's a functions app and pass it in here? 🤔 Should this just be an instance method on `ImmutableAzureAppServiceSettings`?

tracer/src/Datadog.Trace/Configuration/ImmutableAzureAppServiceSettings.cs
```diff
@@ -206,24 +180,60 @@ private static bool ShouldSkipClientSpanWithinFunctions(Scope? scope)
...
+        }
+
+        /// <summary>
+        /// Returns <c>true</c> if the app is running in Azure Functions.
+        /// Checks for the presence of "FUNCTIONS_WORKER_RUNTIME" and "FUNCTIONS_EXTENSION_VERSION" in the configuration.
+        /// </summary>
+        public static bool IsRunningInAzureFunctions(IConfigurationSource source, IConfigurationTelemetry telemetry)
+        {
+            var workerRuntime = new ConfigurationBuilder(source, telemetry)
+                           .WithKeys(ConfigurationKeys.AzureFunctions.FunctionsWorkerRuntime)
```
andrewlock, line 216: I wonder if we should actually just be checking this in the environment instead of all the config sources. Undecided as this is better for testing etc :D

tracer/src/Datadog.Trace/Configuration/ImmutableAzureAppServiceSettings.cs
```diff
@@ -206,24 +180,60 @@ private static bool ShouldSkipClientSpanWithinFunctions(Scope? scope)
...
+            var extensionVersion = new ConfigurationBuilder(source, telemetry)
+                           .WithKeys(ConfigurationKeys.AzureFunctions.FunctionsExtensionVersion)
+                           .AsString();
+
+            return !string.IsNullOrEmpty(workerRuntime) && !string.IsNullOrEmpty(extensionVersion);
+        }
+
+        /// <summary>
+        /// Returns <c>true</c> if the app is instrumented using the Azure App Services Site Extension.
+        /// Checks for the presence of "DD_AZURE_APP_SERVICES=1" and "DD_AAS_DOTNET_EXTENSION_VERSION" in the configuration.
```
andrewlock: Does it _have_ to be both? Wouldn't the first one alone be sufficient? 🤔

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -669,8 +666,11 @@ _ when x.ToBoolean() is { } boolean => boolean,
             _isDataStreamsMonitoringEnabled = config
                                             .WithKeys(ConfigurationKeys.DataStreamsMonitoring.Enabled)
                                             .AsBool(
-                                                  !EnvironmentHelpers.IsServerlessEnvironment() &&
-                                                  !IsRunningInAzureAppService);
+                                                  !EnvironmentHelpers.IsAwsLambda() &&
```
andrewlock, line 669: I feel like there's something more specific we're trying to capture here tbh. Like "Is running with full trace agent) for example 😅

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -685,10 +685,12 @@ _ when x.ToBoolean() is { } boolean => boolean,
 
             StatsComputationEnabled = config
                                      .WithKeys(ConfigurationKeys.StatsComputationEnabled)
-                                     .AsBool(defaultValue: (IsRunningInGCPFunctions || IsRunningMiniAgentInAzureFunctions));
-            if (!ApmTracingEnabled && StatsComputationEnabled)
+                                     .AsBool(false); // default is false, but user config can be overridden below
```
andrewlock, line 688: So this is a change in behaviour? i.e. it's enabled by default in GCP and with the mini agent, no? Is that change intentional?

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -1339,15 +1347,16 @@ public bool DiagnosticSourceEnabled
...
+        /// so is not available in some scenarios. It may also be explicitly disabled.
         /// </summary>
+        // NOTE: when we clean this up, see also EnvironmentHelpers.IsServerlessEnvironment()
         internal bool IsRemoteConfigurationAvailable =>
             RemoteConfigurationEnabled &&
-            !(IsRunningInAzureAppService
-           || IsRunningMiniAgentInAzureFunctions
-           || IsRunningInGCPFunctions
-           || LambdaMetadata.IsRunningInLambda);
+            !IsRunningInAzureAppService &&
```
andrewlock, line 1356: Shouldn't we be looking for the extension rather than AAS? Or do these also not have remote config?

tracer/src/Datadog.Trace/Debugger/Caching/DefaultEnvironmentChecker.cs
```diff
@@ -24,6 +24,8 @@ private DefaultEnvironmentChecker()
     private bool CheckServerlessEnvironment()
     {
         // Checking serverless environment based on environment variables
-        return EnvironmentHelpers.IsServerlessEnvironment();
+        return EnvironmentHelpers.IsAwsLambda() ||
+               EnvironmentHelpers.IsAzureFunctions() ||
+               EnvironmentHelpers.IsGoogleCloudFunctions();
```
andrewlock, line 29: This is a different set of "serverless environmment" checks, which seems wrong to me... it's not available with the AAS extension for example

tracer/src/Datadog.Trace/LibDatadog/NativeInterop.cs
```diff
@@ -17,10 +18,14 @@ internal class NativeInterop
...
-    // This confirms that we are in an automatic instrumentation environment (and so PInvokes have been re-written)
-    // and that the libdatadog library has been deployed (which is not the case in many serverless environments).
+    // This confirms that we are in an automatic instrumentation environment (and so P/Invokes have been re-written)
+    // and that the libdatadog library has been deployed (which is not the case in some serverless environments).
     // We should add or remove conditions from here as our deployment requirements change.
-    private static readonly Lazy<bool> LibDatadogAvailable = new(() => !Util.EnvironmentHelpers.IsServerlessEnvironment() && Instrumentation.ProfilerAttached);
+    private static readonly Lazy<bool> LibDatadogAvailable = new(() =>
+        Instrumentation.ProfilerAttached &&
+        !EnvironmentHelpers.IsAwsLambda() &&
+        (!EnvironmentHelpers.IsAzureAppServices() || EnvironmentHelpers.IsUsingAzureAppServicesSiteExtension()) &&
```
andrewlock, line 27: This isn't true is it? e.g. it _is_ available when you use the bundle...

tracer/src/Datadog.Trace/Telemetry/TelemetryController.cs
```diff
@@ -387,8 +387,12 @@ public void Update(TracerSettings settings)
...
                 { IsRunningInGCPFunctions: true } => ",gcp",
                 { LambdaMetadata.IsRunningInLambda: true } => ",aws",
-                { IsRunningMiniAgentInAzureFunctions: true } => ",azf",
-                { IsRunningInAzureAppService: true } => ",aas",
+
+                // check for Azure Functions first because it's a subset of Azure App Service
+                { IsRunningInAzureFunctions: true } => ",azf",
+                // IsRunningInAzureFunctions: true is handled above, but it doesn't hurt to be safe,
+                // so test for IsRunningInAzureFunctions: false
+                { IsRunningInAzureAppService: true, IsRunningInAzureFunctions: false } => ",aas",
```
andrewlock, line 395: FWIW, you don't need the explicit `IsRunningInAzureFunctions: false` check, and it will likely be prompted to be removed by the IDEs I think

tracer/src/Datadog.Trace/Util/EnvironmentHelpers.cs
```diff
@@ -96,15 +96,68 @@ public static IDictionary GetEnvironmentVariables()
...
+            return EnvironmentVariableExists(LambdaMetadata.FunctionNameEnvVar);
+        }
+
+        /// <summary>
+        /// Check if the current environment is Google Cloud Functions
+        /// by checking for the presence of either "K_SERVICE" and "FUNCTION_TARGET",
+        /// or "FUNCTION_NAME" and "GCP_PROJECT".
+        /// This method reads environment variables directly and bypasses the configuration system.
+        /// </summary>
+        public static bool IsGoogleCloudFunctions()
```
andrewlock, line 149: nit: should we be caching these checks? How often do we hit them all?

tracer/src/Datadog.Trace/Util/EnvironmentHelpers.cs
```diff
@@ -96,15 +96,68 @@ public static IDictionary GetEnvironmentVariables()
...
-        internal static bool IsServerlessEnvironment()
+        /// <summary>
+        /// Check if the current environment is Azure App Services
+        /// by checking for the presence of "WEBSITE_SITE_NAME".
+        /// Note that this is a superset of IsAzureFunctions().
+        /// This method reads environment variables directly and bypasses the configuration system.
+        /// </summary>
+        public static bool IsAzureAppServices()
+        {
+            return EnvironmentVariableExists(ConfigurationKeys.AzureAppService.SiteNameKey);
```
andrewlock, line 107: Isn't this a _slightly_ different check to the one we're _already_ doing in `ImmutableAzureAppServiceSettings`? Do we need/want both?

tracer/src/Datadog.Trace/Util/EnvironmentHelpers.cs
```diff
@@ -96,15 +96,68 @@ public static IDictionary GetEnvironmentVariables()
...
-            return !string.IsNullOrEmpty(GetEnvironmentVariable("AZURE_FUNCTIONS_ENVIRONMENT")) ||
-                   !string.IsNullOrEmpty(GetEnvironmentVariable(ConfigurationKeys.AzureAppService.FunctionsWorkerRuntimeKey)) ||
-                   !string.IsNullOrEmpty(GetEnvironmentVariable(LambdaMetadata.FunctionNameEnvVar)) ||
-                   !string.IsNullOrEmpty(GetEnvironmentVariable(ConfigurationKeys.GCPFunction.DeprecatedFunctionNameKey)) ||
-                   !string.IsNullOrEmpty(GetEnvironmentVariable(ConfigurationKeys.GCPFunction.DeprecatedProjectKey)) ||
-                   !string.IsNullOrEmpty(GetEnvironmentVariable(ConfigurationKeys.GCPFunction.FunctionNameKey)) ||
-                   !string.IsNullOrEmpty(GetEnvironmentVariable(ConfigurationKeys.GCPFunction.FunctionTargetKey));
+            return IsAzureAppServices() &&
+                   EnvironmentVariableExists(ConfigurationKeys.AzureFunctions.FunctionsWorkerRuntime) &&
+                   EnvironmentVariableExists(ConfigurationKeys.AzureFunctions.FunctionsExtensionVersion);
```
andrewlock, line 120: Same question as above, isn't this a duplicate with tweaks? Do we need both?

tracer/test/Datadog.Trace.Tests/Configuration/TracerSettingsServerlessTests.cs
```diff
@@ -80,19 +70,9 @@ public void HttpClientExcludedUrlSubstrings_AzureAppServices(string value, bool
...
-            { LambdaMetadata.HandlerEnvVar, System.Environment.GetEnvironmentVariable(LambdaMetadata.HandlerEnvVar) },
-            { LambdaMetadata.ExtensionPathEnvVar, System.Environment.GetEnvironmentVariable(LambdaMetadata.ExtensionPathEnvVar) },
-        };
-
-        System.Environment.SetEnvironmentVariable(LambdaMetadata.FunctionNameEnvVar, functionName);
-        System.Environment.SetEnvironmentVariable(LambdaMetadata.HandlerEnvVar, handlerName);
-        System.Environment.SetEnvironmentVariable(LambdaMetadata.ExtensionPathEnvVar, extensionPath);
-        return previous;
+        // using [CallerFilePath] to ensure we can a file path that exists
+        return callerFilePath!;
```
andrewlock, line 76: This won't always be true. If you enable deterministic builds, you get a dummy prefix `D:/_/` which won't point to a real file...

tracer/test/Datadog.Trace.Tests/Util/EnvironmentHelpersTests.cs
```diff
@@ -0,0 +1,100 @@
...
+        if (value)
+        {
+            Environment.SetEnvironmentVariable("AWS_LAMBDA_FUNCTION_NAME", "test");
+        }
+
+        EnvironmentHelpers.IsAwsLambda().Should().Be(value);
+    }
+
+    [Theory]
+    [PairwiseData]
```
andrewlock, line 80: You probably want `Combinatorial` so you actually test all combinations, seeing as these tests are fast

tracer/src/Datadog.Trace/ClrProfiler/ScopeFactory.cs
lucaspimentel: Thanks. I'll restore the null checks.

tracer/src/Datadog.Trace/ClrProfiler/Helpers/HttpBypassHelper.cs
lucaspimentel: Thanks. I'll restore the null checks.

tracer/src/Datadog.Trace/Configuration/ImmutableAzureAppServiceSettings.cs
lucaspimentel: Yup, I had the same thoughts. Definitely something to discuss for an upcoming refactor of all the "is this a serverless environment?" code. They are all different for Lambda, Azure, and GCP 😅

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
lucaspimentel: 100%

I would like to refactor many of these checks, maybe more "feature detection" than "environment detection." We could call it a _service for discovery_, if you will. 😝

tracer/test/Datadog.Trace.Tests/Configuration/TracerSettingsServerlessTests.cs
lucaspimentel: Huh, TIL. The code was already using `[CallerFilePath]`, I was just documenting _why_. I'll add this to the comment for now.

tracer/test/Datadog.Trace.Tests/Util/EnvironmentHelpersTests.cs
lucaspimentel: Thanks. Do we have docs on these attributes?

tracer/src/Datadog.Trace/ClrProfiler/ScopeFactory.cs
lucaspimentel: I mean, they all come from `HttpWebRequest.RequestUri`, which apparently can never be null, but better to be safe.

lucaspimentel: Nope, I lied: https://github.com/DataDog/dd-trace-dotnet/blob/1752abdc32c02c702b7c8374bfd1bae8b381d7db/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Remoting/Client/HttpProcessMessageIntegration.cs#L62-L62

tracer/src/Datadog.Trace/ClrProfiler/Helpers/HttpBypassHelper.cs
lucaspimentel: Done.

tracer/src/Datadog.Trace/Configuration/ImmutableAzureAppServiceSettings.cs
lucaspimentel: It was only used from the ctor, so I inlined it there for now. The mini-agent will not be used anymore so this code is going away, I just didn't want to rip it out just yet. I can do it in a separate PR for the next release.

tracer/src/Datadog.Trace/Configuration/ImmutableAzureAppServiceSettings.cs
lucaspimentel: Sure, I can change it. This method isn't called from anywhere yet, but I think we'll need it soon.

tracer/src/Datadog.Trace/Configuration/ImmutableAzureAppServiceSettings.cs
lucaspimentel: uhmm, looking at the code, `AgentProcessManager` is static and its `Initialize()` creates its own instance of `ImmutableAzureAppServiceSettings` (instead of using the one we already create in `TracerSettings`).

More stuff to refactor someday 😅

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
lucaspimentel: It was intentional. The _experimental_ "Serverless Mini-Agent" (aka SMA) has been deprecated for a while in favor of the "Serverless Compatibility Layer", which is what we are using for GA. I'll remove the rest of the SMA code in a separate PR.

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
lucaspimentel: Nothing has remote config except the full Go agent.

tracer/src/Datadog.Trace/Debugger/Caching/DefaultEnvironmentChecker.cs
lucaspimentel: This `CheckServerlessEnvironment()` is only used by `ConcurrentAdaptiveCache` to switch from the default 2048 capacity to 512.

https://github.com/DataDog/dd-trace-dotnet/blob/64feb5d50a15bba0b3ac471bf1473cfb380293d3/tracer/src/Datadog.Trace/Debugger/Caching/ConcurrentAdaptiveCache.cs#L386-L389

The previous implementation called `EnvironmentHelpers.IsServerlessEnvironment()` (the only code that did), and that also didn't check for AAS.

https://github.com/DataDog/dd-trace-dotnet/blob/1752abdc32c02c702b7c8374bfd1bae8b381d7db/tracer/src/Datadog.Trace/Util/EnvironmentHelpers.cs#L99-L108

I guess AAS is not "serverless" enough in this context, maybe because more memory is available?

lucaspimentel: The only other place the started using this recently was `LibDatadogAvailable` (incorrectly, IMO).

tracer/src/Datadog.Trace/LibDatadog/NativeInterop.cs
lucaspimentel: I suppose that's true, but we can't really check if we're using the bundle, can we? Do we even "support" using the bundle in AAS or AzFunc?

I'd rather we check for the _presence_ of the file itself rather that try to guess if it's there based other heuristics like hosting environment tracer deployment strategy.

lucaspimentel: Checking `IsUsingAzureAppServicesSiteExtension()` is actually a _fix_, because the previous checks would return `false` for Azure Functions using the AAS site extension, which does include `libdatadog`, even though they are "serverless."

tracer/src/Datadog.Trace/Util/EnvironmentHelpers.cs
lucaspimentel: > Isn't this a slightly different check to the one we're already doing in `ImmutableAzureAppServiceSettings?` 

If you mean the check we do in `ImmutableAzureAppServiceSettings` _after_ this PR: yes, they both check `WEBSITE_SITE_NAME` and should give the same result.

If you mean the check we do in `ImmutableAzureAppServiceSettings` _before_ this PR: that checked for `DD_AZURE_APP_SERVICES`, which is only set by the AAS Site Extension. This incorrectly returned `false` if you are instrumenting an Azure Function without the Site Extension. I split that out into a different check: `IsUsingAzureAppServicesSiteExtension()`.

>  Do we need/want both?

No, but for now my goal was to make the checks in `EnvironmentHelpers` and `ImmutableAzureAppServiceSettings` consistent (and correct), while we come back later and do a larger refactor to consolidate them.

tracer/src/Datadog.Trace/Util/EnvironmentHelpers.cs
lucaspimentel: See reply in the other thread 😅 
but for `IsAzureFunctions()`, I made them a bit more consistent after this comment (check the same env vars).

tracer/src/Datadog.Trace/Util/EnvironmentHelpers.cs
lucaspimentel: Yes, on my list of things to fix soon in a separate PR.

tracer/src/Datadog.Trace/Configuration/ImmutableAzureAppServiceSettings.cs
andrewlock: > It was only used from the ctor

Ah, I see, and presumably only public for testing - wouldn't have said anything if it was private 😄

tracer/src/Datadog.Trace/LibDatadog/NativeInterop.cs
andrewlock: > I'd rather we check for the presence of the file itself

This is generally a much harder proposition than it sounds. You know what would be best? Just always include the file 😉

tracer/src/Datadog.Trace/Util/EnvironmentHelpers.cs
andrewlock: > Isn't this a slightly different check

I actually meant that the settings version checks all the configuration sources for the presence of the variable, and sends all those checks to telemetry. This check _just_ checks the environment variables, and doesn't log anything.

> No, but for now my goal was to make the checks in EnvironmentHelpers and ImmutableAzureAppServiceSettings consistent

My point is mostly that they're _not_ consistent. Because if for some reason I set the web api key in my web.config (for example) then these would return different values 🤔

tracer/test/Datadog.Trace.Tests/Util/EnvironmentHelpersTests.cs
andrewlock: > Do we have docs on these attributes

https://andrewlock.net/simplifying-theory-test-data-with-xunit-combinatorial/

😉

andrewlock: Though also: https://github.com/DataDog/dd-trace-dotnet/blob/master/docs/development/XunitCombinatorial.md

PR #7262, opened 2025-07-22
[Test Optimization] Add pr.number tag to Github provider

tonyredondo:
## Summary of changes

This PR adds `pr.number` test tag for github action jobs.

## Reason for change

We have `pr.number` tag for several providers except github

### DIFF
--- a/tracer/src/Datadog.Trace/Ci/CiEnvironment/GithubActionsEnvironmentValues.cs
+++ b/tracer/src/Datadog.Trace/Ci/CiEnvironment/GithubActionsEnvironmentValues.cs
@@ -6,6 +6,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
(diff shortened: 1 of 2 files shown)


tracer/src/Datadog.Trace/Ci/CiEnvironment/GithubActionsEnvironmentValues.cs
```diff
@@ -102,6 +102,12 @@ private void LoadGithubEventJson()
             {
                 var githubEvent = File.ReadAllText(githubEventPath);
                 var githubEventObject = JObject.Parse(githubEvent);
+                var number = githubEventObject["number"]?.Value<int>();
+                if (number is > 0)
+                {
+                    PrNumber = number.ToString();
```
andrewlock:
```suggestion
                    PrNumber = number.ToString(CultureInfo.InvariantCulture);
```

PR #7265, opened 2025-07-23
Use `TrySetResult` instead of `SetResult`

andrewlock:
## Summary of changes

Use `TaskCompletionSource.TrySetResult` instead of `SetResult` (which throws)

## Reason for change

In #7200 we replaced `CancellationTokenSource` with `TaskCompletionSource`, but this caused some exceptions on shutdown due to double-disposal. In #7240 we fixed that, but did it in a way that doesn't make sense. We don't need to use `Interlocked.Exchange` we can just use `TrySetResult`.

## Implementation details

- Use `TrySetResult` instead of `SetResult` as the latter just calls the former and throws
- Did a sweep to update some other usages of `SetResult` which are unnecessarily risky

## Test coverage

Covered by existing test


tonyredondo (approved):
LGTM

PR #7271, opened 2025-07-24
[Azure Functions] Ignore additional URLs in outgoing HTTP client requests

lucaspimentel:
## Summary of changes

Added more Azure service endpoints to avoid creating http client spans for them.

Bonus:
- switch from read-only properties to`const string`:
  - `ImmutableAzureAppServiceSettings.DefaultHttpClientExclusions`
  - `LambdaMetadata.DefaultHttpClientExclusions`
- avoid the extra string allocation from `ToUpperInvariant()` in `ImmutableAzureAppServiceSettings.DefaultHttpClientExclusions`

## Reason for change

Discovered while testing new Azure Functions support.

## Implementation details

- Added default exclusions in Azure App Services (including Azure Functions):
  - Azure Monitor endpoints (`monitor.azure.com`)
  - Azure Functions table storage endpoints (`table.core.windows.net`)
  - Azure Functions diagnostic events endpoints
  - Azure Functions CDN endpoints (`cdn.functions.azure.com`)
- Code refactoring: Extracted HTTP client exclusion logic into a dedicated method `GetDefaultHttpClientExclusions()` in `TracerSettings` for better maintainability

## Test coverage

Added tests for new exclusions.

## Other details
Fixes APMSVLS-59


tracer/test/Datadog.Trace.Tests/Configuration/TracerSettingsServerlessTests.cs
```diff
@@ -57,6 +48,40 @@ public void HttpClientExcludedUrlSubstrings(string value, bool isRunningInAppSer
         }
     }
 
+    // These tests rely on Lambda.Create() which uses environment variables
+    // See TracerSettingsTests for tests which don't rely on any environment variables
```
andrewlock: Is this a lie? 🤔 Or are they used _internally_? Because if the latter, eww, that's gross, but a problem for another time I guess 😅

tracer/src/Datadog.Trace/Configuration/ImmutableAzureAppServiceSettings.cs
```diff
@@ -21,9 +21,24 @@ namespace Datadog.Trace.Configuration
...
+#pragma warning disable SA1025 // CodeMustNotContainMultipleWhitespaceInARow
+            ".LOGS.DATADOGHQ.," +
+            ".SERVICES.VISUALSTUDIO.," +
+            ".APPLICATIONINSIGHTS.AZURE.," +
+            ".MONITOR.AZURE.," +
+            ".BLOB.CORE.WINDOWS.NET/AZURE-WEBJOBS," +                   // no trailing slash, /azure-webjobs-hosts/*, /azure-webjobs-secrets/*, etc
+            ".TABLE.CORE.WINDOWS.NET/TABLES," +                         // no trailing slash, /Tables, /Tables('AzureFunctionsDiagnosticEventsCheck')
+            ".TABLE.CORE.WINDOWS.NET/AZUREFUNCTIONSDIAGNOSTICEVENTS," + // no trailing slash, /AzureFunctionsDiagnosticEvents202507()?$format=...
+            ".AZUREWEBSITES.NET/ADMIN/," +                              // trailing slash, /admin/*
+            "CDN.FUNCTIONS.AZURE.COM/PUBLIC/";                          // trailing slash, /public/*
```
andrewlock: How sure are you that some of these won't be used by customers too? 🤔e.g. the CDN one looks like it could happen?

andrewlock (approved):

tracer/test/Datadog.Trace.Tests/Configuration/TracerSettingsServerlessTests.cs
lucaspimentel: Oops. Copy/paste from the original test which handled both AWS Lambda and AAS. I split them up, so this now only true for `HttpClientExcludedUrlSubstrings_AwsLambda`, not for `HttpClientExcludedUrlSubstrings_AzureAppServices`.

edit: removed this comment on the AAS test, left the comment on the Lambda test

lucaspimentel: But yes, `Lambda.Create()` reads env vars directly to check if we're running in AWS Lambda, so that test has to set env vars.

> _a problem for another time I guess_

Yes. I want to refactor all the "detect serverless environments" code when I have time. Each one is completely different and they all have different issues 😅

tracer/src/Datadog.Trace/Configuration/ImmutableAzureAppServiceSettings.cs
lucaspimentel: > _How sure are you..._

I'm not, but the same question applies to many of the others we already ignored before this change. Some of these are so noisy that I'm convinced it's better to ignore them by default even if we get false positives. If it comes up, users can override the default value and remove any URLs they want to see.

lucaspimentel: I narrowed the CDN one to `CDN.FUNCTIONS.AZURE.COM/PUBLIC/EXTENSIONBUNDLES/`.

bouwkast (approved):

PR #7272, opened 2025-07-24
[Azure Functions] Fix operation name

lucaspimentel:
## Summary of changes

Fix operation name.

## Reason for change

@bouwkast:
> I think this is a breaking change no?

The Agent is _already_ normalizing `"azure-functions.invoke"` to `"azure_functions.invoke"` before any trace stats are computed and the traces reach the backend because the hyphen is not allowed, so I don't think it's a breaking change.

This change fixes sampling rules that try to match by operation name, because the UI shows the operation name as `"azure_functions.invoke"` to users, but if they try that in a sampling rule, it won't match.

## Implementation details

Change `"azure-functions.invoke"` to `"azure_functions.invoke"`.

## Test coverage

Updated test snapshots.

## Other details
Fixes APMSVLS-125


bouwkast (approved):
Thanks 👍

PR #7280, opened 2025-07-25
Update xunit to latest, fix analyzer warnings, and update analyzer tests

andrewlock:
## Summary of changes

- Update `xunit` to latest v2 package version (`v2.9.3`)
- Update Analyzer testing libraries to the new test-framework agnostic versions
- Fix all the errors xunit now flags

## Reason for change

As part of the .NET 10 work we tried updating to the latest xunit. It's not _entirely_ clear if that work was actually necessary (there appeared to be a bug in preview 5 that didn't impact preview 6) but rather than waste the effort, pulling this out into a separate PR.

## Implementation details

Mostly just bumping package versions. For the analyzer library changes, I followed [this guide](https://github.com/dotnet/roslyn-sdk/blob/main/src/Microsoft.CodeAnalysis.Testing/README.md) to update. For xunit I bumped required packages, and then fixed a bunch of errors the xunit.analyzers started flagging:

- `Assert.Empty` -> use FluentAssertions
- "Incorrect type" -> Use `int` instead of `HttpStatusCode`, `NullableBooleanTestCases` instead of `BooleanTestCases` (for example)
- "Don't use `ConfigureAwait(false)`" -> Remove it (xunit has a sync context)
- "xUnit1031 - Test methods should not use Blocking operations" - I _mostly_ ignored this one, because the tests are in testing for low-level stuff (call target machinery) and didn't want to change the behavior

## Test coverage

This is the test. Need to double check the tests are definitely _actually_ running though... 

## Other details

Related to
- #7170 
- https://datadoghq.atlassian.net/browse/LANGPLAT-632


tracer/test/Datadog.Trace.DuckTyping.Tests/Datadog.Trace.DuckTyping.Tests.csproj
```diff
@@ -3,7 +3,7 @@
   <ItemGroup>
     <PackageReference Include="Serilog" Version="2.10.0" />
     <PackageReference Include="Serilog.Sinks.Console" Version="3.1.1" />
-    <PackageReference Include="System.Collections.Immutable" Version="5.0.0" />
+    <PackageReference Include="System.Collections.Immutable" Version="6.0.0" />
```
bouwkast, line 6: Was this a required bump? Don't think it is important or not just curious.

andrewlock: > Was this a required bump? Don't think it is important or not just curious.

Yup, it was! Can't remember exactly where from, but it was required 🤷‍♂️

lucaspimentel (approved):

dudikeleti (approved):
Thanks

daniel-romano-DD (approved):
TY!

PR #7282, opened 2025-07-25
[HandsOff Config] Read sources from libdatadog

anna-git:
## Summary of changes

- Introducing stable config managed part with libdatadog: reading 2 yml files through libdatadog: one local file, one fleet file.
- Added some unit testing for GlobalConfigurationSource

## Reason for change
Stable configuration managed implementation

## Implementation details

- Added DD_APPLICATION_MONITORING_CONFIG_FILE_ENABLED  to deactivate hands off config for support cases. Env var is added here as well: https://github.com/DataDog/dd-go/pull/193905
- Prevent logging when initializing stable config from libdatadog as the Lazy IsLibDatadogAvailable ends up calling itself again even if it uses a class with a static logger field
      - Using results instead of logging to log later on
      - transforming logger to lazy<logger> in some class      - 

## Test coverage

- Added some unit testing for GlobalConfigurationSource

### DIFF
--- a/tracer/src/Datadog.Trace/LibDatadog/HandsOffConfiguration/ConfiguratorHelper.cs
+++ b/tracer/src/Datadog.Trace/LibDatadog/HandsOffConfiguration/ConfiguratorHelper.cs
@@ -17,7 +17,7 @@ internal struct ConfiguratorHelper
     {
         if (isLibdatadogAvailable is false)
         {
-            return new ConfigurationResult(null, "Skipping hands-off configuration: as LibDatadog is not available", Result.LibDatadogUnavailable);
+            return new ConfigurationResult(null, "Skipping hands-off configuration because LibDatadog is not available", Result.LibDatadogUnavailable);
         }
 
         if (isLibdatadogAvailable is null)
--- a/tracer/src/Datadog.Trace/LibDatadog/LibDatadogAvailableResult.cs
+++ b/tracer/src/Datadog.Trace/LibDatadog/LibDatadogAvailableResult.cs
@@ -8,7 +8,7 @@ using System;
 
 namespace Datadog.Trace.LibDatadog;
 
-internal struct LibDatadogAvailableResult(bool isAvailable, Exception? exception = null)
+internal readonly struct LibDatadogAvailableResult(bool isAvailable, Exception? exception = null)
 {
     public bool IsAvailable { get; } = isAvailable;
 
--- a/tracer/src/Datadog.Trace/TracerManagerFactory.cs
+++ b/tracer/src/Datadog.Trace/TracerManagerFactory.cs
@@ -118,7 +118,7 @@ namespace Datadog.Trace
             var libdatadogAvailaibility = LibDatadogAvailaibilityHelper.IsLibDatadogAvailable;
             if (libdatadogAvailaibility.Exception is not null)
             {
-                Log.Warning(libdatadogAvailaibility.Exception, "An exception happened when trying to see if libdatadog was available");
+                Log.Warning(libdatadogAvailaibility.Exception, "An exception occurred while checking if libdatadog is available");
             }
 
             var defaultServiceName = settings.ServiceName ??


tracer/src/Datadog.Trace/Configuration/ConfigurationSources/GlobalConfigurationSource.cs
```diff
@@ -18,21 +21,57 @@ namespace Datadog.Trace.Configuration;
...
+        Exception? exception = null;
+        var resultType = Result.Success;
         var configurationSource = new CompositeConfigurationSource();
-        configurationSource.Add(new EnvironmentConfigurationSource());
+        var environmentSource = new EnvironmentConfigurationSource();
+        var configBuilder = new ConfigurationBuilder(environmentSource, TelemetryFactory.Config);
+        var applicationMonitoringConfigFileEnabled = configBuilder.WithKeys(ConfigurationKeys.ApplicationMonitoringConfigFileEnabled).AsBool(true);
+        var debugEnabled = configBuilder.WithKeys(ConfigurationKeys.DebugEnabled).AsBool(false);
+        if (applicationMonitoringConfigFileEnabled)
+        {
```
andrewlock: nit: `debugEnabled` can be moved closer to usage
```suggestion
        if (applicationMonitoringConfigFileEnabled)
        {
            var debugEnabled = configBuilder.WithKeys(ConfigurationKeys.DebugEnabled).AsBool(false);
```

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/GlobalConfigurationSource.cs
```diff
@@ -18,21 +21,57 @@ namespace Datadog.Trace.Configuration;
...
-    internal static CompositeConfigurationSource CreateDefaultConfigurationSource()
+    internal static GlobalConfigurationSourceResult CreateDefaultConfigurationSource(string? handsOffLocalConfigPath = null, string? handsOffFleetConfigPath = null, bool? isLibdatadogAvailable = null)
     {
         // env > AppSettings > datadog.json
+        string? message = null;
+        Exception? exception = null;
+        var resultType = Result.Success;
         var configurationSource = new CompositeConfigurationSource();
-        configurationSource.Add(new EnvironmentConfigurationSource());
+        var environmentSource = new EnvironmentConfigurationSource();
```
andrewlock, line 46: [According to the RFC](https://docs.google.com/document/d/1TFn7IjzpU4kMLmPZHb5NRE98RDyJfbqaZ-wni7ZRmxw/edit?tab=t.0#heading=h.vw7xhmwnlkbk) I think the fleet installer stuff should all be added _after_ the datadog.json and web.config sources?

tracer/src/Datadog.Trace/Configuration/GlobalSettings.cs
```diff
@@ -121,8 +122,13 @@ public static void Reload()
...
             TelemetryFactory.Metrics.Record(PublicApiUsage.GlobalSettings_Reload);
             DatadogLogging.Reset();
-            LibDatadog.Logger.Instance.SetLogLevel(debugEnabled: false);
-            GlobalConfigurationSource.Reload();
+            var isLibdatadogAvailable = LibDatadogAvailaibilityHelper.IsLibDatadogAvailable;
+            if (isLibdatadogAvailable.IsAvailable)
+            {
+                LibDatadog.Logger.Instance.SetLogLevel(debugEnabled: false);
+                GlobalConfigurationSource.Reload(isLibdatadogAvailable.IsAvailable);
+            }
```
andrewlock: This should be outside the if:

```suggestion
            }

            GlobalConfigurationSource.Reload(isLibdatadogAvailable.IsAvailable);
```

tracer/src/Datadog.Trace/FrameworkDescription.cs
```diff
@@ -4,18 +4,15 @@
...
 using Datadog.Trace.Logging;
-using Microsoft.Win32;
 
 namespace Datadog.Trace
 {
     internal partial class FrameworkDescription
     {
-        private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(FrameworkDescription));
+        // don't use a direct logger field as this will trigger ConfigurationSources building which are themselves calling this path though IsProfilerAttached
+        private static readonly Lazy<IDatadogLogger> Log = new(() => DatadogLogging.GetLoggerFor(typeof(FrameworkDescription)));
```
andrewlock: This makes me very nervous tbh :/ I don't this actually prevents the recursive loop issue if an error actually occurs and we try to write a log... It "fixed" it here because you didn't hit the errors, but if you _do_... 💥

tracer/src/Datadog.Trace/LibDatadog/CString.cs
```diff
@@ -0,0 +1,53 @@
...
+#nullable enable
+using System;
+using System.Runtime.InteropServices;
+using System.Text;
+using Datadog.Trace.Util;
+
+namespace Datadog.Trace.LibDatadog;
+
+[StructLayout(LayoutKind.Sequential)]
+internal struct CString : IDisposable
```
andrewlock, line 15: I think this needs to be derived from `SafeHandle` so that we have a finalizer etc too? 🤔

tracer/src/Datadog.Trace/LibDatadog/CString.cs
```diff
@@ -0,0 +1,53 @@
...
+    internal CString(string? str)
+    {
+        if (string.IsNullOrEmpty(str))
+        {
+            Ptr = IntPtr.Zero;
+            Length = UIntPtr.Zero;
+        }
+        else
+        {
+            var encoding = Encoding.UTF8;
```
andrewlock: This will add a BOM unless you use the `StringEncoding` version (requires an extra namespace)
```suggestion
            var encoding = StringEncoding.UTF8;
```

tracer/src/Datadog.Trace/LibDatadog/CString.cs
```diff
@@ -0,0 +1,53 @@
...
+            var maxBytesCount = encoding.GetMaxByteCount(str!.Length);
+            Ptr = Marshal.AllocHGlobal(maxBytesCount);
```
andrewlock, line 31: You might need a try-catch here, so that if `GetBytes` throws and the instance is not created, you still free the ptr.

tracer/src/Datadog.Trace/LibDatadog/HandsOffConfiguration/ConfigurationSuccessResult.cs
```diff
@@ -0,0 +1,17 @@
...
+
+using System.Collections.Generic;
+
+namespace Datadog.Trace.LibDatadog.HandsOffConfiguration;
+
+internal readonly struct ConfigurationSuccessResult(IDictionary<string, string> configEntriesLocal, IDictionary<string, string> configEntriesFleet)
+{
+    public IDictionary<string, string> ConfigEntriesLocal { get; } = configEntriesLocal;
+
+    public IDictionary<string, string> ConfigEntriesFleet { get; } = configEntriesFleet;
```
andrewlock: nit: you get slightly better performance if you use the "real" type instead of the interface (something we should look at more generally!)
```suggestion
    public Dictionary<string, string> ConfigEntriesLocal { get; } = configEntriesLocal;

    public Dictionary<string, string> ConfigEntriesFleet { get; } = configEntriesFleet;
```

tracer/src/Datadog.Trace/Ci/Configuration/TestOptimizationSettings.cs
```diff
@@ -345,7 +345,7 @@ internal void SetDefaultManualInstrumentationSettings()
         }
 
         private TracerSettings InitializeTracerSettings()
-            => InitializeTracerSettings(GlobalConfigurationSource.CreateDefaultConfigurationSource());
+            => InitializeTracerSettings(GlobalConfigurationSource.CreateDefaultConfigurationSource().ConfigurationSource);
```
andrewlock: This looks wrong, I think it should probably be accessing the Instance cc @tonyredondo  
```suggestion
            => InitializeTracerSettings(GlobalConfigurationSource.Instance);
```

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
```diff
@@ -52,6 +52,7 @@ public static class Instrumentation
 
         /// <summary>
         /// Gets a value indicating whether Datadog's profiler is attached to the current process.
+        /// Should not log anything. Logging here could be an issue as this is accessed before Configuration objects are built. Logging here could create a loop where Configuration building tests if profiler is attached to access libdatadog, the test wants to log, the Logger being created for the first time tried to access the Configuration object.
```
andrewlock: We should probably add this note to `NativeMethods.IsProfilerAttached` too

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/GlobalConfigurationSource.cs
```diff
@@ -18,21 +21,57 @@ namespace Datadog.Trace.Configuration;
...
-    internal static IConfigurationSource Instance { get; private set;  } = CreateDefaultConfigurationSource();
+    internal static IConfigurationSource Instance { get; private set; }
 
     /// <summary>
     /// Creates a <see cref="IConfigurationSource"/> by combining environment variables,
     /// AppSettings where available, and a local datadog.json file, if present.
     /// </summary>
     /// <returns>A new <see cref="IConfigurationSource"/> instance.</returns>
-    internal static CompositeConfigurationSource CreateDefaultConfigurationSource()
+    internal static GlobalConfigurationSourceResult CreateDefaultConfigurationSource(string? handsOffLocalConfigPath = null, string? handsOffFleetConfigPath = null, bool? isLibdatadogAvailable = null)
```
andrewlock, line 40: AFAICT we never _do_ anything with the returned result other than throw it away, so I'm not sure it's adding much value 🤔 If we want to (for example) _log_ it, we should likely store the result somewhere it can be retrieved later and logged. e.g. 

```csharp
internal static IConfigurationSource Instance => CreationResult.ConfigurationSource;

internal static GlobalConfigurationSourceResult CreationResult { get; private set; }
```

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/HandsOffConfigurationSource.cs
```diff
@@ -0,0 +1,27 @@
...
+
+#nullable enable
+using System;
+using System.Collections.Generic;
+using Datadog.Trace.Configuration.Telemetry;
+using Datadog.Trace.LibDatadog.HandsOffConfiguration;
+
+namespace Datadog.Trace.Configuration.ConfigurationSources;
+
+internal class HandsOffConfigurationSource(IDictionary<string, string> configurations, bool localFile)
```
andrewlock: nit: we can prob just make these `Dictionary<string, string>` (avoids interface dispatch later)

tracer/src/Datadog.Trace/LibDatadog/HandsOffConfiguration/ConfigurationResult.cs
```diff
@@ -0,0 +1,29 @@
...
+using System;
+
+namespace Datadog.Trace.LibDatadog.HandsOffConfiguration;
+
+internal enum Result
+{
+    LibDatadogCallError,
+    LibDatadogUnavailable,
+    Success,
+    ApplicationMonitoringConfigFileEnabled
```
andrewlock: Is this one actually used? What does it mean? Should it be `ApplicationMonitoringConfigFileDisabled` instead?

tracer/src/Datadog.Trace/LibDatadog/HandsOffConfiguration/ConfiguratorHelper.cs
```diff
@@ -0,0 +1,98 @@
...
+namespace Datadog.Trace.LibDatadog.HandsOffConfiguration;
+
+internal struct ConfiguratorHelper
+{
+    internal static ConfigurationResult GetConfiguration(bool debugEnabled, string? handsOffLocalConfigPath, string? handsOffFleetConfigPath, bool? isLibdatadogAvailable = null)
+    {
+        if (isLibdatadogAvailable is false or null)
+        {
+            var isLibdatadogAvailableEval = LibDatadogAvailaibilityHelper.IsLibDatadogAvailable;
+            if (!isLibdatadogAvailableEval.IsAvailable)
```
andrewlock: This doesn't look right to me 🤔 If `isLibdatadogAvailable` == false, you shouldn't be ignoring that, you should just not do it I think? Tbh, I'd rather we made it a non-null required field instead. Calling global static stateful methods is generally problematic

tracer/src/Datadog.Trace/LibDatadog/HandsOffConfiguration/ConfiguratorHelper.cs
```diff
@@ -0,0 +1,98 @@
...
+            {
+                return new ConfigurationResult(null, "Skipping hands-off configuration: as LibDatadog is not available", Result.LibDatadogUnavailable, isLibdatadogAvailableEval.Exception);
+            }
+        }
+
+        try
+        {
+            var configHandle = NativeInterop.LibraryConfig.ConfiguratorNew(debugEnabled ? (byte)1 : (byte)0, new CharSlice(TracerConstants.Language));
```
andrewlock: I think we're leaking stuff here

- `CharSlice` needs to be disposed
- We should wrap the created `configHandle` in a try-finally and put the `ConfiguratorDrop` in the finally block, to make sure it's always freed

tracer/src/Datadog.Trace/LibDatadog/HandsOffConfiguration/ConfiguratorHelper.cs
```diff
@@ -0,0 +1,98 @@
...
+            }
+
+            var configurationResult = NativeInterop.LibraryConfig.ConfiguratorGet(configHandle);
+            var result = configurationResult.Result;
+            ConfigurationResult configurationResultReturned;
+            if (configurationResult.Tag == ResultTag.Err)
+            {
+                var error = result.Error.Message.ToUtf8String();
+                NativeInterop.Common.DropError(ref result.Error);
+                configurationResultReturned = new ConfigurationResult(null, error, Result.LibDatadogCallError);
```
andrewlock: nit: I would just return early here instead of having the big nested `else`. If you use the `try-finally` above then that's easy to do here. We should also probably guard to make sure we always drop the error too:

```suggestion
            if (configurationResult.Tag == ResultTag.Err)
            {
                string error;
                try
                {
                    error = result.Error.Message.ToUtf8String();
                }
                finally
                {
                    NativeInterop.Common.DropError(ref result.Error);
                }
                
                return new ConfigurationResult(null, error, Result.LibDatadogCallError);
```

We need to remember that _any_ NativeInterop call could throw, so we need to make sure our try-catch-finally etc are all in order so that there's no way for an error to escape or for us to leak memory here if it does!

tracer/src/Datadog.Trace/LibDatadog/HandsOffConfiguration/ConfiguratorHelper.cs
```diff
@@ -0,0 +1,98 @@
...
+            }
+            else
+            {
+                ref var configurationResultRef = ref configurationResult.Result;
```
andrewlock: Pinging @tonyredondo to look in on unsafe code as requested 😄

tracer/src/Datadog.Trace/LibDatadog/HandsOffConfiguration/ConfiguratorHelper.cs
```diff
@@ -0,0 +1,98 @@
...
+                            configEntriesRemote.Add(name, value);
+                        }
+                        else if (libraryConfig->Source == LibraryConfigSource.LocalStableConfig)
+                        {
+                            configEntriesLocal.Add(name, value);
+                        }
+                    }
+                }
+
+                NativeInterop.LibraryConfig.LibraryConfigDrop(configurationResultRef.Ok);
```
andrewlock: I don't quite follow the need for the `ref var`, but I'll take your word for it. Either way, all this cleanup should be in one or more `finally` calls (potentially, multiple, so that they're _all_ always called regardless of whether a `Dispose()` or `Drop` call fails

tracer/src/Datadog.Trace/LibDatadog/ServiceDiscovery/ServiceDiscoveryHelper.cs
```diff
@@ -61,6 +61,11 @@ internal static StoreMetadataResult StoreTracerMetadata(TracerSettings tracerSet
             "Skipping storage of tracer metadata with LibDatadog: Platform supported: {PlatformIsSupported}, Deployment supported: {DeploymentIsSupported}",
             platformIsSupported,
             deploymentIsSupported);
+        if (deploymentIsSupported.Exception is not null)
+        {
+            Log.Warning(deploymentIsSupported.Exception, "An exception happened when trying to see if libdatadog was available");
```
andrewlock: There's potentially lots of places we _Could_ log this, but we really only want to log it once, so it probably make sense in TracerManagerFactory instead? (where we're calling libdatadog already)

tracer/src/Datadog.Trace/Logging/Internal/DatadogLoggingFactory.cs
```diff
@@ -166,7 +166,7 @@ private static ConsoleLoggingConfiguration GetConsoleLoggingConfiguration(IConfi
 
     // Internal for testing
     internal static string GetLogDirectory(IConfigurationTelemetry telemetry)
-        => GetLogDirectory(GlobalConfigurationSource.CreateDefaultConfigurationSource(), telemetry);
+        => GetLogDirectory(GlobalConfigurationSource.CreateDefaultConfigurationSource().ConfigurationSource, telemetry);
```
andrewlock:
```suggestion
        => GetLogDirectory(GlobalConfigurationSource.Instance, telemetry);
```

tracer/src/Datadog.Trace/Util/NativeStringHelper.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public static unsafe string GetString(nint pointer, nuint length)
+    {
+#if NETCOREAPP3_0_OR_GREATER
+        var span = new ReadOnlySpan<byte>((byte*)pointer, (int)length);
+        return Encoding.UTF8.GetString(span);
+#else
+        return Encoding.UTF8.GetString((byte*)pointer, (int)length);
+#endif
+    }
```
andrewlock: I think we should probably move this into the LibDatadog folder, instead of the `Util` dumping ground. Also, there's not really any benefit to using readonlyspan here + we need to use the BOM-free version for libdatadog

```suggestion
    public static unsafe string GetString(nint pointer, nuint length)
        => StringEncoding.UTF8.GetString((byte*)pointer, (int)length);
```

tracer/src/Datadog.Tracer.Native/cor_profiler.cpp
```diff
@@ -971,6 +971,7 @@ HRESULT CorProfiler::TryRejitModule(ModuleID module_id, std::vector<ModuleID>& m
             RewritingPInvokeMaps(module_metadata, libdatadog_exporter_nativemethods_type, libdatadog_filepath);
             RewritingPInvokeMaps(module_metadata, libdatadog_config_nativemethods_type, libdatadog_filepath);
             RewritingPInvokeMaps(module_metadata, libdatadog_logger_nativemethods_type, libdatadog_filepath);
+            RewritingPInvokeMaps(module_metadata, libdatadog_libraryconfig_nativemethods_type, libdatadog_filepath);
```
andrewlock, line 974: I'm guessing this is left over from a refactor? You moved it all into `config` before (which is fine I think
```suggestion
```

tracer/src/Datadog.Tracer.Native/dd_profiler_constants.h
```diff
@@ -155,6 +155,7 @@ const shared::WSTRING libdatadog_exporter_nativemethods_type = WStr("Datadog.Tra
 const shared::WSTRING libdatadog_config_nativemethods_type = WStr("Datadog.Trace.LibDatadog.NativeInterop+Config");
 const shared::WSTRING libdatadog_common_nativemethods_type = WStr("Datadog.Trace.LibDatadog.NativeInterop+Common");
 const shared::WSTRING libdatadog_logger_nativemethods_type = WStr("Datadog.Trace.LibDatadog.NativeInterop+Logger");
+const shared::WSTRING libdatadog_libraryconfig_nativemethods_type = WStr("Datadog.Trace.LibDatadog.NativeInterop+LibraryConfig");
```
andrewlock, line 158:
```suggestion
```

tracer/test/Datadog.Trace.Tests/Configuration/GlobalConfigurationSourceTests.cs
```diff
@@ -0,0 +1,81 @@
...
+        // Precedence is as follows:
+        // 1. fleet file
+        // 2. env var
+        // 3. local file
+
+        // KEY1: defined in fleet file: this should be ignored
+        Environment.SetEnvironmentVariable("KEY1", "local_env_var", EnvironmentVariableTarget.Process);
+
+        // KEY2: defined false in local file: this should be applied over the local file
+        Environment.SetEnvironmentVariable("KEY2", "true", EnvironmentVariableTarget.Process);
```
andrewlock: If you're setting environment variables in tests, you need to use the `[EnvironmentRestorer]` attribute with the required env vars, to ensure it's not flaky

tracer/test/Datadog.Trace.Tests/Configuration/GlobalConfigurationSourceTests.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        // KEY2: only defined in env var, should prevail
+        Environment.SetEnvironmentVariable("KEY3", "false", EnvironmentVariableTarget.Process);
+        var result = GlobalConfigurationSource.CreateDefaultConfigurationSource(
+            handsOffLocalConfigPath: Path.Combine("Configuration", "HandsOffConfigData", "application_monitoring.yml"),
+            handsOffFleetConfigPath: Path.Combine("Configuration", "HandsOffConfigData", "application_monitoring_fleet.yml"),
+            isLibdatadogAvailable: true);
+
+        var source = result.ConfigurationSource;
```
andrewlock: This is nice, but I think you can probably test it more directly by testing the actual configuration sources that are in the returned value. Currently your results are dependent on a hole stack of additional behaviour (how `GetBool()` works etc)

Instead, I think you can do something like this:

```csharp
var result = GlobalConfigurationSource.CreateDefaultConfigurationSource(...)

result.Should().Be(something);
var source = result.ConfigurationSources.Should().NotBeNull().And.BeOfType<CompositeConfigurationSource>().Subject;

var sourcesInComposite = source.ToList();

source[0].Should().BeOfType<EnvironmentConfigurationSource>();
source[1].Should()...
// etc
```

That way you don't have to set env vars at all

tracer/test/Datadog.Trace.Tests/Configuration/GlobalConfigurationSourceTests.cs
```diff
@@ -0,0 +1,81 @@
...
+
+    [Fact]
+    public void TestErrorHandsOffConfigFile()
+    {
+        var handsOffErrorPath = Path.Combine("Configuration", "HandsOffConfigData", "corrupt_file.yml");
+        var result = LibDatadog.HandsOffConfiguration.ConfiguratorHelper.GetConfiguration(debugEnabled: true, handsOffLocalConfigPath: handsOffErrorPath, handsOffFleetConfigPath: handsOffErrorPath, isLibdatadogAvailable: true);
+        result.ConfigurationSuccessResult.Should().BeNull();
+        result.ErrorMessage.Should().NotBeNull();
+        result.ErrorMessage.Should().Be("apm_configuration_default: invalid type: string \"DD_TRACE_DEBUG': true, DD_ENV: \", expected struct ConfigMap(HashMap<String, String>) at line 3 column 3");
+    }
```
andrewlock: I would recommend moving this to a separate `HandsOffConfigurationTests` type, and test a broader set of values, including as many edge cases as you can e.g. missing files/incorrect paths, incorrect types etc

tracer/test/Datadog.Trace.Tests/Datadog.Trace.Tests.csproj
```diff
@@ -7,6 +7,27 @@
...
+    </None>
+    <None Update="Configuration\HandsOffConfigData\application_monitoring.yml">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Update="Configuration\HandsOffConfigData\application_monitoring_fleet.yml">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Update="Configuration\HandsOffConfigData\corrupt_file.yml">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
```
andrewlock: There's links to 7 files here, but only 3 files in the PR? 🤔

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
lucaspimentel: _super_ nitty 😅 , but I would place this new text in `<remarks>`, not `<summary>`.

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
```diff
@@ -52,6 +52,7 @@ public static class Instrumentation
 
         /// <summary>
         /// Gets a value indicating whether Datadog's profiler is attached to the current process.
```
lucaspimentel: While you're here, do you mind change this line a bit? There's a lot of old mentions of "profiler" which pre-date the Continuous Profiler product, which is confusing.

Something like:
```suggestion
        /// Gets a value indicating whether Datadog's instrumentation library (aka CLR profiler) is attached to the current process.
```


(Even better if we could rename `ProfilerAttached`, but I'm sure it's accessed via reflection somewhere so we probably shouldn't.)

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/GlobalConfigurationSource.cs
```diff
@@ -18,21 +21,57 @@ namespace Datadog.Trace.Configuration;
...
     {
         // env > AppSettings > datadog.json
```
lucaspimentel: We should update this comment to include the precedence of the new sources. From the code, it looks like:

```diff
+ fleet hands-off config, if enabled
environment variables
+ local hands-off config, if enabled
AppSettings, if .NET Framework
JSON file, if file exists
```

Is that correct?

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/GlobalConfigurationSource.cs
```diff
@@ -18,21 +21,57 @@ namespace Datadog.Trace.Configuration;
...
+                message = configsResult.ErrorMessage;
+                exception = configsResult.Exception;
+                resultType = configsResult.Result;
+                configurationSource.Add(environmentSource);
+            }
+        }
+        else
+        {
+            resultType = Result.ApplicationMonitoringConfigFileEnabled;
+            message = $"As {nameof(ConfigurationKeys.ApplicationMonitoringConfigFileEnabled)}  is disabled, not using hands-off configuration";
```
lucaspimentel:
```suggestion
            message = $"{nameof(ConfigurationKeys.ApplicationMonitoringConfigFileEnabled)} is disabled, not using hands-off configuration";
```

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -559,6 +559,13 @@ internal static partial class ConfigurationKeys
         /// <seealso cref="TracerSettings.GraphQLErrorExtensions"/>
         public const string GraphQLErrorExtensions = "DD_TRACE_GRAPHQL_ERROR_EXTENSIONS";
 
+        /// <summary>
+        /// Configuration key for deactivating reading the application monitoring config file through libdatadog (hands off config).
+        /// True by default
+        /// </summary>
+        /// <seealso cref="TracerSettings.GraphQLErrorExtensions"/>
```
lucaspimentel: `GraphQLErrorExtensions` looks like copy/paste error

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/HandsOffConfigurationSource.cs
lucaspimentel: related, we can make classes `sealed` as well:
```
internal sealed class HandsOffConfigurationSource
```

tracer/src/Datadog.Trace/LibDatadog/LibDatadogAvailaibilityHelper.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        var res = IsServerlessEnvironment();
+        var isServerless = res.Item1;
+        var possibleException = res.Item2;
+        return new(!isServerless && Instrumentation.ProfilerAttached, possibleException);
+    });
+
+    public static LibDatadogAvailableResult IsLibDatadogAvailable => LibDatadogAvailable.Value;
+
+    private static Tuple<bool, Exception?> IsServerlessEnvironment()
```
lucaspimentel: Not a big deal since it only happens once, but we can use two `out` parameters and avoid the `Tuple<>` allocation. Sadly, we can't use `ValueTuple<>`.

tracer/src/Datadog.Trace/LibDatadog/LibDatadogAvailaibilityHelper.cs
```diff
@@ -0,0 +1,64 @@
...
+        bool EnvVarExistsNoLogging(string key)
+        {
+            string? value = null;
+            try
+            {
+                value = Environment.GetEnvironmentVariable(key);
+            }
+            catch (Exception ex)
+            {
+                exception = ex;
```
lucaspimentel: This local method is capturing the outer `exception`, which incurs a closure allocation. We should try to declare local methods as `static` if possible to avoid doing this accidentally. Maybe try refactoring the method to use `out` parameters to avoid the closure.

tracer/src/Datadog.Trace/LibDatadog/LibDatadogAvailaibilityHelper.cs
```diff
@@ -0,0 +1,64 @@
...
+
+    public static LibDatadogAvailableResult IsLibDatadogAvailable => LibDatadogAvailable.Value;
+
+    private static Tuple<bool, Exception?> IsServerlessEnvironment()
+    {
+        Exception? exception = null;
+        if (EnvVarExistsNoLogging(LambdaMetadata.FunctionNameEnvVar)
+         || (EnvVarExistsNoLogging(ConfigurationKeys.AzureAppService.SiteNameKey) && !EnvVarExistsNoLogging(ConfigurationKeys.AzureAppService.AzureAppServicesContextKey))
+         || (EnvVarExistsNoLogging(ConfigurationKeys.GCPFunction.FunctionNameKey) && EnvVarExistsNoLogging(ConfigurationKeys.GCPFunction.FunctionTargetKey))
+         || (EnvVarExistsNoLogging(ConfigurationKeys.GCPFunction.DeprecatedFunctionNameKey) && EnvVarExistsNoLogging(ConfigurationKeys.GCPFunction.DeprecatedProjectKey)))
```
lucaspimentel: Hmmm, it's not great that we're adding _yet another_ place where we check for these env vars 😅 . I made some changes recently to these and I had to do it in several places because it was duplicated in several places (and not all of them were consistent!).

In nothing else, maybe for now just create a copy of `Datadog.Trace.Util.EnvironmentHelpers` that doesn't have a logger? At least that way the code is side-by-side and not hidden away in a LibDatadog-specific file.

(All the "detect if serverless" code needs more refactoring than that, but that's a separate issue.)

tracer/src/Datadog.Trace/LibDatadog/LibDatadogAvailaibilityHelper.cs
```diff
@@ -0,0 +1,64 @@
...
+    // This will never change, so we use a lazy to cache the result.
+    // This confirms that we are in an automatic instrumentation environment (and so P/Invokes have been re-written)
+    // and that the libdatadog library has been deployed (which is not the case in some serverless environments).
+    // We should add or remove conditions from here as our deployment requirements change.
+    private static readonly Lazy<LibDatadogAvailableResult> LibDatadogAvailable = new(() =>
+    {
+        var res = IsServerlessEnvironment();
+        var isServerless = res.Item1;
+        var possibleException = res.Item2;
+        return new(!isServerless && Instrumentation.ProfilerAttached, possibleException);
```
lucaspimentel: `Instrumentation.ProfilerAttached` only check that we loaded `Datadog.Trace.ClrProfiler.Native.dll` or OS equivalent. Should this check that we loaded   `datadog_profiling_ffi.dll` to be safe?

Just checking "is serverless" based on env vars seems dangerous if we get it wrong. What happens if we think `libdatadog` is available but the file is not actually there?

lucaspimentel: (This is not specific to this PR, I've brought it up several times before.)

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/GlobalConfigurationSource.cs
anna-git: so the order of precedence is fleet hands-off config > env > local hands-off config > AppSettings > datadog.json, so it should be correct no? 
image

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/GlobalConfigurationSource.cs
anna-git: yes! fixed in https://github.com/DataDog/dd-trace-dotnet/pull/7282/commits/3911f6a1fab06f020e503d19417e0a5ba401a6ad

tracer/src/Datadog.Trace/Configuration/GlobalSettings.cs
anna-git: good catch! fixed in https://github.com/DataDog/dd-trace-dotnet/pull/7282/commits/3911f6a1fab06f020e503d19417e0a5ba401a6ad

tracer/src/Datadog.Trace/LibDatadog/HandsOffConfiguration/ConfigurationSuccessResult.cs
anna-git: good to know, changed in https://github.com/DataDog/dd-trace-dotnet/pull/7282/commits/3911f6a1fab06f020e503d19417e0a5ba401a6ad

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
anna-git: added in https://github.com/DataDog/dd-trace-dotnet/pull/7282/commits/3911f6a1fab06f020e503d19417e0a5ba401a6ad

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/HandsOffConfigurationSource.cs
anna-git: yes, changed in https://github.com/DataDog/dd-trace-dotnet/pull/7282/commits/3911f6a1fab06f020e503d19417e0a5ba401a6ad

tracer/test/Datadog.Trace.Tests/Datadog.Trace.Tests.csproj
anna-git: good point 😅  fixed in https://github.com/DataDog/dd-trace-dotnet/pull/7282/commits/3911f6a1fab06f020e503d19417e0a5ba401a6ad

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
anna-git: yep! changed in https://github.com/DataDog/dd-trace-dotnet/pull/7282/commits/3911f6a1fab06f020e503d19417e0a5ba401a6ad

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
anna-git: Changed doc in https://github.com/DataDog/dd-trace-dotnet/pull/7282/commits/3911f6a1fab06f020e503d19417e0a5ba401a6ad 
A bit scared to rename `ProfilerAttached` in this PR though 😅

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/GlobalConfigurationSource.cs
anna-git: thanks yes, changed in https://github.com/DataDog/dd-trace-dotnet/pull/7282/commits/3911f6a1fab06f020e503d19417e0a5ba401a6ad

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/GlobalConfigurationSource.cs
anna-git: changed in https://github.com/DataDog/dd-trace-dotnet/pull/7282/commits/3911f6a1fab06f020e503d19417e0a5ba401a6ad

tracer/test/Datadog.Trace.Tests/Configuration/GlobalConfigurationSourceTests.cs
anna-git: fixed in ce361d118c49a2782db5644bcde134212bf3e5cb removed env  vars setting

tracer/test/Datadog.Trace.Tests/Configuration/GlobalConfigurationSourceTests.cs
anna-git: fixed in ce361d118c49a2782db5644bcde134212bf3e5cb

tracer/test/Datadog.Trace.Tests/Configuration/GlobalConfigurationSourceTests.cs
anna-git: changed in ce361d118c49a2782db5644bcde134212bf3e5cb

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
anna-git: good catch fixed in ce361d118c49a2782db5644bcde134212bf3e5cb

tracer/src/Datadog.Trace/LibDatadog/LibDatadogAvailaibilityHelper.cs
anna-git: yes added another `EnvironmentHelpersNoLogging` in ce361d118c49a2782db5644bcde134212bf3e5cb

tracer/src/Datadog.Trace/LibDatadog/LibDatadogAvailaibilityHelper.cs
anna-git: yes maybe we could do the same as `ProfilerAttached` and expose `IsLibdatadogAvailable` as [the test already exists in native](https://github.com/DataDog/dd-trace-dotnet/blob/ce361d118c49a2782db5644bcde134212bf3e5cb/tracer/src/Datadog.Tracer.Native/cor_profiler.cpp?plain=1#L965C1-L965C64) and we could interop expose it  but maybe not in this PR 😅

tracer/src/Datadog.Trace/FrameworkDescription.cs
anna-git: changed in ce361d118c49a2782db5644bcde134212bf3e5cb using now an `EnvironmentHelpersNoLogging` class instead

tracer/src/Datadog.Trace/LibDatadog/CString.cs
anna-git: added in ce361d118c49a2782db5644bcde134212bf3e5cb

tracer/src/Datadog.Trace/LibDatadog/CString.cs
anna-git: good point, I thought it was an error in another class to use the `MessagePack` tool, fixed in ce361d118c49a2782db5644bcde134212bf3e5cb

tracer/src/Datadog.Trace/LibDatadog/HandsOffConfiguration/ConfiguratorHelper.cs
anna-git: very good catch! fixed in 23907fa09b376cb80575e1bf76b2eaadf0afcfa8

tracer/src/Datadog.Tracer.Native/cor_profiler.cpp
anna-git: I didn't I created another class `LibraryConfig` so this is needed... to separate from trace exporter stuff.
But maybe I should rename the trace exporter one which is not really library config? or is it..

tracer/src/Datadog.Trace/LibDatadog/CString.cs
anna-git: This type, like `CharSlice` might have different disposable strategies. If they're delivered by `libdatadog` we don't want them to run their finalizers automatically because they will be handled by libdatadog, and one scenario is:
- libd returns a mother type with various fields
- in the end we drop the whole type with libdatadog like  `ddog_library_config_drop(_: ffi::Vec<LibraryConfig>)` taking care of dropping everything inside
- but if it's us who needed to create CString or CharSlices to supply parameters to libdatadog functions for example, then we should dispose them ourselves with the Dispose() code.. 
so maybe we should leave it a `IDisposable` like `CharSlice` already is and let the caller decides which disposal strategy it wants. If we have a finalizer that tries disposing a CString already disposed by libdatadog, it could crash or be unpredictable.

tracer/src/Datadog.Trace/LibDatadog/LibDatadogAvailaibilityHelper.cs
anna-git: thanks fixed in [d05db22](https://github.com/DataDog/dd-trace-dotnet/pull/7282/commits/d05db229dbabbbd44953491abb8fc991538d554c)

tracer/src/Datadog.Trace/LibDatadog/LibDatadogAvailaibilityHelper.cs
anna-git: yes fixed in [d05db22](https://github.com/DataDog/dd-trace-dotnet/pull/7282/commits/d05db229dbabbbd44953491abb8fc991538d554c)

tracer/src/Datadog.Tracer.Native/dd_profiler_constants.h
anna-git: So after double checking, I confirm this is correct, this maps to all classes in: https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/src/Datadog.Trace/LibDatadog/NativeInterop.cs. The problem is [having the `Config` class](https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/src/Datadog.Trace/LibDatadog/NativeInterop.cs#L53) in the first place which maps to `TraceExporterConfig`, I just checked with libdatadog structure. But the new LibraryConfig is actually LibraryConfig in libdatadog, it's more the `Config` that doesn't seem quite right, I think it should have been all in the `Exporter` class above, as this `Config` for the tracer exporter... but this is beyond this PR context, as LibraryConfig is actually right here..
Maybe we could open another PR renaming the `Config` class later

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/GlobalConfigurationSource.cs
anna-git: yes changed here https://github.com/DataDog/dd-trace-dotnet/blob/anna/stable-configuration-managed/tracer/src/Datadog.Trace/TracerManagerFactory.cs#L112

tracer/src/Datadog.Trace/LibDatadog/HandsOffConfiguration/ConfigurationResult.cs
anna-git: good point renamed it here https://github.com/DataDog/dd-trace-dotnet/blob/e00ae0b2301e0a93402a608ef8b01900f5e2e2a0/tracer/src/Datadog.Trace/LibDatadog/HandsOffConfiguration/ConfigurationResult.cs#L17 and enum is read [here ](https://github.com/DataDog/dd-trace-dotnet/blob/a013c17a6d548446e417b69cea9676b29fd5d7cb/tracer/src/Datadog.Trace/TracerManagerFactory.cs#L115)

tracer/src/Datadog.Trace/LibDatadog/HandsOffConfiguration/ConfiguratorHelper.cs
anna-git: good point all fixed [here ](https://github.com/DataDog/dd-trace-dotnet/blob/a013c17a6d548446e417b69cea9676b29fd5d7cb/tracer/src/Datadog.Trace/LibDatadog/HandsOffConfiguration/ConfiguratorHelper.cs#L97)

tracer/src/Datadog.Trace/LibDatadog/HandsOffConfiguration/ConfiguratorHelper.cs
anna-git: all fixed in [here](https://github.com/DataDog/dd-trace-dotnet/blob/a013c17a6d548446e417b69cea9676b29fd5d7cb/tracer/src/Datadog.Trace/LibDatadog/HandsOffConfiguration/ConfiguratorHelper.cs#L97)

tracer/src/Datadog.Trace/LibDatadog/HandsOffConfiguration/ConfiguratorHelper.cs
anna-git: it's all been changed [here ](https://github.com/DataDog/dd-trace-dotnet/blob/a013c17a6d548446e417b69cea9676b29fd5d7cb/tracer/src/Datadog.Trace/LibDatadog/HandsOffConfiguration/ConfiguratorHelper.cs#L97)

tracer/src/Datadog.Trace/LibDatadog/HandsOffConfiguration/ConfiguratorHelper.cs
anna-git: changed in [here](https://github.com/DataDog/dd-trace-dotnet/blob/a013c17a6d548446e417b69cea9676b29fd5d7cb/tracer/src/Datadog.Trace/LibDatadog/HandsOffConfiguration/ConfiguratorHelper.cs#L97)

tracer/src/Datadog.Trace/LibDatadog/ServiceDiscovery/ServiceDiscoveryHelper.cs
anna-git: good point I [removed it](https://github.com/DataDog/dd-trace-dotnet/blob/a013c17a6d548446e417b69cea9676b29fd5d7cb/tracer/src/Datadog.Trace/LibDatadog/ServiceDiscovery/ServiceDiscoveryHelper.cs#L60) and [logged here](https://github.com/DataDog/dd-trace-dotnet/blob/a013c17a6d548446e417b69cea9676b29fd5d7cb/tracer/src/Datadog.Trace/TracerManagerFactory.cs#L121)

tracer/src/Datadog.Trace/Logging/Internal/DatadogLoggingFactory.cs
anna-git: [Changed it here](https://github.com/DataDog/dd-trace-dotnet/blob/a013c17a6d548446e417b69cea9676b29fd5d7cb/tracer/src/Datadog.Trace/Logging/Internal/DatadogLoggingFactory.cs#L168)

tracer/src/Datadog.Trace/Util/NativeStringHelper.cs
anna-git: fixed in 5b9ecf58156e907fa0ca98120f4ea44bd4dc8573

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/GlobalConfigurationSource.cs
andrewlock: oh... oh no... their whole stack there is just _wrong_ 🤦‍♂️ 

Our stack from highest to lowest _Was_ this:

- Config at runtime
- Programmatic config
- datadog.json
- web.config
- env vars
- default values

So they have the web.config and datadog.json in the wrong place 🙈 And we can't change that, their spec is just wrong.

So the question is - where _should_ they be. Because what they have described cannot work.

I _think_ we need to do something like this:

image

Which is _close_ to what you have, but not quite, as the fleet managed would need to be added last

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/GlobalConfigurationSource.cs
```diff
@@ -18,33 +20,90 @@ namespace Datadog.Trace.Configuration;
...
+    {
+        get
+        {
+            if (_instance != null)
+            {
+                return _instance;
+            }
+
+            CreationResult = CreateDefaultConfigurationSource();
+            _instance = CreationResult.ConfigurationSource;
```
andrewlock: We might need to use `Interlocked.Exchange` on these to ensure they're exposed properly to other threads. Alternatively, use something like this instead (simpler IMO)

```csharp
internal static IConfigurationSource Instance => CreationResult.ConfigurationSource;

internal static GlobalConfigurationSourceResult CreationResult { get; private set; } = CreateDefaultConfigurationSource();


```

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -435,11 +436,21 @@ internal TracerSettings(IConfigurationSource? source, IConfigurationTelemetry te
...
 
-                if (!isLibDatadogAvailable)
+                if (!isLibDatadogAvailable.IsAvailable)
                 {
                     DataPipelineEnabled = false;
-                    Log.Warning(
-                        $"{ConfigurationKeys.TraceDataPipelineEnabled} is enabled, but libdatadog is not available. Disabling data pipeline.");
+                    if (isLibDatadogAvailable.Exception is not null)
+                    {
+                        Log.Error(
```
andrewlock: It was a warning before, I think it should remain a warning.
```suggestion
                        Log.Warning(
```

tracer/src/Datadog.Trace/LibDatadog/CString.cs
andrewlock: > If we have a finalizer that tries disposing a CString already disposed by libdatadog, it could crash or be unpredictable.

~That's what the dispose pattern is for 😄 If we have already called `Dispose()` then the finalizer will be suppressed. That said, if we're confident that we will never miss a dispose here then it's not a problem. My suggestion was mostly because we _did_ miss a dispose 😄~

EDIT: Oh, I see, the difficulty is due to the fact that _sometimes_ we use this type for retrieving _from_ libdatadog, and _sometimes_ we use it to send data _to_ libdatadog 😕 It seems like this could be an ongoing source of issues, i.e. one way you _must_ dispose, the other way you must _not_ dispose.

I would prefer if we could split this into two different types, one for "reading" and one for "writing"; that way we can more easily understand when a usage is correct or not? 🤔 

Happy for that to be in a follow up PR if you agree in principal 🙂

tracer/src/Datadog.Trace/LibDatadog/CString.cs
```diff
@@ -0,0 +1,61 @@
...
+
+[StructLayout(LayoutKind.Sequential)]
+internal struct CString : IDisposable
+{
+    public nint Ptr; // char*
+    public nuint Length; // size of the string, excluding null terminator
+
+    internal CString(string? str)
+    {
+        if (string.IsNullOrEmpty(str))
```
andrewlock: If you use `StringUtil` instead, you don't need the `str!` later (because it has the correct annotations 🙂)

tracer/src/Datadog.Trace/LibDatadog/CString.cs
```diff
@@ -0,0 +1,61 @@
...
+                    {
+                        Marshal.FreeHGlobal(Ptr);
+                        Ptr = IntPtr.Zero;
+                    }
+                }
+            }
+        }
+    }
+
+    public unsafe string ToUtf8String() => StringEncoding.UTF8.GetString((byte*)Ptr, (int)Length);
```
andrewlock, line 50: ~I'm a little confused when we would ever need to call this 🤔 We create a `CString` _from_ a `string` to pass _to_ libdatadog, right?~

EDIT: Oh I see, we use it for both sending to _and_ fetching from 🤔

tracer/src/Datadog.Trace/LibDatadog/HandsOffConfiguration/ConfiguratorHelper.cs
```diff
@@ -0,0 +1,117 @@
...
+            {
+                unsafe
+                {
+                    var ptr = new IntPtr(libraryConfigs.Value.Ptr + (structSize * i));
+                    var libraryConfig = (LibraryConfig*)ptr;
+                    var name = libraryConfig->Name.ToUtf8String();
+                    var value = libraryConfig->Value.ToUtf8String();
+                    if (libraryConfig->Source == LibraryConfigSource.FleetStableConfig)
+                    {
+                        configEntriesRemote.Add(name, value);
```
andrewlock: Do we need to handle the case where the same `name` is returned multiple times? 🤔 It would likely only be due to a bug on the libdatadog side, but using `Add` here would cause it to throw an Exception. If we'd rather _overwrite_ the previous value you could do it safely using
```suggestion
                        configEntriesRemote[name] = value;
```

Maybe throwing an exception as we are currently is the _best_ thing here in practice, just flagging it 🙂

tracer/src/Datadog.Trace/LibDatadog/LibDatadogAvailaibilityHelper.cs
```diff
@@ -0,0 +1,32 @@
...
+using Datadog.Trace.ClrProfiler;
+using Datadog.Trace.Util;
+
+namespace Datadog.Trace.LibDatadog;
+
+/// <summary>
+/// *This class should NOT contain any direct Logger field, nor methods should log*.
+/// LibDatadogAvailable factory is used when building settings for the tracer, if a logger is instantiated in this path, it creates an infinite loop as the logger itself will try to build settings
+/// The Lazy will loop on itself and end up in a InvalidOperationException as Value ends up calling itself
+/// Even calling a method on a class that has a Logger as a static field is enough to have it instantiated, that's why ProfilerAttached eventually call FrameworkDescription that has a Lazy`Logger`
```
andrewlock: nit: I think we need to update this comment?

tracer/src/Datadog.Tracer.Native/cor_profiler.cpp
andrewlock: oooh, I didn't notice that you had moved `StoreTracerMetadata` and that this didn't already exit😄 I think this looks good. WE should probably rename the `Config` type to `ExporterConfig`, but that's completely separate 🙂

tracer/src/Datadog.Tracer.Native/dd_profiler_constants.h
andrewlock: Yep, I agree with all that!

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/GlobalConfigurationSource.cs
andrewlock: so, as discussed, I got this completely wrong. Sorry 😅

tracer/test/Datadog.Trace.Tests/Configuration/GlobalConfigurationSourceTests.cs
```diff
@@ -0,0 +1,55 @@
...
+        sources.Count.Should().Be(3);
+#endif
+        var fleetConfigSource = sources[0].Should().BeOfType<HandsOffConfigurationSource>().Subject;
+        fleetConfigSource.Origin.Should().Be(ConfigurationOrigins.FleetStableConfig);
+        fleetConfigSource.GetString("KEY1", NullConfigurationTelemetry.Instance, null, false).Result.Should().Be("fleet_file_env");
+        fleetConfigSource.GetString("KEY5", NullConfigurationTelemetry.Instance, null, false).Result.Should().Be("fleet_file_service");
+        sources[1].Should().BeOfType<EnvironmentConfigurationSource>();
+        var localConfigSource = sources[2].Should().BeOfType<HandsOffConfigurationSource>().Subject;
+        localConfigSource.Origin.Should().Be(ConfigurationOrigins.LocalStableConfig);
+        // libdatadog already applies precedence so doesnt get these keys already defined by fleet, but we want to account for this in later stages where it should report everything for telemetry
```
andrewlock, line 49: Just to clarify, this is a libdatadog limitation, which we expect to be updated in the future, and we should update this test when it is?

andrewlock (approved):
LGTM, just a few suggestions, nice work!

tracer/test/Datadog.Trace.Tests/Configuration/GlobalConfigurationSourceTests.cs
anna-git: Yes it is a limitation as for telemetry we want to report all sources later on, so it will have to be adapted to report everything and not filter out

tracer/src/Datadog.Trace/LibDatadog/CString.cs
anna-git: That's a good idea, we actually had that the issue with the waf too if I remember correctly having types used in both directions with different ownership🤔

tracer/src/Datadog.Trace/LibDatadog/LibDatadogAvailaibilityHelper.cs
anna-git: Something maybe like https://github.com/DataDog/dd-trace-dotnet/pull/7380/

tracer/src/Datadog.Trace/LibDatadog/CString.cs
lucaspimentel: > This will add a BOM 

I think this only happens when writing to streams, not when using `GetBytes()` or `GetString()`.

> I thought it was an error in another class to use the `MessagePack` tool

Yeah, it feels a bit dirty to me too to depend on `MessagePack`, but it's just creating a new `UTF8Encoding` with the BOM disabled:
https://github.com/DataDog/dd-trace-dotnet/blob/fec482205d2f1ab7fa766134923c1d3f4c2347e3/tracer/src/Datadog.Trace/Vendors/MessagePack/StringEncoding.cs#L10-L13

(The one from `Encoding.UTF8` has the BOM enabled.)

We could just copy that `StringEncoding.UTF8` into our own code to remove the dependency on `MessagePack`, but... meh.

tracer/src/Datadog.Trace/Util/NativeStringHelper.cs
lucaspimentel: > we need to use the BOM-free version for libdatadog

fwiw, `Encoding.UTF8.GetString()` doesn't add the BOM. That only applies when writing to streams or similar.

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
lucaspimentel: > A bit scared to rename `ProfilerAttached` in this PR though

yeah, don't do that, hah

tracer/src/Datadog.Trace/LibDatadog/LibDatadogAvailaibilityHelper.cs
lucaspimentel: Thanks! I'll try to do more cleanup in all the code that detects "serverless" soon.

tracer/src/Datadog.Trace/LibDatadog/LibDatadogAvailaibilityHelper.cs
lucaspimentel: > Something maybe like #7380

LGTM, thanks!

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
```diff
@@ -50,27 +50,6 @@ public static class Instrumentation
 
         private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(Instrumentation));
 
-        /// <summary>
-        /// Gets a value indicating whether Datadog's profiler is attached to the current process.
-        /// </summary>
-        /// <value>
-        ///   <c>true</c> if the profiler is currently attached; <c>false</c> otherwise.
-        /// </value>
-        public static bool ProfilerAttached
```
lucaspimentel, line 59: Any reason to remove this? It's fine if it's not used anywhere and all test pass. I'm just curious since you said were you were afraid of renaming it 😅

tracer/src/Datadog.Trace/LibDatadog/HandsOffConfiguration/ConfiguratorHelper.cs
```diff
@@ -0,0 +1,117 @@
...
+
+namespace Datadog.Trace.LibDatadog.HandsOffConfiguration;
+
+internal struct ConfiguratorHelper
+{
+    internal static ConfigurationResult GetConfiguration(bool debugEnabled, string? handsOffLocalConfigPath, string? handsOffFleetConfigPath, bool? isLibdatadogAvailable = null)
+    {
+        if (isLibdatadogAvailable is false)
+        {
+            return new ConfigurationResult(null, "Skipping hands-off configuration: as LibDatadog is not available", Result.LibDatadogUnavailable);
```
lucaspimentel:
```suggestion
            return new ConfigurationResult(null, "Skipping hands-off configuration because LibDatadog is not available", Result.LibDatadogUnavailable);
```

tracer/src/Datadog.Trace/LibDatadog/HandsOffConfiguration/ConfiguratorHelper.cs
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        if (isLibdatadogAvailable is null)
+        {
+            var isLibdatadogAvailableEval = LibDatadogAvailaibilityHelper.IsLibDatadogAvailable;
+            if (!isLibdatadogAvailableEval.IsAvailable)
+            {
+                return new ConfigurationResult(null, "Skipping hands-off configuration: as LibDatadog is not available", Result.LibDatadogUnavailable, isLibdatadogAvailableEval.Exception);
```
lucaspimentel, line 28:
```suggestion
                return new ConfigurationResult(null, "Skipping hands-off configuration because LibDatadog is not available", Result.LibDatadogUnavailable, isLibdatadogAvailableEval.Exception);
```

tracer/src/Datadog.Trace/LibDatadog/LibDatadogAvailableResult.cs
```diff
@@ -0,0 +1,16 @@
...
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2 License.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
+// </copyright>
+
+#nullable enable
+using System;
+
+namespace Datadog.Trace.LibDatadog;
+
+internal struct LibDatadogAvailableResult(bool isAvailable, Exception? exception = null)
```
lucaspimentel: I think this can be `readonly`? (it's good practice to use `readonly struct` whenever possible)

```suggestion
internal readonly struct LibDatadogAvailableResult(bool isAvailable, Exception? exception = null)
```

tracer/src/Datadog.Trace/TracerManagerFactory.cs
```diff
@@ -107,6 +109,17 @@ internal TracerManager CreateTracerManager(
...
+            var result = GlobalConfigurationSource.CreationResult;
+            if (result.Result is not Result.Success)
+            {
+                Log.Warning(result.Exception, "Failed to create the global configuration source with status: {Status} and error message: {ErrorMessage}", result.Result.ToString(), result.ErrorMessage);
+            }
+
+            var libdatadogAvailaibility = LibDatadogAvailaibilityHelper.IsLibDatadogAvailable;
+            if (libdatadogAvailaibility.Exception is not null)
+            {
+                Log.Warning(libdatadogAvailaibility.Exception, "An exception happened when trying to see if libdatadog was available");
```
lucaspimentel:
```suggestion
                Log.Warning(libdatadogAvailaibility.Exception, "An exception occurred while checking if libdatadog is available");
```

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
anna-git: Oh yes I got mixed up, I was scared of renaming `NativeMethods.IsProfilerAttached` but this one I checked and it doesnt seem to be used anywhere, maybe a relicate from before

tracer/src/Datadog.Trace/LibDatadog/LibDatadogAvailableResult.cs
anna-git: fixed in 841472276dece6591de6f6cbbc5228a71aa48dc2

tracer/src/Datadog.Trace/LibDatadog/HandsOffConfiguration/ConfiguratorHelper.cs
anna-git: fixed in 841472276dece6591de6f6cbbc5228a71aa48dc2

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/GlobalConfigurationSource.cs
anna-git: good point changed in e6417baf8094782bd245ca9ab76d3088c1c38484

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
anna-git: my bad, this happens when moving things around 😅  changed in e6417baf8094782bd245ca9ab76d3088c1c38484

tracer/src/Datadog.Trace/LibDatadog/CString.cs
anna-git: fixed in e6417baf8094782bd245ca9ab76d3088c1c38484

tracer/src/Datadog.Trace/LibDatadog/LibDatadogAvailaibilityHelper.cs
anna-git: changed in e6417baf8094782bd245ca9ab76d3088c1c38484

tracer/src/Datadog.Trace/LibDatadog/HandsOffConfiguration/ConfiguratorHelper.cs
anna-git: good point! I committed your suggestions of `configEntriesRemote[name] = value;` . This indeed shouldn't happen but as we can't really log anything in this code path, and I didn't want to add more complexity to the result type 😅 For now let's not make it crash 😅

PR #7284, opened 2025-07-25
Dispose the inferred proxy span last

bouwkast:
## Summary of changes

Swaps the ordering when disposing the API Gateway inferred span to be after the "root" span.
Realistically the API Gateway span will be the parent of the root span for the ASP NET Core instrumentation.

## Reason for change

Strangely only came up with xUnit 2.9.3 upgrade and goes away when downgrading 😕 

But I think this is technically more correct.

## Implementation details

Reversed the ordering.

## Test coverage

Updated snapshots.

## Other details

I still don't fully understand _why_ the xUnit upgrade detected / caused this?

### DIFF
--- a/tracer/test/snapshots/AspNetCoreMvc31InferredProxySpansTests.Enabled.__path=_status-code_203_statusCode=203_spanCount=3.verified.txt
+++ b/tracer/test/snapshots/AspNetCoreMvc31InferredProxySpansTests.Enabled.__path=_status-code_203_statusCode=203_spanCount=3.verified.txt
@@ -20,33 +20,6 @@
   },
   {
     TraceId: Id_1,
-    SpanId: Id_4,
-    Name: aws.apigateway,
-    Resource: GET /api/test/?,
-    Service: test.api.com,
-    Type: web,
-    ParentId: Id_5,
-    Tags: {
-      component: aws-apigateway,
-      env: integration_tests,
-      http.method: GET,
-      http.response.headers.sample_correlation_identifier: 0000-0000-0000,
-      http.response.headers.server: Kestrel,
-      http.route: /api/test/?,
-      http.status_code: 203,
-      http.url: test.api.com/api/test/1,
-      language: dotnet,
-      span.kind: internal,
-      stage: prod,
-      _dd.base_service: Samples.AspNetCoreMvc31
-    },
-    Metrics: {
-      _dd.inferred_span: 1.0,
-      _sampling_priority_v1: 1.0
-    }
-  },
-  {
-    TraceId: Id_1,
     SpanId: Id_3,
     Name: aspnet_core.request,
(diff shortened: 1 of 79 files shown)


tracer/test/snapshots/AspNetCoreMvc31InferredProxySpansTests.Enabled.__path=_status-code_203_statusCode=203_spanCount=3.verified.txt
```diff
@@ -18,33 +18,6 @@
...
     }
   },
-  {
-    TraceId: Id_1,
-    SpanId: Id_4,
-    Name: aws.apigateway,
-    Resource: GET /api/test/?,
-    Service: test.api.com,
-    Type: web,
-    ParentId: Id_5,
```
zacharycmontoya, line 28: It's interesting to note that BEFORE this change, the API Gateway span had a parent span (according to the snapshots). Now, it doesn't. Is the new behavior correct?

zacharycmontoya (approved):
LGTM, looking for confirmation on one specific change in the snapshots

tracer/test/snapshots/AspNetCoreMvc31InferredProxySpansTests.Enabled.__path=_status-code_203_statusCode=203_spanCount=3.verified.txt
andrewlock: Yeah, this looks very strange 🤔 AFAICT, there _shouldn't_ be a parent ID on the incoming span, because the request is being sent from the test itself... Which suggests that it was incorrect previously and is now correct, but then, what was it doing before? 🙈

andrewlock (approved):
LGTM _if_ the change in parenting for the API gateway span is actually correct :/

tracer/test/snapshots/AspNetCoreMvc31InferredProxySpansTests.Enabled.__path=_status-code_203_statusCode=203_spanCount=3.verified.txt
bouwkast: Correct the new behavior is the expected correct behavior, there shouldn't be a parent ID on the span it should be the parent, I honestly couldn't figure out what was going on but should probably determine what was causing it.

The Parent ID from what I could tell was the _previous_ API Gateway span from previous test runs 😕

bouwkast:
Verified that the cause of this was the proxy getting closed first when it wasn't the active scope which meant that the close did nothing so it got left open 😬

bouwkast:
Closing in favor of https://github.com/DataDog/dd-trace-dotnet/pull/7293

PR #7287, opened 2025-07-28
[Profiler] Support Stable Config

chrisnas:
## Summary of changes
The managed layer of the tracer should be able to set basic configuration to the profiling including per runtimeID details

## Reason for change
Support Stable Configuration 

## Implementation details
- add p/invoke method
- update profiler configuration management and SSI 

## Test coverage
- update existing ones and add new needed scenarios
-

### DIFF
--- a/.github/workflows/verify_app_trimming_changes_are_persisted.yml
+++ b/.github/workflows/verify_app_trimming_changes_are_persisted.yml
@@ -26,7 +26,7 @@ jobs:
         run: Get-ChildItem –Path ".\tracer\src\Datadog.Trace.Trimming\build\Datadog.Trace.Trimming.xml" -Recurse -File | Remove-Item
 
       - name: "Regenerating Datadog.Trace.Trimming.xml"
-        run: .\tracer\build.ps1 BuildTracerHome CreateTrimmingFile
+        run: .\tracer\build.ps1 CreateRequiredDirectories Restore CompileManagedLoader CompileManagedSrc CreateTrimmingFile
 
       - name: "Verify no changes in Datadog.Trace.Trimming.xml"
         run: |


profiler/src/ProfilerEngine/Datadog.Profiler.Native/PInvoke.h
```diff
@@ -8,6 +8,28 @@
...
 
+
+enum ProfilingEnabled
+{
+    ProfilingDisabled = 0,
+    ProfilingEnabledTrue = 1,
+    ProfilingAuto = 2
+};
+
+struct SharedConfig
```
anna-git: Could this struct maybe moved to the Shared code in `\shared\src\native-src ?` so that it's reusable by all products?

chrisnas: ✅ Good idea: moved to `\shared\src\native-src\stable_config.h`

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
```diff
@@ -91,6 +92,39 @@ public static void Initialize()
...
+                            IastEnabled = Iast.Iast.Instance.Settings.Enabled,
+                            RaspEnabled = Security.Instance.Settings.RaspEnabled,
+                            DynamicInstrumentationEnabled = false,  // TODO: find where to get this value from but for the other native p/invoke call
+                            RuntimeId = RuntimeId.Get(),
+                            Environment = tracerSettings.Environment,
+                            ServiceName = TraceUtil.NormalizeTag(tracerSettings.ServiceName ?? string.Empty),
+                            Version = tracerSettings.ServiceVersion
+                        };
+
+                        // It is possible that the profiler binary is not "there" such as Azure Function and some CI tests
```
anna-git: fwiw, there is a[ function here](https://github.com/DataDog/dd-trace-dotnet/blob/303c4f9d3dc4e4ac35ccd1094a7f7e254663ebce/tracer/src/Datadog.Trace/Util/EnvironmentHelpersNoLogging.cs#L21) telling you if it's serverless , maybe it could be a guard

tracer/src/Datadog.Trace.Trimming/build/Datadog.Trace.Trimming.xml
```diff
@@ -227,6 +227,7 @@
    <assembly fullname="System.Data.SqlClient" />
    <assembly fullname="System.Data.SQLite" />
    <assembly fullname="System.Diagnostics.Debug">
+      <type fullname="System.Diagnostics.Debug" />
```
andrewlock: I think these snuck in by default
```suggestion
```

tracer/src/Datadog.Trace.Trimming/build/Datadog.Trace.Trimming.xml
```diff
@@ -562,6 +563,8 @@
       <type fullname="System.Diagnostics.CodeAnalysis.RequiresUnreferencedCodeAttribute" />
       <type fullname="System.Diagnostics.CodeAnalysis.UnconditionalSuppressMessageAttribute" />
       <type fullname="System.Diagnostics.ConditionalAttribute" />
+      <type fullname="System.Diagnostics.Debug" />
+      <type fullname="System.Diagnostics.Debug/AssertInterpolatedStringHandler" />
```
andrewlock:
```suggestion
```

profiler/src/ProfilerEngine/Datadog.Profiler.Native/RuntimeIdStore.cpp
```diff
@@ -86,6 +86,7 @@ const char* RuntimeIdStore::GetName()
 
 const char* RuntimeIdStore::GetId(AppDomainID appDomainId)
 {
+    // TODO: why not using the cache?
```
gleocadie, line 89: because the cache is already in the native loader
This handles the case where the profiler did not find the native loader

shared/src/native-src/stable_config.h
```diff
@@ -0,0 +1,31 @@
+#ifndef DD_CLR_PROFILER_STABLE_CONFIG_H_
```
gleocadie, line 1: add the license header

profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/SystemCallsShield.h
```diff
@@ -22,6 +22,7 @@ class SystemCallsShield : public IService
 
     bool Start() override;
     bool Stop() override;
+    bool IsStarted() override;
```
gleocadie, line 25: why overriding it ?
why not rely on the one in ServiceBase ?

profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/SystemCallsShield.cpp
```diff
@@ -34,6 +35,12 @@ bool SystemCallsShield::ShouldEnable(IConfiguration* configuration)
 
 bool SystemCallsShield::Start()
 {
+    if (_isStarted)
+    {
+        return true;
+    }
+    _isStarted = true;
+
```
gleocadie, line 43: why overriding it ?
why not rely on the one in ServiceBase ?

profiler/test/Datadog.Profiler.IntegrationTests/SingleStepInstrumentation/SingleStepInstrumentationTest.cs
```diff
@@ -51,7 +51,8 @@ public void ManualAndProfilingEnvVarNotSet(string appName, string framework, str
         [TestAppFact("Samples.Computer01")]
         public void ManualAndProfilingEnvVarTrue(string appName, string framework, string appAssembly)
         {
-            var runner = new TestApplicationRunner(appName, framework, appAssembly, _output, commandLine: "--scenario 1", enableProfiler: true);
+            var runner = new TestApplicationRunner(appName, framework, appAssembly, _output, commandLine: "--scenario 1", enableProfiler: false);
+            runner.Environment.SetVariable(EnvironmentVariables.ProfilerEnabled, "true");
```
gleocadie: set enableProfiler when the env var is true

gleocadie (approved):
LGTM

profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/SystemCallsShield.cpp
chrisnas: > why overriding it ? why not rely on the one in ServiceBase ?

The class does not inherit from ServiceBase and I did not want to change it in this PR

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
```diff
@@ -60,6 +61,51 @@ public static class Instrumentation
...
+
+            Log.Debug("Setting Stable Configuration in Continuous Profiler native library.");
+            var tracerSettings = Tracer.Instance.Settings;
+            var profilerSettings = Profiler.Instance.Settings;
+
+            NativeInterop.SharedConfig config = new NativeInterop.SharedConfig
+            {
+                ProfilingEnabled = (profilerSettings.ProfilerState == ProfilerState.Auto) ? NativeInterop.ProfilingEnabled.Auto :
+                                   (profilerSettings.ProfilerState == ProfilerState.Enabled) ? NativeInterop.ProfilingEnabled.Enabled :
+                                   NativeInterop.ProfilingEnabled.Disabled,
```
andrewlock: nit (it looks better in the non-GitHub view, I promise!)
```suggestion
                ProfilingEnabled = profilerSettings.ProfilerState switch
                                   {
                                       ProfilerState.Auto => NativeInterop.ProfilingEnabled.Auto,
                                       ProfilerState.Enabled => NativeInterop.ProfilingEnabled.Enabled,
                                       _ => NativeInterop.ProfilingEnabled.Disabled
                                   },
```

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
```diff
@@ -60,6 +61,51 @@ public static class Instrumentation
...
+                TracingEnabled = tracerSettings.TraceEnabled,
+                IastEnabled = Iast.Iast.Instance.Settings.Enabled,
+                RaspEnabled = Security.Instance.Settings.RaspEnabled,
+                DynamicInstrumentationEnabled = false,  // TODO: find where to get this value from but for the other native p/invoke call
+                RuntimeId = RuntimeId.Get(),
+                Environment = tracerSettings.Environment,
+                ServiceName = TraceUtil.NormalizeTag(tracerSettings.ServiceName ?? string.Empty),
```
andrewlock: TODO: We should _probably_ move this normalization to the settings instead of doing this everywhere (I've seen it elsewhere too)

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
```diff
@@ -60,6 +61,51 @@ public static class Instrumentation
...
+                Version = tracerSettings.ServiceVersion
+            };
+
+            // It is possible that the profiler binary is not "there" such as Azure Function and some CI tests
```
andrewlock: I would suggest we explicitly check for scenarios that we _know_ the profiler isn't available and don't try to do this. We can likely enumerate all of those and encapsulate it in a method, similar to the way we do with serverless🤔 

My main concern is that I would like this to be an `Error()` in cases where we _Expect_ it to work, so that we have visibility of issues

andrewlock (approved):
LGTM from the tracer side apart from some minor suggestions

tracer/src/Datadog.Trace.Trimming/build/Datadog.Trace.Trimming.xml
chrisnas: ✅ done

tracer/src/Datadog.Trace.Trimming/build/Datadog.Trace.Trimming.xml
chrisnas: ✅ done

shared/src/native-src/stable_config.h
chrisnas: I did not add the header because the other files in that folder don't have one

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
chrisnas: ✅ done

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
chrisnas: I've seen it in `TracerManagerFactory` but I don't know if these two are connected.
However, the constructor of `TracerSettings` is using `NormalizerTraceProcessor.NormalizeService` and does not check against empty names. Not sure what to do...

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
chrisnas: Sure: where can I find the code that detects all these "known" scenarios?
I'm already checking the ARM case at the top of the method

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
andrewlock: The TODO was more for me, for later 😄 I've seen other places explicitly normalizing the tag and it shouldn't be necessary to do it more than once...

Although, now you mention it, I think this actually _isn't_ passing the correct value, it should be passing `Tracer.DefaultServiceName`

I think ideally, you should make this change above:

```diff
-var tracerSettings = Tracer.Instance.Settings;
+var tracer = Tracer.Instance;
+var tracerSettings = tracer.Settings;
```

and then update the code here:

```suggestion
                ServiceName = tracer.DefaultServiceName,
```

which is what we're using today: 🙂 
https://github.com/DataDog/dd-trace-dotnet/blob/f0e9905238e7fe6c65a78a2a92dbd691fb36613e/tracer/src/Datadog.Trace/TracerManagerFactory.cs#L78

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
andrewlock: > Sure: where can I find the code that detects all these "known" scenarios?

It doesn't not exist yet 😅 I think we'll need to discuss with @lucaspimentel and @tonyredondo as those are the two main scenarios that can be impacted, but I think we can do that _post_ this PR tbh.

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
```diff
@@ -60,6 +61,62 @@ public static class Instrumentation
...
+                    _ => NativeInterop.ProfilingEnabled.Disabled
+                },
+
+                TracingEnabled = tracerSettings.TraceEnabled,
+                IastEnabled = Iast.Iast.Instance.Settings.Enabled,
+                RaspEnabled = Security.Instance.Settings.RaspEnabled,
+                DynamicInstrumentationEnabled = false,  // TODO: find where to get this value from but for the other native p/invoke call
+                RuntimeId = RuntimeId.Get(),
+                Environment = tracerSettings.Environment,
+                ServiceName = TraceUtil.NormalizeTag(tracerSettings.ServiceName ?? string.Empty),
```
andrewlock: This is using the wrong value, you need to use `tracer.DefaultServiceName`:
```suggestion
                ServiceName = TraceUtil.NormalizeTag(tracer.DefaultServiceName),
```

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
chrisnas: ✅ done

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
```diff
@@ -60,6 +61,63 @@ public static class Instrumentation
...
+                Log.Information("Profiling is not available on ARM.");
+                return;
+            }
+
+            // it is also possible that the profiler binary is not "there" such as Azure Function and some CI tests
+            if (string.IsNullOrEmpty(Environment.GetEnvironmentVariable("DD_INTERNAL_PROFILING_NATIVE_ENGINE_PATH")))
+            {
+                Log.Information("Profiling native library path is not set, skipping setting profiler configuration.");
+                return;
+            }
```
andrewlock:
```suggestion
            // The profiler is not available in various environments, so don't try to P/Invoke in those cases
            // as the binary won't be there
            if (!ProfilerAvailabilityHelper.IsContinuousProfilerAvailable)
            {
                Log.Information("The Continuous Profiler is not available");
                return;
            }
```

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
andrewlock: We now handle these scenarios above, so this is just a sensible "don't crash" backstop

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
```diff
@@ -60,6 +61,63 @@ public static class Instrumentation
...
+                RaspEnabled = Security.Instance.Settings.RaspEnabled,
+                DynamicInstrumentationEnabled = false,  // TODO: find where to get this value from but for the other native p/invoke call
+                RuntimeId = RuntimeId.Get(),
+                Environment = tracerSettings.Environment,
+                ServiceName = tracer.DefaultServiceName,
+                Version = tracerSettings.ServiceVersion
+            };
+
+            // It is possible that the profiler binary is not "there" such as Azure Function and some CI tests
+            // So, ensure that no exception bubbles up
```
andrewlock: We handle those cases in general now, but still want to play it safe

```suggestion
            // Make sure nothing bubbles up, even if there are issues
```

PR #7290, opened 2025-07-28
Filter connection header in flaky tests

NachoEchevarria:
## Summary of changes

Some security WebApi tests that rely on the Sample.security.WebAPI (.NET framework) does not show consistent result. The result of the test depends on the order of the tests. The problem is that in some requests, the Connection header is returned (or not) by the framework depending on the request order. That causes the generated security header fingerprint to vary.

The fingerprint hash and format is defined [here](https://docs.google.com/document/d/1DivOa9XsCggmZVzMI57vyxH2_EBJ0-qqIkRHm_sEvSs/edit?tab=t.0).

In order to avoid that, a scrubber has been added to skip that particular flag in the fingerprint.

Other approaches have been tried to get a consistent result that failed:
* Adding a filter in FilterConfig.cs in the sample to overwrite the connection header.
* Set UseProxy = false in HttpClientHandler in the test
* Moved to an HttpModule and added Connection: keep-alive in EndRequest

In all these cases, the framework seemed to overwrite the connection header value even when set explicitly.

## Reason for change

Avoid having tests that depend on test execution order.

### DIFF
--- a/tracer/test/Datadog.Trace.Security.IntegrationTests/AspNetWebApi.cs
+++ b/tracer/test/Datadog.Trace.Security.IntegrationTests/AspNetWebApi.cs
@@ -8,7 +8,6 @@ using System;
 using System.Collections.Immutable;
 using System.Linq;
 using System.Net;
-using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Datadog.Trace.AppSec;
 using Datadog.Trace.TestHelpers;


tracer/test/Datadog.Trace.Security.IntegrationTests/AspNetBase.cs
```diff
@@ -256,6 +256,14 @@ await VerifyHelper.VerifySpans(spans, settings)
             }
         }
 
+        protected static void FilterConnectionHeader(VerifySettings settings)
+        {
+            Regex appSecConnectionHeader0 = new(@"_dd.appsec.fp.http.header: hdr-0\d", RegexOptions.IgnoreCase | RegexOptions.Compiled);
```
andrewlock, line 261: So just to clarify, this regex simply sets the _second_ digit of the regex to `X`, correct? That's the only expected change?

andrewlock (approved):
Thanks!

tracer/test/Datadog.Trace.Security.IntegrationTests/AspNetBase.cs
NachoEchevarria: Yes, it only sets the second digit. Later, it is replaced by "_dd.appsec.fp.http.header: hdr-0X"

PR #7293, opened 2025-07-28
Swap disposal ordering of inferred span

bouwkast:
## Summary of changes

Swaps the ordering when disposing the API Gateway inferred span to be after the "root" span.
Realistically the API Gateway span will be the parent of the root span for the ASP NET Core instrumentation.

## Reason for change

If the proxy scope is the parent of the `ASPNET Core` scope AND that `ASPNET Core` scope is active this was hitting the following [piece of code](https://github.com/DataDog/dd-trace-dotnet/blob/1752abdc32c02c702b7c8374bfd1bae8b381d7db/tracer/src/Datadog.Trace/AsyncLocalScopeManager.cs#L44-L48) in `Async

```csharp
            if (current == null || current != scope)
            {
                // This is not the current scope for this context, bail out
                return;
            }
```

This would cause the scope for the proxy to remain open 😢 

## Implementation details

Swapped the ordering of them.

## Test coverage

- Identified and covered by existing tests and was caught in [this](https://github.com/DataDog/dd-trace-dotnet/pull/7280) PR when updating xUnit

## Other details

Cherry picked out and making it's own PR instead of bundling into the [xUnit update one ](https://github.com/DataDog/dd-trace-dotnet/pull/7284) because it passes and works without that.

### DIFF
--- a/tracer/test/snapshots/AspNetCoreMvc31InferredProxySpansTests.Enabled.__path=__statusCode=200_spanCount=3.verified.txt
+++ b/tracer/test/snapshots/AspNetCoreMvc31InferredProxySpansTests.Enabled.__path=__statusCode=200_spanCount=3.verified.txt
@@ -20,33 +20,6 @@
   },
   {
     TraceId: Id_1,
-    SpanId: Id_4,
-    Name: aws.apigateway,
-    Resource: GET /api/test/?,
-    Service: test.api.com,
-    Type: web,
-    ParentId: Id_5,
-    Tags: {
-      component: aws-apigateway,
-      env: integration_tests,
-      http.method: GET,
-      http.response.headers.sample_correlation_identifier: 0000-0000-0000,
-      http.response.headers.server: Kestrel,
-      http.route: /api/test/?,
-      http.status_code: 200,
-      http.url: test.api.com/api/test/1,
-      language: dotnet,
-      span.kind: internal,
-      stage: prod,
-      _dd.base_service: Samples.AspNetCoreMvc31
-    },
-    Metrics: {
-      _dd.inferred_span: 1.0,
-      _sampling_priority_v1: 1.0
-    }
-  },
-  {
-    TraceId: Id_1,
     SpanId: Id_3,
     Name: aspnet_core.request,
     Resource: GET /home/index,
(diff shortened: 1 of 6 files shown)


tracer/test/snapshots/AspNetCoreMvc31InferredProxySpansTests.Enabled.__path=__statusCode=200_spanCount=3.verified.txt
```diff
@@ -18,33 +18,6 @@
...
     }
   },
-  {
-    TraceId: Id_1,
-    SpanId: Id_4,
-    Name: aws.apigateway,
-    Resource: GET /api/test/?,
-    Service: test.api.com,
-    Type: web,
-    ParentId: Id_5,
```
bouwkast, line 28: This ParentID should NOT be set, there shouldn't be a parent for this span, and this parent was the proxy scope of the previous test as it wasn't closed.

tracer/test/snapshots/AspNetCoreMvc31InferredProxySpansTests.Enabled.__path=_bad-request_statusCode=500_spanCount=3.verified.txt
```diff
@@ -20,71 +20,74 @@
...
     }
   },
   {
     TraceId: Id_1,
-    SpanId: Id_3,
-    Name: aspnet_core.request,
-    Resource: GET /bad-request,
-    Service: Samples.AspNetCoreMvc31,
+    SpanId: Id_4,
+    Name: aws.apigateway,
```
bouwkast, line 61: Note the correct lack of Parent ID

tracer/test/snapshots/AspNetCoreMvc31InferredProxySpansTests.Enabled.__path=_handled-exception_statusCode=500_spanCount=3.verified.txt
```diff
@@ -24,35 +24,6 @@ System.Exception: Exception of type 'System.Exception' was thrown.
...
     }
   },
-  {
-    TraceId: Id_1,
-    SpanId: Id_4,
-    Name: aws.apigateway,
-    Resource: GET /api/test/?,
-    Service: test.api.com,
-    Type: web,
-    ParentId: Id_5,
```
bouwkast, line 34: This ParentID should NOT be set, there shouldn't be a parent for this span, and this parent was the proxy scope of the previous test as it wasn't closed.

tracer/test/snapshots/AspNetCoreMvc31InferredProxySpansTests.Enabled.__path=_not-found_statusCode=404_spanCount=2.verified.txt
```diff
@@ -2,58 +2,58 @@
...
   },
   {
     TraceId: Id_1,
     SpanId: Id_3,
-    Name: aspnet_core.request,
-    Resource: GET /not-found,
-    Service: Samples.AspNetCoreMvc31,
+    Name: aws.apigateway,
+    Resource: GET /api/test/?,
+    Service: test.api.com,
```
bouwkast, line 35: Note the correct lack of Parent ID

tracer/test/snapshots/AspNetCoreMvc31InferredProxySpansTests.Enabled.__path=_status-code_203_statusCode=203_spanCount=3.verified.txt
```diff
@@ -18,33 +18,6 @@
...
     }
   },
-  {
-    TraceId: Id_1,
-    SpanId: Id_4,
-    Name: aws.apigateway,
-    Resource: GET /api/test/?,
-    Service: test.api.com,
-    Type: web,
-    ParentId: Id_5,
```
bouwkast, line 28: This ParentID should NOT be set, there shouldn't be a parent for this span, and this parent was the proxy scope of the previous test as it wasn't closed.

andrewlock (approved):
Great, thanks! I'll rebase the xunit PR on top of this once it's merged 🙂

PR #7294, opened 2025-07-29
[Profiler] Fix and bump timeitsharp

tonyredondo:
## Summary of changes

This PR fixes the currently broken execution benchmark for the profiler product.

## Reason for change

It's broken


profiler/build/OutgoingHttpRequests.windows.json
```diff
@@ -56,5 +58,6 @@
     "benchmark.job.runtime.name": ".NET 9.0",
     "benchmark.job.runtime.moniker": "net9.0"
   },
-  "metricsFilePath": "*metrics.json"
+  "metricsFilePath": "*metrics.json",
+  "overheadThreshold": 100
```
andrewlock, line 62: Are these thresholds ok - asking because they weren't for .NET FX before 😉

profiler/build/run-outgoinghttprequests.cmd
andrewlock, line 1: all these cmd files need to have the stupid error handling, to make sure they fail if the benchmarks actually fail. I would recommend putting this into a single file and calling it.

For example, create this file `run_timeit.cmd`:

```cmd
@echo off
setlocal enabledelayedexpansion

set FAILED=0

echo *********************
echo Run x64
echo *********************
dotnet timeit %1 --variable arch=x64

IF ERRORLEVEL 1 set FAILED=1

echo *********************
echo Run x86
echo *********************
dotnet timeit %1 --variable arch=x64

IF ERRORLEVEL 1 set FAILED=1

IF "!FAILED!"=="1" (
    echo One or more benchmarks failed.
    exit /b 1
)
echo All benchmarks completed successfully.
exit /b 0
```

And then change this file to be 

```cmd
@echo off

call install_timeit.cmd

call run_timeit.cmd  OutgoingHttpRequests.windows.json

exit /b %ERRORLEVEL%
```

andrewlock (approved):
I think we should improve the cmd files to make sure we return as expected:

profiler/build/run-outgoinghttprequests.cmd
tonyredondo: done in [b45b1b7](https://github.com/DataDog/dd-trace-dotnet/pull/7294/commits/b45b1b7b10539c91f2a864ce6cbe6157a44ef3e4)

PR #7296, opened 2025-07-29
Ensure file_opened_ is getting updated when rotating

tonyredondo:
## Summary of changes

This PR just make sure that the `file_opened_` flag gets updated properly.

### DIFF
--- a/shared/src/native-src/lazy_rotating_file_sink-inl.h
+++ b/shared/src/native-src/lazy_rotating_file_sink-inl.h
@@ -138,6 +138,7 @@ SPDLOG_INLINE void lazy_rotating_file_sink<Mutex>::rotate_()
             {
                 file_helper_.reopen(true); // truncate the log file anyway to prevent it to grow beyond its limit!
                 current_size_ = 0;
+                file_opened_ = true;
                 throw_spdlog_ex("lazy_rotating_file_sink: failed renaming " + filename_to_str(src) + " to " + filename_to_str(target), errno);
             }
         }
--- a/shared/src/native-src/lazy_rotating_file_sink-inl.h
+++ b/shared/src/native-src/lazy_rotating_file_sink-inl.h
@@ -93,8 +93,12 @@ SPDLOG_INLINE void lazy_rotating_file_sink<Mutex>::sink_it_(const details::log_m
             new_size = formatted.size();
         }
     }
-    file_helper_.write(formatted);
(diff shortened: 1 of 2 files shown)


shared/src/native-src/lazy_rotating_file_sink-inl.h
```diff
@@ -69,12 +69,12 @@ SPDLOG_INLINE filename_t lazy_rotating_file_sink<Mutex>::filename()
...
-        // I _think_ this is called under a mutex, so we don't need to lock here.
-        file_opened_ = true;
-        file_helper_.open(calc_filename(base_filename_, 0));
-        current_size_ = file_helper_.size(); // expensive. called only once
+    try {
+        // let's open the file if is needed
+        open_if_needed_();
+    } catch (const spdlog_ex&) {
+        file_opened_ = false;
+        return;
```
andrewlock, line 77: Exceptions thrown by sink_it_ are handled automatically. I was wondering if we should rethrow here, but as far as I can tell, the spdlog catch (`SPDLOG_TRY` macro`) is always _inside_ the for loop, so doesn't actually make any difference I think 🤔

shared/src/native-src/lazy_rotating_file_sink-inl.h
```diff
@@ -137,11 +142,24 @@ SPDLOG_INLINE void lazy_rotating_file_sink<Mutex>::rotate_()
...
                 throw_spdlog_ex("lazy_rotating_file_sink: failed renaming " + filename_to_str(src) + " to " + filename_to_str(target), errno);
             }
         }
     }
     file_helper_.reopen(true);
+    file_opened_ = true;
+}
+
+template<typename Mutex>
+SPDLOG_INLINE void lazy_rotating_file_sink<Mutex>::open_if_needed_()
```
andrewlock, line 155: Is it safe to have a mutex for this when it's called from methods that will _already_ have the mutex? 🤔 Actually, I assume so, given `rotate_` was already called from `sink_it_`

tonyredondo: There isn't a mutex here, the template works as a generic here, the mutex is in the base_sink class, and the factory define the final Mutex type.

andrewlock: aaahh, yes 👍

PR #7299, opened 2025-07-29
Update Datadog.Sketches vendored code

bouwkast:
## Summary of changes

Re-vendor the [Datadog.Sketches](https://github.com/DataDog/sketches-dotnet) code.

## Reason for change

We'd like the changes from @brontosaurusgym in https://github.com/DataDog/sketches-dotnet/pull/18 which fixed a bug with `MergeWith()` calls.

## Implementation details

I opted to not cut a new release for Datadog.Sketches and just pointed it to grab from `main` instead to update the code.

~~I then undid the `UpdateVendoredCode` changes and re-ran to just get the correct URL back which means that if we do cut a new release it will work normally, but if we don't cut a release it won't make changes.~~

Main changes are the `store.` additions, it appears the other files haven't been run since `UpdateVendoredCode` has been changed.

## Other details

Thanks @brontosaurusgym


andrewlock (approved):
Perfect, thanks!

The only thing is - should we change this to permanently grab from the current main sha? I don't think we necessarily need to cut a release of the sketches repository, my only concern is if we have to re-run the vendoring for some reason, we will lose the bug fix? 🤔

bouwkast:
So if you rerun the vendoring tool it won't change it because it still points to the 1.0.0 in the last downloaded URL (even though it isn't the latest)
But I can just have it point to main as well in the tool - seems reasonable

PR #7300, opened 2025-07-30
[Test Optimization] Fix base branch SHA usage

tonyredondo:
## Summary of changes

This PR fixes the use of the base branch sha value, we thought this sha was the BASE sha, but we realized it is the HEAD sha of the base branch.

## Reason for change

Jira: https://datadoghq.atlassian.net/browse/SDTEST-2352

## Implementation details

Small updates on the tag names

## Test coverage

Update the tests json files and small modification on the github test.

### DIFF
--- a/tracer/src/Datadog.Trace/Ci/CiEnvironment/GithubActionsEnvironmentValues.cs
+++ b/tracer/src/Datadog.Trace/Ci/CiEnvironment/GithubActionsEnvironmentValues.cs
@@ -121,7 +121,7 @@ internal sealed class GithubActionsEnvironmentValues<TValueProvider>(TValueProvi
                     var prBaseSha = pullRequestObject["base"]?["sha"]?.Value<string>();
                     if (!string.IsNullOrWhiteSpace(prBaseSha))
                     {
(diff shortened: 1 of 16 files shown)


tracer/src/Datadog.Trace/Ci/CiEnvironment/GithubActionsEnvironmentValues.cs
```diff
@@ -121,7 +121,7 @@ private void LoadGithubEventJson()
                     var prBaseSha = pullRequestObject["base"]?["sha"]?.Value<string>();
                     if (!string.IsNullOrWhiteSpace(prBaseSha))
                     {
-                        PrBaseCommit = prBaseSha;
+                        PrBaseHeadCommit = prBaseSha;
```
andrewlock, line 124: So we can't get the `PrBaseCommit` any more right?

tonyredondo: nope, we don't have it for Github actions

PR #7303, opened 2025-07-30
Refactor the profiling configuration reading

andrewlock:
## Summary of changes

Refactors how we determine the profiler's current configuration state

## Reason for change

Today, the profiler determines if it's enabled or disabled independently of the tracer, in the native code. We then try to mirror that state in the managed code, for several reasons

- To know whether we need to send app-domain information to the profiler (e.g. service name, env)
- To know whether we need to send thread-context-change events to the profiler (for CodeHotspots/EndpointProfiling)
- For telemetry reasons
- For recording the state in DSM payloads

As part of the hands-off config work, we will calculate whether to enable the profiler based on the _full_ managed config, instead of just using environment variables. However, when managed activation is disabled (for back-compat investigation purposes), we _do_ need to just use env vars, to "mirror" the state that will be calculated by the native continuous profiler, essentially working in the existing "legacy" state.

The refactoring in this PR is intended to lay the groundwork for the future hands-off config work in the profiler, as well as to solve some existing issues with inconsistent/missing telemetry

## Implementation details

- Add a `ProfilerState` `enum` to track the profiler as a 3-state enablement which represents its real state, instead of trying to force it as a boolean
- Use `EnvironmentConfigurationSource` to record variables read from env in telemetry correctly.
- Add support for the `DD_PROFILER_MANAGED_ACTIVATION_ENABLED` which will be used shortly

## Test coverage

Added unit tests for the new `ProfilerSettings` which encapsulates most of the logic

## Other details

This PR technically requires support for `DD_PROFILER_MANAGED_ACTIVATION_ENABLED` has been merged in order to report the correct values. This isn't a big deal, as it will likely be merged soon, but if it isn't, we should tweak the logic in `ProfilerSettings` so that it's correct. Alternatively, I can split this into two PRs - one that is correct _today_, and one that is correct only with the profiler PR


tracer/src/Datadog.Trace/ContinuousProfiler/ProfilerSettings.cs
```diff
@@ -0,0 +1,101 @@
...
+                             {
+                                 // If there's no explicit `DD_PROFILING_ENABLED` key,
+                                 // we set the state based on the SSI value, only checking env vars (not the full stack)
+                                 var isSsiDeployment = envConfigBuilder
+                                                      .WithKeys(ConfigurationKeys.SsiDeployed)
+                                                      .AsString();
+
+                                 return isSsiDeployment switch
+                                 {
+                                     not null => ContinuousProfiler.ProfilerState.Auto,
```
chrisnas: Should check that it contains `"profiler"`. 
If it does not, should be set to `false`.
This is what is done by the profiler.

We don't emit telemetry metrics anymore so it does not make sense to even p/invoke spans/endpoints when this env var does not contains "profiler"

tracer/src/Datadog.Trace/ContinuousProfiler/ProfilerSettings.cs
```diff
@@ -0,0 +1,101 @@
...
+
+    // Internal for testing only
+    internal ProfilerSettings(ProfilerState state)
+    {
+        ProfilerState = state;
+    }
+
+    /// <summary>
+    /// Gets a value indicating whether the profiler is supported on this platform at all.
+    /// If it's not supported, we should not try to P/Invoke to it or do any context tracking.
```
chrisnas, line 90: the p/invoke should also be skipped if profiler is disabled (=not (`true` or `auto`))

tracer/src/Datadog.Trace/ContinuousProfiler/ProfilerSettings.cs
```diff
@@ -0,0 +1,107 @@
...
+    /// </summary>
+    public static bool IsProfilingSupported
```
chrisnas, line 92: should be private

tracer/src/Datadog.Trace/ContinuousProfiler/ProfilerState.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// </summary>
+    Disabled = 0,
+
+    /// <summary>
+    /// The profiler is explicitly enabled via configuration
+    /// </summary>
+    Enabled = 1,
+
+    /// <summary>
+    /// The profiler is in "auto" mode
```
chrisnas:
```suggestion
    /// The profiler is in "auto" mode; i.e. will start after a delay and if traces are created
```

tracer/src/Datadog.Trace/ContinuousProfiler/ProfilerStatus.cs
```diff
@@ -5,48 +5,23 @@
...
-                {
-                    // the profiler is declared "enabled" just if the SSI environment variable exists to be sure that telemetry metrics
-                    // will contain the right status (i.e. we need the tracer to send the spans even if the profiler is not started yet)
-                    _isProfilingEnabled = (EnvironmentHelpers.GetEnvironmentVariable(ConfigurationKeys.SsiDeployed) != null);
-                }
-            }
-
-            Log.Information("Continuous Profiler is {IsEnabled}.", _isProfilingEnabled ? "enabled" : "disabled");
+            _profilerState = settings.ProfilerState;
+            Log.Information("Continuous Profiler is {IsEnabled}.", _profilerState);
```
chrisnas:
```suggestion
            Log.Information("Continuous Profiler mode = '", ((_profilerState == ProfilerState::Enabled) ? "enabled" : (_profilerState == ProfilerState::Auto) ? "auto" ? "disabled"), "'.");
```

tracer/test/Datadog.Trace.Tests/Telemetry/Collectors/ConfigurationTelemetryCollectorTests.cs
```diff
@@ -231,12 +232,13 @@ public void ConfigurationDataShouldReportDefaultValues()
...
 
         _ = new TracerSettings(source, collector, new OverrideErrorLog());
+        _ = new ProfilerSettings(source, source, collector);
         _ = new SecuritySettings(source, collector);
 
         var data = collector.GetData();
 
         GetLatestValueFromConfig(data, "DD_TRACE_ENABLED", ConfigurationOrigins.Default).Should().Be(true);
-        GetLatestValueFromConfig(data, "DD_PROFILING_ENABLED", ConfigurationOrigins.Default).Should().Be(false);
+        GetLatestValueFromConfig(data, "DD_PROFILING_ENABLED", ConfigurationOrigins.Default).Should().Be("false");
```
chrisnas: Why `"false"` instead of `false`?

tracer/src/Datadog.Trace/ContinuousProfiler/ConfigurationKeys.cs
```diff
@@ -11,5 +11,6 @@ internal static class ConfigurationKeys
         public const string CodeHotspotsEnabled = "DD_PROFILING_CODEHOTSPOTS_ENABLED";
         public const string EndpointProfilingEnabled = "DD_PROFILING_ENDPOINT_COLLECTION_ENABLED";
         public const string SsiDeployed = "DD_INJECTION_ENABLED";
+        public const string ProfilerManagedActivationEnabled = "DD_PROFILING_MANAGED_ACTIVATION_ENABLED ";
```
chrisnas:
```suggestion
        public const string ProfilerManagedActivationEnabled = "DD_PROFILER_MANAGED_ACTIVATION_ENABLED";
```

tracer/test/Datadog.Trace.Tests/ContinuousProfiler/ProfilerSettingsTests.cs
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    [Theory]
+    [InlineData("1", null, (int)ProfilerState.Enabled)]
+    [InlineData("0", null, (int)ProfilerState.Disabled)]
+    [InlineData("auto", null, (int)ProfilerState.Auto)]
+    [InlineData(null, "1", (int)ProfilerState.Disabled)]
+    [InlineData(null, "0", (int)ProfilerState.Disabled)]
+    [InlineData(null, "auto", (int)ProfilerState.Disabled)]
+    public void ProfilerState_WhenManagedActivationIsEnabledOrMissing_ReadsFromAllConfig(string configProfilingEnabled, string envProfilingEnabled, int expected)
```
chrisnas:
```csharp
    public void ProfilerState_WhenManagedActivationIsMissingOrOnByDefault_DontReadFromEnvVars(string configProfilingEnabled, string envProfilingEnabled, int expected)
```

chrisnas (approved):
LGTM except the env var name that should be changed to DD_PROFILER_MANAGED_ACTIVATION_ENABLED instead of DD_PROFILING_MANAGED_ACTIVATION_ENABLED

tracer/src/Datadog.Trace/ContinuousProfiler/ProfilerSettings.cs
andrewlock: I left this public because you'll need it in your stable-config PR I think. i.e. if the profiler isn't even _available_ because it's not supported, we shouldn't call `SetConfiguration`, right? I mean we _could_, but it's better not to I think (avoid the exception for a failed P/Invoke)

tracer/src/Datadog.Trace/ContinuousProfiler/ProfilerSettings.cs
andrewlock: 🤔 Not true I think - we still need to do the `SetConfiguration()` call in the disabled case. But if the profiler isn't running _at all_ (because it's linux arm64, for example) then we shouldn't do that P/Invoke, right?

tracer/test/Datadog.Trace.Tests/ContinuousProfiler/ProfilerSettingsTests.cs
andrewlock: _Technically_ the new name is misleading, because it _will_ read from env vars, just not _just_ env vars, so I tweaked it😄 
```suggestion
    public void ProfilerState_WhenManagedActivationIsMissingOrOnByDefault_DontReadFromOnlyEnvVars(string configProfilingEnabled, string envProfilingEnabled, int expected)
```

tracer/test/Datadog.Trace.Tests/Telemetry/Collectors/ConfigurationTelemetryCollectorTests.cs
andrewlock: In the telemetry configuration, we're now recording a `string`, one of `auto`, `false`, `true`, whereas previously we were recording a `boolean` value. It's just an artifact of the API really, and it all _ultimately_ gets converted to a string in the intake, so it's not a big deal 🙂

tracer/src/Datadog.Trace/ContinuousProfiler/ProfilerSettings.cs
chrisnas: You're right: I was thinking about the span/endpoint p/invoke calls

tracer/src/Datadog.Trace/ContinuousProfiler/ProfilerStatus.cs
```diff
@@ -5,48 +5,30 @@
...
-                    _isProfilingEnabled = (EnvironmentHelpers.GetEnvironmentVariable(ConfigurationKeys.SsiDeployed) != null);
-                }
-            }
+                ProfilerState.Enabled => "enabled",
+                ProfilerState.Auto => "auto",
+                _ => "disabled"
+            };
 
-            Log.Information("Continuous Profiler is {IsEnabled}.", _isProfilingEnabled ? "enabled" : "disabled");
+            Log.Information("Continuous Profiler mode = {ProfilerState}", state);
```
chrisnas, line 31: Maybe I'm not used to the pattern used to log in the tracer, but why do you add {ProfilerState} in the log as a string and not an interpolated string?

I'm expecting more something like:
`   Log.Information("Continuous Profiler mode = ", state);`
or 
`   Log.Information($"Continuous Profiler mode = {state}");`

chrisnas (approved):
LGTM

tracer/src/Datadog.Trace/ContinuousProfiler/ProfilerStatus.cs
andrewlock: In general, you should never use interpolated strings for logging. We even have analyzers that warns you not to 😄 There's three reasons for that:

- It breaks the caching that the log framework does related to parsing out parameters
- It means you don't get structured logging (not actually an issue for us, we log to a flat file, but 'good practice' in general)
- For error logs, we explicitly are only allowed to log "constant" values for security reasons, and interpolation doesn't satisfy that (unless all the parameters are _also_ constants)

In this case, it actually _would_ be fine, because the log is only written once and it's not an error, but I didn't think it was worth creating an explicit `#pragma` to allow it 😄

chrisnas: I get it but why {ProfilerState} as "{ProfilerState}" immediately concatenated to the value of `state`?
We'll get something like the following in the logs:
`   Continuous Profiler = {ProfilerState}Enabled`

I would expect instead:
`   Continuous Profiler mode = Enabled`

andrewlock: That's not how the logging APIs work 😉 

> We'll get something like the following in the logs

No we won't 😄 

> I would expect instead:

That's what you'll get! 😄

andrewlock: Think of it as somewhat equivalent to

```csharp
Log.Information(string.Format("Continuous Profiler mode = {0}", state));
```

but where you _also_ have an associated `ProfilerState=state` ambient variable 🙂

chrisnas: Sorry... My mind is so twisted by C++   ;^)

lucaspimentel (approved):

PR #7304, opened 2025-07-30
Refactor debugger code

dudikeleti:


tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
```diff
@@ -490,51 +464,18 @@ private static void StartDiagnosticManager()
...
-
-                        if (isDiscoverySuccessful)
-                        {
-                            var liveDebugger = LiveDebuggerFactory.Create(discoveryService, RcmSubscriptionManager.Instance, settings, serviceName, tracer.TracerManager.Telemetry, debuggerSettings, tracer.TracerManager.GitMetadataTagsProvider);
-
-                            Log.Debug("Initializing live debugger.");
-
-                            await InitializeLiveDebugger(liveDebugger).ConfigureAwait(false);
-                        }
+                        await DebuggerManager.Instance.UpdateConfiguration().ConfigureAwait(false);
```
andrewlock: I know I've talked about it a few times, but I think we really need to refactor this again to move the debugger management into `TracerManager` and `TracerManagerFactory`, so we can use the correct settings for all our products. It's something I've been thinking about more recently, but doesn't mean anything needs to change here, just an FYI 🙂

tracer/src/Datadog.Trace/Debugger/DebuggerFactory.cs
```diff
@@ -103,17 +90,18 @@ private static DiagnosticsUploader CreateDiagnosticsUploader(IDiscoveryService d
...
         return symbolsUploader;
     }
 
-    private static IApiRequestFactory GetApiFactory(TracerSettings tracerSettings, bool isMultipart)
+    private static IApiRequestFactory GetApiFactory(bool isMultipart)
     {
+        // TODO: we need to be able to update the tracer settings dynamically
         return AgentTransportStrategy.Get(
-            tracerSettings.Exporter,
+            TracerManager.Instance.Settings.Exporter,
```
andrewlock: Yeah, this smells like trouble to me 😞 At the very least, we should be passing in the settings into this method, so that we can control them in unit tests. By accessing the `TracerManager.Instance` here you've basically made this entire type untestable. The tracer settings should likely be passed in from the outer edges, e.g. in `Instrumentation.Initialize`

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
```diff
@@ -31,182 +29,162 @@
...
         private readonly List<ProbeDefinition> _unboundProbes;
         private readonly IProbeStatusPoller _probeStatusPoller;
         private readonly ConfigurationUpdater _configurationUpdater;
         private readonly IDogStatsd _dogStats;
+        private readonly DebuggerSettings _settings;
         private readonly object _instanceLock = new();
-        private bool _isRcmAvailable;
+        private volatile bool _isRcmAvailable;
+        private long _initState = 0; // 0=not initialized, 1=initializing, 2=initialized
+        private long _disposeState = 0; // 0=not disposed, 1=disposing or disposed
```
andrewlock: I know you made these long so that you can use them with `Interlocked.Read()` but that's really not necessary, you can use them as int and just use `Volatile.Read()` 🙂 You can't use `Interlocked.Read()` with `int`s because it's not necessary 😄 
```suggestion
        private int _initState = 0; // 0=not initialized, 1=initializing, 2=initialized
        private int _disposeState = 0; // 0=not disposed, 1=disposing or disposed
```

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
```diff
@@ -31,182 +29,162 @@
...
+                return;
             }
 
             try
             {
-                Log.Information("Live Debugger initialization started");
-                _subscriptionManager.SubscribeToChanges(_subscription);
+                if (!_settings.DynamicInstrumentationEnabled)
+                {
+                    Log.Information("Dynamic Instrumentation is disabled. To enable it, please set DD_DYNAMIC_INSTRUMENTATION_ENABLED environment variable to 'true'.");
```
andrewlock, line 106: nit: I don't think this should be an info log (or any log at all tbh). The customer didn't enable dynamic instrumentation, adding noise to the logs is very low reward and very unlikely to pay off

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
```diff
@@ -31,182 +29,162 @@
...
+                    // Reset to "not initialized"
+                    Interlocked.Exchange(ref _initState, 0);
```
andrewlock, line 108: So we can run this multiple times then?

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
```diff
@@ -31,182 +29,162 @@
...
+                }
 
-                DebuggerSnapshotSerializer.SetConfig(Settings);
-                Redaction.Instance.SetConfig(Settings.RedactedIdentifiers, Settings.RedactedExcludedIdentifiers, Settings.RedactedTypes);
-                AppDomain.CurrentDomain.AssemblyLoad += (sender, args) => CheckUnboundProbes();
+                Log.Information("Dynamic Instrumentation initialization started");
 
-                await StartAsync().ConfigureAwait(false);
+                // Initialize without blocking
+                _ = Task.Run(async () => await InitializeAsync().ConfigureAwait(false), _cancellationTokenSource.Token).ConfigureAwait(false);
```
andrewlock: This whole initialization is already non-blocking though, right? due to other `Task.Run()` elsewhere? I'm worried that creating a bunch of confusion.

 Also, I believe this could cause `UnobservedTaskException` and theoretically bring down the process with the way you're currently passing in the cancellation token. If the token is cancelled, the task fails, and because the output is never observed, if the customer has opted in, we would crash. 

Also, all the ConfigureAwait() etc is unnecessary if you're not awaiting 😄

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
```diff
@@ -31,182 +29,162 @@
...
+                // Reset to "not initialized"
+                Interlocked.Exchange(ref _initState, 0);
             }
         }
 
+        private void StartInBackground()
+        {
+            _probeStatusPoller.StartPolling();
+            _ = _diagnosticsUploader.StartFlushingAsync();
+            _ = _snapshotUploader.StartFlushingAsync();
```
andrewlock, line 170: These should probably be started as explicitly long-running jobs so as to not swamp the threadpool, e.g. like we do elsewhere in the tracer: https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/src/Datadog.Trace/Agent/AgentWriter.cs#L95-L96

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
```diff
@@ -527,31 +550,72 @@ internal void SendMetrics(ProbeInfo probe, MetricKind metricKind, string metricN
...
+                        .Execute(() => _cancellationTokenSource.Cancel(), "cancel dynamic instrumentation operations")
+                        .Execute(() => _discoveryService.RemoveSubscription(DiscoveryCallback), "removing discovery service subscription")
+                        .Execute(() => _subscriptionManager.Unsubscribe(_subscription), "unsubscribing from RCM")
+                        .Add(_snapshotUploader)
+                        .Add(_diagnosticsUploader)
+                        .Add(_probeStatusPoller)
+                        .Add(_dogStats)
+                        .Add(_rcmAvailabilitySemaphore)
+                        .Add(_cancellationTokenSource)
+                        .DisposeAll();
```
andrewlock, line 618: Just FYI, adding `IsDisposed()` into the methods that call these _doesn't_ make them thread safe. e.g. You can have the following sequence

- `Dispose()` is called
- `AddSnapshot()` is called
- `Dispose()` continues, disposing the snapshotuploader
- `AddSnpahot()` continues, and tries to used disposed stuff

tracer/src/Datadog.Trace/Debugger/Expressions/ProbeProcessor.cs
```diff
@@ -419,12 +426,12 @@ private void SetSpanDecoration(DebuggerSnapshotCreator snapshotCreator, ref bool
 
                 if (targetSpan == null)
                 {
-                    Log.Warning("No root span or active span is available, so we can't set the {Tag} tag. Probe ID: {ProbeId}", decoration.TagName, this.ProbeInfo.ProbeId);
+                    Log.Warning("No root span or active span is available, so we can't set the {Tag} tag. Probe ID: {ProbeId}", decoration.TagName, ProbeInfo.ProbeId);
```
andrewlock, line 429: Unrelated to this PR, but should this be a debug log?

tracer/src/Datadog.Trace/Debugger/Expressions/ProbeProcessor.cs
```diff
@@ -472,11 +479,11 @@ private bool TryGetScope([NotNullWhen(true)] out Scope? scope)
                     return scope != null;
                 }
 
-                Log.Warning("Unable to find active scope in WCF context for span decoration. Probe ID: {ProbeId}", this.ProbeInfo.ProbeId);
+                Log.Warning("Unable to find active scope in WCF context for span decoration. Probe ID: {ProbeId}", ProbeInfo.ProbeId);
```
andrewlock, line 482: Same question, should these actually be debug?

tracer/src/Datadog.Trace/Debugger/SpanCodeOrigin/SpanCodeOrigin.cs
```diff
@@ -17,25 +18,24 @@
...
     {
-        private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(SpanCodeOriginManager));
+        private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(SpanCodeOrigin));
 
         private readonly ConcurrentAdaptiveCache<Assembly, AssemblyPdbInfo?> _assemblyPdbCache = new();
-        private readonly DebuggerSettings _settings;
         private readonly CodeOriginTags _tags;
 
-        internal SpanCodeOriginManager()
+        internal SpanCodeOrigin(DebuggerSettings settings)
```
andrewlock, line 28: Big fan of the setting injection like this, as it makes things much more testable 🙂 👍

tracer/src/Datadog.Trace/Debugger/Symbols/SymbolsUploader.cs
```diff
@@ -119,7 +122,7 @@ private void ConfigurationChanged(AgentConfiguration configuration)
             _discoveryService = null;
         }
 
-        public static IDebuggerUploader Create(IBatchUploadApi api, IDiscoveryService discoveryService, IRcmSubscriptionManager remoteConfigurationManager, DebuggerSettings settings, TracerSettings tracerSettings, string serviceName)
+        public static IDebuggerUploader Create(IBatchUploadApi api, IDiscoveryService discoveryService, IRcmSubscriptionManager remoteConfigurationManager, DebuggerSettings settings, string serviceName)
```
andrewlock: Sad to see this _not_ being injected, this is a step back 🙁

tracer/src/Datadog.Trace/Debugger/Symbols/SymbolsUploader.cs
```diff
@@ -385,41 +413,98 @@ private async Task<bool> WaitForDiscoveryServiceAsync()
...
+                catch (ObjectDisposedException)
+                {
+                    // Already disposed, ignore
+                }
+
+                _cancellationToken.Dispose();
+                _assemblySemaphore.Dispose();
+                _enablementSemaphore.Dispose();
+                _discoveryServiceSemaphore.Dispose();
+                _disposed = true;
```
andrewlock: By setting this at the end, `Dispose()` could be called on these multiple times, which will typically cause exceptions. You'll also end up calling `_subscriptionManager.Unsubscribe` multiple times, etc.

tracer/src/Datadog.Trace/Span.cs
```diff
@@ -432,7 +433,7 @@ internal void SetExceptionTags(Exception exception)
                     SetTag(Trace.Tags.ErrorType, exception.GetType().ToString());
                     SetTag(Trace.Tags.ErrorStack, exception.ToString());
 
-                    ExceptionDebugging.Report(this, exception);
+                    DebuggerManager.Instance.ExceptionReplay?.Report(this, exception);
```
andrewlock, line 436: This could be a problem 😟 Today, a Span is tied to a _specific_ `Tracer` instance (via a `TraceContext`, which is _not_ a global instance. Calling `DebuggerManager.Instance` here will cause the live debugger to be initialized in all our unit tests etc, which could cause a bunch of issues.

This really all goes back to what I have been talking about for a while, about having `DebuggerManager` exposed on `TracerManager` (which is itself available from a `Tracer` instance).

tracer/src/Datadog.Trace/Debugger/DebuggerFactory.cs
dudikeleti: Fixed in e3fcc166ef63085f0d14e1a47982dd747efaa271

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
dudikeleti: Sure. Fixed in 6251269dc863baf000af2a21e0565bd6b39f3871

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
dudikeleti: We use that log line in integration tests

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
dudikeleti: Theoratically (remote enabement/other initialization flow like LiveDebugger in the future)

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
dudikeleti: Yes, I thought maybe wrapping it all with lock (or other solution) but on the other hand, can we live with ending with exception here? if we decide to make it real safe we have to make it like this for DI and ER.
I can do that in the next PR that will implement the dynamic settings.

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
dudikeleti: I prefer to make inner class changes in a future PR (after the refactore and the dynamic configuration will be merged), you know where it starts but not where it ends.

tracer/src/Datadog.Trace/Debugger/Expressions/ProbeProcessor.cs
dudikeleti: Making that Warning helped me a couple of times to understand customer issues without asking them to change to Debug. But I agree with you, I'll probably change it later to Debug.

tracer/src/Datadog.Trace/Debugger/Expressions/ProbeProcessor.cs
dudikeleti: Same :)

tracer/src/Datadog.Trace/Debugger/Symbols/SymbolsUploader.cs
dudikeleti: Fixed in 038fd6f7ca201e3377ff9d9622ae8bdbf24909ec

tracer/src/Datadog.Trace/Debugger/Symbols/SymbolsUploader.cs
dudikeleti: Fixed in 5e09178e208e8ee5859df399adc31c210c269ed1

tracer/src/Datadog.Trace/Span.cs
dudikeleti: I understand. We defently have to change it in the future but for now, only by calling DebuggerManager.Instance you don't initialized DI, ER etc.
Anyway, lets do what you saying about having DebuggerManager exposed on TracerManager as a follow-up PR.

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
dudikeleti: > This whole initialization is already non-blocking though, right?
> I believe this could cause UnobservedTaskException

I will address it in the next PR, when dynamic configuration changes will come into the picture

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
andrewlock: > Yes, I thought maybe wrapping it all with lock (or other solution) 

Yeah, the trouble with that is that all of the `AddSnapshot` calls etc would need a lock too 😬 Fundamentally it's kind of tricky. The "safe" thing is probably just to wrap each of the callsites for these types in try/catch blocks and just accept that there could be an ObjectDisposed Exception.

> can we live with ending with exception here

As long as it's caught, that's ok. The problem for me here is the callsites, where they _aren't_ wrapped in try-catch universally I think.

tracer/src/Datadog.Trace/Debugger/Expressions/ProbeProcessor.cs
andrewlock: My main concern is when we see the logs flooded with them. If we don't see that, it's fine 🙂

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
dudikeleti: > wrap each of the callsites for these types in try/catch blocks
The problem for me here is the callsites, where they aren't wrapped in try-catch universally

Exactly. That's why I thought about handle it now or not, because it involves changes in many places. I can protect the call site with a try catch or use reference counting everywhere. We'll take care of that in the next PR.

PR #7308, opened 2025-07-31
Bump timeitsharp to v0.4.5

tonyredondo:

### DIFF
--- a/profiler/build/install_timeit.cmd
+++ b/profiler/build/install_timeit.cmd
@@ -1,6 +1,6 @@
 @echo off
 
-dotnet tool update -g timeitsharp --version 0.4.4
+dotnet tool update -g timeitsharp --version 0.4.5
 
 rem Add %USERPROFILE%\.dotnet\tools to Path if it is not already there
 path|find /i "%USERPROFILE%\.dotnet\tools" >nul || set path=%path%;"%USERPROFILE%\.dotnet\tools"
--- a/profiler/build/install_timeit.sh
+++ b/profiler/build/install_timeit.sh
@@ -1,2 +1,2 @@
 #!/usr/bin/env bash
-dotnet tool update -g timeitsharp --version 0.4.4
+dotnet tool update -g timeitsharp --version 0.4.5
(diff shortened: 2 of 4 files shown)


profiler/build/install_timeit.cmd
```diff
@@ -1,6 +1,6 @@
 @echo off
 
-dotnet tool update -g timeitsharp --version 0.4.4
+dotnet tool update -g timeitsharp --version 0.4.5
```
andrewlock:
```suggestion
dotnet tool update -g timeitsharp --version 0.4.5 --allow-downgrade
```

profiler/build/install_timeit.sh
```diff
@@ -1,2 +1,2 @@
 #!/usr/bin/env bash
-dotnet tool update -g timeitsharp --version 0.4.4
+dotnet tool update -g timeitsharp --version 0.4.5
```
andrewlock:
```suggestion
dotnet tool update -g timeitsharp --version 0.4.5 --allow-downgrade
```

tracer/build/timeit/Samples.FakeDbCommand/run.cmd
```diff
@@ -11,7 +11,7 @@ set FAILED=0
 echo *********************
 echo Installing timeitsharp
 echo *********************
-dotnet tool update -g timeitsharp --version 0.4.4
+dotnet tool update -g timeitsharp --version 0.4.5
```
andrewlock:
```suggestion
dotnet tool update -g timeitsharp --version 0.4.5 --allow-downgrade
```

tracer/build/timeit/Samples.HttpMessageHandler/run.cmd
```diff
@@ -11,7 +11,7 @@ set FAILED=0
 echo *********************
 echo Installing timeitsharp
 echo *********************
-dotnet tool update -g timeitsharp --version 0.4.4
+dotnet tool update -g timeitsharp --version 0.4.5
```
andrewlock:
```suggestion
dotnet tool update -g timeitsharp --version 0.4.5 --allow-downgrade
```

andrewlock (approved):
Because this is doing global installs, every time you update timeitsharp, we break master and everyone else's branch 😅 

We should add the `--allow-downgrade` flag _now_ so that we don't break them with future updates.

Alternatively, we could install local tools. Or use `dnx` once we ship the .NET 10 SDK

PR #7310, opened 2025-07-31
Stop using GITHUB_TOKEN for pr creation

xopham:

### DIFF
--- a/.github/workflows/auto_bump_test_package_versions.yml
+++ b/.github/workflows/auto_bump_test_package_versions.yml
@@ -2,9 +2,9 @@ name: Auto bump test package versions
 
 on:
   schedule:
-    - cron: '0 0 * * 0' # Every Sunday at midnight
+    - cron: "0 0 * * 0" # Every Sunday at midnight
   pull_request_target:
-    branches: [master, main]
+    branches: [master]
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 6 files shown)


.github/workflows/auto_bump_test_package_versions.yml
```diff
@@ -2,9 +2,9 @@ name: Auto bump test package versions
 
 on:
   schedule:
-    - cron: '0 0 * * 0' # Every Sunday at midnight
+    - cron: "0 0 * * 0" # Every Sunday at midnight
   pull_request_target:
-    branches: [master, main]
+    branches: [master]
```
xopham, line 7: there is no `main` branch, so using "main" actually makes this insecure

andrewlock: we had the intention to move from master -> main at some point, but you know, never happened 😅

.github/workflows/auto_create_version_bump_pr.yml
```diff
@@ -10,20 +10,27 @@ jobs:
...
+
     if: endsWith(github.event.release.tag_name, '.0') || endsWith(github.event.release.tag_name, '.0-prerelease')
     runs-on: windows-latest
     permissions:
       contents: write # Creates a branch
-      pull-requests: write # Creates a PR
       issues: write # Closes milestones
+      id-token: write # Required for dd-octo-sts authentication
     env:
       GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
```
andrewlock, line 21: Do we still need this? Should we instead use the id-token everywhere? It's currently used for calling the GitHub API, for example, but would it be better to use the dd-octo-sts token instead?

bouwkast (approved):
Thanks!

andrewlock (commented):
Thanks! I just did a test run with one of the workflows though, and it didn't work 😅  https://github.com/DataDog/dd-trace-dotnet/actions/runs/16649587516/job/47118484563

.github/workflows/auto_bump_test_package_versions.yml
xopham: gotcha, see how that happens. Sorry, for side fixing things here 😅

PR #7311, opened 2025-07-31
Refactor `TraceSampler` to separate building from evaluation

andrewlock:
## Summary of changes

Refactors the `TraceSampler` and `ITraceSampler` abstractions to separate the building of the rules from the evaluation

## Reason for change

It's neater, and it makes some subsequent changes much simpler, easier and more performant

## Implementation details

Create a nested `Builder` type in `TraceSampler` and move all the `RegisterRule()` methods on it. Call `Build` to get a final list of rules to be applied.

## Test coverage

Covered by existing tests

## Other details

Part of a stack of branches related to sampling improvements

- https://github.com/DataDog/dd-trace-dotnet/pull/7311 👈
- https://github.com/DataDog/dd-trace-dotnet/pull/7312

### DIFF
--- a/tracer/src/Datadog.Trace/Sampling/TraceSampler.cs
+++ b/tracer/src/Datadog.Trace/Sampling/TraceSampler.cs
@@ -136,11 +136,10 @@ namespace Datadog.Trace.Sampling
             public void RegisterAgentSamplingRule(AgentSamplingRule rule)
             {
                 // only register the one AgentSamplingRule
+                // keep a reference to this rule so we can call SetDefaultSampleRates() later
+                // to update the agent sampling rates
                 if (Interlocked.Exchange(ref _agentSamplingRule, rule) == null)
                 {
-                    // keep a reference to this rule so we can call SetDefaultSampleRates() later
-                    // to update the agent sampling rates
-
                     RegisterRule(rule);
                 }
                 else
--- a/tracer/src/Datadog.Trace/Sampling/TraceSampler.cs
+++ b/tracer/src/Datadog.Trace/Sampling/TraceSampler.cs
@@ -140,7 +140,6 @@ namespace Datadog.Trace.Sampling
                 {
                     // keep a reference to this rule so we can call SetDefaultSampleRates() later
                     // to update the agent sampling rates
-                    _agentSamplingRule = rule;
 
                     RegisterRule(rule);
                 }
--- a/tracer/src/Datadog.Trace/Sampling/TraceSampler.cs
+++ b/tracer/src/Datadog.Trace/Sampling/TraceSampler.cs
@@ -17,13 +17,14 @@ namespace Datadog.Trace.Sampling
         private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor<TraceSampler>();
 
         private readonly IRateLimiter _limiter;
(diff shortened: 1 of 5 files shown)


tracer/src/Datadog.Trace/Sampling/TraceSampler.cs
```diff
@@ -137,5 +87,68 @@ private SamplingDecision MakeSamplingDecision(Span span, float rate, string mech
...
+            /// The first rule that matches will be used to determine the sampling rate.
+            /// </remarks>
+            public void RegisterAgentSamplingRule(AgentSamplingRule rule)
+            {
+                // only register the one AgentSamplingRule
+                if (Interlocked.Exchange(ref _agentSamplingRule, rule) == null)
+                {
+                    // keep a reference to this rule so we can call SetDefaultSampleRates() later
+                    // to update the agent sampling rates
+                    _agentSamplingRule = rule;
```
bouwkast:
```suggestion
```

Just seeing this but I don't think we need this set as we are doing a `Interlocked.Exchange` on it?

Was already there though.

bouwkast (approved):

tracer/src/Datadog.Trace/Sampling/TraceSampler.cs
andrewlock: Oh yeah, I didn't read this code at all 😅 I agree, I don't see why the set is necessary 👍

tracer/src/Datadog.Trace/Sampling/TraceSampler.cs
```diff
@@ -137,5 +87,67 @@ private SamplingDecision MakeSamplingDecision(Span span, float rate, string mech
...
+            /// </remarks>
+            public void RegisterAgentSamplingRule(AgentSamplingRule rule)
+            {
+                // only register the one AgentSamplingRule
+                if (Interlocked.Exchange(ref _agentSamplingRule, rule) == null)
+                {
+                    // keep a reference to this rule so we can call SetDefaultSampleRates() later
+                    // to update the agent sampling rates
+
+                    RegisterRule(rule);
```
andrewlock:
```suggestion
                // only register the one AgentSamplingRule
                // keep a reference to this rule so we can call SetDefaultSampleRates() later
                // to update the agent sampling rates
                if (Interlocked.Exchange(ref _agentSamplingRule, rule) == null)
                {
                    RegisterRule(rule);
```

tracer/src/Datadog.Trace/Sampling/TraceSampler.cs
```diff
@@ -17,13 +17,14 @@ internal class TraceSampler : ITraceSampler
...
+        private readonly List<ISamplingRule> _rules;
+        private readonly AgentSamplingRule? _agentSamplingRule;
 
-        private AgentSamplingRule? _agentSamplingRule;
-
-        public TraceSampler(IRateLimiter limiter)
+        public TraceSampler(IRateLimiter limiter, List<ISamplingRule> rules, AgentSamplingRule? agentSamplingRule)
         {
             _limiter = limiter;
+            _rules = [..rules];
```
tonyredondo, line 26: ohhh niceee

tonyredondo (approved):

PR #7312, opened 2025-07-31
Avoid generating resource name that will be thrown away if it's not required

andrewlock:
## Summary of changes

- Add `HasResourceBasedSamplingRule` to `ITraceSampler` and `ISpanSampler`
- If there are no resource-based sampling rule, we can avoid calculating the resource name

## Reason for change

In #6936, we fixed an issue where resource-based sampling wasn't work. This was because we can't calculate the "final" resource name for a span until after the MVC framework has run. Previously, we were delaying the resource name calculation to this point. For the fix, we calculate a _provisional_ resource name, which we subsequently change. It's sub-optimal in that 

- The resource name changes
- If the customer _isn't_ using resource based sampling, we are allocating for ultimately no benefit

This PR exposes a property which lets us check whether resource based sampling is being used. If it is, then we _must_ calculate the resource name up-front. If it's not, then we can delay creation for performance reasons.

## Implementation details

- Expose `IsResourceBasedSamplingRule` on `ISamplingRule` and `ISpanSamplingRule`
- Expose `HasResourceBasedSamplingRule` on `ISampler` and `ISpanSampler`
- Expose `HasResourceBasedSamplingRule` on `PerTraceSettings`
- Check this value prior to creating the initial resource name in `TracingHttpModule`

## Test coverage

- Added unit tests for the new properties
- Existing integration tests cover the implementation in IIS

## Other details

This is mostly a setup for adding a fix to the ASP.NET Core provider to give the same functionality.

In general, it feels a bit hacky, and it assumes that nothing _else_ (other than sampling) is dependent on the resource name. _Ideally_, we would have a system like @dougqh has discussed previously to evaluate properties lazily based on the available data in the span. But until then, this is a a stop gap which means we can avoid the perf hit for the majority of customers who _don't_ have resource-based sampling.

Part of a stack
- https://github.com/DataDog/dd-trace-dotnet/pull/7311
- https://github.com/DataDog/dd-trace-dotnet/pull/7312 👈
- https://github.com/DataDog/dd-trace-dotnet/pull/7316

### DIFF
--- a/tracer/src/Datadog.Trace/Sampling/SpanSampler.cs
+++ b/tracer/src/Datadog.Trace/Sampling/SpanSampler.cs
@@ -25,7 +25,15 @@ internal class SpanSampler : ISpanSampler
         }
 
         _rules = rules.ToList();
-        HasResourceBasedSamplingRule = _rules.Any(rule => rule.IsResourceBasedSamplingRule);
+        HasResourceBasedSamplingRule = false;
+        foreach (var rule in _rules)
+        {
+            if (rule.IsResourceBasedSamplingRule)
+            {
+                HasResourceBasedSamplingRule = true;
+                break;
+            }
+        }
     }
 
     /// <inheritdoc/>
--- a/tracer/src/Datadog.Trace/Sampling/TraceSampler.cs
+++ b/tracer/src/Datadog.Trace/Sampling/TraceSampler.cs
@@ -26,7 +26,15 @@ namespace Datadog.Trace.Sampling
             _limiter = limiter;
             _rules = [..rules];
             _agentSamplingRule = agentSamplingRule;
-            HasResourceBasedSamplingRule = rules.Any(rule => rule.IsResourceBasedSamplingRule);
+            HasResourceBasedSamplingRule = false;
+            foreach (var rule in rules)
+            {
+                if (rule.IsResourceBasedSamplingRule)
+                {
+                    HasResourceBasedSamplingRule = true;
+                    break;
+                }
+            }
         }
 
         public bool HasResourceBasedSamplingRule { get; }


tracer/src/Datadog.Trace/Configuration/PerTraceSettings.cs
```diff
@@ -22,6 +22,7 @@ public PerTraceSettings(ITraceSampler? traceSampler, ISpanSampler? spanSampler,
             SpanSampler = spanSampler;
             ServiceNames = serviceNames;
             Schema = schema;
+            HasResourceBasedSamplingRule = (traceSampler?.HasResourceBasedSamplingRule ?? false) || (traceSampler?.HasResourceBasedSamplingRule ?? false);
```
bouwkast:
```suggestion
            HasResourceBasedSamplingRule = (traceSampler?.HasResourceBasedSamplingRule ?? false) || (spanSampler?.HasResourceBasedSamplingRule ?? false);
```

tracer/src/Datadog.Trace/Sampling/CustomSamplingRule.cs
```diff
@@ -50,6 +50,8 @@ _resourceNameRegex is null &&
             }
         }
 
+        public bool IsResourceBasedSamplingRule => _resourceNameRegex is not null;
```
bouwkast, line 53:
```suggestion
        public bool IsResourceBasedSamplingRule => _resourceNameRegex is not null && !_alwaysMatch;
```

Wondering if we should treat `_alwaysMatch` (`*`) as being something that isn't a resource-based sampling rule. As I think as is, if it was always match we'd go and compute that provisional resource name?

bouwkast: Oh wait, that above logic only is `_alwaysMatch` when they are all null.

Maybe we want to go and add a different setting to catch when the resource name is a always match?

bouwkast (approved):

tracer/src/Datadog.Trace/Configuration/PerTraceSettings.cs
bouwkast: Copy/paste error

tracer/src/Datadog.Trace/Sampling/CustomSamplingRule.cs
andrewlock: Yeah `_alwaysMatch` is included so we don't need that. And even better, when the resource name is a always match, then `_resourceNameRegex` is null, so both of those are accounted for 🙂

tracer/src/Datadog.Trace/Configuration/PerTraceSettings.cs
andrewlock: The _one_ type where I was like "shall I create a unit test... nah, it's obvious from the code that it's correct" 🤦‍♂️ 💀

zacharycmontoya (approved):

tracer/src/Datadog.Trace/Sampling/SpanSampler.cs
```diff
@@ -25,8 +25,12 @@ public SpanSampler(IEnumerable<ISpanSamplingRule> rules)
         }
 
         _rules = rules.ToList();
+        HasResourceBasedSamplingRule = _rules.Any(rule => rule.IsResourceBasedSamplingRule);
```
lucaspimentel: MY EYES

lucaspimentel: Think of all the avoidable heap allocations! Iterators, delegates, oh my!

tracer/src/Datadog.Trace/Sampling/TraceSampler.cs
```diff
@@ -25,8 +26,11 @@ public TraceSampler(IRateLimiter limiter, List<ISamplingRule> rules, AgentSampli
             _limiter = limiter;
             _rules = [..rules];
             _agentSamplingRule = agentSamplingRule;
+            HasResourceBasedSamplingRule = rules.Any(rule => rule.IsResourceBasedSamplingRule);
```
lucaspimentel: More LINQ?!

lucaspimentel (approved):

tracer/src/Datadog.Trace/Configuration/PerTraceSettings.cs
lucaspimentel: Unresolving thread since it's not fixed yet.

tracer/src/Datadog.Trace/Sampling/SpanSampler.cs
andrewlock: It's a one-off call, but I can refactor it as it bothers you so much 😝

andrewlock: _also_ newer versions of Linq are actually sometimes more performant because they use SIMD instructions, so they can out-perform `foreach` 😉

PR #7316, opened 2025-07-31
Fix resource-based sampling for ASP.NET Core

andrewlock:
## Summary of changes

Eagerly calculate a stand-in resource name if resource-based sampling is being used

## Reason for change

In https://github.com/DataDog/dd-trace-dotnet/pull/6936, we fixed an issue where resource-based sampling wasn't working in IIS. This was because we can't calculate the "final" resource name for a span until after the MVC framework has run. Previously, we were delaying the resource name calculation to this point. For the fix, we calculated a provisional resource name, which we subsequently change. In #7312 we  updated the logic to only generate the resource name when resource-based sampling was being used.

This PR fixes the same issue for ASP.NET Core

## Implementation details

- Added unit and integration tests to demonstrate the issue
  - For the integration tests, we have to make a sampling decision after the pipeline starts, but before routing is hit.
  - To avoid creating new samples or breaking existing ones, I manually forced a sampling decision to the existing `PingMiddleware` (which is invoked before routing kicks in).
  - The initial commit contains the snapshots in their "incorrect" state. The second commit shows the "diff" showing the sampling working
- Make resource name generation eager in `AspNetCoreDiagnosticObserver` if we have resource-based sampling

## Test coverage

Unit tests + integration tests, IIS vs standalone, In and out of process

## Other details

Part of a stack

- https://github.com/DataDog/dd-trace-dotnet/pull/7311
- https://github.com/DataDog/dd-trace-dotnet/pull/7312
- https://github.com/DataDog/dd-trace-dotnet/pull/7316 👈

https://datadoghq.atlassian.net/browse/APMS-16673

### DIFF
--- a/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNetCore/AspNetCoreMvcResourceBasedSamplingTest.cs
+++ b/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNetCore/AspNetCoreMvcResourceBasedSamplingTest.cs
@@ -109,7 +109,8 @@ public abstract class AspNetCoreMvcResourceBasedSamplingTestBase : AspNetCoreMvc
         _testName = testName;
     }
 
-    [Fact]
+    [SkippableFact]
+    [Trait("RunOnWindows", "True")]
     public async Task TestSampledSpan()
     {
         var path = "/ping";
@@ -146,7 +147,10 @@ public abstract class AspNetCoreIisMvcResourceBasedSamplingTestBase : AspNetCore
         SetEnvironmentVariable(ConfigurationKeys.CustomSamplingRulesFormat, SamplingRulesFormat.Glob); // for ease of use
     }
 
-    [Fact]
+    [SkippableFact]
+    [Trait("RunOnWindows", "True")]
+    [Trait("Category", "EndToEnd")]
+    [Trait("Category", "LinuxUnsupported")]
     public async Task TestSampledSpan()
     {
         var path = "/ping";


tracer/src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
```diff
@@ -442,7 +442,10 @@ private void OnHostingHttpRequestInStart(object arg)
                 {
                     // Use an empty resource name here, as we will likely replace it as part of the request
                     // If we don't, update it in OnHostingHttpRequestInStop or OnHostingUnhandledException
-                    var scope = AspNetCoreRequestHandler.StartAspNetCorePipelineScope(tracer, CurrentSecurity, httpContext, resourceName: string.Empty);
+                    // If the app is using resource-based sampling rules, then we need to set a resource straight
+                    // away, so force that by using null.
+                    var resourceName = tracer.TracerManager.PerTraceSettings.HasResourceBasedSamplingRule ? null : string.Empty;
+                    var scope = AspNetCoreRequestHandler.StartAspNetCorePipelineScope(tracer, CurrentSecurity, httpContext, resourceName);
```
andrewlock, line 448: This is the actual fix

tracer/test/snapshots/AspNetCoreIisMvc21MvcResourceBasedSamplingTests._.verified.txt
```diff
@@ -0,0 +1,30 @@
...
+      span.kind: server,
+      version: 1.0.0
+    },
+    Metrics: {
+      process_id: 0,
+      _dd.rule_psr: 0.0,
+      _dd.top_level: 1.0,
+      _dd.tracer_kr: 1.0,
+      _sampling_priority_v1: -1.0
+    }
```
andrewlock, line 28: FYI, the fix changes the snapshots from the following:

```
     Metrics: {
       process_id: 0,
       _dd.top_level: 1.0,
       _dd.tracer_kr: 1.0,
       _sampling_priority_v1: 1.0
     }
```


to what you see now. i.e.

```diff
     Metrics: {
       process_id: 0,
+      _dd.rule_psr: 0.0,
       _dd.top_level: 1.0,
       _dd.tracer_kr: 1.0,
-      _sampling_priority_v1: 1.0
+      _sampling_priority_v1: -1.0
     }
```

bouwkast (approved):
Looks good to me!

lucaspimentel (approved):

PR #6733, opened 2025-03-03
[ASM] API Security Endpoints collection (kestrel)

e-n-0:
## Summary of changes

This PR introduce the collection of endpoints using ASP.NET Core in part of Api Security.
[RFC](https://docs.google.com/document/d/1txwuurIiSUWjYX7Xa0let7e49XKW2uhm1djgqjl_gL0/edit?pli=1&tab=t.0#heading=h.8d3o7vtyu1y1).

- 2 new environment variables [(PR on dd-go)](https://github.com/DataDog/dd-go/pull/172566):
	- `DD_API_SECURITY_ENDPOINT_COLLECTION_ENABLED`
	- `DD_API_SECURITY_ENDPOINT_COLLECTION_MESSAGE_LIMIT`

- Sending a new telemetry payload `app-endpoints` that includes the detected endpoints.

- The feature will be merged with a default activation value of `false` because the backend part is not done yet. When this will be available, it will be changed to `true`

## Implementation details

This PR adds 2 new instrumentation points:
- On `KestrelServerImpl` just before the server starts accepting connection to collect all router endpoints
- On MapExtensions ([Map](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.mapextensions.map?view=aspnetcore-9.0)) and RunExtensions ([Run](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.runextensions.run?view=aspnetcore-9.0#microsoft-aspnetcore-builder-runextensions-run(microsoft-aspnetcore-builder-iapplicationbuilder-microsoft-aspnetcore-http-requestdelegate))) to fully get the endpoints configured this way

## Test coverage

- Unit tests for the new environment variables (checking their values and apply states).
- Unit tests on the Telemetry Controller to see if the telemetry data is correctly reported
- Integration tests with endpoints collection enabled/disabled and checking if the telemetry data is correctly reported.
	- Also checks if special endpoints are reported
	- On AspNetCore MVC and Bare

# Other details

- The PR for ASP.NET (framework) will come in a second part.

- Another PR will come to add extra metadata to the telemetry payload (as display name, authentication information, ..)

- There is some endpoints that we can't report at the moment with the current implementation. Endpoints configured as middleware other than with `.Map()` needs to be instrumented (like for example `UseDirectoryBrowser`). The listing of them and their support will come later.

PR #7317, opened 2025-07-31
[AAP] Add resource name to the collected endpoints

e-n-0:
## Summary of changes

Added the ResourceName to the endpoints collected based on the last update of the [RFC](https://docs.google.com/document/d/1txwuurIiSUWjYX7Xa0let7e49XKW2uhm1djgqjl_gL0).
## Implementation details

- The resource name is defined as `method + path` except in the case of an endpoint detected as a wildcard, the resource name is set to the path.

## Test coverage

Updated the tests.

## Other details

#6733

### DIFF
--- a/tracer/src/Datadog.Trace/Telemetry/DTOs/AppEndpointData.cs
+++ b/tracer/src/Datadog.Trace/Telemetry/DTOs/AppEndpointData.cs
@@ -35,6 +35,7 @@ internal class AppEndpointData
         Method = method;
         Path = path;
         OperationName = "http.request";
+        ResourceName = $"{method} {path}";
     }
 
     /// <summary>
@@ -58,4 +59,9 @@ internal class AppEndpointData
     /// Gets or sets the operation name for the endpoint.
     /// </summary>
     public string OperationName { get; set; }
+
+    /// <summary>
+    /// Gets or sets the resource name for the endpoint.
+    /// </summary>
+    public string ResourceName { get; set; }
 }
--- a/tracer/src/Datadog.Trace/Telemetry/DTOs/AppEndpointData.cs
+++ b/tracer/src/Datadog.Trace/Telemetry/DTOs/AppEndpointData.cs
@@ -35,7 +35,7 @@ internal class AppEndpointData
         Method = method;
         Path = path;
         OperationName = "http.request";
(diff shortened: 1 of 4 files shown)


tracer/src/Datadog.Trace/Telemetry/DTOs/AppEndpointData.cs
```diff
@@ -35,6 +35,7 @@ public AppEndpointData(string method, string path)
         Method = method;
         Path = path;
         OperationName = "http.request";
+        ResourceName = method == "*" ? path : $"{method} {path}";
```
andrewlock, line 38: Shouldn't this be the _actual_ resource name, e.g. the resource name that is applied to spans? 🤔 Because you will end up with an explosion of resource names for paths that include an identifier, which is generally not useful for customers in APM so feels suspect here too

i.e. instead of 1000 different endpoints called 
- `/customer/1`
- `/customer/234`
- `/customer/464`

You really want one endpoint called
- `/customer/{customerId}`

e-n-0: The path should have been passed to the method `AspNetCoreResourceNameHelper.SimplifyRoutePattern` before.

https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/src/Datadog.Trace/AppSec/ApiSec/EndpointsCollection.cs#L66
```cs
path = AspNetCoreResourceNameHelper.SimplifyRoutePattern(routePattern, routePatternV3.RequiredValues, areaName, controllerName, actionName, false);
```

Isn't that is doing a simple quantization before?

dromanol: I'd add a specific test confirming this point

andrewlock: > The path should have been passed to the method

Ah ok, so it's not a path at all, it's a route 😅 In which case it's just a naming issue and you're correct, it should be fine 🙂

e-n-0: I added a test using an endpoint with quantized params 83ef131

PR #7320, opened 2025-08-01
Cache response in order to avoid unnecessary deserializations

hoolioh:
## Summary of changes
Cache agent's response to avoid unnecessary deserializations in case sample rates haven't changed from previous response.
## Reason for change
Reduce performance overhead.
## Other details
This PR implements part of the functionality of this [one](https://github.com/DataDog/dd-trace-dotnet/pull/7251). As commented the rates versioning payload will be handled internally in the native exporter but until it is available the response will be cached in the same way as in the managed exporter.


andrewlock (approved):
LGTM, I think we can enhance the `TraceExporterTests.SendsTracesUsingDataPipeline` test to test this functionality, by sending 2 separate traces, returning the same agent response, and confirming that `_update_sampleRates` is only invoked once 🙂

PR #7321, opened 2025-08-01
Send Warm-up request in iisexpress integration tests

NachoEchevarria:
## Summary of changes

We have identified some flaky tests when running under iisexpress. To address part of the issue, we have already scrubbed certain connection headers from the snapshots, as the presence of specific flags from a given request may not be consistent. This inconsistency is influenced by test execution order and cannot always be predicted.

Additionally, in ASM, certain tags, such as the WAF version, are sent during the first request. In other frameworks, this is typically handled by warm-up calls, which helps avoid such issues. However, iisexpress tests currently do not include warm-up calls. Introducing a warm-up call would help produce more predictable test results and resolve some of the inconsistencies we have been encountering.


tracer/test/Datadog.Trace.TestHelpers.AutoInstrumentation/IisFixture.cs
```diff
@@ -95,5 +99,84 @@ public void Dispose()
                 }
             }
         }
+
```
NachoEchevarria, line 102: Adapted from AspNetCoreTestFixture

tracer/test/Datadog.Trace.Tools.dd_dotnet.ArtifactTests/Checks/IisCheckTests.cs
```diff
@@ -230,7 +230,7 @@ private async Task<IisFixture> StartIis(IisAppType appType)
 
         try
         {
-            await fixture.TryStartIis(this, appType);
+            await fixture.TryStartIis(this, appType, false);
```
NachoEchevarria, line 233: This class has a more relaxed httpget below that allows forbidden status codes to pass the check.

tracer/test/snapshots/Security.AspNetMvc5AsmData.Classic.enableSecurity=True.__test=blocking-ips_url=_.verified.txt
```diff
@@ -44,17 +44,14 @@
       runtime-id: Guid_1,
       span.kind: server,
       _dd.appsec.event_rules.version: 1.13.3,
-      _dd.appsec.waf.version: 1.27.0,
```
andrewlock, line 47: I guess these changes are expected because they are now sent in the health check request, correct?

tracer/test/snapshots/Security.AspNetWebFormsAsmData.Classic.enableSecurity=True.__test=blocking-user_url=_user.verified.txt
```diff
@@ -21,17 +21,14 @@
...
     },
     Metrics: {
       process_id: 0,
       _dd.appsec.enabled: 1.0,
-      _dd.appsec.event_rules.error_count: 0.0,
-      _dd.appsec.event_rules.loaded: 159.0,
       _dd.top_level: 1.0,
       _dd.tracer_kr: 1.0,
-      _sampling_priority_v1: 2.0
+      _sampling_priority_v1: 1.0
```
andrewlock, line 31: Is this change again because normally we would force-keep the span as it contains the ASM data, but we don't have to because the health check has that instead (and so will have `_sampling_priority_v1: 2.0` instead)?

tracer/test/snapshots/Security.AspNetMvc5AsmData.Classic.enableSecurity=True.__test=blocking-ips_url=_.verified.txt
NachoEchevarria: Yes, these values are sent during the first request. The problem is that we had some snapshots having them and some others not based on test order.

tracer/test/snapshots/Security.AspNetWebFormsAsmData.Classic.enableSecurity=True.__test=blocking-user_url=_user.verified.txt
NachoEchevarria: That's correct :)

PR #7327, opened 2025-08-01
Minor `IConfigurationSource` clean up

andrewlock:
## Summary of changes

- Remove unused methods in `IConfigurationSource`
- Add a test for the recording of "error" telemetry in configuration sources

## Reason for change

I'm working on some things for configuration, and needed to do a bit of cleanup first. Nothing major here.

## Implementation details

- Delete unused `CompositeConfigurationSource.Insert()` method
- Delete unused `IConfigurationSource.IsPresent` method
- Add a test that we record "invalid" errors, and that the final telemetry value is the "real" value

## Test coverage

Covered by existing + a new one

### DIFF
--- a/tracer/src/Datadog.Trace/Configuration/ConfigurationSources/Telemetry/ConfigurationBuilder.cs
+++ b/tracer/src/Datadog.Trace/Configuration/ConfigurationSources/Telemetry/ConfigurationBuilder.cs
@@ -66,6 +66,7 @@ internal readonly struct ConfigurationBuilder
         bool recordValue,
         Func<DefaultResult<T>>? getDefaultValue,
         [NotNullIfNotNull(nameof(getDefaultValue))] out T? value)
+        where T : notnull
     {
         if (result is { Result: { } ddResult, IsValid: true })
         {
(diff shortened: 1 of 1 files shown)


tracer/src/Datadog.Trace/Configuration/ConfigurationSources/CompositeConfigurationSource.cs
```diff
@@ -41,34 +37,18 @@ public CompositeConfigurationSource(IEnumerable<IConfigurationSource> sources)
         /// <param name="source">The configuration source to add.</param>
         public void Add(IConfigurationSource source)
         {
-            if (source == null) { ThrowHelper.ThrowArgumentNullException(nameof(source)); }
+            if (source == null!) { ThrowHelper.ThrowArgumentNullException(nameof(source)); }
```
anna-git, line 40: this `null!` is funny, why is it needed here 🤔

anna-git (approved):

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/CompositeConfigurationSource.cs
andrewlock: Yeah 😅 It's because the nullable reference types indicate that `source` _can't_ be `null`, so it's to stop a warning. Once we have null reference types everywhere, we can remove the arg check completely, but for now, it's safest to keep it! 😄

lucaspimentel (approved):

PR #7335, opened 2025-08-04
chore: clean up baggage span tags implementation

rachelyangdog:
## Summary of changes
- Removed identical AddBaggageTagsToSpan methods
- Added to SpanContextPropagator method AddBaggageToSpanAsTags with centralized exception handling and logging
- Replaced all method calls with direct calls to the propagator
- Optimized loop consolidation for better performance
- Original PR: #7020

### DIFF
--- a/tracer/src/Datadog.Trace/Propagators/SpanContextPropagator.cs
+++ b/tracer/src/Datadog.Trace/Propagators/SpanContextPropagator.cs
@@ -329,13 +329,13 @@ namespace Datadog.Trace.Propagators
             try
             {
                 var addAllItems = baggageTagKeys.Length == 1 && baggageTagKeys[0] == "*";
-                var keysToProcess = addAllItems ? baggage.Select(item => item.Key) : baggageTagKeys;
+                var allowedKeys = addAllItems ? null : new HashSet<string>(baggageTagKeys);
 
-                foreach (var key in keysToProcess)
+                foreach (var item in baggage)
                 {
-                    if (baggage.TryGetValue(key, out var value))
+                    if (addAllItems || allowedKeys!.Contains(item.Key))
                     {
-                        span.SetTag("baggage." + key, value);
+                        span.SetTag("baggage." + item.Key, item.Value);
                     }
                 }
             }
--- a/tracer/src/Datadog.Trace/Propagators/SpanContextPropagator.cs
+++ b/tracer/src/Datadog.Trace/Propagators/SpanContextPropagator.cs
@@ -10,6 +10,7 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using Datadog.Trace.Headers;
+using Datadog.Trace.Logging;
 using Datadog.Trace.Tagging;
 using Datadog.Trace.Util;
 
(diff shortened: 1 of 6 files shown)


tracer/src/Datadog.Trace/Propagators/SpanContextPropagator.cs
```diff
@@ -323,24 +326,22 @@ internal void AddBaggageToSpanAsTags(ISpan span, Baggage? baggage, string[] bagg
                 return;
             }
 
-            if (baggageTagKeys.Length == 1 && baggageTagKeys[0] == "*")
+            try
             {
-                // add all baggage items as tags
+                var addAllItems = baggageTagKeys.Length == 1 && baggageTagKeys[0] == "*";
+                var allowedKeys = addAllItems ? null : new HashSet<string>(baggageTagKeys);
```
lucaspimentel: Using `HashSet<string>` is a good idea to make the `Contains()` faster for larger sizes of `TracerSettings.BaggageTagKeys`, but one issue this introduces is that we're now creating a new `HashSet` in the hot path every time we call `AddBaggageToSpanAsTags()`. More heap allocations lead to more garbage collections, reducing overall throughput.

Another (smaller?) issue is that `HashSet` is probably only faster for larger `TracerSettings.BaggageTagKeys` collections, but a simple array can be faster for smaller ones. We _could_ try to get fancy and switch between `string[]` and `HashSet<string>` based on how many baggage items are configured in `DD_TRACE_BAGGAGE_TAG_KEYS`, but I don't think it's worth it. 😅 

I suggest either not using `HashSet<string>` at all and sticking to `string[]`, or changing the `TracerSettings.BaggageTagKeys` property from `string[]` to a `HashSet<string>` so that only one `HashSet` is ever created (during tracer initialization) and is re-used over and over here in `AddBaggageToSpanAsTags()`, avoiding the additional heap allocations.

zacharycmontoya: > I suggest either not using `HashSet<string>`, or changing the `TracerSettings.BaggageTagKeys` property from `string[]` to a `HashSet<string>`. That way, the `HashSet` is only created once (during tracer initialization) and is re-used over and over here in `AddBaggageToSpanAsTags()`, avoiding the additional heap allocations.

+1 changing the original property to a `HashSet<string>`, this should be sufficient for what we're trying to do without requiring more advanced microbenchmarking

lucaspimentel: (I made some edits to my original comment before Zach's reply, but still basically the same message 😅 )

zacharycmontoya: Doing that would also allow you to remove the `allowedKeys` variable entirely, as Lucas was saying since we want to avoid creating unneeded objects when we're running default / hot-path code

rachelyangdog: Thanks for the suggestions @lucaspimentel @zacharycmontoya! The changes have been pushed! Lmk if it's correct and if there's anything else I should update. Thank youu!

lucaspimentel: Thanks!

PR #7336, opened 2025-08-04
Update default value of DD_LOGS_INJECTION to true

rachelyangdog:
## Summary of changes
- updates the default value of DD_LOGS_INJECTION to be true
- this change will enable log injection for structured logging frameworks by default 

## Reason for change
- Phase 1 of Trace Log Correlation [RFC](https://docs.google.com/document/d/1Okl7hQcJgA9NdQ8msSVu5P2cPQOx139K9s9f3C_bTho/edit?tab=t.0)

### DIFF
--- a/tracer/src/Datadog.Trace/Logging/DirectSubmission/DirectLogSubmissionSettings.cs
+++ b/tracer/src/Datadog.Trace/Logging/DirectSubmission/DirectLogSubmissionSettings.cs
@@ -170,8 +170,8 @@ namespace Datadog.Trace.Logging.DirectSubmission
             ValidationErrors = validationErrors;
             IsEnabled = isEnabled;
 
-            // Logs injection is enabled by default if direct log submission is enabled, otherwise disabled by default
-            LogsInjectionEnabled = config.WithKeys(ConfigurationKeys.LogsInjectionEnabled).AsBool(defaultValue: isEnabled);
+            // Logs injection is enabled by default
+            LogsInjectionEnabled = config.WithKeys(ConfigurationKeys.LogsInjectionEnabled).AsBool(defaultValue: true);
         }
 
         /// <summary>


tracer/src/Datadog.Trace/Logging/DirectSubmission/DirectLogSubmissionSettings.cs
```diff
@@ -170,8 +170,8 @@ public DirectLogSubmissionSettings(IConfigurationSource? source, IConfigurationT
             ValidationErrors = validationErrors;
             IsEnabled = isEnabled;
 
-            // Logs injection is enabled by default if direct log submission is enabled, otherwise disabled by default
-            LogsInjectionEnabled = config.WithKeys(ConfigurationKeys.LogsInjectionEnabled).AsBool(defaultValue: isEnabled);
+            // Logs injection is enabled by default when not explicitly configured
```
mabdinur: nit: we can omit this detail
```suggestion
            // Logs injection is enabled by default
```

tracer/build/_build/Build.ExplorationTests.cs
```diff
@@ -166,7 +166,9 @@ Dictionary<string, string> GetEnvironmentVariables(ExplorationTestDescription te
         {
             ["DD_TRACE_LOG_DIRECTORY"] = TestLogsDirectory,
             ["DD_SERVICE"] = "exploration_tests",
-            ["DD_VERSION"] = Version
+            ["DD_VERSION"] = Version,
+            // Disable logs injection for exploration tests to avoid interfering with third-party test expectations
+            ["DD_LOGS_INJECTION"] = "false"
```
mabdinur, line 171: Should we add a test case to integrations to validate the default behavior?

mabdinur (approved):
Left some questions but I'll defer to dotnet maintainers. The behavior change looks good to me!

tracer/build/_build/Build.ExplorationTests.cs
rachelyangdog: Yup, this test` WhenLogsInjectionIsNotSetThenValueDefaultsToTrue` should validate the default behavior.

rachelyangdog: I can edit the above tests to get rid of LogSubmissionSettings assertions since they aren't needed anymore.

tracer/test/Datadog.Trace.Tests/Logging/DirectSubmission/DirectLogSubmissionSettingsTests.cs
```diff
@@ -392,7 +392,7 @@ public void WhenLogsInjectionIsExplicitlySetViaEnvironmentThenValueIsUsed(bool l
         [Theory]
         [InlineData(true)]
         [InlineData(false)]
-        public void WhenLogsInjectionIsNotSetThenValueIsSameAsLogSubmissionEnabled(bool directLogSubmissionEnabled)
+        public void WhenLogsInjectionIsNotSetThenValueDefaultsToTrue(bool directLogSubmissionEnabled)
```
zacharycmontoya: I think we can simply delete this test. With this PR's proposed changes, the two settings are now independent of each other so this test no longer adds value

tracer/src/Datadog.Trace/Logging/DirectSubmission/DirectLogSubmissionSettings.cs
```diff
@@ -170,8 +170,8 @@ public DirectLogSubmissionSettings(IConfigurationSource? source, IConfigurationT
...
+            LogsInjectionEnabled = config.WithKeys(ConfigurationKeys.LogsInjectionEnabled).AsBool(defaultValue: true);
```
zacharycmontoya: Since we're making these configurations independent now, can you move the config loading this from this class and place it in `tracer/src/Datadog.Trace/Configuration/TracerSettings.cs`? I think the best example you can follow is how we load the `_apmTracingEnabled` field

tracer/build/_build/Build.ExplorationTests.cs
zacharycmontoya, line 171: For integration tests, can you go through the folder `tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests` and delete all uses of  `SetEnvironmentVariable("DD_LOGS_INJECTION", "true");` since this should be on by default now?

Please do not update any instances of `SetEnvironmentVariable("DD_LOGS_INJECTION", "false");` since we are testing that the feature can be turned off.

tracer/test/Datadog.Trace.Tests/Logging/DirectSubmission/DirectLogSubmissionSettingsTests.cs
```diff
@@ -210,13 +210,13 @@ public void ApiKey(string value, string expected)
         }
 
         [Theory]
-        [MemberData(nameof(BooleanTestCases), false)]
+        [MemberData(nameof(BooleanTestCases), true)]
```
zacharycmontoya: Now that we have the setting back in the `TracerSettings` class, can you move this entire method to the `tracer/test/Datadog.Trace.Tests/Configuration/TracerSettingsTests.cs` file?

link04 (changes requested):
The exploration tests fail for Serilog and seems to be because we now inject the trace data into the log, not sure what the process is for adapting those test but to have them work as before we could add the env var with a value of `false` below and I imagine it would pass as before: https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/build/_build/Build.ExplorationTests.cs#L702-L711

tracer/build/_build/Build.ExplorationTests.cs
```diff
@@ -167,6 +167,8 @@ Dictionary<string, string> GetEnvironmentVariables(ExplorationTestDescription te
             ["DD_TRACE_LOG_DIRECTORY"] = TestLogsDirectory,
             ["DD_SERVICE"] = "exploration_tests",
             ["DD_VERSION"] = Version
```
zacharycmontoya:
```suggestion
            ["DD_VERSION"] = Version,
```

PR #7337, opened 2025-08-04
[CI] Creates PR Label Based Docker Image

link04:
## Summary of changes
Adding new stage to check if a label has been added to the PR for Azure to decide wether a Docker image should be created based on the PR or not(currently we only do it for master after each merge).

- Related system test PR where I tested the image being used correctly(a 4 lines change): 
https://github.com/DataDog/system-tests/pull/5024
- Location where the image gets used and all tests pass: 
https://github.com/DataDog/system-tests/actions/runs/16763951714/job/47465265591?pr=5024#step:7:72

## Reason for change
For us to be able to review/merge system-test PRs with higher confidence and remove the need to run locally for cases where the user may not want/need to.

## Implementation details
Updated the action, workflow and pipeline files needed to take into account the label and use the branch name instead.

## Test coverage
This is the test, the first commits add the docker image and tag it as expected: [Location](https://github.com/DataDog/dd-trace-dotnet/pkgs/container/dd-trace-dotnet%2Fdd-trace-dotnet)
The last commit makes sure the image gets deleted when this PR(with its label) closes, so the above should not have it anymore: [Location](https://github.com/DataDog/dd-trace-dotnet/pkgs/container/dd-trace-dotnet%2Fdd-trace-dotnet)

There could still be a scenario where the label gets removed and then the PR closed where the image could still remain but I doubt any of us would that...

### DIFF
--- a/.azure-pipelines/ultimate-pipeline.yml
+++ b/.azure-pipelines/ultimate-pipeline.yml
@@ -4161,9 +4161,18 @@ stages:
             AZURE_STORAGE_ACCOUNT: $(AZURE_STORAGE_ACCOUNT_NAME)
             AZURE_STORAGE_SAS_TOKEN: $(AZURE_STORAGE_SHARED_ACCESS_TOKEN)
 
-# Only run these on merges to master etc
+# Only run these on merges to master and PRs with the docker_image_artifacts label
 - stage: upload_container_images
-  condition: and(succeeded(), eq(variables['Build.Reason'], 'IndividualCI'), eq(variables.isMainOrReleaseBranch, true))
+  condition: |
+      or(
+        and( succeeded()
+          , eq( variables['Build.Reason'], 'IndividualCI')
+          , eq( variables['isMainOrReleaseBranch'], true)
+        ),
+        and( in( dependencies.upload_to_azure.result, 'Succeeded')
+          , eq( variables['Build.Reason'], 'PullRequest')
+        )
+      )
   dependsOn: [upload_to_azure]
   jobs:
     - job: upload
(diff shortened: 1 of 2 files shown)


.github/workflows/delete-pr-image.yml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+      - name: Sanitize branch name to match image tag
+        id: sanitize
+        run: |
+          BRANCH="${{ github.event.pull_request.head.ref }}"
+          TAG="${BRANCH//\//_}"
+          echo "tag=$TAG" >> "$GITHUB_OUTPUT"
+
+      - name: Delete image from GHCR
+        uses: actions/delete-package-versions@v4
```
andrewlock: nit: this should have a SHA specified. I think Dependabot will fix this for us though, so not a big deal

.azure-pipelines/ultimate-pipeline.yml
```diff
@@ -4183,6 +4183,39 @@ stages:
...
+  jobs:
+    - job: upload
+      timeoutInMinutes: 60
+      pool:
+        vmImage: ubuntu-latest
+      steps:
+      - checkout: none
+      - template: steps/create-github-app-token.yml
+      - bash: |
+          LABELS=$(curl -s -H "Authorization: token $GITHUB_TOKEN" \
```
andrewlock: ~~`$(System.PullRequest.PullRequestNumber)` won't have a value _unless_ it's running in the context of a PR. Have you tested that this works as expected? 🤔 I think we should gate all this logic behind a check that `$(System.PullRequest.PullRequestNumber)` has a value.~~ Just seen you have `eq(variables['Build.Reason'], 'PullRequest')` so that's probably good enough 👍

.azure-pipelines/ultimate-pipeline.yml
```diff
@@ -4183,6 +4183,39 @@ stages:
...
+            BRANCH_NAME="$(System.PullRequest.SourceBranch)"
+            echo "Creating dispatch event for https://github.com/DataDog/dd-trace-dotnet/actions/workflows/create-system-test-docker-base-images.yml with ref=$REF and azdo_build_id=$(Build.BuildId) and is_release_version=False"
+            curl \
+              -X POST \
+              -H "Accept: application/vnd.github+json" \
+              -H "Authorization: token $GITHUB_TOKEN"\
+              -H "X-GitHub-Api-Version: 2022-11-28" \
+              https://api.github.com/repos/DataDog/dd-trace-dotnet/actions/workflows/create-system-test-docker-base-images.yml/dispatches \
+              -d "{\"ref\":\"$REF\",\"inputs\":{\"azdo_build_id\":\"$(Build.BuildId)\",\"branch_name\":\"$BRANCH_NAME\"}}"
+          fi
```
andrewlock: _Technically_ this doesn't actually create a docker image with what you _might_ expect...

When we test in CI, we do a `git merge` between the tested branch and `master` in a PR (which is effectively what GitHub does). However, this will trigger the image to run using the `ref` which is the tip of the branch _not_ the merged version.

That said, I actually don't think it matters 😄 Seeing as GitHub is pulling the _build_ artifacts anyway, this seems fine 👍

.azure-pipelines/ultimate-pipeline.yml
```diff
@@ -4183,6 +4183,39 @@ stages:
         env:
           GITHUB_TOKEN: $(retrieve_github_token.GITHUB_APP_TOKEN)
 
+# New stage for PR Docker builds when label is present
+- stage: upload_container_images_pr
```
andrewlock: We already have `upload_container_images`, could we merge these two? 🤔 Seems like a lot of the logic is duplicated, might be nice to reduce that?

link04: Done in my last commit! https://github.com/DataDog/dd-trace-dotnet/pull/7337/commits/f15b5540c6b251db93baff1544ec252403e567f5

.github/workflows/delete-pr-image.yml
link04: Ty! Updated it in the last commit: https://github.com/DataDog/dd-trace-dotnet/pull/7337/commits/f15b5540c6b251db93baff1544ec252403e567f5 (this one I test when the PR closes and if it doesn't remove it will re-open the PR)

PR #7342, opened 2025-08-05
Include `telemetry_forwarder.exe` in the Windows SSI OCI images

andrewlock:
## Summary of changes

Downloads and verifies the telemetry_forwarder.exe binary and includes it in the OCI image

## Reason for change

We want to ship the binary side-by-side with the fleet installer. At some point we'll configure a proper release process for the `telemetry_forwarder.exe` dll, but in the mean time we're simply pulling from a public location for simplicity.

## Implementation details

- Download the telemetry forwarder from our Azure blob storage
- Sign the binary as part of our normal GitLab build
- Include the forwarder in our GitLab build output
- Set the `DD_TELEMETRY_FORWARDER_PATH` when installing the tracer

## Test coverage

Less than I'd like, but ultimately this is a non-critical component, that is tested independently. It would be nice to have some testing that the telemetry is forwarded correctly, but ultimately I think this should be added in the system-tests/onboarding tests repo. I included the telemetry forwarder in our smoke tests so we should at least get warnings/errors there if there's anything egregious. 

## Other details

Stacked on 
- https://github.com/DataDog/dd-trace-dotnet/pull/7341

### DIFF
--- a/tracer/build/_build/Build.Gitlab.cs
+++ b/tracer/build/_build/Build.Gitlab.cs
@@ -31,7 +31,7 @@ partial class Build
         {
             // Download the forwarder from Azure for now.
             // We will likely change this in the future, but it'll do for now.
-            const string url = "https://apmdotnetci.blob.core.windows.net/apm-datadog-win-ssi-telemetry-forwarder/telemetry_forwarder.exe";
+            const string url = "https://apmdotnetci.blob.core.windows.net/apm-datadog-win-ssi-telemetry-forwarder/c83ee9ad2f93c7314779051662e2e00086a213e0/telemetry_forwarder.exe";
             const string expectedHash = "0B192C1901C670FC9A55464AFDF39774AB7CD0D667ECFB37BC22C27184B49C37D4658383E021F792A2F0C7024E1091F35C3CAD046EC68871FAEEE3C98A40163A";
 
             var tempFile = await DownloadFile(url);
--- a/tracer/build/_build/Build.Gitlab.cs
+++ b/tracer/build/_build/Build.Gitlab.cs
@@ -34,7 +34,7 @@ partial class Build
             const string url = "https://apmdotnetci.blob.core.windows.net/apm-datadog-win-ssi-telemetry-forwarder/telemetry_forwarder.exe";
             const string expectedHash = "3A207302683E2BE8CB3D3A6AFF4F40721B7C85763A5A9B8F441D6BFD1B419F02D3EA8155634CEB3A09473F4745CE49679A55095F0EAD91F0582C6CB08C725782";
 
-            var tempFile = await TryDownloadForwarder();
+            var tempFile = await DownloadFile(url);
             var actualHash = GetSha512Hash(tempFile);
             if (!string.Equals(expectedHash, actualHash, StringComparison.Ordinal))
             {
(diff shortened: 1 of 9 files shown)


tracer/src/Datadog.FleetInstaller/PathHelper.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.IO;
+using System.Reflection;
+
+namespace Datadog.FleetInstaller;
+
+internal static class PathHelper
+{
+    private const string ForwarderFileName = "telemetry_forwarder.exe";
+
+    public static string GetTelemetryForwarderPath()
```
bmermet: Why not use the "--home-path" passed to the fleet installer commands to compute the path?

tracer/build/_build/Build.Gitlab.cs
```diff
@@ -19,6 +22,97 @@
...
+                throw new Exception($"Downloaded file did not have expected hash. Expected hash {expectedHash}, actual hash {actualHash}");
+            }
+
+            Logger.Information("Hash verified: '{Hash}'", expectedHash);
+
+            // Move to expected location
+            var output = ArtifactsDirectory / "telemetry_forwarder.exe";
+            FileSystemTasks.CopyFile(tempFile, output, FileExistsPolicy.Overwrite);
+
+            static async Task<string> TryDownloadForwarder()
```
zacharycmontoya: Both of these local methods seem like good utility methods to have in general, could we place them in a common location?

tracer/src/Datadog.FleetInstaller/PathHelper.cs
andrewlock: > Why not use the "--home-path" passed to the fleet installer commands to compute the path?

Because that's a different path 😄 --home-path is the path of all the tracer files, the telemetry_forwarder.exe lives next to the Datadog.FleetInstaller.exe. I really _don't_ want to put the telemetry_forwarder.exe inside the --home-path, because that's something that has a standard shape across all our deployments, and dumping a win-ssi file in there too just feels wrong 😅

bmermet: I was thinking more of deriving the location of the forwarder from the home path (something like "$HomePath/../installer/") rather than putting the forwarder in the home path. Anyway I don't think that makes much of a difference on the result.

andrewlock: Ah I see. Yeah, we could do that, as long as we never change the relative location of those components. I wasn't a fan of my current approach tbh 😄

andrewlock: Changed the approach in https://github.com/DataDog/dd-trace-dotnet/pull/7342/commits/2597a36ea5a53ae13a85170c92a36c0598dcb5bc

tracer/build/_build/Build.Gitlab.cs
andrewlock: Moved them to `Build.Utilities` in https://github.com/DataDog/dd-trace-dotnet/pull/7342/commits/a1891190e6d2df17c3866803c225f16257254999 we could move them somewhere else even more general, not fussed 🤷‍♂️

tracer/build/_build/Build.Gitlab.cs
```diff
@@ -19,6 +22,32 @@
...
+       .Description("Downloads the telemetry forwarder executable used by SSI ")
+       .Unlisted()
+       .Requires(() => IsWin)
+       .Before(SignDlls)
+       .Executes(async () =>
+        {
+            // Download the forwarder from Azure for now.
+            // We will likely change this in the future, but it'll do for now.
+            const string url = "https://apmdotnetci.blob.core.windows.net/apm-datadog-win-ssi-telemetry-forwarder/telemetry_forwarder.exe";
+            const string expectedHash = "3A207302683E2BE8CB3D3A6AFF4F40721B7C85763A5A9B8F441D6BFD1B419F02D3EA8155634CEB3A09473F4745CE49679A55095F0EAD91F0582C6CB08C725782";
```
bmermet: This is the previous version hash. The build fails so the validation works 👌

tracer/build/_build/Build.Gitlab.cs
```diff
@@ -19,6 +22,32 @@
...
+    Target DownloadWinSsiTelemetryForwarder => _ => _
+       .Description("Downloads the telemetry forwarder executable used by SSI ")
+       .Unlisted()
+       .Requires(() => IsWin)
+       .Before(SignDlls)
+       .Executes(async () =>
+        {
+            // Download the forwarder from Azure for now.
+            // We will likely change this in the future, but it'll do for now.
+            const string url = "https://apmdotnetci.blob.core.windows.net/apm-datadog-win-ssi-telemetry-forwarder/telemetry_forwarder.exe";
```
bmermet: We don't need to do do it now, but we'll need to version this path by commit or version number when we'll update the telemetry_forwarder.

bmermet (approved):

tracer/build/_build/Build.Gitlab.cs
andrewlock: Done in https://github.com/DataDog/dd-trace-dotnet/pull/7342/commits/938414821b5ac41c98e605ba722252b1834c5d2d

PR #7345, opened 2025-08-05
Fix `TracerSettings` OTel config reporting order

andrewlock:
## Summary of changes

Ensures we read the DD config values _before_ the OTel values

## Reason for change

When we read config, we report it to telemetry. Config read _later_ takes precedence (in telemetry) to config read _earlier_. OTel config that _overrides_ Datadog config values should therefore be read _after_ the datadog config is read.

## Implementation details

Change the order from 
- Read OTel config
- Read DD config
- Calculate overrides

to
- Read DD config
- Read OTel config
- Calculate overrides

## Test coverage

Should be covered by existing tests (and not easy to test for explicitly). We will just need to catch it in review I think. We may able to redesign the `ConfigurationBuilder` APIs to force this, but I think this is good enough for now.

## Other details

Discovered while working on other config stuff


zacharycmontoya:
Hey Andrew, the current behavior when both an OpenTelemetry configuration and Datadog configuration apply to a setting, we prefer the Datadog one, so it seems like the telemetry order is correct and there's nothing to do here. Perhaps it's the `OverrideWith(in otelConfig)` method that's misleading?

However, in the future we do plan to swap the precedence such that OpenTelemetry configurations take precedence as the final value, so I wonder if this may be the time to try and introduce an ergonomic toggle for which one is read last / set as the final value? This is something I can take on later though if you have better things to do

andrewlock:
> Hey Andrew, the current behavior when both an OpenTelemetry configuration and Datadog configuration apply to a setting, we prefer the Datadog one, so it seems like the telemetry order is correct and there's nothing to do here.

🤦‍♂️ You're right @zacharycmontoya, thanks!

> Perhaps it's the `OverrideWith(in otelConfig)` method that's misleading?

Yeah, I was reading that as "override the datadog value with the otel value", which isn't really what it's doing 🤔 
 
> However, in the future we do plan to swap the precedence such that OpenTelemetry configurations take precedence as the final value, so I wonder if this may be the time to try and introduce an ergonomic toggle for which one is read last / set as the final value? This is something I can take on later though if you have better things to do


Meh, I'm not sure if it's worth having this configurable 🤔 It _is_ difficult to understand the requirements (i.e. what I _thought_ was necessary in this PR) but I think that could require quite a bit of refactoring work to pull off 🤔 I don't have a feel for what it should look like tbh.

I think for now we can probably just close this PR, and then either revive it when you need to do the swap (cough, breaking change 😉 ) or work on a toggle then?

andrewlock:
Closing for now because Zach pointed out that it's _already_ correct 😅

PR #7346, opened 2025-08-05
Add `DOTNET_ROOT` values for windows integration tests

andrewlock:
## Summary of changes

Add  `DOTNET_ROOT` values for Windows integration tests

## Reason for change

We think that missing these values is causing failures in the .NET 10 update branch. These _should_ be set in the VMs, but until they are, set them here for now. If merging this to .NET 10 solves the issue, we know the VM fix will too.

## Implementation details

Add the variables [that control .NET lookup-paths](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables#net-sdk-and-cli-environment-variables)

## Test coverage

If the tests pass (and pass in the .NET 10 branch too) then we can merge it. Alternatively, we could _just_ include this in the .NET 10 branch, and update the VMs


andrewlock:
Set in the VMs directly now anyway

PR #7347, opened 2025-08-05
Include the target framework in our `InstrumentationTests`

andrewlock:
## Summary of changes

Make sure we always specify the target framework when creating a test console app

## Reason for change

We have some tests where we explicitly create, publish, and run a .NET Console app. The behaviour we're looking for is dependent on the target framework, however previously we _weren't_ specifying the TFM to use. That meant that it was always using the SDK version, even though that is not expected by the tests (and breaks in some scenarios, e.g. when using preview SDK builds).

## Implementation details

Explicitly set the target framework to use when creating the console app

## Test coverage

If the tests pass, we're fine.

## Other details

We currently have workarounds in the .NET 10 branch for this; if this is merged, we can revert those changes.


bouwkast:
Hmm yeah same issue, but for .NET 9.0 😕 
```
17:07:25 [DBG]  Error: Invalid option(s):
17:07:25 [DBG]  -f net8.0
17:07:25 [DBG]     'net8.0' is not a valid value for -f. The possible values are:
17:07:25 [DBG]        net9.0   - Target net9.0
```

bouwkast (approved):

PR #7350, opened 2025-08-06
Update the configuration key for controlling managed activation

andrewlock:
## Summary of changes

Update the configuration key for controlling managed activation that was added in #7303

## Reason for change

Other keys are `DD_PROFILING_` so should be consistent

## Implementation details

`DD_PROFILER_` => ``DD_PROFILING_`

## Test coverage

N/A


chrisnas (approved):
LGTM

PR #7351, opened 2025-08-06
Do not inject DSM legacy headers by default

kr-igor:
## Summary of changes
Reversed the default state of `IsDataStreamsLegacyHeadersEnabled` for `enabled by default` state. 

## Reason for change
See [ticket](https://datadoghq.atlassian.net/browse/APMS-16734)


tracer/output.txt
andrewlock, line 1: Pretty sure this shouldn't be here 😄

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -662,7 +662,7 @@ internal TracerSettings(IConfigurationSource? source, IConfigurationTelemetry te
 
             IsDataStreamsLegacyHeadersEnabled = config
                                                .WithKeys(ConfigurationKeys.DataStreamsMonitoring.LegacyHeadersEnabled)
-                                               .AsBool(true);
+                                               .AsBool(false);
```
```suggestion
                                               .AsBool( !_isDataStreamsMonitoringInDefaultState);
```

bouwkast (approved):
Thanks!

I think some tests will need to be updated to accommodate this as they are failing

PR #7354, opened 2025-08-06
Refactor some telemetry 

andrewlock:
## Summary of changes

Refactor `ConfigurationBuilder`

## Reason for change

As part of some upcoming work, we need to record additional telemetry about default values (so that these can be exposed to customers). As an indirect consequence, it's preferable to provide an explicit default value instead of a `Func<T>`, so this refactoring is _mostly_ using `T` instead of `Func<T>` where it makes sense.

## Implementation details

Mostly standard refactoring stuff. Added some overloads, inlined some methods, moved some other things around. Nothing drastic.

One major difference is that `DefaultResult<T>` now _requires_ that you provide an explicit associated "telemetry value" for recording, whereas it was previously optional. However, we always provided it in practice, so this is a noop.

## Test coverage

Should all be covered by existing tests.

## Other details

### DIFF
--- a/tracer/src/Datadog.Trace/Configuration/ConfigurationSources/Telemetry/ConfigurationBuilder.cs
+++ b/tracer/src/Datadog.Trace/Configuration/ConfigurationSources/Telemetry/ConfigurationBuilder.cs
@@ -159,17 +159,17 @@ internal readonly struct ConfigurationBuilder(IConfigurationSource source, IConf
         // String accessors
         // ****************
         public string? AsRedactedString()
-            => AsString(getDefaultValue: null, validator: null, converter: null, recordValue: false);
+            => AsString(defaultValue: null, validator: null, recordValue: false);
 
         public string AsRedactedString(string defaultValue)
-            => AsString(() => defaultValue, validator: null, converter: null, recordValue: false);
+            => AsString(defaultValue, validator: null, recordValue: false);
 
         /// <summary>
         /// Beware, this function won't record telemetry if the config isn't explicitly set.
         /// If you can, use <see cref="AsString(string)"/> instead or record telemetry manually.
         /// </summary>
         /// <returns>the string value of the configuration if set</returns>
-        public string? AsString() => AsString(getDefaultValue: null, validator: null, converter: null, recordValue: true);
+        public string? AsString() => AsString(defaultValue: null, validator: null, recordValue: true);
 
         public string AsString(string defaultValue) => AsString(defaultValue, validator: null);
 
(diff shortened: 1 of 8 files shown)


tracer/test/Datadog.Trace.Tests/Configuration/Telemetry/ConfigurationBuilderTests.cs
andrewlock, line 1: The changes in this file are a consequence of making the `DefaultResult<T>` telemetryValue required

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/Telemetry/ConfigurationBuilder.cs
```diff
@@ -12,44 +12,10 @@
 
 namespace Datadog.Trace.Configuration.Telemetry;
 
-internal readonly struct ConfigurationBuilder
+internal readonly struct ConfigurationBuilder(IConfigurationSource source, IConfigurationTelemetry telemetry)
 {
-    // static accessor functions
```
andrewlock, line 17: These all just moved to the bottom of the class, out of the way!

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/Telemetry/ConfigurationBuilder.cs
```diff
@@ -59,41 +25,13 @@ public ConfigurationBuilder(IConfigurationSource source, IConfigurationTelemetry
 
     public HasKeys WithKeys(string key, string fallbackKey1, string fallbackKey2, string fallbackKey3) => new(_source, _telemetry, key, fallbackKey1, fallbackKey2, fallbackKey3);
 
-    private static bool TryHandleResult<T>(
```
andrewlock, line 62: This was inlined as it was only ultimately used in 1 place, and is much more complicated to understand than is worth it

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/Telemetry/ConfigurationBuilder.cs
```diff
@@ -103,8 +41,14 @@ private static void RecordTelemetry<T>(IConfigurationTelemetry telemetry, string
...
                 telemetry.Record(key, boolVal, ConfigurationOrigins.Default);
                 break;
+            case string stringVal:
+                telemetry.Record(key, stringVal, recordValue, ConfigurationOrigins.Default);
+                break;
+            case null: // can't actually be called in practice
+                break;
             default:
-                telemetry.Record(key, defaultValue.TelemetryValue, recordValue, ConfigurationOrigins.Default);
+                // TODO: this shouldn't be calleable in practice, we need to revise it
```
andrewlock, line 50: Would require a bunch more refactoring and copy-pasting, so will handle in a follow up (if at all)

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/Telemetry/ConfigurationBuilder.cs
```diff
@@ -137,17 +81,17 @@ public HasKeys(IConfigurationSource source, IConfigurationTelemetry telemetry, s
         // String accessors
         // ****************
         public string? AsRedactedString()
-            => AsString(getDefaultValue: null, validator: null, converter: null, recordValue: false);
+            => AsString(defaultValue: null, validator: null, recordValue: false);
 
         public string AsRedactedString(string defaultValue)
-            => AsString(() => defaultValue, validator: null, converter: null, recordValue: false);
+            => AsString(defaultValue, validator: null, recordValue: false);
```
andrewlock, line 87: This is the _main_ "point" of this PR. Stop creating all these extras `Func<T>` when we don't have to

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/Telemetry/ConfigurationBuilder.cs
```diff
@@ -156,80 +100,156 @@ public string AsRedactedString(string defaultValue)
...
+        public string? AsString(Func<string>? getDefaultValue, Func<string, bool>? validator, Func<string, ParsingResult<string>> converter)
             => AsString(getDefaultValue, validator, converter, recordValue: true);
 
         [return: NotNullIfNotNull(nameof(getDefaultValue))]
-        private string? AsString(Func<DefaultResult<string>>? getDefaultValue, Func<string, bool>? validator, bool recordValue)
+        private string? AsString(Func<string>? getDefaultValue, Func<string, bool>? validator, bool recordValue)
             => AsString(getDefaultValue, validator, converter: null, recordValue);
 
+        [return: NotNullIfNotNull(nameof(defaultValue))]
+        private string? AsString(string? defaultValue, Func<string, bool>? validator, bool recordValue)
```
andrewlock, line 121: The difference between these two methods is the first is `T? defaultValue` and the second is `Func<T>? defaultValue`. We should favour the former where possible

PR #7355, opened 2025-08-06
Add Nuke steps to generate simplified "supported" integrations CSVs

bouwkast:
## Summary of changes

Adds a Nuke step `tracer/build.ps1 GenerateSupportedIntegrationsCsv` to generate three CSVs around our supported integrations

It generates three CSVs: one for .NET Framework, one for .NET Core, one as a combination.

_I have not validated the accuracy of the results but seems close enough to start looking at internally and build on IMO_
## Reason for change

IDM team is now interested in collating data across tracing libraries of what various libraries are supported in various tracers.

## Implementation details

- `Build.SupportedIntegrationsTable` Nuke target to generate the CSVs
- `SupportedIntegrationsTableGenerator` class that does the actual conversion of `supported_versions.json` into the CSVs.

Note that this is run locally.

## Test coverage

Ran it locally seems fine.

## Other details

This is NOT prescriptive in that these generated CSVs aren't the source of truth.
This was something I messed around with for a day during the last R&D week to see how to use various LLM coding assistants mainly looking into parsing our `supported_versions.json` file into something a bit more consumable.

A better approach that I also experimented with was based off our generated calltargets and a Roslyn analyzer to go through our various attributes to pull out _actual_ source of truth data (and can include other bits like what methods we actually instrument) which seems more useful, but probably too much information at the moment.

All of this information is basically pulled from `supported_versions.json` but easier this way to disseminate the information internally without caring about "Assembly" vs "NuGet" (something that must be considered though long term).


bouwkast:
Sounds like we want to go a different route so closing this as not needed

We still have the supported_versions.json file that contains information that we can leverage and/or our generated calltargets file

PR #7356, opened 2025-08-07
[AAP] Added WAF metrics. Reorganized other metrics to match RFC

daniel-romano-DD:
## Summary of changes
Implemented some WAF missing metrics, as `WafTimeouts`
Reorganized existing metrics which were defined as Tags instead of Metrics, matching the definition in the RFC

## Reason for change
WAF metrics [RFC](https://docs.google.com/document/d/1D4hkC0jwwUyeo0hEQgyKP54kM1LZU98GL8MaP60tQrA/edit?tab=t.0)

### DIFF
--- a/tracer/src/Datadog.Trace/AppSec/AppSecRequestContext.cs
+++ b/tracer/src/Datadog.Trace/AppSec/AppSecRequestContext.cs
@@ -30,7 +30,7 @@ internal partial class AppSecRequestContext
     private int? _wafRaspError = null;
     private Dictionary<string, List<Dictionary<string, object>>>? _raspStackTraces;
 
-    internal void CloseWebSpan(TraceTagCollection tags, Span span)
+    internal void CloseWebSpan(Span span)
     {
         lock (_sync)
         {
@@ -45,7 +45,7 @@ internal partial class AppSecRequestContext
                 else
                 {
                     var triggers = JsonConvert.SerializeObject(_wafSecurityEvents);
-                    tags.SetTag(Tags.AppSecJson, "{\"triggers\":" + triggers + "}");
+                    span.Tags.SetTag(Tags.AppSecJson, "{\"triggers\":" + triggers + "}");
                 }
             }
 
--- a/tracer/src/Datadog.Trace/AppSec/AppSecRequestContext.cs
+++ b/tracer/src/Datadog.Trace/AppSec/AppSecRequestContext.cs
@@ -6,6 +6,7 @@
 #nullable enable
 
 using System.Collections.Generic;
+using System.Threading;
 using Datadog.Trace.AppSec.Rasp;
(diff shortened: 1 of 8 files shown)


e-n-0 (approved):
Thanks 😄

tracer/test/Datadog.Trace.Security.Unit.Tests/RASP/RaspTelemetryHelperTests.cs
andrewlock, line 1: nit: this should be renamed to `RaspMetricsHelperTests` 🙂

tracer/src/Datadog.Trace/AppSec/AppSecRequestContext.cs
```diff
@@ -47,27 +49,38 @@ internal void CloseWebSpan(TraceTagCollection tags, Span span)
...
+            _raspMetricsHelper?.GenerateRaspSpanMetricTags(span.Tags);
         }
     }
 
-    internal void CheckWAFError(int code, bool isRasp)
+    internal void CheckWAFError(IResult result, bool isRasp)
     {
+        if (result.Timeout)
+        {
+            Interlocked.Increment(ref _wafTimeout);
```
andrewlock: I'm a little unsure about the concurrency expectations in this class - the interlocked here implies it's meant to be able to be called from multiple threads, but basically nothing else in the type is thread safe as far as I can tell 😅

tracer/src/Datadog.Trace/AppSec/AppSecRequestContext.cs
```diff
@@ -47,27 +49,38 @@ internal void CloseWebSpan(TraceTagCollection tags, Span span)
                 }
             }
 
-            if (_raspStackTraces?.Count > 0)
+            if (_wafTimeout > 0)
             {
-                span.SetMetaStruct(StackKey, MetaStructHelper.ObjectToByteArray(_raspStackTraces));
+                span.Tags.SetMetric(Metrics.WafTimeouts, _wafTimeout);
```
andrewlock, line 54: Is there a reason you're calling `span.Tags` and not using the `tags` object provided? 🤔  How do you know which to use?

andrewlock (approved):

tracer/src/Datadog.Trace/AppSec/AppSecRequestContext.cs
dromanol: Fixed, so all tags go in the root span

tracer/src/Datadog.Trace/AppSec/AppSecRequestContext.cs
dromanol: Better safe than sorry, but you are right

PR #7358, opened 2025-08-07
Update libdatadog to 20.0.0

hoolioh:


tracer/src/Datadog.Trace/LibDatadog/TraceExporter.cs
```diff
@@ -141,14 +124,22 @@ private unsafe void ProcessResponse(IntPtr response)
         }
 
         // TODO: replace GetBodyLen with a native function in order to avoid iterating over the response to get its length.
-        var body = NativeInterop.Exporter.GetResponseBody(response);
-        var len = GetBodyLen(body);
-        if (len <= 0)
+        var len = UIntPtr.Zero;
+        var body = NativeInterop.Exporter.GetResponseBody(response, ref len);
+        var bodyLen = (ulong)len;
+        if (body != IntPtr.Zero || bodyLen == 0)
```
andrewlock: I think this is a typo? 😄 
```suggestion
        if (body == IntPtr.Zero || bodyLen == 0)
```

hoolioh: Good catch

tracer/src/Datadog.Trace/LibDatadog/TraceExporter.cs
```diff
@@ -141,14 +124,22 @@ private unsafe void ProcessResponse(IntPtr response)
         }
 
         // TODO: replace GetBodyLen with a native function in order to avoid iterating over the response to get its length.
```
bouwkast: nit: looks like this can be removed now?
```suggestion
```

PR #7360, opened 2025-08-07
Implement "enahanced telemetry reporting"

andrewlock:
## Summary of changes

Implements "enhance telemetry reporting" for .NET

## Reason for change

We want to report the telemetry found from _all_ sources, to implement [this RFC](https://docs.google.com/document/d/1vhIimn2vt4tDRSxsHn6vWSc8zYHl0Lv0Fk7CQps04C4/edit?tab=t.0#heading=h.7tw832mj0ztp), so that customers can more easily understand where their telemetry comes from.

## Implementation details

The .NET implementation was already pretty close to the desired result, so there were just some main changes to make

- Read sources in lowest to highest priority, and always report all found values.
  - This differs from previously, where we would read from highest to lowest and stop on first found values.
  - In order to ensure that the "last" reported telemetry value is the _actual_ value we use, we have to record the telemetry value _twice_ in cases where we have a parse failure (which should be rare in practice).
  - To do this properly, we had to explicitly expose the `Origin` on `IConfigurationSource` and captuate the "value for telemetry" where required (for non-simple types).
- Report the "default" value to be used, even if it's going to be overridden by an explicit value.
  - This is in keeping with the general "report all sources" principle.
  - We have to do the same "fixing" of values when we have parsing errors as above.
  - To avoid expensive unnecessary computation where a default value is not "simple" we go against this approach when a `Func<T>` is provided for calculating the default.
- For "fallback" keys, we just treat them the same as we do today. The backend will handle merging the keys

## Test coverage

Mostly covered by existing tests, and added/updated unit tests to meet new expectations

## Other details

Spec: https://docs.google.com/document/d/1vhIimn2vt4tDRSxsHn6vWSc8zYHl0Lv0Fk7CQps04C4/edit?tab=t.0#heading=h.7tw832mj0ztp

Dependent (stacked) on
- https://github.com/DataDog/dd-trace-dotnet/pull/7327
- https://github.com/DataDog/dd-trace-dotnet/pull/7354
- https://github.com/DataDog/dd-trace-dotnet/pull/7345

### DIFF
--- a/tracer/src/Datadog.Trace/Configuration/ConfigurationSources/HandsOffConfigurationSource.cs
+++ b/tracer/src/Datadog.Trace/Configuration/ConfigurationSources/HandsOffConfigurationSource.cs
@@ -17,7 +17,7 @@ internal sealed class HandsOffConfigurationSource(Dictionary<string, string> con
     private readonly Dictionary<string, string> _configurations = configurations;
     private readonly bool _localFile = localFile;
 
-    internal override ConfigurationOrigins Origin => _localFile ? ConfigurationOrigins.LocalStableConfig : ConfigurationOrigins.FleetStableConfig;
+    public override ConfigurationOrigins Origin => _localFile ? ConfigurationOrigins.LocalStableConfig : ConfigurationOrigins.FleetStableConfig;
 
     protected override string? GetString(string key)
     {


tracer/src/Datadog.Trace/Configuration/ConfigurationSources/CompositeConfigurationSource.cs
```diff
@@ -51,50 +53,260 @@ public void Add(IConfigurationSource source)
 
         /// <inheritdoc />
         public ConfigurationResult<string> GetString(string key, IConfigurationTelemetry telemetry, Func<string, bool>? validator, bool recordValue)
-            => _sources
-              .Select(source => source.GetString(key, telemetry, validator, recordValue))
-              .FirstOrDefault(value => value.IsValid, ConfigurationResult<string>.NotFound());
+        {
```
dromanol, line 56: The only difference in all these `GetXXX` is the value retrieving function, which could be a lambda, to avoid the n-plicated code

andrewlock: Yeah, I've gone back and forth on the lambda thing. I extracted these to lambdas in `ConfigurationBuilder` and it makes everything seem that much more complicated, that I'm really not sure which is best. Add to that the fact that there's _slightly_ different behaviour between each of the cases and it ends up really not being clear-cut which approach is better tbh.

PR #7365, opened 2025-08-07
Protect against `NullReferenceExceptions` in AWS S3 Integration

### LINKED ISSUE #7364: Profiler fails with some S3 Errors
When trying to trace an endpoint, we had an error uploading an S3 file (permissions) and it caused the endpoint trace not to appear in datadog at all. Turning on the trace debugging, I found this in the logs:

```
025-08-07 14:08:28.275 -04:00 [ERR] Exception occurred when calling the CallTarget integration continuation. System.NullReferenceException: Object reference not set to an instance of an object.
   at AWSSDK_S3__885C28607F98E604.Amazon_S3_Model_PutObjectResponse.Datadog_Trace_ClrProfiler_AutoInstrumentation_AWS_S3_IS3EtagResponse_30.get_ETag()
   at Datadog.Trace.ClrProfiler.AutoInstrumentation.AWS.S3.ObjectManagement.PutObjectAsyncIntegration.OnAsyncMethodEnd[TTarget,TReturn](TTarget instance, TReturn returnValue, Exception exception, CallTargetState& state) in /_/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/S3/ObjectManagement/PutObjectAsyncIntegration.cs:line 54
   at PutObjectAsyncIntegration.OnAsyncMethodEnd.AmazonS3Client.PutObjectResponse(AmazonS3Client, PutObjectResponse, Exception, CallTargetState&)
   at Datadog.Trace.ClrProfiler.CallTarget.Handlers.Continuations.TaskContinuationGenerator`4.SyncCallbackHandler.ContinuationAction(Task`1 previousTask, TTarget target, CallTargetState state)
 { MachineName: ".", Process: "[1209 Seeds.EntryPoin]", AppDomain: "[1 Seeds.EntryPoints.Api]", AssemblyLoadContext: "\"\" Datadog.Trace.ClrProfiler.Managed.Loader.ManagedProfilerAssemblyLoadContext #1", TracerVersion: "3.23.0.0" }
```

It looks like returnValue is a nullable object in the method signature below, but there's no guard against a null reference in the method itself, and that's what I seem to be running into.

`https://github.com/DataDog/dd-trace-dotnet/blame/015d21801674f10d8d72c17276723ec9c322df69/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/S3/ObjectManagement/PutObjectAsyncIntegration.cs#L54`

bouwkast:
## Summary of changes

Protect against `NullReferenceException` in the S3 integration by checking `returnValue?.Instance is not null` instead of `returnValue is not null`

## Reason for change

`returnValue` won't be null, but the `Instance` may be, so should check that.

## Implementation details

Added `?.Instance` to checks.

## Test coverage

I manually tested this by creating a S3 client pointing to a non-existent endpoint and saw the `NullReferenceException`. After making the changes and rebuilding the exception went away and instead a span was created.

_However I didn't commit this and only tested it for the `PutObjectAsyncIntegration`_. Let me know if you think we should update tests to check for this. 

## Other details

Fixes #7364


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/S3/MultipartUploadManagement/CompleteMultipartUploadAsyncIntegration.cs
```diff
@@ -47,7 +47,7 @@ internal static CallTargetState OnMethodBegin<TTarget, TRequest>(TTarget instanc
     internal static TReturn? OnAsyncMethodEnd<TTarget, TReturn>(TTarget instance, TReturn? returnValue, Exception exception, in CallTargetState state)
```
andrewlock, line 47: I'm thinking that TReturn should be non-nullable in this case too, seeing as it can never be null🤔

andrewlock (approved):
Thanks! I'm a bit torn on the testing, but I think it would be preferable to include it if it's not too much hassle

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/S3/MultipartUploadManagement/CompleteMultipartUploadAsyncIntegration.cs
bouwkast: I was wondering that too, but wasn't 100% sure 

I'll swap it to be non-nullable while adding test cases

bouwkast: Changed the `TReturn?` in the params of the files I touched to be `TReturn` and also added some tests that would throw the exception by pointing to an invalid endpoint.

Issue at the moment is the tests are quite slow so I'll follow up on that in a next PR

And for some reason throwing that caused the .NET Framework sample to hang 😭 so disabled those, will follow up on those as well

bouwkast: Okay they don't hang, but they end up taking ~8minutes per test run so something to do with the timeouts

I'll plan on reducing the test cases for the invalid endpoints for a PR next week

bouwkast:
Fixed in this PR

https://github.com/DataDog/dd-trace-dotnet/pull/7448

The tests in this PR did show the that there was an issue, but they were very slow and honestly I think provide not a lot of value, so opting to close this for that one.

PR #7366, opened 2025-08-07
[Debugger Default-On] DEBUG-3322 Debugger in-product enablement

dudikeleti:
## Summary of changes
This PR adds remote enablement support for debugger products.
depends on https://github.com/DataDog/dd-trace-dotnet/pull/7441

## Reason for change
Currently, the only option to enable DI (for example) is by setting an environment variable. We want to give customers the ability to enable products on demand.

## Implementation details
Gets RC events with new configuration and update the products based on defined rules.

## Test coverage
All existing tests
New system tests (https://github.com/DataDog/system-tests/pull/4991)
DebuggerManagerTests
DebuggerManagerDynamicTests


tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
```diff
@@ -445,18 +445,16 @@ private static void StartDiagnosticManager()
...
-                });
+            DebuggerManager.Instance.UpdateConfiguration(tracerSettings)
+                           .ContinueWith(
+                                t =>
+                                {
+                                    if (t is { Exception: not null })
+                                    {
+                                        Log.Error(t.Exception.Flatten(), "Error initializing debugger");
+                                    }
+                                },
```
andrewlock: nit: this will only be called if there's an exception, and the api takes a nullable anyway, so can simplify. Also, I don't think we need to call flatten for the logging, it should still include all the inner exceptions? 🤔 

```suggestion
                                t => Log.Error(t?.Exception, "Error initializing debugger"),
```

tracer/src/Datadog.Trace/Configuration/DynamicConfigurationManager.cs
```diff
@@ -91,27 +99,68 @@ private static void OnConfigurationChanged(ConfigurationBuilder settings)
...
+
+            DebuggerManager.Instance.UpdateConfiguration(newSettings, newDebuggerSettings)
+                           .ContinueWith(
+                                t =>
+                                {
+                                    if (t is { Exception: not null })
+                                    {
+                                        Log.Error(t.Exception.Flatten(), "Error updating dynamic configuration for debugger");
+                                    }
+                                },
```
andrewlock: nit: we can simplify
```suggestion
                                t => Log.Error(t?.Exception, "Error updating dynamic configuration for debugger"),
```

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -67,32 +70,60 @@ private DebuggerManager(DebuggerSettings debuggerSettings, ExceptionReplaySettin
 
         internal string ServiceName { get; private set; }
 
-        private async Task<bool> WaitForDiscoveryServiceAsync(IDiscoveryService discoveryService, Task shutdownTask)
+        private string GetServiceName(TracerSettings tracerSettings)
```
andrewlock, line 87: I know it existed before, but the fact that it _does_ is more evidence that our layering here is just wrong 😢 I still think we need to refactor this significantly to accept the fact that there's a dependence of debugger on tracing. It shouldn't be intiailizing independently like this, as it's a recipe for mismatch. All for later of course, but it's becoming clearer and clearer

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -67,32 +70,60 @@ private DebuggerManager(DebuggerSettings debuggerSettings, ExceptionReplaySettin
...
         {
-            if (_discoveryServiceReady)
+            try
             {
-                return true;
+                return TraceUtil.NormalizeTag(tracerSettings.ServiceName ?? TracerManager.Instance.DefaultServiceName);
```
andrewlock, line 91: The `TracerManager.Instance` access is a big issue frankly. It could refer to a completely different `TracerSettings` object for example. Essentially, this `DebuggerManager` should be created as part of the `TracerManager` creation, and then it can just pass in the values you need. Again, something for a different PR that we can work on.

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -102,45 +133,152 @@ private void SetGeneralConfig(TracerSettings tracerSettings, DebuggerSettings se
...
+        }
+
+        private Task UpdateProductsState(TracerSettings tracerSettings, DebuggerSettings newDebuggerSettings)
+        {
+            if (_processExit.Task.IsCompleted)
+            {
+                return _processExit.Task;
+            }
+
+            OneTimeSetup(tracerSettings);
```
andrewlock, line 160: I have some suspicions that that is the source of the extra overhead we saw coming from the refactor. Essentially you're doing a bunch of one-time setup, on the startup hot path, even if the debugger is never used. I think we should try to do this work lazily (I started working on a PR to do that)

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -102,45 +133,152 @@ private void SetGeneralConfig(TracerSettings tracerSettings, DebuggerSettings se
...
             {
-                return TraceUtil.NormalizeTag(tracerSettings.ServiceName ?? TracerManager.Instance.DefaultServiceName);
+                if (_processExit.Task.IsCompleted || !DebuggerSettings.SymbolDatabaseUploadEnabled)
+                {
+                    return;
+                }
+
+                if (!tracerSettings.IsRemoteConfigurationAvailable)
+                {
+                    Log.Information("Remote configuration is not available in this environment, so we don't upload symbols.");
```
andrewlock: This seems noisy, it's written to all the logs, every time regardless of whether a customer is ever going to enable the debugger
```suggestion
                    Log.Debug("Remote configuration is not available in this environment, so we don't upload symbols.");
```

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -102,45 +133,152 @@ private void SetGeneralConfig(TracerSettings tracerSettings, DebuggerSettings se
...
+                }
+
+                if (CodeOrigin != null)
+                {
+                    Log.Debug("Code Origin for Spans is already initialized");
+                    return;
                 }
+
+                CodeOrigin = new SpanCodeOrigin.SpanCodeOrigin(debuggerSettings);
+                tracerSettings.Telemetry.Record(ConfigurationKeys.Debugger.CodeOriginForSpansEnabled, true, debuggerSettings.DynamicSettings.CodeOriginEnabled == true ? ConfigurationOrigins.RemoteConfig : ConfigurationOrigins.AppConfig);
```
andrewlock: A couple of things here:
- This telemetry will never be recorded. You should not change the `tracerSettings.Telemetry` object outside of `TracerSettings` itself _unless_ you're doing so as part of the startup path (which we definitely should).
- The use of `ConfigurationOrigins.AppConfig` is incorrect here - that is referring to _specifically_ to the web.config configuration source, whereas the value could be coming from any of the one 10 different sources we use 😄 

The _good_ news, is that you shouldn't need this line at all. As long as you have your config extraction set up correctly in RC and DebuggerSettings (i.e. the same as we do for `TracerSettings` then the telemetry should already be recorded correctly, and with the correct origins 🙂 
```suggestion
```

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -102,45 +133,152 @@ private void SetGeneralConfig(TracerSettings tracerSettings, DebuggerSettings se
...
+
                 if (!ExceptionReplaySettings.Enabled)
                 {
-                    Log.Information("Exception Replay is disabled. To enable it, please set {ExceptionReplayEnabled} environment variable to '1'/'true'.", ConfigurationKeys.Debugger.ExceptionReplayEnabled);
+                    SafeDisposal.TryDispose(ExceptionReplay);
+                    ExceptionReplay = null;
+                    if (debuggerSettings.DynamicSettings.ExceptionReplayEnabled == false)
+                    {
+                        Log.Information("Exception Replay is disabled by remote enablement. To enable it, re-enable it via Datadog UI");
+                        tracerSettings.Telemetry.Record(ConfigurationKeys.Debugger.ExceptionReplayEnabled, false, ConfigurationOrigins.RemoteConfig);
```
andrewlock: This is already recorded by the `ConfigurationBuilder` in `DynamicConfigurationManager`
```suggestion
```

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -102,45 +133,152 @@ private void SetGeneralConfig(TracerSettings tracerSettings, DebuggerSettings se
...
+                }
+
                 if (!ExceptionReplaySettings.Enabled)
                 {
-                    Log.Information("Exception Replay is disabled. To enable it, please set {ExceptionReplayEnabled} environment variable to '1'/'true'.", ConfigurationKeys.Debugger.ExceptionReplayEnabled);
+                    SafeDisposal.TryDispose(ExceptionReplay);
+                    ExceptionReplay = null;
+                    if (debuggerSettings.DynamicSettings.ExceptionReplayEnabled == false)
+                    {
+                        Log.Information("Exception Replay is disabled by remote enablement. To enable it, re-enable it via Datadog UI");
```
andrewlock, line 302: IMO, all these info logs are too noisy - we shouldn't be trying to tell customers about features in our logs, that should be in the UI. Doing it in the logs makes it harder to debug (noise), it impacts startup time (logs are slow), it uses up drive space, and customers don't look at them anyway 😅

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -153,202 +291,216 @@ private void SetExceptionReplayState()
                 exceptionReplay.Initialize();
                 ExceptionReplay = exceptionReplay;
 
-                // TODO:
-                // TracerManager.Instance.Telemetry.ProductChanged(TelemetryProductType.ExceptionReplay, enabled: true, error: null);
+                tracerSettings.Telemetry.Record(ConfigurationKeys.Debugger.ExceptionReplayEnabled, true, debuggerSettings.DynamicSettings.ExceptionReplayEnabled == true ? ConfigurationOrigins.RemoteConfig : ConfigurationOrigins.AppConfig);
```
andrewlock: This is already recorded by the `ConfigurationBuilder` in `DynamicConfigurationManager`
```suggestion
```

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -153,202 +291,216 @@ private void SetExceptionReplayState()
...
+            }
 
-                if (DynamicInstrumentation != null)
+            if (!tracerSettings.IsRemoteConfigurationAvailable)
+            {
+                if (debuggerSettings.DynamicInstrumentationEnabled)
                 {
-                    Log.Debug("Dynamic Instrumentation is already initialized");
-                    return;
+                    Log.Warning("Remote configuration is not available in this environment, so Dynamic Instrumentation cannot be enabled.");
```
andrewlock, line 342: In contrast to my above comment, this one is valid to keep👍

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -153,202 +291,216 @@ private void SetExceptionReplayState()
...
-                return;
+                Interlocked.CompareExchange(ref _diDebounceCts, null, cts);
+                SafeDisposal.TryDispose(cts);
+                SafeDisposal.TryDispose(prevCts);
             }
 
-            OneTimeSetup(tracerSettings);
+            bool ShouldSkipUpdate(bool requested, bool current)
+            {
+                if (requested && (requested == current))
```
andrewlock: Is this correct? 🤔 It means you _will_ update if you're currently disabled and requested is _also_ disabled. Shouldn't it be "update if there's any change"?
```suggestion
                if (requested == current)
```

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -153,202 +291,216 @@ private void SetExceptionReplayState()
...
+            {
+                Log.Debug("Waiting {Timeout}ms before updating Dynamic Instrumentation", _diDebounceDelay.TotalMilliseconds);
+                await Task.Delay(_diDebounceDelay, cts.Token).ConfigureAwait(false);
+
+                // Re-check if state change is still needed after debounce
+                if (!cts.IsCancellationRequested)
                 {
-                    var isDiscoverySuccessful = await WaitForDiscoveryServiceAsync(discoveryService, _processExit.Task).ConfigureAwait(false);
-                    if (!isDiscoverySuccessful)
+                    if (ShouldSkipUpdate(DebuggerSettings.DynamicInstrumentationEnabled, DynamicInstrumentation is { IsDisposed: false }))
```
andrewlock: This seems like a curious condition check 🤔 Because `ShouldSkipUpdate` takes a "requested" and a "current" but this is testing something completely different? 🤔

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -153,202 +291,216 @@ private void SetExceptionReplayState()
...
+            if (ShouldSkipUpdate(debuggerSettings.DynamicInstrumentationEnabled, DynamicInstrumentation is { IsDisposed: false }))
+            {
+                return;
+            }
 
-                    tracerManager.Telemetry.ProductChanged(TelemetryProductType.DynamicInstrumentation, enabled: false, error: null);
-                    return;
-                }
+            // Cancel any pending operation and start new one
+            var cts = new CancellationTokenSource();
```
andrewlock: Careful about CTS, remember the shutdown exceptions issue 🙂

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -153,202 +291,216 @@ private void SetExceptionReplayState()
...
-
-        private async Task UpdateProductsState(TracerSettings tracerSettings, DebuggerSettings newDebuggerSettings)
-        {
-            if (_processExit.Task.IsCompleted)
+            finally
             {
-                return;
+                Interlocked.CompareExchange(ref _diDebounceCts, null, cts);
+                SafeDisposal.TryDispose(cts);
+                SafeDisposal.TryDispose(prevCts);
```
andrewlock: Remember that `SafeDisposal` isn't safe 😉

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -153,202 +291,216 @@ private void SetExceptionReplayState()
...
+                if (_processExit.Task.IsCompleted || cancellationToken.IsCancellationRequested)
                 {
-                    attemptsRemaining--;
-                    semaphoreAcquired = await _semaphore.WaitAsync(TimeSpan.FromSeconds(5)).ConfigureAwait(false);
+                    return;
                 }
 
-                if (_processExit.Task.IsCompleted)
+                var requestedState = debuggerSettings.DynamicInstrumentationEnabled;
+                var currentState = DynamicInstrumentation is { IsDisposed: false };
```
andrewlock: You're accessing global state here that could have changed separately, which seems dangerous. You have no locking around this or anything

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -153,202 +291,216 @@ private void SetExceptionReplayState()
...
                 {
-                    try
-                    {
-                        _semaphore.Release();
-                    }
-                    catch
-                    {
-                        // ignore ObjectDisposedException or SemaphoreFullException
-                    }
+                    SafeDisposal.TryDispose(DynamicInstrumentation);
```
andrewlock: This worries me, you have other places in this type accessing `DynamicInstrumentation` and I don't see indication that it's safe to assume that it works when it's disposed. 😕

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -153,202 +291,216 @@ private void SetExceptionReplayState()
...
-            finally
+
+            void DisableDynamicInstrumentation(bool currentState)
             {
-                if (semaphoreAcquired)
+                Log.Debug("Disabling Dynamic Instrumentation");
+
+                var origin = debuggerSettings.DynamicSettings.DynamicInstrumentationEnabled == false
+                                 ? ConfigurationOrigins.RemoteConfig
+                                 : ConfigurationOrigins.AppConfig;
```
andrewlock: I would not use origins here
```suggestion
                var disabledViaRemoteConfig = debuggerSettings.DynamicSettings.DynamicInstrumentationEnabled == false;
```

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -153,202 +291,216 @@ private void SetExceptionReplayState()
...
-                        _semaphore.Release();
-                    }
-                    catch
-                    {
-                        // ignore ObjectDisposedException or SemaphoreFullException
-                    }
+                    SafeDisposal.TryDispose(DynamicInstrumentation);
+                    DynamicInstrumentation = null;
+                    TracerManager.Instance.Telemetry.ProductChanged(TelemetryProductType.DynamicInstrumentation, enabled: false, error: null);
+                    tracerSettings.Telemetry.Record(ConfigurationKeys.Debugger.DynamicInstrumentationEnabled, false, origin);
```
andrewlock:
```suggestion
```

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -153,202 +291,216 @@ private void SetExceptionReplayState()
...
+                }
+
+                if (origin == ConfigurationOrigins.RemoteConfig)
```
andrewlock:
```suggestion
                if (disabledViaRemoteConfig)
```

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -153,202 +291,216 @@ private void SetExceptionReplayState()
...
+                    discoveryService,
+                    RcmSubscriptionManager.Instance,
+                    tracerSettings,
+                    ServiceName,
+                    debuggerSettings,
+                    tracerManager.GitMetadataTagsProvider);
+                Log.Debug("Dynamic Instrumentation has been created");
+                DynamicInstrumentation.Initialize();
+                tracerManager.Telemetry.ProductChanged(TelemetryProductType.DynamicInstrumentation, enabled: true, error: null);
+                tracerSettings.Telemetry.Record(ConfigurationKeys.Debugger.DynamicInstrumentationEnabled, true, debuggerSettings.DynamicSettings.DynamicInstrumentationEnabled == true ? ConfigurationOrigins.RemoteConfig : ConfigurationOrigins.AppConfig);
```
andrewlock:
```suggestion
```

tracer/src/Datadog.Trace/Debugger/DebuggerSettings.cs
```diff
@@ -29,12 +31,31 @@ internal class DebuggerSettings
...
+                           null,
+                           x => x switch
+                           {
+                               _ when x.ToBoolean() is { } boolean => boolean ? "true" : "false",
+                               _ => ParsingResult<string>.Failure()
+                           });
+
+            _internalDynamicInstrumentationEnabled = diEnabledAsString is null or "null"
+                                                         ? null
+                                                         : diEnabledAsString.ToBoolean();
```
andrewlock: Maybe I'm missing something, but I think this can be rewritten as
```suggestion
            _internalDynamicInstrumentationEnabled =
                config.WithKeys(ConfigurationKeys.Debugger.DynamicInstrumentationEnabled)
                      .AsBool();
```

But anyway, I think you can delete this property entirely (described below)

tracer/src/Datadog.Trace/Debugger/DebuggerSettings.cs
```diff
@@ -133,6 +154,21 @@ public DebuggerSettings(IConfigurationSource? source, IConfigurationTelemetry te
...
+                           x => x switch
+                           {
+                               _ when x.ToBoolean() is { } boolean => boolean ? "true" : "false",
+                               _ => ParsingResult<string>.Failure()
+                           });
+
+            _internalCodeOriginForSpansEnabled = coEnabledAsString is null or "null"
+                                                     ? null
+                                                     : coEnabledAsString.ToBoolean();
+
```
andrewlock: I think you should delete all this

tracer/src/Datadog.Trace/Debugger/DebuggerSettings.cs
```diff
@@ -141,7 +177,15 @@ public DebuggerSettings(IConfigurationSource? source, IConfigurationTelemetry te
...
-        public bool DynamicInstrumentationEnabled { get; }
+        internal ImmutableDynamicDebuggerSettings DynamicSettings { get; init; } = new();
+
+        public bool DynamicInstrumentationEnabled
+        {
+            get =>
+                (_internalDynamicInstrumentationEnabled == true && DynamicSettings.DynamicInstrumentationEnabled == null)
+             || (_internalDynamicInstrumentationEnabled == null && DynamicSettings.DynamicInstrumentationEnabled == true);
+            init { }
+        }
```
andrewlock: I'd suggest you rewrite this to be similar to the tracer version e.g. 
```suggestion
        public bool DynamicInstrumentationEnabled
            => DynamicSettings.DynamicInstrumentationEnabled ?? _internalDynamicInstrumentationEnabled;
```

And then remove all the other changes above, make `_internalDynamicInstrumentationEnabled` a `bool` and do this change:

```diff
- DynamicInstrumentationEnabled = config
+ _internalDynamicInstrumentationEnabled = config
    .WithKeys(ConfigurationKeys.Debugger.DynamicInstrumentationEnabled)
    .AsBool(false);
```

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
```diff
@@ -87,87 +82,82 @@ internal DynamicInstrumentation(
...
+            // Start initialization in background
+            _ = Task.Run(InitializeAsync)
+                    .ContinueWith(
+                         t =>
+                         {
+                             if (t is { Exception: not null })
+                             {
+                                 Log.Error(t.Exception.Flatten(), "Error in DI initialization");
+                             }
+                         },
```
andrewlock:
```suggestion
                         t => Log.Error(t?.Exception, "Error in DI initialization"),
```

tracer/src/Datadog.Trace/Debugger/ExceptionAutoInstrumentation/ExceptionTrackManager.cs
```diff
@@ -52,7 +52,8 @@ private ExceptionTrackManager(ExceptionReplaySettings settings)
             _cachedInvalidatedCases = new();
 
             _exceptionProcessorTask = Task.Factory.StartNew(
-                                               async () => await StartExceptionProcessingAsync(_cts.Token).ConfigureAwait(false), TaskCreationOptions.LongRunning)
+                                               async () => await StartExceptionProcessingAsync().ConfigureAwait(false),
```
andrewlock:
```suggestion
                                               () => StartExceptionProcessingAsync(),
```

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
dudikeleti: My thought was to avoid log the error in case the exception is null (even though that should call only on fault), but I'm ok with your suggestion.

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
dudikeleti: In case the symbol upload is disabled, or all 3 products (CO, ER and DI) are explicitly disabled, the symbols upload should be skipped. The thing is that if the features are not explicitly disabled, but empty, we will want to upload symbols as soon as possible even if they are not currently enabled, in case it is enabled remotely. Otherwise, the user may not see symbols until we upload them all. Note, that we don't start to upload symbols untill we got a rc request to do so.

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
dudikeleti: I was thinking on AppConfig as the opposite of RemoteConfig 😄🤦‍♂️
Thanks for that knowledge.

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
dudikeleti: I get it. I'll change it to debug.

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
dudikeleti: You right about that, but I did it to make DI tests pass, we don't do anything in the case where the requested is `false` and the current is `false`, but simply postpone this choice until later.
But it's too ugly, I'll fix the tests.

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
dudikeleti: Maybe I miss something but I think the condition is correct. First parameter is requested state and second paramter is current state

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
dudikeleti: That correct. We access it inside the class during enabling which is inside try/catch block or outside the class with null checking. I'll make sure that when using outside the class will always get null or a valid object.

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
dudikeleti: Fixed in a514b2898b89a756e9f5280e276e17c62fbc9c73

tracer/src/Datadog.Trace/Debugger/DebuggerSettings.cs
dudikeleti: Fixed in 79622fa5719c458094d4a7aa65ae596b58c7890b

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
dudikeleti: Fixed in 41e5d626c9471697451540f2fb579cf6f8f8835d

tracer/src/Datadog.Trace/Configuration/DynamicConfigurationManager.cs
dudikeleti: Fixed in 41e5d626c9471697451540f2fb579cf6f8f8835d

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
dudikeleti: Fixed in 41e5d626c9471697451540f2fb579cf6f8f8835d

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
dudikeleti: Fixed in 41e5d626c9471697451540f2fb579cf6f8f8835d

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
dudikeleti: Fixed in 41e5d626c9471697451540f2fb579cf6f8f8835d

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
dudikeleti: Fixed in 41e5d626c9471697451540f2fb579cf6f8f8835d

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
dudikeleti: Fixed in 41e5d626c9471697451540f2fb579cf6f8f8835d

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
dudikeleti: Fixed in 41e5d626c9471697451540f2fb579cf6f8f8835d

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
dudikeleti: Fixed in 41e5d626c9471697451540f2fb579cf6f8f8835d

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
dudikeleti: Fixed in 41e5d626c9471697451540f2fb579cf6f8f8835d

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
dudikeleti: Fixed in 41e5d626c9471697451540f2fb579cf6f8f8835d

tracer/src/Datadog.Trace/Debugger/DebuggerSettings.cs
dudikeleti: But then how can I distinguish between `empty/null` to explicit `false`? I need to know that to decide if a feature might be enabled in the future or not.

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
dudikeleti: Fixed in ad909d4c3e6dd2df59137357b3761ba23d184175, also in DynamicInstrumentation we check for disposed.

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
dudikeleti: Fixed in 25c890ad31dc9b80868d533c911c6c8fb9e68d72 and 40cbacfbedf676217436d13cf3d1a2ffaccc6b8b

tracer/src/Datadog.Trace/Debugger/DebuggerSettings.cs
dudikeleti: Fixed in 79622fa5719c458094d4a7aa65ae596b58c7890b

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
dudikeleti: Fixed in 79622fa5719c458094d4a7aa65ae596b58c7890b
Let me know if you still want to to disable it completly untill we got the first rc request to enable it.

dudikeleti: See also ddfb2c474bc1f9a98f04ff1266b6a8581aae7c35

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
```diff
@@ -446,26 +446,20 @@ private static void StartDiagnosticManager()
         private static void InitializeDebugger(TracerSettings tracerSettings)
         {
             var manager = DebuggerManager.Instance;
-            if (manager.DebuggerSettings.CodeOriginForSpansEnabled
-                || manager.DebuggerSettings.DynamicInstrumentationEnabled
-                || manager.ExceptionReplaySettings.Enabled)
+            if (manager.DebuggerSettings.IsDebuggerProductsDisabled && manager.ExceptionReplaySettings.IsExceptionReplayDisabled)
```
GreenMatan: We cannot do that due to Stable Configuration (PR: #7441). We have two new env vars:
1. `DD_DYNAMIC_INSTRUMENTATION_MANAGED_ACTIVATION_ENABLED `
2. `DD_EXCEPTION_REPLAY_MANAGED_ACTIVATION_ENABLED`

that dictate if Stable Configuration should be taken into account or not. Both of them are `true` by default. If the user sets them to `false` - then we have to look at our own env var to determine if our products are explicitly disabled.
Apparently, Stable Configuration take precedence over local env var (🤷)

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
```diff
@@ -446,26 +446,20 @@ private static void StartDiagnosticManager()
...
-                        {
-                            Log.Error(ex, "Error initializing debugger");
-                        }
-                    });
+                Log.Debug("Debugger products are explicitly disabled via environment variables.");
             }
             else
             {
-                Log.Information($"Dynamic Instrumentation is disabled. To enable it, please set {ConfigurationKeys.Debugger.DynamicInstrumentationEnabled} environment variable to 'true'.");
+                if (!manager.DebuggerSettings.DynamicInstrumentationEnabled)
```
GreenMatan: https://github.com/DataDog/dd-trace-dotnet/pull/7366#discussion_r2315850417

tracer/src/Datadog.Trace/Configuration/DynamicConfigurationManager.cs
```diff
@@ -91,27 +99,60 @@ private static void OnConfigurationChanged(ConfigurationBuilder settings)
             // Needs to be done before returning, to feed the value to the telemetry
             // var debugLogsEnabled = settings.WithKeys(ConfigurationKeys.DebugEnabled).AsBool();
 
+            TracerSettings newSettings;
             if (dynamicSettings.Equals(oldSettings.DynamicSettings))
             {
                 Log.Debug("No changes detected in the new dynamic configuration");
-                return;
+                newSettings = oldSettings;
```
GreenMatan, line 106: There was early exit in this branch. You got rid of it on purpose?

tracer/src/Datadog.Trace/Configuration/DynamicConfigurationManager.cs
```diff
@@ -91,27 +99,60 @@ private static void OnConfigurationChanged(ConfigurationBuilder settings)
...
+                /*
+                if (debugLogsEnabled != null && debugLogsEnabled.Value != GlobalSettings.Instance.DebugEnabled)
+                {
+                    GlobalSettings.SetDebugEnabledInternal(debugLogsEnabled.Value);
+                    Security.Instance.SetDebugEnabled(debugLogsEnabled.Value);
 
-            Log.Information("Applying new dynamic configuration");
+                    NativeMethods.UpdateSettings(new[] { ConfigurationKeys.DebugEnabled }, new[] { debugLogsEnabled.Value ? "1" : "0" });
+                }
+                */
```
GreenMatan, line 122: do we need this comment or it's a left over?

tracer/test/Datadog.Trace.Tests/Debugger/DynamicInstrumentationTests.cs
```diff
@@ -79,7 +79,7 @@ public void DynamicInstrumentationDisabled_ServicesNotCalled()
 
         var debugger = new DynamicInstrumentation(settings, discoveryService, rcmSubscriptionManagerMock, lineProbeResolver, snapshotUploader, diagnosticsUploader, probeStatusPoller, updater, new DogStatsd.NoOpStatsd());
         debugger.Initialize();
-
+        await Task.Delay(1000);
```
GreenMatan: why do we need this delay?

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -102,253 +142,389 @@ private void SetGeneralConfig(TracerSettings tracerSettings, DebuggerSettings se
...
+
+                    instance = _dynamicInstrumentation;
+                }
+
+                if (created && instance != null)
+                {
+                    Log.Debug("Dynamic Instrumentation has been created");
+                    instance.Initialize();
+                    tracerManager.Telemetry.ProductChanged(TelemetryProductType.DynamicInstrumentation, enabled: true, error: null);
                 }
```
GreenMatan: since you lock only specific part, do we care if we end up calling `instance.Initialize` on `DynamicInstrumentation` instance while another thread tries to create/destroy it potentially? Do we have potential race here?

GreenMatan (approved):
submitting first batch of comments, will keep reviewing and potentially send another batch.
Nice work!

tracer/src/Datadog.Trace/Configuration/DynamicConfigurationManager.cs
dudikeleti: It was there before and I'm not sure if should I delete it or not

tracer/test/Datadog.Trace.Tests/Debugger/DynamicInstrumentationTests.cs
dudikeleti: To let the DI background initialization to run theoritically so in case we have a bug in our enabling condition this test will fail.

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
dudikeleti: Fixed in 1d6bfc79afccef50eb29faf1211fa5ffcd3cea83

tracer/src/Datadog.Trace/Debugger/DebuggerSettings.cs
andrewlock: > I need to know that to decide if a feature might be enabled in the future or not.

Remote configuration overrides the value anyway, so it can _always_ be added in the future I think? And if not, then we have a layering problem and confusion issue across products :/

tracer/test/Datadog.Trace.Tests/Debugger/DynamicInstrumentationTests.cs
andrewlock: We should not have any `Task.Delay` in tests, because it _will_ make them flaky. I practically guarantee that this will take longer than 1s in some cases, because the CI machines run at basically 100% capacity. 

I strongly suggest having a deterministic way to know whether initialization has completed. If you can't, that suggests we need to redesign the code - APIs should be testable by default (e.g. no static access too 😉)

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
andrewlock: > We cannot do that due to Stable Configuration

This is already all handled in managed code by the `ConfigurationBuilder` infrastructure. You _shouldn't_ need to do anything in managed code to worry about it 🙂

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -36,19 +36,25 @@ internal class DebuggerManager
...
 
-        private static volatile bool _discoveryServiceReady;
-        private readonly SemaphoreSlim _semaphore;
-        private readonly TaskCompletionSource<bool> _processExit = new(TaskCreationOptions.RunContinuationsAsynchronously);
+        private readonly object _syncLock;
+        private readonly TimeSpan _diDebounceDelay;
+        private readonly TaskCompletionSource<bool> _processExit;
+        private volatile bool _isDebuggerEndpointAvailable;
         private int _initialized;
+        private CancellationTokenSource? _diDebounceCts;
```
andrewlock: We shouldn't use `CancellationTokenSource` in paths that can be hit during shutdown, because it can cause crashes. We generally have to use `TaskCompletionSource` instead (even though it's a bit of a pain)

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
andrewlock: > Let me know if you still want to to disable it completly untill we got the first rc request to enable it

I think we should do that, because otherwise we're doing a bunch of work (and impacting performance) that most of our customers will never benefit from. I think we should definitely be lazy for all of this setup wherever we can be, because it has a clear impact on our benchmarks, and for large apps that could end up being very significant

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -152,203 +298,304 @@ private void SetExceptionReplayState()
...
-                    writer.WriteValue(DebuggerSettings.SymbolDatabaseUploadEnabled);
+                    writer.WriteValue(settings.CodeOriginForSpansEnabled);
                     writer.WritePropertyName("exception_replay_enabled");
                     writer.WriteValue(ExceptionReplaySettings.Enabled);
+                    writer.WritePropertyName("symbol_database_upload_enabled");
+                    writer.WriteValue(settings.SymbolDatabaseUploadEnabled);
+                    writer.WriteEndObject();
                 }
 
                 Log.Information("DATADOG DEBUGGER CONFIGURATION - {Configuration}", stringWriter.ToString());
```
andrewlock, line 593: Given you always write this log (which I think is fine, as it mirrors the tracer version 👍), you can probably change the other logs that are used by tests to use this one, instead of writing additional logs

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
```diff
@@ -87,87 +82,76 @@ internal DynamicInstrumentation(
...
 
         internal void Initialize()
         {
-            var originalState = Interlocked.CompareExchange(ref _initState, 1, 0);
-
-            // If we weren't in "not initialized" state, return early
-            if (originalState != 0)
+            if (!_settings.DynamicInstrumentationEnabled)
             {
+                Log.Information("Dynamic Instrumentation is disabled. To enable it, please set DD_DYNAMIC_INSTRUMENTATION_ENABLED environment variable to 'true'.");
```
andrewlock: nit: You're already writing this log elsewhere, we should remove one of them (this is a problem today, we write the same log twice)

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
```diff
@@ -87,87 +82,76 @@ internal DynamicInstrumentation(
...
-                if (!_settings.DynamicInstrumentationEnabled)
-                {
-                    Log.Information("Dynamic Instrumentation is disabled. To enable it, please set DD_DYNAMIC_INSTRUMENTATION_ENABLED environment variable to 'true'.");
-                    // Reset to "not initialized"
-                    Interlocked.Exchange(ref _initState, 0);
-                    return;
-                }
-
-                Log.Information("Dynamic Instrumentation initialization started");
+            Log.Information("Dynamic Instrumentation initialization started");
```
andrewlock: nit: this should probably be debug

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
```diff
@@ -87,87 +82,76 @@ internal DynamicInstrumentation(
...
-                _ = Task.Run(async () => await InitializeAsync().ConfigureAwait(false), _cancellationTokenSource.Token);
-            }
-            catch (Exception e)
-            {
-                Log.Error(e, "Initializing Dynamic Instrumentation failed.");
-                // Reset to "not initialized"
-                Interlocked.Exchange(ref _initState, 0);
-            }
+            // Start initialization in background
+            _ = Task.Run(InitializeAsync)
```
andrewlock: There's so much "in the background" stuff that I get lost 😢 Given _everything_ is started "in the background" anyway in `Instrumentation`, do we _really_ need to make everything explicitly use `Task.Run`? It would be much easier to follow if the `DebuggerManager` for example, could just do something like this:

```csharp
Task allTasks = Task.WhenAll(
                  DynamicInstrumentation.InitializeAsync(),
                  SymbolUploader.InitializeAsync(),
                  SomethingElse.InitializeAsync());
```

Not required, and could be another PR if you want, but I think it would make a lot of these things much easier to follow

tracer/src/Datadog.Trace/Debugger/ExceptionAutoInstrumentation/ExceptionReplaySettings.cs
```diff
@@ -20,15 +17,16 @@ internal class ExceptionReplaySettings
...
         {
             source ??= NullConfigurationSource.Instance;
             var config = new ConfigurationBuilder(source, telemetry);
 
-#pragma warning disable CS0612 // Type or member is obsolete
-            Enabled = config.WithKeys(ConfigurationKeys.Debugger.ExceptionDebuggingEnabled, ConfigurationKeys.Debugger.ExceptionReplayEnabled).AsBool(false);
-#pragma warning restore CS0612 // Type or member is obsolete
+            Enabled = config.WithKeys(ConfigurationKeys.Debugger.ExceptionReplayEnabled).AsBool(false);
+
+            _internalExceptionReplayEnabled = config.WithKeys(ConfigurationKeys.Debugger.ExceptionReplayEnabled).AsBool();
```
andrewlock: You are reading the same setting from _all_ of the sources, multiple times, this is expensive, and shouldn't be necessary. 

If you need to do this (I'm still not entirely convinced you do - you should always be able to enable via remote config) you can do it like this:

```suggestion
            var enabledResult = config.WithKeys(ConfigurationKeys.Debugger.ExceptionReplayEnabled).AsBool(false);

            Enabled = enabledResult.WithDefault(false);
            CanBeEnabled = !(enabledResult.ConfigurationResult.IsValid && enabledResult.ConfigurationResult.Result == false);
```

Note that I changed the name of the property (and removed the unecessary field), because it's currently confusing to have `ExceptionReplaySettings.Enabled` and `IsExceptionReplayDisabled` meaning different things

tracer/src/Datadog.Trace/Debugger/ExceptionAutoInstrumentation/ExceptionTrackManager.cs
```diff
@@ -52,7 +52,8 @@ private ExceptionTrackManager(ExceptionReplaySettings settings)
             _cachedInvalidatedCases = new();
 
             _exceptionProcessorTask = Task.Factory.StartNew(
-                                               async () => await StartExceptionProcessingAsync(_cts.Token).ConfigureAwait(false), TaskCreationOptions.LongRunning)
+                                               () => StartExceptionProcessingAsync(),
+                                               TaskCreationOptions.LongRunning)
```
andrewlock: You probably need to make sure you're using the default scheduler if you're using `StartNew` instead of `Task.Run()`

andrewlock (commented):
Thanks, I left a few comments, in particular I think we need to try to reduce the impact when the debugger is not enabled.

That said, I think the main thing missing at this point is integration tests for the remote enablement (unless they're already there and I just missed them!). I think we should have tests for:

- start with the debugger enabled, disable it, then re-enable it
- start with the debugger disabled, enable it, then re-disable it

I believe there are some ASM tests that do something similar, so might be worth looking at those for inspiration! Thanks

tracer/test/Datadog.Trace.Tests/Debugger/DynamicInstrumentationTests.cs
tonyredondo: please let's avoid using any Thread.Sleep or Task.Delay in tests.

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
GreenMatan: yes, we're currently discussing it. It seems we're all good :)

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -102,45 +144,149 @@ private void SetGeneralConfig(TracerSettings tracerSettings, DebuggerSettings se
...
+                    {
+                        try
+                        {
+                            await SymbolsUploader.StartFlushingAsync().ConfigureAwait(false);
+                        }
+                        catch (Exception ex)
+                        {
+                            Log.Error(ex, "Failed to initialize symbol uploader");
+                        }
+                    });
```
tonyredondo: Why not 
```
.ContinueWith(
                         t => Log.Error(t?.Exception, "Failed to initialize symbol uploader"),
                         TaskContinuationOptions.OnlyOnFaulted);
```

tracer/src/Datadog.Trace/Debugger/ExceptionAutoInstrumentation/ExceptionTrackManager.cs
tonyredondo: This is a `LongRunning` task, a new thread is created for this one, but in `StartExceptionProcessingAsync` you have a `.ConfigureAwait(false)` so this new thread will be lost at that first Task.WhenAny, because on return the continuation will be schedule to a thread pool thread.

tracer/test/Datadog.Trace.Tests/Debugger/DynamicInstrumentationTests.cs
dudikeleti: Fixed in 5683a7c78bfbca902bc8eb61710df96ff20c78ad

> I strongly suggest having a deterministic way to know whether initialization has completed

We have a way for that, but in this case we need the opposite. Anyway I prefer to leave it as is was before, we don't have to change it now.

tracer/src/Datadog.Trace/Debugger/ExceptionAutoInstrumentation/ExceptionTrackManager.cs
dudikeleti: Thanks for pointing that. Fixed in dd64f6e89a16c1969abd9c1deb54d70d0f8c89c4

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
dudikeleti: Fixed in dd64f6e89a16c1969abd9c1deb54d70d0f8c89c4

tracer/src/Datadog.Trace/Debugger/ExceptionAutoInstrumentation/ExceptionTrackManager.cs
dudikeleti: Fixed in dd64f6e89a16c1969abd9c1deb54d70d0f8c89c4

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
dudikeleti: Thanks for pointing that. Fixed in f3d6f4c949964231419ab232c0b5939abb08ac8e

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
dudikeleti: I think that we use today only this initialization log in tests (there ar emore but not related to initialization)
`Dynamic Instrumentation is disabled. To enable it, please set DD_DYNAMIC_INSTRUMENTATION_ENABLED environment variable to 'true'.`
We can change the test to use the configuration value but we still want this log (without [this ](https://github.com/DataDog/dd-trace-dotnet/pull/7366#discussion_r2318345301) log).

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
dudikeleti: Fixed in 9cc8537e8fce4bd242b9c5798bfc51c00d140e38

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
dudikeleti: Fixed in 9cc8537e8fce4bd242b9c5798bfc51c00d140e38

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
dudikeleti: Deleted some Task.Run in d3b96dbf51b46f1ac593b3faee04ef7803804c7e

tracer/src/Datadog.Trace/Debugger/ExceptionAutoInstrumentation/ExceptionReplaySettings.cs
dudikeleti: Does it make sense now? bfa3f79b5c87db7e916afbdc9375dce6dfcf2ae0

tracer/src/Datadog.Trace/Debugger/DebuggerSettings.cs
dudikeleti: Does it make sense now? bfa3f79b5c87db7e916afbdc9375dce6dfcf2ae0

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
dudikeleti: After bfa3f79b5c87db7e916afbdc9375dce6dfcf2ae0 symdb won't be on by default if di is explicitly disabled

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
```diff
@@ -446,26 +446,29 @@ private static void StartDiagnosticManager()
...
             var manager = DebuggerManager.Instance;
-            if (manager.DebuggerSettings.CodeOriginForSpansEnabled
-                || manager.DebuggerSettings.DynamicInstrumentationEnabled
-                || manager.ExceptionReplaySettings.Enabled)
+            var debuggerSettings = manager.DebuggerSettings;
+            bool diDisabled = !debuggerSettings.DynamicInstrumentationCanBeEnabled;
+            bool coDisabled = !debuggerSettings.CodeOriginForSpansCanBeEnabled;
+            bool erDisabled = !manager.ExceptionReplaySettings.CanBeEnabled || debuggerSettings.DynamicSettings.ExceptionReplayEnabled == false;
+
+            if (diDisabled && coDisabled && erDisabled)
```
andrewlock: Just FYI, my concern is that these will almost never all be set (because customers would have to explicitly disable all of them) so this effectively means we will be initializing for all customers - we'll need to check the execution benchmarks to check that this doesn't have an adverse impact in general 🙂

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
andrewlock: > won't be on by default if di is explicitly disabled

Yeah, the concern is that's basically never going to be the case, so it's effectively still _always_ going to be on by default for customers

tracer/src/Datadog.Trace/Debugger/DebuggerSettings.cs
```diff
@@ -39,11 +39,10 @@ public DebuggerSettings(IConfigurationSource? source, IConfigurationTelemetry te
             source ??= NullConfigurationSource.Instance;
             var config = new ConfigurationBuilder(source, telemetry);
 
-            DynamicInstrumentationEnabled = config.WithKeys(ConfigurationKeys.Debugger.DynamicInstrumentationEnabled).AsBool(false);
-
             _internalDynamicInstrumentationEnabled = config.WithKeys(ConfigurationKeys.Debugger.DynamicInstrumentationEnabled).AsBool();
 
-            SymbolDatabaseUploadEnabled = config.WithKeys(ConfigurationKeys.Debugger.SymbolDatabaseUploadEnabled).AsBool(true);
+            SymbolDatabaseUploadEnabled = config.WithKeys(ConfigurationKeys.Debugger.SymbolDatabaseUploadEnabled).AsBool(true)
+                                       && _internalDynamicInstrumentationEnabled != false;
```
andrewlock: It's important that the default value reflects the actual value, so we shouldn't manipulate the value afterwards. e.g. the following is equivalent, but records the "real" default value

```suggestion
            SymbolDatabaseUploadEnabled = config.WithKeys(ConfigurationKeys.Debugger.SymbolDatabaseUploadEnabled).AsBool(_internalDynamicInstrumentationEnabled != false);
```

tracer/src/Datadog.Trace/Debugger/DebuggerSettings.cs
```diff
@@ -191,11 +184,9 @@ public bool DynamicInstrumentationEnabled
...
         public HashSet<string> RedactedTypes { get; }
 
-        public bool CodeOriginForSpansEnabled
-        {
-            get => DynamicSettings.CodeOriginEnabled ?? _internalCodeOriginForSpansEnabled == true;
-            init { }
-        }
+        public bool CodeOriginForSpansEnabled => _internalCodeOriginForSpansEnabled == true || DynamicSettings.CodeOriginEnabled == true;
+
+        public bool CodeOriginForSpansCanBeEnabled => _internalCodeOriginForSpansEnabled != false && DynamicSettings.CodeOriginEnabled != false;
```
andrewlock: I don't think "can be enabled" should be controlled by dynamic config? 🤔:

```suggestion
        public bool CodeOriginForSpansCanBeEnabled => _internalCodeOriginForSpansEnabled != false;
```

tracer/src/Datadog.Trace/Debugger/ExceptionAutoInstrumentation/ExceptionReplaySettings.cs
```diff
@@ -17,16 +16,15 @@ internal class ExceptionReplaySettings
...
         {
             source ??= NullConfigurationSource.Instance;
             var config = new ConfigurationBuilder(source, telemetry);
 
-            Enabled = config.WithKeys(ConfigurationKeys.Debugger.ExceptionReplayEnabled).AsBool(false);
-
-            _internalExceptionReplayEnabled = config.WithKeys(ConfigurationKeys.Debugger.ExceptionReplayEnabled).AsBool();
+            var enabledResult = config.WithKeys(ConfigurationKeys.Debugger.ExceptionReplayEnabled).AsBool();
+            Enabled = enabledResult ?? false;
+            CanBeEnabled = enabledResult != false;
```
andrewlock: This doesn't record the correct value, so you need to use `AsBoolResult()` and then call `WithDefault()` etc afterwards

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
dudikeleti: That true but we also check later that SymDb is `true` which will be true only if DI is not `false` so even if the customer set only DI to `false`, we will not upload symbols.

>Yeah, the concern is that's basically never going to be the case, so it's effectively still always going to be on by default for customers

Ok, I undrstand. So let's do the opposite, during startap enable initialize DebuggerManager only if one of the products is enabled (== true)

dudikeleti: Fixed in f65a260d976c7b380e2e166f6cb689018d5b6cf4

tracer/src/Datadog.Trace/Debugger/DebuggerSettings.cs
dudikeleti: Fixed in 256bd507d924cc973d55d3bb5467e6c7e5570cf3

tracer/src/Datadog.Trace/Debugger/DebuggerSettings.cs
dudikeleti: Fixed in 256bd507d924cc973d55d3bb5467e6c7e5570cf3

tracer/src/Datadog.Trace/Debugger/ExceptionAutoInstrumentation/ExceptionReplaySettings.cs
dudikeleti: Fixed in 256bd507d924cc973d55d3bb5467e6c7e5570cf3

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -102,253 +143,435 @@ private void SetGeneralConfig(TracerSettings tracerSettings, DebuggerSettings se
...
+        private void OneTimeSetup(TracerSettings tracerSettings)
+        {
+            if (Interlocked.CompareExchange(ref _initialized, 1, 0) != 0)
+            {
+                return;
+            }
+
+            LifetimeManager.Instance.AddShutdownTask(ShutdownTasks);
+            SetGeneralConfig(tracerSettings, DebuggerSettings);
+            InitializeSymbolUploader(tracerSettings);
```
andrewlock: I forgot to mention, discussing with @tylfin he pointed out that the symbol uploading should _not_ be enabled by default, so I think we need to rework this to make sure symbol uploading is lazy, and doesn't start until DI is enabled (i.e. immediately if DI is explicitly enabled, or not until it's enabled remotely)

The concern is that this is a lot of additional network traffic on both sides (client + backend) if it's not going to be used

tracer/src/Datadog.Trace/Debugger/SpanCodeOrigin/SpanCodeOrigin.cs
```diff
@@ -27,6 +27,7 @@ internal class SpanCodeOrigin
 
         internal SpanCodeOrigin(DebuggerSettings settings)
         {
+            Log.Information("Initializing Code Origin for Spans");
```
andrewlock, line 30: I'd rather keep these kinds of "everything ok" logs as debug if possible, they can add a surprising amount to startup latency

tracer/test/Datadog.Trace.Debugger.IntegrationTests/DebuggerManagerDynamicTests.cs
```diff
@@ -0,0 +1,301 @@
...
+    [Trait("Category", "EndToEnd")]
+    [Trait("Category", "ArmUnsupported")]
+    [Trait("RunOnWindows", "True")]
+    [Trait("Category", "LinuxUnsupported")]
+    public async Task DebuggerManager_DynamicInstrumentation_StartDisabled_EnabledViaRemoteConfig()
+    {
+#if NET8_0_OR_GREATER
+        // These tests often hang on x86 on .NET 8+. Needs investigation
+        Skip.If(!EnvironmentTools.IsTestTarget64BitProcess());
+#endif
```
andrewlock: Shouldn't we investigate that _now_, before we merge 😟 I feel a bit uncomfortable knowing we have unknown hangs on all the MS _supported_ versions of .NET 😬

tracer/test/Datadog.Trace.Debugger.IntegrationTests/DebuggerManagerDynamicTests.cs
```diff
@@ -0,0 +1,301 @@
...
+using Datadog.Trace.TestHelpers;
+using Samples.Probes.TestRuns.SmokeTests;
+using VerifyXunit;
+using Xunit;
+using Xunit.Abstractions;
+
+#nullable enable
+namespace Datadog.Trace.Debugger.IntegrationTests;
+
+#if !NETCOREAPP2_1
```
andrewlock, line 23: This isn't a big deal, as we don't test on 2.1 anymore, but why do we need this explicit exclude?

tracer/test/Datadog.Trace.Debugger.IntegrationTests/DebuggerManagerDynamicTests.cs
```diff
@@ -0,0 +1,301 @@
...
+using Datadog.Trace.Debugger.IntegrationTests.Assertions;
+using Datadog.Trace.Debugger.IntegrationTests.Helpers;
+using Datadog.Trace.Debugger.Sink;
+using Datadog.Trace.TestHelpers;
+using Samples.Probes.TestRuns.SmokeTests;
+using VerifyXunit;
+using Xunit;
+using Xunit.Abstractions;
+
+#nullable enable
```
andrewlock, line 20: Just FYI, _in general_ `#nullable enable` isn't in the tests, and we tend not to add it, as it mostly makes the tests a bit more annoying to write 😄

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
dudikeleti: > The concern is that this is a lot of additional network traffic on both sides (client + backend) if it's not going to be used
Note that SymDB initialization by itself do nothing unless we got 2 things:
1. RC request to upload symbols
2. SymbDB endpoint

Fixed in 82ca65b3f4214d6921db30c08c5f7e8f7bdb409a

tracer/src/Datadog.Trace/Debugger/SpanCodeOrigin/SpanCodeOrigin.cs
dudikeleti: Yes I'd rather too but the thing is that we heavily use the logger wathcer for all kind of tests. We can change it, but then I prefer to do that for all tests instead of having difference methods to check things, and better to do that in a different PR if we will decide to change it.

tracer/test/Datadog.Trace.Debugger.IntegrationTests/DebuggerManagerDynamicTests.cs
dudikeleti: We have a crash in ClrMD on NETCOREAPP2_1

tracer/test/Datadog.Trace.Debugger.IntegrationTests/DebuggerManagerDynamicTests.cs
dudikeleti: It's make my life easier to put it everywhere 😄

tracer/test/Datadog.Trace.Debugger.IntegrationTests/DebuggerManagerDynamicTests.cs
dudikeleti: Trying to run without skipping. I'll keep you posted.
b35dfcc3872e2aebc1c9c18d2e5b8b12ba381e83

andrewlock (approved):

PR #7367, opened 2025-08-08
Clean up header when tracing is disabled

williamscs:
## Summary of changes

When tracing is disabled, also remove the "tracing enabled" header.

## Reason for change

We have a use-case where a partner doesn't allow unknown headers. This header allows the tracing headers to not be added, but it still remains when the request is sent, which breaks their system.

## Test coverage

None that I found. This is more of a proof of concept to get the maintainers feedback. Happy to dig in more, if this is a good idea.

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Http/HttpClient/HttpMessageHandlerCommon.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Http/HttpClient/HttpMessageHandlerCommon.cs
@@ -46,6 +46,10 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Http.HttpClient
                     return new CallTargetState(scope);
                 }
             }
+            else
+            {
+                headers.Remove(HttpHeaderNames.TracingEnabled);
+            }
 
             return CallTargetState.GetDefault();
         }
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Http/WebRequest/WebRequestCommon.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Http/WebRequest/WebRequestCommon.cs
@@ -41,7 +41,12 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Http.WebRequest
         /// <returns>Returns <c>true</c> if injection was performed, and <c>/false</c> otherwise</returns>
         public static bool TryInjectHeaders<TTarget>(TTarget instance)
         {
-            if (instance is HttpWebRequest request && IsTracingEnabled(request))
+            if (instance is not HttpWebRequest request)
+            {
+                return false;
+            }
+
+            if (IsTracingEnabled(request))
             {
                 var tracer = Tracer.Instance;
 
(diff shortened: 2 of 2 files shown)


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Http/HttpClient/HttpMessageHandlerCommon.cs
```diff
@@ -46,6 +46,10 @@ public static CallTargetState OnMethodBegin<TTarget, TRequest>(TTarget instance,
                     return new CallTargetState(scope);
                 }
             }
+            else
+            {
+                headers.Remove(HttpHeaderNames.TracingEnabled);
```
andrewlock, line 51: This is nice in theory, but it could run into potential issues 🤔 e.g. there are often nested handlers, and if we remove this header from the first one, the _innermost_ handler may create a span 🤔 I think instead we may need to implement the "black hole" span approach which has been discussed a few times, but which is non-trivial to implement at the moment 🤔

williamscs: Ok! Was worth a shot. I can close this out for now. But before I do...

Do you have any recommendations for folks in my position? I've tried delegating handlers as the PrimaryHttpMessageHandler, HttpClientHandler, but the headers are still present. I'd really like to have tracing enabled for my other services and the one I'm trying to instrument is a bit of a monolith, so there are a lot of other http requests coming in and out of it.

andrewlock: Unfortunately, currently, the tracing (span generation) of HttpClient spans is tied to propagation. Assuming you _do_ want to see spans for the requests, you just _don't_ want to include the headers in the outgoing request, I think there's currently only one avenue:

- Disable tracing (and propagation) for the Url's you _don't_ want to trace by setting `DD_TRACE_HTTP_CLIENT_EXCLUDED_URL_SUBSTRINGS`, e.g. `DD_TRACE_HTTP_CLIENT_EXCLUDED_URL_SUBSTRINGS=some.domain.com,some.other.domain/and/path`.
- _Manually_ create a span around the calls to these `HttpClient` requests to emulate the span that you _would_ have got.

The need to use the manual span is obviously a bit hacky, but I think it's the only real workaround available at the moment. Hope that helps, and we'll bear this request in mind as a possible new feature in the future, thanks!

williamscs: `DD_TRACE_HTTP_CLIENT_EXCLUDED_URL_SUBSTRINGS` is exactly what I needed! Totally worked. Didn't know that existed, for what it's worth, but I see it in the code now. Thanks a ton!

williamscs:
Doesn't fit the current paradigm: https://github.com/DataDog/dd-trace-dotnet/pull/7367#discussion_r2262397750

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Http/HttpClient/HttpMessageHandlerCommon.cs
andrewlock: Cool, glad that solved your issue 🙂

PR #7368, opened 2025-08-08
Fix broken S3 integration

andrewlock:
## Summary of changes

The latest version of S3 is broken

## Reason for change

This is the same issue we've seen several times, the s3 integration is broken, because we're ducktyping the field (which changes) instead of the public property

## Implementation details

Update _all_ the S3 integrations to switch to using properties instead of the backing field, because it's just a matter of time before they all break (I manually checked they're all valid).

Also fixed the nullability, which was incorrect (they can all technically be null).

And fixed potential null ref issue because of that related to span pointers.

## Test coverage

This is the test (I bumped the package versions) but will also do an all-versions test


bouwkast (approved):
Thanks!

PR #7370, opened 2025-08-08
Ignore rejit timeout errors in smoke tests and report a metric

andrewlock:
## Summary of changes

Ignore the "rejit timed out" errors in smoke tests, but report a metric

## Reason for change

We're still seeing flake from this. We don't want to increase the timeout further at this stage, so instead we want to allow it, and report a metric so that we can track the issue

## Implementation details

Added a helper for sending a metric, and added a "reportablePatterns" list for the smoke tests.

## Test coverage

This is the test, will see if any of the jobs hit it and 🤞 

## Other details

We will set up a dashboard to monitor these (and other skip metrics) as part of the pre-release process

### DIFF
--- a/tracer/build/_build/MetricHelper.cs
+++ b/tracer/build/_build/MetricHelper.cs
@@ -31,7 +31,7 @@ public static class MetricHelper
                              "ci.stage:{{SanitizeTagValue(Environment.GetEnvironmentVariable("DD_LOGGER_SYSTEM_STAGEDISPLAYNAME"))}}",
                              "ci.job:{{SanitizeTagValue(Environment.GetEnvironmentVariable("DD_LOGGER_SYSTEM_JOBDISPLAYNAME"))}}",
                              "git.branch:{{SanitizeTagValue(Environment.GetEnvironmentVariable("DD_LOGGER_BUILD_SOURCEBRANCH"))}}",
-                             "error_reason: {{SanitizeTagValue(errorReason)}}"
+                             "error_reason:{{SanitizeTagValue(errorReason)}}"
                          """;
 
             return $$"""


tracer/build/_build/Build.Steps.cs
```diff
@@ -2453,13 +2453,13 @@ string NormalizedPath(AbsolutePath ap)
                new(@".*Some errors were found while applying waf configuration \(RulesFile: rasp-rule-set.json\).*", RegexOptions.Compiled),
            };
 
-           CheckLogsForErrors(knownPatterns, allFilesMustExist: false, minLogLevel: LogLevel.Error);
+           await CheckLogsForErrors(knownPatterns, allFilesMustExist: false, minLogLevel: LogLevel.Error, new ());
```
lucaspimentel, line 2456: Coding-style nit, but I don't like the `new ()` without the explicit type unless the type is obvious. I'm looking at this line and have no idea what the type is 😓

(otoh, I felt the same way about `var`, but I've pretty much ~given up~ embraced using it everywhere, even when the type is not obvious, so maybe just give me some time)

tracer/build/_build/Build.Steps.cs
```diff
@@ -2565,6 +2575,15 @@ private void CheckLogsForErrors(List<Regex> knownPatterns, bool allFilesMustExis
                      || nativeProfilerErrors.Count != 0
                      || nativeLoaderErrors.Count != 0;
 
+        if (reportableMetrics.Count > 0)
+        {
+            Logger.Warning("Found reportable (but ignored) problems in the logs");
+            await MetricHelper.SendReportableErrorMetrics(Logger.Logger, reportableMetrics);
+            // TODO: Remove this before merge, this is just to find it
```
lucaspimentel: Reminder: remove this before merge 😉

tracer/build/_build/Build.Steps.cs
andrewlock: I agree with you in general. Tbh, rather than listing the full, ugly as hell, `List<(string IgnoreReasonTag, Regex Regex)>` type, I _should_ have just done this 🙂 

```suggestion
           await CheckLogsForErrors(knownPatterns, allFilesMustExist: false, minLogLevel: LogLevel.Error, reportablePatterns: new ());
```

lucaspimentel: yeah, that works, too
especially with that ugly type name 😅

PR #7371, opened 2025-08-08
Add the version tag to every span

andrewlock:
## Summary of changes

Adds the version tag to every span

## Reason for change

The agent does this anyway, and sometimes it gets it wrong!

## Implementation details

Remove an `if` and flagellate the snapshots

## Test coverage

This is the test, as long as the standard set pass, I'm happy enough to believe it covers everything


zacharycmontoya:
Can you also update this span tag assertion line so our structured SpanMetadataRules testing enforces that every single span we test for our integrations has a `version` tag? https://github.com/DataDog/dd-trace-dotnet/blob/455ad525e99a198acd81d73ee5669475357ea02c/tracer/test/Datadog.Trace.TestHelpers/SpanTagAssertion.cs#L28

andrewlock:
> Can you also update this span tag assertion line so our structured SpanMetadataRules testing enforces that every single span we test for our integrations has a version tag?

Actually, no, because this isn't the case - it's only the case _if_ the customer has set the version 🙂

PR #7374, opened 2025-08-08
fix: Make OTLP metrics snapshot testing more stable

zacharycmontoya:
## Summary of changes
Makes the OTLP metrics snapshots more stable by adding ordering, doing some deduplication (that wasn't reproduced locally), and consolidating data series by their identifying Resources/Scopes

## Reason for change
We had to mark this test as Flaky, but we really want to make sure it's stable so we can correctly do regression tests against it.

## Implementation details
Processes the OTLP metrics data after receiving it from the mock tracer agent. This includes:
- Grouping multiple series of ScopeMetrics into one series if their Resources match (this is expected and asserted to be true)
- Keeping only one metric data point from a Metric with the same Resource & InstrumentationScope (should fix the issue with multiple exports of the same asynchronous counter)
- Scrubs the `telemetry.sdk.version` attribute with a dummy value so we can successfully test against multiple package versions

## Test coverage
This is the test coverage.

### DIFF
--- a/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/OpenTelemetrySdkTests.cs
+++ b/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/OpenTelemetrySdkTests.cs
@@ -238,7 +238,8 @@ namespace Datadog.Trace.ClrProfiler.IntegrationTests
                 // Group the scope metrics by the resource metrics and schema URL (should only be one unique combination)
                 var resourceMetricsByResource = metricRequests
                                         .SelectMany(r => r.MetricsData.ResourceMetrics)
-                                        .GroupBy(r => new Tuple<global::OpenTelemetry.Proto.Resource.V1.Resource, string>(r.Resource, r.SchemaUrl)).ToList();
+                                        .GroupBy(r => new Tuple<global::OpenTelemetry.Proto.Resource.V1.Resource, string>(r.Resource, r.SchemaUrl))
+                                        .ToList();
                 resourceMetricsByResource.Should().ContainSingle();
 
                 var resourceMetrics = new List<object>();
@@ -250,15 +251,23 @@ namespace Datadog.Trace.ClrProfiler.IntegrationTests
                     // This may result in multiple entries for metrics that are repeated multiple times before the test exits
                     var scopeMetricsByResource = resourceMetricByResource
                                             .SelectMany(r => r.ScopeMetrics)
-                                            .GroupBy(r => new Tuple<global::OpenTelemetry.Proto.Common.V1.InstrumentationScope, string>(r.Scope, r.SchemaUrl)).ToList();
+                                            .GroupBy(r => new Tuple<global::OpenTelemetry.Proto.Common.V1.InstrumentationScope, string>(r.Scope, r.SchemaUrl))
+                                            .OrderBy(group => group.Key.Item1.Name)
+                                            .ToList();
 
                     var scopeMetrics = new List<object>();
                     foreach (var scopeMetricByResource in scopeMetricsByResource)
                     {
+                        var metrics = scopeMetricByResource
+                                        .SelectMany(r => r.Metrics)
+                                        .OrderBy(metric => metric.Name)
+                                        .ToList();
+                        OtlpHelper.DeduplicateOtlpMetrics(metrics);
+
                         scopeMetrics.Add(new
                         {
                             Scope = scopeMetricByResource.Key.Item1,
-                            Metrics = scopeMetricByResource.SelectMany(r => r.Metrics),
+                            Metrics = metrics,
                             SchemaUrl = scopeMetricByResource.Key.Item2
                         });
                     }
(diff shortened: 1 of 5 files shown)


tracer/test/Datadog.Trace.TestHelpers/MockTracerAgent.cs
```diff
@@ -1704,13 +1698,13 @@ private async Task HandleUdsTraces()
...
         {
-            public MockOtlpRequest(string pathAndQuery, Dictionary<string, List<string>> headers, byte[] body, string contentType, object deserializedData)
+            public MockOtlpRequest(string pathAndQuery, Dictionary<string, List<string>> headers, byte[] body, string contentType, MetricsData metricsData)
             {
                 PathAndQuery = pathAndQuery;
                 Headers = headers;
                 Body = body;
                 ContentType = contentType;
-                DeserializedData = deserializedData;
+                MetricsData = metricsData;
```
andrewlock, line 1707: ooooh, I didn't realise we already had the means for deserializing this vendored in, completely missed the part above where it's deserialized 😄

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/OpenTelemetrySdkTests.cs
```diff
@@ -235,22 +234,68 @@ public async Task SubmitsOtlpMetrics(string packageVersion)
...
-                // count values should be 0 in subsequent payloads etc.
-                var snapshotPayload = metricRequests
-                    .Select(r => r.DeserializedData)
-                    .ToList();
+                // Group the scope metrics by the resource metrics and schema URL (should only be one unique combination)
+                var resourceMetricsByResource = metricRequests
+                                        .SelectMany(r => r.MetricsData.ResourceMetrics)
+                                        .GroupBy(r => new Tuple<global::OpenTelemetry.Proto.Resource.V1.Resource, string>(r.Resource, r.SchemaUrl))
+                                        .ToList();
+                resourceMetricsByResource.Should().ContainSingle();
```
andrewlock: nit: if you _don't_ want it as an array (to remove the unnecessary "foreach") you can do this:
```suggestion
                var resourceMetricByResource = metricRequests
                                        .SelectMany(r => r.MetricsData.ResourceMetrics)
                                        .GroupBy(r => new Tuple<global::OpenTelemetry.Proto.Resource.V1.Resource, string>(r.Resource, r.SchemaUrl))
                                        .ToList()
                                        .Should()
                                        .ContainSingle()
                                        .Subject;
```

And `resourceMetricByResource` will contain the one value

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/OpenTelemetrySdkTests.cs
```diff
@@ -235,22 +234,68 @@ public async Task SubmitsOtlpMetrics(string packageVersion)
...
+                // Although there's only one resource, let's still emit snapshot data in the expected array format
+                foreach (var resourceMetricByResource in resourceMetricsByResource)
+                {
+                    // Group the individual metrics by scope metric and schema URL (should only be one unique combination since we're only using one ActivitySource)
+                    // This may result in multiple entries for metrics that are repeated multiple times before the test exits
+                    var scopeMetricsByResource = resourceMetricByResource
+                                            .SelectMany(r => r.ScopeMetrics)
+                                            .GroupBy(r => new Tuple<global::OpenTelemetry.Proto.Common.V1.InstrumentationScope, string>(r.Scope, r.SchemaUrl))
+                                            .OrderBy(group => group.Key.Item1.Name)
+                                            .ToList();
```
andrewlock: nit: only iterated once, so can ditch the `ToList()`
```suggestion
                                            .OrderBy(group => group.Key.Item1.Name);
```

tracer/test/snapshots/OpenTelemetrySdkTests.SubmitsOtlpMetrics.verified.txt
andrewlock, line 1: The diff looks pretty mangled on GitHub, I'll take your word that it looks good to you!

andrewlock (approved):
Thanks for doing what I couldn't be bothered to 🙈

tracer/test/snapshots/OpenTelemetrySdkTests.SubmitsOtlpMetrics.verified.txt
zacharycmontoya: The changes look much smaller when you view without whitespace differences! Sorry, should have mentioned that in the PR description

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/OpenTelemetrySdkTests.cs
zacharycmontoya: Thanks! Updated in 92a9b7395919b89e87b6268c691dfb5ff00b8465

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/OpenTelemetrySdkTests.cs
zacharycmontoya: Thanks for the suggestion! Applied in 92a9b7395919b89e87b6268c691dfb5ff00b8465

PR #7375, opened 2025-08-11
Removes exception throwing during shutdown of dynamic instrumentation

andrewlock:
## Summary of changes

Removes exception throwing in "success" path, even when DI is disabled

## Reason for change

#7304 introduced a bunch of changes, but the use of `CancellationTokenSource` resulted in exceptions being thrown in the "happy" shutdown path, which can cause crashes in buggy versions of .NET (i.e. all of them, currently)

## Implementation details

Replace usages of `TaskCompletionSource<bool>` with `CancellationTokenSource`

## Test coverage

Covered by existing - checked the execution tests to make sure the exception count has gone back own.

## Other details

Discovered some additional issues that need to be addressed I think. Most importantly, `SafeDisposal` looks like a band-aid due to unclear lifetime management. We should refactor the code to not require it i.e. `Disposing` types should be safe and should not throw (regardless of whether we catch the exception).

Additionally, this looks like it does quite a lot of work even when DI is disabled. I would suggest we refactor it so that it doesn't do a bunch of work in cases where it's never enabled.

### DIFF
--- a/tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
+++ b/tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
@@ -38,19 +38,16 @@ namespace Datadog.Trace.Debugger
 
         private static volatile bool _discoveryServiceReady;
         private readonly SemaphoreSlim _semaphore;
-        private readonly CancellationTokenSource _cancellationToken;
-        private volatile bool _isShuttingDown;
+        private readonly TaskCompletionSource<bool> _processExit = new(TaskCreationOptions.RunContinuationsAsynchronously);
         private int _initialized;
 
         private DebuggerManager(DebuggerSettings debuggerSettings, ExceptionReplaySettings exceptionReplaySettings)
         {
             _initialized = 0;
             _discoveryServiceReady = false;
-            _isShuttingDown = false;
             _semaphore = new SemaphoreSlim(1, 1);
             DebuggerSettings = debuggerSettings;
             ExceptionReplaySettings = exceptionReplaySettings;
-            _cancellationToken = new CancellationTokenSource();
             ServiceName = string.Empty;
         }
 
@@ -70,7 +67,7 @@ namespace Datadog.Trace.Debugger
 
         internal string ServiceName { get; private set; }
 
-        private async Task<bool> WaitForDiscoveryServiceAsync(IDiscoveryService discoveryService, CancellationToken cancellationToken)
+        private async Task<bool> WaitForDiscoveryServiceAsync(IDiscoveryService discoveryService, Task shutdownTask)
         {
             if (_discoveryServiceReady)
             {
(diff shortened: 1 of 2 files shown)


tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -248,8 +245,20 @@ private async Task UpdateProductsState(TracerSettings tracerSettings, DebuggerSe
             bool semaphoreAcquired = false;
             try
             {
-                semaphoreAcquired = await _semaphore.WaitAsync(TimeSpan.FromSeconds(30), _cancellationToken.Token).ConfigureAwait(false);
-                if (!semaphoreAcquired || _isShuttingDown)
+                var attemptsRemaining = 6; // 5*6 = 30s timeout
+                while (!_processExit.Task.IsCompleted && !semaphoreAcquired && attemptsRemaining > 0)
```
andrewlock, line 249: This feels nasty, but I couldn't see a way to wait with a timeout _or_ process exit safely, incase the semaphore was acquired in the background, but could refactor it to use `Task.WhenAny()` and ignore that potentially, (as we're already exiting at that point)?

tracer/src/Datadog.Trace/Debugger/Symbols/SymbolsUploader.cs
```diff
@@ -118,8 +117,7 @@ private void ConfigurationChanged(AgentConfiguration configuration)
 
             _symDbEndpoint = configuration.SymbolDbEndpoint;
             _discoveryServiceSemaphore.Release(1);
-            _discoveryService!.RemoveSubscription(ConfigurationChanged);
-            _discoveryService = null;
+            _discoveryService.RemoveSubscription(ConfigurationChanged);
```
andrewlock, line 120: This didn't seem safe or worth the risk, in case the `ConfigurationChange` is invoked twice concurrently

tracer/src/Datadog.Trace/Debugger/Symbols/SymbolsUploader.cs
```diff
@@ -167,10 +165,13 @@ private async Task ProcessItemAsync(Assembly assembly)
             bool semaphoreAcquired = false;
             try
             {
-                await _assemblySemaphore.WaitAsync(_cancellationToken.Token).ConfigureAwait(false);
-                semaphoreAcquired = true;
+                var acquireTimeout = TimeSpan.FromSeconds(5);
+                while (!_processExit.Task.IsCompleted && !semaphoreAcquired)
+                {
+                    semaphoreAcquired = await _assemblySemaphore.WaitAsync(acquireTimeout).ConfigureAwait(false);
+                }
```
bouwkast, line 172: I see in `DebuggerManager` you have a upper limit on the runs should there be something similar here?
```csharp
                var attemptsRemaining = 6; // 5*6 = 30s timeout
                while (!_processExit.Task.IsCompleted && !semaphoreAcquired && attemptsRemaining > 0)
                {
                    attemptsRemaining--;
                    semaphoreAcquired = await _semaphore.WaitAsync(TimeSpan.FromSeconds(5)).ConfigureAwait(false);
                }
```

andrewlock: I'm just replicating what was done before to be honest, so I don't think so - this wants to wait forever currently 🙂

dudikeleti (approved):
Thanks!

PR #7380, opened 2025-08-12
Detect libdatadog available

anna-git:
## Summary of changes

Expose the native check for the libdatadog library to the managed side and use it instead of !IsServerless && ClrProfiler.Attached


tracer/src/Datadog.Trace/ClrProfiler/NativeMethods.cs
```diff
@@ -295,6 +308,9 @@ private static class NonWindows
             [DllImport("Datadog.Tracer.Native")]
             public static extern bool IsProfilerAttached();
 
+            [DllImport("Datadog.Tracer.Native.")]
```
lucaspimentel, line 311:
```suggestion
            [DllImport("Datadog.Tracer.Native")]
```

PR #7381, opened 2025-08-12
Service Bus

pablomartinezbernardo:

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
@@ -6,12 +6,18 @@
 #if !NETFRAMEWORK
 using System;
 using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Text;
 using Datadog.Trace.ClrProfiler.CallTarget;
 using Datadog.Trace.Configuration;
 using Datadog.Trace.DuckTyping;
+using Datadog.Trace.Headers;
 using Datadog.Trace.Logging;
 using Datadog.Trace.Propagators;
 using Datadog.Trace.Tagging;
+using Datadog.Trace.Vendors.Newtonsoft.Json;
 
 #nullable enable
 
@@ -21,7 +27,7 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Azure.Functions
     {
         public const string IntegrationName = nameof(Configuration.IntegrationId.AzureFunctions);
 
-        public const string OperationName = "azure_functions.invoke";
+        public const string OperationName = "azure.functions.invoke";
         public const string SpanType = SpanTypes.Serverless;
         public const IntegrationId IntegrationId = Configuration.IntegrationId.AzureFunctions;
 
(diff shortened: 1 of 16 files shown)


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -21,7 +27,7 @@ internal static class AzureFunctionsCommon
     {
         public const string IntegrationName = nameof(Configuration.IntegrationId.AzureFunctions);
 
-        public const string OperationName = "azure_functions.invoke";
+        public const string OperationName = "azure.functions.invoke";
```
lucaspimentel, line 30: I don't think we should change this.
1. It's a breaking change: trace stats use operation names, so any metrics, dashboards, monitors, etc relying on this name will stop working.
2. The general convention for operation names is something like `<product>.<action>` (single dot).

examples (ignore the `-` vs `_`, that's a separate issue 😅)
```
aspnet.request
aspnet-mvc.request
aspnet-webapi.request
aspnet_core.request
aspnet_core_mvc.request
```

(There are many operation names that don't follow this convention. Mistakes were made that are now hard to fix because of point 1 above)

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -217,7 +224,7 @@ public static CallTargetState OnIsolatedFunctionBegin<T>(T functionContext)
                     {
                         _ when type.Equals("httpTrigger", StringComparison.OrdinalIgnoreCase) => "Http", // Microsoft.Azure.Functions.Worker.Extensions.Http
                         _ when type.Equals("timerTrigger", StringComparison.OrdinalIgnoreCase) => "Timer", // Microsoft.Azure.Functions.Worker.Extensions.Timer
-                        _ when type.Equals("serviceBus", StringComparison.OrdinalIgnoreCase) => "ServiceBus", // Microsoft.Azure.Functions.Worker.Extensions.ServiceBus
+                        _ when type.Equals("serviceBusTrigger", StringComparison.OrdinalIgnoreCase) => "ServiceBus", // Microsoft.Azure.Functions.Worker.Extensions.ServiceBus
```
lucaspimentel, line 227: Were we never hitting this branch?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -330,6 +356,270 @@ private static PropagationContext ExtractPropagatedContextFromHttp<T>(T context,
...
+            {
+                Log.Error(ex, "Error extracting propagated context from ServiceBus binding");
+                return default;
+            }
+        }
+
+        private static SpanContext? ExtractSpanContextFromProperties(Dictionary<string, object> userProperties)
+        {
+            var headerAdapter = new ServiceBusUserPropertiesHeadersCollection(userProperties);
+            var extractedContext = Tracer.Instance.TracerManager.SpanContextPropagator.Extract(headerAdapter);
```
lucaspimentel, line 436: I think these days we prefer to use the overloads that take lambdas instead of implementing `IHeadersCollection`. That was the old way of doing this.

lucaspimentel: Oh, I see you added a `ContextPropagation : ICarrierGetter<...>`.

PR #7383, opened 2025-08-12
Fix for sending duplicate logs when using Agentless Logging in Azure Function host

bouwkast:
## Summary of changes

Disables agentless logging (for the Azure Functions Host Process) if we detect that we are instrumenting an Azure Functions host process. Previously this would cause us to duplicate every log sent from the worker process.

This bug fix can be reverted by setting the new configuration key `DD_LOGS_DIRECT_SUBMISSION_AZURE_FUNCTIONS_HOST_ENABLED` to `true`.

## Reason for change

If customers are running with the Isolated Azure Function model we will be instrumenting two applications, one is the function host process the other is the worker process. If they have direct log submission enabled the function host ends up duplicating the logs from the function process which results in us shipping two nearly identical logs.

This behavior isn't ideal as the duplicate log is likely not valuable so we've disabled agentless logging in the azure functions host process.

## Implementation details

Added `IsRunningInAzureFunctionsHost()` to `EnvironmentHelpers.cs` which allows for a rough detection of whether we are running on the function host using the following logic:

- Is `FUNCTIONS_WORKER_RUNTIME` present AND set to `dotnet-isolated`?
- Are both `--functions-worker-id` or `--workerId` NOT present in the command line?

If both are true we treat that scenario as being running in the function host - otherwise we are likely the worker process

I wasn't able to find a more robust way of checking, but when looking at the various log output that I had the `--functions-worker-id` or `--workerId` seemed to always be called by the function host.

```
[2025-10-03T16:01:42.901Z] Reading functions metadata (Worker)
[2025-10-03T16:01:47.176Z] {
[2025-10-03T16:01:47.177Z]   "ProcessId": 71080,
[2025-10-03T16:01:47.178Z]   "RuntimeIdentifier": "win-x64",
[2025-10-03T16:01:47.179Z]   "WorkerVersion": "2.0.0.0",
[2025-10-03T16:01:47.180Z]   "ProductVersion": "2.0.0\u002Bd8b5fe998a8c92819b8ee41d2569d2525413e9c5",
[2025-10-03T16:01:47.181Z]   "FrameworkDescription": ".NET 9.0.9",
[2025-10-03T16:01:47.182Z]   "OSDescription": "Microsoft Windows 10.0.26100",
[2025-10-03T16:01:47.183Z]   "OSArchitecture": "X64",
[2025-10-03T16:01:47.184Z]   "CommandLine": "C:\\Users\\steven.bouwkamp\\source\\repos\\dd-trace-dotnet\\artifacts\\bin\\Samples.AzureFunctions.V4Isolated.AspNetCore\\debug_net9.0\\Samples.AzureFunctions.V4Isolated.AspNetCore.dll --host 127.0.0.1 --port 65401 --workerId e94d23fd-cd3c-4780-a3e3-4980d7b0f644 --requestId 6dba68ac-1954-466a-aeb4-9570cc9b12c2 --grpcMaxMessageLength 2147483647 --functions-uri http://127.0.0.1:65401/ --functions-worker-id e94d23fd-cd3c-4780-a3e3-4980d7b0f644 --functions-request-id 6dba68ac-1954-466a-aeb4-9570cc9b12c2 --functions-grpc-max-message-length 2147483647"
[2025-10-03T16:01:47.185Z] }
```

- Added `DD_LOGS_DIRECT_SUBMISSION_AZURE_FUNCTIONS_HOST_ENABLED` that defaults to `false` to disable the duplicate logs from being sent.

## Test coverage

- Added a new test project and test for having host logs enabled / disabled.
  - The reason why I added a new test project instead of re-using an existing one was because when I re-ran the Function application in our tests multiple times the `func.exe` would fail to obtain a lock and would need to wait some period of time to for recovery after each subsequent run of the same application. I think this is because we end the `func.exe` process with a `Kill()`. Making a new project wasn't ideal but was a quick and simple workaround.

## Other details
Fixes SLES-2364

### DIFF
--- a/tracer/test/test-applications/azure-functions/Samples.AzureFunctions.V4Isolated.HostLogsDisabled/Samples.AzureFunctions.V4Isolated.HostLogsDisabled.csproj
+++ b/tracer/test/test-applications/azure-functions/Samples.AzureFunctions.V4Isolated.HostLogsDisabled/Samples.AzureFunctions.V4Isolated.HostLogsDisabled.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net6.0;net7.0;net8.0;net9.0</TargetFrameworks>
+    <TargetFrameworks>net6.0;net7.0;net8.0;net9.0;net10.0</TargetFrameworks>
     <AzureFunctionsVersion>v4</AzureFunctionsVersion>
     <OutputType>Exe</OutputType>
     <ImplicitUsings>enable</ImplicitUsings>


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsHostDetector.cs
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    private static bool DetectIfFunctionsHost()
+    {
+        try
+        {
+            var functionsRuntime = Environment.GetEnvironmentVariable("FUNCTIONS_WORKER_RUNTIME");
+            if (string.IsNullOrEmpty(functionsRuntime))
+            {
+                return false; // we are the worker process, not the host
```
lucaspimentel: > we are the worker process, not the host

I think `FUNCTIONS_WORKER_RUNTIME` is set on both the host and worker processes?

lucaspimentel: Btw we have some of these checks here:
https://github.com/DataDog/dd-trace-dotnet/blob/835d79f94261ca1cbf5177c50ebc442329bbb057/tracer/src/Datadog.Trace/Util/EnvironmentHelpers.cs#L110-L121

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsHostDetector.cs
```diff
@@ -0,0 +1,69 @@
...
+            }
+
+            // Check if we are isolated
+            if (!string.Equals(functionsRuntime, "dotnet-isolated", StringComparison.OrdinalIgnoreCase))
```
lucaspimentel: Maybe add these new checks to `EnvironmentHelpers.cs` and the constants to `ConfigurationKeys.AzureFunctions.cs` for now? We need to do some cleanup around reading all these env vars, but it will be easier if they are all in the same place.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsHostDetector.cs
```diff
@@ -0,0 +1,69 @@
...
+                // We're in the worker process, not the host
+                return false;
+            }
+
+            // at this point we can assume that we are in the host process
+            // but again this is just us checking the environment variable(s)
+            // Alternatively, we can check the process name and loaded assemblies if we want to be even more thorough
+            // like so below, but unsure if it is worth it as
+            // this is still guarded by having another env var set to disable ILogger
+            var process = Process.GetCurrentProcess();
```
lucaspimentel: We have a helper in `ProcessHelpers.GetCurrentProcessName()` to avoid allocating a new `Process` instance.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsHostDetector.cs
```diff
@@ -0,0 +1,69 @@
...
+            var processName = process.ProcessName.ToLowerInvariant();
+            var isKnownHostProcess = processName == "func" || processName.Contains("microsoft.azure.webjobs.script.webhost") || processName == "webhost";
```
lucaspimentel: The string comparisons should be case-insensitive.

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.DirectLogSubmission.cs
```diff
@@ -79,6 +79,13 @@ internal static class DirectLogSubmission
             /// </summary>
             /// <seealso cref="DirectLogSubmissionSettings.BatchPeriod"/>
             public const string BatchPeriodSeconds = "DD_LOGS_DIRECT_SUBMISSION_BATCH_PERIOD_SECONDS";
+
+            /// <summary>
+            /// Configuration key to disable direct log submission for Azure Functions host.
+            /// Default value is <c>false</c>.
+            /// </summary>
+            /// <seealso cref="DirectLogSubmissionSettings.DisableForAzureFunctionsHost"/>
+            public const string DisableForAzureFunctionsHost = "DD_LOGS_DIRECT_SUBMISSION_DISABLE_FOR_AZURE_FUNCTIONS_HOST";
```
lucaspimentel: Our config naming convention is usually `DD_<SOMETHING>_ENABLED`. Consider calling this
```
DD_LOGS_DIRECT_SUBMISSION_AZURE_FUNCTIONS_HOST_ENABLED
```
(I prefer this personally to avoid the confusing double negative in `DISABLED=false`)

If you change the name, remember to flip the default value, too (unless we want to ignore the host logs by default).

andrewlock: Agreed, we're trying to be more rigid about these conventions too

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsHostDetector.cs
bouwkast: Ah yes, my comment here is not correct and should be that we aren't in AzureFunctions, but thanks for the EnvironmentHelpers link, I'm shifting stuff to there as it makes more sense

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsHostDetector.cs
bouwkast: Okay I changed this all up and put the logic into the `EnvironmentHelpers.cs`

I opted to go with a different way rather than checking the Process name and instead am relying on reading the command line to see when the host process is passing the ID

e.g. something like this is what I see consistently

```
... Samples.AzureFunctions.V4Isolated.HostLogsDisabled.dll
  --workerId 35a1c218-... 
  --functions-worker-id 35a1c218-...
  --functions-uri http://127.0.0.1:63307/
```

So I check for those instead, but I'm not super sure if that is good practice or not

tracer/test/test-applications/azure-functions/Samples.AzureFunctions.V4Isolated.HostLogsDisabled/AllTriggers.cs
bouwkast, line 1: I copied the files because I was running into frequent "File in use" when attempting to build / run the samples and tests and the only thing I can think of is that there is some form of race condition between linking the various files with the worker extension projects that get generated.

Not really sure what to do or if that was the cause/solution, but it went away after I did this.

tracer/test/Datadog.Trace.Tests/Telemetry/config_norm_rules.json
```diff
@@ -527,6 +527,7 @@
   "DD_LOGS_DIRECT_SUBMISSION_MAX_BATCH_SIZE": "logs_direct_submission_max_batch_size",
   "DD_LOGS_DIRECT_SUBMISSION_MAX_QUEUE_SIZE": "logs_direct_submission_max_queue_size",
   "DD_LOGS_DIRECT_SUBMISSION_BATCH_PERIOD_SECONDS": "logs_direct_submission_batch_period_seconds",
+  "DD_LOGS_DIRECT_SUBMISSION_AZURE_FUNCTIONS_HOST_ENABLED": "logs_direct_submission_azure_functions_host_enabled",
```
bouwkast, line 531: Blocking -> need to create the PR for this.

tracer/src/Datadog.Trace/Logging/DirectSubmission/DirectLogSubmissionSettings.cs
```diff
@@ -98,6 +99,9 @@ public DirectLogSubmissionSettings(IConfigurationSource? source, IConfigurationT
 
             BatchPeriod = TimeSpan.FromSeconds(seconds);
 
+            AzureFunctionsHostEnabled = config.WithKeys(ConfigurationKeys.DirectLogSubmission.AzureFunctionsHostEnabled)
+                                              .AsBool(true);
```
andrewlock: I wonder if this should be `false` (effectively making this a bug-fix that you can revert)?

tracer/src/Datadog.Trace/Util/EnvironmentHelpers.cs
```diff
@@ -130,6 +130,24 @@ public static bool IsUsingAzureAppServicesSiteExtension()
...
+        /// </summary>
+        public static bool IsRunningInAzureFunctionsHost()
+        {
+            var cmd = Environment.CommandLine ?? string.Empty;
+            var hasWorkerId = cmd.IndexOf("--functions-workerid", StringComparison.OrdinalIgnoreCase) >= 0 ||
+                              cmd.IndexOf("--workerId", StringComparison.OrdinalIgnoreCase) >= 0;
+
+            return IsAzureFunctions()
+                   && string.Equals(GetEnvironmentVariable(ConfigurationKeys.AzureFunctions.FunctionsWorkerRuntime, defaultValue: string.Empty), "dotnet-isolated", StringComparison.Ordinal)
+                   && !hasWorkerId;
```
andrewlock, line 156: nit: you've already done the "work" of reading workerId, so may as well check that first
```suggestion
            return !hasWorkerId
                   && IsAzureFunctions()
                   && string.Equals(GetEnvironmentVariable(ConfigurationKeys.AzureFunctions.FunctionsWorkerRuntime, defaultValue: string.Empty), "dotnet-isolated", StringComparison.Ordinal);
```

tracer/test/test-applications/azure-functions/Samples.AzureFunctions.V4Isolated/Program.cs
```diff
@@ -1,7 +1,13 @@
 using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
 
 var host = new HostBuilder()
     .ConfigureFunctionsWorkerDefaults()
+        .ConfigureLogging(logging =>
+        {
+            logging.AddConsole();
+            logging.SetMinimumLevel(LogLevel.Trace);
+        })
```
andrewlock, line 10: You have this in the "disabled" case, do you want them to be the same?
```suggestion
            logging.AddFilter("*", LogLevel.Trace);
        })
```

andrewlock (approved):
LGTM, thanks!

tracer/src/Datadog.Trace/Logging/DirectSubmission/DirectLogSubmissionSettings.cs
bouwkast: @lucaspimentel thoughts on this?

I don't think these host logs are necessarily useful, but it could be a breaking change if someone is using them.

lucaspimentel: My thoughts exactly.

andrewlock: Every bug fix is _technically_ a breaking change 😉 

image

IMO the current behaviour is "wrong" because it causes duplicate logs 🤷‍♂️ But my level of care is low, so will def defer to @lucaspimentel as code ower😄

lucaspimentel:
@codex review

lucaspimentel:
@codex review

tracer/src/Datadog.Trace/Logging/DirectSubmission/DirectLogSubmissionSettings.cs
lucaspimentel: > the current behaviour is "wrong" because it causes duplicate logs

Hmm. Before this PR, we duplicate logs from the worker process, not logs from the host process. This PR fix that duplication, but in the process we also loose all logs from the host process.

Maybe the better solution is to disable agentless logging in the worker process instead of the host process, so we can capture all logs (once) in the host process. This is assuming all the logs from the worker are sent to the host so we can capture them there.

tracer/src/Datadog.Trace/Util/EnvironmentHelpers.cs
```diff
@@ -130,6 +130,24 @@ public static bool IsUsingAzureAppServicesSiteExtension()
...
+        /// This method reads environment variables directly and bypasses the configuration system.
+        /// </summary>
+        public static bool IsRunningInAzureFunctionsHost()
+        {
+            var cmd = Environment.CommandLine ?? string.Empty;
+            var hasWorkerId = cmd.IndexOf("--functions-workerid", StringComparison.OrdinalIgnoreCase) >= 0 ||
+                              cmd.IndexOf("--workerId", StringComparison.OrdinalIgnoreCase) >= 0;
+
+            return !hasWorkerId
+                   && IsAzureFunctions()
```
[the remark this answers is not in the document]
lucaspimentel: Does `hasWorkerId` mean we are running on the host or the worker? I think robot has it backwards. Might want to clarify that in the xml comments above.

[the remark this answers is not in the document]
bouwkast: Clarified, my documentation comment was incorrect (as was the PR description), but the code was correct.

The worker process will have those flags (actually mis-spelled one 😢 )
The host process will **not** have those flags

At least per what I see locally

```
[2025-10-03T16:01:42.901Z] Reading functions metadata (Worker)
[2025-10-03T16:01:47.176Z] {
[2025-10-03T16:01:47.177Z]   "ProcessId": 71080,
[2025-10-03T16:01:47.178Z]   "RuntimeIdentifier": "win-x64",
[2025-10-03T16:01:47.179Z]   "WorkerVersion": "2.0.0.0",
[2025-10-03T16:01:47.180Z]   "ProductVersion": "2.0.0\u002Bd8b5fe998a8c92819b8ee41d2569d2525413e9c5",
[2025-10-03T16:01:47.181Z]   "FrameworkDescription": ".NET 9.0.9",
[2025-10-03T16:01:47.182Z]   "OSDescription": "Microsoft Windows 10.0.26100",
[2025-10-03T16:01:47.183Z]   "OSArchitecture": "X64",
[2025-10-03T16:01:47.184Z]   "CommandLine": "C:\\Users\\steven.bouwkamp\\source\\repos\\dd-trace-dotnet\\artifacts\\bin\\Samples.AzureFunctions.V4Isolated.AspNetCore\\debug_net9.0\\Samples.AzureFunctions.V4Isolated.AspNetCore.dll --host 127.0.0.1 --port 65401 --workerId e94d23fd-cd3c-4780-a3e3-4980d7b0f644 --requestId 6dba68ac-1954-466a-aeb4-9570cc9b12c2 --grpcMaxMessageLength 2147483647 --functions-uri http://127.0.0.1:65401/ --functions-worker-id e94d23fd-cd3c-4780-a3e3-4980d7b0f644 --functions-request-id 6dba68ac-1954-466a-aeb4-9570cc9b12c2 --functions-grpc-max-message-length 2147483647"
[2025-10-03T16:01:47.185Z] }
```

bouwkast:
@codex review

tracer/src/Datadog.Trace/Logging/DirectSubmission/DirectLogSubmissionSettings.cs
```diff
@@ -98,6 +99,9 @@ public DirectLogSubmissionSettings(IConfigurationSource? source, IConfigurationT
...
+
```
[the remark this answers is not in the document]
bouwkast: Fixed

tracer/test/Datadog.Trace.Tests/Telemetry/config_norm_rules.json
bouwkast: Done in https://github.com/DataDog/dd-go/pull/197334

bouwkast:
Tested this out locally and in real Azure Functions

PR #7386, opened 2025-08-13
chore(trace-exporter): disable data pipeline when SSI is enabled

ganeshnj:
## Reason for change

To avoid any surprises.

## Implementation details

Just like other checks we conditionally disable trace exporter for certain platforms and situations, disable trace exporter when SSI in works.

## Other details

Fixes https://datadoghq.atlassian.net/browse/APMSP-2197

### DIFF
--- a/tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
+++ b/tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
@@ -442,6 +442,17 @@ namespace Datadog.Trace.Configuration
                         $"{ConfigurationKeys.TraceDataPipelineEnabled} is enabled, but libdatadog is not available. Disabling data pipeline.");
                     _telemetry.Record(ConfigurationKeys.TraceDataPipelineEnabled, false, ConfigurationOrigins.Calculated);
                 }
+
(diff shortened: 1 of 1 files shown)


tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -442,6 +442,17 @@ internal TracerSettings(IConfigurationSource? source, IConfigurationTelemetry te
                         $"{ConfigurationKeys.TraceDataPipelineEnabled} is enabled, but libdatadog is not available. Disabling data pipeline.");
                     _telemetry.Record(ConfigurationKeys.TraceDataPipelineEnabled, false, ConfigurationOrigins.Calculated);
                 }
+
+                // On Linux, SSI already utilizes libdatadog. To prevent unexpected behavior,
```
andrewlock: Note that SSI is no longer limited to Linux. It is already shipping for Windows (where it is currently only available for IIS applications). This disables it in that scenario too (which is fine, just being clear)

ganeshnj: true, I had Linux from our conversation in my head, just updated the comment.

PR #7387, opened 2025-08-13
[Service Disc] Change type layout to make it x86 safe

anna-git:
## Summary of changes

Change service discovery result type to work in x86 systems.

## Reason for change

tested locally removing x64 guards it works on  both

## Implementation details

use a  mix of sequential and explicit layouts


andrewlock (approved):
Thanks!

PR #7388, opened 2025-08-13
send snapshots to diagnostics endpoint

dudikeleti:

### DIFF
--- a/tracer/src/Datadog.Trace/Debugger/DebuggerFactory.cs
+++ b/tracer/src/Datadog.Trace/Debugger/DebuggerFactory.cs
@@ -33,7 +33,7 @@ internal class DebuggerFactory
         var snapshotStatusSink = SnapshotSink.Create(debuggerSettings, snapshotSlicer);
         var diagnosticsSink = DiagnosticsSink.Create(serviceName, debuggerSettings);
 
-        var debuggerUploader = CreateSnapshotUploader(discoveryService, debuggerSettings, gitMetadataTagsProvider, GetApiFactory(tracerSettings, false), snapshotStatusSink);
+        var debuggerUploader = CreateSnapshotUploader(discoveryService, debuggerSettings, gitMetadataTagsProvider, GetApiFactory(tracerSettings, true), snapshotStatusSink);
         var diagnosticsUploader = CreateDiagnosticsUploader(discoveryService, debuggerSettings, gitMetadataTagsProvider, GetApiFactory(tracerSettings, true), diagnosticsSink);
         var lineProbeResolver = LineProbeResolver.Create(debuggerSettings.ThirdPartyDetectionExcludes, debuggerSettings.ThirdPartyDetectionIncludes);
         var probeStatusPoller = ProbeStatusPoller.Create(diagnosticsSink, debuggerSettings);
--- a/tracer/src/Datadog.Trace/Debugger/Upload/SnapshotUploadApi.cs
+++ b/tracer/src/Datadog.Trace/Debugger/Upload/SnapshotUploadApi.cs
@@ -27,7 +27,7 @@ namespace Datadog.Trace.Debugger.Upload
             : base(apiRequestFactory, gitMetadataTagsProvider)
         {
             _apiRequestFactory = apiRequestFactory;
-            discoveryService.SubscribeToChanges(c => Endpoint = c.DebuggerEndpoint);
+            discoveryService.SubscribeToChanges(c => Endpoint = c.DiagnosticsEndpoint);
         }
 
         public static SnapshotUploadApi Create(


tracer/src/Datadog.Trace/Debugger/Upload/SymbolUploadApi.cs
```diff
@@ -51,7 +52,7 @@ internal static IBatchUploadApi Create(
         {
             ArraySegment<byte> GetEventMetadataAsArraySegment()
             {
-                var eventMetadata = $@"{{""ddsource"": ""dd_debugger"", ""service"": ""{serviceName}"", ""runtimeId"": ""{Tracer.RuntimeId}""}}";
+                var eventMetadata = $@"{{""ddsource"": ""{DebuggerTags.DDSource}"", ""service"": ""{serviceName}"", ""runtimeId"": ""{Tracer.RuntimeId}"", ""debugger.type"": ""{DebuggerTags.DebuggerType.SymDb}""}}";
```
andrewlock: nit: Raw-interpolated strings are cool because it's basically just "normal" JSON (plus interpolation) 🙂 

```suggestion
                var eventMetadata = $$"""{"ddsource": "{{DebuggerTags.DDSource}}", "service": "{{serviceName}}", "runtimeId": "{{Tracer.RuntimeId}}", "debugger.type": {{DebuggerTags.DebuggerType.SymDb}}}""";
```

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
```diff
@@ -493,10 +497,27 @@ internal void AddSnapshot(ProbeInfo probe, string snapshot)
...
+            {
+                AddLog(probe, snapshot);
+                return;
+            }
+
             _snapshotUploader.Add(probe.ProbeId, snapshot);
             SetProbeStatusToEmitting(probe);
         }
 
+        internal void AddLog(ProbeInfo probe, string log)
```
GreenMatan, line 486: nit: we can mark it as `private`

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
```diff
@@ -493,10 +497,27 @@ internal void AddSnapshot(ProbeInfo probe, string snapshot)
                 return;
             }
 
+            if (!probe.IsFullSnapshot)
```
GreenMatan, line 476: why do you direct only non full snapshot to the LogUploader?

tracer/src/Datadog.Trace/Debugger/Upload/LogUploadApi.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor<LogUploadApi>();
+
+        private LogUploadApi(
+            IApiRequestFactory apiRequestFactory,
+            IDiscoveryService discoveryService,
+            IGitMetadataTagsProvider gitMetadataTagsProvider)
+            : base(apiRequestFactory, gitMetadataTagsProvider)
+        {
+            discoveryService.SubscribeToChanges(c => Endpoint = c.DebuggerEndpoint);
```
GreenMatan: Are you sure that we plan to use `c.DebuggerEndpoint` in here and not `DiagnosticsEndpoint` or what you're doing in `SnapshotUploadApi`;
`Endpoint = c.DebuggerV2Endpoint ?? c.DiagnosticsEndpoint ?? c.DebuggerEndpoint;`?

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
dudikeleti: For now we forwarding only snapshot probe

tracer/src/Datadog.Trace/Debugger/Upload/LogUploadApi.cs
dudikeleti: For now we forwarding only snapshot probe

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
dudikeleti: Theoretically we can use it outside the class, like `AddSnapshot` so I'll keep it internal for now. We have to have a final decision about LogProbe, then we can update it accordingly.

tylfin:
FYI, for Python I manually ran the following test which helped QA the code:

1. Start application / tracer with the latest agent
2. Add instrumentation and trigger (snapshots should be uploaded)
3. Kill the agent, wait then trigger (check warning/error logs, not too verbose)
4. Start the agent v7.47.0 and trigger (no snapshots should be uploaded)
5. Kill the agent, and restart latest, and trigger (snapshots should against start being uploaded)

Could we test the same scenario here?

PR #7389, opened 2025-08-13
[Azure Functions] Remove all Serverless Mini-Agent (SMA) code

lucaspimentel:
## Summary of changes

Remove all code related to initializing the "Serverless Mini-Agent" (SMA), which was used during earlier previews of Azure Functions support via nuget packages (instead of the AAS Site Extension).

## Reason for change

The `Datadog.Sma` nuget package was used during the earlier preview period and has been deprecated for some time. It was replaced with the "Serverless Compatibility Layer" (`Datadog.Serverless.Compat`).

Since the tracer is no longer starting the SMA, users need to add a single line of code to start the agent. We will be looking into other ways to remove that friction, since the agent can be used without the tracer (for logs, metrics, etc).
```csharp
Datadog.Serverless.CompatibilityLayer.Start();
```

## Implementation details

Delete the code.

## Test coverage

Delete the tests.

## Other details

Resolves APMSVLS-112

### DIFF
--- a/tracer/src/Datadog.Trace.Trimming/build/Datadog.Trace.Trimming.xml
+++ b/tracer/src/Datadog.Trace.Trimming/build/Datadog.Trace.Trimming.xml
@@ -241,8 +241,6 @@
       <type fullname="System.Diagnostics.DistributedContextPropagator" />
    </assembly>
    <assembly fullname="System.Diagnostics.Process">
-      <type fullname="System.Diagnostics.DataReceivedEventArgs" />
-      <type fullname="System.Diagnostics.DataReceivedEventHandler" />
       <type fullname="System.Diagnostics.Process" />
       <type fullname="System.Diagnostics.ProcessStartInfo" />
       <type fullname="System.Diagnostics.ProcessThreadCollection" />


tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -1271,12 +1270,6 @@ public bool DiagnosticSourceEnabled
         /// </summary>
         internal bool IsRunningInAzureFunctions => AzureAppServiceMetadata?.IsFunctionsApp ?? false;
 
-        /// <summary>
-        /// Gets a value indicating whether the tracer is running in an Azure Function on a
-        /// consumption plan
-        /// </summary>
-        internal bool IsRunningMiniAgentInAzureFunctions => AzureAppServiceMetadata?.IsRunningMiniAgentInAzureFunctions ?? false;
```
andrewlock, line 1278: What were these settings actually _used_ for? 🤔

duncanpharvey (approved):

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
duncanpharvey: It was a way to prevent the Serverless Mini Agent from starting for a .NET Windows Azure Function on an App Service Plan since the .NET APM Extension should be used for that case

lucaspimentel:
### Merge activity

* **Aug 14, 7:57 PM UTC**: @lucaspimentel merged this pull request with [Graphite](https://app.graphite.dev/github/pr/DataDog/dd-trace-dotnet/7389).

PR #7390, opened 2025-08-14
Fix minor warnings and issues and improvements

andrewlock:
## Summary of changes

Fixes a few minor warnings and issues that came up in the .NET 10 migration

## Reason for change

They came up in the .NET 10 branch, but aren't directly related, so thought it would make sense to extract them out

## Implementation details

- Fix nullability issues
- Improve the condition checks in csproj for checking (e.g.) "is TargetFramework >= .NET 5"
- Fix some flake in `TraceContextTests` where you get the same value from `DateTimeOffset.UtcNow` sometimes
- Use the publicly available `ElapsedTicks` property

## Test coverage

Covered by existing


tracer/test/Datadog.Trace.Tests/Util/StopwatchTests.cs
```diff
@@ -22,11 +22,16 @@ public void ComputesCorrectTimespan()
 
             sw.Stop();
 
+#if NET6_0_OR_GREATER
+            // Since at least .NET 6, Stopwatch.GetRawElapsedTicks is directly exposed as ElapsedTicks
+            // and in .NET 10+ the GetRawElapsedTicks method doesn't exist
+            var stopwatchTicks = sw.ElapsedTicks;
```
lucaspimentel: Why can't we use `Stopwatch.ElapsedTicks` always? As far as I can tell it's been available since .NET _Framework_ 2.0.

lucaspimentel (approved):

tracer/test/Datadog.Trace.Tests/Util/StopwatchTests.cs
andrewlock: That's a good question, seeing as it's been that way https://github.com/DataDog/dd-trace-dotnet/pull/756 😅 But as far as I can tell, in all cases it calls GetRawElapsedTicks so yeah, I agree 🤔 Maybe we should ask @kevingosse 😅

andrewlock: Tried simplifying it, let's see what flakes 😄

kevingosse: Well, in #756 the goal was to avoid allocating a new stopwatch for every trace. Then I needed a test to validate that I compute the right timespan for a given amount of ticks, and for that I extracted the raw ticks to compare the result of my helper with Elapsed.

Now in hindsight, instead of doing that, we (and by that I mean you) can rewrite the test to use hardcoded input and expected values.

andrewlock: > Then I needed a test to validate that I compute the right timespan for a given amount of ticks, and for that I extracted the raw ticks to compare the result of my helper with Elapsed.

But... the question is why go to the effort of calling `Stopwatch.GetRawElapsedTicks()` when the implementation of `ElapsedTicks` is this (in all TFMs AFAICT)

```csharp
public class StopWatch
{
    public long ElapsedTicks => GetRawElapsedTicks();
}
```

As for using hardcoded input and expected values... yeah... that's a good point, we don't actually need a stopwatch at all for the test 😅

kevingosse: Oh 😅 yeah that's a good question, maybe past-me missed it

andrewlock: Thanks! 👋

PR #7394, opened 2025-08-15
Fix variable generation for `linux-arm64` unit tests

andrewlock:
## Summary of changes

Fix variable generation for `linux-arm64` unit tests

## Reason for change

We had a minor scare when I saw this in the build logs:
```
No test result files matching '[ 'artifacts/build_data/results/**/*.trx' ]' were found.
unit_tests_arm64 • test glibc_netcoreapp3.0 • publish test results
No test result files matching '[ 'artifacts/build_data/results/**/*.trx' ]' were found.
unit_tests_arm64 • test musl_netcoreapp3.0 • publish test results
No test result files matching '[ 'artifacts/build_data/results/**/*.trx' ]' were found.
unit_tests_arm64 • test glibc_netcoreapp3.1 • publish test results
No test result files matching '[ 'artifacts/build_data/results/**/*.trx' ]' were found.
unit_tests_arm64 • test musl_netcoreapp3.1 • publish test results
```

as it means we weren't running any tests in these scenarios. Luckily, we aren't _supposed_ to run tests in these scenarios, (<.NET 5 linux-arm64) so we can just stop creating these jobs entirely.

## Implementation details

The confusion is that the `TestingFrameworks` property uses ambient properties of the env to decide which frameworks to run (i.e. platform, arm64, test all TFms enabled), but we generate the variable list in a single job (on Windows). 

This PR makes it possible to ask what frameworks a platform/bitness combination _would_ run, and uses that for generating the job matrix.

## Test coverage

This is the test, but I'll do an "all TFMs" test to confirm too

### DIFF
--- a/tracer/build/_build/Build.Steps.Debugger.cs
+++ b/tracer/build/_build/Build.Steps.Debugger.cs
@@ -21,9 +21,6 @@ partial class Build
     [Parameter("Optimize generated code. Used for debugger integrations tests", List = false)]
     readonly bool? Optimize;
 
-    TargetFramework[] TestingFrameworksDebugger =>
-        TargetFramework.GetFrameworks(except: new[] { TargetFramework.NET461, TargetFramework.NETSTANDARD2_0, TargetFramework.NETCOREAPP3_0, TargetFramework.NET5_0 });
-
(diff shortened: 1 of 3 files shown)


zacharycmontoya (approved):
LGTM

PR #7402, opened 2025-08-18
chore: update malformed telemetry name

rachelyangdog:
## Summary of changes
- update malformed baggage telemetry name to `context_header_style.malformed` 
- this follows [dd-go naming ](https://github.com/DataDog/dd-go/blob/b11d80373c74d5a1900ff459a09274cca6223972/trace/apps/tracer-telemetry-intake/telemetry-metrics/static/common_metrics.json#L1713)

### DIFF
--- a/tracer/src/Datadog.Trace/Telemetry/Metrics/Count.cs
+++ b/tracer/src/Datadog.Trace/Telemetry/Metrics/Count.cs
@@ -109,7 +109,7 @@ internal enum Count
     /// <summary>
     /// The number of times a context propagation header is dropped because it is malformed, tagged by the header style (`header_style:baggage`)
     /// </summary>
-    [TelemetryMetric<MetricTags.ContextHeaderMalformed>("context_header.malformed")] ContextHeaderMalformed,
+    [TelemetryMetric<MetricTags.ContextHeaderMalformed>("context_header_style.malformed")] ContextHeaderMalformed,
 
     /// <summary>
     /// The number of requests sent to the stats endopint in the agent, regardless of success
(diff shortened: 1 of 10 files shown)


tracer/src/Datadog.Trace/Telemetry/Metrics/Count.cs
```diff
@@ -109,7 +109,7 @@ internal enum Count
     /// <summary>
     /// The number of times a context propagation header is dropped because it is malformed, tagged by the header style (`header_style:baggage`)
     /// </summary>
-    [TelemetryMetric<MetricTags.ContextHeaderMalformed>("context_header.malformed")] ContextHeaderMalformed,
+    [TelemetryMetric<MetricTags.ContextHeaderMalformed>("context_header_style.malformed")] ContextHeaderMalformed,
```
zacharycmontoya, line 112: ~Are the other metrics above OK or do they also need to be changed from `context_header` to `context_header_style`?~

EDIT: It looks like the only metric name that doesn't have this prefix is `context_header.truncated`. Is that one still correct or does it need to be updated?

rachelyangdog: The other metrics are OK... Dotnet set the precedent for the other metrics and malformed headers is the only new metric being added. The naming follows what exists in dd-go: https://github.com/DataDog/dd-go/blob/b11d80373c74d5a1900ff459a09274cca6223972/trace/apps/tracer-telemetry-intake/telemetry-metrics/static/common_metrics.json#L1682C1-L1721C28

PR #7405, opened 2025-08-19
refactor(libdatadog): namespaces and file organization, error code and typo

ganeshnj:
## Reason for change

Qualify of life improvement

## Implementation details

- created new DataPipeline and logging namespaces 
- fix typo
- updated error codes with new variants

## Test coverage

Nothing to cover since there is no business logic change

### DIFF
--- a/tracer/src/Datadog.Trace/Configuration/GlobalSettings.cs
+++ b/tracer/src/Datadog.Trace/Configuration/GlobalSettings.cs
@@ -123,7 +123,7 @@ namespace Datadog.Trace.Configuration
         {
             TelemetryFactory.Metrics.Record(PublicApiUsage.GlobalSettings_Reload);
             DatadogLogging.Reset();
-            var isLibdatadogAvailable = LibDatadogAvailaibilityHelper.IsLibDatadogAvailable;
+            var isLibdatadogAvailable = LibDatadogAvailabilityHelper.IsLibDatadogAvailable;
             if (isLibdatadogAvailable.IsAvailable)
             {
                 Logger.Instance.SetLogLevel(debugEnabled: false);
--- a/tracer/src/Datadog.Trace/Configuration/GlobalSettings.cs
+++ b/tracer/src/Datadog.Trace/Configuration/GlobalSettings.cs
@@ -9,6 +9,7 @@ using System;
 using Datadog.Trace.Configuration.ConfigurationSources.Telemetry;
 using Datadog.Trace.Configuration.Telemetry;
 using Datadog.Trace.LibDatadog;
+using Datadog.Trace.LibDatadog.Logging;
 using Datadog.Trace.Logging;
 using Datadog.Trace.SourceGenerators;
 using Datadog.Trace.Telemetry;
@@ -108,7 +109,7 @@ namespace Datadog.Trace.Configuration
                 DatadogLogging.UseDefaultLevel();
             }
 
-            LibDatadog.Logger.Instance.SetLogLevel(debugEnabled: enabled);
+            Logger.Instance.SetLogLevel(debugEnabled: enabled);
 
             TelemetryFactory.Config.Record(ConfigurationKeys.DebugEnabled, enabled, ConfigurationOrigins.Code);
         }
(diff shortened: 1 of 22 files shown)


tracer/src/Datadog.Trace/Configuration/GlobalSettings.cs
```diff
@@ -108,7 +109,7 @@ internal static void SetDebugEnabledInternal(bool enabled)
                 DatadogLogging.UseDefaultLevel();
             }
 
-            LibDatadog.Logger.Instance.SetLogLevel(debugEnabled: enabled);
+            Logger.Instance.SetLogLevel(debugEnabled: enabled);
```
andrewlock: I'm actually not a big fan of this because I think it's important to keep the context that this is LibDatadogLogger, and there's already multiple types and variables called `Logger` 😅 

I would suggest adding this `using`:

```csharp
using LibDatadogLog = Datadog.Trace.LibDatadog.Logger;
```

And then using this instead:

```suggestion
            LibDatadogLog .Instance.SetLogLevel(debugEnabled: enabled);
```

(or something similar)

tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporter.cs
```diff
@@ -7,16 +7,14 @@
...
-using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Threading.Tasks;
 using Datadog.Trace.Agent;
 using Datadog.Trace.Logging;
 using Datadog.Trace.Vendors.Newtonsoft.Json;
 using static Datadog.Trace.Agent.Api;
 
-namespace Datadog.Trace.LibDatadog;
+namespace Datadog.Trace.LibDatadog.DataPipeline;
```
andrewlock, line 17: Should this namespace really be TraceExporter (or similar) seeing as it's called TraceExporter everywhere in the code. I worry the overloading here can only cause confusion (you have to know they're the same exact thing 😄)

tracer/src/Datadog.Trace/TracerManagerFactory.cs
```diff
@@ -386,7 +388,7 @@ internal static IApi GetApi(TracerSettings settings, IDogStatsd statsd, Action<D
                     // If this was previously initialized, it will be re-initialized with the new settings, which is fine
                     if (Log.FileLoggingConfiguration is { } fileConfig)
                     {
-                        var logger = LibDatadog.Logger.Instance;
+                        var logger = Logger.Instance;
```
andrewlock: Same here, I'd _really_ like to disambiguate which logger we're talking about here

andrewlock (approved):
Thanks! I left a few suggestions 🙂

tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporter.cs
ganeshnj: Started with TraceExporter namespace but settled on data pipeline since this is how it maps to Rust side.

It would be easier to reason about later https://github.com/DataDog/libdatadog/tree/eeecfbafc6e799122f67c506771aa20b720dbee2/data-pipeline

tracer/src/Datadog.Trace/Configuration/GlobalSettings.cs
ganeshnj: fair enough!

I can be convinced with keeping the namespace prefix.

tracer/src/Datadog.Trace/TracerManagerFactory.cs
ganeshnj: fair enough!

tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporter.cs
andrewlock: In which case, I'd argue that we should rename all the TraceExporter types to be DataPipeline instead 😉 

Unless I'm conflating two things that are _actually_ different? Does the data-pipeline encapsulate anything _other_ than the trace exporter? Will it ever (from the managed side point of view)?

ganeshnj: it already does, CSS, dropping spans that are not required and some other things perhaps in future as we get more requirements.

currently we are handling telemetry as well with data pipeline which we shouldn't ideally.

lucaspimentel (approved):

PR #7410, opened 2025-08-19
Scrape native tracer stats from logs in smoke tests and track in Datadog

andrewlock:
## Summary of changes

Scrapes the stats that we write from the native tracer in to the logs and sends them to Datadog

## Reason for change

We aren't currently collecting these stats, and it would be good to understand the distribution of values we get, how they change over time, and how they vary per runtime.

## Implementation details

- Parse all the native logs looking for the stats line
- Parse the stats out
- Send metrics:
- `dd_trace_dotnet.ci.smoke_tests.init_duration_ms`
- `dd_trace_dotnet.ci.smoke_tests.callback_duration_ms`
- `dd_trace_dotnet.ci.smoke_tests.callback_count`
- `dd_trace_dotnet.ci.smoke_tests.callback_mean_duration_ms`

Where the latter three are tagged with
- `native:initialize`
- `native:module_load_finished`
- `native:calltarget_request_rejit`
- `native:calltarget_request_rewriter`
- `native:assembly_load_finished`
- `native:module_unload_started`
- `native:jit_compilation_started`
- `native:jit_inlining`
- `native:jit_cache_function_search_started`
- `native:initialize_profiler`
- `native:enquue_rejit_for_loaded_modules`

## Test coverage

Tested locally that the scraping works etc

## Other details

Requires https://github.com/DataDog/dd-trace-dotnet/pull/7409 for correct tagging:
- https://github.com/DataDog/dd-trace-dotnet/pull/7409


tracer/build/_build/LogParsing/LogParser.cs
```diff
@@ -0,0 +1,474 @@
...
+                    if (currentLine is not null)
+                    {
+                        allLogs.Add(currentLine);
+                        currentLine = null;
+                    }
+
+                    try
+                    {
+                        // native logs are on one line
+                        var timestamp = DateTimeOffset.ParseExact(match.Groups[1].Value, dateFormat, null);
```
NachoEchevarria: NIT: Should we use CultureInfo.InvariantCulture instead of null?

tracer/build/_build/LogParsing/LogParser.cs
```diff
@@ -0,0 +1,474 @@
...
+                if (pattern.IsMatch(logLine.Message))
+                {
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
+        bool ContainsCanary(ParsedLogLine logLine)
```
NachoEchevarria, line 177: NIT: Should we use StringComparison.OrdinalIgnoreCase to allow variations?

tracer/build/_build/LogParsing/LogParser.cs
andrewlock: Makes sense 👍

PR #7411, opened 2025-08-19
refactor(libdatadog): error handling and trace exporter response management

ganeshnj:
…gement


tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporter.cs
```diff
@@ -43,18 +43,35 @@ public Task<bool> SendTracesAsync(ArraySegment<byte> traces, int numberOfTraces,
...
+            else
             {
-                _log.Error(ex, "An error occurred deserializing the response.");
+                var json = response.ReadAsString();
+
+                if (string.IsNullOrEmpty(json))
+                {
+                    _log.Warning("Traces sent successfully to the Agent, but the response is empty");
+                }
+                else if (json != null && json != _cachedResponse)
```
ganeshnj: this null check is required to make the static analyzer happy, because `string.IsNullOrEmpty(json)` doesn't have the same impact as `!= null` check.

andrewlock: We have a helper that means you can avoid this, `StringUtil.IsNullOrEmpty(json)` instead of `string.IsNullOrEmpty()` will mean you can ditch the second null check 🙂

ganeshnj: nice

tracer/src/Datadog.Trace/LibDatadog/ErrorCode.cs
ganeshnj, line 1: https://github.com/DataDog/dd-trace-dotnet/blob/56b6954ed94d5115ea137e939ec4d10bf4682ec1/tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporterErrorCode.cs#L13

tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporterResponse.cs
```diff
@@ -0,0 +1,69 @@
...
+        if (!IsInvalid)
+        {
+            try
+            {
+                NativeInterop.Exporter.FreeResponse(handle);
+            }
+            catch (Exception ex)
+            {
+                Logger.Error(ex, "Failed to free TraceExporterResponse handle");
+                return false;
```
andrewlock: Do we ever actually want to return `false` from here? 🤔 Based on the docs: 

> `true` if the handle is released successfully; otherwise, in the event of a catastrophic failure, `false`. In this case, it generates a [releaseHandleFailed](https://learn.microsoft.com/en-us/dotnet/framework/debug-trace-profile/releasehandlefailed-mda) Managed Debugging Assistant.

Is this really a 'catastrophic' failure? Do we really want to risk triggering that additional Managed Debugging Assistant? IMO no, that behaviour prob risks causing issues for customers

tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporterResponse.cs
```diff
@@ -0,0 +1,69 @@
...
+            return null;
+        }
+
+        if (bodyLen > int.MaxValue)
+        {
+            Logger.Warning("Agent response is too large");
+            return null;
+        }
+
+        return System.Text.Encoding.UTF8.GetString((byte*)body, (int)bodyLen);
```
andrewlock: nit (didn't catch it before), but as this will always be a non-BOM utf-8 array, we should use the `StringEncoding` instance instead (I know, it's a weird namespace 😅 )
```suggestion
        return Datadog.Trace.Vendors.MessagePack.StringEncoding.UTF8.GetString((byte*)body, (int)bodyLen);
```

andrewlock (commented):
Thanks - just a question about whether we should ever return `false` from `ReleaseHandle()` (I suspect no)

tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporterResponse.cs
ganeshnj: I was reading the same doc yesterday and decided to add return `false` in case of a failure in drop/free

```
This MDA has no effect on the CLR.
```

As far as I can tell this shouldn't impact the working of the customer applications but while debugging gives us more useful information and potentially catch issues otherwise these will just slip through.

Although, I'm fine with just always returning `true` if this what we usually prefer.

andrewlock: > As far as I can tell this shouldn't impact the working of the customer applications

That's true as long as customers aren't _debugging_ their applications, otherwise it could be disruptive (and yeah, customers do that with the tracer attached sometimes). IMO it's not worth the risk tbh - we're already recording telemetry error logs, so these will already be reported and caught in CI 🤷‍♂️

ganeshnj: fair enough!

tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporter.cs
```diff
@@ -43,18 +43,35 @@ public Task<bool> SendTracesAsync(ArraySegment<byte> traces, int numberOfTraces,
...
+            if (response.IsInvalid)
             {
-                ProcessResponse(responsePtr);
+                _log.Warning("Traces sent successfully to the Agent, but the response is invalid");
             }
-            catch (Exception ex)
+            else
             {
-                _log.Error(ex, "An error occurred deserializing the response.");
+                var json = response.ReadAsString();
```
andrewlock, line 56: Totally random thought... this libdatadog API _always_ returns the response body, right? Was just wondering - if it _actually_ returned the same (cached) memory instead, then instead of having to deserialize into a string (and allocating) we could just directly compare the pointer. Terrible idea?

ganeshnj: It isn't a bad idea theoretically speaking but native implementation is quite straightforward here in this case, we always [box](https://github.com/DataDog/libdatadog/blob/c285c41e0465d5f19e12ce128b996a5acab9f708/data-pipeline-ffi/src/trace_exporter.rs#L533) (make a pointer) when returning the response, so it would not work as it is today.

although, this is more on the native side that we shouldn't return anything if stuff hasn't changed. So the logic stays in the native side and implemented only once.

NachoEchevarria (approved):

tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporter.cs
andrewlock: Yeah, agreed, that would be the best approach

PR #7413, opened 2025-08-19
[Tracer] Add Azure Service Bus instrumentation

pablomartinezbernardo:
## Summary of changes

New integration disabled by default, to be enabled with `DD_TRACE_AZURESERVICEBUS_ENABLED=true`

### Created [ServiceBusSenderSendMessagesAsyncIntegration](https://github.com/Azure/azure-sdk-for-net/blob/25a961e365e147fa446cc76b30ecfb8f4abfc90e/sdk/servicebus/Azure.Messaging.ServiceBus/src/Sender/ServiceBusSender.cs#L226)
* Creates a span

### Created [ServiceBusReceiverReceiveMessagesAsyncIntegration](https://github.com/Azure/azure-sdk-for-net/blob/25a961e365e147fa446cc76b30ecfb8f4abfc90e/sdk/servicebus/Azure.Messaging.ServiceBus/src/Receiver/ServiceBusReceiver.cs#L298)
* On begin: Does nothing. On serverless message reception can block for 60s, so we wait until something is actually read
* On end:
1. Creates a new span extracting context from Service Bus message and ~~parents/~~ links. ~~Parenting is used when all contexts are the same, span links otherwise.~~
2. Rewrites the contex for all messages so they point to the span we just created

### AzureFunctionsCommon changes
On service bus trigger, extracts context from service bus messages to set the parent. Given ServiceBusReceiverReceiveMessagesAsyncIntegration this should be one single context, warn otherwise.

### InstrumentMessageIntegration changes
Add context to all messages.

### Test coverage
Service Bus integration tests for sending, receiving and ~~parenting/~~ linking.

### Limitations
* Batching not included

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ServiceBusReceiverReceiveMessagesAsyncIntegration.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ServiceBusReceiverReceiveMessagesAsyncIntegration.cs
@@ -167,6 +167,12 @@ public class ServiceBusReceiverReceiveMessagesAsyncIntegration
             }
         }
 
+        var endpoint = receiverInstance.Connection?.ServiceEndpoint;
+        if (endpoint != null)
+        {
+            tags.ServerAddress = endpoint.Host;
+        }
+
         return scope;
     }
 
--- a/tracer/src/Datadog.Trace/Tagging/AzureServiceBusTags.cs
+++ b/tracer/src/Datadog.Trace/Tagging/AzureServiceBusTags.cs
@@ -38,6 +38,9 @@ namespace Datadog.Trace.Tagging
         [Tag(Trace.Tags.NetworkDestinationPort)]
         public string NetworkDestinationPort { get; set; }
 
+        [Tag(Trace.Tags.ServerAddress)]
+        public string ServerAddress { get; set; }
+
         public override string SpanKind
         {
             get => MessagingOperation switch
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/IServiceBusReceiver.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/IServiceBusReceiver.cs
@@ -12,4 +12,7 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Azure.ServiceBus;
 internal interface IServiceBusReceiver : IDuckType
 {
     string EntityPath { get; }
+
+    [DuckField(Name = "_connection")]
+    IServiceBusConnection Connection { get; }
 }
--- a/tracer/src/Datadog.Trace/Generated/net461/Datadog.Trace.SourceGenerators/TagListGenerator/AzureServiceBusTags.g.cs
+++ b/tracer/src/Datadog.Trace/Generated/net461/Datadog.Trace.SourceGenerators/TagListGenerator/AzureServiceBusTags.g.cs
@@ -30,6 +30,8 @@ namespace Datadog.Trace.Tagging
         private static ReadOnlySpan<byte> NetworkDestinationNameBytes => new byte[] { 184, 110, 101, 116, 119, 111, 114, 107, 46, 100, 101, 115, 116, 105, 110, 97, 116, 105, 111, 110, 46, 110, 97, 109, 101 };
         // NetworkDestinationPortBytes = MessagePack.Serialize("network.destination.port");
         private static ReadOnlySpan<byte> NetworkDestinationPortBytes => new byte[] { 184, 110, 101, 116, 119, 111, 114, 107, 46, 100, 101, 115, 116, 105, 110, 97, 116, 105, 111, 110, 46, 112, 111, 114, 116 };
+        // ServerAddressBytes = MessagePack.Serialize("server.address");
+        private static ReadOnlySpan<byte> ServerAddressBytes => new byte[] { 174, 115, 101, 114, 118, 101, 114, 46, 97, 100, 100, 114, 101, 115, 115 };
 
         public override string? GetTag(string key)
         {
@@ -41,6 +43,7 @@ namespace Datadog.Trace.Tagging
                 "message_bus.destination" => LegacyMessageBusDestination,
                 "network.destination.name" => NetworkDestinationName,
                 "network.destination.port" => NetworkDestinationPort,
+                "server.address" => ServerAddress,
                 _ => base.GetTag(key),
             };
         }
@@ -66,6 +69,9 @@ namespace Datadog.Trace.Tagging
                     break;
                 case "network.destination.port": 
                     NetworkDestinationPort = value;
+                    break;
+                case "server.address": 
+                    ServerAddress = value;
                     break;
                 default: 
                     base.SetTag(key, value);
@@ -103,6 +109,11 @@ namespace Datadog.Trace.Tagging
             if (NetworkDestinationPort is not null)
             {
                 processor.Process(new TagItem<string>("network.destination.port", NetworkDestinationPort, NetworkDestinationPortBytes));
+            }
+
+            if (ServerAddress is not null)
+            {
+                processor.Process(new TagItem<string>("server.address", ServerAddress, ServerAddressBytes));
             }
 
             base.EnumerateTags(ref processor);
@@ -149,6 +160,13 @@ namespace Datadog.Trace.Tagging
             {
                 sb.Append("network.destination.port (tag):")
                   .Append(NetworkDestinationPort)
+                  .Append(',');
+            }
+
+            if (ServerAddress is not null)
+            {
+                sb.Append("server.address (tag):")
+                  .Append(ServerAddress)
                   .Append(',');
             }
 
--- a/tracer/src/Datadog.Trace/Generated/net6.0/Datadog.Trace.SourceGenerators/TagListGenerator/AzureServiceBusTags.g.cs
+++ b/tracer/src/Datadog.Trace/Generated/net6.0/Datadog.Trace.SourceGenerators/TagListGenerator/AzureServiceBusTags.g.cs
@@ -30,6 +30,8 @@ namespace Datadog.Trace.Tagging
         private static ReadOnlySpan<byte> NetworkDestinationNameBytes => new byte[] { 184, 110, 101, 116, 119, 111, 114, 107, 46, 100, 101, 115, 116, 105, 110, 97, 116, 105, 111, 110, 46, 110, 97, 109, 101 };
         // NetworkDestinationPortBytes = MessagePack.Serialize("network.destination.port");
         private static ReadOnlySpan<byte> NetworkDestinationPortBytes => new byte[] { 184, 110, 101, 116, 119, 111, 114, 107, 46, 100, 101, 115, 116, 105, 110, 97, 116, 105, 111, 110, 46, 112, 111, 114, 116 };
+        // ServerAddressBytes = MessagePack.Serialize("server.address");
+        private static ReadOnlySpan<byte> ServerAddressBytes => new byte[] { 174, 115, 101, 114, 118, 101, 114, 46, 97, 100, 100, 114, 101, 115, 115 };
 
         public override string? GetTag(string key)
         {
@@ -41,6 +43,7 @@ namespace Datadog.Trace.Tagging
                 "message_bus.destination" => LegacyMessageBusDestination,
                 "network.destination.name" => NetworkDestinationName,
                 "network.destination.port" => NetworkDestinationPort,
+                "server.address" => ServerAddress,
                 _ => base.GetTag(key),
             };
         }
@@ -66,6 +69,9 @@ namespace Datadog.Trace.Tagging
                     break;
                 case "network.destination.port": 
                     NetworkDestinationPort = value;
+                    break;
+                case "server.address": 
+                    ServerAddress = value;
                     break;
                 default: 
                     base.SetTag(key, value);
@@ -103,6 +109,11 @@ namespace Datadog.Trace.Tagging
             if (NetworkDestinationPort is not null)
             {
                 processor.Process(new TagItem<string>("network.destination.port", NetworkDestinationPort, NetworkDestinationPortBytes));
+            }
+
+            if (ServerAddress is not null)
+            {
+                processor.Process(new TagItem<string>("server.address", ServerAddress, ServerAddressBytes));
             }
 
             base.EnumerateTags(ref processor);
@@ -149,6 +160,13 @@ namespace Datadog.Trace.Tagging
             {
                 sb.Append("network.destination.port (tag):")
                   .Append(NetworkDestinationPort)
+                  .Append(',');
+            }
+
+            if (ServerAddress is not null)
+            {
+                sb.Append("server.address (tag):")
+                  .Append(ServerAddress)
                   .Append(',');
             }
 
--- a/tracer/src/Datadog.Trace/Generated/netcoreapp3.1/Datadog.Trace.SourceGenerators/TagListGenerator/AzureServiceBusTags.g.cs
+++ b/tracer/src/Datadog.Trace/Generated/netcoreapp3.1/Datadog.Trace.SourceGenerators/TagListGenerator/AzureServiceBusTags.g.cs
@@ -30,6 +30,8 @@ namespace Datadog.Trace.Tagging
         private static ReadOnlySpan<byte> NetworkDestinationNameBytes => new byte[] { 184, 110, 101, 116, 119, 111, 114, 107, 46, 100, 101, 115, 116, 105, 110, 97, 116, 105, 111, 110, 46, 110, 97, 109, 101 };
         // NetworkDestinationPortBytes = MessagePack.Serialize("network.destination.port");
         private static ReadOnlySpan<byte> NetworkDestinationPortBytes => new byte[] { 184, 110, 101, 116, 119, 111, 114, 107, 46, 100, 101, 115, 116, 105, 110, 97, 116, 105, 111, 110, 46, 112, 111, 114, 116 };
+        // ServerAddressBytes = MessagePack.Serialize("server.address");
+        private static ReadOnlySpan<byte> ServerAddressBytes => new byte[] { 174, 115, 101, 114, 118, 101, 114, 46, 97, 100, 100, 114, 101, 115, 115 };
 
         public override string? GetTag(string key)
         {
@@ -41,6 +43,7 @@ namespace Datadog.Trace.Tagging
                 "message_bus.destination" => LegacyMessageBusDestination,
                 "network.destination.name" => NetworkDestinationName,
                 "network.destination.port" => NetworkDestinationPort,
+                "server.address" => ServerAddress,
                 _ => base.GetTag(key),
             };
         }
@@ -66,6 +69,9 @@ namespace Datadog.Trace.Tagging
                     break;
                 case "network.destination.port": 
                     NetworkDestinationPort = value;
+                    break;
+                case "server.address": 
+                    ServerAddress = value;
                     break;
                 default: 
                     base.SetTag(key, value);
@@ -103,6 +109,11 @@ namespace Datadog.Trace.Tagging
             if (NetworkDestinationPort is not null)
             {
                 processor.Process(new TagItem<string>("network.destination.port", NetworkDestinationPort, NetworkDestinationPortBytes));
+            }
+
+            if (ServerAddress is not null)
+            {
+                processor.Process(new TagItem<string>("server.address", ServerAddress, ServerAddressBytes));
             }
 
             base.EnumerateTags(ref processor);
@@ -149,6 +160,13 @@ namespace Datadog.Trace.Tagging
             {
                 sb.Append("network.destination.port (tag):")
                   .Append(NetworkDestinationPort)
+                  .Append(',');
+            }
+
+            if (ServerAddress is not null)
+            {
+                sb.Append("server.address (tag):")
+                  .Append(ServerAddress)
                   .Append(',');
             }
 
--- a/tracer/src/Datadog.Trace/Generated/netstandard2.0/Datadog.Trace.SourceGenerators/TagListGenerator/AzureServiceBusTags.g.cs
+++ b/tracer/src/Datadog.Trace/Generated/netstandard2.0/Datadog.Trace.SourceGenerators/TagListGenerator/AzureServiceBusTags.g.cs
@@ -30,6 +30,8 @@ namespace Datadog.Trace.Tagging
         private static ReadOnlySpan<byte> NetworkDestinationNameBytes => new byte[] { 184, 110, 101, 116, 119, 111, 114, 107, 46, 100, 101, 115, 116, 105, 110, 97, 116, 105, 111, 110, 46, 110, 97, 109, 101 };
         // NetworkDestinationPortBytes = MessagePack.Serialize("network.destination.port");
         private static ReadOnlySpan<byte> NetworkDestinationPortBytes => new byte[] { 184, 110, 101, 116, 119, 111, 114, 107, 46, 100, 101, 115, 116, 105, 110, 97, 116, 105, 111, 110, 46, 112, 111, 114, 116 };
+        // ServerAddressBytes = MessagePack.Serialize("server.address");
+        private static ReadOnlySpan<byte> ServerAddressBytes => new byte[] { 174, 115, 101, 114, 118, 101, 114, 46, 97, 100, 100, 114, 101, 115, 115 };
 
         public override string? GetTag(string key)
         {
@@ -41,6 +43,7 @@ namespace Datadog.Trace.Tagging
                 "message_bus.destination" => LegacyMessageBusDestination,
                 "network.destination.name" => NetworkDestinationName,
                 "network.destination.port" => NetworkDestinationPort,
+                "server.address" => ServerAddress,
                 _ => base.GetTag(key),
             };
         }
@@ -66,6 +69,9 @@ namespace Datadog.Trace.Tagging
                     break;
                 case "network.destination.port": 
                     NetworkDestinationPort = value;
+                    break;
+                case "server.address": 
+                    ServerAddress = value;
                     break;
                 default: 
                     base.SetTag(key, value);
@@ -103,6 +109,11 @@ namespace Datadog.Trace.Tagging
             if (NetworkDestinationPort is not null)
             {
                 processor.Process(new TagItem<string>("network.destination.port", NetworkDestinationPort, NetworkDestinationPortBytes));
+            }
+
+            if (ServerAddress is not null)
+            {
+                processor.Process(new TagItem<string>("server.address", ServerAddress, ServerAddressBytes));
             }
 
             base.EnumerateTags(ref processor);
@@ -149,6 +160,13 @@ namespace Datadog.Trace.Tagging
             {
                 sb.Append("network.destination.port (tag):")
                   .Append(NetworkDestinationPort)
+                  .Append(',');
+            }
+
+            if (ServerAddress is not null)
+            {
+                sb.Append("server.address (tag):")
+                  .Append(ServerAddress)
                   .Append(',');
             }
 
--- a/tracer/src/Datadog.Trace/Tags.cs
+++ b/tracer/src/Datadog.Trace/Tags.cs
@@ -146,6 +146,11 @@ namespace Datadog.Trace
         public const string OutPort = "out.port";
 
         /// <summary>
+        /// The server address for the remote service.
+        /// </summary>
+        public const string ServerAddress = "server.address";
+
+        /// <summary>
         /// The size of the message.
         /// </summary>
         public const string MessageSize = "message.size";
--- a/tracer/test/Datadog.Trace.TestHelpers/SpanMetadataV0Rules.cs
+++ b/tracer/test/Datadog.Trace.TestHelpers/SpanMetadataV0Rules.cs
@@ -316,7 +316,7 @@ namespace Datadog.Trace.TestHelpers
                 .IsOptional("messaging.message_id")
                 .IsOptional("net.peer.name")
                 .IsOptional("peer.address")
-                .IsOptional("server.address")
+                .IsPresent("server.address")
                 .Matches("component", "AzureServiceBus")
                 .IfPresentMatches("kind", "consumer")
                 .IsOptional("_dd.base_service")
--- a/tracer/test/Datadog.Trace.TestHelpers/SpanMetadataV1Rules.cs
+++ b/tracer/test/Datadog.Trace.TestHelpers/SpanMetadataV1Rules.cs
@@ -517,7 +517,7 @@ namespace Datadog.Trace.TestHelpers
                 .IsOptional("messaging.message_id")
                 .IsOptional("net.peer.name")
                 .IsOptional("peer.address")
-                .IsOptional("server.address")
+                .IsPresent("server.address")
                 .Matches("component", "AzureServiceBus")
                 .IfPresentMatches("kind", "consumer")
                 .IsOptional("_dd.base_service")


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -330,6 +338,117 @@ private static PropagationContext ExtractPropagatedContextFromHttp<T>(T context,
...
+                {
+                    return default;
+                }
+
+                bool areAllTheSame = spanContexts.Count == 1 ||
+                                     (spanContexts.Count > 1 && AreAllContextsIdentical(spanContexts));
+
+                if (!areAllTheSame)
+                {
+                    Log.Warning("Multiple different contexts found in ServiceBus messages. Using first context for parentship.");
```
lucaspimentel, line 404: If this is a valid use-case, consider lowering the log severity. Otherwise some users will get these warnings all the time and they are not actionable.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -330,6 +338,117 @@ private static PropagationContext ExtractPropagatedContextFromHttp<T>(T context,
...
+                {
+                    return default;
+                }
+
+                GrpcBindingsFeatureStruct? bindingsFeature = null;
+                foreach (var kvp in context.Features)
+                {
+                    if (kvp.Key.FullName?.Equals("Microsoft.Azure.Functions.Worker.Context.Features.IFunctionBindingsFeature") == true)
+                    {
+                        bindingsFeature = kvp.Value?.TryDuckCast<GrpcBindingsFeatureStruct>(out var feature) == true ? feature : null;
```
lucaspimentel, line 358: Can you do this instead? Assuming `TryDuckCast()` sets the `out` parameter to `null` when it fails (it should).

```suggestion
                        kvp.Value?.TryDuckCast<GrpcBindingsFeatureStruct>(out var bindingsFeature);
```

(level of care: low)

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ContextPropagation.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Azure.ServiceBus;
+
+internal readonly struct ContextPropagation : ICarrierGetter<IDictionary<string, object>>, ICarrierSetter<IDictionary<string, object>>
+{
+    public IEnumerable<string> Get(IDictionary<string, object> carrier, string key)
+    {
+        if (carrier.TryGetValue(key, out var value) && value is string stringValue)
+        {
+            return new[] { stringValue };
```
lucaspimentel:
```suggestion
            return [stringValue];
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ContextPropagation.cs
```diff
@@ -0,0 +1,32 @@
...
+        }
+
+        return Enumerable.Empty<string>();
```
lucaspimentel:
```suggestion
        return [];
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ContextPropagation.cs
```diff
@@ -0,0 +1,32 @@
...
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2 License.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
+// </copyright>
+
+#nullable enable
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
```
lucaspimentel:
```suggestion
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/InstrumentMessageIntegration.cs
```diff
@@ -10,6 +10,7 @@
 using System.ComponentModel;
 using Datadog.Trace.ClrProfiler.CallTarget;
 using Datadog.Trace.Configuration;
+using Datadog.Trace.Logging;
```
lucaspimentel: Looks like you didn't use this.
```suggestion
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/InstrumentMessageIntegration.cs
```diff
@@ -41,15 +42,25 @@ public class InstrumentMessageIntegration
...
+                    // save the active message properties object to an AsyncLocal field. This will limit
+                    // our lookup to one AsyncLocal field and one static field
+                    AzureServiceBusCommon.ActiveMessageProperties.Value = properties;
+                }
+
+                // Inject tracing context into the final message properties
+                var activeScope = Tracer.Instance.ActiveScope;
+                if (activeScope?.Span?.Context != null && properties != null)
+                {
+                    var context = new Datadog.Trace.Propagators.PropagationContext(activeScope.Span.Context as Datadog.Trace.SpanContext, Datadog.Trace.Baggage.Current);
```
lucaspimentel: [coding style]

```suggestion
                if (activeScope?.Span?.Context is SpanContext spanContext && properties != null)
                {
                    var context = new PropagationContext(spanContext, Baggage.Current);
```

You'll need `using Datadog.Trace.Propagators;` up top.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/IServiceBusReceivedMessage.cs
```diff
@@ -7,10 +7,15 @@
 
 using System;
 using System.Collections.Generic;
+using Datadog.Trace.DuckTyping;
```
lucaspimentel:
```suggestion
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
pablomartinezbernardo: > Assuming TryDuckCast() sets the out parameter to null when it fails (it should).

It doesn't, it even [requires non-null](https://github.com/DataDog/dd-trace-dotnet/blob/6baf47efb56362d5e8e4fd1cbb712e977d80dc1d/tracer/src/Datadog.Trace/DuckTyping/DuckTypeExtensions.cs#L49) and the object is a nullable struct

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
pablomartinezbernardo: This is not a valid use-case, when messages are received by a function they should have already gone through the ReceiveMessageIntegration and have their messages reparented, so this situation means that for some reason the ReceiveMessagesIntegration was never ran, which means things will not work as expected.

Still, I'm not completely used to the usual severity of the logs, so if you still think it should be lower please say!

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ContextPropagation.cs
pablomartinezbernardo: Removed only System.Text because of the other comments in this file

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ContextPropagation.cs
pablomartinezbernardo: That would be incompatible with older net fwks right?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ContextPropagation.cs
pablomartinezbernardo: Same as the other comment, this would not be compatible with older net fwks, might be missing something though 🤔

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/InstrumentMessageIntegration.cs
pablomartinezbernardo: 👍

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/InstrumentMessageIntegration.cs
pablomartinezbernardo: 👍

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/IServiceBusReceivedMessage.cs
pablomartinezbernardo: 👍

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ContextPropagation.cs
bouwkast: It's actually a C# language feature added in C# 12 so not tied to a framework-> https://andrewlock.net/behind-the-scenes-of-collection-expressions-part-1-introducing-collection-expressions-in-csharp12/

tracer/build/PackageVersionsGeneratorDefinitions.json
```diff
@@ -864,6 +864,17 @@
...
+    "IntegrationName": "AzureServiceBusAPM",
+    "SampleProjectName": "Samples.AzureServiceBus.APM",
+    "NugetPackageSearchName": "Azure.Messaging.ServiceBus",
+    "MinVersion": "7.18.0",
+    "MaxVersionExclusive": "7.21.0",
+    "SpecificVersions": [
+      "7.18.*",
+      "7.*.*"
+    ]
+  },
```
bouwkast, line 877: Just to clarify, we made a new one here because the sample application is different?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/Isolated/GrpcBindingsFeatureStruct.cs
```diff
@@ -0,0 +1,24 @@
+// <copyright file="GrpcBindingsFeatureStruct.cs" company="Datadog">
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2 License.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
+// </copyright>
+
+using System.Collections.Generic;
+using Datadog.Trace.DuckTyping;
+
+#if !NETFRAMEWORK
```
bouwkast:
```suggestion
#if !NETFRAMEWORK

using System.Collections.Generic;
using Datadog.Trace.DuckTyping;

```

nitpick but the `using`s can be moved within the preprocessor

docker-compose.yml
```diff
@@ -209,6 +209,19 @@ services:
     - ACCEPT_EULA=Y
     - SA_PASSWORD=Strong!Passw0rd
 
+  azureservicebus-emulator:
```
bouwkast, line 218: Awesome thanks!

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -330,6 +338,117 @@ private static PropagationContext ExtractPropagatedContextFromHttp<T>(T context,
...
+            }
+        }
+
+        private static SpanContext? ExtractSpanContextFromProperties(Dictionary<string, object> userProperties)
+        {
+            var extractedContext = Tracer.Instance.TracerManager.SpanContextPropagator.Extract(userProperties, default(ServiceBus.ContextPropagation));
+            return extractedContext.SpanContext;
+        }
+
+        private static T? TryParseJson<T>(object? jsonObj)
```
bouwkast: nitpick but I think changing this to return a `bool` and have an `out T?` would be better (you'd need to update usages potentially) just to be more consistent with `Try...` functions


```suggestion
        private static bool TryParseJson<T>(object? jsonObj, out T? result)
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ServiceBusReceiverReceiveMessagesAsyncIntegration.cs
```diff
@@ -0,0 +1,240 @@
...
+
+            if (extractedContexts.Count == 0)
+            {
+                return new ContextExtractionResult(null, null);
+            }
+
+            // Check if all contexts are the same
+            var firstContext = extractedContexts[0];
+            var comparer = new SpanContextComparer();
+            var allSame = extractedContexts.All(ctx => comparer.Equals(ctx, firstContext));
```
bouwkast: Any chance we could modify this function so we aren't doing two iterations over the contexts?
Maybe we could do a `HashSet` 

Maybe it isn't super important though I don't know how many contexts  we'd actually expect here so it may be very marginal.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ServiceBusReceiverReceiveMessagesAsyncIntegration.cs
```diff
@@ -0,0 +1,240 @@
...
+        if (!tracer.Settings.IsIntegrationEnabled(IntegrationId.AzureServiceBus, false))
+        {
+            return returnValue;
+        }
+
+        var messagesList = returnValue as System.Collections.IList;
+        var messageCount = messagesList?.Count ?? 0;
+
+        // Don't create spans when there's an exception or no messages
+        if (exception != null || messageCount == 0)
```
bouwkast, line 61: Why don't we create a span when there is an exception?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ServiceBusReceiverReceiveMessagesAsyncIntegration.cs
```diff
@@ -0,0 +1,240 @@
...
+/// <summary>
+/// System.Threading.Tasks.Task`1[System.Collections.Generic.IReadOnlyList`1[Azure.Messaging.ServiceBus.ServiceBusReceivedMessage]] Azure.Messaging.ServiceBus.ServiceBusReceiver::ReceiveMessagesAsync(System.Int32,System.Nullable`1[System.TimeSpan],System.Boolean,System.Threading.CancellationToken) calltarget instrumentation
+/// </summary>
+[InstrumentMethod(
+    AssemblyName = "Azure.Messaging.ServiceBus",
+    TypeName = "Azure.Messaging.ServiceBus.ServiceBusReceiver",
+    MethodName = "ReceiveMessagesAsync",
+    ReturnTypeName = "System.Threading.Tasks.Task`1[System.Collections.Generic.IReadOnlyList`1[Azure.Messaging.ServiceBus.ServiceBusReceivedMessage]]",
+    ParameterTypeNames = [ClrNames.Int32, "System.Nullable`1[System.TimeSpan]", ClrNames.Bool, ClrNames.CancellationToken],
+    MinimumVersion = "7.14.0",
```
bouwkast, line 31: Just want to make sure that the `7.14.0` is right (haven't checked anything just wondering)

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ServiceBusSenderSendMessagesAsyncIntegration.cs
```diff
@@ -0,0 +1,103 @@
...
+            {
+                scope.Span.SetException(exception);
+            }
+        }
+        finally
+        {
+            scope.Dispose();
+        }
+
+        return returnValue;
```
bouwkast: We have a helper for this that looks like

```suggestion
        state.Scope.DisposeWithException(exception);
        return returnValue;
```

tracer/src/Datadog.Trace/Tagging/AzureServiceBusTags.cs
```diff
@@ -17,6 +17,9 @@ internal partial class AzureServiceBusTags : OpenTelemetryTags
     {
         private string _spanKind;
 
+        [Tag(Trace.Tags.InstrumentationName)]
+        public string InstrumentationName { get; set; }
```
bouwkast, line 21: Just to ensure it is always set (you'll need to update instances where it was being set explicitly though)

```suggestion
        public string InstrumentationName => nameof(IntegrationId.AzureServiceBus);
```

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/Azure/AzureServiceBusAPMTests.cs
bouwkast, line 1: Do we have tests for the span links?

I'm actually not entirely sure what testing Span Links looks like though at the moment, but I'd think we would have an example

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ContextPropagation.cs
pablomartinezbernardo: Thanks!

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ContextPropagation.cs
pablomartinezbernardo: Done!

tracer/build/PackageVersionsGeneratorDefinitions.json
pablomartinezbernardo: The other samples app was built by DSM and was kept for manual testing, never included as part of CI. We want to test the new stuff, but these changes should be unrelated to DSM instrumentations, so I created a new sample app and included this flow in CI

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/Isolated/GrpcBindingsFeatureStruct.cs
pablomartinezbernardo: Moved!

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
pablomartinezbernardo: 👍

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ServiceBusReceiverReceiveMessagesAsyncIntegration.cs
pablomartinezbernardo: Not super important, but using a HashSet makes it easier to read as well so 👍

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ServiceBusReceiverReceiveMessagesAsyncIntegration.cs
pablomartinezbernardo: 7.14 forwards I confirmed everything is there, unsure about earlier versions. The reason to pick 7.14 was simply because that is what the DSM integration supports, maintaining the same version support.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ServiceBusReceiverReceiveMessagesAsyncIntegration.cs
pablomartinezbernardo: Open to alternatives. The main reasoning is that when using a ServiceBusTrigger, ReceiveMessagesAsync is periodically called from the host until something is received. Because of that, whenever shut down/restart operations are done on the funcion the message reception ends up being cancelled with an exception, which doesn't feel completely right when everything is working as expected.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ServiceBusSenderSendMessagesAsyncIntegration.cs
pablomartinezbernardo: Didn't know, done!

tracer/src/Datadog.Trace/Tagging/AzureServiceBusTags.cs
pablomartinezbernardo: This is actually on purpose: AzureServiceBusTags was already being used by the other service bus integrations, so I didn't want to be modifying those. But again, open to alternatives

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/Azure/AzureServiceBusAPMTests.cs
pablomartinezbernardo: No there was nothing, only parenting was being tested. I added them!

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ContextPropagation.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Collections.Generic;
+using Datadog.Trace.Propagators;
+
+namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Azure.ServiceBus;
+
+internal readonly struct ContextPropagation : ICarrierGetter<IDictionary<string, object>>, ICarrierSetter<IDictionary<string, object>>
+{
+    public IEnumerable<string> Get(IDictionary<string, object> carrier, string key)
+    {
+        if (carrier.TryGetValue(key, out var value) && value is string stringValue)
```
bricefriha, line 18: Is there any risk of ending up with a `null reference exception`? 🤔

Is there any scenario in which carrier can be null?

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/Azure/AzureServiceBusAPMTests.cs
```diff
@@ -0,0 +1,184 @@
...
+        {
+            SetEnvironmentVariable("DD_TRACE_SPAN_ATTRIBUTE_SCHEMA", metadataSchemaVersion);
+            SetEnvironmentVariable("DD_TRACE_AZURESERVICEBUS_ENABLED", "true");
+
+            using (var agent = EnvironmentHelper.GetMockAgent())
+            using (await RunSampleAndWaitForExit(agent, packageVersion: packageVersion))
+            {
+                var spans = await agent.WaitForSpansAsync(6, timeoutInMilliseconds: 30000);
+
+                using var s = new AssertionScope();
```
bricefriha, line 130: Are we reusing this variable in the scope?
(sorry if I missed it but I can't see it being used again)

```suggestion
```

lucaspimentel:
@codex review

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
lucaspimentel: Huh, I guess `TryDuckCas()` doesn't follow the .NET convention for `TryABC()` methods 😓 
Thanks.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ContextPropagation.cs
lucaspimentel: Right. It's only syntax sugar. The compiler converts it to a correct type like `string[]`, `List<string>`, `IEnumerable<string>`, etc as needed.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -330,6 +338,119 @@ private static PropagationContext ExtractPropagatedContextFromHttp<T>(T context,
...
+            {
+                Log.Error(ex, "Error extracting propagated context from ServiceBus binding");
+                return default;
+            }
+        }
+
+        private static SpanContext? ExtractSpanContextFromProperties(Dictionary<string, object> userProperties)
+        {
+            var extractedContext = Tracer.Instance.TracerManager.SpanContextPropagator.Extract(userProperties, default(ServiceBus.ContextPropagation));
+            return extractedContext.SpanContext;
```
lucaspimentel, line 419: This effectively throws away any baggage propagated in this message. This might be ok, since I think users expect baggage to propagate over HTTP or gRPC only, but it might be something we have to come back and add in later.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -330,6 +338,119 @@ private static PropagationContext ExtractPropagatedContextFromHttp<T>(T context,
...
+        }
+
+        private static bool TryParseJson<T>(object? jsonObj, out T? result)
```
lucaspimentel: This helps the compiler's nullability analyzer known that `result` is not null when the method returns `true`.
```suggestion
        private static bool TryParseJson<T>(object? jsonObj, [NotNullWhen(true)] out T? result)
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ServiceBusReceiverReceiveMessagesAsyncIntegration.cs
```diff
@@ -0,0 +1,209 @@
...
+                }
+            }
+
+            if (extractedContexts.Count == 0)
+            {
+                return null;
+            }
+
+            var uniqueContexts = new HashSet<SpanContext>(extractedContexts, new SpanContextComparer());
+            return uniqueContexts.Select(ctx => new SpanLink(ctx)).ToList();
```
lucaspimentel: We try to avoid LINQ (`.Select()` and .`ToList()`) to reduce heap allocations and delegate calls. In this case, you know the size of `uniqueContexts`, so you can do something like:
```csharp
// initialize with known length to avoid growing dynamically
var spanLinks = new List<SpanLink>(uniqueContext.Length);

// foreach over HashSet uses a struct enumerator instead of allocating one in the heap
foreach (var ctx in uniqueContexts)
{
    spanLinks.Add(new SpanLink(ctx));
}

return spanLinks;
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ServiceBusReceiverReceiveMessagesAsyncIntegration.cs
```diff
@@ -0,0 +1,209 @@
...
+    {
+        var tags = tracer.CurrentTraceSettings.Schema.Messaging.CreateAzureServiceBusTags(SpanKinds.Consumer);
+
+        var entityPath = receiverInstance.EntityPath ?? "unknown";
+        tags.MessagingDestinationName = entityPath;
+        tags.MessagingOperation = "receive";
+        tags.MessagingSystem = "servicebus";
+        tags.InstrumentationName = "AzureServiceBus";
+
+        string serviceName = tracer.CurrentTraceSettings.Schema.Messaging.GetServiceName("azureservicebus");
```
lucaspimentel, line 140: What will the service name be here?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ServiceBusReceiverReceiveMessagesAsyncIntegration.cs
lucaspimentel: I think we can go even further. Instead of creating a `List<SpanContext>()` first and then populating a `HashSet<SpanContext>` with the contents of the list, we can start with the `HashSet` and avoid the `List` entirely.

```diff
- var extractedContexts = new List<SpanContext>();
+ var extractedContexts = new HashSet<SpanContext>(new SpanContextComparer());
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ServiceBusSenderSendMessagesAsyncIntegration.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        var tags = tracer.CurrentTraceSettings.Schema.Messaging.CreateAzureServiceBusTags(SpanKinds.Producer);
+
+        var entityPath = instance.EntityPath ?? "unknown";
+        tags.MessagingDestinationName = entityPath;
+        tags.MessagingOperation = "send";
+        tags.MessagingSystem = "servicebus";
+        tags.InstrumentationName = "AzureServiceBus";
+
+        string serviceName = tracer.CurrentTraceSettings.Schema.Messaging.GetServiceName("azureservicebus");
```
lucaspimentel: What will the service name be here?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ServiceBusSenderSendMessagesAsyncIntegration.cs
```diff
@@ -0,0 +1,85 @@
...
+        span.Type = SpanTypes.Queue;
+        span.ResourceName = entityPath;
+
+        var endpoint = instance.Connection?.ServiceEndpoint;
+        if (endpoint != null)
+        {
+            tags.NetworkDestinationName = endpoint.Host;
+            // https://learn.microsoft.com/en-us/dotnet/api/system.uri.port?view=net-8.0#remarks
+            var port = endpoint.Port == -1 ? DefaultServiceBusPort : endpoint.Port;
+            tags.NetworkDestinationPort = port.ToString();
```
lucaspimentel: We should save `DefaultServiceBusPort` as a string to avoid the allocation in `int.ToString()` below.

```csharp
private const string DefaultServiceBusPort = "5671";

// ...

tags.NetworkDestinationPort = endpoint.Port is -1 or 5671 ?
                                DefaultServiceBusPort :
                                endpoint.Port.ToString();
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ServiceBusReceiverReceiveMessagesAsyncIntegration.cs
bouwkast: Ah interesting thanks for the context makes sense then the way it is

tracer/src/Datadog.Trace/Tagging/AzureServiceBusTags.cs
bouwkast: Ah hmm okay 

Are those ServiceBus integrations not Azure Service Bus? 😅

bouwkast (commented):
Looks good to me, but seeing 2 minute timeouts everywhere in the build/test related to the docker image?

`nc: bad address 'azureservicebus-emulator'`

Eventually it looks like the 2nd retry gets it going and I do see the tests running successfully 

`2025-09-09T12:01:20.8665192Z 12:01:20 [DBG] [xUnit.net 00:06:51.66] Datadog.Trace.ClrProfiler.IntegrationTests: SUCCESS: Datadog.Trace.ClrProfiler.IntegrationTests.Azure.AzureServiceBusAPMTests.TestSendMessagesAsyncIntegration(7.18.4, v1) (1.1988216s)`

lucaspimentel (approved):
LGTM. I left some comments about reducing heap allocations, but since this is disabled by default and we're doing a second pass, we can do those later, if you want.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -330,6 +339,119 @@ private static PropagationContext ExtractPropagatedContextFromHttp<T>(T context,
...
+                {
+                    return default;
+                }
+
+                bool areAllTheSame = spanContexts.Count == 1 ||
+                                     (spanContexts.Count > 1 && AreAllContextsIdentical(spanContexts));
+
+                if (!areAllTheSame)
+                {
+                    Log.Warning("Multiple different contexts found in ServiceBus messages. Using first context for parentship.");
```
duncanpharvey, line 404: Is this logic needed if we're only using span links and not reparenting?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ServiceBusReceiverReceiveMessagesAsyncIntegration.cs
```diff
@@ -0,0 +1,210 @@
...
+            if (message?.TryDuckCast<IServiceBusReceivedMessage>(out var serviceBusMessage) == true)
+            {
+                var messageId = serviceBusMessage.MessageId;
+                if (!string.IsNullOrEmpty(messageId))
+                {
+                    span.SetTag(Tags.MessagingMessageId, messageId);
+                }
+            }
+        }
+
```
duncanpharvey, line 169: Can `messaging.batch.message_count` (multiple messages received, either in a batch or a list) be set on the consumer spans as well?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ServiceBusSenderSendMessagesAsyncIntegration.cs
```diff
@@ -0,0 +1,86 @@
...
+        var endpoint = instance.Connection?.ServiceEndpoint;
+        if (endpoint != null)
+        {
+            tags.NetworkDestinationName = endpoint.Host;
+            // https://learn.microsoft.com/en-us/dotnet/api/system.uri.port?view=net-8.0#remarks
+            tags.NetworkDestinationPort = endpoint.Port is -1 or 5671 ?
+                                DefaultServiceBusPort :
+                                endpoint.Port.ToString();
+        }
+
```
duncanpharvey: Can `messaging.batch.message_count` (multiple messages sent, either in a batch or a list) and `messaging.message_id` (single message sent) be set on the producer spans as well?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ServiceBusSenderSendMessagesAsyncIntegration.cs
```diff
@@ -0,0 +1,86 @@
...
+[EditorBrowsable(EditorBrowsableState.Never)]
+public class ServiceBusSenderSendMessagesAsyncIntegration
+{
+    private const string OperationName = "azure_servicebus.send";
+    private const string MessagingType = "servicebus";
+    private const string DefaultServiceBusPort = "5671";
+
+    private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(ServiceBusSenderSendMessagesAsyncIntegration));
+
+    internal static CallTargetState OnMethodBegin<TTarget>(TTarget instance, ref IEnumerable messages, ref CancellationToken cancellationToken)
```
duncanpharvey, line 35: Does this support when `ServiceBusMessageBatch` is sent with `ServiceBusSender.SendMessagesAsync`?

Also what about the other methods?

- `ServiceBusSender.SendMessageAsync`
- `ServiceBusSender.ScheduleMessageAsync`
- `ServiceBusSender.ScheduleMessagesAsync`

https://learn.microsoft.com/en-us/dotnet/api/azure.messaging.servicebus.servicebussender?view=azure-dotnet

Screenshot 2025-09-17 at 5 17 29 PM

duncanpharvey (commented):
Some additional questions:
- Will the batch instrumentation be implemented in a future PR? This is where we would have a `create` span for each message added to a batch and each `create` span would link to the `send` and `receive` spans. It would be on by default and disabled with `DD_TRACE_AZURE_SERVICEBUS_BATCH_LINKS_ENABLED`
- Are the `messaging.*` and `network.*` span attributes added for Azure Function consumers? I see those span attributes for Azure Service Bus producers and consumers
- Are there any differences between queues and topic that would make it worth testing topics as well? I see the Azure Service Bus Emulator is configured with a queue but no topic

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
pablomartinezbernardo: Not really **needed**, the logic is only used to decide whether to write a warning log. Explained in more detail in [this convo](https://github.com/DataDog/dd-trace-dotnet/pull/7413/files#r2321255888), the function span is still reparented to the receive span

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ServiceBusReceiverReceiveMessagesAsyncIntegration.cs
pablomartinezbernardo: Done! See 3b7488416028c1a5c25668b7d7457d7caa2af4bc. Originally thought this was a tag meant for batches (Like the batch entity, not many messages)

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ServiceBusSenderSendMessagesAsyncIntegration.cs
pablomartinezbernardo: For SendMessageAsync

(At least for ServiceBus) it is a common pattern that the single element override [calls the multiple elements override](https://github.com/Azure/azure-sdk-for-net/blob/deb2a0c3c0b853a21c76f18463f70bb2510de858/sdk/servicebus/Azure.Messaging.ServiceBus/src/Sender/ServiceBusSender.cs#L193).

For ScheduleMessages... thanks for that, because I didn't know it existed. Added it, extracted the logic to the common utils because it is the same as SendMessages, see 3b7488416028c1a5c25668b7d7457d7caa2af4bc

While I was at it, refactored the sample app to make it simpler to understand because it was starting to get too complex

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/ServiceBusSenderSendMessagesAsyncIntegration.cs
pablomartinezbernardo: 👍  see 3b7488416028c1a5c25668b7d7457d7caa2af4bc

pablomartinezbernardo:
> * Will the batch instrumentation be implemented in a future PR? This is where we would have a `create` span for each message added to a batch and each `create` span would link to the `send` and `receive` spans. It would be on by default and disabled with `DD_TRACE_AZURE_SERVICEBUS_BATCH_LINKS_ENABLED`

Yes, initially split it because we hadn't defined batching, now I kept it like this to try and get a smaller PR

> * Are the `messaging.*` and `network.*` span attributes added for Azure Function consumers? I see those span attributes for Azure Service Bus producers and consumers

Are `network.destination.*` tags expected to exist in consumers?

Edit: To answer to myself: No they are not. But server.address is, so I added it fc0bc7bc7e07e44b47ca16558b9ba50954185a3d

> * Are there any differences between queues and topic that would make it worth testing topics as well? I see the Azure Service Bus Emulator is configured with a queue but no topic

There's differences in the receiver constructor, but once you have the instance the usage is identical, we are not tracing any function that has anything to do with queue or topic distinctions

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
lucaspimentel: This is probably fine, then. I'm just worried about spamming the logs if it happens too much.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/AzureServiceBusCommon.cs
```diff
@@ -55,5 +59,68 @@ internal static long GetMessageSize<T>(T message)
...
+                span.SetTag(Tags.MessagingBatchMessageCount, messageCount.ToString());
+            }
+
+            if (messageCount == 1 && messages != null)
+            {
+                foreach (var message in messages)
+                {
+                    var duckTypedMessage = message?.DuckCast<IServiceBusMessage>();
+                    singleMessageId = duckTypedMessage?.MessageId;
+                    break;
```
bricefriha, line 104: I have a hard time understanding the reason behind the choice of putting a `break` here.

If we know there is only one message, implied by the earlier check, wouldn't it be simpler to access it using an index?  something like: 'message[0].DuckCast<IServiceBusMessage>'

PR #7417, opened 2025-08-19
Handle TypeLoadException when retrieving HostingEnvironment

NachoEchevarria:
## Summary of changes

This PR improves resilience in the application name resolution logic for ASP.NET sites.

When the tracer is loaded into w3wp.exe, it attempts to probe System.Web.Hosting.HostingEnvironment to determine the IIS site name. In scenarios where ASP.NET Core is running on the .NET Framework CLR (an unsupported hosting model), HostingEnvironment cannot be loaded and triggers a TypeLoadException.

Previously, this bubbled up into error tracking, creating noisy/unhelpful error reports.

Changes:
Wrap the System.Web.Hosting.HostingEnvironment probe in a try/catch block.
On TypeLoadException, log a warning instead of letting the exception propagate.

### DIFF
--- a/tracer/src/Datadog.Trace/TracerManagerFactory.cs
+++ b/tracer/src/Datadog.Trace/TracerManagerFactory.cs
@@ -590,9 +590,9 @@ namespace Datadog.Trace
                     return true;
                 }
             }
-            catch (TypeLoadException)
+            catch (TypeLoadException ex)
             {
-                Log.Warning("Unable to determine ASP.NET site name: HostingEnvironment type could not be loaded. This is expected when running ASP.NET Core on the .NET Framework CLR, which is not supported.");
+                Log.Warning(ex, "Unable to determine ASP.NET site name: HostingEnvironment type could not be loaded. This is expected when running ASP.NET Core on the .NET Framework CLR, which is not supported.");
             }
 #endif
             siteName = default;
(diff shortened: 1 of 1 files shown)


tracer/src/Datadog.Trace/TracerManagerFactory.cs
```diff
@@ -579,15 +579,21 @@ private static string GetApplicationName(TracerSettings settings)
...
+                // System.Web.dll is only available on .NET Framework
+                if (System.Web.Hosting.HostingEnvironment.IsHosted)
+                {
+                    // if this app is an ASP.NET application, return "SiteName/ApplicationVirtualPath".
+                    // note that ApplicationVirtualPath includes a leading slash.
+                    siteName = (System.Web.Hosting.HostingEnvironment.SiteName + System.Web.Hosting.HostingEnvironment.ApplicationVirtualPath).TrimEnd('/');
+                    return true;
+                }
+            }
+            catch (TypeLoadException)
```
andrewlock: Is it worth including the exception in the log message? I'm not sure if it is, or if it would just confuse customers 🤷‍♂️

andrewlock (approved):

tracer/src/Datadog.Trace/TracerManagerFactory.cs
NachoEchevarria: I will add it. I guess it will not hurt. Thanks!

PR #7419, opened 2025-08-19
[Azure Functions] Restore files moved by Azure Functions SDK (in-process functions)

lucaspimentel:
## Summary of changes

Improve support for in-process Azure Functions my restoring the files moved by the SDK.

## Reason for change

The Azure Functions SDK for in-process functions moves files from `$(PublishDir)**\*.dll` to `$(PublishDir)bin\` (keeping the directory structure). This breaks the tracer because the files are not found where expected. Furthermore, it splits the tracer files into two paths because it moves our managed files like `net6.0/Datadog.Trace.dll` and our Windows native files like `win-x64/Datadog.Tracer.Native.dll`, but it doesn't move the `loader.conf` files or Linux native files like `linux-x64/Datadog.Tracer.Native.so`.

See https://github.com/Azure/azure-functions-vs-build-sdk/blob/7a3e505c2382ee9c461985baef1603e980740ca6/src/Microsoft.NET.Sdk.Functions.MSBuild/Targets/Microsoft.NET.Sdk.Functions.Publish.targets#L134-L154

## Implementation details

Add a build target that runs after the SDK's `_GenerateFunctionsAndCopyContentFiles` target. This new target moves any files in `$(PublishDir)bin\datadog\**\*.dll` back to `$(PublishDir)datadog\`, maintaining the directory structure.

## Test coverage

TODO

## Other details
Fixes APMSVLS-136


tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.props
```diff
@@ -0,0 +1,6 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <!-- Allow opting out of fix for in-process Azure Functions -->
+    <RestoreDatadogFiles Condition="'$(RestoreDatadogFiles)' == ''">true</RestoreDatadogFiles>
```
andrewlock, line 4: IIRC, .props files are added at the _start_ of the build, _before_ the csproj contents, right? In which case `$(RestoreDatadogFiles)` will always be empty, so this will always be added. Which is fine, because you can always override it. In other words, the Condition is harmless but unnecessary I think?

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.targets
```diff
@@ -0,0 +1,22 @@
+﻿<!-- These build step is a workaround for the Azure Functions SDK moving our *.dll files from datadog/ to bin/datadog/
+     See https://github.com/Azure/azure-functions-vs-build-sdk/blob/7a3e505c2382ee9c461985baef1603e980740ca6/src/Microsoft.NET.Sdk.Functions.MSBuild/Targets/Microsoft.NET.Sdk.Functions.Publish.targets#L134-L154 -->
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Target Name="_RestoreDatadogFiles"
+          AfterTargets="_GenerateFunctionsAndCopyContentFiles"
```
andrewlock, line 5: This ensures it runs after `_GenerateFunctionsAndCopyContentFiles`, but it doesn't ensure _when_ after, right? Does that matter? Do we need a `BeforeTargets` check too, to ensure correct ordering?

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.targets
```diff
@@ -0,0 +1,22 @@
...
+    </ItemGroup>
+
+    <!-- move the files back -->
+    <Move SourceFiles="@(DatadogFilesAfter)"
+          DestinationFiles="$(PublishDir)datadog\%(RecursiveDir)%(Filename)%(Extension)"
+          OverwriteReadOnlyFiles="true" />
+
+    <RemoveDir Directories="$(PublishDir)bin\datadog"/>
+
+    <Message Importance="High"
```
andrewlock: I wonder if we should have any policies about the message levels 🤔 High seems a bit intrusive perhaps? 🤷‍♂️

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
```diff
@@ -58,6 +58,9 @@
       <PackageCopyToOutput>false</PackageCopyToOutput>
     </None>
 
+    <None Include="Datadog.AzureFunctions.props" Pack="true" PackagePath="buildTransitive\Datadog.AzureFunctions.props" />
```
andrewlock, line 61: Does the `Datadog.AzureFunctions` project actually _work_ if it's transitively referenced? 🤔 if not, should we just be in `build` for simplicity?

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.props
lucaspimentel: You can define msbuild properties from the command-line, which I gather are read first because this works:
```
dotnet publish -p:RestoreDatadogFiles=false
```

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.targets
lucaspimentel: Oops, copy/paste mistake, not intentional.
I removed `Importance="High"` to use the default importance (`Normal`?).

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.targets
lucaspimentel: Good idea. Added a `BeforeTargets` to be safe, although it seems like to could have run _last_ and still work.

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
lucaspimentel: Content files (all our tracer files) are not transitive by default because `PrivateAssets` defaults to `contentfiles;analyzers;build`, but users can make them transitive by removing `contentfiles` from the default list, e.g. `PrivateAssets="none"`.

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.targets
```diff
@@ -0,0 +1,22 @@
...
+      <DatadogFilesAfter Include="$(PublishDir)bin\datadog\**\*.dll"/>
+    </ItemGroup>
+
+    <!-- move the files back -->
+    <Move SourceFiles="@(DatadogFilesAfter)"
+          DestinationFiles="$(PublishDir)datadog\%(RecursiveDir)%(Filename)%(Extension)"
+          OverwriteReadOnlyFiles="true" />
+
+    <RemoveDir Directories="$(PublishDir)bin\datadog"/>
+
```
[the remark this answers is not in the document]
lucaspimentel: Good call out, friend, but the directory will be empty, since it can only contain dll files moved by the SDK, which we have moved back before `<RemoveDir>`.

tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.props
andrewlock: oooh, good point, I didn't think about that 👍 (side note, I may need to update some of my oss projects 😅 )

PR #7420, opened 2025-08-20
[OTEL] Metrics API Support - Configurations & Telemetry

link04:
## Summary of changes
This PR implements configuration parsing needed for OpenTelemetry OTLP Metrics Collecting and Exporting as part of the broader [OpenTelemetry Metrics API support RFC](https://docs.google.com/document/d/1Zkpk6QdlRQ67459fqHk8_SlJKb3-ZqRg-3juLHtWBtU/edit?tab=t.0#heading=h.ps254zut7txp). Key additions include:
- OTLP Configuration Keys: Added all standard OpenTelemetry environment variables for metrics export configuration
- Fallback Hierarchy: Implemented proper fallback from metrics-specific configs to general OTLP configs to defaults

This is the first of a series of 3 PRs I hope to put together to complete this work.

## Reason for change
The [RFC] OpenTelemetry Metrics API support requires the .NET tracer to act as a drop-in replacement for the OpenTelemetry SDK, allowing customers to migrate from OTel SDK to Datadog SDK with zero friction. This requires supporting all standard OpenTelemetry configuration environment variables for metrics export.
Business Impact: Enables customers using OpenTelemetry Metrics API to switch to Datadog without code changes, supporting our strategic initiative to be OTel-native.

## Implementation details
Added the configurations below, for some I needed to add both metrics configs and fallbacks also tracer settings, for others I needed to add only testing and to telemetry.     :white_check_mark: Added | 🟣 Updated | 🟦 Untouched

**Datadog SDK Configurations (2/2):**
```1. DD_METRICS_OTEL_ENABLED``` - 🟣
```2. DD_TAGS``` - 🟦

**OpenTelemetry Metrics SDK Configurations (4/4):**
```1. OTEL_RESOURCE_ATTRIBUTES``` - 🟦
```2. OTEL_METRICS_EXPORTER``` - 🟣
```3. OTEL_METRIC_EXPORT_INTERVAL``` - :white_check_mark:
```4. OTEL_METRIC_EXPORT_TIMEOUT``` - :white_check_mark:

**OTLP Exporter Configurations (9/9):**
```1. OTEL_EXPORTER_OTLP_METRICS_PROTOCOL``` - :white_check_mark:
```2. OTEL_EXPORTER_OTLP_METRICS_ENDPOINT``` - :white_check_mark:
```3. OTEL_EXPORTER_OTLP_METRICS_HEADERS``` - :white_check_mark:
```4. OTEL_EXPORTER_OTLP_METRICS_TIMEOUT``` - :white_check_mark:
```5. OTEL_EXPORTER_OTLP_PROTOCOL (fallback)``` - :white_check_mark:
```6. OTEL_EXPORTER_OTLP_ENDPOINT (fallback)``` - :white_check_mark:
```7. OTEL_EXPORTER_OTLP_HEADERS (fallback)```  - :white_check_mark:
```8. OTEL_EXPORTER_OTLP_TIMEOUT (fallback)```  - :white_check_mark:
```9. OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE``` - :white_check_mark:

## Test coverage
Added to TracerSettingsTests different tests for each configuration combos and fallbacks expected behavior, also accounting for customers adding invalid/unsupported inputs for the multiple choice configs.
(Will add/update the tests as the other PRs are coming to completion).

## Other details

Jira: [APMAPI-1571](https://datadoghq.atlassian.net/browse/APMAPI-1571)

### CODE REVIEW

File: tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.OpenTelemetry.cs [hunk 1]
```diff
@@ -67,6 +59,95 @@ internal class OpenTelemetry
...
+            /// </summary>
+            public const string MetricsExporter = "OTEL_METRICS_EXPORTER";
+
+            /// <summary>
+            /// Configuration key to set the export interval for metrics in milliseconds.
+            /// Specifies the time interval between the start of two export attempts.
+            /// Default value is 10000ms (10s) for Datadog.
+            /// This deviates from OpenTelemetry specification default of 60000ms (60s).
+            /// </summary>
+            public const string MetricExportInterval = "OTEL_METRIC_EXPORT_INTERVAL";
```
andrewlock: tiny nit: Generally I prefer having the unit in the name for the settings so that it's clear this is `ms` not `s`. But won't fight it if you disagree 😅 
```suggestion
            public const string MetricExportIntervalMs = "OTEL_METRIC_EXPORT_INTERVAL";
```

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.OpenTelemetry.cs
```diff
@@ -67,6 +59,95 @@ internal class OpenTelemetry
...
+            /// Takes precedence over <see cref="ExporterOtlpEndpoint"/>.
+            /// Expects values like `http://localhost:4317` for grpc, `unix:///path/to/socket.sock` for UDS,
+            /// `\\.\pipename\` for Windows Named Pipes.
+            /// When using HTTP protocol with fallback, v1/metrics should be appended.
+            /// Default value for http/protobuf: `http://localhost:4318/v1/metrics`.
+            /// </summary>
+            public const string ExporterOtlpMetricsEndpoint = "OTEL_EXPORTER_OTLP_METRICS_ENDPOINT";
+
+            /// <summary>
+            /// Configuration key to set custom headers for OTLP metrics export.
```
andrewlock, line 105:
```suggestion
            /// Configuration key to set custom headers for OTLP metrics export.
            /// Takes precedence over <see cref="ExporterOtlpHeaders"/>.

```

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.OpenTelemetry.cs
```diff
@@ -67,6 +59,95 @@ internal class OpenTelemetry
...
+            /// Format: api-key=key,other=value.
+            /// </summary>
+            public const string ExporterOtlpMetricsHeaders = "OTEL_EXPORTER_OTLP_METRICS_HEADERS";
+
+            /// <summary>
+            /// Configuration key to set the request timeout for OTLP metrics export in milliseconds.
```
andrewlock, line 112:
```suggestion
            /// Configuration key to set the request timeout for OTLP metrics export in milliseconds.
            /// Takes precedence over <see cref="ExporterOtlpTimeout"/>.

```

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.OpenTelemetry.cs
```diff
@@ -67,6 +59,95 @@ internal class OpenTelemetry
...
+            /// Default value is 10000ms.
+            /// </summary>
+            public const string ExporterOtlpMetricsTimeout = "OTEL_EXPORTER_OTLP_METRICS_TIMEOUT";
+
+            /// <summary>
+            /// Configuration key to set the temporality preference for OTLP metrics export.
+            /// Supported values: delta, cumulative, lowmemory.
+            /// Default value is delta for Datadog.
```
andrewlock, line 121:
```suggestion
            /// Default value is delta for Datadog.
            /// This deviates from OpenTelemetry specification default of cumulative.

```

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.OpenTelemetry.cs
```diff
@@ -67,6 +59,95 @@ internal class OpenTelemetry
...
+            /// </summary>
+            public const string ExporterOtlpMetricsTemporalityPreference = "OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE";
+
+            /// <summary>
+            /// Configuration key to set the OTLP protocol (fallback for metrics-specific protocol).
+            /// Used when <see cref="ExporterOtlpMetricsProtocol"/> is not set.
+            /// Valid values: grpc, http/protobuf, http/json.
```
andrewlock:
```suggestion
            /// Valid values: grpc, http/protobuf, http/json, defaults to http/protobuf.
```

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.OpenTelemetry.cs
```diff
@@ -67,6 +59,95 @@ internal class OpenTelemetry
...
+            /// </summary>
+            public const string ExporterOtlpProtocol = "OTEL_EXPORTER_OTLP_PROTOCOL";
+
+            /// <summary>
+            /// Configuration key to set the OTLP endpoint URL (fallback for metrics-specific endpoint).
+            /// Used when <see cref="ExporterOtlpMetricsEndpoint"/> is not set.
+            /// When using HTTP protocol with fallback, v1/metrics should be appended.
```
andrewlock:
```suggestion
            /// Expects values like `http://localhost:4317` for grpc, `unix:///path/to/socket.sock` for UDS,
            /// `\\.\pipename\` for Windows Named Pipes.
            /// When using HTTP protocol with fallback, v1/metrics should be appended.
            /// Default value for http/protobuf: `http://localhost:4318/v1/metrics`.
```

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -398,20 +398,78 @@ internal TracerSettings(IConfigurationSource? source, IConfigurationTelemetry te
 
             StatsComputationInterval = config.WithKeys(ConfigurationKeys.StatsComputationInterval).AsInt32(defaultValue: 10);
 
+            OtelMetricsExporter = config
+                        .WithKeys(ConfigurationKeys.OpenTelemetry.MetricsExporter)
+                        .AsString(defaultValue: "otlp");
```
andrewlock: I think this should actually be `OtelMetricsExporterEnabled`, it should be a boolean, and it should be set based on the `otelRuntimeMetricsEnabled` property below. i.e. it should be set like this:

```csharp
            var otelRuntimeMetricsEnabled = config
                                           .WithKeys(ConfigurationKeys.OpenTelemetry.MetricsExporter)
                                           .AsBoolResult(value => value switch
                                            {
                                                not null when string.Equals(value, "otlp", StringComparison.OrdinalIgnoreCase)
                                                    => ParsingResult<bool>.Success(result: false),
                                                not null when string.Equals(value, "none", StringComparison.OrdinalIgnoreCase)
                                                    => ParsingResult<bool>.Success(result: false),
                                                _ => ParsingResult<bool>.Failure()
                                            });
            OtelMetricsExporterEnabled  = otelRuntimeMetricsEnabled.WithDefault(true); // enabled by default
```

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -398,20 +398,78 @@ internal TracerSettings(IConfigurationSource? source, IConfigurationTelemetry te
...
+                              .AsBoolResult(
+                                   value => string.Equals(value, "none", StringComparison.OrdinalIgnoreCase)
+                                                ? ParsingResult<bool>.Success(result: false)
+                                                : ParsingResult<bool>.Failure());
             _runtimeMetricsEnabled = config
-                                   .WithKeys(ConfigurationKeys.RuntimeMetricsEnabled)
-                                   .AsBoolResult()
-                                   .OverrideWith(in otelRuntimeMetricsEnabled, ErrorLog, defaultValue: false);
+                            .WithKeys(ConfigurationKeys.RuntimeMetricsEnabled)
+                            .AsBoolResult()
```
andrewlock: The ordering here is important. Because we're now _overriding_ `_runtimeMetricsEnabled` based on the value of `otelRuntimeMetricsEnabled`, we need to make some changes:

- We have to call `_runtimeMetricsEnabled = config.WithKeys(ConfigurationKeys.RuntimeMetricsEnabled).AsBoolResult()` _before_ setting `otelRuntimeMetricsEnabled`.
- We need to do the overriding the other way around.

e.g. 

```csharp
var runtimeMetricsEnabledResult = config
                            .WithKeys(ConfigurationKeys.RuntimeMetricsEnabled)
                            .AsBoolResult();

var otelRuntimeMetricsEnabled = config
                              .WithKeys(ConfigurationKeys.OpenTelemetry.MetricsExporter)
                              .AsBoolResult( /* */ );

_runtimeMetricsEnabled = otelRuntimeMetricsEnabled
                              .OverrideWith(in runtimeMetricsEnabledResult, ErrorLog, defaultValue: false);
```

_Although_ I'm not even sure this is correct, because I think we need to factor `DD_METRICS_OTEL_ENABLED` in there somehow too 😅 That's a _lot_ of confusing configuration - I'm not actually sure `ConfigurationBuilder` supports it currently 🤔

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -398,20 +398,78 @@ internal TracerSettings(IConfigurationSource? source, IConfigurationTelemetry te
...
+                            .OverrideWith(in otelRuntimeMetricsEnabled, ErrorLog, defaultValue: false);
+
+            OtelMetricExportInterval = config
+                            .WithKeys(ConfigurationKeys.OpenTelemetry.MetricExportInterval)
+                            .AsInt32(defaultValue: 10000);
```
andrewlock: nit:
```suggestion
                            .AsInt32(defaultValue: 10_000);
```

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -398,20 +398,78 @@ internal TracerSettings(IConfigurationSource? source, IConfigurationTelemetry te
...
+
+            OtelMetricExportTimeout = config
+                            .WithKeys(ConfigurationKeys.OpenTelemetry.MetricExportTimeout)
+                            .AsInt32(defaultValue: 7500);
```
```suggestion
                            .AsInt32(defaultValue: 7_500);
```

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -398,20 +398,78 @@ internal TracerSettings(IConfigurationSource? source, IConfigurationTelemetry te
...
+                                 .WithKeys(ConfigurationKeys.OpenTelemetry.ExporterOtlpMetricsProtocol, ConfigurationKeys.OpenTelemetry.ExporterOtlpProtocol)
+                                 .GetAs(
+                                      defaultValue: new(OtlpProtocol.HttpProtobuf, "http/protobuf"),
+                                      converter: x => x.ToLowerInvariant() switch
+                                      {
+                                          "http/protobuf" => OtlpProtocol.HttpProtobuf,
+                                          "grpc" => OtlpProtocol.Grpc,
+                                          "http/json" => OtlpProtocol.HttpJson,
+                                          _ => UnsupportedOtlpProtocol(x),
+                                      },
```
andrewlock: Calling `ToLowerInvariant()` allocates, it's better to do a case-insensitive comparison, even though it makes the `switch` uglier!

```suggestion
                                      converter: x => x.ToLowerInvariant() switch
                                      {
                                          not null when string.Equals(x, "http/protobuf", StringComparison.OrdinalIgnoreCase) => OtlpProtocol.HttpProtobuf,
                                          not null when string.Equals(x, "grpc", StringComparison.OrdinalIgnoreCase) => OtlpProtocol.Grpc,
                                          not null when string.Equals(x, "http/json", StringComparison.OrdinalIgnoreCase) => OtlpProtocol.HttpJson,
                                          _ => UnsupportedOtlpProtocol(x),
                                      },
```

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -398,20 +398,78 @@ internal TracerSettings(IConfigurationSource? source, IConfigurationTelemetry te
...
+                                      validator: null);
+
+            OtlpMetricsEndpoint = config
+                          .WithKeys(ConfigurationKeys.OpenTelemetry.ExporterOtlpMetricsEndpoint, ConfigurationKeys.OpenTelemetry.ExporterOtlpEndpoint)
+                          .AsString(defaultValue: "http://localhost:4318/v1/metrics");
```
andrewlock: You might want to pass a `validator:` here to validate that the default value is one of the valid formats, e.g. using `Uri.TryParse()`, and potentially making this type a `Uri` -  it will probably save you headaches later, e.g. in the block below

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -398,20 +398,78 @@ internal TracerSettings(IConfigurationSource? source, IConfigurationTelemetry te
...
+
+            if ((OtlpMetricsProtocol == OtlpProtocol.HttpProtobuf || OtlpMetricsProtocol == OtlpProtocol.HttpJson) && !OtlpMetricsEndpoint.EndsWith("/v1/metrics"))
+            {
+                OtlpMetricsEndpoint = $"{OtlpMetricsEndpoint.TrimEnd('/')}/v1/metrics";
+            }
```
andrewlock: I think this could be problematic 🤔 e.g. if the customer sets `"http://localhost:4318/v1/metrics"` then this will set it to `http://localhost:4318/v1/metrics/v1/metrics" 😅 What's the expected (desired) behaviour when a customer sets it to any of these, for example?

- `http://localhost:4318/proxy`
- `http://localhost:4318/v1/metrics/proxy`

You probably should work with the `Uri` type directly in any case, it will make it easier to avoid the edge cases I think

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -398,20 +398,78 @@ internal TracerSettings(IConfigurationSource? source, IConfigurationTelemetry te
...
+
+            OtlpMetricsHeaders = config
+                                .WithKeys(ConfigurationKeys.OpenTelemetry.ExporterOtlpMetricsHeaders, ConfigurationKeys.OpenTelemetry.ExporterOtlpHeaders)
+                                .AsDictionaryResult(separator: '=')
+                                .WithDefault(defaultValue: ReadOnlyDictionary.Empty)
+                                .Where(kvp => !string.IsNullOrWhiteSpace(kvp.Key))
+                                .ToDictionary(kvp => kvp.Key.Trim(), kvp => kvp.Value?.Trim() ?? string.Empty);
```
andrewlock: Congratulations, you found a bug in `ConfigurationBuilder`, this shouldn't be possible 😂 
```
                                .AsDictionaryResult(separator: '=')
                                .WithDefault(defaultValue: ReadOnlyDictionary.Empty)
```

Fixed here:
- https://github.com/DataDog/dd-trace-dotnet/pull/7421

Or at least I _thought_ it was, but I realised you actually needed to set the separator, which is why you did this 😅 I guess we should probably allow providing that too, but in the mean time, you can do this instead. Note that the bug is that you should only be passing `DefaultResult<T>` to `WithDefault`. 

Fixed that one here:
- https://github.com/DataDog/dd-trace-dotnet/pull/7422

Also, we can just pass in a null value instead of `ReadOnlyDictionary.Empty`, otherwise we end up creating a new (empty) dictionary in the final `ToDictionary()`.

You should then be able to do this:
```suggestion
                                .AsDictionary(separator: '=')
                                .WithDefault(defaultValue: null, "[]")
                                ?.Where(kvp => !string.IsNullOrWhiteSpace(kvp.Key))
                                .ToDictionary(kvp => kvp.Key.Trim(), kvp => kvp.Value?.Trim() ?? string.Empty) ?? ReadOnlyDictionary.Empty;
```

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -398,20 +398,78 @@ internal TracerSettings(IConfigurationSource? source, IConfigurationTelemetry te
...
+
+            OtlpMetricsTimeout = config
+                            .WithKeys(ConfigurationKeys.OpenTelemetry.ExporterOtlpMetricsTimeout, ConfigurationKeys.OpenTelemetry.ExporterOtlpTimeout)
+                            .AsInt32(defaultValue: 10000);
```
andrewlock: nit:
```suggestion
                            .AsInt32(defaultValue: 10_000);
```

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -1621,6 +1737,12 @@ internal string GetDefaultHttpClientExclusions()
             }
         }
 
+        private static ParsingResult<OtlpProtocol> UnsupportedOtlpProtocol(string inputValue)
+        {
+            Log.Error("Unsupported OTLP protocol '{Protocol}'. Supported values are 'http/protobuf', 'grpc', 'http/json'. Using default: http/protobuf", inputValue);
```
andrewlock: This should only be a warning (we don't want error logs for this), but is it actually necessary at all? 🤔 We don't log warnings for other incorrect values?
```suggestion
            Log.Warning("Unsupported OTLP protocol '{Protocol}'. Supported values are 'http/protobuf', 'grpc', 'http/json'. Using default: http/protobuf", inputValue);
```

tracer/test/Datadog.Trace.Tests/Configuration/TracerSettingsTests.cs
```diff
@@ -1382,6 +1382,125 @@ public void LogsInjectionEnabled(string value, bool expected)
...
+            settings.OtelMetricExportTimeout.Should().Be(expected);
+        }
+
+        [Theory]
+        [InlineData(null, null, OtlpProtocol.HttpProtobuf)]
+        [InlineData("invalid", null, OtlpProtocol.HttpProtobuf)]
+        [InlineData("grpc", null, OtlpProtocol.Grpc)]
+        [InlineData("grpc", "http/protobuf", OtlpProtocol.Grpc)]
+        [InlineData(null, "http/json", OtlpProtocol.HttpJson)]
+        public void OtlpProtocolFallbacks(string metricsProtocol, string generalProtocol, object expected)
```
andrewlock, line 1450: Nice, some good tests here 👍

tracer/test/Datadog.Trace.Tests/Configuration/TracerSettingsTests.cs
```diff
@@ -1382,6 +1382,125 @@ public void LogsInjectionEnabled(string value, bool expected)
...
+            settings.OtlpMetricsProtocol.Should().Be((OtlpProtocol)expected);
+        }
+
+        [Theory]
+        [InlineData("v1", SchemaVersion.V1)]
+        [InlineData("V1", SchemaVersion.V1)]
+        [InlineData("", SchemaVersion.V0)]
+        [InlineData(null, SchemaVersion.V0)]
+        [InlineData("v1 ", SchemaVersion.V0)]
+        public void MetadatsaSchemaVersion(string value, object expected)
```
andrewlock: Did this one slip in when it shouldn't? 😅

tracer/test/Datadog.Trace.Tests/Configuration/TracerSettingsTests.cs
```diff
@@ -1382,6 +1382,125 @@ public void LogsInjectionEnabled(string value, bool expected)
             tracerSettings.LogsInjectionEnabled.Should().Be(expected);
         }
 
+        [Theory]
+        [MemberData(nameof(BooleanTestCases), false)]
+        public void OpenTelemetryMetricsEnabled(string value, bool expected)
```
andrewlock, line 1391: We should probably test the impact on runtime metrics enabled setting too...

andrewlock (commented):
Nice, some great documentation for the public APIs 👍 Most of the comments are just tiny nits, but there's some things we'll need to address with the ordering in the telemetry I think

tracer/src/Datadog.Trace/Configuration/OtlpProtocol.cs
```diff
@@ -0,0 +1,29 @@
...
+/// </summary>
+internal enum OtlpProtocol
+{
+    /// <summary>
+    /// HTTP with Protocol Buffers encoding
+    /// </summary>
+    HttpProtobuf = 0,
+
+    /// <summary>
+    /// gRPC protocol
```
zacharycmontoya: Adding a small suggestion to help other reviewers who haven't read all the OTLP specs

```suggestion
    /// gRPC with Protocol Buffers encoding
```

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.OpenTelemetry.cs
zacharycmontoya: I think this suggestion is good, but I want to call out that the default values you should include are:
- gRPC: `http://localhost:4317`
- http/protobuf: `http://localhost:4318`

zacharycmontoya: To clarify, since this is a signal-agnostic configuration, there is no additional `v1/metrics` path for the http/protobuf protocol

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.OpenTelemetry.cs
```diff
@@ -67,6 +59,95 @@ internal class OpenTelemetry
...
+            /// </summary>
+            public const string ExporterOtlpMetricsProtocol = "OTEL_EXPORTER_OTLP_METRICS_PROTOCOL";
+
+            /// <summary>
+            /// Configuration key to set the OTLP endpoint URL for metrics.
+            /// Takes precedence over <see cref="ExporterOtlpEndpoint"/>.
+            /// Expects values like `http://localhost:4317` for grpc, `unix:///path/to/socket.sock` for UDS,
+            /// `\\.\pipename\` for Windows Named Pipes.
+            /// When using HTTP protocol with fallback, v1/metrics should be appended.
+            /// Default value for http/protobuf: `http://localhost:4318/v1/metrics`.
```
zacharycmontoya:
```suggestion
            /// Default value for grpc: `http://localhost:4317`.
            /// Default value for http/protobuf: `http://localhost:4318/v1/metrics`.
```

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
zacharycmontoya: You should also remove the `ToLowerInvariant()` call on x for the switch expression!

tracer/test/Datadog.Trace.Tests/Configuration/TracerSettingsTests.cs
```diff
@@ -1382,6 +1382,125 @@ public void LogsInjectionEnabled(string value, bool expected)
...
+            var source = CreateConfigurationSource((ConfigurationKeys.FeatureFlags.OpenTelemetryMetricsEnabled, value));
+            var settings = new TracerSettings(source);
+
+            settings.OpenTelemetryMetricsEnabled.Should().Be(expected);
+        }
+
+        [Theory]
+        [InlineData("otlp", "otlp")]
+        [InlineData("none", "none")]
+        [InlineData("prometheus", "prometheus")]
```
zacharycmontoya: Currently an output value of `"prometheus"` doesn't really mean anything to us since we don't implement a prometheus exporter (though the `"prometheus"` input is a good test case!). I think you should follow Andrew's other suggestions to test whether we are emitting OTLP metrics or not and then unit test that behavior. Then you could test what result we'd expect when the `"prometheus"` input is passed in

tracer/test/Datadog.Trace.Tests/Configuration/TracerSettingsTests.cs
```diff
@@ -1382,6 +1382,125 @@ public void LogsInjectionEnabled(string value, bool expected)
...
+
+            settings.MetadataSchemaVersion.Should().Be((SchemaVersion)expected);
+        }
+
+        [Theory]
+        [InlineData("http://custom:4318/v1/metrics", null, "http://custom:4318/v1/metrics")]
+        [InlineData("http://custom:4318/v1/metrics", "http://fallback:4318", "http://custom:4318/v1/metrics")]
+        [InlineData(null, "http://fallback:8080", "http://fallback:8080/v1/metrics")]  // HTTP fallback gets v1/metrics
+        [InlineData(null, null, "http://localhost:4318/v1/metrics")]  // Default
+        public void OtlpEndpointFallbacks(string metricsEndpoint, string generalEndpoint, string expected)
```
zacharycmontoya: Two small suggestions:
1. Can you explicitly set the protocol to `http/protobuf` in the test code below?
2. Can you similar methods to test the fallback logic when the `grpc` protocol or the `http/json` protocol is used?

tracer/test/Datadog.Trace.Tests/Configuration/TracerSettingsTests.cs
```diff
@@ -1382,6 +1382,125 @@ public void LogsInjectionEnabled(string value, bool expected)
...
+            var settings = new TracerSettings(source);
+
+            settings.OtlpMetricsEndpoint.Should().Be(expected);
+        }
+
+        [Theory]
+        [InlineData("DELTA", OtlpTemporality.Delta)]
+        [InlineData("cumulative", OtlpTemporality.Cumulative)]
+        [InlineData("loWmemOry", OtlpTemporality.LowMemory)]
+        [InlineData(null, OtlpTemporality.Delta)]
```
zacharycmontoya, line 1496: I think we could also use a bogus value like this
```suggestion
        [InlineData(null, OtlpTemporality.Delta)]
        [InlineData("bad-value", OtlpTemporality.Delta)]
```

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
zacharycmontoya: Ideally I would like:
1. The in-code property like Andrew's suggested `OtelMetricsExporterEnabled` to reflect whether we should turn on/off the exporter.
2. The telemetry to display a string value to see if users have tried to configure other exporters.

@andrewlock is it possible to achieve both and do you think that's useful?

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
zacharycmontoya: Hey Andrew, back to report that `OverrideWith` was a terrible name that I think I introduced, because it doesn't actually override. 😅 The logic here looks fine because we only take this "override" OTel value when the Datadog equivalent is not set.

zacharycmontoya: So in general, this looks OK to me. Here's the logic/reasoning why I think so:
- Runtime metrics enabled is controlled by our `DD_RUNTIME_METRICS_ENABLED` flag, and based on line 414 it has a default value of `false` if not set.
- If both `DD_RUNTIME_METRICS_ENABLED` and `OTEL_METRICS_EXPORTER` are set, the OTel value is ignored when calculating runtime metrics
- If `DD_RUNTIME_METRICS_ENABLED` is not explicitly set, we'll enable/disable based on how we've configured `OTEL_METRICS_EXPORTER`. If `OTEL_METRICS_EXPORTER=none` we disable emitting runtime metrics.
- For our purposes, `DD_METRICS_OTEL_ENABLED` controls listening to custom metrics so we don't factor that in here

😄

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
zacharycmontoya: I'm also thinking that you could get the general `ConfigurationKeys.OpenTelemetry.ExporterOtlpEndpoint` first. That would help make sure you properly handle the general URL versus the metric-specific URL

zacharycmontoya: Also, let's make sure to have the following logic:
- If the user has set `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT`, use exactly that value provided
- If the user has set `OTEL_EXPORTER_OTLP_ENDPOINT` and they're using HTTP, let's append the `v1/metrics` path

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
link04: Yeah this one is part of the RFC but after talking with Zach we've agreed to change it to ` Warning`.

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
link04: Hey @andrewlock even after merging your changes, I am unable to use  `.WithDefault(new(null, "[]")` with null that using `ReadOnlyDictionary.Empty` allows me to bypass, not sure if I am doing something wrong or my understanding is wrong here.
`Cannot convert null literal to non-nullable reference type.`

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
andrewlock: > The telemetry to display a string value to see if users have tried to configure other exporters.

I'm pretty sure that's exactly what the above should do 🙂 But we could/should also write a unit test and check!

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
andrewlock: > If DD_RUNTIME_METRICS_ENABLED is not explicitly set, we'll enable/disable based on how we've configured OTEL_METRICS_EXPORTER. If OTEL_METRICS_EXPORTER=none we disable emitting runtime metrics.

Ah ok, I actually thought this from the RFC:

> A value of none disables the emission of metrics, as well as disabled APM runtime metrics

I thought that meant that it was supposed to override the DD value. In which case, all good!

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
andrewlock: > I am unable to use `.WithDefault(new(null, "[]")`

Yeah, it's a bit awkward, because there are _two_ overloads, and _neither_ of these take a `DefaultValue<T>` 😅 

So you actually want this:

```csharp
.AsDictionary(defaultValue: null, "[]")
```

Sorry! (I updated my suggestion above)

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -398,20 +398,95 @@ internal TracerSettings(IConfigurationSource? source, IConfigurationTelemetry te
...
+                        telemetryValue: $"{defaultUri}/v1/metrics"),
+                    validator: null,
+                    converter: uriString => new Uri(uriString));
+
+            OtlpMetricsHeaders = config
+                            .WithKeys(ConfigurationKeys.OpenTelemetry.ExporterOtlpMetricsHeaders, ConfigurationKeys.OpenTelemetry.ExporterOtlpHeaders)
+                            .AsDictionaryResult(separator: '=')
+                            .WithDefault(new DefaultResult<IDictionary<string, string>>(new Dictionary<string, string>(), "[]"))
+                            .Where(kvp => !string.IsNullOrWhiteSpace(kvp.Key))
+                            .ToDictionary(kvp => kvp.Key.Trim(), kvp => kvp.Value?.Trim() ?? string.Empty);
```
link04, line 478: @andrewlock this what I end up coming up with, let me know if you think this is okay or can be improved on this PR.

tracer/test/Datadog.Trace.Tests/Configuration/TracerSettingsTests.cs
link04: Thank you @zacharycmontoya, updated some here: https://github.com/DataDog/dd-trace-dotnet/pull/7420/files#diff-6c397a3040cd9399ed76241b767d80d81c92020e7ed8650bae95e3835e201337R1397-R1418

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
link04: Since the value of "otlp" doesn't enable runtime metrics I've use the approach in the latest commit where I do make into OtelMetricsExporterEnabled boolean but only set the value to true if **otlp** is found and false for anything else which is then checked for **none** for runtime metrics, test: https://github.com/DataDog/dd-trace-dotnet/pull/7420/files#diff-6c397a3040cd9399ed76241b767d80d81c92020e7ed8650bae95e3835e201337R1397-R1418

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
link04: Hey guys Ive added a comment here: https://github.com/DataDog/dd-trace-dotnet/pull/7420/files#r2324112089

And wanrted to check if there are any other recommendations based on the decision that otlp doesn't override DD runtime metrics.

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -398,20 +398,95 @@ internal TracerSettings(IConfigurationSource? source, IConfigurationTelemetry te
...
+                .GetAs(
+                    defaultValue: new DefaultResult<Uri>(
+                        result: OtlpMetricsProtocol switch
+                        {
+                            OtlpProtocol.Grpc => OtlpEndpoint,
+                            _ => new Uri(OtlpEndpoint, "/v1/metrics")
+                        },
+                        telemetryValue: $"{defaultUri}/v1/metrics"),
+                    validator: null,
+                    converter: uriString => new Uri(uriString));
```
link04: Hey @zacharycmontoya, please let me know what you think of this approach or if it looks good to go.

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
andrewlock: LGTM. I hate that this still doesn't work like I want it too with the nullability but that's on me 😅

tracer/src/Datadog.Trace/Configuration/OtlpProtocol.cs
```diff
@@ -0,0 +1,29 @@
...
+
+/// <summary>
+/// Defines the available OTLP protocol types available for exporting.
+/// </summary>
+internal enum OtlpProtocol
+{
+    /// <summary>
+    /// HTTP with Protocol Buffers encoding
+    /// </summary>
+    HttpProtobuf = 0,
```
zacharycmontoya: Can you please re-order these values so Grpc=0 and HttpProbuf=1, just like in the OTel exporter? https://github.com/open-telemetry/opentelemetry-dotnet/blob/f153e130b46844b023ccde867d00ed757260969c/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/OtlpExportProtocol.cs

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -398,16 +398,91 @@ internal TracerSettings(IConfigurationSource? source, IConfigurationTelemetry te
...
+
+            OtelMetricsExporterEnabled = string.Equals(OtelMetricsExporter, "otlp", StringComparison.OrdinalIgnoreCase);
+
             var otelRuntimeMetricsEnabled = config
-                                          .WithKeys(ConfigurationKeys.OpenTelemetry.MetricsExporter)
-                                          .AsBoolResult(
-                                               value => string.Equals(value, "none", StringComparison.OrdinalIgnoreCase)
-                                                            ? ParsingResult<bool>.Success(result: false)
-                                                            : ParsingResult<bool>.Failure());
+                            .WithKeys(ConfigurationKeys.OpenTelemetry.MetricsExporter)
```
zacharycmontoya: @andrewlock is this going to cause multiple telemetry calls? If we need to do only one call, I think we may be able to add this as a validator on the first call to only accept `none` and `otlp` as valid

Note: I don't think this is a blocking concern

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
zacharycmontoya: I think there's a small tweak that the default `telemetryValue` should also switch based on Grpc vs HTTP. Other than that, this looks good!

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -398,16 +398,91 @@ internal TracerSettings(IConfigurationSource? source, IConfigurationTelemetry te
...
+                            .AsBoolResult(value => value switch
+                            {
+                                not null when string.Equals(value, "none", StringComparison.OrdinalIgnoreCase)
```
zacharycmontoya: Blocking comment: Since `"otlp"` is now a valid variable for the `OTEL_METRICS_EXPORTER` environment variable, we should no longer mark it as `ParsingResult<bool>.Failure()`. This means in the test `tracer/test/Datadog.Trace.Tests/Configuration/TracerSettingsTests.cs::RuntimeMetricsEnabled` where you have the existing test:
```csharp
                (null, "otlp") => Count.OpenTelemetryConfigInvalid,
```

it should now be:
```csharp
                (null, "otlp") => null,
```

Maybe what you should do is change the existing `otelRuntimeMetricsEnabled` variable so it's more like `otelDisableAllMetrics`. Then, you'd have the following logic:
- `"none"` => `ParsingResult<bool>.Success(result: true)`
- `"otlp"` => `ParsingResult<bool>.Success(result: false)`
- default => `ParsingResult<bool>.Failure()`

And then you can pass `!otelDisableAllMetrics` to the `_runtimeMetricsEnabled` OverrideWith call

zacharycmontoya (commented):
I have one blocking concern with the weird interaction between OTEL_METRICS_EXPORTER and the runtime metrics config. Otherwise, just left some comments for smaller tweaks

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
link04: Ty added some tests here: 

https://github.com/DataDog/dd-trace-dotnet/pull/7420/files#diff-6c397a3040cd9399ed76241b767d80d81c92020e7ed8650bae95e3835e201337R1458-R1488

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
link04: So I tried that an a couple of other things but becuase of the parameter OverrideWith works with and how `otlp` basically doesn't do anything for runtime metrics I've done the below where it seems we collect telemetry correctly for the enabling exporter setting itself tho still return invalid if Runtime Metrics not set and the exporter is `otlp`, is this something we should look into deeper to have the `otlp` value also enable runtime metrics?
- https://github.com/DataDog/dd-trace-dotnet/pull/7420/files#diff-8afc30ae77b35082c438827c3ab04d7dce486226317aeef23e40e7bd62ba2b75R401-R416

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
link04: Also discussed [here](https://github.com/DataDog/dd-trace-dotnet/pull/7420#discussion_r2325553989) and TLDR I am not validating `otlp` as a value when trying to check it for Runtime Metrics, Andrew and I also discussed this `otlp` value not doing anything for Runtime Metrics before so let me know what you guys think, or if we should revisit this override scenario...

andrewlock: Ah yes, it will 🤔 And essentially the latter result overwrites the first. I _think_ we can get a similar result like this, but honestly I could definitely be wrong depending on the exact behaviour we _want_ here 😅 

```csharp
var otelExporterResult = config
                        .WithKeys(ConfigurationKeys.OpenTelemetry.MetricsExporter)
                        .AsBoolResult(value => value switch
                          {
                              not null when string.Equals(value, "none", StringComparison.OrdinalIgnoreCase) => ParsingResult<bool>.Success(result: false),
                              not null when string.Equals(value, "otlp", StringComparison.OrdinalIgnoreCase) => ParsingResult<bool>.Success(result: true),
                               _ => ParsingResult<bool>.Failure()
                            });

OtelMetricsExporter = otelExporterResult.ConfigurationResult switch
            {
                { IsValid: true, Result: var result } => result ? "otel" : "none",
                _ => "otel", // invalid or default value
            }

OtelMetricsExporterEnabled = otelExporterResult.ConfigurationResult is { IsValid: true, Result: true };
```

Happy to discuss offline if needs be

andrewlock: As discussed, I think this fixes the runtime metrics aspect:

```c#
var runtimeMetricsEnabledResult = config
                .WithKeys(ConfigurationKeys.RuntimeMetricsEnabled)
                .AsBoolResult()

if (runtimeMetricsEnabledResult.IsPresent && otelExporterResult.IsPresent)
{
    ErrorLog.LogDuplicateConfiguration(ConfigurationKeys.RuntimeMetricsEnabled, ConfigurationKeys.OpenTelemetry.MetricsExporter);
}

_runtimeMetricsEnabled = runtimeMetricsEnabledResult.WithDefault(false);
```

link04: I've added some updates based on this convo and some other offline ones were some tests broke and we discussed  how to get default telemetry and proper logging:

https://github.com/DataDog/dd-trace-dotnet/pull/7420/files#diff-8afc30ae77b35082c438827c3ab04d7dce486226317aeef23e40e7bd62ba2b75R401

PR #7421, opened 2025-08-20
Add missing `GetDictionary()` method to `ConfigurationBuilder`

andrewlock:
## Summary of changes

Add missing `GetDictionary()` method to `ConfigurationBuilder`

## Reason for change

We were missing a method that allows providing a default instance for a dictionary up-front. This was forcing using other overloads that are suboptimal for telemetry.

## Implementation details

Add the missing helper that allows providing a default value. I allowed passing a null value in here, as it avoids unnecessary allocations if you don't actually need the dictionary. A good example was the fixed usage in `InitializeHeaderTags` which was creating an empty dictionary instead of taking the fast-path `null` route (and potentially actually giving _incorrect_ settings in the `DynamicSettings` I think 

## Test coverage

Meh

## Other details

Spotted when reviewing
- https://github.com/DataDog/dd-trace-dotnet/pull/7420

### DIFF
--- a/tracer/src/Datadog.Trace/Configuration/ConfigurationSources/Telemetry/ConfigurationBuilder.cs
+++ b/tracer/src/Datadog.Trace/Configuration/ConfigurationSources/Telemetry/ConfigurationBuilder.cs
@@ -354,6 +354,10 @@ internal readonly struct ConfigurationBuilder(IConfigurationSource source, IConf
         public IDictionary<string, string> AsDictionary(Func<IDictionary<string, string>> getDefaultValue, string defaultValueForTelemetry)
             => AsDictionary(allowOptionalMappings: false, getDefaultValue: getDefaultValue, defaultValueForTelemetry);
 
+        [return: NotNullIfNotNull(nameof(defaultValue))]
+        public IDictionary<string, string>? AsDictionary(IDictionary<string, string>? defaultValue, string defaultValueForTelemetry)
+            => AsDictionary(allowOptionalMappings: false, defaultValue, defaultValueForTelemetry);
+
         [return: NotNullIfNotNull(nameof(getDefaultValue))]
         public IDictionary<string, string>? AsDictionary(
             bool allowOptionalMappings,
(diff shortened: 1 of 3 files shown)


tracer/src/Datadog.Trace/Configuration/ConfigurationSources/Telemetry/ConfigurationBuilder.cs
```diff
@@ -375,6 +379,32 @@ public IDictionary<string, string> AsDictionary(Func<IDictionary<string, string>
...
+                Telemetry.Record(Key, defaultValueForTelemetry, recordValue: true, ConfigurationOrigins.Default);
+            }
+
+            var result = GetDictionaryResult(allowOptionalMappings, separator: ':');
+            if (result is { Result: { } ddResult, IsValid: true })
+            {
+                return ddResult;
+            }
+
+            if (result.IsPresent)
```
bouwkast, line 400:
```suggestion
            if (defaultValue is not null && result.IsPresent)
```

I see this pattern in others do we need to do the same here?

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -1397,7 +1397,7 @@ internal static TracerSettings FromDefaultSourcesInternal()
         {
             var configurationDictionary = config
                    .WithKeys(key)
-                   .AsDictionary(allowOptionalMappings: true, () => new Dictionary<string, string>(), string.Empty);
+                   .AsDictionary(allowOptionalMappings: true, defaultValue: null, "[]");
```
bouwkast, line 1400: I wonder if this could have caused some of the issues where headers as tags would sometimes seemingly not work?
But I don't think it was via dynamic so probably not.

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/Telemetry/ConfigurationBuilder.cs
andrewlock: Yeah, that was deliberate, so that we can do things like the other change made in `TracerSettings` i.e. you can pass in a `null` dictionary value, and then use `?? new Dictionary<string, string>()` so that you can avoid creating the default value up front 🙂

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
andrewlock: Yeah, I think the only place this actually changes behaviour is if the tags are set via dynamic settings (and I actually have other concerns about how that works now 😅)

PR #7422, opened 2025-08-20
More `ConfigurationBuilder` refactoring to avoid incorrect usages

andrewlock:
## Summary of changes

- Fix `ClassConfigurationResultWithKey.WithDefault()` to take a `DefaultResult<T>`
- Don't allow creating arbitrary `StructConfigurationResultWithKey`
- Remove (inline) buggy `RecordTelemetry<>` method

## Reason for change

When reviewing https://github.com/DataDog/dd-trace-dotnet/pull/7420, I noticed there was an incorrect usage of `WithDefault()` due to limitations in the type system, so this tries to close that gap. Unfortunately, that required more changes than I would like, due to limitations in generics and constraints.

## Implementation details

- Inline the `RecordTelemetry<T>` method
- Explicitly limit the types of `StructConfigurationResultWithKey<T>` that may be created by using static factory methods and a `private` ctor
- Fix `ClassConfigurationResultWithKey.WithDefault()` to take a `DefaultResult<T>` so that we always have a telemetry value to record. This will be slightly clunky if you try to do it with `string`, but we can close that gap later if we need to

## Test coverage

Added a couple of small unit tests

## Other details

Spotted when reviewing

- https://github.com/DataDog/dd-trace-dotnet/pull/7420


zacharycmontoya (approved):
That was a doozy, nice work!

PR #7423, opened 2025-08-20
Safe get process start info variables.

NachoEchevarria:


NachoEchevarria:
> Great work! Just to clarify, was this actually causing incorrect behaviour for the customer? Or was this just an error in telemetry logs? because it _looks_ like it should be the latter?

Thanks! If ProcessStartInfo.Environment throws and that exception isn’t handled inside the integration, the ProcessStart instrumentation won’t create the command_execution span. This would also affect ASM ability to detect certain vulnerabilities and attacks in IAST and RASP.

PR #7424, opened 2025-08-20
[Phase 2] chore(trace-exporter): enable data pipeline based on Windows

ganeshnj:
## Summary of changes

Where is phase 1?
https://github.com/DataDog/datadog-aas-extension/pull/336

### DIFF
--- a/tracer/src/Datadog.Trace/FrameworkDescription.cs
+++ b/tracer/src/Datadog.Trace/FrameworkDescription.cs
@@ -69,6 +69,11 @@ namespace Datadog.Trace
             return string.Equals(OSPlatform, OSPlatformName.Windows, StringComparison.OrdinalIgnoreCase);
         }
 
+        public bool IsLinux()
+        {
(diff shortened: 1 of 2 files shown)


tracer/src/Datadog.Trace/FrameworkDescription.cs
```diff
@@ -69,6 +69,11 @@ public bool IsWindows()
             return string.Equals(OSPlatform, OSPlatformName.Windows, StringComparison.OrdinalIgnoreCase);
         }
 
+        public bool IsLinux()
```
ganeshnj: this should have been in the https://github.com/DataDog/dd-trace-dotnet/pull/7425/files PR

andrewlock: Do we really need this? Aren't we going to go Windows then everywhere? 🤔 I don't think there's much reason to not tbh

andrewlock (approved):

PR #7322, opened 2025-08-01
[Profiler] Make timer_create-based CPU profiler default

gleocadie:
## Summary of changes

This is another take at making `timer_create`-based CPU profiler the default one on linux ([previous attempt](https://github.com/DataDog/dd-trace-dotnet/pull/6315))

## Reason for change
The new Linux Cpu profiler is based on timer_create. It has been under development last quarter. We enabled it recently by default and let it run in CIs to find remaining issues.
Those issues have been fixed.
## Implementation details
In the configuration set the default CPU profiler (linux) to timer_create
## Test coverage

Tests are updated accordingly

PR #7427, opened 2025-08-21
Revert "[Profiler] Make `timer_create`-based CPU profiler default (#7322)"

andrewlock:
## Summary of changes

This reverts commit 5676931004042cfb4af0468567f00c056d2d467f.

## Reason for change

We noticed a lot of (flaky) crashing on the _trimmed_ smoke tests, shortly after #7322  was merged. Investigating this by testing a revert of the PR and re-running in CI resulted in no crashes. Out of an abundance of caution, we are choosing to revert the change to investigate before the next release.

## Implementation details

This reverts commit 5676931004042cfb4af0468567f00c056d2d467f 
- #7322

## Test coverage

This is the test, but I also re-ran all the trimmed tests 6 times and saw no failures

## Other details

We'll aim to revert this revert and investigate next week


chrisnas (approved):
LGTM

PR #7428, opened 2025-08-22
Fix null reference exception

NachoEchevarria:

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AspNet/AsyncControllerActionInvoker_EndInvokeAction_Integration.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AspNet/AsyncControllerActionInvoker_EndInvokeAction_Integration.cs
@@ -10,8 +10,9 @@ using System.Web;
 using Datadog.Trace.AspNet;
 using Datadog.Trace.ClrProfiler.CallTarget;
 using Datadog.Trace.Configuration;
-using Datadog.Trace.ExtensionMethods;
 using Datadog.Trace.Logging;
+
+#nullable enable
 
 namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.AspNet
 {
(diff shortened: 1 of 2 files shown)


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AspNet/AsyncControllerActionInvoker_EndInvokeAction_Integration.cs
```diff
@@ -52,8 +53,8 @@ public class AsyncControllerActionInvoker_EndInvokeAction_Integration
         /// <returns>Return value of the method</returns>
         internal static CallTargetReturn<TResult> OnMethodEnd<TTarget, TResult>(TTarget instance, TResult returnValue, Exception exception, in CallTargetState state)
         {
-            Scope scope = null;
-            Scope proxyScope = null;
+            Scope? scope = null;
+            Scope? proxyScope = null;
             var httpContext = HttpContext.Current;
```
andrewlock: `HttpContext.Current` can technically be null in some cases, so we should probably guard that in some places

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AspNet/AsyncControllerActionInvoker_EndInvokeAction_Integration.cs
```diff
@@ -131,7 +132,7 @@ private static void OnRequestCompletedAfterException(HttpContext httpContext, Sc
                     proxyScope.Span.Finish(finishTime);
                 }
 
-                HttpContextHelper.AddHeaderTagsFromHttpResponse(httpContext, proxyScope);
+                HttpContextHelper.AddHeaderTagsFromHttpResponse(httpContext, scope);
```
andrewlock, line 136: Good catch 👍

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AspNet/AsyncControllerActionInvoker_EndInvokeAction_Integration.cs
NachoEchevarria: Thanks! I have updated the code.

PR #7431, opened 2025-08-22
Use mock agent in exploration tests

andrewlock:
## Summary of changes

Use the mock agent in exploration tests instead of the real agent

## Reason for change

In the .NET 10 branch, we updated the VMs, and also pulled a new version of the agent. For _some_ reason, (e.g. new agent) the agent docker container often doesn't start up, making the tests super flaky, due to failing to send traces to the agent.

I've spent _far_ to long trying to get the real agent to work in CI, and in the end I give up. We absolutely don't need the "real" agent, so it's not worth the hassle IMO.

## Implementation details

- Fix the standalone "MockAgent" project (which just hosts the `MockTracerAgent` class)
- Rework it to be based on the standard .NET hosting model instead of doing custom stuff
- Run the mock agent in the Windows Exploration tests only.

## Test coverage

[It works in the .NET 10 branch](https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=186178&view=logs&j=8345afb5-d2a0-5c0b-09c6-12ff46d9917b) which is the important thing.

## Other details

Yeah, we _could_ have used the Python test-agent here (dd-apm-test-agent), but we'd have to use the docker container, and rebuild all the VMs etc. Meh, this works.

### DIFF
--- a/tracer/tools/MockAgent/Program.cs
+++ b/tracer/tools/MockAgent/Program.cs
@@ -1,117 +1,148 @@
-// See https://aka.ms/new-console-template for more information
 using CommandLine;
 using Datadog.Trace.TestHelpers;
 using MockAgent;
+using Xunit.Abstractions;
 
-var showTraces = true;
-var showMetrics = true;
+Options? options = null;
+Parser.Default.ParseArguments<Options>(args)
+    .WithParsed(o => options = o)
+    .WithNotParsed(o => throw new Exception("Error parsing options"));
 
-EventHandler<EventArgs<IList<IList<MockSpan>>>> displayTraces = (sender, args) =>
+var builder = Host.CreateApplicationBuilder(args);
+builder.Services.AddSingleton(options!);
+builder.Services.AddHostedService<Worker>();
+builder.Services.AddWindowsService();
+
+var host = builder.Build();
+host.Run();
+
+public class Worker : BackgroundService, ITestOutputHelper
 {
-    if (!showTraces)
+    private readonly Options _opts;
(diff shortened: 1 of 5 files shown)


tracer/tools/MockAgent/Program.cs
```diff
@@ -1,117 +1,148 @@
...
+           {
+               agent.MetricsReceived += DisplayMetrics;
+           }
+       }
+
+       if (_opts.UnixDomainSockets)
+       {
+           // can't enable on windows
+           string? metricsUds = null;
+           bool useMetrics = !string.IsNullOrEmpty(metricsUds);
```
NachoEchevarria, line 76: useMetrics will always be false

tracer/tools/MockAgent/Program.cs
```diff
@@ -1,117 +1,148 @@
...
+           if (Environment.OSVersion.Platform is PlatformID.Unix or PlatformID.MacOSX)
+           {
+               metricsUds = _opts.MetricsUnixDomainSocketPath;
+           }
+
```
NachoEchevarria:
```suggestion
           string? metricsUds = null;
           if (Environment.OSVersion.Platform is PlatformID.Unix or PlatformID.MacOSX)
           {
               metricsUds = _opts.MetricsUnixDomainSocketPath;
           }
           bool useMetrics = !string.IsNullOrEmpty(metricsUds);
```
useMetrics will always be false otherwise.

andrewlock: Oops, good catch, thanks!

PR #7434, opened 2025-08-22
[Tracer] fix: Re-use runtime metrics writer resources to limit memory growth

zacharycmontoya:
## Summary of changes
Updates `TracerManagerFactory.CreateTracerManager` to pass and re-use the previous `RuntimeMetricsWriter`, if runtime metrics is enabled for the new TracerManager. We still create a new `IDogStatsd` client every time, so any updated Agent settings are adopted by the DogStatsd client, but this could be further optimized at a later time to re-use the client if none of its configuration (e.g. host/port/tags) have changed.

## Reason for change
We've observed a scenario where the number of `RuntimeEventListener` instances continues to grow, consuming more and more memory. This happens whenever new Dynamic Configuration settings are received by the tracer and runtime metrics are enabled. This PR resolves the issue.

## Implementation details
When creating the new `TracerManager`, pass in the previous `RuntimeMetricsWriter` instance and only update the `IDogStatsd` object with new settings. This makes sure that we maintain only one `RuntimeMetricsWriter` instance while getting up-to-date DogStatsD settings throughout the application lifetime.

## Test coverage
Adds a small unit test to confirm that the previous `RuntimeMetricsWriter` is re-used. Additionally, local testing was done to confirm that the number of `Datadog.Trace.RuntimeMetrics.RuntimeEventListener` objects does not grow when Dynamic Configuration is updated.

### Without the fix
After some number of Dynamic Configuration settings were made in the Datadog UI, a dump was taken with the following analysis:
```
> dumpheap -type Datadog.Trace.RuntimeMetrics.RuntimeEventListener
         Address               MT           Size
    0159c3c7fe88     7ffe4e801810             88
    0159c442ed28     7ffe4e801810             88
    0159c44eac60     7ffe4e801810             88
    0159c4500f00     7ffe4e801810             88
    0159c454ea68     7ffe4e801810             88

Statistics:
          MT Count TotalSize Class Name
7ffe4e801810     5       440 Datadog.Trace.RuntimeMetrics.RuntimeEventListener
Total 5 objects, 440 bytes
```

Then after one additional Dynamic Configuration update was made in the Datadog UI, a dump was taken with the following analysis:
```
> dumpheap -type Datadog.Trace.RuntimeMetrics.RuntimeEventListener
         Address               MT           Size
    0159c3c7fe88     7ffe4e801810             88
    0159c442ed28     7ffe4e801810             88
    0159c44eac60     7ffe4e801810             88
    0159c4500f00     7ffe4e801810             88
    0159c454ea68     7ffe4e801810             88
    0159c61fef48     7ffe4e801810             88

Statistics:
          MT Count TotalSize Class Name
7ffe4e801810     6       528 Datadog.Trace.RuntimeMetrics.RuntimeEventListener
Total 6 objects, 528 bytes
```

### With the fix
After some number of Dynamic Configuration settings were made in the Datadog UI, a dump was taken with the following analysis:
```
> dumpheap -type Datadog.Trace.RuntimeMetrics.RuntimeEventListener
         Address               MT           Size
    01527107faf0     7ffe4e7acc00             88

Statistics:
          MT Count TotalSize Class Name
7ffe4e7acc00     1        88 Datadog.Trace.RuntimeMetrics.RuntimeEventListener
Total 1 objects, 88 bytes
```

Then after one additional Dynamic Configuration update was made in the Datadog UI, a dump was taken with the following analysis:
```
> dumpheap -type Datadog.Trace.RuntimeMetrics.RuntimeEventListener
         Address               MT           Size
    01527107faf0     7ffe4e7acc00             88

Statistics:
          MT Count TotalSize Class Name
7ffe4e7acc00     1        88 Datadog.Trace.RuntimeMetrics.RuntimeEventListener
Total 1 objects, 88 bytes
```

### DIFF
--- a/tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
+++ b/tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
@@ -39,7 +39,6 @@ namespace Datadog.Trace.RuntimeMetrics
 
         private readonly TimeSpan _delay;
 
-        private readonly IDogStatsd _statsd;
         private readonly Timer _timer;
 
         private readonly IRuntimeMetricsListener _listener;
@@ -52,6 +51,7 @@ namespace Datadog.Trace.RuntimeMetrics
 #endif
 
         private readonly ConcurrentDictionary<string, int> _exceptionCounts = new ConcurrentDictionary<string, int>();
+        private IDogStatsd _statsd;
         private int _outOfMemoryCount;
 
         // The time when the runtime metrics were last pushed
@@ -156,6 +156,11 @@ namespace Datadog.Trace.RuntimeMetrics
             _listener?.Dispose();
 #endif
             _exceptionCounts.Clear();
+        }
+
+        internal void UpdateStatsd(IDogStatsd statsd)
+        {
+            _statsd = statsd;
         }
 
         internal void PushEvents()
--- a/tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
+++ b/tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
@@ -160,7 +160,7 @@ namespace Datadog.Trace.RuntimeMetrics
 
         internal void UpdateStatsd(IDogStatsd statsd)
         {
-            _statsd = statsd;
+            Interlocked.Exchange(ref _statsd, statsd);
         }
 
         internal void PushEvents()
--- a/tracer/src/Datadog.Trace/TracerManagerFactory.cs
+++ b/tracer/src/Datadog.Trace/TracerManagerFactory.cs
@@ -132,7 +132,7 @@ namespace Datadog.Trace
 
             // Even if a previous statsd instance is provided, we need to update it with potentially new settings
             statsd = (settings.TracerMetricsEnabled || runtimeMetricsEnabled)
-                         ? (statsd ?? CreateDogStatsdClient(settings, defaultServiceName))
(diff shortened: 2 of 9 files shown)


tracer/src/Datadog.Trace/TracerManagerFactory.cs
```diff
@@ -137,9 +137,13 @@ internal TracerManager CreateTracerManager(
             agentWriter ??= GetAgentWriter(settings, settings.TracerMetricsEnabled ? statsd : null, rates => sampler.SetDefaultSampleRates(rates), discoveryService);
             scopeManager ??= new AsyncLocalScopeManager();
 
-            if (runtimeMetricsEnabled)
+            if (runtimeMetricsEnabled && runtimeMetrics is { })
```
anna-git, line 140: NIT maybe to avoid double testing `runtimeMetricsEnabled` 

```
if (!runtimeMetricsEnabled)
{
    runtimeMetrics = null;
}
else if (runtimeMetrics is { })
{
    runtimeMetrics.UpdateStatsd(statsd);
}
else
{
    runtimeMetrics = new RuntimeMetricsWriter(
        statsd,
        TimeSpan.FromSeconds(10),
        settings.IsRunningInAzureAppService);
}
````

**Review by anna-git — APPROVED** [review]
Nicely investigated, fixed and tested!

File: tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs [hunk 2]
```diff
@@ -158,6 +158,11 @@ public void Dispose()
             _exceptionCounts.Clear();
         }
 
+        internal void UpdateStatsd(IDogStatsd statsd)
+        {
+            _statsd = statsd;
```
andrewlock: Given it's a background thread that's running here, I think we need an interlocked (or volatile read and write, but this is easier) to ensure cross-thread visibility?

```suggestion
            Interlocked.Exchange(ref _statsd, statsd);
```

andrewlock (approved):
Thanks, I have Plans™️about how to back this out in the future, so that we can avoid all the disposes (because disposing `IDogStatsd` is problematic too unfortunately!), but glad to see this solves the immediate issue!

PR #7440, opened 2025-08-26
[IAST] Disable all Dataflow operations until aspects are loaded

daniel-romano-DD:
## Summary of changes
Disable all Dataflow operations (more specifically, `OnModuleLoad` and `OnModuleUnload`) when it's in iddle mode waiting for activation signal from the managed side

## Reason for change
With this change, Dataflow (call site instrumentation for IAST and RASP) is completely idle until a call to `InitEmbeddedCallSiteDefinitions` happens, effectively enabling all the mechanisms.

## Implementation details
Until now, Dataflow would collect modules info as they were loaded in the `OnModuleLoad` event. This has been disabled, and this module info retrieval happens on demand, when someone asks for it, if it's not found in the `ModuleInfo` dictionary.

### DIFF
--- a/tracer/src/Datadog.Tracer.Native/cor_profiler.cpp
+++ b/tracer/src/Datadog.Tracer.Native/cor_profiler.cpp
@@ -460,7 +460,7 @@ HRESULT STDMETHODCALLTYPE CorProfiler::AssemblyLoadFinished(AssemblyID assembly_
         {
             Logger::Info("AssemblyLoadFinished: Datadog.Trace.dll v", assembly_version, " matched profiler version v",
                          expected_version);
-            managed_profiler_loaded_app_domains.insert({assembly_info.app_domain_id, assembly_metadata.version});
+            managed_profiler_loaded_app_domains.insert({assembly_info.app_domain_id, assembly_info.manifest_module_id});
 
             // Load defaults values if the version are the same as expected
             if (assembly_metadata.version == expected_assembly_reference.version)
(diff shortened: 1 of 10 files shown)


tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
```diff
@@ -188,6 +188,7 @@ void Dataflow::Destroy()
     if (_initialized)
     {
         _initialized = false;
+        _loaded = false;
```
anna-git: I'm a bit confused regarding the difference between loaded and initialized?  maybe this is `_aspectsLoaded`? and maybe `_initialized` is `_disposed`?

dromanol: Yeah, that's mainly it. Poorly chosen names

tracer/src/Datadog.Tracer.Native/cor_profiler.cpp
```diff
@@ -2501,12 +2501,29 @@ bool CorProfiler::GetIntegrationTypeRef(ModuleMetadata& module_metadata, ModuleI
     return true;
 }
 
-bool CorProfiler::ProfilerAssemblyIsLoadedIntoAppDomain(AppDomainID app_domain_id)
+bool CorProfiler::ProfilerAssemblyIsLoadedIntoAppDomain(AppDomainID app_domain_id, ModuleID* pModuleId)
```
tonyredondo: you don't need to change the signature.

dromanol: Yeah, that was an unmade unneeded change. Fixed

tonyredondo (approved):

tracer/src/Datadog.Tracer.Native/iast/module_info.h
```diff
@@ -92,7 +87,7 @@ namespace iast
 
         virtual HRESULT InstrumentModule_Internal() { return S_FALSE; }
 
-        ModuleInfo* GetModuleInfoByName(WSTRING moduleName);
+        // ModuleInfo* GetModuleInfoByName(WSTRING moduleName);
```
andrewlock: nit:

```suggestion
```

robertpi (approved):

tracer/src/Datadog.Tracer.Native/iast/module_info.h
dromanol: And in the cpp also. Cleaned up

PR #7441, opened 2025-08-27
[Dynamic Instrumentation] Added native hot standby mode to support Remote Enablement and Stable Config

GreenMatan:
## Summary of changes
Added hot standby mode to Dynamic Instrumentation and Exception Replay products in the native side of the tracer to support Remote Enablement and Stable Config.

For DI/ER products to function correctly, the native side, among other things, need to inject a new field into each and every `StateMachine` type that is generated for async methods. Our injected field is used during the execution of DI/ER instrumentation to capture continuation of async flows.
The only point in time where it's possible to change type's layout is in `ModuleLoadFinished` event, and this event is fired only once per module.
Until Stable Configuration we only injected that field when DI/ER env vars were enable, but now we can not rely on these env var anymore as they're no longer the source of truth. With Stable Config we can not know if DI/ER will be enabled or not at startup. To accommodate that, the native side now acts as if DI/ER is enabled, just in case it will.
This is true not only for Stable Config, but also for Remote Enablement.

There are two new env vars:
1. `DD_DYNAMIC_INSTRUMENTATION_MANAGED_ACTIVATION_ENABLED`
2. `DD_EXCEPTION_REPLAY_MANAGED_ACTIVATION_ENABLED`

These env vars, as described in [Stable Config Standby mode doc](https://docs.google.com/document/d/1KEu4rgOnB3VOTiB_uztJGwT6w4dQ6FNCryVnMQItiP8/edit?tab=t.6njjxcv5o5he#heading=h.hgsgqqfp3ubt), are used to deactivate hot standby for ER/DI.

A big contribution to support remote enablement for DI/ER involved changing our injection strategy of async methods.
In DI/ER instrumentation, on `ModuleLoadFinished`, we inject a field of type `AsyncDebuggerState` into all `StateMachine` types. This field is defined in `Datadog.Trace`. When the managed loader fails/bails out in loading `Datadog.Trace`, it results in TypeLoadException as the runtime tried to implicitly load `Datadog.Trace` to no avail.
In this PR, the field type is changed to be `System.Object` to avoid that cumbersome.

## Implementation
The field that is injected and used in async methods to determine continuation is changed to be `System.Object`. To support that, the instrumentation of async methods for log probes has been changed - instead of accepting `AsyncDebuggerState` as the state, it's now accepting `System.Object`.
The entrypoints now goes through a new shim class `AsyncMethodDebuggerInvokerV2`, and from there to the underlying `AsyncMethodDebuggerInvoker` after materializing the state from `System.Object` -> `AsyncDebuggerState`.

In span probes on async methods the same concept has been applied, but, instead of a new class shim, new signatures were added that accept the state as `System.Object` and in similar fashion materializing to `AsyncDebuggerState` before transferring control the the previous functions.

## Test
Relying on existing test suite of debugger & tracer to cover the instrumentation changes

## Reason for change
Fully support Remote Enablement and Stable Config.

Resolves DEBUG-4421


tracer/src/Datadog.Tracer.Native/debugger_probes_instrumentation_requester.cpp
```diff
@@ -1027,13 +1014,8 @@ void DebuggerProbesInstrumentationRequester::ModuleLoadFinished_AddMetadataToMod
 
 HRESULT STDMETHODCALLTYPE DebuggerProbesInstrumentationRequester::ModuleLoadFinished(const ModuleID moduleId)
 {
-    if (!is_debugger_or_exception_debugging_enabled)
```
anna-git, line 1030: iiuc, this is the hot standby code that always needs to execute so that, in case, settings now read in the managed tracer turn on DI, it will work. Could we still implement the safety kill switch though `DD_DYNAMIC_INSTRUMENTATION_MANAGED_ACTIVATION_ENABLED` ? https://docs.google.com/document/d/1KEu4rgOnB3VOTiB_uztJGwT6w4dQ6FNCryVnMQItiP8/edit?tab=t.6njjxcv5o5he#heading=h.hgsgqqfp3ubt

So that basically, if `DD_DYNAMIC_INSTRUMENTATION_MANAGED_ACTIVATION_ENABLED` is false (it should be most of the time true), we dont have any native footprint for the product. Which means, you still need to read the env vars locally to fallback on the legacy mechanism (the current one until now)

GreenMatan: done in badbe44391babeb0cbadb3d186bc21700d3f6de9

GreenMatan: Nice catch! done in 04e74150bd1593deb24eed0cc7f43479ce1638bc

tylfin (approved):

tracer/src/Datadog.Tracer.Native/debugger_environment_variables.h
```diff
@@ -16,13 +16,18 @@ namespace environment
...
-    // Determines if the Dynamic Instrumentation (aka live debugger) is enabled.
+    // Determines if the Dynamic Instrumentation (aka live debugger) product is enabled.
     const WSTRING dynamic_instrumentation_enabled = WStr("DD_DYNAMIC_INSTRUMENTATION_ENABLED");
 
     // Determines if the Exception Replay product is enabled.
-    const WSTRING exception_debugging_enabled = WStr("DD_EXCEPTION_DEBUGGING_ENABLED"); // Old name
     const WSTRING exception_replay_enabled = WStr("DD_EXCEPTION_REPLAY_ENABLED");
 
+    // Determines if Dynamic Instrumentation should be controlled by Stable Configuration
+    const WSTRING dynamic_instrumentation_stable_config_enabled = WStr("DD_DYNAMIC_INSTRUMENTATION_MANAGED_ACTIVATION_ENABLED");
```
anna-git: NIT: I think in the naming and in comments we might not want to add any stable configuration notion as this is really just managed activation. Managed activation takes into [account multiple configuration sources](https://github.com/DataDog/dd-trace-dotnet/blob/16951283ad056fd5fe4e74aeaebf31ba4e92c3fa/tracer/src/Datadog.Trace/Configuration/ConfigurationSources/GlobalConfigurationSource.cs#L36), datadog.json, appconfig.xml, webconfig.xml, env vars, now hands off configs files, but tomorrow it could potentially be any config sources merged on the managed side that the native side doesn't really need to be aware of. Technically, if we didnt have stable config, it would still work the same, we could read `DD_DYNAMIC_INSTRUMENTATION_ENABLED` from datadog.json for example and apply it as long as `DD_DYNAMIC_INSTRUMENTATION_MANAGED_ACTIVATION_ENABLED` is true

tracer/src/Datadog.Tracer.Native/debugger_environment_variables_util.cpp
```diff
@@ -11,15 +11,17 @@ bool IsDynamicInstrumentationEnabled()
...
-        const auto old_exception_replay_flag = GetEnvironmentValue(environment::exception_debugging_enabled);
-        const auto new_exception_replay_flag = GetEnvironmentValue(environment::exception_replay_enabled);
-        sValue = (IsTrue(old_exception_replay_flag) || IsTrue(new_exception_replay_flag)) ? 1 : 0;
-    }
-
-    return sValue == 1;
+    CheckIfTrue(GetEnvironmentValue(environment::exception_replay_enabled));
+}
+
+bool IsDynamicInstrumentationStableConfigDisabled()
```
anna-git:
```suggestion
bool IsDynamicInstrumentationManagedActivationDisabled()
```

anna-git (approved):
thank you! I left one small NIT comment

tracer/src/Datadog.Tracer.Native/debugger_environment_variables_util.cpp
```diff
@@ -11,15 +11,17 @@ bool IsDynamicInstrumentationEnabled()
...
+{
+    CheckIfFalse(GetEnvironmentValue(environment::dynamic_instrumentation_stable_config_enabled));
```
anna-git: make sure if the env var isn't defined, we fallback on true but maybe it's the case already as we're testing false, mentioning just in case 😅

GreenMatan: Done in d21eeec872cc18a0cfb752c1f56dfe0957cde68d.

tracer/src/Datadog.Tracer.Native/debugger_environment_variables_util.cpp
GreenMatan: Done in d21eeec872cc18a0cfb752c1f56dfe0957cde68d.

tracer/src/Datadog.Tracer.Native/debugger_environment_variables.h
GreenMatan: Done in b6023b364bedf12ae204eb7b4e1b680510b59fba.

PR #7443, opened 2025-08-27
Add parameter names to MongoDb `IBsonSerializerProxy.Serialize()` method

andrewlock:
## Summary of changes

Adds duck-typing parameter names to MongoDb `IBsonSerializerProxy.Serialize()` method

## Reason for change

We have seen some errors with stack traces like the following:

My theory was that it's somehow related to:
- We're returning [a proxy](https://github.com/DataDog/dd-trace-dotnet/blob/73c63d6327adaa2a9532778a8fffd88a82caab7f/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/MongoDb/BsonSerialization/IBsonSerializerProxy.cs#L9) for [this type](https://github.com/mongodb/mongo-csharp-driver/blob/main/src/MongoDB.Bson/Serialization/IBsonSerializer.cs#L23), which has a non-generic `Serialize` method:

```csharp
void Serialize(object context, object args, object value);

// proxying this:
void Serialize(BsonSerializationContext context, BsonSerializationArgs args, object value);
```

- We get this [proxy](https://github.com/DataDog/dd-trace-dotnet/blob/d13598215b82b225430767e841838ac19b4b2f43/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/MongoDb/BsonSerialization/IBsonSerializerLookupProxy.cs#L15) via proxying another [static method](https://github.com/mongodb/mongo-csharp-driver/blob/main/src/MongoDB.Bson/Serialization/BsonSerializer.cs#L553):

```c#
public static IBsonSerializer LookupSerializer(Type type)
{
    return __serializerRegistry.GetSerializer(type);
}
```

The returned type, `EnumerableInterfaceImplementerSerializer<BsonDocument, BsonElement>` that we're proxying in the errors above explicitly implements the `IBsonSerializer` interface that we're ducktyping in a base type:

```c#
void IBsonSerializer.Serialize(BsonSerializationContext context, BsonSerializationArgs args, object value)
{
    Serialize(context, args, (TValue)value);
}
```

as well as [implicitly implementing ](https://github.com/mongodb/mongo-csharp-driver/blob/main/src/MongoDB.Bson/Serialization/Serializers/EnumerableSerializerBase.cs#L164)a generic version in a base type (which the non-generic version calls into)

```c#
public override void Serialize(BsonSerializationContext context, BsonSerializationArgs args, TValue value)
```

My theory was that we're "accidentally" trying to invoke the generic version.

@tonyredondo helped confirm that, by translating our dumped byte array body (`typeof(ducktype).GetMethod("Serialize").GetMethodBody().GetIlAsByteArray()`)  to this:

```msil
// Offsets are byte offsets into the IL stream

IL_0000: ldarg.0                                       // load the proxy
IL_0001: ldfld        /* field   0x04000001 */         // load the instance field in the proxy
IL_0006: ldarg.1                                       // load the first argument
IL_0007: castclass    /* typeref 0x0100000B */         // cast the object to the actual type of the target argument
IL_000C: ldarg.2                                       // load the second argument
IL_000D: dup                                           // duplicate the second argument in the stack (instance <- first arg casted <- second arg <- second arg)
IL_000E: isinst       /* typeref 0x0100000C */         // check if the second argument is an instance of something.... (this pop the dup) . Why we need to check if is instance of???
IL_0013: brtrue.s     IL_001C                          // if is true (is an instance, we jump to IL_001C)

IL_0015: pop                                           // We pop the second arg
IL_0016: newobj       /* memberref 0x0A000004 */       // Create the exception object
IL_001B: throw                                         // Throw the exception

IL_001C: unbox.any    /* typeref 0x0100000C */         // unbox? so the target second argument is a value type!!!?
IL_0021: ldarg.3                                       // load the third argument
IL_0022: castclass    /* typeref 0x01000006 */         // Cast the third argument to the target third argument type (if both sides were System.Object this castclass shouldn't be here, this could probably be the target `T` from the generic.
IL_0027: callvirt     /* memberref 0x0A000005 */       // Call the target method with the stack ( instance <- first arg casted <- second arg unboxed <- third arg casted )
IL_002C: ret                                           // Return the value.

// nothing left on the stack
```

## Implementation details

Add `ParameterTypeNames` to avoid ambiguity

## Test coverage

The existing tests should ensure we haven't broken anything key, but we won't know if this actually fixes the underlying issue until we release, because we can't repro the invalid cast locally.

Did a full test against all package versions, and the only failures were due to missing `language` tag on old (untested) package version snapshots, so updated those at the same time

## Other details


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/MongoDb/BsonSerialization/IBsonSerializerProxy.cs
```diff
@@ -4,9 +4,15 @@
...
+using Datadog.Trace.DuckTyping;
+
 namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.MongoDb.BsonSerialization;
 
+/// <summary>
+/// Proxy for https://github.com/mongodb/mongo-csharp-driver/blob/4027d482d14960364c14be1ca59f7f6e350042a3/src/MongoDB.Bson/Serialization/IBsonSerializer.cs#L23
+/// </summary>
 internal interface IBsonSerializerProxy
 {
+    [Duck(ParameterTypeNames = ["MongoDB.Bson.Serialization.BsonSerializationContext", "MongoDB.Bson.Serialization.BsonSerializationArgs", ClrNames.Object])]
```
andrewlock: I think we need the assembly actually
```suggestion
    [Duck(ParameterTypeNames = ["MongoDB.Bson.Serialization.BsonSerializationContext, MongoDB.Bson", "MongoDB.Bson.Serialization.BsonSerializationArgs, MongoDB.Bson", ClrNames.Object])]
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/MongoDb/BsonSerialization/IBsonSerializerProxy.cs
```diff
@@ -4,9 +4,15 @@
 // </copyright>
 
 #nullable enable
+using Datadog.Trace.DuckTyping;
+
 namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.MongoDb.BsonSerialization;
 
+/// <summary>
+/// Proxy for https://github.com/mongodb/mongo-csharp-driver/blob/4027d482d14960364c14be1ca59f7f6e350042a3/src/MongoDB.Bson/Serialization/IBsonSerializer.cs#L23
```
bouwkast, line 12: Thanks!

bouwkast (approved):

tracer/test/snapshots/MongoDbTests.SubmitsTraces_packageVersion=2_2.SchemaV0.verified.txt
```diff
@@ -89,6 +89,7 @@
       component: MongoDb,
       db.name: test-db,
       env: integration_tests,
+      language: dotnet,
```
lucaspimentel, line 92: Why did `language: dotnet` start showing up? Is that expected?

lucaspimentel (approved):

tracer/test/snapshots/MongoDbTests.SubmitsTraces_packageVersion=2_2.SchemaV0.verified.txt
andrewlock: Yeah, basically these snapshots are out of date from when we started adding this to every span. These snapshots are only used when we do the "comprehensive testing" (i.e. hitting all package versions). So this is entirely incidental

PR #7445, opened 2025-08-28
Support `final="true"` in NLog Logging rules for direct log submission

andrewlock:
## Summary of changes

Adds support for direct-logs injection for logging rules that have `final="true"` set

## Reason for change

A recent escalation flagged that NLog configuration for rules where `final="true"` were not being sent via direct log submission. This is due to how NLog processes rules - specifically, it processes rules in order, sending the logs to the specified targets, and _stops_ processing if it finds `final="true"`.

Previously we were adding our logging rule to the end of the list, which meant that it was never getting processed if a customer rule set `final="true"`. We change that to insert our rule at the start of the list of rules (and make sure _our_ rule isn't marked as final)

## Implementation details

This was generally a bit of a pain, because it's only "easy" to add a rule at the _end_ of the of the rules list, so we have to add it there, and manipulate it into the right place. It's a bit hacky, but as it should only happen once in general, it's hopefully not a big deal

## Test coverage

- Marked all the existing "manual" rules as `final="true"` - confirmed that this caused all the tests to fail
- Add support for inserting the rule at the start of the list, and confirm that the tests now pass.

I will run an "all package versions" attempt separately

## Other details

https://datadoghq.atlassian.net/browse/APMS-16994


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Logging/NLog/DirectSubmission/NLogCommon.cs
```diff
@@ -287,12 +308,62 @@ internal static bool AddDatadogTargetNLog43To45(object loggingConfiguration, obj
 
             // need to add the new target to the logging configuraiton
             loggingConfigurationProxy.AddTarget(NLogConstants.DatadogTargetName, targetProxy);
+            // This adds the rule at the end, but we need to add it to the start instead
             loggingConfigurationProxy.AddRuleForAllLevels(targetProxy, "**");
 
+            if (loggingConfigurationProxy.LoggingRules is { } rules)
```
bouwkast, line 314: I'm assuming that `LoggingRules` here isn't a List so we can't reuse `ReorderLoggingRules`?

bouwkast: Ah or maybe it doesn't `lock`?

andrewlock: > Ah or maybe it doesn't lock?

Exactly, that's the difference. That said, _technically_ we _could_ `lock` here, it's just not "necessary" because nothing in NLog is doing it... I'm not sure whether it's better to reduce the duplication and just lock, or to leave it as-is given the locking is kinda bad practice anyway 😅 WDYT?

bouwkast: Leaving it as is is fine IMO

PR #7446, opened 2025-08-28
Update Managed DevOps pools images

andrewlock:
## Summary of changes

Updates the VM images to be based on Ubuntu 24.04 instead of 20.04

## Reason for change

We've been plagued by `##[error]The remote provider was unable to process the request.` errors which cause jobs to just never start. After a call with Microsoft, they suggested the issue was related to using Ubuntu 20.04 as our base image for the VMs. Updating the images to test it out.

## Implementation details

Rebuild the Linux VMs to use the new base images. The bulk of the implementation happens in dockerfiles so hopefully there won't be any issues

## Test coverage

This is the test

### DIFF
--- a/.azure-pipelines/ultimate-pipeline.yml
+++ b/.azure-pipelines/ultimate-pipeline.yml
@@ -334,7 +334,7 @@ stages:
           useNativeSdkVersion: ""
           artifactSuffix: linux-musl-x64
     pool:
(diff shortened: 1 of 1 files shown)


bouwkast (approved):
Thanks!

lucaspimentel (approved):

PR #7447, opened 2025-08-28
Extract the Managed DevOps pools names to variables

andrewlock:
## Summary of changes

Extract the pool names to variables

## Reason for change

Reduces duplication, makes it easier to update them

## Implementation details

Find and replace

## Test coverage

This is the test

### DIFF
--- a/.azure-pipelines/ultimate-pipeline.yml
+++ b/.azure-pipelines/ultimate-pipeline.yml
@@ -61,6 +61,13 @@ variables:
   # After GA we can do a find and replace and get rid of this
   dotnetCoreSdkLatestVersionShort: 10.0.100-preview.7
   nativeBuildDotnetSdkVersion: 7.0.306
+  # These are the Managed DevOps pool names we use
+  linuxTasksPool: azure-managed-linux-tasks
+  linuxX64SmokePool: azure-managed-linux-smoke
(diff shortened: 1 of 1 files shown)


.azure-pipelines/ultimate-pipeline.yml
```diff
@@ -2965,7 +2972,7 @@ stages:
     timeoutInMinutes: 60 #default value
 
     pool:
-      name: azure-managed-linux-x64-1
```
bouwkast, line 2968: Guessing that this one was missed in https://github.com/DataDog/dd-trace-dotnet/pull/7446?

bouwkast (approved):
👍

.azure-pipelines/ultimate-pipeline.yml
bouwkast: Just to make sure it isn't different for a specific reason the others were on `-2` and being changed to `-1`, this one is on `-1` and staying on `-1`

lucaspimentel (approved):

.azure-pipelines/ultimate-pipeline.yml
andrewlock: Yup, looks like it - I think it was something that got out of sync when we added this stage and updated the pools seperately. This PR should fix it luckily 🙂

PR #7448, opened 2025-08-28
Add a code fix provider for invalid DuckType null checks

bouwkast:
## Summary of changes

This introduces a new Roslyn analyzer to flag instances where we are directly checking if an `IDuckType` is or isn't `null`.
In most cases this is incorrect and the correct equality check would be to check if the `IDuckType.Instance` is or isn't `null`.

The analyzer flags all instances of the former as a new rule called `DDDUCK001` 🦆 .

In addition to the analyzer, I've added a code fix provider that rewrites the flagged expression to use `?.Instance` instead. This allows us to correct the issue _and_ still check for `null` on the base `IDuckType` just to be safe.

**It is important to note though that not every instance of this detection is 100% accurate - there are and will be cases where checking if the `IDuckType` is `null` is the correct thing to do.**

However, using the conditional accessor should still be safe, but for transparency I've added some current cases to a `pramga` and to `tracer/GlobalSuppressions.cs`.

To ensure that we do not potentially disrupt developer workflows drastically with this the analyzer is flagged as a Warning, it will be bumped to an Error at a later point.

## Reason for change

Help enforce us to not accidentally check the IDuckType for `null` and instead check the `Instance` for `null`.

Using a null conditional (`?.`) also means that we are _still_ checking if the `IDuckType` is `null` on the off chance that it is.

## Implementation details

The analyzer is based off of our other analyzers (somewhat) and largely was based on stepping through the code in the tests in debug mode (works well in VS).

It will match on the following expressions: `==`, `!=`, `is`, `is not`.
This is only when the left and right hand sides implement an `IDuckType` and the other side is the `null` keyword.

I went through each of the current violations of the analyzer and either applied the code fix in VS automatically or ignored them on a case-by-case basis.

## Test coverage

- Tests have been added for the following usage patterns (code fix and analyzer included)
  - binary and `is` patterns: `duckType != null` / `duckType is not null`
  - generic constraints: where T : IDuckType
  - `(object)duckType == null` -> `duckType?.Instance == null`

## Other details

There was some ChatGPT usage here around the parenthesis / casting and namespace exclusions as that really tripped me up.

### DIFF
--- a/tracer/src/Datadog.Trace.Tools.Analyzers/DuckTypeAnalyzer/DuckDiagnostics.cs
+++ b/tracer/src/Datadog.Trace.Tools.Analyzers/DuckTypeAnalyzer/DuckDiagnostics.cs
@@ -22,6 +22,6 @@ public class DuckDiagnostics
         title: "Checking IDuckType for null",
         messageFormat: "The IDuckType is almost always non-null, check the Instance for null to ensure we have access to the duck typed object",
         category: "CodeQuality",
-        defaultSeverity: DiagnosticSeverity.Error,
+        defaultSeverity: DiagnosticSeverity.Warning,
         isEnabledByDefault: true);
 }
--- a/tracer/test/Datadog.Trace.Tools.Analyzers.Tests/DuckTypeAnalyzers/DuckTypeNullCheckAnalyzerTests.cs
+++ b/tracer/test/Datadog.Trace.Tools.Analyzers.Tests/DuckTypeAnalyzers/DuckTypeNullCheckAnalyzerTests.cs
@@ -17,7 +17,7 @@ namespace Datadog.Trace.Tools.Analyzers.Tests.DuckTypeAnalyzers;
 
 public class DuckTypeNullCheckAnalyzerTests
 {
-    private const DiagnosticSeverity Severity = DiagnosticSeverity.Error;
+    private const DiagnosticSeverity Severity = DiagnosticSeverity.Warning;
     private const string DiagnosticId = DuckDiagnostics.DuckTypeNullCheckDiagnosticId;
 
     private static string DuckTypeDefinitions { get; } = """


bouwkast:
Build failures are due to the detected errors, some I already have PRs for but will aim to address them before merging this

tracer/src/Datadog.Trace.Tools.Analyzers/DuckTypeAnalyzer/DuckDiagnostics.cs
```diff
@@ -0,0 +1,28 @@
...
+    public const string DuckTypeNullCheckDiagnosticId = "DDDUCK001";
+
+    internal static readonly DiagnosticDescriptor ADuckIsNeverNullRule = new(
+        DuckTypeNullCheckDiagnosticId,
+        title: "Checking IDuckType for null",
+        messageFormat: "{0}",
+        category: "CodeQuality",
+        defaultSeverity: DiagnosticSeverity.Error,
+        isEnabledByDefault: true,
+        description: "The IDuckType is almost always never null, check the Instance for null to ensure we have access to the ducked object.");
```
andrewlock: "almost always never" 😅 

```suggestion
        description: "The IDuckType is almost always non-null, check the Instance for null to ensure we have access to the duck typed object.");
```

tracer/src/Datadog.Trace.Tools.Analyzers/DuckTypeAnalyzer/DuckTypeNullCheckAnalyzer.cs
```diff
@@ -0,0 +1,275 @@
...
+
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Datadog.Trace.Tools.Analyzers.DuckTypeAnalyzer
+{
+    /// <summary>
+    /// Checks fo r null checks against IDuckType instances.
```
andrewlock:
```suggestion
    /// Checks for null checks against IDuckType instances.
```

tracer/src/Datadog.Trace.Tools.Analyzers/DuckTypeAnalyzer/DuckTypeNullCheckAnalyzer.cs
```diff
@@ -0,0 +1,275 @@
...
+            var rightIsNull = IsNullLiteral(bin.RightOperand);
+            if (!leftIsNull && !rightIsNull)
+            {
+                return;
+            }
+
+            // Look at the non-null side and unwrap casts/boxing to object/dynamic
+            // candidate here is: ConversionOperation Type: object
+            // it is an implicit cast / box to object (unlike the `is null` pattern)
+            // so we need to undo that before we can check if it is a IDuckType
```
andrewlock: This is very interesting, were you aware of this @tonyredondo? If so, we should probably change the analyzer to make you never use `==` either and always use `is`😄

tracer/src/Datadog.Trace.Tools.Analyzers/DuckTypeAnalyzer/DuckTypeNullCheckAnalyzer.cs
```diff
@@ -0,0 +1,275 @@
...
+                    if (ImplementsDuckType(c, duckType))
+                    {
+                        return true;
+                    }
+                }
+
+                return false;
+            }
+
+            // NOTE: IDuckType? == IDuckType which surprised me
```
andrewlock: I think that's because `IDuckType` is effectively a reference type by default, so the `?` is an annotation on that rather than making the struct nullable.

That's why this is bad (it has to box to an object):
```c#
public void MyMethod(IDuckType duck);
```

And this is good (no boxing)

```c#
public void MyMethod<T>(T duck);
    where T : IDuckType
```

tracer/test/Datadog.Trace.Tools.Analyzers.Tests/DuckTypeAnalyzers/DuckTypeNullCheckAnalyzerTests.cs
```diff
@@ -0,0 +1,292 @@
...
+        var fixedSource = $$"""
+        using Datadog.Trace.DuckTyping;
+
+        {{DuckTypeDefinitions}}
+
+        class TestClass
+        {
+            void TestMethod(IDuckType? duckType)
+            {
+                if (duckType.Instance {{operand}} null)
```
andrewlock: Given the incoming type is nullable, it's boxed, so I _think_ this actually needs to be:

```
if (duckType?.Instance {{operand}} null)
```

tracer/test/Datadog.Trace.Tools.Analyzers.Tests/DuckTypeAnalyzers/DuckTypeNullCheckAnalyzerTests.cs
```diff
@@ -0,0 +1,292 @@
...
+                    var bar = returnValue.Bar;
+                }
+        
+                return returnValue; // Dummy return
+            }
+        }
+        """;
+
+        // NOTE: The source fixedSource currently only replaces the condition, it does not remove the unnecessary nullability in the constraint.
+        // NOTE: Leaving that as an exercise for the reader
```
andrewlock: _Technically_ the nullability _isn't_ unnecessary (I think), so that's good 😄

tracer/src/Datadog.Trace.Tools.Analyzers/DuckTypeAnalyzer/DuckTypeNullCheckAnalyzer.cs
bouwkast: One thing I'm not sure of is whether this is just a Roslyn thing

tracer/src/Datadog.Trace.Tools.Analyzers/DuckTypeAnalyzer/DuckTypeNullCheckAnalyzer.cs
bouwkast: Ohhh very interesting thanks!

tracer/test/Datadog.Trace.Tools.Analyzers.Tests/DuckTypeAnalyzers/DuckTypeNullCheckAnalyzerTests.cs
bouwkast: Yeah that's correct but I don't know how to do that 😆 

I can poke around a bit more to see how to make it do that

tracer/test/Datadog.Trace.Tools.Analyzers.Tests/DuckTypeAnalyzers/DuckTypeNullCheckAnalyzerTests.cs
bouwkast: Oh interesting okay, so I really should get that supported

tracer/src/Datadog.Trace.Tools.Analyzers/DuckTypeAnalyzer/DuckDiagnostics.cs
bouwkast: I spent an embarrassingly amount of time coming up with how to phrase this 🪦

bouwkast: Applied in 3d3d344a6

Forgot that the `description` is not the "Description" the `messageFormat` is the "Description" and the `description` is just optional 😖

tracer/test/Datadog.Trace.Tools.Analyzers.Tests/DuckTypeAnalyzers/DuckTypeNullCheckAnalyzerTests.cs
bouwkast: Hmm yes, this causes Visual Studio to not allow the the code fix to be applied, so I'll work on fixing it.

tracer/src/Datadog.Trace.Tools.Analyzers/DuckTypeAnalyzer/DuckTypeNullCheckAnalyzer.cs
bouwkast: Okay I think that this is purely when you are in Visual Studio in the code file itself the syntax tree changes, I have reworked this to hopefully simplify it, but it is still a case to handle (and only affects the `==` and `!=`)

(still need to push the change(s) up though)

tracer/test/Datadog.Trace.ClrProfiler.Managed.Tests/AutoInstrumentation/AWS/Shared/SpanPointersTests.cs
```diff
@@ -116,6 +116,7 @@ public void AddDynamoDbSpanPointer_ShouldAddCorrectSpanLink()
         var stringAttr = new Mock<IDynamoDbAttributeValue>();
         stringAttr.Setup(a => a.S).Returns("testvalue");
         var keys = new Mock<IDynamoDbKeysObject>();
+        keys.Setup(k => k.Instance).Returns(new object()); // not important of what it returns as long as it isn't null
```
bouwkast, line 119: Done due to the new (correct) check if the keys.Instance is null

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Shared/SpanPointers.cs

bouwkast:
@codex review

bouwkast:
@codex review

bouwkast:
@codex review

PR #7449, opened 2025-08-29
Make sure we generate a build-id using SHA-1

gleocadie:
## Summary of changes

Make sure build-id for native libraries are generated using SHA-1

## Reason for change

The deobfuscation-api requires build-id to be generated by `SHA-1`. For some reason, build-ids for tracer and profiler were generated using SHA-1 but build-ids for universal binaries look like MD5 hash.

## Implementation details

In CMake files use `-Wl,--build-id=sha1`

## Test coverage

Downloaded the binaries and checked that the build-id correct:
for x86_64
image
image

for aarch64
image


andrewlock (approved):
🙈

PR #7451, opened 2025-08-29
[IAST] Delay dataflow creation

daniel-romano-DD:
## Summary of changes
Delay dataflow instance creation until activation is received from managed side

### DIFF
--- a/tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
+++ b/tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
@@ -179,25 +179,16 @@ Dataflow::Dataflow(ICorProfilerInfo* profiler, std::shared_ptr<RejitHandler> rej
         trace::Logger::Error("Dataflow::Dataflow -> Something very wrong happened, as QI on ICorProfilerInfo3 failed. Disabling Dataflow. HRESULT : ", Hex(hr));
     }
 
-    _initialized = (_profiler != nullptr);
     _aspectsLoaded = false;
 }
+
 Dataflow::~Dataflow()
 {
-    Destroy();
-}
-
-void Dataflow::Destroy()
-{
-    if (_initialized)
-    {
-        _initialized = false;
-        _aspectsLoaded = false;
-        REL(_profiler);
-        DEL_MAP_VALUES(_modules);
-        DEL_MAP_VALUES(_appDomains);
-        DEL_MAP_VALUES(_moduleAspects);
-    }
+    _aspectsLoaded = false;
+    REL(_profiler);
+    DEL_MAP_VALUES(_modules);
+    DEL_MAP_VALUES(_appDomains);
+    DEL_MAP_VALUES(_moduleAspects);
 }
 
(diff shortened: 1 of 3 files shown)


anna-git (approved):
thank you!

tracer/src/Datadog.Tracer.Native/rejit_handler.cpp
```diff
@@ -325,11 +325,13 @@ void RejitHandler::EnqueueRequestRejit(std::vector<MethodIdentifier>& rejitReque
 RejitHandler::RejitHandler(ICorProfilerInfo7* pInfo, std::shared_ptr<RejitWorkOffloader> work_offloader) :
     m_profilerInfo(pInfo), m_profilerInfo10(nullptr), m_work_offloader(work_offloader)
 {
+    m_rejitters.reserve(5);
```
andrewlock, line 328: where is this `5` coming from?

dromanol: There are 3 possible rejitters. 5 is to future proof this and to ensure there's no resizing, which was the cause of the crashes

dromanol: I could add a comment

andrewlock: > There are 3 possible rejitters. 5 is to future proof this and to ensure there's no resizing, which was the cause of the crashes

Yeah, I would be included to set this to the expected value of 3

> I could add a comment

Yeah, definitely 👍 I would suggest adding a comment here, something like this:
```suggestion
    // We need to set this to the maximum number of instances that we will register during the
    // lifetime of the 
    m_rejitters.reserve(3);
```

I would also suggest we add an explicit check in `RegisterRejitter` to make sure we never try to grow it.

In fact 🤔  Given that `vector` is _designed_ to grow (and to hide the underlying array), maybe we should actually switch this to _explicitly_ being an array? Then there's no risk of "accidentally" growing it and causing a crash (you always would have to check the length, but that seems fine)

I'm also fine with us doing this in a follow up PR (straight away 🙂) - the fact that the vector resize was the cause of crashes is well worth explicitly calling out in an explicit PR IMO

tonyredondo: +1 for the comment and the array idea.

tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
```diff
@@ -179,32 +179,29 @@ Dataflow::Dataflow(ICorProfilerInfo* profiler, std::shared_ptr<RejitHandler> rej
...
 void Dataflow::LoadAspects(WCHAR** aspects, int aspectsLength, UINT32 enabledCategories, UINT32 platform)
 {
     // Init aspects
     DBG("Dataflow::LoadAspects -> Processing aspects... ", aspectsLength, " Enabled categories: ", enabledCategories, " Platform: ", platform);
 
+    if (aspectsLength > 10)
+    {
+        _aspectClasses.reserve(aspectsLength / 10);
+        _aspects.reserve(aspectsLength);
+    }
```
tonyredondo, line 203: a comment here also would be nice.

PR #7453, opened 2025-08-29
Use P/Invoke to call  `lstat` instead of shelling out to `stat` to retrieve the inode

andrewlock:
## Summary of changes

Adds a P/Invoke path for retrieving the inode on linux and only later falls back to using `stat -c %i <path>`

## Reason for change

Calling `stat` will fail if `stat` isn't available (as in chiseled containers, for example). It also results in additional spans being created when we're running in a version-mismatch scenario(because the `Process` invocation is traced)

## Implementation details

Based my initial PoC on P/Invoking into `System.Native`, which is a shim shipped with recent .NET that  handles a bunch of marshalling issues etc and returns [this type](https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/Interop/Unix/System.Native/Interop.Stat.cs). That's [how recent .NET versions](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/IO/FileStatus.Unix.cs#L498) handle various bits like this.

However, System.Native only gives the correct inode value in .NET 7+ (from my local testing) and isn't available in early .NET Core. I then tried the "raw" route, trying to P/Invoke directly into `libc`. I also investigated making the syscall directly, but libc papers over differences in old kernels etc, so it seemed the easiest route. However, older versions of `libc` don't export the `lstat` or `fstatat` symbols as public symbols, and instead they're just macros.

Finally settled on a similar approach to System.Native, i.e. we P/Invoke into Datadog.Trace.Native which then makes the `lstat` call.

## Test coverage

Should _mostly_ be covered by existing tests (and smoke tests). I added an additional unit test to make sure the `stat -c %i` call and the P/Invoke give the same value, otherwise we should be covered. 

I also rebased on top of [this branch](https://github.com/DataDog/dd-trace-dotnet/pull/7446) which is where the error appeared - and did a [full-installer run](https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=186579&view=results), and it fixed all the issues, so we should be good 🤞 And we still have the existing fallback if not

Note that I explicitly disabled the process spans in the version-conflict tests, as whether we get the spans or not is dependent on the underlying host system, which is a bit gross.

## Other details

~~AI assisted in various places, including generating the structs.~~ None of that worked in the end. Great job AI.

I made the failure an error as it _shouldn't_ fail in normal execution, and we'd like to report if it does, however there are some cases where the P/Invoke calls aren't re-written so we know this won't work. In case of failure, we explicitly check for that, and log the error as debug instead.

To get everything to compile easily I split the `ContainerMetadata` implementation for .NET FX as it's not supported in that scenario anyway.

Also, to be able to unit test the P/Invoke work, I copied the native libs into the test directories.

Finally, it's worth us looking into [this method](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.nativelibrary.setdllimportresolver?view=net-9.0) to allow P/Invokes to happen in the native runner, given we know where the native libs are, they're just not re-written. Something to think about in a separate PR I think

### DIFF
--- a/tracer/build/_build/Build.Steps.cs
+++ b/tracer/build/_build/Build.Steps.cs
@@ -680,18 +680,27 @@ partial class Build
 
                 var testBinFolder = testDir / "bin" / BuildConfiguration;
 
-                var (ext, source) = Platform switch
+                var (ext, source, libdatadog) = Platform switch
                 {
-                    PlatformFamily.Windows => ("dll", MonitoringHomeDirectory / $"win-{TargetPlatform}" / "datadog_profiling_ffi.dll"),
-                    PlatformFamily.Linux => ("so", MonitoringHomeDirectory / GetUnixArchitectureAndExtension().Arch / "libdatadog_profiling.so"),
-                    PlatformFamily.OSX => ("dylib", MonitoringHomeDirectory / "osx" / $"libdatadog_profiling.dylib"),
+                    PlatformFamily.Windows => ("dll", MonitoringHomeDirectory / $"win-{TargetPlatform}", "datadog_profiling_ffi.dll"),
+                    PlatformFamily.Linux => ("so", MonitoringHomeDirectory / GetUnixArchitectureAndExtension().Arch, "libdatadog_profiling.so"),
+                    PlatformFamily.OSX => ("dylib", MonitoringHomeDirectory / "osx", "libdatadog_profiling.dylib"),
                     _ => throw new NotSupportedException($"Unsupported platform: {Platform}")
+                };
+
+                var libs = new[]
+                {
+                    (libdatadog, $"LibDatadog.{ext}"),
+                    ($"Datadog.Tracer.Native.{ext}", $"Datadog.Tracer.Native.{ext}"),
                 };
 
                 foreach (var framework in frameworks)
                 {
-                    var dest = testBinFolder / framework / $"LibDatadog.{ext}";
-                    CopyFile(source, dest, FileExistsPolicy.Overwrite);
+                    foreach (var lib in libs)
+                    {
+                        var dest = testBinFolder / framework / lib.Item2;
+                        CopyFile(source / lib.Item1, dest, FileExistsPolicy.Overwrite);
+                    }
                 }
             }
         });
--- a/tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
+++ b/tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
@@ -5,8 +5,10 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Linq;
+using System.Runtime.InteropServices;
 using System.Text.RegularExpressions;
 using System.Threading;
 using Datadog.Trace.Logging;
@@ -117,8 +119,8 @@ namespace Datadog.Trace.PlatformHelpers
             {
                 var tuple = ParseControllerAndPathFromCgroupLine(line);
                 if (tuple is not null
-                    && !string.IsNullOrEmpty(tuple.Item2)
-                    && (tuple.Item1 == string.Empty || string.Equals(tuple.Item1, "memory", StringComparison.OrdinalIgnoreCase)))
+                 && !string.IsNullOrEmpty(tuple.Item2)
+                 && (tuple.Item1 == string.Empty || string.Equals(tuple.Item1, "memory", StringComparison.OrdinalIgnoreCase)))
                 {
                     string controller = tuple.Item1;
                     string cgroupNodePath = tuple.Item2;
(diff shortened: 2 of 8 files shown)


tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
```diff
@@ -221,5 +246,109 @@ private static bool IsHostCgroupNamespaceInternal()
...
+        [SuppressMessage("StyleCop.CSharp.NamingRules", "SA1307:Accessible fields should begin with upper-case letter", Justification = "Keeping the names matching the spec definitions")]
+        private static class NativeInterop
+        {
+            /// <summary>
+            /// Get the inode for the provided path. Does so portably on x64 and arm64 Linux only.
+            /// </summary>
+            public static bool GetInode(string path, out long inode)
+            {
+                var arch = RuntimeInformation.ProcessArchitecture;
+                // using AT_FDCWD and AT_SYMLINK_NOFOLLOW for lstat-like behaviour
```
zacharycmontoya: Was there a particular reason you didn't use lstat if that is the behavior you're trying to achieve?

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
```diff
@@ -221,5 +246,109 @@ private static bool IsHostCgroupNamespaceInternal()
...
+            [StructLayout(LayoutKind.Sequential)]
+            private struct Timespec
+            {
+                public nint tv_sec; // time_t (64-bit on 64-bit Linux)
+                public nint tv_nsec; // long
+            }
+
+            // x86_64 (glibc & musl)
+            [StructLayout(LayoutKind.Sequential, Pack = 8)]
+            private struct StatX64
```
zacharycmontoya: Maybe this doesn't matter for this use-case since we only care about `st_ino`, but I noticed that the order of the fields do not match between x86_64 and aarch64/arm64, is that intentional? For instance, in x86_64 you have `st_ino` followed by `st_nlink` but in aarch64/arm64 you have `st_ino` followed by `st_mode`

lucaspimentel (approved):

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
andrewlock: Yeah, it was because `lstat` isn't actually exported as a public symbol in some early versions of glibc, and is rather just a macro, so it's not possible to P/Invoke it directly. Unfortunately, it seems that's _also_ true of `fstatat`, so that doesn't work either 😅 I'm trying a different approach in which we just use the native tracer for interop instead... 🤞

tracer/build/_build/Build.Steps.cs
```diff
@@ -680,18 +680,27 @@ async Task DownloadWafVersion(string libddwafVersion = null, string uncompressFo
...
+                    PlatformFamily.Windows => ("dll", MonitoringHomeDirectory / $"win-{TargetPlatform}", "datadog_profiling_ffi.dll"),
+                    PlatformFamily.Linux => ("so", MonitoringHomeDirectory / GetUnixArchitectureAndExtension().Arch, "libdatadog_profiling.so"),
+                    PlatformFamily.OSX => ("dylib", MonitoringHomeDirectory / "osx", "libdatadog_profiling.dylib"),
                     _ => throw new NotSupportedException($"Unsupported platform: {Platform}")
                 };
 
+                var libs = new[]
+                {
+                    (libdatadog, $"LibDatadog.{ext}"),
+                    ("Datadog.Trace.Native.dll", $"Datadog.Trace.Native.{ext}"),
```
andrewlock:
```suggestion
                    ("Datadog.Tracer.Native.dll", $"Datadog.Tracer.Native.{ext}"),
```

tracer/build/_build/Build.Steps.cs
```diff
@@ -680,18 +680,27 @@ async Task DownloadWafVersion(string libddwafVersion = null, string uncompressFo
...
+                    PlatformFamily.Windows => ("dll", MonitoringHomeDirectory / $"win-{TargetPlatform}", "datadog_profiling_ffi.dll"),
+                    PlatformFamily.Linux => ("so", MonitoringHomeDirectory / GetUnixArchitectureAndExtension().Arch, "libdatadog_profiling.so"),
+                    PlatformFamily.OSX => ("dylib", MonitoringHomeDirectory / "osx", "libdatadog_profiling.dylib"),
                     _ => throw new NotSupportedException($"Unsupported platform: {Platform}")
                 };
 
+                var libs = new[]
+                {
+                    (libdatadog, $"LibDatadog.{ext}"),
+                    ("Datadog.Tracer.Native.dll", $"Datadog.Tracer.Native.{ext}"),
```
andrewlock:
```suggestion
                    ($"Datadog.Tracer.Native.{ext}", $"Datadog.Tracer.Native.{ext}"),
```

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
andrewlock: not an issue now that we're calling it directly from C++ 🙂

PR #6834, opened 2025-04-09
Protect against more `NullReferenceExceptions` in `Kafka`

bouwkast:
## Summary of changes

This adds more null checking in to the the Kafka integrations that I spotted popping up in Error Tracking recently:
- `KafkaProduceAsyncIntegration`
- `KafkaProducerConstructorIntegration`

There weren't many of these exceptions being thrown by us, but when I looked at the integration code I noticed that it didn't have as much null-checking as I think we should.

The first was easy to reproduce, I couldn't reproduce the second one.

_Most likely the rest of the integration code should get a once over, but only address these as these are what we know are getting hit._

## Reason for change

Integration code was throwing `NullReferenceExceptions`

## Implementation details

More protections against `null`

## Test coverage

Added two calls that reproduced the null reference in `KafkaProduceAsyncIntegration`
I couldn't reproduce the ctor one though.

PR #7454, opened 2025-08-29
Fix all instances of DDDUCK001 (invalid `null` check for `IDuckType`)

bouwkast:
## Summary of changes

This fixes all instances of `DDDUCK001` where we are checking if a `IDuckType` equals `null` when we should likely be checking whether `IDuckType.Instance` equals `null`.

## Reason for change

Usually, but not always, the `IDuckType` is 
If the `Instance` is null we'd end up throwing a null reference exception when if/when we called a property / method.

## Implementation details

Applied the code fix.

## Test coverage

Existing tests

## Other details

Seems that with `IDuckType?` the code fix doesn't work in VS because I don't yet support that. Known issue, so I'll work on that part of the code fixer.


bouwkast:
Sortof a super set

I have two other old PRs I need to get merged that supersede this one

https://github.com/DataDog/dd-trace-dotnet/pull/6834
https://github.com/DataDog/dd-trace-dotnet/pull/7365

I had to re-arrange code in some places due to `?` causing errors for potential dereferencing (mainly in `Kafka`), but I _think_ the functionality here should all be the same.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Testing/MsTestV2/TestMethodAttributeExecuteIntegration.cs
bouwkast, line 1: @tonyredondo I wasn't super sure if these test related ones are accurate I initially had them excluded from the analyzer but figured I'd add them in here

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -164,7 +164,7 @@ private static long GetMessageSize<T>(T message)
                 PathwayContext? pathwayContext = null;
 
                 // Try to extract propagated context from headers
-                if (message?.Headers is not null)
+                if (message.Instance is not null && message.Headers is not null)
```
andrewlock, line 167: The interesting thing here is that we're passing in an `IMessage`, which means this actually _could_ be null I think 🤔 At the very least, the `?` approach is safe

```suggestion
                if (message?.Instance is not null && message.Headers is not null)
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -218,16 +218,19 @@ private static long GetMessageSize<T>(T message)
                     tags.BootstrapServers = bootstrapServers;
                 }
 
-                if (message is not null && message.Timestamp.Type != 0)
+                if (message.Instance is not null)
```
andrewlock, line 221:
```suggestion
                if (message?.Instance is not null)
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -252,7 +255,7 @@ private static long GetMessageSize<T>(T message)
                         dataStreamsManager,
                         CheckpointKind.Consume,
                         edgeTags,
-                        message is null || dataStreamsManager.IsInDefaultState ? 0 : GetMessageSize(message),
+                        message.Instance is null || dataStreamsManager.IsInDefaultState ? 0 : GetMessageSize(message),
```
andrewlock, line 258:
```suggestion
                        message?.Instance is null || dataStreamsManager.IsInDefaultState ? 0 : GetMessageSize(message),
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Logging/NLog/DirectSubmission/LogFactoryBuildLoggerConfiguration.cs
```diff
@@ -45,7 +45,7 @@ internal static CallTargetState OnMethodBegin<TTarget, TLoggingRuleList>(TTarget
         // when logging is being stopped/shutdown this instrumentation gets called again
         // The logging rules passed are "null"
         // but we have an "_isDisposing" field that is set to true in the LogFactory that we can check
-        if (instance.TryDuckCast<ILogFactoryProxy>(out var logFactoryProxy) && logFactoryProxy.IsDisposing)
+        if (instance.TryDuckCast<ILogFactoryProxy>(out var logFactoryProxy) && logFactoryProxy.Instance is not null && logFactoryProxy.IsDisposing)
```
andrewlock, line 48: _technically_ the `Instance` check isn't necessary here - if `instance.TryDuckCast` returns `true`, then `instance` must be non-null, and therefore `logFactoryProxy.Instance` is non-null 😅

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Testing/MsTestV2/TestMethodRunnerExecuteIntegration.cs
```diff
@@ -54,7 +54,7 @@ public static class TestMethodRunnerExecuteIntegration
                 {
                     if (unitTestResult.Outcome is UnitTestResultOutcome.Inconclusive)
                     {
-                        if (instance.TestMethodInfo is not null)
+                        if (instance?.TestMethodInfo?.Instance is not null)
```
andrewlock, line 57: I believe this is an instance method so `instance` cannot be null

bouwkast:
Put everything in this one with a conditional access for safety

https://github.com/DataDog/dd-trace-dotnet/pull/7448

PR #6894, opened 2025-04-24
Steven/more combinatorial tests

bouwkast:
## Summary of changes

Attempts to speed up: `SerilogTests`, `GrpcTests`, and `RabbitMqTests`.

## Reason for change

- CI speed improvement

## Implementation details

- Swapped `SerilogTests` to `CombinatorialOrPairwiseData`
  - The skips will only be hit when running locally - unfortunately not a clean way to exclude
- Reduced sleeps in GRPC tests, these were causing them to take 1:30 to 3:00 minutes with most of the time in a sleep.
- Re-use RabbitMQ connections - each connection takes ~2 seconds to create, amounting to around 20 to 30 seconds of time just spent creating connections per `SubmitTraces` run, causing the entire suite to take ~5 minutes - now each `SubmitTraces` should be around 10s or so.
- 
## Other details

Outside of `NLogTests` https://github.com/DataDog/dd-trace-dotnet/pull/6871 I don't see many other quick wins for Combinatorial mode.

If the speed changes cause flake they can be reverted.

PR #7457, opened 2025-09-02
Fix GRPC instrumentation for BuildHttpErrorResponse. Enable integration tests.

NachoEchevarria:
## Summary of changes

[This error](https://app.datadoghq.com/error-tracking?query=source%3Adotnet%20-%2AStackOverflowException%2A%20-%2Adebugger%2A%20-%2Aappsec%2A&et-side=data&et-viz=events&order=total_count&refresh_mode=sliding&source=all&sp=%5B%7B%22p%22%3A%7B%22issueId%22%3A%225b4a9f2a-3ae6-11f0-9077-da7ad0900002%22%7D%2C%22i%22%3A%22error-tracking-issue%22%7D%5D&view=spans&from_ts=1756801478434&to_ts=1756887878434&live=true) was logged:

```
System.TypeInitializationException ---> Datadog.Trace.ClrProfiler.CallTarget.CallTargetInvokerException ---> System.ArgumentException
at Datadog.Trace.Util.ThrowHelper.ThrowArgumentException(String message)
at Datadog.Trace.ClrProfiler.CallTarget.Handlers.IntegrationMapper.CreateBeginMethodDelegate(Type integrationType, Type targetType, Type[] argumentsTypes)
at Datadog.Trace.ClrProfiler.CallTarget.Handlers.BeginMethodHandler`6..cctor()
```

This error would only happen when method Grpc.AspNetCore.Server.Internal.GrpcProtocolHelpers.BuildHttpErrorResponse is called, which happens when there is an error during the Grpc call.

In order to fix it, the signature of OnMethodBegin<TTarget, TGrpcStatusCode> in GrpcProtocolHelpersBuildHttpErrorResponseIntegration.cs had to be changed by adding generic types.

To test this change, a new case has been added to the GRPC sample application. By adding this header to the request (context.Request.Headers["content-type"] = "application/json";), we can test this instrumentation by causing an error during the GRPC call:

Content-Type 'application/json' is not supported

Additionally, some fixes have been done in the test app. In the following code, we were always throwing the exception because the delay set in theVerySlow method was 300 msecs, lower than 600, so the method would finish on time, which is not the expected outcome :
```
        try
        {
            _logger.LogInformation("Sending very slow request to self");
            await client.VerySlowAsync(new HelloRequest { Name = "GreeterClient" }, deadline: DateTime.UtcNow.AddMilliseconds(600));
            throw new Exception("Received reply, when should have exceeded deadline");
        }
```

By setting a lower deadline, we make sure that the deadline is exceeded. We were not detecting this unexpected behavior because in grpctests.cs, we were capturing the exception ExitCodeException:

```
            catch (ExitCodeException)
            {
                // There is a race condition in GRPC version < v2.43.0 that can cause ObjectDisposedException
                // when a deadline is exceeded. Skip the test if we hit it: https://github.com/grpc/grpc-dotnet/pull/1550
                if (!string.IsNullOrEmpty(packageVersion)
                 && new Version(packageVersion) < new Version("2.43.0")
                 && processResult is not null
                 && processResult.StandardError.Contains("ObjectDisposedException"))
                {
                    throw new SkipException("Hit race condition in GRPC deadline exceeded");
                }
            }
```

After capturing the exception, we would not launch an SkipException but we would end the test without errors. A new throw statement has been added to control that we only skip the tests in the case of ObjectDisposedException.

This change in the GRPC tests has also effectively enabled the GRPC legacy tests, that have a different instrumentation but were suffering the same issue (throw exception and skip). The snapshots of the legacy tests were not matching exactly the result obtained after enabling these tests. The instrumentation of the legacy methods or the legacy sample app have not changed significantly, so we would probably need to evaluate if the results that we are getting are the expected ones. Anyway, this work would probably be out of the scope of this PR.


tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/GrpcTests.cs
```diff
@@ -426,6 +424,8 @@ static void FixVerySlowClientSpans(IImmutableList<MockSpan> spans)
                 {
                     throw new SkipException("Hit race condition in GRPC deadline exceeded");
                 }
+
+                throw;
```
NachoEchevarria, line 428: This line will prevent test from passing if the return code is not correct

tracer/test/snapshots/GrpcLegacy.SubmitTraces_httpclient=False.SchemaV0.verified.txt
```diff
@@ -1291,39 +1291,5 @@ Grpc.Core.Internal.CoreErrorDetailException: {"created":"@00000000000.000000000"
       _dd.tracer_kr: 1.0,
       _sampling_priority_v1: 1.0
     }
-  },
```
NachoEchevarria, line 1294: The legacy tests were not being executed against the snapshots because of the skipException. While the GRCP legacy instrumentation has not changed, it seems that what we had in the snapshots is different from what we are getting, with all the error information in a single span in the case of deadline exceeded.

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/GrpcTests.cs
andrewlock: Good catch - I assume you're not seeing any other failures from this, right?

tracer/test/snapshots/GrpcLegacy.SubmitTraces_httpclient=False.SchemaV0.verified.txt
andrewlock: AFAICT, these _should_ still be produced, right 🤔 assuming we're still creating the deadline exceeded span. It looks like we broke this in https://github.com/DataDog/dd-trace-dotnet/pull/6894, but I don't understand why this span would now be missing 🤔

tracer/test/snapshots/GrpcDotNet.SubmitTraces_httpclient=True.SchemaV1.verified.txt
andrewlock, line 1: These LGTM - AFAICT the diff is just a bit rubbish, and all we're really seeing is 4 new spans from the `InvalidContentTypeMethod` test call

andrewlock (approved):
Good catch, thanks!

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/GrpcTests.cs
NachoEchevarria: Yes, tests are passing

tracer/test/test-applications/integrations/Samples.GrpcDotNet/Worker.cs
```diff
@@ -214,7 +215,8 @@ private async Task SendVerySlowRequestAsync(Greeter.GreeterClient client)
         try
         {
             _logger.LogInformation("Sending very slow request to self");
-            await client.VerySlowAsync(new HelloRequest { Name = "GreeterClient" }, deadline: DateTime.UtcNow.AddMilliseconds(600));
+            // The delay in the method is 200 miliseconds.
```
bouwkast, line 218: Ah nice thanks I think I did this oops

tracer/test/test-applications/integrations/Samples.GrpcDotNet/Worker.cs
```diff
@@ -214,7 +215,8 @@ private async Task SendVerySlowRequestAsync(Greeter.GreeterClient client)
...
+            await client.VerySlowAsync(new HelloRequest { Name = "GreeterClient" }, deadline: DateTime.UtcNow.AddMilliseconds(20));
```
bouwkast, line 219: Just for my own clarification, I thought we were expecting the server delay to be less than the client delay in these tests now that is flipped.
Is that correct?

Or was it incorrect previously?

andrewlock: Yeah, good point, something doesn't look quite right here 🤔 it's very confusing to be fair 😅 

I'm pretty sure that that the _client_ should be smaller than the _server_... otherwise there will be no deadline exceeded, so this _should_ now be correct...

However, the fact the snapshots are missing from the _legacy_ tests, suggests there's something wrong there to me 🤔

NachoEchevarria: Yes, we should have a smaller delay in the client so we get the exception. Otherwise, we would end the app with an exception error (that was previously ignored)

With the non legacy tests, everything works as expected. In the case of legacy, the snapshots that we had don't seem to match the results that we are getting now. It's weird because the legacy instrumentation has not changed in this PR. Still, we have not run these tests for quite some time because we were skipping them.

PR #7460, opened 2025-09-02
[Native] Convert Rejitters vector to a fixed array

daniel-romano-DD:
## Summary of changes
After converting the creation of Dataflow to dynamic, when adding it to the rejiters vector, this could cause a race condition when it grows. So, in order to avoid this, the rejitters vector has been converted to a fixed array.

As there are 3 known rejitters, ATM, the size of the array has been set to 4 to have a minimal buffer.

## Reason for change
Previous vector reserve implementation could be unsafe if a new rejitter was added without modifying the reserved size. This way if a new one is added without modifying the max count, an exception will be thrown

## Implementation details
Replace the previous rejitters vector with a fixed array.

## Test coverage
Unit tests have been added for the array insert function

### DIFF
--- a/tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
+++ b/tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
@@ -160,7 +160,7 @@ AspectFilter* ModuleAspects::GetFilter(DataflowAspectFilterValue filterValue)
 
 Dataflow::Dataflow(ICorProfilerInfo* profiler, std::shared_ptr<RejitHandler> rejitHandler,
                    const RuntimeInformation& runtimeInfo) :
-    Rejitter(rejitHandler, RejitterPriority::Low)
+    Rejitter(RejitterPriority::Low)
 {
     m_runtimeType = runtimeInfo.runtime_type;
     m_runtimeVersion = VersionInfo{runtimeInfo.major_version, runtimeInfo.minor_version, runtimeInfo.build_version, 0};
@@ -180,6 +180,7 @@ Dataflow::Dataflow(ICorProfilerInfo* profiler, std::shared_ptr<RejitHandler> rej
     }
 
     _initialized = false;
+    rejitHandler->RegisterRejitter(this);
 }
 
 Dataflow::~Dataflow()
--- a/tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
+++ b/tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
@@ -198,8 +198,8 @@ void Dataflow::LoadAspects(WCHAR** aspects, int aspectsLength, UINT32 enabledCat
 
     if (aspectsLength > 10)
     {
-        _aspectClasses.reserve(aspectsLength / 10);
-        _aspects.reserve(aspectsLength);
+        _aspects.reserve(aspectsLength); // We know the max number of aspects we are going to have, so reserve the space to avoid vector resizes
+        _aspectClasses.reserve(aspectsLength / 10); // We don't know exactly the number of aspects which are class aspects, but 1/10 is a fine approach
     }
 
     DataflowAspectClass* aspectClass = nullptr;
--- a/tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
+++ b/tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
@@ -160,7 +160,7 @@ AspectFilter* ModuleAspects::GetFilter(DataflowAspectFilterValue filterValue)
 
 Dataflow::Dataflow(ICorProfilerInfo* profiler, std::shared_ptr<RejitHandler> rejitHandler,
                    const RuntimeInformation& runtimeInfo) :
-    Rejitter(RejitterPriority::Low)
+    Rejitter(rejitHandler, RejitterPriority::Low, false)
 {
     m_runtimeType = runtimeInfo.runtime_type;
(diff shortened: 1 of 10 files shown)


tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
```diff
@@ -198,8 +198,8 @@ void Dataflow::LoadAspects(WCHAR** aspects, int aspectsLength, UINT32 enabledCat
 
     if (aspectsLength > 10)
     {
-        _aspectClasses.reserve(aspectsLength / 10);
-        _aspects.reserve(aspectsLength);
+        _aspects.reserve(aspectsLength); // We know the max number of aspects we are going to have, so reserve the space to avoid vector sesizes
```
andrewlock:
```suggestion
        _aspects.reserve(aspectsLength); // We know the max number of aspects we are going to have, so reserve the space to avoid vector resizes
```

tonyredondo (approved):

andrewlock (approved):
Thanks!

robertpi (approved):

tracer/test/Datadog.Tracer.Native.Tests/util_test.cpp
```diff
@@ -0,0 +1,86 @@
...
+// === Test cases ===
+TEST(UtilTests, ArrayInsert_Middle)
+{
+    int arr[10] = {1, 2, 3, 4, 5};
+    size_t count = 5;
+
+    Insert(arr, count, 2, 99);
+    int expected[] = {1, 2, 99, 3, 4, 5};
+
+    assert(count == 6);
```
gleocadie: why no use GTest asserts instead ? Here you could use ASSERT_EQ instead
https://github.com/google/googletest/blob/main/docs/reference/assertions.md

tracer/test/Datadog.Tracer.Native.Tests/util_test.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+TEST(UtilTests, ArrayInsert_Beginning)
+{
+    int arr[10] = {10, 20, 30};
+    size_t count = 3;
+
+    Insert(arr, count, 0, 5);
+    int expected[] = {5, 10, 20, 30};
+
+    assert(count == 4);
```
gleocadie: same

tracer/test/Datadog.Tracer.Native.Tests/util_test.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+TEST(UtilTests, ArrayInsert_End)
+{
+    int arr[10] = {7, 8, 9};
+    size_t count = 3;
+
+    Insert(arr, count, 3, 99);
+    int expected[] = {7, 8, 9, 99};
+
+    assert(count == 4);
```
gleocadie: same

tracer/test/Datadog.Tracer.Native.Tests/util_test.cpp
```diff
@@ -0,0 +1,86 @@
...
+    bool thrown = false;
+    try
+    {
+        Insert(arr, count, 5, 99); // invalid position
+    }
+    catch (const std::out_of_range&)
+    {
+        thrown = true;
+    }
+    assert(thrown);
```
gleocadie: here you can replace by 
`ASSERT_THROW` https://github.com/google/googletest/blob/main/docs/reference/assertions.md#expect_throw-expect_throw

tracer/test/Datadog.Tracer.Native.Tests/util_test.cpp
```diff
@@ -0,0 +1,86 @@
...
+    bool thrown = false;
+    try
+    {
+        Insert(arr, count, 1, 99); // array is already full
+    }
+    catch (const std::out_of_range&)
+    {
+        thrown = true;
+    }
+    assert(thrown);
```
gleocadie: same

tracer/test/Datadog.Tracer.Native.Tests/util_test.cpp
dromanol: You are right. Fixed in 10a464344a708684f9c34196a688821f323b9876

tracer/test/Datadog.Tracer.Native.Tests/util_test.cpp
dromanol: Fixed in 1b99ddfe76c2febca642c6b0836651ebd7c84cb6

tracer/src/Datadog.Tracer.Native/rejit_handler.h
```diff
@@ -112,7 +112,9 @@ class RejitHandler
     bool enable_by_ref_instrumentation = false;
     bool enable_calltarget_state_by_ref = false;
 
-    std::vector<Rejitter*> m_rejitters;
+    #define MAX_REJITTERS 4 // Increase this number when a new rejitter is added
+    Rejitter* m_rejitters[MAX_REJITTERS];
+    size_t m_rejittersCount = 0;
```
gleocadie, line 117: in a previous commit, this was a std::atomic<size_t>, why rolling back ?
if this can be accessed/modified by multiple threads at the same time, shouldn't be an atomic ?

gleocadie: discussed offline and it accessed by only one thread so 👍

PR #7462, opened 2025-09-03
Consolidate bailout checks into native loader

andrewlock:
## Summary of changes

Move some of the bail-out checks from the native loader into the native tracer

## Reason for change

As far as I can tell, these checks really _should_ be in the native loader, and they're only in the native tracer currently for historical reasons. This results in both duplication of work _and_ mismatch between traced processes for the profiler and the tracer. This was discovered as part of the work in #7287.

## Implementation details

- Removed duplicate checks (process name variables) from the native tracer
- Moved the additional CI Visibility checks to the native loader. I'll defer to @tonyredondo if we want to tidy these up at all
- Moved the AAS checks to the native loader. I'll defer to @lucaspimentel if we want to change any of these.

## Test coverage

This should be covered by the existing tests in general, but I may add an additional instrumentation test to make sure we get no tracer logs for previous cases where we would have done.

## Other details

Currently the native tracer prints a bunch of env vars for debugging purposes. We could move this to the native loader too if we prefer.

### DIFF
--- a/shared/src/Datadog.Trace.ClrProfiler.Native/EnvironmentVariables.h
+++ b/shared/src/Datadog.Trace.ClrProfiler.Native/EnvironmentVariables.h
@@ -23,8 +23,8 @@ public:
     // Sets whether the current process must run in CI Visibility mode or not.
     inline static const shared::WSTRING CiVisibilityEnabled = WStr("DD_CIVISIBILITY_ENABLED");
 
-    // Indicates whether the profiler is running in the context
-    // of Azure App Services with the extension installed
+    // Indicates whether the CLR profiler is running from
+    // the Azure App Services site extension
     inline static const shared::WSTRING IsAzureAppServicesExtension = WStr("DD_AZURE_APP_SERVICES");
 
     // The app_pool_id in the context of azure app services
@@ -33,7 +33,8 @@ public:
     // The DOTNET_CLI_TELEMETRY_PROFILE in the context of azure app services
     inline static const shared::WSTRING AzureAppServicesCliTelemetryProfilerValue = WStr("DOTNET_CLI_TELEMETRY_PROFILE");
 
-    // The FUNCTIONS_WORKER_RUNTIME in the context of azure app services
+    // The FUNCTIONS_WORKER_RUNTIME in Azure Functions.
+    // Valid values: "dotnet" (in-process functions) or "dotnet-isolated" (isolated functions).
     // Used as a flag to determine that we are running within a functions app.
     inline static const shared::WSTRING AzureAppServicesFunctionsWorkerRuntime = WStr("FUNCTIONS_WORKER_RUNTIME");
 
--- a/shared/src/Datadog.Trace.ClrProfiler.Native/EnvironmentVariables.h
+++ b/shared/src/Datadog.Trace.ClrProfiler.Native/EnvironmentVariables.h
@@ -19,4 +19,25 @@ public:
     inline static const shared::WSTRING SingleStepInstrumentationEnabled = WStr("DD_INJECTION_ENABLED");
     inline static const shared::WSTRING ForceEolInstrumentation = WStr("DD_INJECT_FORCE");
     inline static const shared::WSTRING SingleStepInstrumentationTelemetryForwarderPath = WStr("DD_TELEMETRY_FORWARDER_PATH");
+
+    // Sets whether the current process must run in CI Visibility mode or not.
+    inline static const shared::WSTRING CiVisibilityEnabled = WStr("DD_CIVISIBILITY_ENABLED");
+
+    // Indicates whether the profiler is running in the context
+    // of Azure App Services with the extension installed
+    inline static const shared::WSTRING IsAzureAppServicesExtension = WStr("DD_AZURE_APP_SERVICES");
+
+    // The app_pool_id in the context of azure app services
+    inline static const shared::WSTRING AzureAppServicesAppPoolId = WStr("APP_POOL_ID");
+
+    // The DOTNET_CLI_TELEMETRY_PROFILE in the context of azure app services
+    inline static const shared::WSTRING AzureAppServicesCliTelemetryProfilerValue = WStr("DOTNET_CLI_TELEMETRY_PROFILE");
+
+    // The FUNCTIONS_WORKER_RUNTIME in the context of azure app services
+    // Used as a flag to determine that we are running within a functions app.
+    inline static const shared::WSTRING AzureAppServicesFunctionsWorkerRuntime = WStr("FUNCTIONS_WORKER_RUNTIME");
+
+    // Determine whether to instrument within azure functions.
+    // Default is true.
+    inline static const shared::WSTRING AzureFunctionsInstrumentationEnabled = WStr("DD_TRACE_AZURE_FUNCTIONS_ENABLED");
 };
--- a/shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
+++ b/shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
@@ -250,10 +250,9 @@ namespace datadog::shared::nativeloader
         }
 
         // AAS checks
-        const auto isRunningInAas = TryParseBooleanEnvironmentValue(
-                                        GetEnvironmentValue(EnvironmentVariables::IsAzureAppServicesExtension),
-                                        isRunningInAas) && isRunningInAas;
-        if (isRunningInAas)
+        bool isRunningInAas;
+        if (TryParseBooleanEnvironmentValue(GetEnvironmentValue(EnvironmentVariables::IsAzureAppServicesExtension),
+                                            isRunningInAas) && isRunningInAas)
         {
             Log::Info("Azure App Services detected.");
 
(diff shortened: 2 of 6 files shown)


shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -228,6 +228,71 @@ namespace datadog::shared::nativeloader
...
+                    if (TryParseBooleanEnvironmentValue(GetEnvironmentValue(EnvironmentVariables::CiVisibilityEnabled), is_ci_visibility_enabled)
+                        && is_ci_visibility_enabled)
+                    {
+                        if (tokenized_command_line[1] != WStr("test") &&
+                            process_command_line.find(WStr("testhost")) == WSTRING::npos &&
+                            process_command_line.find(WStr("exec")) == WSTRING::npos &&
+                            process_command_line.find(WStr("datacollector")) == WSTRING::npos &&
+                            process_command_line.find(WStr("vstest.console.dll")) == WSTRING::npos)
+                        {
+                            Log::Info("The Tracer Profiler has been disabled because the process is running in CI Visibility "
```
lucaspimentel, line 243: "Tracer Profiler"? How about we use "CLR Profiler" everywhere? or "Datadog instrumentation library" or `dd-trace-dotnet` or really anything else 😝 

```suggestion
                            Log::Info("The CLR Profiler has been disabled because the process is running in CI Visibility "
```

lucaspimentel: Maybe we should avoid "profiler" entirely. I see it often causes confusion in support tickets. People think it means the Continuous Profiler.

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -228,6 +228,71 @@ namespace datadog::shared::nativeloader
...
+            }
+        }
+
+        // AAS checks
+        const auto isRunningInAas = TryParseBooleanEnvironmentValue(
+                                        GetEnvironmentValue(EnvironmentVariables::IsAzureAppServicesExtension),
+                                        isRunningInAas) && isRunningInAas;
+        if (isRunningInAas)
+        {
+            Log::Info("ClrProfiler is operating within Azure App Services context.");
```
lucaspimentel: So verbose 😅 
```suggestion
            Log::Info("Azure App Services detected.");
```

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -228,6 +228,71 @@ namespace datadog::shared::nativeloader
...
+
+            const auto& app_pool_id_value = GetEnvironmentValue(EnvironmentVariables::AzureAppServicesAppPoolId);
+
+            if (app_pool_id_value.size() > 1 && app_pool_id_value.at(0) == '~')
+            {
+                Log::Info(
+                    "DATADOG TRACER DIAGNOSTICS - ClrProfiler disabled: ", EnvironmentVariables::AzureAppServicesAppPoolId, " ",
+                    app_pool_id_value, " is recognized as an Azure App Services infrastructure process.");
```
lucaspimentel:
```suggestion
                    app_pool_id_value, " is an Azure App Services infrastructure process.");
```

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -228,6 +228,71 @@ namespace datadog::shared::nativeloader
...
+                return CORPROF_E_PROFILER_CANCEL_ACTIVATION;
+            }
+
+            const auto& cli_telemetry_profile_value =
+                GetEnvironmentValue(EnvironmentVariables::AzureAppServicesCliTelemetryProfilerValue);
+
+            if (cli_telemetry_profile_value == WStr("AzureKudu"))
+            {
+                Log::Info("DATADOG TRACER DIAGNOSTICS - ClrProfiler disabled: ", app_pool_id_value,
+                             " is recognized as Kudu, an Azure App Services reserved process.");
```
lucaspimentel:
```suggestion
                             " is Kudu, an Azure App Services reserved process.");
```

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -228,6 +228,71 @@ namespace datadog::shared::nativeloader
...
+                // enabled by default
+                const auto azure_functions_enabled = !TryParseBooleanEnvironmentValue(
+                                                         GetEnvironmentValue(
+                                                             EnvironmentVariables::AzureFunctionsInstrumentationEnabled),
+                                                         azure_functions_enabled)
+                                                     || azure_functions_enabled;
+
+                if (!azure_functions_enabled)
+                {
+                    Log::Info("DATADOG TRACER DIAGNOSTICS - ClrProfiler explicitly disabled for Azure Functions.");
```
lucaspimentel, line 291: TIL we had a env var for this.

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -228,6 +228,71 @@ namespace datadog::shared::nativeloader
...
+                                "mode, the name is 'dotnet' but the commandline doesn't contain 'testhost' or 'datacollector' or 'vstest.console.dll' or 'exec'");
+                            return CORPROF_E_PROFILER_CANCEL_ACTIVATION;
+                        }
+                    }
+                }
+            }
+        }
+
+        // AAS checks
+        const auto isRunningInAas = TryParseBooleanEnvironmentValue(
```
lucaspimentel: We should be careful with `DD_AZURE_APP_SERVICES`. It doesn't mean "we're running inside AAS," it means "we're instrumenting _with the site extension_."

We now have a nuget package that can be used instead of the site extension, and `DD_AZURE_APP_SERVICES` won't be set in that case.

shared/src/Datadog.Trace.ClrProfiler.Native/EnvironmentVariables.h
```diff
@@ -19,4 +19,25 @@ class EnvironmentVariables final
     inline static const shared::WSTRING SingleStepInstrumentationEnabled = WStr("DD_INJECTION_ENABLED");
     inline static const shared::WSTRING ForceEolInstrumentation = WStr("DD_INJECT_FORCE");
     inline static const shared::WSTRING SingleStepInstrumentationTelemetryForwarderPath = WStr("DD_TELEMETRY_FORWARDER_PATH");
+
+    // Sets whether the current process must run in CI Visibility mode or not.
+    inline static const shared::WSTRING CiVisibilityEnabled = WStr("DD_CIVISIBILITY_ENABLED");
+
+    // Indicates whether the profiler is running in the context
+    // of Azure App Services with the extension installed
```
lucaspimentel:
```suggestion
    // Indicates whether the CLR profiler is running from
    // the Azure App Services site extension
```

shared/src/Datadog.Trace.ClrProfiler.Native/EnvironmentVariables.h
```diff
@@ -19,4 +19,25 @@ class EnvironmentVariables final
...
+    inline static const shared::WSTRING IsAzureAppServicesExtension = WStr("DD_AZURE_APP_SERVICES");
+
+    // The app_pool_id in the context of azure app services
+    inline static const shared::WSTRING AzureAppServicesAppPoolId = WStr("APP_POOL_ID");
+
+    // The DOTNET_CLI_TELEMETRY_PROFILE in the context of azure app services
+    inline static const shared::WSTRING AzureAppServicesCliTelemetryProfilerValue = WStr("DOTNET_CLI_TELEMETRY_PROFILE");
+
+    // The FUNCTIONS_WORKER_RUNTIME in the context of azure app services
```
lucaspimentel:
```suggestion
    // The FUNCTIONS_WORKER_RUNTIME in Azure Functions.
    // Valid values: "dotnet" (in-process functions) or "dotnet-isolated" (isolated functions).
```

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
lucaspimentel: In managed code, we now use `WEBSITE_SITE_NAME` to determine if we are running in Azure App Services.

lucaspimentel: (no action required in this PR, just an observation)

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
andrewlock: Yeah, this was a copy-paste, there's a bunch of these, all with slight variations. I think we should consolidate everywhere. But maybe in a separate PR, so that we have a chance to agree on the naming 😅

How about 
- `Datadog SDK` when we mean the whole product
- `Continuous profiler`
- ? for the tracer? 😅

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
andrewlock: Same 😅

chrisnas (approved):
LGTM
I like the idea of renaming "profiler" depending on native loader (CLR profiler), tracer and continuous profiler

tonyredondo (approved):

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
lucaspimentel: For posterity: discussed offline. See #7467.

PR #7464, opened 2025-09-04
chore(configuration): enable DataPipelineEnabled on AAS + Windows

ganeshnj:
## Reason for change

Due to limitation AAS extension https://github.com/DataDog/datadog-aas-extension/pull/388

TLDR; customer set env var have lower priority than the AAS configs.

### DIFF
--- a/tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
+++ b/tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
@@ -411,7 +411,7 @@ namespace Datadog.Trace.Configuration
 
             DataPipelineEnabled = config
                                   .WithKeys(ConfigurationKeys.TraceDataPipelineEnabled)
-                                  .AsBool(defaultValue: false);
+                                  .AsBool(defaultValue: EnvironmentHelpers.IsAzureAppServices() && FrameworkDescription.Instance.IsWindows());
 
             if (DataPipelineEnabled)
             {


tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -411,7 +411,7 @@ internal TracerSettings(IConfigurationSource? source, IConfigurationTelemetry te
 
             DataPipelineEnabled = config
                                   .WithKeys(ConfigurationKeys.TraceDataPipelineEnabled)
-                                  .AsBool(defaultValue: false);
+                                  .AsBool(defaultValue: EnvironmentHelpers.IsAzureAppServices() && FrameworkDescription.Instance.IsWindows());
```
andrewlock: I think we actually want to check if the extension is installed to scope this as previously discussed
```suggestion
                                  .AsBool(defaultValue: EnvironmentHelpers.IsUsingAzureAppServicesSiteExtension() && FrameworkDescription.Instance.IsWindows());
```

andrewlock (approved):

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -411,7 +411,7 @@ internal TracerSettings(IConfigurationSource? source, IConfigurationTelemetry te
 
             DataPipelineEnabled = config
                                   .WithKeys(ConfigurationKeys.TraceDataPipelineEnabled)
-                                  .AsBool(defaultValue: false);
+                                  .AsBool(defaultValue: EnvironmentHelpers.IsUsingAzureAppServicesSiteExtension() && FrameworkDescription.Instance.IsWindows());
```
lucaspimentel: We don't really need to check for Windows. It's already implied since AAS only supports site extensions on Windows.

lucaspimentel: This check will also enable the setting for Azure Functions that are using the site extension. Do we want to include those, or should we exclude them with `!EnvironmentHelpers.IsAzureFunctions()`?

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
lucaspimentel: Good catch. This would have been enabled in Azure Functions using the nuget package, which doesn't include `libdatadog`.

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
lucaspimentel:
```suggestion
                                  .AsBool(defaultValue: EnvironmentHelpers.IsUsingAzureAppServicesSiteExtension() && !EnvironmentHelpers.IsAzureFunctions(());
```

PR #7465, opened 2025-09-04
[IAST] Delay dataflow registration

daniel-romano-DD:
## Summary of changes
Register Dataflow in the `RejitManager` after the aspects have been loaded, not on object construction

## Reason for change
Avoid potential race conditions

### DIFF
--- a/tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
+++ b/tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
@@ -191,55 +191,59 @@ Dataflow::~Dataflow()
 
 void Dataflow::LoadAspects(WCHAR** aspects, int aspectsLength, UINT32 enabledCategories, UINT32 platform)
 {
-    // Init aspects
-    DBG("Dataflow::LoadAspects -> Processing aspects... ", aspectsLength, " Enabled categories: ", enabledCategories, " Platform: ", platform);
-
-    if (aspectsLength > 10)
+    if (!_initialized)
     {
-        _aspects.reserve(aspectsLength); // We know the max number of aspects we are going to have, so reserve the space to avoid vector resizes
-        _aspectClasses.reserve(aspectsLength / 10); // We don't know exactly the number of aspects which are class aspects, but 1/10 is a fine approach
-    }
+        _initialized = true;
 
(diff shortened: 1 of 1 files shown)


e-n-0 (approved):

tracer/src/Datadog.Tracer.Native/rejit_handler.cpp
```diff
@@ -394,7 +394,7 @@ void RejitHandler::RegisterRejitter(Rejitter* rejitter)
     if (m_rejittersCount == 0)
     {
         m_rejitters[m_rejittersCount++] = rejitter;
-        DBG("RejitHandler::RegisterRejitter -> Registered Rejitter. Count : ", m_rejittersCount);
+        Logger::Info("RejitHandler::RegisterRejitter -> Registered Rejitter. Count : ", m_rejittersCount);
```
andrewlock, line 397: nit: I wonder if we should update `RegisterRejitter()` to pass in a rejitter name for the logs 🤔 We can likely deduce it, but don't know, might be nice to be explicit?

anna-git (approved):

tracer/src/Datadog.Tracer.Native/rejit_handler.cpp
```diff
@@ -427,7 +427,8 @@ HRESULT RejitHandler::NotifyReJITParameters(ModuleID moduleId, mdMethodDef metho
     FunctionControlWrapper functionControl((ICorProfilerInfo*)m_profilerInfo, moduleId, methodId);
 
     // Call all rejitters sequentially
-    for (auto x = 0; x < m_rejittersCount; x++)
+    auto c = m_rejittersCount;
```
andrewlock, line 430: Is this incase the count changes between iterations? 🤔 AFAICT this _shouldn't_ be necessary, because the count only increases, never decreases, and it's incremented _after_ the value is set in the array...

And if that's _not_ the case, and the value _can_ decrease, then we have other problems here 😅 

If it's for some other reason, then nit:
```suggestion
    const auto c = m_rejittersCount;
```

dromanol: My only concern is when inserting a new one, that we could possible call the same one twice while it's being shifted, hence the change.

PR #7467, opened 2025-09-04
Update language used in native log messages

andrewlock:
## Summary of changes

Updates some of the language used in native logs to avoid ambiguous use of `Profiler`

## Reason for change

`Profiler` is tricky. We use the .NET profiling APIs for all our native functionality, so our native components are technically all "profilers". But this is ambiguous with the Continuous Profiler _product_, and can often confuse both customers and support.

To reduce confusion, we should aim to _avoid_ using "Profiler" to refer to the native components where possible. In particular

- Inside the native loader, when we disable the whole product, use `Datadog SDK`
  - e.g. `The Datadog SDK has been disabled`
- Always refer to the Continuous Profiler as `Continuous Profiler`
  - e.g. `The Continuous Profiler has been disabled`
- Refer to the native tracer as either `Instrumentation` or `Instrumentation component` as appropriate
  - e.g. `Instrumentation has been disabled` or `The Instrumentation component failed to initialize`
- Don't refer to the "managed profiler", instead use `Datadog.Trace.dll` (I hope it's not necessary to refer to the Datadog.Trace.ClrProfiler.Managed.Loader.dll 😅)
  -  e.g. `Unable to initialize: Datadog.Trace.dll was not yet loaded into the App Domain`

Note that we're _primarily_ concerned with the "high level" log messages which are "customer facing". See other details for exceptions

## Implementation details

Updated the logs in the native loader and native tracer in line with this guidance. I'm assuming the CP already follows this.

## Test coverage

N/A

## Other details

Note that internal/technical naming is still valid, so we still have the same distinct components, e.g.

- Native loader
- Native tracer
- Managed tracer loader
- Managed tracer
- Libdatadog
- Continuous Profiler

Additionally, we can still use `CorProfiler` / `ICorProfiler` / `COR Profiler` to refer to the specific runtime components if necessary.

### DIFF
--- a/shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
+++ b/shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
@@ -224,7 +224,7 @@ namespace datadog::shared::nativeloader
 
                     if (is_ignored_command)
                     {
-                        Log::Info("Instrumentation has been disabled because the process is 'dotnet' "
+                        Log::Info("Instrumentation has been disabled because the process name is 'dotnet' "
                             "but an unsupported command was detected");
                         return CORPROF_E_PROFILER_CANCEL_ACTIVATION;
                     }
@@ -240,8 +240,8 @@ namespace datadog::shared::nativeloader
                             process_command_line.find(WStr("datacollector")) == WSTRING::npos &&
                             process_command_line.find(WStr("vstest.console.dll")) == WSTRING::npos)
                         {
-                            Log::Info("Instrumentation has been disabled because the process is running in CI Visibility "
-                                "mode, the name is 'dotnet' but the commandline doesn't contain 'testhost' or 'datacollector' or 'vstest.console.dll' or 'exec'");
+                            Log::Info("Instrumentation disabled because the process is running in CI Visibility "
+                                "mode and its name is 'dotnet', but the command line doesn't contain 'testhost', 'datacollector', 'vstest.console.dll', or 'exec'");
                             return CORPROF_E_PROFILER_CANCEL_ACTIVATION;
                         }
                     }
--- a/tracer/src/Datadog.Tracer.Native/cor_profiler.cpp
+++ b/tracer/src/Datadog.Tracer.Native/cor_profiler.cpp
@@ -62,8 +62,8 @@ HRESULT STDMETHODCALLTYPE CorProfiler::Initialize(IUnknown* cor_profiler_info_un
         else
         {
             Logger::Error("Instrumentation is initialized multiple times. This may cause unpredictable failures.",
-                " When running aspnetcore in IIS, make sure to disable managed code in the application pool settings.",
-                " https://learn.microsoft.com/en-us/aspnet/core/host-and-deploy/iis/advanced?view=aspnetcore-9.0#create-the-iis-site");
+                " When running ASP.NET Core in IIS, make sure to disable managed code in the Application Pool settings.",
+                " https://learn.microsoft.com/en-us/aspnet/core/host-and-deploy/iis/advanced#create-the-iis-site");
         }
     }
 
@@ -87,8 +87,7 @@ HRESULT STDMETHODCALLTYPE CorProfiler::Initialize(IUnknown* cor_profiler_info_un
     }
     else
     {
-        Logger::Warn("DATADOG TRACER DIAGNOSTICS - Instrumentation disabled: .NET 5.0 runtime or greater is required on this "
-                     "architecture.");
+        Logger::Warn("DATADOG TRACER DIAGNOSTICS - Instrumentation disabled: .NET 5.0 runtime or greater is required on ARM architectures.");
         return E_FAIL;
     }
 #endif


tonyredondo (approved):
LGTM, 
Also, I think there's a good opportunity here to change two small nits that I hate about our current native logger:

1. The date format: `09/04/25 04:02:45.112 PM` .... can we move to the managed format: `2025-09-04 16:02:45.723 +02:00`
2. Status names: `[info]` in the native and `[INF]` in the managed.

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -488,7 +488,7 @@ namespace datadog::shared::nativeloader
 
                     Log::Debug("CorProfiler::Initialize: *LocalMaskLow: ", local_mask_low);
                     Log::Debug("CorProfiler::Initialize: *LocalMaskHi : ", local_mask_hi);
-                    Log::Info("CorProfiler::Initialize: Tracer Profiler initialized successfully.");
+                    Log::Info("CorProfiler::Initialize: Instrumetation initialized successfully.");
```
zacharycmontoya:
```suggestion
                    Log::Info("CorProfiler::Initialize: Instrumentation initialized successfully.");
```

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -497,7 +497,7 @@ namespace datadog::shared::nativeloader
             }
             else
             {
-                Log::Warn("CorProfiler::Initialize: Error Initializing the Tracer Profiler, unloading the dynamic library.");
+                Log::Warn("CorProfiler::Initialize: Error Initializing the Instrumetation component, unloading the dynamic library.");
```
zacharycmontoya:
```suggestion
                Log::Warn("CorProfiler::Initialize: Error Initializing the Instrumentation component, unloading the dynamic library.");
```

zacharycmontoya (approved):
Thanks for taking the time to do this! We'll also need to update our own internal support docs with updated log lines, so team members can CTRL+F on the new debug logs

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -130,7 +130,7 @@ namespace datadog::shared::nativeloader
         // process's name is on the list
         if (!include_process_names.empty() && !Contains(include_process_names, process_name))
         {
-            Log::Info("CorProfiler::Initialize ClrProfiler disabled: ", process_name, " not found in ",
+            Log::Info("CorProfiler::Initialize Datadog SDK disabled: ", process_name, " not found in ",
```
lucaspimentel, line 133: > Datadog SDK

"but the auto-instrumentation library is not an SDK!!" :trollface:

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -224,7 +224,7 @@ namespace datadog::shared::nativeloader
 
                     if (is_ignored_command)
                     {
-                        Log::Info("The Tracer Profiler has been disabled because the process is 'dotnet' "
+                        Log::Info("Instrumentation has been disabled because the process is 'dotnet' "
```
lucaspimentel: nits
```suggestion
                        Log::Info("Instrumentation has been disabled because the process name is 'dotnet' "
```

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -240,7 +240,7 @@ namespace datadog::shared::nativeloader
                             process_command_line.find(WStr("datacollector")) == WSTRING::npos &&
                             process_command_line.find(WStr("vstest.console.dll")) == WSTRING::npos)
                         {
-                            Log::Info("The Tracer Profiler has been disabled because the process is running in CI Visibility "
+                            Log::Info("Instrumentation has been disabled because the process is running in CI Visibility "
                                 "mode, the name is 'dotnet' but the commandline doesn't contain 'testhost' or 'datacollector' or 'vstest.console.dll' or 'exec'");
```
lucaspimentel:
```suggestion
                            Log::Info("Instrumentation disabled because the process is running in CI Visibility "
                                "mode and its name is 'dotnet', but the command line doesn't contain 'testhost', 'datacollector', 'vstest.console.dll', or 'exec'");
```

tracer/src/Datadog.Tracer.Native/class_factory.cpp
```diff
@@ -54,7 +54,7 @@ HRESULT STDMETHODCALLTYPE ClassFactory::CreateInstance(IUnknown* pUnkOuter, REFI
         return CLASS_E_NOAGGREGATION;
     }
 
-    trace::Logger::Info("Datadog CLR Profiler ", PROFILER_VERSION, " on",
+    trace::Logger::Info("Datadog Instrumentation ", PROFILER_VERSION, " on",
```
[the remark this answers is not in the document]
lucaspimentel: Oh, maybe we were logging the versions of the native and manage libraries separately to help with version mismatch scenarios. In that case `Instrumentation` make sense (as the user-facing name for the CLR Profiler).

tracer/src/Datadog.Tracer.Native/cor_profiler.cpp
```diff
@@ -57,11 +57,11 @@ HRESULT STDMETHODCALLTYPE CorProfiler::Initialize(IUnknown* cor_profiler_info_un
...
         if (isRunningInAas)
         {
-            Logger::Info("The Tracer Profiler is initialized multiple times. This is expected and currently unavoidable when running in AAS.");
+            Logger::Info("Instrumentation is initialized multiple times. This is expected and currently unavoidable when running in AAS.");
         }
         else
         {
-            Logger::Error("The Tracer Profiler is initialized multiple times. This may cause unpredictable failures.",
+            Logger::Error("Instrumentation is initialized multiple times. This may cause unpredictable failures.",
                 " When running aspnetcore in IIS, make sure to disable managed code in the application pool settings.",
```
lucaspimentel:
```suggestion
                " When running ASP.NET Core in IIS, make sure to disable managed code in the Application Pool settings.",
```

tracer/src/Datadog.Tracer.Native/cor_profiler.cpp
```diff
@@ -57,11 +57,11 @@ HRESULT STDMETHODCALLTYPE CorProfiler::Initialize(IUnknown* cor_profiler_info_un
...
                 " https://learn.microsoft.com/en-us/aspnet/core/host-and-deploy/iis/advanced?view=aspnetcore-9.0#create-the-iis-site");
```
lucaspimentel: If you remove `view` from the query string, the page will automatically select the latest version, so we don't need to update this when new versions are released.
```suggestion
                " https://learn.microsoft.com/en-us/aspnet/core/host-and-deploy/iis/advanced#create-the-iis-site");
```

tracer/src/Datadog.Tracer.Native/cor_profiler.cpp
```diff
@@ -87,7 +87,7 @@ HRESULT STDMETHODCALLTYPE CorProfiler::Initialize(IUnknown* cor_profiler_info_un
     }
     else
     {
-        Logger::Warn("DATADOG TRACER DIAGNOSTICS - Profiler disabled: .NET 5.0 runtime or greater is required on this "
+        Logger::Warn("DATADOG TRACER DIAGNOSTICS - Instrumentation disabled: .NET 5.0 runtime or greater is required on this "
                      "architecture.");
```
lucaspimentel:
```suggestion
        Logger::Warn("DATADOG TRACER DIAGNOSTICS - Instrumentation disabled: .NET 5.0 runtime or greater is required on ARM architectures.");
```

tracer/src/Datadog.Tracer.Native/interop.cpp
```diff
@@ -27,7 +27,7 @@ EXTERN_C VOID STDAPICALLTYPE GetAssemblyAndSymbolsBytes(BYTE** pAssemblyArray, i
 {
     if (trace::profiler == nullptr)
     {
-        trace::Logger::Error("Error in GetAssemblyAndSymbolsBytes call. Tracer CLR Profiler was not initialized.");
+        trace::Logger::Error("Error in GetAssemblyAndSymbolsBytes call. Instrumentation component was not initialized.");
```
lucaspimentel, line 30: Not a blocker, but I think
```
Instrumentation library was not initialized.
```
sounds better.

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -151,7 +151,7 @@ namespace datadog::shared::nativeloader
             {
                 if (process_name == exclude_assembly)
                 {
-                    Log::Info("CorProfiler::Initialize ClrProfiler disabled: ", process_name," found in default exclude list");
+                    Log::Info("CorProfiler::Initialize Datadog SDK disabled: ", process_name," found in default exclude list");
```
lucaspimentel, line 154: Unrelated, but why are they called `default_exclude_assemblies` and `exclude_assembly` when it's a list of process names? 😅

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
andrewlock: You're preaching to the choir man, just give me an alternative suggestion 😂 Datadog Library? Or we can just drop SDK entirely - 'Datadog disabled" works too 🤷‍♂️

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
andrewlock: A totally reasonable question 😅 As far as I can tell, it all traces back years ago and was essentially copy pasta. Going to do a bit more cleanup separately

andrewlock: Did a PR to fix that (and other minor annoyances) here: https://github.com/DataDog/dd-trace-dotnet/pull/7475

PR #7469, opened 2025-09-04
Remove `Samples.TraceAnnotations.VersionMismatch.NewerNuget`

andrewlock:
## Summary of changes

Removes the `Samples.TraceAnnotations.VersionMismatch.NewerNuget` sample and associated test

## Reason for change

This test was originally intended for testing scenarios where a _newer_ `Datadog.Trace.dll` is loaded via a NuGet than the native component version. However, since v3, we don't ship `Datadog.Trace.dll` in the dll, so this isn't technically a real scenario.

What's more, in #7287, this test is currently broken, because the `Datadog.Trace.dll` doesn't contain the API call we need. The "correct" way to fix this test is to create a _new_ version of the datadog package, which contains Datadog.Trace.dll with a package version of `255.x.x` and push that to our internal feed. However, that's a real pain, and like I say doesn't represent reality any more, so I propose we just drop the test.

## Implementation details

- Remove `Samples.TraceAnnotations.VersionMismatch.NewerNuget`
- Remove `VersionMismatchNewerNugetTests` test
- Remove references to `VersionMismatchNewerNugetTests`

## Test coverage

Less than before, but meh

## Other details

Blocker for 


NachoEchevarria (approved):
Love PRs that simplify things. Thanks!

PR #7471, opened 2025-09-04
Use a 2.x version of the tracer in CI vis version mismatch test

andrewlock:
## Summary of changes

Uses a 2.x version instead of a fictional 255.x.x version

## Reason for change

This scenario isn't a "real" scenario any more, as we don't ship the Datadog.Trace.dll in the nuget package. 

## Implementation details

Switch the version from our fictional (private, repack of 2.x.x) to use a real 2.x.x. Note that this hits different paths on the native side, but this scenario can actually happen whereas the previous one can't.

## Test coverage

Unchanged, except that no we test a real scenario instead of a fake one.

## Other details

Blocker for

### DIFF
--- a/tracer/test/test-applications/integrations/Samples.CIVisibilityVersionMismatch/Samples.CIVisibilityVersionMismatch.csproj
+++ b/tracer/test/test-applications/integrations/Samples.CIVisibilityVersionMismatch/Samples.CIVisibilityVersionMismatch.csproj
@@ -25,7 +25,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Datadog.Trace" Version="255.1.6-prerelease" />
+    <PackageReference Include="Datadog.Trace" Version="2.60.0" />
   </ItemGroup>
 
   <ItemGroup>


tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/CI/VersionMismatchTests.cs
```diff
@@ -33,6 +33,12 @@ public VersionMismatchTests(ITestOutputHelper output)
     {
         SetServiceName("version-mismatch-tests");
         SetServiceVersion("1.0.0");
+
+        // We disable the process integration to prevent the `stat` process spans that
+        // _may_ be present depending on the underlying host system (cgroup v1/v2).
+        // The "do not trace" helper that normally blocks tracing these spans doesn't
+        // work in version conflict scenarios.
+        SetEnvironmentVariable("DD_TRACE_Process_ENABLED", "0");
```
chrisnas:
```suggestion
        SetEnvironmentVariable("DD_TRACE_PROCESS_ENABLED", "0");
```
I guess it should be all UPPER CASE

chrisnas (approved):
LGTM except maybe typo to checl

PR #7472, opened 2025-09-04
Improve logging for `InstrumentationTests` and update dynamic code scenario

andrewlock:
## Summary of changes

- Add some missing `SetLogDirectory()` calls for `InstrumentationTests`
- Update the "dynamic code" instrumentation tests to throw away the logs

## Reason for change

`InstrumentationTests` tests a bunch of very different scenarios. To make it easier to understand the failures, we generally put the logs for each test into its own subfolder by calling `SetLogDirectory()`. We were missing this call for some tests, so added them.

Additionally, while working on #7287 we found a scenario which is currently "broken"
- We instrument an app that has "dynamic code disabled"
- We don't support this, so we bail out in the _managed loader_
- The `Tracer` is never initialized
- The Continuous Profiler writes an error on shutdown because the Tracer never sends the stable config.
- In `CheckBuildLogsForErrors` we fail the CI because of the error log.

_Ideally_ we would bail out in the _native_ loader so that nothing's loaded, but we don't know how to read the context switch on the native side ATM, so we can't. However, as this isn't a supported scenario, we're ok with the error log being there, and don't think it's worth trying to avoid it at this stage.

## Implementation details

- Add missing `SetLogDirectory()`
- Write the "dynamic code" logs to a "temp" directory so that they're not checked by the `CheckBuildLogsForErrors` stage

## Test coverage

Coverage is the same

## Other details

Blocker for 

### DIFF
--- a/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/InstrumentationTests.cs
+++ b/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/InstrumentationTests.cs
@@ -222,6 +222,7 @@ namespace Foo
         [Trait("RunOnWindows", "True")]
         public async Task WhenUsingRelativeTracerHome_InstrumentsApp()
         {
+            SetLogDirectory();
             // the working dir when we run the app is the _test_ project, not the app itself, so we need to be relative to that
             // This is a perfect example of why we don't recommend using relative paths for these variables
             var workingDir = Environment.CurrentDirectory;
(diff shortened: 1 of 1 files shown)


tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/InstrumentationTests.cs
```diff
@@ -635,6 +638,18 @@ public async Task WhenDynamicCodeIsEnabled_InstrumentsApp()
...
+            // We move the logs to a throw-away temp location so that they're not checked by the logs-checker in CI.
+            // That's because in this scenario the Profiler will never receive the stable config results, so will
+            // never initialize, because we bail-out in the managed loader (i.e. we never initialize the tracer).
+            // This isn't ideal - we'd rather bail out in the native loader - but it's not possible to check the context
+            // switch on the native side AFAIK (though we should check again to see if we can).
+            // Regardless, the CP is loaded, and generates an error in its logs, but as this isn't a supported scenario
+            // anyway, that's fine.
+
+            var logDir = Path.Combine(Path.GetTempPath(), Path.GetFileNameWithoutExtension(Path.GetRandomFileName()));
+            Directory.CreateDirectory(logDir);
```
chrisnas, line 650: Is it needed to delete it at the end of the test?

andrewlock: meh, it's in the temp folder and less than 1kb😅

chrisnas (approved):
LGTM

bouwkast (approved):

PR #7474, opened 2025-09-04
Attempt to improve Kafka timeout flake in tests

bouwkast:
## Summary of changes

We've been seeing timeouts of produced messages in the Kafka tests.
This increases the timeout slightly from 3 seconds to 5 seconds.
This also removes what appeared to be an accidental flush after every single synchronous produce.

## Reason for change

We've been seeing timeouts of produced messages in the Kafka tests.

## Implementation details

Changed `3000` -> `5_000`

Removed the flush call (after all messages are produced it still flushes).

## Test coverage

Same tests


andrewlock (approved):
Thanks, I really don't understand _why_ this would work, but why not give it a try 😄

PR #7475, opened 2025-09-05
More cleanup in native loader code

andrewlock:
## Summary of changes

- Cleanup of some of the native loader code
- Add `createdump` to the exclude list

## Reason for change

Various minor things that have been bugging me for a while.

## Implementation details

- Remove `default_exclude_assemblies` from the native tracer as it's no longer used (after #7642)
- Remove `default_exclude_assemblies` to `default_exclude_processes` as it was named incorrectly!
- Move environment variables that were in `util.h` into the `EnvironmentVariables` type that's designed for this (and remove duplicates)
- Rename `EnvironmentVariables.h` to `environment.h` for consistency with the majoring of the native loader code + native tracer. Use snake casing instead of PascalCasing
- Move constants out of `util.h` into `constants.h` to avoid cluttering it
- Add `createdump` to the process exclude list (I noticed we apparently instrument the `createdump` calls themselves in our CI!) 

## Test coverage

Covered by existing

## Other details

Commit-by-commit review is easiest, as they're all small self-contained updates!

### DIFF
--- a/shared/src/Datadog.Trace.ClrProfiler.Native/constants.h
+++ b/shared/src/Datadog.Trace.ClrProfiler.Native/constants.h
@@ -15,8 +15,8 @@ const shared::WSTRING default_exclude_processes[]{
     WStr("aspnet_compiler.exe"),
     WStr("aspnet_state.exe"),
     WStr("CollectGuestLogs.exe"), // https://github.com/Azure/WindowsVMAgent
-    WStr("csc.exe"),
     WStr("createdump"),
+    WStr("csc.exe"),
     WStr("dd-trace"),
     WStr("dd-trace.exe"),
     WStr("devenv.exe"),
--- a/shared/src/Datadog.Trace.ClrProfiler.Native/environment.h
+++ b/shared/src/Datadog.Trace.ClrProfiler.Native/environment.h
@@ -23,8 +23,7 @@ public:
     // Sets whether the current process must run in CI Visibility mode or not.
     inline static const shared::WSTRING ci_visibility_enabled = WStr("DD_CIVISIBILITY_ENABLED");
 
-    // Indicates whether the CLR profiler is running from
-    // the Azure App Services site extension
+    // Indicates whether the Datadog SDK is running from the Azure App Services site extension
     inline static const shared::WSTRING is_azure_app_services_extension = WStr("DD_AZURE_APP_SERVICES");
 
     // The app_pool_id in the context of azure app services
@@ -38,8 +37,7 @@ public:
     // Used as a flag to determine that we are running within a functions app.
     inline static const shared::WSTRING azure_app_services_functions_worker_runtime = WStr("FUNCTIONS_WORKER_RUNTIME");
 
-    // Determine whether to instrument within azure functions.
-    // Default is true.
+    // Enables instrumentation of Azure Functions. Default value is true.
     inline static const shared::WSTRING azure_functions_instrumentation_enabled = WStr("DD_TRACE_AZURE_FUNCTIONS_ENABLED");
 
     inline static const ::shared::WSTRING config_filepath = WStr("DD_NATIVELOADER_CONFIGFILE");


zacharycmontoya (approved):
Looks good!

shared/src/Datadog.Trace.ClrProfiler.Native/environment.h
```diff
@@ -0,0 +1,49 @@
...
+
+class environment final
+{
+public:
+    inline static const shared::WSTRING log_path = WStr("DD_TRACE_LOG_PATH");
+    inline static const shared::WSTRING log_directory = WStr("DD_TRACE_LOG_DIRECTORY");
+    inline static const shared::WSTRING log_buffering_enabled = WStr("DD_TRACE_LOG_BUFFERING_ENABLED");
+    inline static const shared::WSTRING debug_log_enabled = WStr("DD_TRACE_DEBUG");
+    inline static const shared::WSTRING include_process_names = WStr("DD_PROFILER_PROCESSES");
+    inline static const shared::WSTRING exclude_process_names = WStr("DD_PROFILER_EXCLUDE_PROCESSES");
```
lucaspimentel, line 16: Not for this PR, but we should consider renaming these two env vars to remove `PROFILER` from their names (while keeping backwards support for the old names). Something like:
```
DD_PROFILER_PROCESSES         -> DD_INCLUDE_PROCESS_NAMES
DD_PROFILER_EXCLUDE_PROCESSES -> DD_EXCLUDE_PROCESS_NAMES
```

shared/src/Datadog.Trace.ClrProfiler.Native/environment.h
```diff
@@ -0,0 +1,49 @@
...
+    // DD_INJECTION_ENABLED is non-empty when SSI is enabled
+    inline static const shared::WSTRING single_step_instrumentation_enabled = WStr("DD_INJECTION_ENABLED");
+    inline static const shared::WSTRING force_eol_instrumentation = WStr("DD_INJECT_FORCE");
+    inline static const shared::WSTRING single_step_instrumentation_telemetry_forwarder_path = WStr("DD_TELEMETRY_FORWARDER_PATH");
+
+    // Sets whether the current process must run in CI Visibility mode or not.
+    inline static const shared::WSTRING ci_visibility_enabled = WStr("DD_CIVISIBILITY_ENABLED");
+
+    // Indicates whether the CLR profiler is running from
+    // the Azure App Services site extension
```
lucaspimentel:
```suggestion
    // Indicates whether the Datadog SDK is running from the Azure App Services site extension
```

shared/src/Datadog.Trace.ClrProfiler.Native/environment.h
```diff
@@ -0,0 +1,49 @@
...
+    // The DOTNET_CLI_TELEMETRY_PROFILE in the context of azure app services
+    inline static const shared::WSTRING azure_app_services_cli_telemetry_profiler_value = WStr("DOTNET_CLI_TELEMETRY_PROFILE");
+
+    // The FUNCTIONS_WORKER_RUNTIME in Azure Functions.
+    // Valid values: "dotnet" (in-process functions) or "dotnet-isolated" (isolated functions).
+    // Used as a flag to determine that we are running within a functions app.
+    inline static const shared::WSTRING azure_app_services_functions_worker_runtime = WStr("FUNCTIONS_WORKER_RUNTIME");
+
+    // Determine whether to instrument within azure functions.
+    // Default is true.
```
lucaspimentel:
```suggestion
    // Enables instrumentation of Azure Functions. Default value is true.
```

shared/src/Datadog.Trace.ClrProfiler.Native/constants.h
```diff
@@ -0,0 +1,51 @@
...
+
+// Note that you should also consider adding to the SSI tracer/build/artifacts/requirements.json file
+// FIXME: this should also take into account case insensitivity, but that is not yet supported
+// https://devblogs.microsoft.com/oldnewthing/20241007-00/?p=110345
+const shared::WSTRING default_exclude_processes[]{
+    WStr("aspnet_compiler.exe"),
+    WStr("aspnet_state.exe"),
+    WStr("CollectGuestLogs.exe"), // https://github.com/Azure/WindowsVMAgent
+    WStr("csc.exe"),
+    WStr("createdump"),
```
lucaspimentel: We're trying to keep them in alphabetical order
```suggestion
    WStr("createdump"),
    WStr("csc.exe"),
```

tracer/src/Datadog.Tracer.Native/dd_profiler_constants.h
```diff
@@ -62,48 +62,6 @@ const shared::WSTRING include_assemblies[]{
     WStr("System.Runtime.Remoting"),
 };
 
-// Note that this list should be kept in sync with the values in shared/src/Datadog.Trace.ClrProfiler.Native/util.h
-// Note that you should also consider adding to the SSI tracer/build/artifacts/requirements.json file
-// FIXME: this should also take into account case insensitivity, but that is not yet supported
-// https://devblogs.microsoft.com/oldnewthing/20241007-00/?p=110345
-const shared::WSTRING default_exclude_assemblies[]{
```
lucaspimentel, line 69: (commenting here because the new UI doesn't support file-level comments)
Should we rename this file to remove "profiler" from the name? Maybe `constants.h`?

shared/src/Datadog.Trace.ClrProfiler.Native/constants.h
andrewlock: It's amazing how I clearly knew this and yet failed in my task 🤦

shared/src/Datadog.Trace.ClrProfiler.Native/environment.h
andrewlock: Yeah, while I totally agree in principal, this is very low down on the priority list (as I think you'd agree) given it's essentially an undocumented feature too. It would also be the first and only case of supporting "fallbacks" on the native-side I think

tracer/src/Datadog.Tracer.Native/dd_profiler_constants.h
andrewlock: Yeah, why not I guess 😄

andrewlock: But given this touches like 30 files, I'll defer to a different PR 😅

PR #7476, opened 2025-09-05
Try fixing ASAN job crashes

gleocadie:
## Summary of changes

This PR fixes the ASAN job which started crashing recently.

## Reason for change

This PR is the same as #7452 where we fix TSAN job which failed since a kernel update (6.6+). 
There are information [here](https://stackoverflow.com/questions/78293129/c-programs-fail-with-asan-addresssanitizerdeadlysignal) 
## Implementation details

Same as TSAN: set the Linux kernel parameter controlling the number of randomization bits used for memory regions allocated via mmap.

## Test coverage

ASAN job must be green.

### DIFF
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
@@ -7,6 +7,7 @@
 #include "corprof.h"
 // end
 
+/* TO REMOVE before merging. Added to force ASAN/UBSAN/TSAN jobs */
 #include "CorProfilerCallback.h"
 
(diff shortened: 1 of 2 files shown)


andrewlock (approved):
Worth a try!

profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
```diff
@@ -7,6 +7,7 @@
 #include "corprof.h"
 // end
 
+/* TO REMOVE before merging. Added to force ASAN/UBSAN/TSAN jobs */
```
andrewlock: I guess we remove this now? 😄 
```suggestion
```

PR #7483, opened 2025-09-05
[Hands-off config] add integration test

anna-git:
## Summary of changes

Adds an integration test for hands off config testing precedence, various sources.

## Implementation details

Creates the hands off config files in the system and deletes them
 later on


tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/Configuration/HandsOffConfigurationTests.cs
```diff
@@ -0,0 +1,199 @@
...
+        {
+            SetEnvironmentVariable(ConfigurationKeys.DebugEnabled, "true");
+            await SetupAndCheck(
+                nameof(FleetShouldWinOverEnvVar),
+                null,
+                null,
+                ConfigurationKeys.DebugEnabled,
+                "false",
+                "debug",
+                "false");
```
andrewlock: nit: these would be easier to read if you pass the parameter names too 🙂 
e.g.
```suggestion
            SetEnvironmentVariable(ConfigurationKeys.DebugEnabled, "true");
            await SetupAndCheck(
                logFolder: nameof(FleetShouldWinOverEnvVar),
                localKey: null,
                localValue: null,
                // etc

```

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/Configuration/HandsOffConfigurationTests.cs
```diff
@@ -0,0 +1,199 @@
...
+
+            if (File.Exists(localDestPath))
+            {
+                File.Delete(localDestPath);
+            }
+
+            if (File.Exists(fleetDestPath))
+            {
+                File.Delete(fleetDestPath);
+            }
```
andrewlock: I'm now wondering if we'll potentially break "real" deployments on our dev machine by doing this 😅

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/Configuration/HandsOffConfigurationTests.cs
```diff
@@ -0,0 +1,199 @@
...
+                fleetLines.Insert(0, "apm_configuration_default:");
+                File.AppendAllLines(fleetDestPath, fleetLines);
+            }
+
+            return (localDestPath, fleetDestPath);
+        }
+
+        private async Task SetupAndCheck(string logFolder, string? localKey, string? localValue, string? fleetKey, string? fleetValue, string logKey, string expectedValue)
+        {
+            // SkipOn.Platform(SkipOn.PlatformValue.MacOs);
```
andrewlock: Does this need to be uncommented or deleted? Does it work in CI?

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/Configuration/HandsOffConfigurationTests.cs
```diff
@@ -0,0 +1,199 @@
...
+
+            var logDir = Path.Combine(LogDirectory, logFolder);
+            Directory.CreateDirectory(logDir);
+            SetEnvironmentVariable(ConfigurationKeys.LogDirectory, logDir);
+
+            using var agent = MockTracerAgent.Create(Output, useTelemetry: false);
+            Output.WriteLine($"Assigned port {agent.Port} for the agentPort.");
+
+            var processName = EnvironmentHelper.IsCoreClr() ? "dotnet" : "Samples.Console";
+            using var logEntryWatcher = new LogEntryWatcher($"{LogFileNamePrefix}{processName}*", logDir, Output);
```
andrewlock, line 181: I would suggest not bothering with LogEntryWatcher tbh. That's useful when you need to read the logs in realtime, but it's prone to race conditions. Just read the logs after the sample has exited instead

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/Configuration/HandsOffConfigurationTests.cs
```diff
@@ -0,0 +1,199 @@
...
+
+            var (localDestPath, fleetDestPath) = CreateHandsOffConfigFiles(
+                localKey is not null ? [(Key: localKey, Value: localValue!)] : [],
+                fleetKey is not null ? [(Key: fleetKey, Value: fleetValue!)] : []);
+
+            using var sample = await StartSample(agent, "wait", string.Empty, aspNetCorePort: 5000);
```
andrewlock, line 187: I think you should just run the sample, and exit. No need to `wait` or generate traces or anything. That will be faster and less flaky

andrewlock (approved):
Thanks! Just a few suggestions to improve readability and reduce the chance of flake 🙂

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/Configuration/HandsOffConfigurationTests.cs
anna-git: I didn't think we used the same machines 😅  I can always add some new env vars...

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/Configuration/HandsOffConfigurationTests.cs
anna-git: Yes added in 167f482a74032c90598820f303387e93abebac76
As rider is configured to show param names by default I didnt notice the inconvenience 😅

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/Configuration/HandsOffConfigurationTests.cs
anna-git: removed in [f1aa830](https://github.com/DataDog/dd-trace-dotnet/pull/7483/commits/f1aa830cfa30d04795fc5bd85215dbe4c2ea49db)

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/Configuration/HandsOffConfigurationTests.cs
anna-git: as discussed, have to as logs dont have time to be flushed before app is shutdown, we end up with a log without the CONFIGURATION part

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/Configuration/HandsOffConfigurationTests.cs
anna-git: same as above , logs dont have time to be flushed before app is shutdown, we end up with a log without the CONFIGURATION part

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/Configuration/HandsOffConfigurationTests.cs
andrewlock: I mean literally our own machines that we develop with - we deploy the agent locally, if they decide to deploy the fleet config via that too, we'll clobber it if we run this test locally. But tbh, meh 😄

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/Configuration/HandsOffConfigurationTests.cs
andrewlock: Yep, totally understand that one! Thanks! 😄

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/Configuration/HandsOffConfigurationTests.cs
anna-git: oh I see now 😅 
Probably not very nice but we could: **not** delete the files, and create them overwriting potential existing files, and writing a comment at the beginning of the file like `This file has been created when running integration test `HandsOffConfigurationTests`  and might potentially have overwritten system files` ..
That way, when investigating why hands off config is weird locally, and browsing to the files.. we would understand 😅

andrewlock: Yeah, alternatively an improved option would be to:
- If the file exists, move it to a temporary path
- Run the test (ideally with your comment too)
- On test end, restore the file

I'm not too worried though, we can always do this later, and it's relatively unlikely for people to run the test locally explicitly (now that you're not!)

anna-git: Much better idea! Hopefully [f734c41](https://github.com/DataDog/dd-trace-dotnet/pull/7483/commits/f734c415e318bff131e6a1d394921a80b8f4aee4) works 😅

PR #7490, opened 2025-09-09
Check for errors when signing

NachoEchevarria:

### DIFF
--- a/tracer/build/_build/Build.Gitlab.cs
+++ b/tracer/build/_build/Build.Gitlab.cs
@@ -102,9 +102,52 @@ partial class Build
                     logOutput: false,
                     logInvocation: false);
             signProcess.WaitForExit();
+
+            var output = signProcess.Output.Select(o => o.Text);
+            foreach (var line in output)
+            {
+                Logger.Information("[dd-wcs] {Line}", line);
+
+                // dd-wcs will return 0 even if there are errors
+                if (line.StartsWith("ERROR:", StringComparison.OrdinalIgnoreCase))
+                {
+                    throw new Exception($"Error found when signing {binaryPath}: {line}");
+                }
+            }
+
             if (signProcess.ExitCode == 0)
             {
-                PowerShellTasks.PowerShell($"Get-AuthenticodeSignature {binaryPath}");
+                var sigResult = PowerShellTasks.PowerShell(
+                    $"Get-AuthenticodeSignature '{binaryPath}' | Format-List",
+                    logOutput: false,
+                    logInvocation: false);
+
+                var lines = sigResult
+                    .Select(o => o.Text?.Trim())
+                    .Where(l => !string.IsNullOrEmpty(l))
+                    .ToList();
+
+                foreach (var line in lines)
+                {
+                    Logger.Information("[sigcheck] {Line}", line);
+                }
+
+                var statusLine = lines
+                    .FirstOrDefault(l => l.StartsWith("Status", StringComparison.OrdinalIgnoreCase));
+
+                if (statusLine != null)
+                {
+                    var status = statusLine.Split(':').Last().Trim();
+
+                    if (!status.Equals("Valid", StringComparison.OrdinalIgnoreCase))
+                    {
+                        throw new Exception($"Signature verification failed for {binaryPath}: {status}");
+                    }
+                }
+                else
+                {
+                    throw new Exception($"Signature verification failed for {binaryPath}: no status line found");
+                }
             }
             else
             {


tracer/build/_build/Build.Gitlab.cs
```diff
@@ -102,9 +102,52 @@ void SignBinary(AbsolutePath binaryPath)
...
+            var output = signProcess.Output.Select(o => o.Text);
+            foreach (var line in output)
+            {
+                Logger.Information("[dd-wcs] {Line}", line);
+
+                // dd-wcs will return 0 even if there are errors
+                if (line.StartsWith("ERROR:", StringComparison.OrdinalIgnoreCase))
+                {
+                    throw new Exception($"Error found when signing {binaryPath}: {line}");
+                }
```
bouwkast, line 117: And if I understood correctly you were able to test this out locally right?

bouwkast (approved):
Awesome thanks!

tracer/build/_build/Build.Gitlab.cs
NachoEchevarria: Yes, I was getting some error messages in the tool's log starting with the word "ERROR:". The tool would return 0 though.

tracer/build/_build/Build.Gitlab.cs
```diff
@@ -102,9 +102,52 @@ void SignBinary(AbsolutePath binaryPath)
...
+                    logInvocation: false);
+
+                var lines = sigResult
+                    .Select(o => o.Text?.Trim())
+                    .Where(l => !string.IsNullOrEmpty(l))
+                    .ToList();
+
+                foreach (var line in lines)
+                {
+                    Logger.Information("[sigcheck] {Line}", line);
```
lucaspimentel: Should we use the name of the command instead of "sigcheck"?

```suggestion
                    Logger.Information("[Get-AuthenticodeSignature] {Line}", line);
```

Otherwise, if you see this in the logs, you have to come to the code here to know where it comes from.

tracer/build/_build/Build.Gitlab.cs
```diff
@@ -102,9 +102,52 @@ void SignBinary(AbsolutePath binaryPath)
...
+                }
+
+                var statusLine = lines
+                    .FirstOrDefault(l => l.StartsWith("Status", StringComparison.OrdinalIgnoreCase));
```
lucaspimentel: We should probably be more strict here because the output has 2 lines that start with `"Status"` and the order is probably not guaranteed:

```
...
Status                 : <status>
StatusMessage          : <message>
...
```

lucaspimentel: We could run
```
(Get-AuthenticodeSignature <filename>).Status
```
which returns the status only (`Valid`, `NotSigned`, etc).

(full enum here: https://learn.microsoft.com/en-us/dotnet/api/system.management.automation.signaturestatus)

tracer/build/_build/Build.Gitlab.cs
NachoEchevarria: Sure! Thanks!

tracer/build/_build/Build.Gitlab.cs
lucaspimentel: Also, if `Status == Valid`, we could go an extra step and confirm it was signed with the correct certificate
```
(Get-AuthenticodeSignature datadog-dotnet-apm-3.26.1-x64.msi).SignerCertificate
```
should return `59063C826DAA5B628B5CE8A2B32015019F164BF0`

NachoEchevarria: Thanks for the info! I have added these suggestions to the code!

tracer/build/_build/Build.Gitlab.cs
```diff
@@ -88,6 +88,8 @@ partial class Build
 
     void SignFiles(IReadOnlyCollection<AbsolutePath> filesToSign)
     {
+        var validSignature = "59063C826DAA5B628B5CE8A2B32015019F164BF0";
```
lucaspimentel:
```suggestion
        const string validSignature = "59063C826DAA5B628B5CE8A2B32015019F164BF0";
```

tracer/build/_build/Build.Gitlab.cs
```diff
@@ -102,9 +104,44 @@ void SignBinary(AbsolutePath binaryPath)
...
+                    $"(Get-AuthenticodeSignature '{binaryPath}').Status",
+                    logOutput: false,
+                    logInvocation: false);
+
+                var lines = status.Select(o => o.Text?.Trim()).Where(l => !string.IsNullOrEmpty(l)).ToList();
+
+                if (lines.Count == 0 || !lines[0].Equals("Valid", StringComparison.OrdinalIgnoreCase))
+                {
+                    throw new Exception($"Signature verification failed for {binaryPath}. Status: {(lines.Count > 0 ? lines[0] : "Empty")}");
+                }
```
lucaspimentel: You can simplify this a little and remove `.Where().ToList()` and the checks for `lines.Count` by using `FirstOrDefault()`. Something like:

```csharp
var result = status.Select(o => o.Text)
                   .FirstOrDefault(l => !string.IsNullOrEmpty(l))
                  ?.Trim();

// result can be null
if (!string.Equals(result, "Valid", StringComparison.OrdinalIgnoreCase))
{
    throw new Exception($"Signature verification failed for {binaryPath}. Status: {result ?? "Empty"}");
}
```

Same comment below for the second command.

NachoEchevarria: Done! Thanks!

lucaspimentel (approved):
Thanks!

tracer/build/_build/Build.Gitlab.cs
andrewlock: Can/should we also see about pushing a fix upstream to the `dd wcs` tool so that it doesn't return `0` in those cases?

NachoEchevarria:
Thanks for the feedback, suggestions and reviews!

PR #7491, opened 2025-09-09
feat(ssi): add workload selection

dmehala:


tracer/build/_build/Build.Shared.Steps.cs
```diff
@@ -47,6 +51,47 @@ partial class Build
...
+                if (!string.Equals(expectedHash, actualHash, StringComparison.Ordinal))
+                {
+                    throw new Exception($"Downloaded file did not have expected hash. Expected hash {expectedHash}, actual hash {actualHash}");
+                }
+
+                Logger.Information("Hash verified: '{Hash}'", expectedHash);
+
+                // Unzip to expected location
+                EnsureExistingDirectory(output);
+                foreach (var p in Directory.GetFiles(output).Where(x => Path.GetFileName(x) != ".gitignore"))
```
bmermet, line 81: Why do we delete files in this directory?

tracer/build/_build/Build.Shared.Steps.cs
```diff
@@ -47,6 +51,47 @@ partial class Build
...
+                {
+                    DeleteFile(p);
+                }
+
+                foreach (var d in Directory.GetDirectories(output))
```
bmermet: Since we are looking for a single directory I think it would be clearer to skip the iteration on all subdirectories, like:
```
var policyDir = Path.Combine(output, "dd-policy-engine");
if (Directory.Exists(policyDir))
{
    DeleteDirectory(policyDir);
}
```

tracer/build/_build/Build.Shared.Steps.cs
```diff
@@ -32,8 +33,11 @@ partial class Build
     Target CompileNativeLoaderWindows => _ => _
         .Unlisted()
         .OnlyWhenStatic(() => IsWin)
-        .Executes(() =>
+        .Executes(async () =>
         {
+            // This will clean and download every time, we likely want to add
```
bmermet, line 38: Such suspense, the public wants to know the end of this sentence.

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -316,7 +323,59 @@ namespace datadog::shared::nativeloader
...
+
+        wls::init();
+        wls::set_params(wls::StringEvaluator::PROCESS_EXE, p);
+        wls::set_params(wls::StringEvaluator::RUNTIME_LANGUAGE, "dotnet");
+        wls::set_params(wls::NumericEvaluator::RUNTIME_VERSION_MAJOR, (unsigned long)runtimeInformation.major_version);
+        wls::set_params(wls::NumericEvaluator::RUNTIME_VERSION_MINOR, (unsigned long)runtimeInformation.minor_version);
+        wls::set_params(wls::NumericEvaluator::RUNTIME_VERSION_PATCH, (unsigned long)runtimeInformation.build_version);
+
+        // TODO: Remove once the policy engine will be able to log the context.
+        Log::Info("CorProfiler::Initialize: Workload Selection Context: \"process.name:", p, " runtime:dotnet\"");
```
bmermet: nit: we can log the runtime version too since they are also added to the context

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -316,7 +323,59 @@ namespace datadog::shared::nativeloader
...
+                injection_status = InjectionStatus::DENY;
+            }
+            
+            return std::nullopt;
+        });
+
+        if (const auto wls_file = fs::path{GetDatadogProgramDataFolderPath()} / "protected" / "workload_selection.fb"; fs::exists(wls_file))
+        {
+          auto maybe_error = wls::evaluate_buffer_from_file(wls_file);
+          if (maybe_error || injection_status != InjectionStatus::ALLOW) {
```
bmermet: I think it would be helpful to log errors differently from deny here.

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -316,7 +323,59 @@ namespace datadog::shared::nativeloader
...
+        wls::set_params(wls::StringEvaluator::RUNTIME_LANGUAGE, "dotnet");
+        wls::set_params(wls::NumericEvaluator::RUNTIME_VERSION_MAJOR, (unsigned long)runtimeInformation.major_version);
+        wls::set_params(wls::NumericEvaluator::RUNTIME_VERSION_MINOR, (unsigned long)runtimeInformation.minor_version);
+        wls::set_params(wls::NumericEvaluator::RUNTIME_VERSION_PATCH, (unsigned long)runtimeInformation.build_version);
+
+        // TODO: Remove once the policy engine will be able to log the context.
+        Log::Info("CorProfiler::Initialize: Workload Selection Context: \"process.name:", p, " runtime:dotnet\"");
+
+        wls::register_action(wls::Action::INJECT_DENY, [&injection_status](wls::Result eval_result, const std::vector<const char*>&, const char* desc) -> std::optional<wls::Error> {
+            injection_status = InjectionStatus::DENY;
```
bmermet: Shouldn't we also check the eval_result?

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -316,7 +323,59 @@ namespace datadog::shared::nativeloader
...
+            
+            return std::nullopt;
+        });
+
+        if (const auto wls_file = fs::path{GetDatadogProgramDataFolderPath()} / "protected" / "workload_selection.fb"; fs::exists(wls_file))
+        {
+          auto maybe_error = wls::evaluate_buffer_from_file(wls_file);
+          if (maybe_error || injection_status != InjectionStatus::ALLOW) {
+            Log::Info("CorProfiler::Initialize: Instrumentation denied due to workload selection.");
+            return CORPROF_E_PROFILER_CANCEL_ACTIVATION;
```
bmermet: The default decision should be Allow for IIS

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -316,7 +323,59 @@ namespace datadog::shared::nativeloader
             return E_FAIL;
         }
 
-        // Guard rails have all passed, so we enable (and flush) logs if necessary
+        // Workload Selection
+        #if defined(BUILD_WITH_WORKLOAD_SELECTION)
+        enum class InjectionStatus : uint8_t {
```
bmermet: For the custom build it may work to have it behind a build flag, but we'll also need to guard it behind a runtime flag that turns on this behavior only when the tracer is loaded by SSI.

tracer/build/_build/Build.Shared.Steps.cs
andrewlock: Because if we're downloading a new version, we want to make sure that we don't have anything lying around

tracer/build/_build/Build.Shared.Steps.cs
andrewlock: 😅 I _started_ writing this and then decided to implement caching instead 😅 

```suggestion
```

tracer/build/_build/Build.Shared.Steps.cs
andrewlock: That would make a lot more sense, yeah

tracer/build/_build/Build.Shared.Steps.cs
```diff
@@ -47,6 +51,47 @@ partial class Build
...
+                {
+                    DeleteFile(p);
+                }
+
+                foreach (var d in Directory.GetDirectories(output))
+                {
+                    if (Path.GetFileName(d) == "dd-policy-engine") {
+                        DeleteDirectory(d);
+                    }
+                }
```
andrewlock:
```suggestion
                var policyDir = Path.Combine(output, "dd-policy-engine");
                if (Directory.Exists(policyDir))
                {
                    DeleteDirectory(policyDir);
                }
```

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -316,7 +322,55 @@ namespace datadog::shared::nativeloader
...
+          plcs::init();
+          plcs::set_params(plcs::StringEvaluator::PROCESS_EXE, ToString(process_name));
+          plcs::set_params(plcs::StringEvaluator::RUNTIME_LANGUAGE, "dotnet");
+
+          // TODO: Remove once the policy engine will be able to log the context.
+          Log::Info("CorProfiler::Initialize: Workload Selection Context: \"process.name:", process_name, " runtime:dotnet\"");
+
+          plcs::register_action(plcs::Action::INJECT_ALLOW, [&injection_status](plcs::Result eval_result, const std::vector<const char*>&, const char* desc) -> std::optional<plcs::Error> {
+              if (injection_status == InjectionStatus::ALLOW) return std::nullopt;
+              if (eval_result == plcs::Result::ALLOW)
```
bmermet: Discussed offline, let's use another name than Allow for the result to be consistent with the C implem and avoid the confusion with the action name.

bmermet (approved):

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -316,7 +322,56 @@ namespace datadog::shared::nativeloader
             return E_FAIL;
         }
 
-        // Guard rails have all passed, so we enable (and flush) logs if necessary
+        // Workload Selection
+        #if defined(BUILD_WITH_WORKLOAD_SELECTION)
```
andrewlock: nit: formatting, plus we don't need the `defined()` stuff, right?

```suggestion
#ifdef BUILD_WITH_WORKLOAD_SELECTION
```

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -316,7 +322,56 @@ namespace datadog::shared::nativeloader
...
+        if (IsSingleStepInstrumentation() && !IsRunningOnIIS())
```
andrewlock, line 327: I thought we _were_ going to have workload selection for IIS, we were just going to default inject into it? 🤔

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -316,7 +322,56 @@ namespace datadog::shared::nativeloader
...
+            ALLOW,
+            DENY,
+            UNKNOWN
+          } injection_status = InjectionStatus::UNKNOWN;
+
+          plcs::init();
+          plcs::set_params(plcs::StringEvaluator::PROCESS_EXE, ToString(process_name));
+          plcs::set_params(plcs::StringEvaluator::RUNTIME_LANGUAGE, "dotnet");
+
+          // TODO: Remove once the policy engine will be able to log the context.
```
andrewlock, line 339: How are we proposing for the policy engine to do logging?

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -316,7 +322,56 @@ namespace datadog::shared::nativeloader
...
+          Log::Info("CorProfiler::Initialize: Workload Selection Context: \"process.name:", process_name, " runtime:dotnet\"");
+
+          plcs::register_action(plcs::Action::INJECT_ALLOW, [&injection_status](plcs::Result eval_result, const std::vector<const char*>&, const char* desc) -> std::optional<plcs::Error> {
+              if (injection_status == InjectionStatus::ALLOW) return std::nullopt;
```
andrewlock: nit: formatting, brackets are always required
```suggestion
              if (injection_status == InjectionStatus::ALLOW)
              {
                  return std::nullopt;
              }
```

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -316,7 +322,56 @@ namespace datadog::shared::nativeloader
...
+          } injection_status = InjectionStatus::UNKNOWN;
+
+          plcs::init();
+          plcs::set_params(plcs::StringEvaluator::PROCESS_EXE, ToString(process_name));
+          plcs::set_params(plcs::StringEvaluator::RUNTIME_LANGUAGE, "dotnet");
+
+          // TODO: Remove once the policy engine will be able to log the context.
+          Log::Info("CorProfiler::Initialize: Workload Selection Context: \"process.name:", process_name, " runtime:dotnet\"");
+
+          plcs::register_action(plcs::Action::INJECT_ALLOW, [&injection_status](plcs::Result eval_result, const std::vector<const char*>&, const char* desc) -> std::optional<plcs::Error> {
```
andrewlock, line 342: I'd love some comments explaining this tbh, I don't really know what's going on here 😅

shared/src/Datadog.Trace.ClrProfiler.Native/util.h
```diff
@@ -26,14 +26,32 @@ static fs::path GetCurrentModuleFolderPath()
...
+        program_data_path = WStr(R"(C:\ProgramData)");
+    }
+
+    // on Windows WSTRING == wstring
+    return program_data_path / "Datadog";
+#else
+    // Is it the correct path on linux?
+    return shared::ToWSTRING("/var/log/datadog/dotnet/");
+#endif
+}
```
andrewlock: IIUC, this isn't really about program data, it's about where the workload selection flatbuffers file should be. I think we should rename this helper to be explicit about that, and since we'll _only_ use it on windows, probably wrap the whole method in an `#ifdef`

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -9,6 +9,11 @@
 #include "instrumented_assembly_generator/instrumented_assembly_generator_cor_profiler_info.h"
 #include "instrumented_assembly_generator/instrumented_assembly_generator_helper.h"
 
+#if defined(BUILD_WITH_WORKLOAD_SELECTION)
```
andrewlock:
```suggestion
#ifdef BUILD_WITH_WORKLOAD_SELECTION
```

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -316,7 +322,56 @@ namespace datadog::shared::nativeloader
...
+              }
+              else
+              {
+                  injection_status = InjectionStatus::DENY;
+              }
+
+              return std::nullopt;
+          });
+
+          if (const auto wls_file = fs::path{GetDatadogProgramDataFolderPath()} / "protected" / "workload_selection.fb"; fs::exists(wls_file))
```
andrewlock: Could we remove the compound statement here? People are generally less familiar with C++ in this repo, so we should favour explicitness over terseness I think 🙂

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -316,7 +322,56 @@ namespace datadog::shared::nativeloader
...
+          {
+            auto maybe_error = plcs::evaluate_buffer_from_file(wls_file);
+            if (maybe_error) {
+              Log::Error("CorProfiler::Initialize: An error occured while evaluating workload selection (reason: ", *maybe_error, ")");
+              return E_FAIL;
```
andrewlock, line 365: This is an interesting one. I agree bailing out is probably the correct option. IIRC `E_FAIL` means we write to the event log, which may or may not be what we want to do, given this implies that ever .NET process will write to the log, which could be an issue? 🤔

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -316,7 +322,56 @@ namespace datadog::shared::nativeloader
...
+            }
+
+            if (injection_status != InjectionStatus::ALLOW) {
+              Log::Info("CorProfiler::Initialize: Instrumentation denied due to workload selection.");
+              return CORPROF_E_PROFILER_CANCEL_ACTIVATION;
+            }
+          } else {
+            Log::Warn("CorProfiler::Initialize: Missing workload selection file.");
```
andrewlock: Won't this be normal and the default, at least initially? Should it really be a warning?

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -316,7 +322,56 @@ namespace datadog::shared::nativeloader
...
+          }
+        }
+        #endif
```
andrewlock: nit:
```suggestion
#endif
```

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
dmehala: Indeed, that's the next step.

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
dmehala: I am currently working on an upstream change.

andrewlock: I really meant the mechanism - callback? Logging to its own file?

dmehala: Got it. The library will expose an API to set a callback with a log level. WDYT?

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
andrewlock: Also, `IsRunningOnIIS` isn't exhaustive today - it's just checking if the process name is running _inside_ IIS. However, if the app is an out-of-process ASP.NET Core app (for example) it will still be hosted in IIS, and therefore should be instrumented, but it will currently return `false` here

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
dmehala: addressed in b9b268e85ec3689b49a2b5c8cd27b2f8e2d93b1e

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
dmehala: addressed in b9b268e85ec3689b49a2b5c8cd27b2f8e2d93b1e

shared/src/Datadog.Trace.ClrProfiler.Native/util.h
dmehala: addressed in b9b268e85ec3689b49a2b5c8cd27b2f8e2d93b1e

PR #7493, opened 2025-09-09
Decrease Nuke restore verbosity job

NachoEchevarria:
## Summary of changes

This PR decreases the verbosity of the restore job that currently displays thousand of lines in the logs, flooding the gitlab output and surpassing the limit, which hides later logs.

In windows, we use Nuget.exe. Locally, we still will get the same messages, but in CI, we will only get the messages sent to Stderr for errors and warnings.

In the case of Linux, we use dotnet restore. Setting it to a minimal log level would be enough to suppress the too detailed logs and keep the warnings and errors.

### DIFF
--- a/tracer/build/_build/Build.Steps.cs
+++ b/tracer/build/_build/Build.Steps.cs
@@ -275,15 +275,7 @@ partial class Build
                 NuGetTasks.NuGetRestore(s => s
                     .SetTargetPath(Solution)
                     .SetVerbosity(NuGetVerbosity.Normal)
-                    .SetProcessCustomLogger((type, text) =>
-                    {
-                        if (type == OutputType.Std && IsServerBuild && (!text.StartsWith("Restored", StringComparison.OrdinalIgnoreCase) && !text.Contains("Warning", StringComparison.OrdinalIgnoreCase)))
-                        {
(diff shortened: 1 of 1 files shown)


tracer/build/_build/Build.Steps.cs
```diff
@@ -274,15 +274,16 @@ bool RequiresThoroughTesting()
             {
                 NuGetTasks.NuGetRestore(s => s
                     .SetTargetPath(Solution)
-                    .SetVerbosity(NuGetVerbosity.Normal)
+                    .SetVerbosity(NuGetVerbosity.Normal)   // keep warnings/errors
+                    .SetProcessLogOutput(false)          // Drop stdout, keep stderr, no [DBG] messages
```
andrewlock: This seems to drop _all_ the logs in normal, which isn't a big deal, but it would be nice if it didn't drop them locally. Maybe we should tweak it to show logs locally using something like this? 🤔 
```suggestion
                    .SetVerbosity(NuGetVerbosity.Normal)   // keep warnings/errors
                    .SetProcessLogOutput(!IsServerBuild)          // Drop stdout, keep stderr, no [DBG] messages
```

PR #7494, opened 2025-09-09
[AAP] Update waf version to v1.28.0

daniel-romano-DD:
## Summary of changes
Update waf to v1.28.0

### DIFF
--- a/tracer/src/Datadog.Trace/AppSec/Security.cs
+++ b/tracer/src/Datadog.Trace/AppSec/Security.cs
@@ -348,9 +348,9 @@ namespace Datadog.Trace.AppSec
             {
                 information.TryGetValue("status_code", out var actionStatusCode);
 
-                if (actionStatusCode is int statusCodeInt)
+                if (actionStatusCode is ulong statusCodeInt)
                 {
-                    return statusCodeInt;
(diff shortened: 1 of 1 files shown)


tracer/src/Datadog.Trace/AppSec/Security.cs
```diff
@@ -348,7 +348,11 @@ int GetStatusCode(Dictionary<string, object?> information, int defaultValue)
             {
                 information.TryGetValue("status_code", out var actionStatusCode);
 
-                if (actionStatusCode is string statusCodeString && int.TryParse(statusCodeString, out var statusCode))
+                if (actionStatusCode is ulong statusCodeInt)
+                {
+                    return (int)statusCodeInt;
```
andrewlock, line 353: Is this cast definitely safe? i.e. it's typed as a `ulong` but will only ever be `0`-`int.MaxValue`?

dromanol: It's http status. I guess they modeled it as `ulong` instead of `ushort` for convenience reasons. Previously the status was an int as a string, and so was parsed (I left that part in the `else if` for compatibility).

PR #7496, opened 2025-09-09
Silently uncompress

NachoEchevarria:

### DIFF
--- a/tracer/build/_build/Build.Steps.cs
+++ b/tracer/build/_build/Build.Steps.cs
@@ -10,6 +10,7 @@ using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using CodeGenerators;
+using ICSharpCode.SharpZipLib.Zip;
 using LogParsing;
 using Mono.Cecil;
 using Nuke.Common;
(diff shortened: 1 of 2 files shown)


bouwkast (approved):

tracer/build/_build/Build.Steps.cs
```diff
@@ -2772,11 +2773,34 @@ private async Task DownloadAndExtractVcpkg(AbsolutePath destinationFolder)
...
+        using var zipFile = new ZipFile(fileStream);
+
+        var entries = zipFile.Cast<ZipEntry>().Where(x => !x.IsDirectory);
+        foreach (var entry in entries)
+        {
+            var file = PathConstruction.Combine(directory, entry.Name);
+            var path = Path.GetDirectoryName(file);
+
+            if (!Directory.Exists(path) && !string.IsNullOrEmpty(path))
+                Directory.CreateDirectory(path);
```
lucaspimentel: coding style: we always use `{}` even for single-line blocks
```suggestion
                {
                    Directory.CreateDirectory(path);
                }
```

NachoEchevarria:
> Does the job, but I wonder if we could instead add a log filter to fix cases like these to avoid the duplication🤔 Nuke just uses Serilog, so depending on how they grab the logger it might be easy to add a filter for the compression tasks logs in the entry point?

I tried, but if I add a filter in the constructor of Build, it initially works, but after writing the first line, the filter is lost and is overwritten by one created by Nuke. If I add it later, seems to work, but it does not really seem very clean or reliable.

NachoEchevarria:
Thanks for your feedback!

PR #7500, opened 2025-09-11
chore: refactor error handling in TraceExporter to catch all exceptions

ganeshnj:
## Summary of changes (Generated)

This pull request refactors error handling and response processing in the trace exporting pipeline to improve robustness and clarity. The main changes include switching from exception-based flow to null checks, enhancing logging for error cases, and updating error conversion utilities.

**Error handling improvements:**

* The `SendTracesAsync` method in `TraceExporter.cs` now checks for a null response instead of relying on exception handling, returning `false` if sending fails, and removes the outer try-catch block for non-exporter exceptions. [[1]](diffhunk://#diff-b0f42e84d63dcf674fd1e42d5e0fe5b68f536b7e572bb75b7e9fa2363aeb5a86L46-R50) [[2]](diffhunk://#diff-b0f42e84d63dcf674fd1e42d5e0fe5b68f536b7e572bb75b7e9fa2363aeb5a86L80-L85)
* The internal `Send` method now returns a nullable `TraceExporterResponse?` and logs errors directly without throwing exceptions for non-exporter errors; it returns `null` on failure. [[1]](diffhunk://#diff-b0f42e84d63dcf674fd1e42d5e0fe5b68f536b7e572bb75b7e9fa2363aeb5a86L107-R103) [[2]](diffhunk://#diff-b0f42e84d63dcf674fd1e42d5e0fe5b68f536b7e572bb75b7e9fa2363aeb5a86L117-L133)

**Error conversion and propagation:**

* The `TraceExporterErrorHandle` class replaces the `ToException()` method with `ToError()`, returning a `TraceExporterError` struct, and updates `ThrowIfError()` to use this new method for constructing exceptions.

**Response body processing:**

* The `TraceExporterResponse` class now wraps response body extraction in a try-catch block, logging any exceptions and returning `null` if reading the response body fails. [[1]](diffhunk://#diff-cea4c5186fbd4b8519500192b67c467f022a583c8606c1a8b558d1948fa20422R52-R53) [[2]](diffhunk://#diff-cea4c5186fbd4b8519500192b67c467f022a583c8606c1a8b558d1948fa20422L66-R75)

**Minor code improvements:**

* Adds a missing using directive for `Datadog.Trace.Vendors.MessagePack` in `TraceExporterResponse.cs`.

## Reason for change

We don't want to bubble any trace exporter exception to caller similar to `Api` implementation.

## Implementation details

- try-catch all.
- native method calls have their own try-catch
- some cleanup in `Send` method

### DIFF
--- a/tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporter.cs
+++ b/tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporter.cs
@@ -43,49 +43,41 @@ internal class TraceExporter : SafeHandle, IApi
     {
         _log.Debug<int>("Sending {Count} traces to the Datadog Agent.", numberOfTraces);
 
-        try
+        using var response = Send(traces, numberOfTraces);
+        if (response == null)
         {
-            using var response = Send(traces, numberOfTraces);
-            if (response == null)
-            {
-                return Task.FromResult(false);
-            }
-
-            if (response.IsInvalid)
-            {
-                _log.Warning("Traces sent successfully to the Agent, but the response is invalid");
-            }
-            else
-            {
-                var json = response.ReadAsString();
-
-                if (StringUtil.IsNullOrEmpty(json))
-                {
-                    _log.Warning("Traces sent successfully to the Agent, but the response is empty");
-                }
-                else if (json != _cachedResponse)
-                {
-                    try
-                    {
-                        var apiResponse = JsonConvert.DeserializeObject<ApiResponse>(json);
-                        _updateSampleRates(apiResponse.RateByService);
-                        _cachedResponse = json;
-                    }
-                    catch (Exception ex)
-                    {
-                        _log.Error(ex, "Traces sent successfully to the Agent, but an error occurred deserializing the response.");
-                    }
-                }
-            }
-
-            _log.Debug<int>("Successfully sent {Count} traces to the Datadog Agent.", numberOfTraces);
-            return Task.FromResult(true);
-        }
-        catch (Exception ex)
-        {
-            _log.Error(ex, "An error occurred while sending data to the agent.");
             return Task.FromResult(false);
         }
+
+        if (response.IsInvalid)
+        {
+            _log.Warning("Traces sent successfully to the Agent, but the response is invalid");
+        }
+        else
+        {
+            var json = response.ReadAsString();
+
+            if (StringUtil.IsNullOrEmpty(json))
+            {
+                _log.Warning("Traces sent successfully to the Agent, but the response is empty");
+            }
+            else if (json != _cachedResponse)
+            {
+                try
+                {
+                    var apiResponse = JsonConvert.DeserializeObject<ApiResponse>(json);
+                    _updateSampleRates(apiResponse.RateByService);
+                    _cachedResponse = json;
+                }
+                catch (Exception ex)
+                {
+                    _log.Error(ex, "Traces sent successfully to the Agent, but an error occurred deserializing the response.");
+                }
+            }
+        }
+
+        _log.Debug<int>("Successfully sent {Count} traces to the Datadog Agent.", numberOfTraces);
+        return Task.FromResult(true);
     }
 
     public Task<bool> SendStatsAsync(StatsBuffer stats, long bucketDuration)
--- a/tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporterResponse.cs
+++ b/tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporterResponse.cs
@@ -8,6 +8,7 @@
 using System;
 using System.Runtime.InteropServices;
 using Datadog.Trace.Logging;
+using Datadog.Trace.Vendors.MessagePack;
 
 namespace Datadog.Trace.LibDatadog.DataPipeline;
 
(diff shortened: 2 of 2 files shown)


tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporter.cs
```diff
@@ -77,7 +77,7 @@ public Task<bool> SendTracesAsync(ArraySegment<byte> traces, int numberOfTraces,
             _log.Debug<int>("Successfully sent {Count} traces to the Datadog Agent.", numberOfTraces);
             return Task.FromResult(true);
         }
-        catch (Exception ex) when (ex is not TraceExporterException)
+        catch (Exception ex)
```
andrewlock: This will still cause double-logging because we throw `error.ToException()` in the `Send` method. 

Is it possible to _not_ throw the exception in `Send()`. If we're just going to immediately catch it again anyway, and we don't want to log it, it seems like unnecessary expense? 🤔

ganeshnj: fair call, TraceExporterException is basically not serving any purpose with throwing. Technically we can remove this type, just keeping it as we have a convenience over logging exception.

tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporter.cs
```diff
@@ -122,10 +122,9 @@ private unsafe TraceExporterResponse Send(ArraySegment<byte> traces, int numberO
                 {
                     var ex = error.ToException();
                     _log.Error(ex, "An error occurred while sending data to the agent. Error Code: {ErrorCode}, Message: {Message}", ex.ErrorCode, ex.Message);
-                    throw ex;
```
andrewlock, line 125: This means we're going to return an invalid `TracerExporterResponse` I think, with a potentially zero `responsePtr` 🤔 It also means we _won't_ return `false` from `SendTracesAsync`. I think it may require a bigger refactoring unfortunately

ganeshnj: yeah I'm looking into it, likely deleting the excpetion type completely

ganeshnj: endeup keeping the exception type since we use it other places outside the `Send` method.

tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporter.cs
```diff
@@ -43,45 +43,41 @@ public Task<bool> SendTracesAsync(ArraySegment<byte> traces, int numberOfTraces,
     {
         _log.Debug<int>("Sending {Count} traces to the Datadog Agent.", numberOfTraces);
 
-        try
```
ganeshnj, line 46: we don't need this try-catch each smaller methods which can throw has their own

and if something still goes wrong, we want to know what's the problem using managed exception

tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporter.cs
```diff
@@ -43,45 +43,41 @@ public Task<bool> SendTracesAsync(ArraySegment<byte> traces, int numberOfTraces,
...
+        }
+
+        if (response.IsInvalid)
+        {
+            _log.Warning("Traces sent successfully to the Agent, but the response is invalid");
+        }
+        else
         {
-            using var response = Send(traces, numberOfTraces);
+            var json = response.ReadAsString();
```
andrewlock, line 58: I think this can technically throw? 🤔

tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporter.cs
```diff
@@ -114,23 +110,24 @@ private unsafe TraceExporterResponse Send(ArraySegment<byte> traces, int numberO
...
+                var responsePtr = IntPtr.Zero;
                 using var error = NativeInterop.Exporter.Send(this, traceSlice, (UIntPtr)numberOfTraces, ref responsePtr);
                 if (!error.IsInvalid)
                 {
-                    var ex = error.ToException();
-                    _log.Error(ex, "An error occurred while sending data to the agent. Error Code: {ErrorCode}, Message: {Message}", ex.ErrorCode, ex.Message);
-                    throw ex;
+                    var err = error.ToError();
+                    var message = Marshal.PtrToStringAnsi(err.Msg);
+                    _log.Error("An error occurred while sending data to the agent. Error Code: {ErrorCode}, Message: {Message}", err.Code, message);
```
andrewlock, line 121: If there's an error here, I think we need to return null, so that we return `false` in the callee? I think that's ok in terms of disposals, given we were previously throwing anyway...

```suggestion
                    _log.Error("An error occurred while sending data to the agent. Error Code: {ErrorCode}, Message: {Message}", err.Code, message);
                    return null;
```

tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporter.cs
ganeshnj: not really, I wrapped inside a try-catch, if you scroll to bottom and expand the diff.

https://github.com/DataDog/dd-trace-dotnet/blob/ganeshnj/chore/catch-all-ex-te/tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporterResponse.cs#L52-L75

tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporter.cs
ganeshnj: I was thinking the same but my hypothesis was, if we have an error, the response will be invalid (zero ptr) and we will return zero ptr.

that will be taken care by callee. basically an expected error response

andrewlock: AFAICT, with the current code:
- `Send()` logs and error but returns non-null
- `SendTracesAsync()` logs a warning because the response was invalid
- `SendTracesAsync()` then logs a debug message
- `SendTracesAsync()` returns `true`

So I definitely think we want that `null` 😅

PR #7501, opened 2025-09-11
Add additional compiler flags to allow `required`

andrewlock:
## Summary of changes

Adds a couple of compiler-required attributes for down level target frameworks

## Reason for change

These attributes are required for [the C#11 feature `required`](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/required). They're available by default in .NET 7+, but need to be explicitly provided for earlier TFMs, just like we do for other compiler attributes.

## Implementation details

Copy/paste the attributes as polyfills and `#if` them

## Test coverage

As long as we compile, we're fine. I've confirmed `required` is usable separately

## Other details

### DIFF
--- a/tracer/src/Datadog.Trace/Util/System.Runtime.CompilerServices.Attributes.cs
+++ b/tracer/src/Datadog.Trace/Util/System.Runtime.CompilerServices.Attributes.cs
@@ -48,6 +48,8 @@ internal sealed class SkipLocalsInitAttribute : Attribute
 {
 }
 
+#endif
+
 #if !NET7_0_OR_GREATER
 /// <summary>
 /// Specifies that a type has required members or that a member is required.
@@ -75,5 +77,4 @@ internal sealed class CompilerFeatureRequiredAttribute : Attribute
 
     public bool IsOptional { get; init; }
 }
-#endif
 #endif


lucaspimentel:
@codex review

tracer/src/Datadog.Trace/Util/System.Runtime.CompilerServices.Attributes.cs
```diff
@@ -48,4 +48,32 @@ internal sealed class SkipLocalsInitAttribute : Attribute
...
+
+    public CompilerFeatureRequiredAttribute(string featureName)
+    {
+        FeatureName = featureName;
+    }
+
+    public string FeatureName { get; }
+
+    public bool IsOptional { get; init; }
+}
```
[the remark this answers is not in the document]
lucaspimentel: I think the bot is correct here. The new code is nested inside an existing `#if !NET5_0_OR_GREATER`.

tracer/src/Datadog.Trace/Util/System.Runtime.CompilerServices.Attributes.cs
```diff
@@ -48,4 +48,32 @@ internal sealed class SkipLocalsInitAttribute : Attribute
...
+#endif
```
andrewlock, line 80:
```suggestion
```

tracer/src/Datadog.Trace/Util/System.Runtime.CompilerServices.Attributes.cs
```diff
@@ -48,4 +48,32 @@ internal sealed class SkipLocalsInitAttribute : Attribute
 {
 }
 
+#if !NET7_0_OR_GREATER
```
andrewlock, line 53:
```suggestion
#endif

#if !NET7_0_OR_GREATER
```

PR #7510, opened 2025-09-12
Allow passing fallback values to `IntegrationSettingsCollection`

andrewlock:
## Summary of changes

Allow passing an existing `IntegrationSettingsCollection` in to use as "fallback" values

## Reason for change

Needed functionality as part of a general config refactor (coming in subsequent PRs). We need to be able to find a configuration source to look in, but if the values aren't there, use the existing collection. 

## Implementation details

Allow optionally passing in an `IntegrationSettingsCollection` when creating a new `IntegrationSettingsCollection` to use as the fallback values

## Test coverage

Added a unit test for the behaviour, but it's otherwise not used yet.

## Other details

### DIFF
--- a/tracer/test/Datadog.Trace.Tests/Configuration/IntegrationSettingsCollectionTests.cs
+++ b/tracer/test/Datadog.Trace.Tests/Configuration/IntegrationSettingsCollectionTests.cs
@@ -103,7 +103,7 @@ namespace Datadog.Trace.Tests.Configuration
         {
             var originalSource = new NameValueConfigurationSource(new NameValueCollection
             {
-                { "DD_TRACE_Kafka_ENABLED", "false" },
+                { "DD_TRACE_Kafka_ENABLED", "true" },
                 { "DD_TRACE_Kafka_ANALYTICS_ENABLED", "true" },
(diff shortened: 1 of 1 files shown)


tracer/test/Datadog.Trace.Tests/Configuration/IntegrationSettingsCollectionTests.cs
```diff
@@ -97,5 +97,31 @@ public void PopulatesFromBuilderCorrectly()
...
+            kafka.AnalyticsEnabled.Should().BeTrue();
+            kafka.AnalyticsSampleRate.Should().Be(0.2);
+
+            var updated = new IntegrationSettingsCollection(NullConfigurationSource.Instance, [], original);
+
+            var kafka2 = updated[IntegrationId.Kafka];
+            kafka.Enabled.Should().BeTrue();
+            kafka.AnalyticsEnabled.Should().BeTrue();
+            kafka.AnalyticsSampleRate.Should().Be(0.2);
+
```
andrewlock:
```suggestion
```

tonyredondo (approved):

PR #7511, opened 2025-09-13
[OTEL] Metrics API Support - Collection Implementation

link04:
## Summary of changes
This PR implements the metrics collection foundation for OpenTelemetry Metrics API supportKey additions include:

- MetricReader: Component for collecting instrument measurements and aggregating them
- MetricReaderHandler: High-performance measurement processing with proper aggregation (Sum, Last Value, Histogram)
- MetricPoint: Thread-safe data structure for metric aggregation with histogram bucket support
- Testing: Unit test covering all 7 instrument types (4 sync + 3 async) with exact OTLP value verification

This is the second of a series of 3 PRs to complete the OpenTelemetry Metrics API support.
1. [[OTEL] Metrics API Support - Configurations & Telemetry #7420](https://github.com/DataDog/dd-trace-dotnet/pull/7420)
2. [[OTEL] Metrics API Support -  Collection Implementation #7511](https://github.com/DataDog/dd-trace-dotnet/pull/7511)  <- You are here.
3. [[OTEL] Metrics API Support -  Metrics Reader & OTLP Exporter #7514](https://github.com/DataDog/dd-trace-dotnet/pull/7514)

## Reason for change
The [RFC] OpenTelemetry Metrics API support requires implementing metrics collection that's compatible with the OpenTelemetry data model. This PR provides the collection foundation that will enable customers to use .NET's `System.Diagnostics.Metrics` API with Datadog as a drop-in replacement for the OpenTelemetry SDK.

Business Impact: Enables the collection phase of OpenTelemetry Metrics API support, laying the groundwork for OTLP export in the next PR.

## Implementation details
✅ Metrics Collection Components:
- MetricReader - RFC-compliant component for initialization and async collection
- MetricReaderHandler - Measurement processing and aggregation
- MetricPoint - Aggregation data structure with histogram bucket support

✅ Aggregation Mapping:
- Counter/Asynchronous Counter → Sum Aggregation → Sum Metric Point ✅
- UpDownCounter/Asynchronous UpDownCounter → Sum Aggregation → Sum Metric Point ✅
- Gauge/Asynchronous Gauge → Last Value Aggregation → Gauge Metric Point ✅
- Histogram → Explicit Bucket Histogram Aggregation → Histogram Metric Point ✅

## Test coverage
Added unit test `MeterListenerCapturesAllMetrics()` that verifies:
- All 7 instrument types (Counter, Histogram, Gauge, UpDownCounter + 3 Async variants)
- Value verification matching OTLP snapshot expectations
Histogram bucket logic verification
- Sync + Async instrument collection
- Tag capture and temporality settings

## Other details
Jira: [APMAPI-1572](https://datadoghq.atlassian.net/browse/APMAPI-1572)

[APMAPI-1572]: https://datadoghq.atlassian.net/browse/APMAPI-1572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tracer/missing-nullability-files.csv
+++ b/tracer/missing-nullability-files.csv
@@ -137,7 +137,6 @@ src/Datadog.Trace/HttpOverStreams/HttpResponse.cs
 src/Datadog.Trace/HttpOverStreams/IHttpContent.cs
 src/Datadog.Trace/Iast/ITaintedMap.cs
 src/Datadog.Trace/Iast/SourceType.cs
-src/Datadog.Trace/OTelMetrics/AggregationTemporality.cs
 src/Datadog.Trace/PlatformHelpers/AspNetCoreHttpRequestHandler.cs
 src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
 src/Datadog.Trace/PlatformHelpers/ServiceFabric.cs
--- a/tracer/src/Datadog.Trace/OTelMetrics/AggregationTemporality.cs
+++ b/tracer/src/Datadog.Trace/OTelMetrics/AggregationTemporality.cs
@@ -5,22 +5,23 @@
 
 #if NET6_0_OR_GREATER
 
-namespace Datadog.Trace.OTelMetrics
+#nullable enable
+
+namespace Datadog.Trace.OTelMetrics;
+
+/// <summary>
+/// Represents the aggregation temporality of a metric.
+/// </summary>
+public enum AggregationTemporality
 {
     /// <summary>
-    /// Represents the aggregation temporality of a metric.
+    /// Delta temporality, representing changes since the last measurement.
     /// </summary>
-    public enum AggregationTemporality
-    {
-        /// <summary>
-        /// Delta temporality, representing changes since the last measurement.
-        /// </summary>
-        Delta = 0,
+    Delta = 0,
 
-        /// <summary>
-        /// Cumulative temporality, representing the total value since the start.
-        /// </summary>
-        Cumulative = 1,
-    }
+    /// <summary>
+    /// Cumulative temporality, representing the total value since the start.
+    /// </summary>
+    Cumulative = 1,
 }
 #endif
--- a/tracer/src/Datadog.Trace/OTelMetrics/MetricPoint.cs
+++ b/tracer/src/Datadog.Trace/OTelMetrics/MetricPoint.cs
@@ -9,98 +9,82 @@
 
 using System;
 using System.Collections.Generic;
-using System.Runtime.CompilerServices;
 using System.Threading;
 
-namespace Datadog.Trace.OTelMetrics
+namespace Datadog.Trace.OTelMetrics;
+
+internal class MetricPoint(string instrumentName, string meterName, InstrumentType instrumentType, AggregationTemporality? temporality, Dictionary<string, object?> tags)
 {
-    internal class MetricPoint(string instrumentName, string meterName, InstrumentType instrumentType, AggregationTemporality? temporality, Dictionary<string, object?> tags)
+    internal static readonly double[] DefaultHistogramBounds = [0, 5, 10, 25, 50, 75, 100, 250, 500, 750, 1000, 2500, 5000, 7500, 10000];
+    private readonly long[] _runningBucketCounts = instrumentType == InstrumentType.Histogram ? new long[DefaultHistogramBounds.Length + 1] : [];
+    private readonly object _histogramLock = new();
+    private long _runningCountValue;
+    private double _runningDoubleValue;
+    private double _runningMin = double.PositiveInfinity;
+    private double _runningMax = double.NegativeInfinity;
+
+    public string InstrumentName { get; } = instrumentName;
+
+    public string MeterName { get; } = meterName;
+
+    public InstrumentType InstrumentType { get; } = instrumentType;
+
+    public AggregationTemporality? AggregationTemporality { get; } = temporality;
+
+    public Dictionary<string, object?> Tags { get; } = tags;
+
+    internal long RunningCount => _runningCountValue;
+
+    internal double RunningSum => _runningDoubleValue;
+
+    internal double RunningMin => _runningMin;
+
+    internal double RunningMax => _runningMax;
+
+    internal long[] RunningBucketCounts => _runningBucketCounts;
+
+    internal void UpdateCounter(double value)
     {
-        // Static fields first
-        internal static readonly double[] DefaultHistogramBounds = [0, 5, 10, 25, 50, 75, 100, 250, 500, 750, 1000, 2500, 5000, 7500, 10000];
-
-        // Instance fields
-        private readonly long[] _runningBucketCounts = instrumentType == InstrumentType.Histogram ? new long[DefaultHistogramBounds.Length + 1] : [];
-        private readonly object _histogramLock = new();
-        private long _runningCountValue;     // For counters and histogram count
-        private double _runningDoubleValue;  // For gauges and histogram sum
-        private double _runningMin = double.PositiveInfinity;
-        private double _runningMax = double.NegativeInfinity;
-
-        // Constructor
-
-        // Public properties
-        public string InstrumentName { get; } = instrumentName;
-
-        public string MeterName { get; } = meterName;
-
-        public InstrumentType InstrumentType { get; } = instrumentType;
-
-        public AggregationTemporality? AggregationTemporality { get; } = temporality;
-
-        public Dictionary<string, object?> Tags { get; } = tags;
-
-        // Internal properties for collection access
-        internal long RunningCount => _runningCountValue;
-
-        internal double RunningSum => _runningDoubleValue;
-
-        internal double RunningMin => _runningMin;
-
-        internal double RunningMax => _runningMax;
-
-        internal long[] RunningBucketCounts => _runningBucketCounts;
-
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        internal void UpdateCounter(double value)
+        lock (_histogramLock)
         {
-            // Use lock to avoid floating-point precision issues with CompareExchange
-            lock (_histogramLock)
+            _runningDoubleValue += value;
+        }
+    }
+
+    internal void UpdateGauge(double value)
+    {
+        Interlocked.Exchange(ref _runningDoubleValue, value);
+    }
+
+    internal void UpdateHistogram(double value)
+    {
+        var bucketIndex = FindBucketIndex(value);
+
+        lock (_histogramLock)
+        {
+            unchecked
             {
+                _runningCountValue++;
                 _runningDoubleValue += value;
+                _runningBucketCounts[bucketIndex]++;
             }
+
+            _runningMin = Math.Min(_runningMin, value);
+            _runningMax = Math.Max(_runningMax, value);
         }
+    }
 
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        internal void UpdateGauge(double value)
+    private static int FindBucketIndex(double value)
+    {
+        for (var i = 0; i < DefaultHistogramBounds.Length; i++)
         {
-            Interlocked.Exchange(ref _runningDoubleValue, value);
-        }
-
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        internal void UpdateHistogram(double value)
-        {
-            // Find bucket index first (outside lock for performance)
-            var bucketIndex = FindBucketIndex(value);
-
-            lock (_histogramLock)
+            if (value <= DefaultHistogramBounds[i])
             {
-                unchecked
-                {
-                    _runningCountValue++;
-                    _runningDoubleValue += value; // Sum
-                    _runningBucketCounts[bucketIndex]++;
-                }
-
-                _runningMin = Math.Min(_runningMin, value);
-                _runningMax = Math.Max(_runningMax, value);
+                return i;
             }
         }
 
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private static int FindBucketIndex(double value)
-        {
-            // Linear search for default 15 boundaries (fast enough)
-            for (var i = 0; i < DefaultHistogramBounds.Length; i++)
-            {
-                if (value <= DefaultHistogramBounds[i])
-                {
-                    return i;
-                }
-            }
-
-            return DefaultHistogramBounds.Length; // Overflow bucket
-        }
+        return DefaultHistogramBounds.Length;
     }
 }
 #endif
--- a/tracer/src/Datadog.Trace/OTelMetrics/MetricReader.cs
+++ b/tracer/src/Datadog.Trace/OTelMetrics/MetricReader.cs
@@ -11,67 +11,75 @@ using System;
 using System.Threading;
 using Datadog.Trace.Logging;
 
-namespace Datadog.Trace.OTelMetrics
+namespace Datadog.Trace.OTelMetrics;
+
+internal static class MetricReader
 {
-    internal static class MetricReader
+    private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(MetricReader));
+
+    private static System.Diagnostics.Metrics.MeterListener? _meterListenerInstance;
+    private static int _initialized;
+    private static int _stopped;
+
+    public static bool IsRunning =>
+        Interlocked.CompareExchange(ref _initialized, 1, 1) == 1 &&
+        Interlocked.CompareExchange(ref _stopped, 0, 0) == 0;
+
+    public static void Initialize()
     {
-        private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(MetricReader));
-
-        private static System.Diagnostics.Metrics.MeterListener? _meterListenerInstance;
-        private static int _initialized;
-        private static int _stopped;
-
-        public static bool IsRunning =>
-            Interlocked.CompareExchange(ref _initialized, 1, 1) == 1 &&
-            Interlocked.CompareExchange(ref _stopped, 0, 0) == 0;
-
-        public static void Initialize()
+        if (Interlocked.CompareExchange(ref _initialized, 1, 0) == 1)
         {
-            if (Interlocked.CompareExchange(ref _initialized, 1, 0) == 1)
-            {
-                return;
-            }
-
-            var meterListener = new System.Diagnostics.Metrics.MeterListener();
-            meterListener.InstrumentPublished = MetricReaderHandler.OnInstrumentPublished;
-
-            meterListener.SetMeasurementEventCallback<byte>(MetricReaderHandler.OnMeasurementRecordedByte);
-            meterListener.SetMeasurementEventCallback<short>(MetricReaderHandler.OnMeasurementRecordedShort);
-            meterListener.SetMeasurementEventCallback<int>(MetricReaderHandler.OnMeasurementRecordedInt);
-            meterListener.SetMeasurementEventCallback<long>(MetricReaderHandler.OnMeasurementRecordedLong);
-            meterListener.SetMeasurementEventCallback<float>(MetricReaderHandler.OnMeasurementRecordedFloat);
-            meterListener.SetMeasurementEventCallback<double>(MetricReaderHandler.OnMeasurementRecordedDouble);
-
-            meterListener.Start();
-            _meterListenerInstance = meterListener;
-
-            Log.Debug("MeterListener initialized successfully.");
+            return;
         }
 
-        public static void Stop()
-        {
-            if (_meterListenerInstance is IDisposable disposableListener)
-            {
-                _meterListenerInstance = null;
-                disposableListener.Dispose();
-                Interlocked.Exchange(ref _stopped, 1);
-                Interlocked.Exchange(ref _initialized, 0);
-                Log.Debug("MeterListener stopped.");
-            }
-        }
+        var meterListener = new System.Diagnostics.Metrics.MeterListener();
 
-        internal static void CollectObservableInstruments()
+#if NET6_0 || NET7_0 || NET8_0
+        // Ensures instruments are fully de-registered on Dispose() for 6–8
+        // Static lambda => no captures/allocations
+        meterListener.MeasurementsCompleted = static (_, __) => { };
+#endif
+
+        meterListener.InstrumentPublished = MetricReaderHandler.OnInstrumentPublished;
+
+        meterListener.SetMeasurementEventCallback<byte>(MetricReaderHandler.OnMeasurementRecordedByte);
+        meterListener.SetMeasurementEventCallback<short>(MetricReaderHandler.OnMeasurementRecordedShort);
+        meterListener.SetMeasurementEventCallback<int>(MetricReaderHandler.OnMeasurementRecordedInt);
+        meterListener.SetMeasurementEventCallback<long>(MetricReaderHandler.OnMeasurementRecordedLong);
+        meterListener.SetMeasurementEventCallback<float>(MetricReaderHandler.OnMeasurementRecordedFloat);
+        meterListener.SetMeasurementEventCallback<double>(MetricReaderHandler.OnMeasurementRecordedDouble);
+
+        meterListener.Start();
+
+        Interlocked.Exchange(ref _meterListenerInstance, meterListener);
+        Interlocked.Exchange(ref _stopped, 0);
+
+        Log.Debug("MeterListener initialized successfully.");
+    }
+
+    public static void Stop()
+    {
+        var listener = Interlocked.Exchange(ref _meterListenerInstance, null);
+        if (listener is IDisposable disposableListener)
         {
-            if (_meterListenerInstance != null)
+            disposableListener.Dispose();
+            Interlocked.Exchange(ref _stopped, 1);
+            Interlocked.Exchange(ref _initialized, 0);
+            Log.Debug("MeterListener stopped.");
+        }
+    }
+
+    internal static void CollectObservableInstruments()
+    {
+        if (_meterListenerInstance != null)
+        {
+            try
             {
-                try
-                {
-                    _meterListenerInstance.RecordObservableInstruments();
-                }
-                catch (Exception ex)
-                {
-                    Log.Warning(ex, "Error collecting observable instruments.");
-                }
+                _meterListenerInstance.RecordObservableInstruments();
+            }
+            catch (Exception ex)
+            {
+                Log.Warning(ex, "Error collecting observable instruments.");
             }
         }
     }
--- a/tracer/src/Datadog.Trace/OTelMetrics/MetricReaderHandler.cs
+++ b/tracer/src/Datadog.Trace/OTelMetrics/MetricReaderHandler.cs
@@ -15,325 +15,257 @@ using System.Runtime.CompilerServices;
 using System.Threading;
 using Datadog.Trace.Logging;
 
-namespace Datadog.Trace.OTelMetrics
+namespace Datadog.Trace.OTelMetrics;
+
+internal static class MetricReaderHandler
 {
-    internal static class MetricReaderHandler
+    private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(MetricReaderHandler));
+
+    private static readonly ConcurrentDictionary<MetricStreamIdentity, MetricState> CapturedMetrics = new();
+    private static readonly HashSet<string> MetricStreamNames = new(StringComparer.OrdinalIgnoreCase);
+
+    private static int _metricCount;
+
+    public static void OnInstrumentPublished(Instrument instrument, MeterListener listener)
     {
-        private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(MetricReaderHandler));
+        var meterName = instrument.Meter.Name;
 
-        // Temporary storage for aggregation (like RuntimeMetrics._exceptionCounts)
-        // Will be cleared after each export to prevent memory leaks
-        private static readonly ConcurrentDictionary<MetricStreamIdentity, MetricState> CapturedMetrics = new();
-        private static readonly HashSet<string> MetricStreamNames = new(StringComparer.OrdinalIgnoreCase);
-
-        private static int _metricCount;
-
-        public static void OnInstrumentPublished(Instrument instrument, MeterListener listener)
+        if (!IsValidInstrumentName(instrument.Name))
         {
-            var meterName = instrument.Meter.Name;
-
-            // RFC requirement: Validate instrument name syntax
-            if (!IsValidInstrumentName(instrument.Name))
-            {
-                Log.Warning("Invalid instrument name '{InstrumentName}' from meter '{MeterName}'. Instrument names must be ASCII, start with letter, contain only alphanumeric characters, '_', '.', '/', '-' and be max 255 characters.", instrument.Name, meterName);
-                return; // Skip this instrument
-            }
-
-            bool shouldEnable;
-            var enabledMeterNames = Tracer.Instance.Settings.OpenTelemetryMeterNames;
-
-            if (enabledMeterNames.Length > 0)
-            {
-                shouldEnable = enabledMeterNames.Contains(meterName);
-            }
-            else
-            {
-                shouldEnable = !meterName.StartsWith("System.", StringComparison.Ordinal);
-            }
-
-            if (shouldEnable)
-            {
-                // Create state object for this instrument to avoid ConcurrentDictionary contention
-                var state = CreateMetricState(instrument);
-                if (state != null)
-                {
-                    listener.EnableMeasurementEvents(instrument, state);
-                    Log.Debug("Enabled measurement events for instrument: {InstrumentName} from meter: {MeterName}", instrument.Name, meterName);
-                }
-                else
-                {
-                    Log.Warning("Failed to create MetricState for instrument '{InstrumentName}' from meter '{MeterName}'. Instrument will be ignored.", instrument.Name, meterName);
-                }
-            }
+            Log.Warning("Invalid instrument name '{InstrumentName}' from meter '{MeterName}'. Instrument names must be ASCII, start with letter, contain only alphanumeric characters, '_', '.', '/', '-' and be max 255 characters.", instrument.Name, meterName);
+            return;
         }
 
-        public static void OnMeasurementRecordedLong(Instrument instrument, long value, ReadOnlySpan<KeyValuePair<string, object?>> tags, object? state)
+        bool shouldEnable;
+        var enabledMeterNames = Tracer.Instance.Settings.OpenTelemetryMeterNames;
+
+        if (enabledMeterNames.Length > 0)
         {
-            if (state is MetricState metricState)
-            {
-                metricState.RecordMeasurementLong(value, tags);
-            }
-            else
-            {
-                // Log error - this should not happen if we follow OTel SDK pattern
-                Log.Warning("Measurement recorded for instrument '{InstrumentName}' but no MetricState provided. Measurement will be dropped.", instrument.Name);
-            }
+            shouldEnable = enabledMeterNames.Contains(meterName);
+        }
+        else
+        {
+            shouldEnable = !meterName.StartsWith("System.", StringComparison.Ordinal);
         }
 
-        public static void OnMeasurementRecordedDouble(Instrument instrument, double value, ReadOnlySpan<KeyValuePair<string, object?>> tags, object? state)
+        if (shouldEnable)
         {
-            if (state is MetricState metricState)
+            var state = CreateMetricState(instrument);
+            if (state != null)
             {
-                metricState.RecordMeasurementDouble(value, tags);
-            }
-            else
-            {
-                // Log error - this should not happen if we follow OTel SDK pattern
-                Log.Warning("Measurement recorded for instrument '{InstrumentName}' but no MetricState provided. Measurement will be dropped.", instrument.Name);
+                listener.EnableMeasurementEvents(instrument, state);
+                Log.Debug("Enabled measurement events for instrument: {InstrumentName} from meter: {MeterName}", instrument.Name, meterName);
             }
         }
+    }
 
-        // Overloads for different numeric types
-        public static void OnMeasurementRecordedByte(Instrument instrument, byte value, ReadOnlySpan<KeyValuePair<string, object?>> tags, object? state)
+    public static void OnMeasurementRecordedLong(Instrument instrument, long value, ReadOnlySpan<KeyValuePair<string, object?>> tags, object? state)
+    {
+        if (state is MetricState metricState)
         {
-            OnMeasurementRecordedLong(instrument, value, tags, state);
+            metricState.RecordMeasurementLong(value, tags);
+        }
+    }
+
+    public static void OnMeasurementRecordedDouble(Instrument instrument, double value, ReadOnlySpan<KeyValuePair<string, object?>> tags, object? state)
+    {
+        if (state is MetricState metricState)
+        {
+            metricState.RecordMeasurementDouble(value, tags);
+        }
+    }
+
+    public static void OnMeasurementRecordedByte(Instrument instrument, byte value, ReadOnlySpan<KeyValuePair<string, object?>> tags, object? state)
+    {
+        OnMeasurementRecordedLong(instrument, value, tags, state);
+    }
+
+    public static void OnMeasurementRecordedShort(Instrument instrument, short value, ReadOnlySpan<KeyValuePair<string, object?>> tags, object? state)
+    {
+        OnMeasurementRecordedLong(instrument, value, tags, state);
+    }
+
+    public static void OnMeasurementRecordedInt(Instrument instrument, int value, ReadOnlySpan<KeyValuePair<string, object?>> tags, object? state)
+    {
+        OnMeasurementRecordedLong(instrument, value, tags, state);
+    }
+
+    public static void OnMeasurementRecordedFloat(Instrument instrument, float value, ReadOnlySpan<KeyValuePair<string, object?>> tags, object? state)
+    {
+        OnMeasurementRecordedDouble(instrument, value, tags, state);
+    }
+
+    internal static IReadOnlyList<MetricPoint> GetMetricsForExport(bool clearAfterGet = false)
+    {
+        var allMetricPoints = new List<MetricPoint>();
+
+        foreach (var kvp in CapturedMetrics)
+        {
+            var metricState = kvp.Value;
+            var metricPoints = metricState.GetMetricPoints();
+            allMetricPoints.AddRange(metricPoints);
         }
 
-        public static void OnMeasurementRecordedShort(Instrument instrument, short value, ReadOnlySpan<KeyValuePair<string, object?>> tags, object? state)
-        {
-            OnMeasurementRecordedLong(instrument, value, tags, state);
-        }
-
-        public static void OnMeasurementRecordedInt(Instrument instrument, int value, ReadOnlySpan<KeyValuePair<string, object?>> tags, object? state)
-        {
-            OnMeasurementRecordedLong(instrument, value, tags, state);
-        }
-
-        public static void OnMeasurementRecordedFloat(Instrument instrument, float value, ReadOnlySpan<KeyValuePair<string, object?>> tags, object? state)
-        {
-            OnMeasurementRecordedDouble(instrument, value, tags, state);
-        }
-
-        internal static IReadOnlyList<MetricPoint> GetMetricsForExport(bool clearAfterGet = false)
-        {
-            var allMetricPoints = new List<MetricPoint>();
-
-            foreach (var kvp in CapturedMetrics)
-            {
-                var metricState = kvp.Value;
-                var metricPoints = metricState.GetMetricPoints();
-                allMetricPoints.AddRange(metricPoints);
-            }
-
-            if (clearAfterGet)
-            {
-                CapturedMetrics.Clear();
-            }
-
-            return allMetricPoints;
-        }
-
-        // For testing only
-        internal static IReadOnlyList<MetricPoint> GetCapturedMetricsForTesting()
-        {
-            var allMetricPoints = new List<MetricPoint>();
-
-            foreach (var kvp in CapturedMetrics)
-            {
-                var metricState = kvp.Value;
-                var metricPoints = metricState.GetMetricPoints();
-                allMetricPoints.AddRange(metricPoints);
-            }
-
-            return allMetricPoints;
-        }
-
-        // For testing only - reset all captured metrics
-        internal static void ResetForTesting()
+        if (clearAfterGet)
         {
             CapturedMetrics.Clear();
-            MetricStreamNames.Clear();
-            Interlocked.Exchange(ref _metricCount, 0);
         }
 
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private static bool IsValidInstrumentName(string name)
+        return allMetricPoints;
+    }
+
+    // For testing only to be deleted in next PR
+    internal static IReadOnlyList<MetricPoint> GetCapturedMetricsForTesting()
+     => GetMetricsForExport(clearAfterGet: false);
+
+    // For testing only - reset all captured metrics
+    internal static void ResetForTesting()
+    {
+        CapturedMetrics.Clear();
+        MetricStreamNames.Clear();
+        Interlocked.Exchange(ref _metricCount, 0);
+    }
+
+    private static bool IsValidInstrumentName(string name)
+    {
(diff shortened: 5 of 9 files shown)


tracer/src/Datadog.Trace/OTelMetrics/MetricReader.cs
```diff
@@ -0,0 +1,75 @@
...
+            Interlocked.CompareExchange(ref _stopped, 0, 0) == 0;
+
+        public static void Initialize()
+        {
+            if (Interlocked.CompareExchange(ref _initialized, 1, 0) == 1)
+            {
+                return;
+            }
+
+            var meterListener = new System.Diagnostics.Metrics.MeterListener();
```
zacharycmontoya: This is interesting to directly reference the MeterListener since we can target the .NET 6.0 target framework. I think this decision is fine for this initial set of PR's so that we can develop this feature with first-class support for .NET 6.0+ where the API's are shipped in the runtime, but we'll want to revisit this later and do reflection/duck-typing so that applications on older runtimes that use the API's via the `System.Diagnostics.DiagnosticSource` NuGet package can also emit metrics.

tracer/src/Datadog.Trace/OTelMetrics/MetricReader.cs
```diff
@@ -0,0 +1,75 @@
...
+            meterListener.InstrumentPublished = MetricReaderHandler.OnInstrumentPublished;
+
+            meterListener.SetMeasurementEventCallback<long>(MetricReaderHandler.OnMeasurementRecordedLong);
```
zacharycmontoya: You should also register the callbacks for the other primitive types like the OTel SDK does. That way if, for example, the user wanted to create a counter with a `short` data type, it would still get mapped our Long callback: https://github.com/open-telemetry/opentelemetry-dotnet/blob/616e2730ec9f21da028a49d8e1e2fca6d81c43e3/src/OpenTelemetry/Metrics/MeterProviderSdk.cs#L173-L181

tracer/src/Datadog.Trace/OTelMetrics/MetricReaderHandler.cs
```diff
@@ -0,0 +1,202 @@
...
+
+            // RFC requirement: Validate negative values for Counter and Histogram
+            if ((aggregationType == "Counter" || aggregationType == "Histogram") && doubleValue < 0)
+            {
+                Log.Warning("Ignoring negative value {Value} for {InstrumentType} instrument: {InstrumentName}. API usage is incorrect.", doubleValue, aggregationType, instrument.Name);
+                return;
+            }
+
+            var meterName = instrument.Meter.Name;
+            var key = string.Concat(meterName, ".", instrument.Name);
```
zacharycmontoya: We need to make sure that we have unique metric points based on unique values for all of the following:
- Instrument name
- Instrument kind
- Instrument unit
- Instrument description

Can you make sure they are all reflected in the key?

tracer/src/Datadog.Trace/OTelMetrics/MetricPoint.cs
```diff
@@ -0,0 +1,108 @@
...
+        private double _runningMax = double.NegativeInfinity;
+
+        // Constructor
+
+        // Public properties
+        public string InstrumentName { get; } = instrumentName;
+
+        public string MeterName { get; } = meterName;
+
+        public string InstrumentType { get; } = instrumentType;
```
zacharycmontoya: Can you define and use a new enum to represent the `InstrumentType` instead of using a string? We only have a fixed set of acceptable values so an enum will be easier to enforce, and it's a valuetype

tracer/src/Datadog.Trace/OTelMetrics/MetricReaderHandler.cs
```diff
@@ -0,0 +1,202 @@
...
+                case "Gauge":
+                    metricPoint.UpdateGauge(value);
+                    break;
+                case "Histogram":
+                    metricPoint.UpdateHistogram(value);
+                    break;
+            }
+        }
+
+        internal static void TriggerAsyncCollection()
```
zacharycmontoya: I don't think `TriggerAsyncCollection` is used in this PR or the next one, since you call `MetricReader.CollectObservableInstruments` in your test code. You can delete this

tracer/src/Datadog.Trace/OTelMetrics/MetricReaderHandler.cs
```diff
@@ -0,0 +1,202 @@
...
+            return metrics;
+        }
+
+        // For testing only
+        internal static IReadOnlyDictionary<string, MetricPoint> GetCapturedMetricsForTesting() => CapturedMetrics;
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static string GetTemporalityString()
+        {
+            return _cachedTemporality ??= Tracer.Instance.Settings.OtlpMetricsTemporalityPreference.ToString();
```
zacharycmontoya: I think your interpretation of `temporality` isn't 100% correct. The user specifies a "TemporalityPreference" as input, and then based on that, each Instrument will get its own `temporality` based on a predefined mapping.

### CumulativePreference
For a TemporalityPreference of CumulativePreference, then all instruments get the `Cumulative` temporality (except for Gauge where there's no such thing as temporality).

### DeltaPreference
For a TemporalityPreference of DeltaPreference, the mapping is the following:
```markdown
| Instrument Kind          | Temporality |
| -------------------------| ----------- |
| Counter                  | Delta       |
| Observable Counter       | Delta       |
| UpDownCounter            | Cumulative  |
| Observable UpDownCounter | Cumulative  |
| Histogram                | Delta       |
| Gauge                    | N/A         |
```

### LowMemoryPreference
For a TemporalityPreference of LowMemoryPreference, the mapping is the following:
```markdown
| Instrument Kind          | Temporality |
| -------------------------| ----------- |
| Counter                  | Delta       |
| Observable Counter       | Cumulative  |
| UpDownCounter            | Cumulative  |
| Observable UpDownCounter | Cumulative  |
| Histogram                | Delta       |
| Gauge                    | N/A         |
```

tracer/src/Datadog.Trace/OTelMetrics/MetricReaderHandler.cs
```diff
@@ -0,0 +1,202 @@
...
+                return;
+            }
+
+            var meterName = instrument.Meter.Name;
+            var key = string.Concat(meterName, ".", instrument.Name);
+            var temporality = GetTemporalityString();
+
+            var tagsDict = tags.Length == 0 ? EmptyTags : CreateTagsDictionary(tags);
+
+            CapturedMetrics.AddOrUpdate(
```
zacharycmontoya: I think this is going to be too slow because theoretically you could have many different threads trying to increment a counter and then they will all wait here to access the ConcurrentDictionary, even if they were all waiting to access a different metric. For correctness, I think that this is OK so while I have a suggestion to make this work, I think that any changes to this should be done in a _single_ commit where you know before the change and after the change all of your tests are passing.

For the actionable suggestion, I think you should follow the example done by the .NET OpenTelemetry SDK [here](https://github.com/open-telemetry/opentelemetry-dotnet/blob/616e2730ec9f21da028a49d8e1e2fca6d81c43e3/src/OpenTelemetry/Metrics/MeterProviderSdk.cs#L169) where the listener is registered with a unique `state` object for each instrument, so that we don't block calls to a global concurrent dictionary. In summary, what they do is to create a small [MetricState](https://github.com/open-telemetry/opentelemetry-dotnet/blob/616e2730ec9f21da028a49d8e1e2fca6d81c43e3/src/OpenTelemetry/Metrics/MetricState.cs#L8) object which points to their [Metric](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Metrics/Metric.cs) object. I think you can do your own version of that

tracer/src/Datadog.Trace/OTelMetrics/MetricReaderHandler.cs
```diff
@@ -0,0 +1,202 @@
...
+        {
+            ProcessMeasurement(instrument, value, tags, isInteger: false);
+        }
+
+        private static void ProcessMeasurement<T>(Instrument instrument, T value, ReadOnlySpan<KeyValuePair<string, object?>> tags, bool isInteger)
+            where T : struct
+        {
+            try
+            {
+                var instrumentType = instrument.GetType().FullName;
```
zacharycmontoya: Using my suggestion from [this comment](https://github.com/DataDog/dd-trace-dotnet/pull/7511/files#r2350327923) to have a `state` object for each Instrument, you will then have a state object when measurement callback is made. I recommend determining the "instrument type" once during the state creation so you don't have to recalculate it every single time a new measurement is made.

tracer/src/Datadog.Trace/OTelMetrics/MetricReaderHandler.cs
```diff
@@ -0,0 +1,202 @@
...
+        }
+
+        private static void ProcessMeasurementCore<T>(Instrument instrument, string instrumentType, T value, ReadOnlySpan<KeyValuePair<string, object?>> tags, bool isInteger)
+            where T : struct
+        {
+            string aggregationType;
+            if (instrumentType.StartsWith("System.Diagnostics.Metrics.Counter`1") ||
+                instrumentType.StartsWith("System.Diagnostics.Metrics.ObservableCounter`1"))
+            {
+                aggregationType = "Counter";
```
zacharycmontoya: Small change: You should use an enum rather than arbitrary strings to keep track of the instrument type. This can also help you determine if you should be recording `long` values vs `double` values, which I think you should implement. Converting `long` values into `double` values is adding overhead that I think we can avoid

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/OpenTelemetrySdkTests.cs
```diff
@@ -298,6 +298,114 @@ await Verifier.Verify(resourceMetrics, settings)
                               .DisableRequireUniquePrefix();
             }
         }
+
+        [SkippableFact]
+        [Trait("Category", "EndToEnd")]
+        [Trait("RunOnWindows", "True")]
+        public void MeterListenerCapturesAllMetrics()
```
zacharycmontoya: This test doesn't invoke a test application, so I think this would be better placed in the `Datadog.Trace.Tests` project. There's a file called `ActivityTests.cs`, so perhaps this could be placed in a `MeterListenerTests.cs` file?

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/OpenTelemetrySdkTests.cs
```diff
@@ -298,6 +298,114 @@ await Verifier.Verify(resourceMetrics, settings)
...
+            expectedMetricsCount += 2;
+#elif NET9_0_OR_GREATER
+            var gauge = meter.CreateGauge<double>("test.gauge");
+            gauge.Record(77L, testTags);
+            expectedMetricsCount += 1;
+#endif
+
+            // Trigger async collection and get metrics for testing
+            OTelMetrics.MetricReader.CollectObservableInstruments();
+            var capturedMetrics = OTelMetrics.MetricReaderHandler.GetCapturedMetricsForTesting();
```
zacharycmontoya: Note: When you implement the Exporter work, I recommend replacing this `GetCapturedMetricsForTesting` method with an InMemoryExporter so you're continuing to test the Export processing. We probably shouldn't expose it without a good use case, but you can make it internal to Datadog.Trace

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/OpenTelemetrySdkTests.cs
```diff
@@ -298,6 +298,114 @@ await Verifier.Verify(resourceMetrics, settings)
...
+            histogramMetric.Tags.Should().ContainKey("http.method").WhoseValue.Should().Be("GET");
+            histogramMetric.Tags.Should().ContainKey("rid").WhoseValue.Should().Be("1234567890");
+            var bucketCounts = histogramMetric.RunningBucketCounts;
+            bucketCounts[4].Should().Be(1L, "Value 33.0 should fall in bucket 4 (25 < 33.0 <= 50)");
+
+#if NET7_0_OR_GREATER
+            if (capturedMetrics.ContainsKey("TestMeter.test.upDownCounter"))
+            {
+                var upDownMetric = capturedMetrics["TestMeter.test.upDownCounter"];
+                upDownMetric.InstrumentType.Should().Be("UpDownCounter");
```
zacharycmontoya:
```suggestion
                upDownMetric.InstrumentType.Should().Be("UpDownCounter");
                upDownMetric.AggregationTemporality.Should().Be("Cumulative");
```

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/OpenTelemetrySdkTests.cs
```diff
@@ -298,6 +298,114 @@ await Verifier.Verify(resourceMetrics, settings)
...
+                upDownMetric.RunningSum.Should().Be(55.0);
+                upDownMetric.Tags.Should().ContainKey("http.method").WhoseValue.Should().Be("GET");
+            }
+
+            if (capturedMetrics.ContainsKey("TestMeter.test.async.upDownCounter"))
+            {
+                var asyncUpDownMetric = capturedMetrics["TestMeter.test.async.upDownCounter"];
+                asyncUpDownMetric.InstrumentType.Should().Be("UpDownCounter");
```
zacharycmontoya:
```suggestion
                asyncUpDownMetric.InstrumentType.Should().Be("UpDownCounter");
                asyncUpDownMetric.AggregationTemporality.Should().Be("Cumulative");
```

zacharycmontoya (commented):
I took a first pass at this, and I left comments with the following high-level feedback:
- The TemporalityPreference to Instrument.Temporality needs to be updated for correctness. And then the system-tests I've started should identify any other correctness issues.
- In general, we can use enums rather than strings since we have a well-defined set of Instruments
- Performance: We can add state to the measurement event callbacks, so let's use that to improve the access to individual metric points when we need to update their values.

Note: I still have several configurations I need to test in system-tests so let's sync offline as you progress on that testing.

tracer/src/Datadog.Trace/OTelMetrics/MetricReaderHandler.cs
link04: Thank you so much! I've added an enum for the value and tests for each option in the code below: https://github.com/DataDog/dd-trace-dotnet/pull/7511/files#diff-d18df43162fe4f99aa3868b370283b16c2829a66f28c8d0979a7d72fc5853eeeR21-R34

tracer/src/Datadog.Trace/OTelMetrics/MetricState.cs
```diff
@@ -0,0 +1,154 @@
...
+                        break;
+                    case InstrumentType.Histogram:
+                        _metricPoint.UpdateHistogram(value);
+                        break;
+                }
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void RecordMeasurementDouble(double value, ReadOnlySpan<KeyValuePair<string, object?>> tags)
```
zacharycmontoya: One thing that's important to know is that if a measurement is recorded with two different sets of tags, then you have two different metrics you are outputting. For example, if you have a Counter instrument and you record a measurement with `tags={"status": "200"}` and you record another measurement with `tags={"status": "500"}`, those should have two different aggregations. I think the easiest way to translate that requirement into your work is to think about your types this way:
- There's a 1:1 relationship from Instrument to MetricState
- There's 1:Many relationship from Instrument/MetricState to MetricPoint

Admittedly, I haven't had time to finish my system-tests so I'm going to be working as fast as I can to codify this in my parametric system-tests

zacharycmontoya: So when it comes to the mapping from each Instrument/MetricState, you can think about as a Dictionary where the key is a set of tags and the value is a MetricPoint

tracer/src/Datadog.Trace/OTelMetrics/MetricReaderHandler.cs
```diff
@@ -44,158 +54,280 @@ public static void OnInstrumentPublished(Instrument instrument, MeterListener li
...
+                var temporality = GetTemporalityString(aggregationType.Value);
+                var tagsDict = new Dictionary<string, object?>();
 
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private static string GetTemporalityString()
-        {
-            return _cachedTemporality ??= Tracer.Instance.Settings.OtlpMetricsTemporalityPreference.ToString();
-        }
+                // RFC requirement: Check for duplicate instrument registration
+                if (CapturedMetrics.Any(kvp =>
```
zacharycmontoya: It would be worth checking to see if the runtime already handles this so we don't have to. It would be awesome if the runtime already did checks so when an instrument is created twice with the same information, it returns the _same_ instrument object (`Object.Equals(instrument1, instrument2) == true`)

tracer/src/Datadog.Trace/OTelMetrics/MetricReaderHandler.cs
```diff
@@ -44,158 +54,280 @@ public static void OnInstrumentPublished(Instrument instrument, MeterListener li
...
+                    kvp.Key.InstrumentType == aggregationType.Value &&
+                    string.Equals(kvp.Key.InstrumentName, instrument.Name, StringComparison.OrdinalIgnoreCase) &&
+                    kvp.Key.Unit == instrument.Unit &&
+                    kvp.Key.Description == instrument.Description))
+                {
+                    Log.Warning(
+                        "Duplicate instrument registration detected: {InstrumentType} '{InstrumentName}' from meter '{MeterName}'. Previous instrument will be reused.",
```
zacharycmontoya: You should also include the Unit and Description in the message since those fields are also tested for equality

tracer/src/Datadog.Trace/OTelMetrics/MetricState.cs
zacharycmontoya: If you end up looking through the OpenTelemetry .NET SDK code-base for a reference, the [Metric class](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Metrics/Metric.cs) represents one overall Metric (similar to what you have in your MetricPoint right now), and the Metric has a reference to an AggregatorStore which stores the mapping of Tags to individual [MetricPoint objects](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Metrics/MetricPoint/MetricPoint.cs) (holds the tags for that stream and the current running value)

tracer/test/Datadog.Trace.Tests/MeterListenerTests.cs
```diff
@@ -3,39 +3,43 @@
...
     {
         [Theory]
-        [InlineData("delta", "Delta", "Cumulative")]
-        [InlineData("cumulative", "Cumulative", "Cumulative")]
-        [InlineData("lowmemory", "Delta", "Cumulative")]
-        public void CapturesAllMetricsWithCorrectTemporality(string temporality, string expectedCounterTemporality, string expectedUpDownTemporality)
+        [InlineData("cumulative", AggregationTemporality.Cumulative, AggregationTemporality.Cumulative, AggregationTemporality.Cumulative, AggregationTemporality.Cumulative)]
+        [InlineData("delta", AggregationTemporality.Delta, AggregationTemporality.Cumulative, AggregationTemporality.Delta, AggregationTemporality.Delta)]
+        [InlineData("lowmemory", AggregationTemporality.Delta, AggregationTemporality.Cumulative, AggregationTemporality.Delta, AggregationTemporality.Cumulative)]
+        public void CapturesAllMetricsWithCorrectTemporality(
```
zacharycmontoya: nit: I think an even stronger way to write this test is to capture your existing method as a helper method which accepts the configuration and "expected values", and then we can have four `[Fact]` test methods:
- `CapturesAllMetricsWithCorrectTemporality_Default()`
- `CapturesAllMetricsWithCorrectTemporality_DeltaPreference()`
- `CapturesAllMetricsWithCorrectTemporality_CumulativePreference()`
- `CapturesAllMetricsWithCorrectTemporality_LowMemoryPreference()`

The reason I propose this is because a `[Theory]` implies that there are an endless possible combinations of inputs and we're only testing 3 of them, whereas a `[Fact]` implies that there are no variable inputs. Since in reality we only have the four possible settings (the temporality preference configuration) we can have one test method for each scenario and we know we've tested all possible scenarios.

tracer/src/Datadog.Trace/OTelMetrics/MetricState.cs
zacharycmontoya: Regarding one instrument creating two data points, I've just pushed up a test case in my system-tests branch to demonstrate this: https://github.com/DataDog/system-tests/pull/5106/files#diff-213dd5f07c95c117f0a724e5139c2238cdffe33c0398e4fdad18037982a3b8b5R301

You'll see that the user code represents just that by submitting one Counter add with `attributes=DEFAULT_MEASUREMENT_ATTRIBUTES` and another with `attributes=NON_DEFAULT_MEASUREMENT_ATTRIBUTES`:
```python
            t.otel_counter_add(DEFAULT_METER_NAME, name, DEFAULT_INSTRUMENT_UNIT, DEFAULT_INSTRUMENT_DESCRIPTION, non_negative_value, DEFAULT_MEASUREMENT_ATTRIBUTES)
            t.otel_counter_add(DEFAULT_METER_NAME, name, DEFAULT_INSTRUMENT_UNIT, DEFAULT_INSTRUMENT_DESCRIPTION, second_non_negative_value, NON_DEFAULT_MEASUREMENT_ATTRIBUTES)
```

The result in the OTLP is that there's one [Sum aggregation](https://github.com/open-telemetry/opentelemetry-proto/blob/0d02f212598f3ec1dda35274e87f59351f619058/opentelemetry/proto/metrics/v1/metrics.proto#L238), but there are now two entries in its [data_points array](https://github.com/open-telemetry/opentelemetry-proto/blob/0d02f212598f3ec1dda35274e87f59351f619058/opentelemetry/proto/metrics/v1/metrics.proto#L241C3-L241C44), and each one has a unique set of [attributes](https://github.com/open-telemetry/opentelemetry-proto/blob/0d02f212598f3ec1dda35274e87f59351f619058/opentelemetry/proto/metrics/v1/metrics.proto#L400)

link04: Ty man, added tests here to make sure the behavior works:
https://github.com/DataDog/dd-trace-dotnet/pull/7511/files?new_files_changed=true#diff-d18df43162fe4f99aa3868b370283b16c2829a66f28c8d0979a7d72fc5853eeeR33-R98

And a related: `internal readonly struct TagSet : IEquatable<TagSet>` file.

tracer/src/Datadog.Trace/OTelMetrics/MetricReaderHandler.cs
```diff
@@ -0,0 +1,349 @@
...
+                    Log.Debug("Skipping unsupported instrument: {InstrumentName} of type: {InstrumentType}", instrument.Name, instrumentType);
+                    return null;
+                }
+
+                var metricStreamIdentity = new MetricStreamIdentity(instrument, aggregationType.Value);
+                var temporality = GetTemporality(aggregationType.Value);
+                var tagsDict = new Dictionary<string, object?>();
+
+                // RFC requirement: Check for duplicate instrument registration
+                if (CapturedMetrics.ContainsKey(metricStreamIdentity))
```
zacharycmontoya: I have an edge case that I think warrants its own test case. Only the name is case-insensitive, so we should make sure if everything is the same but two registrations have the same name but different casing, we should treat it as a duplicate instrument

zacharycmontoya: Previously you had a case-insensitive comparison on the Name field, so I want to make sure we keep that

tracer/src/Datadog.Trace/OTelMetrics/MetricState.cs
```diff
@@ -17,138 +18,127 @@ namespace Datadog.Trace.OTelMetrics
...
+        public IEnumerable<MetricPoint> GetMetricPoints() => _points.Values;
+
+        public MetricStreamIdentity GetIdentity() => _identity;
 
-        public MetricPoint GetMetricPoint()
+        private MetricPoint GetOrCreatePoint(ReadOnlySpan<KeyValuePair<string, object?>> tags)
         {
-            lock (_lock)
+            var tagSet = TagSet.FromSpan(tags);
+            var tagsArray = tags.ToArray();
```
zacharycmontoya: This can be optimized by removing the array creation entirely. The `ReadOnlySpan<T>` allows you to safely iterate over it, so you can just iterate over that instead of creating a new array and iterating over it

tracer/src/Datadog.Trace/OTelMetrics/TagSet.cs
```diff
@@ -0,0 +1,50 @@
...
+            if (tags.Length == 0)
+            {
+                return new TagSet(string.Empty);
+            }
+
+            // Order keys deterministically to ensure stable equality
+            var sorted = tags.ToArray();
+            Array.Sort(sorted, (a, b) => string.CompareOrdinal(a.Key, b.Key));
+
+            var key = string.Join(";", sorted.Select(kv => $"{kv.Key}={kv.Value}"));
```
zacharycmontoya: This TagSet implementation and usage looks good. I was trying to think of a more efficient way to create a unique key from the `ReadOnlySpan<KeyValuePair<string, object?>> tags` and off the top of my head I can't think of one. Perhaps we could intern the resulting string because if we get the same tags over and over again, we could save on memory allocations, but then that string value will not be garbage-collected quickly, so it's not a clear win.

zacharycmontoya: Perhaps at a later time we could create a central cache that maps unique `ReadOnlySpan<KeyValuePair<string, object?>> tags` objects => `TagSet` objects by creating a [trie](https://en.wikipedia.org/wiki/Trie) data structure. But not worth spending more time on it without microbenchmarking

tracer/test/Datadog.Trace.Tests/MeterListenerTests.cs
```diff
@@ -0,0 +1,297 @@
...
+
+            // Verify we have separate metric points for each tag set
+            var counterMetrics = capturedMetrics.Where(m => m.InstrumentName == "test.counter").ToList();
+            counterMetrics.Count.Should().Be(2, "Counter should have 2 separate metric points for different tag sets");
+
+            // Verify the actual tag values for counter metrics
+            var counterWithTestValue = counterMetrics.FirstOrDefault(m => m.Tags.ContainsKey("test_attr") && m.Tags["test_attr"]?.ToString() == "test_value");
+            var counterWithNonDefaultValue = counterMetrics.FirstOrDefault(m => m.Tags.ContainsKey("test_attr") && m.Tags["test_attr"]?.ToString() == "non_default_value");
+
+            counterWithTestValue.Should().NotBeNull("Counter should have a metric point with test_attr=test_value");
```
zacharycmontoya, line 90: Can you also assert the `RunningSum` value for the two different counter data points? And for the other metrics where you created two different data points?

tracer/test/Datadog.Trace.Tests/MeterListenerTests.cs
```diff
@@ -0,0 +1,297 @@
...
+        {
+            CapturesAllMetricsWithCorrectTemporality(
+                "lowmemory",
+                AggregationTemporality.Delta,
+                AggregationTemporality.Cumulative,
+                AggregationTemporality.Delta,
+                AggregationTemporality.Cumulative);
+        }
+
+        private void CapturesAllMetricsWithCorrectTemporality(
```
zacharycmontoya, line 228: nit: Two small suggestions
1. You have some of the expected temporality results as parameters but not all. I think it would be better to either list _all_ of them or list only the ones that change (e.g. Counter). Maybe listing all of them would be better because of my 2nd suggestion...
2. Now that you've organized the tests this way, it's really easy to read the above `[Fact]` methods to understand the inputs and outputs. It would look even better if you added the argument name (e.g. `expectedCounterTemporality: AggregationTemporality.Delta`) to make those calls even more human-readable

tracer/missing-nullability-files.csv
```diff
@@ -137,6 +137,7 @@ src/Datadog.Trace/HttpOverStreams/HttpResponse.cs
 src/Datadog.Trace/HttpOverStreams/IHttpContent.cs
 src/Datadog.Trace/Iast/ITaintedMap.cs
 src/Datadog.Trace/Iast/SourceType.cs
+src/Datadog.Trace/OTelMetrics/AggregationTemporality.cs
```
andrewlock: nit: we shouldn't add new files here, we should add `#nullable enable` to this file and fix any issues instead

tracer/src/Datadog.Trace/OTelMetrics/AggregationTemporality.cs
```diff
@@ -0,0 +1,26 @@
+// <copyright file="AggregationTemporality.cs" company="Datadog">
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2 License.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
+// </copyright>
+
+#if NET6_0_OR_GREATER
+
+namespace Datadog.Trace.OTelMetrics
```
andrewlock: tiny nit: favor namespace declarations to save on indenting:

```
namespace Datadog.Trace.OTelMetrics;
```

tracer/src/Datadog.Trace/OTelMetrics/MetricPoint.cs
```diff
@@ -0,0 +1,106 @@
...
+using System;
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+using System.Threading;
+
+namespace Datadog.Trace.OTelMetrics
+{
+    internal class MetricPoint(string instrumentName, string meterName, InstrumentType instrumentType, AggregationTemporality? temporality, Dictionary<string, object?> tags)
+    {
+        // Static fields first
```
andrewlock: nit: these comments are just (AI) noise 😉

tracer/src/Datadog.Trace/OTelMetrics/MetricPoint.cs
```diff
@@ -0,0 +1,106 @@
...
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal void UpdateGauge(double value)
+        {
+            Interlocked.Exchange(ref _runningDoubleValue, value);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
andrewlock: I'd be inclined to _not_ add these tbh. In general, the JIT is good at deciding when to inline, and it's not without costs. In general, unless it's obvious, or you've benchmarked, it's best not to add these IMO

tracer/src/Datadog.Trace/OTelMetrics/MetricPoint.cs
```diff
@@ -0,0 +1,106 @@
...
+        internal void UpdateHistogram(double value)
+        {
+            // Find bucket index first (outside lock for performance)
+            var bucketIndex = FindBucketIndex(value);
+
+            lock (_histogramLock)
+            {
+                unchecked
```
andrewlock: FWIW, `unchecked` is technically the default, but it doesn't hurt to add this in case someone has explicitly opted in to checked. _However_, is unchecked roll over what you actually want (where a big number suddenly becomes negative?!)? Or do you actually want to clamp to max in this case? Or should we reject the point entirely?

That's maybe a question for @zacharycmontoya tbh

tracer/src/Datadog.Trace/OTelMetrics/MetricReader.cs
```diff
@@ -0,0 +1,80 @@
...
+
+            Log.Debug("MeterListener initialized successfully.");
+        }
+
+        public static void Stop()
+        {
+            if (_meterListenerInstance is IDisposable disposableListener)
+            {
+                _meterListenerInstance = null;
+                disposableListener.Dispose();
```
andrewlock: I'm assuming this de-registers all the callbacks, correct?

tracer/src/Datadog.Trace/OTelMetrics/MetricReaderHandler.cs
```diff
@@ -0,0 +1,340 @@
...
+        private static int _metricCount;
+
+        public static void OnInstrumentPublished(Instrument instrument, MeterListener listener)
+        {
+            var meterName = instrument.Meter.Name;
+
+            // RFC requirement: Validate instrument name syntax
+            if (!IsValidInstrumentName(instrument.Name))
+            {
+                Log.Warning("Invalid instrument name '{InstrumentName}' from meter '{MeterName}'. Instrument names must be ASCII, start with letter, contain only alphanumeric characters, '_', '.', '/', '-' and be max 255 characters.", instrument.Name, meterName);
```
andrewlock: How often is `OnInstrumentPublished` called? Just thinking about log verbosity/frequency - if it's only on app startup/rare that's fine

tracer/src/Datadog.Trace/OTelMetrics/MetricReaderHandler.cs
```diff
@@ -0,0 +1,340 @@
...
+                return; // Skip this instrument
+            }
+
+            bool shouldEnable;
+            var enabledMeterNames = Tracer.Instance.Settings.OpenTelemetryMeterNames;
```
andrewlock: The `Tracer.Instance` call is a red-flag to me. For me, this implies that these shouldn't be `static` classes, and rather they should be _instance_ types, that we can pass initalisation values to in the constructor. That said, we can easily refactor this later, so it doesn't need to be blocking

tracer/src/Datadog.Trace/OTelMetrics/MetricState.cs
```diff
@@ -0,0 +1,135 @@
...
+
+        // 1:N mapping of tag sets to MetricPoints
+        private readonly ConcurrentDictionary<TagSet, MetricPoint> _points = new();
+
+        public MetricState(MetricStreamIdentity identity)
+        {
+            _identity = identity;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
andrewlock: This almost certainly shouldn't be aggressively inlined without benchmarking IMO. Trust the JIT 😄

tracer/src/Datadog.Trace/OTelMetrics/MetricState.cs
```diff
@@ -0,0 +1,135 @@
...
+                    break;
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void RecordMeasurementDouble(double value, ReadOnlySpan<KeyValuePair<string, object?>> tags)
+        {
+            if ((_identity.InstrumentType == InstrumentType.Counter || _identity.InstrumentType == InstrumentType.Histogram) && value < 0)
+            {
+                Log.Warning(
```
andrewlock: This is where I start to get worried about flooding the logs, seeing as it's potentially every time a value is recorded. I'm fine with us worrying about that later though too (adding a rate-limiting/uniqueness check on the instrument name)

tracer/src/Datadog.Trace/OTelMetrics/TagSet.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        public static TagSet FromSpan(ReadOnlySpan<KeyValuePair<string, object?>> tags)
+        {
+            if (tags.Length == 0)
+            {
+                return new TagSet(string.Empty);
+            }
+
+            // Order keys deterministically to ensure stable equality
+            var sorted = tags.ToArray();
```
andrewlock: We should _really_ avoid calling `ToArray()` here if possible, as it completely defeats the point of ReadOnlySpan 🙁 We can probably do something similar with a `Span<>` for small arrays, and then fallback to `ToArray()` just for big collections

tracer/src/Datadog.Trace/OTelMetrics/MetricReaderHandler.cs
```diff
@@ -0,0 +1,340 @@
...
+        public static void OnMeasurementRecordedLong(Instrument instrument, long value, ReadOnlySpan<KeyValuePair<string, object?>> tags, object? state)
+        {
+            if (state is MetricState metricState)
+            {
+                metricState.RecordMeasurementLong(value, tags);
+            }
+            else
+            {
+                // Log error - this should not happen if we follow OTel SDK pattern
+                Log.Warning("Measurement recorded for instrument '{InstrumentName}' but no MetricState provided. Measurement will be dropped.", instrument.Name);
```
andrewlock: All these logs in potential "hot path" concern me a bit...

tracer/src/Datadog.Trace/OTelMetrics/MetricReaderHandler.cs
```diff
@@ -0,0 +1,340 @@
...
+        public static void OnMeasurementRecordedFloat(Instrument instrument, float value, ReadOnlySpan<KeyValuePair<string, object?>> tags, object? state)
+        {
+            OnMeasurementRecordedDouble(instrument, value, tags, state);
+        }
+
+        internal static IReadOnlyList<MetricPoint> GetMetricsForExport(bool clearAfterGet = false)
+        {
+            var allMetricPoints = new List<MetricPoint>();
+
+            foreach (var kvp in CapturedMetrics)
```
andrewlock: According to the docs:

> The enumerator returned from the dictionary is safe to use concurrently with reads and writes to the dictionary, however it does not represent a moment-in-time snapshot of the dictionary. The contents exposed through the enumerator may contain modifications made to the dictionary after [GetEnumerator](https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getenumerator?view=net-9.0) was called.

I wonder if we should be working with a snapshot, to avoid missing metrics. They could be being added while you're enumerating, get missed, and then get cleared 🤔

tracer/src/Datadog.Trace/OTelMetrics/MetricReaderHandler.cs
```diff
@@ -0,0 +1,340 @@
...
+            {
+                var metricState = kvp.Value;
+                var metricPoints = metricState.GetMetricPoints();
+                allMetricPoints.AddRange(metricPoints);
+            }
+
+            return allMetricPoints;
+        }
```
andrewlock: nit: AFACIT this is identical
```suggestion
        internal static IReadOnlyList<MetricPoint> GetCapturedMetricsForTesting()
            => GetMetricsForExport(clearAfterGet: false);

```

tracer/src/Datadog.Trace/OTelMetrics/MetricReaderHandler.cs
```diff
@@ -0,0 +1,340 @@
...
+
+        // For testing only - reset all captured metrics
+        internal static void ResetForTesting()
+        {
+            CapturedMetrics.Clear();
+            MetricStreamNames.Clear();
+            Interlocked.Exchange(ref _metricCount, 0);
+        }
```
andrewlock: nit: this is another sign for me that this whole type shouldn't be static. This is the sort of thing that's super easy to test and reason about if you can just do `new MetricReaderHandler()`. It also can make testing faster, as you can run all those tests in parallel if desired.

tracer/src/Datadog.Trace/OTelMetrics/MetricReaderHandler.cs
```diff
@@ -0,0 +1,340 @@
...
+            }
+
+            // 2. Maximum length 255 characters
+            if (name.Length > 255)
+            {
+                return false;
+            }
+
+            // 3. ASCII string
+            if (!IsAsciiString(name))
```
andrewlock: nit: I think you can remove this check, as it's covered by 4 and 5?

tracer/src/Datadog.Trace/OTelMetrics/MetricReaderHandler.cs
```diff
@@ -0,0 +1,340 @@
...
+                if (!MetricStreamNames.Add(metricStreamIdentity.MetricStreamName))
+                {
+                    Log.Warning("Duplicate metric stream detected: {MetricStreamName}. Measurements from this instrument will still be exported but may result in conflicts.", metricStreamIdentity.MetricStreamName);
+                }
+
+                var state = new MetricState(metricStreamIdentity);
+
+                // Store in global dictionary for export
+                CapturedMetrics.TryAdd(metricStreamIdentity, state);
+                Interlocked.Increment(ref _metricCount);
```
andrewlock: This field seems unnecessary, it's unused AFAICT? 🤔

tracer/src/Datadog.Trace/OTelMetrics/MetricState.cs
```diff
@@ -0,0 +1,135 @@
...
+        private MetricPoint GetOrCreatePoint(ReadOnlySpan<KeyValuePair<string, object?>> tags)
+        {
+            var tagSet = TagSet.FromSpan(tags);
+
+            var tagDict = new Dictionary<string, object?>();
+            for (int i = 0; i < tags.Length; i++)
+            {
+                var kv = tags[i];
+                tagDict[kv.Key] = kv.Value;
+            }
```
andrewlock: Why can't we do this? 🤔 

```
var tagDict = new Dictionary<string, object?>(tags);
```

Also, we're creating this new dictionary every time you try to get a point, even if it already exists. I would suggest moving the construction into the lambda as a minimum, e.g.:

```csharp
return _points.GetOrAdd(tagSet, _ => new MetricPoint(
                _identity.InstrumentName,
                _identity.MeterName,
                _identity.InstrumentType,
                MetricReaderHandler.GetTemporality(_identity.InstrumentType),
                new Dictionary<string, object?>(tags)));
```

Also, this will still allocate a closure every time you call this method... you _might_ want to consider a non-allocating TryGetValue first:

```csharp
        private MetricPoint GetOrCreatePoint(ReadOnlySpan<KeyValuePair<string, object?>> tags)
        {
            var tagSet = TagSet.FromSpan(tags);
            if (_points.TryGetValue(tagset, out var value))
            {
                return value;
            }

            // fallback, have to allocate
            return _points.GetOrAdd(tagSet, _ => new MetricPoint(
                _identity.InstrumentName,
                _identity.MeterName,
                _identity.InstrumentType,
                MetricReaderHandler.GetTemporality(_identity.InstrumentType),
                tagDict));
        }
```

andrewlock (approved):
LGTM in general, nice job! 

Some overall comments:
- Lose the `[MethodImpl(MethodImplOptions.AggressiveInlining)]`. There's a lot of places where that is likely going to be harmful overall, and there's no real need for it unless benchmarking shows it's definitely going to be useful IMO.
- The use of statics makes testing things more complicated. I would consider converting to instance objects and then creating a single static instance for the lifetime management.
- There's a couple of places that might need "fixing" for correctness, or at least adding comments acknowledging/working around the issue e.g. overflow

tracer/src/Datadog.Trace/OTelMetrics/MetricReaderHandler.cs
zacharycmontoya: Yes this is called only when creating instruments for the first time, so typically at startup

andrewlock: Cool, this is fine then 👍

tracer/src/Datadog.Trace/OTelMetrics/MetricPoint.cs
zacharycmontoya: Looking at the equivalent history in opentelemetry-dotnet, they currently implement a similar `unchecked` section:
```csharp
unchecked
{
    this.runningValue.AsLong++;
    this.histogramBuckets.RunningSum += number;
    this.histogramBuckets.RunningBucketCounts[i]++;
}
```

This `unchecked` was added in this [PR](https://github.com/open-telemetry/opentelemetry-dotnet/pull/2951) so that they could [avoid throwing an exception](https://github.com/open-telemetry/opentelemetry-dotnet/pull/2951#discussion_r815210688) on a possible increment operation since they are hold a critical section lock.

In terms of correctness, I'm not aware of any mandates for handling integer overflows. I think it would be OK to accept this, and it would be consistent with opentelemetry-dotnet

tracer/src/Datadog.Trace/OTelMetrics/TagSet.cs
zacharycmontoya: So I don't know how we get away from creating an array every time because we want to do the following operations:
1. Check if the metric has any stream _with the same set of tags_
2. If not, create a new metric stream

We can of course split this into two operations:
1. Do a check on the tags before they're sorted (the tag order isn't likely to change)
2. If that fails, sort and then check the tags

However, I think we'll need to move the tags in the `ReadOnlySpan` to an `Array` so we can store it in any sort of helper class. Maybe in this PR we should implement a small cache like in [opentelemetry-dotnet](https://github.com/open-telemetry/opentelemetry-dotnet/blob/4fed6b58e657ba9fbeb719e3ffa3a9cb7103fc38/src/OpenTelemetry/Metrics/ThreadStaticStorage.cs#L37). This uses a small cache to hold arrays of several sizes so each time we need to create an array, it's pre-allocated and we do a `ReadOnlySpan<T>.CopyTo` to reuse that space.

Would you consider this blocking for the PR?

tracer/src/Datadog.Trace/OTelMetrics/MetricReaderHandler.cs
zacharycmontoya: +1 +1

tracer/src/Datadog.Trace/OTelMetrics/MetricState.cs
zacharycmontoya: +1, I think in a previous commit the dict was constructed in the lambda. Let's make sure to move the dictionary construction into the lambda so it's only calculated when creating a new metric point

tracer/src/Datadog.Trace/OTelMetrics/TagSet.cs
andrewlock: > Would you consider this blocking for the PR?

Nah, not unless it meaningfully shifts benchmarks I think. Given this isn't enabled by default yet either, that's fine IMO. 

That said, all we _actually_ need to do is sort the collection right and then enumerate it, so we can probably do that using a stackalloc array (esp as this is .NET 6) and do the sorting that way? We don't actually _need_ the tag array itself I think?

tracer/src/Datadog.Trace/OTelMetrics/TagSet.cs
andrewlock: Arguably a cached stringbuilder with a foreach over the tags would make sense here, e.g. something like this?

```csharp
var sb = StringBuilderCache.Acquire();

foreach(var pair in sorted)
{
    sb.Append(kv.Key);
    sb.Append('=');
    sb.Append(kv.Value);
    sb.Append(';');
}

var key = StringBuilderCache.GetStringAndRelease(sb);
```

tracer/src/Datadog.Trace/OTelMetrics/MetricPoint.cs
link04: Ty applied in last commit: https://github.com/DataDog/dd-trace-dotnet/pull/7511/commits/387023c516283c8ee4c8485f90986f0a0a062da2

tracer/src/Datadog.Trace/OTelMetrics/TagSet.cs
link04: Will address this in the next PR: https://github.com/DataDog/dd-trace-dotnet/pull/7514

tracer/src/Datadog.Trace/OTelMetrics/MetricReaderHandler.cs
link04: Will remove static methods in next PR: https://github.com/DataDog/dd-trace-dotnet/pull/7514

tracer/src/Datadog.Trace/OTelMetrics/MetricReader.cs
link04: Looked into this and found that for before .NET 9 I had to end up adding a no-op MeasurementsCompleted under #if NET6_0 || NET7_0 || NET8_0 so Dispose() to truly de-registers callbacks on those runtimes. Also made Stop() idempotent/race-free with Interlocked.Exchange. On .NET 9+ there’s no behavioral change.

PR #7513, opened 2025-09-13
[Dynamic Instrumentation] Avoid async instrumentation if `AsyncMethodDebuggerInvokerV2` does not exist

GreenMatan:
## Summary of changes
In #7441, the injected field of async methods has been changed from `AsyncDebuggerState` to `System.Object`. In this PR, we make sure instrumentation is not being applied if the loaded `Datadog.Trace` does not contain the new signatures that accept `System.Object` for log/span probes.

## Reason for change
Avoid instrumentation if there are no appropriate signatures to be called. The only place where we can change type layout is in `ModuleLoadFinished`, this is why injection of our field is done there. It might happen before `Datadog.Trace` dll has been loaded into the application, thus we cannot know at injection time if we have the new signatures or not.
Once `Datadog.Trace` actually loads, we try to lookup for the type `AsyncMethodDebuggerInvokerV2` as an anchor to know if the new signatures are there. If they are not, we skip instrumentation on async methods and log appropriate message.


tracer/src/Datadog.Tracer.Native/debugger_method_rewriter.cpp
```diff
@@ -1633,6 +1631,12 @@ HRESULT DebuggerMethodRewriter::ApplyAsyncMethodProbe(
         return E_NOTIMPL; // We do not fail the whole instrumentation as there could be Line Probes that we want to emit. They do not suffer from the absence of the IsReEntry field.
     }
 
+    if (!m_corProfiler->IsAsyncMethodDebuggerInvokerV2TypeAvailable())
+    {
+        Logger::Warn("DebuggerMethodRewriter::ApplyAsyncMethodProbe: The type `AsyncMethodDebuggerInvokerV2` is not defined in `Datadog.Trace`. Please upgrade your `Datadog.Trace` to be able to probe async methods.");
+        return E_UNEXPECTED; // We do not fail the whole instrumentation as there could be Line Probes that we want to emit. They do not suffer from the absence of the IsReEntry field.
```
dudikeleti, line 1637: Nit: we currently return E_NOTIMPL in some places and E_UNEXPECTED in others, with the same comment (“we do not fail the whole instrumentation…”).
It might make sense to use different error codes, but that’s also more bug prone (we need to remember to check both and account for differences from E_FAIL).
I’m wondering if we should just stick with one.

tracer/src/Datadog.Tracer.Native/cor_profiler.cpp
```diff
@@ -876,6 +876,8 @@ HRESULT CorProfiler::TryRejitModule(ModuleID module_id, std::vector<ModuleID>& m
 
         call_target_state_skip_method_body_function_available = IsSkipMethodBodyEnabled() && EnsureCallTargetStateSkipMethodBodyFunctionAvailable(module_metadata);
 
+        asyncmethoddebuggerinvokerv2_type_available = EnsureAsyncMethodDebuggerInvokerV2TypeAvailable(module_metadata);
```
dudikeleti: Can this run twice in case two managed profiler assemblies are loaded into the process (for example, one with the new type and one without)?

tracer/src/Datadog.Tracer.Native/debugger_method_rewriter.cpp
```diff
@@ -2394,10 +2404,13 @@ HRESULT DebuggerMethodRewriter::Rewrite(RejitHandlerModule* moduleHandler,
 
         if (hr == E_NOTIMPL)
         {
-            ProbesMetadataTracker::Instance()->SetErrorProbeStatus(spanProbeId,
-                                                                   invalid_method_probe_probe_is_not_supported);
+            ProbesMetadataTracker::Instance()->SetErrorProbeStatus(spanProbeId, invalid_method_probe_probe_is_not_supported);
             Logger::Info("Emplacement of a span probe is not supported.");
         }
+        else if (hr == E_UNEXPECTED)
```
dudikeleti, line 2410: Nit: maybe check first `E_NOTIMPL  && IsAsyncMethodDebuggerInvokerV2TypeAvailable` to avoid `E_UNEXPECTED`?

tracer/src/Datadog.Tracer.Native/debugger_method_rewriter.cpp
GreenMatan: I agree with you sentiment. I cannot stick with `E_NOTIMPL`, though, as the upper layers dictate what the probe status is and I have to distinguish between `E_NOTIMPL` (that assigns certain probe status) and the new one. Technically it is _unexpected_ in the sense that it should not happen - in V3 of our tracer we expect the native and managed to always match. There might be additional `Datadog.Trace` loaded into the app with older version, but it has to be alongside the newer one. So if we cannot find proper `Datadog.Trace` version we consider it as an invalid state

tracer/src/Datadog.Tracer.Native/debugger_method_rewriter.cpp
GreenMatan: but then I expose the reason why I bail out instrumenting that async method to that layer, where currently this layer only handles the HRESULT that returns
I chose specifically `E_UNEXPECTED` since I consider this state to be invalid. We don't anticipate it happening in the wild, this is mainly a safeguard against the unknown unknown
In v3 of the tracer we anticipate to always have a match between native <-> managed for automatic instrumentation

tracer/src/Datadog.Tracer.Native/cor_profiler.cpp
GreenMatan: That's a very good question. To accommodate that, I'll conditionally assign a value to `asyncmethoddebuggerinvokerv2_type_available` if it's false. That way, if we encounter _at least_ one `Datadog.Trace` with that type in it, we consider it safe to instrument async methods. In our instrumentation we always reference the latest `Datadog.Trace` version so we won't end up calling a function inside an older `Datdog.Trace`

GreenMatan: done in https://github.com/DataDog/dd-trace-dotnet/pull/7513/commits/91fd7fd963a2b4f38bde71851a808d9f2bfdb36e#diff-974ab4b3d49453bf1b92fa22c50acabddbce626ea392310160fbe7c6c48186dcR879-R883.

tracer/src/Datadog.Tracer.Native/dd_profiler_constants.h
```diff
@@ -118,6 +118,9 @@ const shared::WSTRING libdatadog_libraryconfig_nativemethods_type = WStr("Datado
 const shared::WSTRING calltargetstate_type_name = WStr("Datadog.Trace.ClrProfiler.CallTarget.CallTargetState");
 const shared::WSTRING calltargetstate_skipmethodbody_function_name = WStr("GetSkipMethodBody");
 
+// Also defined in `debugger_tokens.h` as `managed_profiler_debugger_async_method_invoker_type`
+const shared::WSTRING asyncmethoddebuggerinvokerv2_type_name = WStr("Datadog.Trace.Debugger.Instrumentation.AsyncMethodDebuggerInvokerV2");
+
```
andrewlock, line 123: Why do we need to define it twice? 🤔

andrewlock (approved):
LGTM on the non-debugger side (I defer to you on the correctness of that) but I'll let @tonyredondo weigh in too seeing as it's native code 😄

tonyredondo (approved):
LGTM

PR #7514, opened 2025-09-13
[OTEL] Metrics API Support - Metrics Reader & OTLP Exporter

link04:
## Summary of changes
This PR implements the OTLP Metrics Exporter for OpenTelemetry Metrics API support, completing the final piece of the RFC implementation. Key additions include:
- OTLP Metrics Exporter: Implemented `MetricExporter` abstract class and `OtlpExporter` concrete implementation initially with only ` HTTP/protobuf` transport support ATM.
- Performance Optimizations: Custom protobuf serialization without external dependencies, async operations with ConfigureAwait(false)
- Retry Logic: Exponential backoff with Retry-After header handling for HTTP 429, 502, 503, 504 responses
- Telemetry Metrics: Internal metrics tracking (`metrics_export_attempts`, `metrics_export_successes`, `metrics_export_failures`) with protocol tag
- Sample App Integration: Enhanced Samples.OpenTelemetrySdk to support both DD and OTel SDK approaches for comparative snapshot testing

This completes the series of 3 PRs for OpenTelemetry Metrics API support.
1. [[OTEL] Metrics API Support - Configurations & Telemetry #7420](https://github.com/DataDog/dd-trace-dotnet/pull/7420)
2. [[OTEL] Metrics API Support -  Collection Implementation #7511](https://github.com/DataDog/dd-trace-dotnet/pull/7511)
3. [[OTEL] Metrics API Support -  Metrics Reader & OTLP Exporter #7513](https://github.com/DataDog/dd-trace-dotnet/pull/7513) <- You are here.

## Reason for change
The [RFC] OpenTelemetry Metrics API support requires the .NET tracer to act as a drop-in replacement for the OpenTelemetry SDK. This PR implements the OTLP export functionality, enabling customers to migrate from OTel SDK to Datadog SDK with zero friction.
Business Impact: Enables customers using OpenTelemetry Metrics API to switch to Datadog without code changes, supporting our strategic initiative to be OTel-native.

## Implementation details
:white_check_mark: OTLP Metrics Exporter Components:
- MetricExporter - Abstract base class defining export contract
- OtlpExporter - Concrete implementation with HTTP/protobuf transport
- OtlpMetricsSerializer - Custom protobuf serialization for performance
- MetricReader - Updated to use the new exporter architecture

:white_check_mark: Key Features:
- HTTP/protobuf transport support
- Resource attribute precedence following RFC (DD_* → OTEL_* → defaults)
- Retry logic with exponential backoff and proper HTTP status handling
- Telemetry metrics with protocol tags (protocol:http/protobuf, protocol:grpc, protocol:http/json)
- Performance optimizations with async operations and custom serialization

:white_check_mark: Sample App Enhancements:
- Conditional OTel SDK exporter via OTEL_METRICS_EXPORTER_ENABLED environment variable for testing
- Comparative testing between DD and OTel SDK approaches
- Proper MeterProvider lifecycle management with disposal

## Test coverage
Added comprehensive integration test `SubmitsOtlpMetrics` that:
- Tests both approaches: DD enabled/OTel disabled vs DD disabled/OTel enabled
- Verifies identical results: Both approaches produce the same OTLP metric requests
- Cross-version testing: Runs across all OpenTelemetry package versions
- Snapshot verification: Ensures RFC compliance and data model compatibility

## Other details

Jira: [APMAPI-1573](https://datadoghq.atlassian.net/browse/APMAPI-1573)


tracer/src/Datadog.Trace/OTelMetrics/OtlpExporter.cs
```diff
@@ -0,0 +1,275 @@
...
+            {
+                Log.Warning(ex, "Error during OTLP exporter shutdown");
+                return false;
+            }
+        }
+
+        /// <summary>
+        /// Gets the protocol tag for telemetry metrics based on the configured OTLP protocol.
+        /// </summary>
+        private Telemetry.Metrics.MetricTags.Protocol GetProtocolTag()
```
zacharycmontoya: super nit: You can calculate this string during the constructor so it's just a field access rather than calling a method

tracer/src/Datadog.Trace/OTelMetrics/MetricReader.cs
```diff
@@ -0,0 +1,138 @@
...
+    public static void Initialize()
+    {
+        if (Interlocked.CompareExchange(ref _initialized, 1, 0) == 1)
+        {
+            return;
+        }
+
+        var meterListener = new System.Diagnostics.Metrics.MeterListener();
+
+#if NET6_0 || NET7_0 || NET8_0
```
zacharycmontoya: nit: You can delete this ifdef since it's already covered by your initial `#if NET6_0_OR_GREATER`. Also since we only currently build for net6.0 you should only need `#if NET6_0`

tracer/src/Datadog.Trace/OTelMetrics/MetricReader.cs
```diff
@@ -0,0 +1,138 @@
...
+            {
+                _meterListenerInstance.RecordObservableInstruments();
+            }
+            catch (Exception ex)
+            {
+                Log.Warning(ex, "Error collecting observable instruments.");
+            }
+        }
+
+        private static async void ExportMetrics()
```
zacharycmontoya: HIGH-LEVEL: I think in general you will want to do everything synchronously instead of asynchronously, because ideally we would have our metrics reader / metrics exporter run synchronously on its own thread. That way we will always have a thread available for writing, even in the case that there's too many tasks trying to be assigned to threads.

tracer/src/Datadog.Trace/OTelMetrics/OtlpExporter.cs
```diff
@@ -0,0 +1,275 @@
...
+            // Record export attempt
+            TelemetryFactory.Metrics.RecordCountMetricsExportAttempts(GetProtocolTag());
+
+            try
+            {
+                // Take snapshots of all metrics for export
+                var metricPoints = new List<MetricPoint>();
+                foreach (var metric in metrics)
+                {
+                    metric.TakeSnapshot(outputDelta: _settings.OtlpMetricsTemporalityPreference == Configuration.OtlpTemporality.Delta);
```
zacharycmontoya: nit: Rather than have to access the settings everytime, I think design-wise it would be better if the `MetricPoint` knew at construction what temporality it is reporting. That way you only have to access the TemporalityPreference settings once per metric creation

tracer/src/Datadog.Trace/OTelMetrics/OtlpExporter.cs
```diff
@@ -0,0 +1,275 @@
...
+                    metricPoints.Add(metric);
+                }
+
+                // Serialize to OTLP protobuf format
+                var otlpPayload = OtlpMetricsSerializer.SerializeMetrics(metricPoints, _settings);
```
zacharycmontoya: I see a lot of the `_settings` being passed around through the serialization. Since you passed the `TracerSettings` to this `OtlpExporter` instance, I also recommend creating an individual instance of the `OtlpMetricsSerializer` class, that way you can prepopulate the serializer with all of the information you need (like the Resources, which you should only need to change if settings change). So in general, see if you can make this entire sequence of calls use object instances, and pre-populate as much as you can so you don't have to spend valuable CPU re-calculating things during the Export call

tracer/src/Datadog.Trace/OTelMetrics/MetricReader.cs
zacharycmontoya: I think this part may take some additional insight and discussion from the rest of the team before converting to synchronous, so let's keep the async structure of your changes intact for now, no urgency to change this part of the design

tracer/src/Datadog.Trace/OTelMetrics/OtlpMetricsSerializer.cs
```diff
@@ -0,0 +1,636 @@
...
+            return buffer.ToArray();
+        }
+
+        private static byte[] SerializeResource(TracerSettings settings)
+        {
+            using var buffer = new MemoryStream();
+            using var writer = new BinaryWriter(buffer, Encoding.UTF8);
+
+            // telemetry.sdk.name attribute (use 'opentelemetry' for ecosystem compatibility)
+            var sdkNameAttr = SerializeKeyValue("telemetry.sdk.name", "opentelemetry");
```
zacharycmontoya: We haven't discussed this yet, but I imagine `datadog` would be a good value here. Since we're not an OpenTelemetry distribution, we must not set this to `opentelemetry`.

```suggestion
            var sdkNameAttr = SerializeKeyValue("telemetry.sdk.name", "datadog");
```

tracer/src/Datadog.Trace/OTelMetrics/OtlpMetricsSerializer.cs
```diff
@@ -0,0 +1,636 @@
...
+    /// but does not require Google.Protobuf dependency (DDSketch + OpenTelemetry patterns).
+    /// </summary>
+    internal static class OtlpMetricsSerializer
+    {
+        // Protobuf wire types
+        private const int VarInt = 0;
+        private const int Fixed64 = 1;
+        private const int LengthDelimited = 2;
+
+        private static byte[] SerializeResourceMetrics(List<MetricPoint> metrics, TracerSettings settings)
```
zacharycmontoya: Did you consider vendoring a protobuf library for this? I suppose that if the OTLP collector can parse this correctly, it works for this version of the OTLP metrics proto, but this isn't really a maintainable solution since any updates to proto fields means we have to go in and update the serialization of the affected fields.

My two cents is that if all of our correctness testing passes, then this this is acceptable for merging, but we'll need to track follow-up work to either vendor in a protobuf library or set up a code-gen process to compile a proto file into the corresponding serialization logic.

tracer/src/Datadog.Trace/OTelMetrics/OtlpExporter.cs
link04: I tried applying this one in the last commit: https://github.com/DataDog/dd-trace-dotnet/pull/7514/commits/bd07ebf7b238c74551b3693267294b1a9255fd87

tracer/src/Datadog.Trace/OTelMetrics/OtlpExporter.cs
link04: I tried applying this one in the last commit: https://github.com/DataDog/dd-trace-dotnet/pull/7514/commits/bd07ebf7b238c74551b3693267294b1a9255fd87

tracer/src/Datadog.Trace/OTelMetrics/TagSet.cs
```diff
@@ -25,16 +26,55 @@ namespace Datadog.Trace.OTelMetrics;
...
         {
             return new TagSet(string.Empty);
         }
 
-        var sorted = tags.ToArray();
-        Array.Sort(sorted, (a, b) => string.CompareOrdinal(a.Key, b.Key));
+        if (len == 1)
+        {
+            ref readonly var kv = ref tags[0];
+            var sb = StringBuilderCache.Acquire(32 + kv.Key.Length);
```
zacharycmontoya: nit: Is there a strong reason for the `32 + kv.Key.Length`? If not, I think you're better off acquiring a StringBuilder without specifying a capacity.

```suggestion
            var sb = StringBuilderCache.Acquire();
```

tracer/src/Datadog.Trace/OTelMetrics/TagSet.cs
```diff
@@ -25,16 +26,55 @@ namespace Datadog.Trace.OTelMetrics;
...
+            }
+
+            return new TagSet(StringBuilderCache.GetStringAndRelease(sb));
+        }
+
+        var sorted = ArrayPool<KeyValuePair<string, object?>>.Shared.Rent(len);
+        try
+        {
+            tags.CopyTo(sorted);
+            Array.Sort(sorted, 0, len, Comparer<KeyValuePair<string, object?>>.Create(
```
zacharycmontoya: nit: I think you should create the Comparer once (the third argument to the `Array.Sort` call) and store it in a `static readonly Comparer<KeyValuePair<string, object?>>` field, so you only invoke that API once, rather than every time we need to construct a TagSet

tracer/src/Datadog.Trace/OTelMetrics/TagSet.cs
```diff
@@ -25,16 +26,55 @@ namespace Datadog.Trace.OTelMetrics;
...
+                static (a, b) => string.CompareOrdinal(a.Key, b.Key)));
 
-        var key = string.Join(";", sorted.Select(kv => $"{kv.Key}={kv.Value}"));
-        return new TagSet(key);
+            var sb = StringBuilderCache.Acquire(len * 24);
```
zacharycmontoya: nit: And here too, I encourage you to simply use `StringBuilderCache.Acquire()` unless you have a strong reason for specifying a capacity

tracer/src/Datadog.Trace/OTelMetrics/TagSet.cs
```diff
@@ -25,16 +26,55 @@ namespace Datadog.Trace.OTelMetrics;
...
+                    sb.Append(kv.Value);
+                }
+            }
+
+            var key = StringBuilderCache.GetStringAndRelease(sb);
+            return new TagSet(key);
+        }
+        finally
+        {
+            ArrayPool<KeyValuePair<string, object?>>.Shared.Return(sorted, clearArray: true);
```
zacharycmontoya: nit: I think you can leave `clearArray: false` (the default value). You're safely accessing only the first `len` items so each time you get here, you'll only copy over and iterate over the applicable tags

tracer/src/Datadog.Trace/OTelMetrics/MetricReader.cs
```diff
@@ -8,78 +8,131 @@
...
+                period: interval);
+
+            Log.Debug<int>("MetricReader started with {IntervalMs}ms export interval", _settings.OtelMetricExportIntervalMs);
+        }
 
-        Interlocked.Exchange(ref _meterListenerInstance, meterListener);
-        Interlocked.Exchange(ref _stopped, 0);
+        public async Task StopAsync()
+        {
+            _timer?.DisposeAsync();
```
zacharycmontoya:
```suggestion
            await _timer?.DisposeAsync().ConfigureAwait(false);
```

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/OpenTelemetrySdkTests.cs
```diff
@@ -297,10 +339,14 @@ await Verifier.Verify(resourceMetrics, settings)
                               .UseFileName(fileName)
                               .DisableRequireUniquePrefix();
             }
+
+            grpcServer?.DisposeAsync();
```
zacharycmontoya:
```suggestion
            await grpcServer?.DisposeAsync();
```

tracer/src/Datadog.Trace/OTelMetrics/InMemoryExporter.cs
```diff
@@ -0,0 +1,80 @@
...
+            return Task.FromResult(result);
+        }
+
+        public override bool Shutdown(int timeoutMilliseconds)
+        {
+            // No-op for in-memory exporter
+            return true;
+        }
+
+        public List<MetricPoint> Drain()
```
zacharycmontoya: Are `Drain()` and `Clear()` still being used? I don't see any references so I think you can delete them

tracer/src/Datadog.Trace/OTelMetrics/MetricPoint.cs
```diff
@@ -33,6 +33,10 @@ internal class MetricPoint(string instrumentName, string meterName, InstrumentTy
 
     public Dictionary<string, object?> Tags { get; } = tags;
 
+    public string Unit { get; } = unit ?? string.Empty;
```
zacharycmontoya: nit: Can the inputs `unit` and `description` be null? It seems like you'll never hit the null-coalescing operator `??`

tracer/src/Datadog.Trace/OTelMetrics/OtlpMetricsSerializer.cs
```diff
@@ -0,0 +1,573 @@
...
+
+            var dataPointData = SerializeNumberDataPoint(metric);
+            WriteTag(writer, FieldNumbers.DataPoints, LengthDelimited);
+            WriteVarInt(writer, dataPointData.Length);
+            writer.Write(dataPointData);
+
+            int temporality;
+            bool isMonotonic;
+
+            if (metric.InstrumentName.Contains("upDownCounter"))
```
zacharycmontoya: nit: I suggest using the instrumentType here since you can easily assert that it's one of two values

tracer/src/Datadog.Trace/OTelMetrics/OtlpMetricsSerializer.cs
```diff
@@ -0,0 +1,573 @@
...
+            {
+                temporality = 2;
```
zacharycmontoya: For the temporality values, can you use the `AggregationTemporality` enum instead? If you need to change that enum to align with the protobuf values, that's ok, though you might get a compiler warning so you'll have to write a `#pragma warning disable <CODE>` to ignore it

zacharycmontoya: And the reason I'm suggesting using the enum is because it will clearly read whether we are using `Delta` or `Cumulative` for a given aggregation, which is much more readable than a simple integer

tracer/src/Datadog.Trace/OTelMetrics/OtlpExporter.cs
```diff
@@ -0,0 +1,468 @@
...
+                Configuration.OtlpProtocol.HttpProtobuf => Telemetry.Metrics.MetricTags.Protocol.HttpProtobuf,
+                Configuration.OtlpProtocol.HttpJson => Telemetry.Metrics.MetricTags.Protocol.HttpJson,
+                _ => Telemetry.Metrics.MetricTags.Protocol.HttpProtobuf
+            };
+            _serializer = new OtlpMetricsSerializer(settings);
+
+            // Enable h2c (HTTP/2 without TLS) for gRPC only, and only for non-HTTPS endpoints
+            if (_protocol == Configuration.OtlpProtocol.Grpc && _endpoint.Scheme != Uri.UriSchemeHttps)
+            {
+                AppContext.SetSwitch("System.Net.Http.SocketsHttpHandler.Http2UnencryptedSupport", true);
```
zacharycmontoya: This seems worthy of writing an Information level log to indicate that we're enabling this runtime switch

tracer/src/Datadog.Trace/OTelMetrics/OtlpExporter.cs
```diff
@@ -0,0 +1,468 @@
...
+            _httpClient = CreateHttpClient(_endpoint);
+        }
+
+        /// <summary>
+        /// Exports a batch of metrics using OTLP protocol.
+        /// This method implements the MetricExporter base class Export operation.
+        /// </summary>
+        /// <param name="metrics">Batch of metrics to export</param>
+        /// <returns>ExportResult indicating success or failure</returns>
+        public override ExportResult Export(IReadOnlyList<MetricPoint> metrics)
```
zacharycmontoya: Can we delete this method? The reason I ask is the below `GetAwaiter().GetResult();` is risky, so if we don't need the code then let's remove it altogether. Otherwise, we will likely need to modify how you're calling sync over async code

tracer/src/Datadog.Trace/OTelMetrics/OtlpExporter.cs
```diff
@@ -0,0 +1,468 @@
...
+        /// <summary>
+        /// Shuts down the exporter and ensures all pending exports complete.
+        /// </summary>
+        /// <param name="timeoutMilliseconds">Maximum time to wait for shutdown</param>
+        /// <returns>True if shutdown completed successfully, false otherwise</returns>
+        public override bool Shutdown(int timeoutMilliseconds)
+        {
+            try
+            {
+                using var cts = new CancellationTokenSource(timeoutMilliseconds);
```
zacharycmontoya: This token source isn't doing anything, so I think you can get rid of this

tracer/src/Datadog.Trace/OTelMetrics/MetricsRuntime.cs
```diff
@@ -0,0 +1,61 @@
...
+            }
+
+            var exporter = new OtlpExporter(settings);
+            _instance = new OtelMetricsPipeline(settings, exporter);
+            _instance.Start();
+
+            LifetimeManager.Instance.AddAsyncShutdownTask((_) => StopAsync());
+        }
+
+        public static Task StopAsync()
```
zacharycmontoya: You should make `StopAsync` and `FlushAsync` async and `await` on the calls to `_instance`, to make sure we correctly wait for these operations to finish running

tracer/src/Datadog.Trace/OTelMetrics/MetricsRuntime.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            if (_instance == null)
+            {
+                return Task.CompletedTask;
+            }
+
+            return _instance.StopAsync();
+        }
+
+        public static Task FlushAsync()
```
zacharycmontoya: nit: Is `FlushAsync` used anywhere? I think only `StopAsync` is called

tracer/src/Datadog.Trace/OTelMetrics/MetricState.cs
```diff
@@ -91,39 +97,51 @@ public void RecordMeasurementDouble(double value, ReadOnlySpan<KeyValuePair<stri
...
+            return existingPoint;
         }
 
-        return _points.GetOrAdd(tagSet, _ => new MetricPoint(
-            _identity.InstrumentName,
-            _identity.MeterName,
-            _identity.InstrumentType,
-            MetricReaderHandler.GetTemporality(_identity.InstrumentType),
-            tagDict));
+        var tagsArray = tags.ToArray();  // Materialize span to heap array
```
zacharycmontoya: You should be able to remove `tagsArray` entirely and instead iterate over `tags` inside the callback

tracer/src/Datadog.Trace/OTelMetrics/MetricState.cs
```diff
@@ -91,39 +97,51 @@ public void RecordMeasurementDouble(double value, ReadOnlySpan<KeyValuePair<stri
...
-        => GetOrCreatePoint(tags).UpdateHistogram(value);
-
-    public IEnumerable<MetricPoint> GetMetricPoints() => _points.Values;
-
-    public MetricStreamIdentity GetIdentity() => _identity;
+    /// <summary>
+    /// Builds metric point snapshots and adds them to the provided list.
+    /// For Delta temporality, resets the running values after snapshotting.
+    /// </summary>
+    public void TryBuildPoints(List<MetricPoint> into)
```
zacharycmontoya: nit: You should remove the `Try` prefix. There's a standard C# convention to do `bool TrySomeOperation(out result)` and this goes against that convention

```suggestion
    public void BuildPoints(List<MetricPoint> into)
```

tracer/src/Datadog.Trace/OTelMetrics/MetricReaderHandler.cs
```diff
@@ -155,41 +175,38 @@ private static bool IsValidInstrumentName(string name)
         return true;
     }
 
-    internal static AggregationTemporality? GetTemporality(InstrumentType instrumentType)
+    private static AggregationTemporality? TemporalityFor(InstrumentType kind, OtlpTemporality preference)
```
zacharycmontoya: nit: I think the previous name, or `GetAggregationTemporality`, sounds more conventional than `TemporalityFor`

```suggestion
    private static AggregationTemporality? GetTemporality(InstrumentType kind, OtlpTemporality preference)
```

tracer/src/Datadog.Trace/OTelMetrics/MetricReaderHandler.cs
```diff
@@ -37,93 +43,107 @@ public static void OnInstrumentPublished(Instrument instrument, MeterListener li
...
+
+        var instrumentType = GetInstrumentType(instrument.GetType().FullName);
+        if (instrumentType == null)
+        {
+            Log.Debug("Skipping unsupported instrument: {InstrumentName} of type: {InstrumentType}", instrument.Name, instrument.GetType().FullName);
+            return;
+        }
+
+        var temporality = TemporalityFor(instrumentType.Value, _settings.OtlpMetricsTemporalityPreference);
+        var identity = new MetricStreamIdentity(instrument, instrumentType.Value);
```
zacharycmontoya, line 70: As a follow-up item, I think we should track adding properties for `MeterVersion` and `MeterTags` since Meters are also unique based on those fields (in addition to `MeterName`)

tracer/src/Datadog.Trace/OTelMetrics/MetricReaderHandler.cs
```diff
@@ -37,93 +43,107 @@ public static void OnInstrumentPublished(Instrument instrument, MeterListener li
         }
 
         bool shouldEnable;
-        var enabledMeterNames = Tracer.Instance.Settings.OpenTelemetryMeterNames;
+        var enabledMeterNames = _settings.OpenTelemetryMeterNames;
```
zacharycmontoya, line 46: nit: It _might_ be more performant to make `TracerSettings.OpenTelemetryMeterNames` a `HashSet<string>` to improve the lookup time. This is a potential followup item

tracer/src/Datadog.Trace/OTelMetrics/MetricReader.cs
```diff
@@ -8,78 +8,131 @@
...
+        private readonly TracerSettings _settings;
+        private readonly MetricReaderHandler _handler;
+        private readonly MetricExporter _exporter;
+        private MeterListener? _listener;
+        private Timer? _timer;
+
+        public MetricReader(TracerSettings settings, MetricReaderHandler handler, MetricExporter exporter)
         {
-            return;
+            _settings = settings;
```
zacharycmontoya: Rather than keep a reference to the settings object, can you copy over the values you need as fields?

tracer/src/Datadog.Trace/OTelMetrics/MetricPoint.cs
```diff
@@ -84,7 +112,58 @@ private static int FindBucketIndex(double value)
...
+    /// </summary>
+    /// <returns>A new MetricPoint containing the snapshot values</returns>
+    public MetricPoint CreateSnapshotAndReset()
+    {
+        lock (_histogramLock)
+        {
+            var endTime = DateTimeOffset.UtcNow;
+
+            // Create a snapshot copy with current values
+            var snapshot = new MetricPoint(InstrumentName, MeterName, InstrumentType, AggregationTemporality, Tags, Unit, Description)
```
zacharycmontoya: nit: For follow-up, I'd like to avoid creating a new MetricPoint each time we do a snapshot. Instead, I feel like we only need to update the `Snapshot*` fields. I don't necessarily think we should block on this, but this seems like the next optimization we can make in terms of allocations

tracer/src/Datadog.Trace/OTelMetrics/MetricPoint.cs
```diff
@@ -84,7 +112,58 @@ private static int FindBucketIndex(double value)
...
+                _runningDoubleValue = 0.0;
+                _runningMin = double.PositiveInfinity;
+                _runningMax = double.NegativeInfinity;
+                if (_runningBucketCounts.Length > 0)
+                {
+                    Array.Clear(_runningBucketCounts, 0, _runningBucketCounts.Length);
+                }
+
+                // Update start time for next delta window
+                StartTime = endTime;
```
zacharycmontoya, line 198: I'm going to follow up on this, but I don't know if we are supposed to update StartTime each time we collect...I think we only need to update EndTime. I'll verify this offline and come back to this comment

tracer/src/Datadog.Trace/OTelMetrics/MetricPoint.cs
```diff
@@ -51,6 +71,14 @@ internal void UpdateCounter(double value)
         }
     }
 
+    internal void UpdateObservableCounter(double currentValue)
+    {
+        lock (_histogramLock)
+        {
+            _runningDoubleValue = currentValue;
```
zacharycmontoya, line 96: nit: For follow-up, we should be able to faithfully report both longs and doubles, so I think we'll need to do some refactoring. For example, a Counter may either be a long or a double, and we should store the intended type

tracer/src/Datadog.Trace/OTelMetrics/OtlpExporter.cs
```diff
@@ -0,0 +1,468 @@
...
+            if (metrics.Count == 0)
+            {
+                return ExportResult.Success;
+            }
+
+            TelemetryFactory.Metrics.RecordCountMetricsExportAttempts(_protocolTag);
+
+            try
+            {
+                var otlpPayload = _serializer.SerializeMetrics(metrics);
```
zacharycmontoya: Design-wise, I think `SerializeMetrics` should be called inside your `SendOtlpRequest` method since there you are making a decision between sending Protobuf or JSON content. As the code is written now, the `otlpPayload` could either be Protobuf or JSON data and then you're sending that ambiguous content to the `SendOtlpRequest` method

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/OpenTelemetrySdkTests.cs
```diff
@@ -218,20 +233,43 @@ public async Task SubmitsOtlpMetrics(string packageVersion)
 
             var initialAgentPort = TcpPortProvider.GetOpenPort();
 
-            SetEnvironmentVariable("DD_METRICS_OTEL_ENABLED", "true");
+            MockOtlpGrpcServer grpcServer = null;
+
+            string udsPath = networkType == "unix" ? "/tmp/otel-grpc-server.sock" : null;
+
+            if (udsPath is not null && OperatingSystem.IsWindows())
```
zacharycmontoya: I think this is actually supported on Windows, so we should be able to enable this scenario. One reference: https://learn.microsoft.com/en-us/aspnet/core/grpc/interprocess-uds?view=aspnetcore-9.0

tracer/test/Datadog.Trace.Tests/MeterListenerTests.cs
zacharycmontoya, line 1: This test file looks good, nice work with the InMemoryExporter!

tracer/src/Datadog.Trace/OTelMetrics/MetricPoint.cs
link04: Looking at this it seems like we can leave it as it is, WDYT?
- https://buf.build/opentelemetry/opentelemetry/docs/main:opentelemetry.proto.metrics.v1#:~:text=rejected%20by%20consumers.-,StartTimeUnixNano,-StartTimeUnixNano%20in%20general

zacharycmontoya: Ohhh I see, in the Python OTel SDK implementation, this gets updated each time for Delta aggregations (what you're doing!). But for Cumulative, it's not updated and it's just the original time. https://github.com/open-telemetry/opentelemetry-python/blob/daae23bc7f5514003d1e262730c409875d12d77d/opentelemetry-sdk/src/opentelemetry/sdk/metrics/_internal/aggregation.py#L333-L366

So this change looks good! Feel free to resolve this thread.

tracer/src/Datadog.Trace/OTelMetrics/OtlpMetricsSerializer.cs
```diff
@@ -252,28 +252,31 @@ private byte[] SerializeSum(MetricPoint metric)
...
             else
             {
+                // Counter and ObservableCounter use configured temporality and are monotonic
                 temporality = _settings.OtlpMetricsTemporalityPreference switch
                 {
-                    OtlpTemporality.Delta => 1,
-                    OtlpTemporality.LowMemory => 1,
-                    OtlpTemporality.Cumulative => 2,
-                    _ => 1
+                    OtlpTemporality.Delta => AggregationTemporality.Delta,
```
zacharycmontoya: 😍 Looks great! The only other improvement I can think of is renaming the class name for the `OtlpTemporality` enum to make the distinction a little clearer that this one is the "Preference". For people looking at this code for the first time, it might not be obvious what the difference is between the right and left side of the switch expressions. Can definitely do this as a follow-up if you want to pursue it

tracer/test/Datadog.Trace.TestHelpers/OtelProto/MetricsService.cs
zacharycmontoya, line 1: Since we're going to test the gRPC work in a separate PR, I think we can delete this file, right?

tracer/src/Datadog.Trace/OTelMetrics/OtlpExporter.cs
```diff
@@ -0,0 +1,460 @@
...
+
+            foreach (var header in _headers)
+            {
+                httpRequest.Headers.TryAddWithoutValidation(header.Key, header.Value);
+            }
+
+            return await SendWithRetry(httpRequest).ConfigureAwait(false);
+        }
+
+        private async Task<bool> SendGrpcRequest(byte[] exportRequest)
```
zacharycmontoya: The more I think about it, the more I think that we should exclude the gRPC component from this PR. Even though the gRPC protocol is opt-in, I don't think we should be rolling our own gRPC client code as I expect it to be very unreliable. Can we move the gRPC implementation into its own PR so we can merge this PR as soon as we validate the end-to-end `http/protobuf` support?

tracer/src/Datadog.Trace/OTelMetrics/OtlpMetricsSerializer.cs
zacharycmontoya, line 1: I'll just leave a file-level comment to make sure it doesn't get lost, but as a follow-up task we should vendor this code so that it's easy to update how we emit protobuf if we need to update to a newer proto file to implement. In general it seems like we're creating a bunch of intermediate byte arrays which could be optimized away, but it's not worth spending time on this if we're going to change the code soon. Otherwise, as long as we are able to emit correct OTLP, this is good for this PR

zacharycmontoya (commented):
The last overall comment I have is that we should remove the gRPC emission from this PR and focus on the `http/protobuf` serialization, but nothing else major from me!

bouwkast:
@codex review

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
```diff
@@ -376,25 +376,20 @@ internal static void InitializeNoNativeParts(Stopwatch sw = null)
...
             catch (Exception ex)
             {
-                Log.Error(ex, "Error initializing OTel Metrics Reader");
+                Log.Error(ex, "Error initializing OTel Metrics collection.");
             }
 #else
             if (Tracer.Instance.Settings.OpenTelemetryMetricsEnabled)
             {
-                Log.Information("Unable to initialize OTel Metrics collection, this is only available starting with .NET 6.0.");
+                Log.Information("Unable to initialize Opentelemetry Protocol Metrics collection, this is only available starting with .NET 6.0.");
```
bouwkast, line 392: nitpick: feel free to pick up in a next PR

```suggestion
                Log.Information("Unable to initialize OpenTelemetry Protocol Metrics collection, this is only available starting with .NET 6.0.");
```

tracer/src/Datadog.Trace/OTelMetrics/MetricStreamIdentity.cs
```diff
@@ -21,6 +23,14 @@ public MetricStreamIdentity(Instrument instrument, InstrumentType instrumentType
...
             InstrumentName = instrument.Name;
             MeterName = instrument.Meter.Name;
+            MeterVersion = instrument.Meter.Version ?? string.Empty;
+
+#if NET8_0_OR_GREATER
+            MeterTags = instrument.Meter.Tags?.ToArray() ?? [];
+#else
+            MeterTags = [];
+#endif
+
```
bouwkast: Hmm I think since we build against anything higher than .NET 6 so this always hits the `else` per my understanding

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
bouwkast: Feel free to ignore this, was reviewing initially this morning

tracer/src/Datadog.Trace/OTelMetrics/MetricPoint.cs
```diff
@@ -51,6 +75,14 @@ internal void UpdateCounter(double value)
...
```
bouwkast, line 96:
```suggestion
            _runningDoubleValue += currentValue;
```

Is it supposed to be an `=` to replace or is it supposed to be a `+=` like the `UpdateCounter` above?

Still going through the PR so this may be obvious later

bouwkast: Maybe I just don't understand what "Observable" means

tracer/src/Datadog.Trace/OTelMetrics/MetricStreamIdentity.cs
```diff
@@ -21,6 +24,19 @@ public MetricStreamIdentity(Instrument instrument, InstrumentType instrumentType
...
+            // Duck typing works at runtime - checks if Tags property exists (.NET 8+)
+            try
+            {
+                var meterDuck = instrument.Meter.DuckCast<IMeterDuck>();
+                MeterTags = meterDuck.Tags?.ToArray() ?? [];
+            }
+            catch
+            {
+                MeterTags = [];
+            }
```
bouwkast: I'd rather we not rely on throwing an exception here (in fact I don't actually know if it does throw 🤔 )

Could we swap to `TryDuckCast` or `DuckAs`?

tracer/src/Datadog.Trace/OTelMetrics/MetricState.cs
```diff
@@ -91,39 +97,47 @@ public void RecordMeasurementDouble(double value, ReadOnlySpan<KeyValuePair<stri
...
+        {
+            return existingPoint;
+        }
+
+        var dict = new Dictionary<string, object?>(tags.Length);
         for (int i = 0; i < tags.Length; i++)
         {
-            var kv = tags[i];
-            tagDict[kv.Key] = kv.Value;
+            ref readonly var kv = ref tags[i];
```
bouwkast: This looks strange to me what was the rational to go to this from?

Not saying that this is incorrect as I haven't tested this out or looked _entirely_ at the complete call chain


```suggestion
            var kv = tags[i];
```

tracer/src/Datadog.Trace/OTelMetrics/TagSet.cs
```diff
@@ -19,22 +20,63 @@ namespace Datadog.Trace.OTelMetrics;
...
-        return new TagSet(key);
+            var sb = StringBuilderCache.Acquire();
+            for (int i = 0; i < len; i++)
+            {
+                if (i > 0)
+                {
+                    sb.Append(';');
+                }
+
+                ref readonly var kv = ref sorted[i];
```
bouwkast: Hm maybe I'm missing something as there is the `ref readonly` and `ref` here too as in the other location.

Why exactly do we need to do this? Probably worthwhile to note down as a comment as it isn't immediately clear to me.

bouwkast (approved):
Just a few comments but LGTM overall

Note that I didn't inspect the snapshots very closely as I'm not super sure on what they should be 😅

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/OpenTelemetrySdkTests.cs
```diff
@@ -202,34 +213,40 @@ public async Task IntegrationDisabled(string packageVersion)
...
+        public async Task SubmitsOtlpMetrics(string packageVersion, string datadogMetricsEnabled, string otelMetricsEnabled)
         {
             var parsedVersion = Version.Parse(!string.IsNullOrEmpty(packageVersion) ? packageVersion : "1.12.0");
             var runtimeMajor = Environment.Version.Major;
 
             var snapshotName = runtimeMajor switch
             {
-                6 when parsedVersion >= new Version("1.3.2") && parsedVersion < new Version("1.5.0") => ".NET_6",
+                6 when parsedVersion >= new Version("1.3.2") && parsedVersion < new Version("1.5.0") => otelMetricsEnabled.Equals("true") ? ".NET_6_OTEL" : ".NET_6_DD",
                 7 or 8 when parsedVersion >= new Version("1.5.1") && parsedVersion < new Version("1.10.0") => ".NET_7_8",
```
bouwkast, line 225: The .NET 8 snapshots need to be split from .NET 7 as we have the metric tags in .NET 8 but not .NET 7 (causing test failures)

tracer/src/Datadog.Trace/OTelMetrics/MetricPoint.cs
link04: So _runningDoubleValue = currentValue; is correct since ObservableCounter reports cumulative values (like an odometer.

tracer/src/Datadog.Trace/OTelMetrics/MetricStreamIdentity.cs
link04: Done in my last commit:

tracer/src/Datadog.Trace/OTelMetrics/MetricState.cs
link04: Not really needed ATM just AI things that claim performance.

PR #7521, opened 2025-09-16
Upload execution benchmark log files and cleanup

andrewlock:
## Summary of changes

- Fix error trying to delete log file in use
- Upload execution benchmark log files for the job
- Make failures more obvious

## Reason for change

We want to try to understand spurious results, don't want to be confused by unrelated messages, and want to make it easier to understand failurs

## Implementation details

Tweak some scripts

## Test coverage

I'll check that this run is doing what we expect


tonyredondo (approved):
Nice!

PR #7522, opened 2025-09-16
Extract `MutableSettings` from `TracerSettings`

andrewlock:
## Summary of changes

- Extract a `MutableSettings` type from `TracerSettings`
- Extract a `Raw` type from `ExporterSettings`

## Reason for change

We're working on refactoring how we handle dynamic/remote/config in code settings i.e. settings which can change at runtime. As a first step, this PR extracts those settings to their own type, called `MutableSettings` because they mutable during the lifetime of the app.

> Feel free to suggest other names for this type, or we can alternatively bikeshed it later.

Additionally, extracted the "raw" settings from `ExporterSettings`. These are the values which are read from config sources. The actual values of `ExporterSettings` are set based on these values, using a highly convoluted (backward compatible) series of methods, but the idea is: if the `Raw` settings haven't changed, the `ExporterSettings` haven't changed.

> This isn't _strictly_ true due to the `File.Exists` call we have, but I believe this is good enough for our purposes.

## Implementation details

- Create `MutableSettings` and move all the properties from `TracerSettings` that _can_ change to it.
- Update `TracerSettings` to create an instance of `MutableSettings`, and simply pass-through properties to it.
  - This should mean existing functionality is unaffected by this PR
- Implement `IEquatable<MutableSettings>` for future comparisons between `MutableSettings` instances
  - Unfortunately, can't use a `record` here or auto-gen the implementation, because we need to handle equivalence of the dictionaries.
- Extract the "raw" setting reading to an `ExporterSettings` nested-type
  - Opted for nested here, because unlike `MutableSettings` (which will eventually live separately from `TracerSettings`) we won't expose `Raw` to consumers -  they'll still use `ExporterSettings`

## Test coverage

This is just a refactoring, so it's covered by existing tests.

Additionally, I added a test for the `IEquatable` implementation (which is similar to the test we have for `ImmutableDynamicSettings`) to ensure the implementation is updated if we add more properties.

## Other details

Part of a config stack

### DIFF
--- a/tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
+++ b/tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
@@ -84,52 +84,47 @@ namespace Datadog.Trace.Configuration
         /// Direct use in tests only.
         /// </summary>
         internal ExporterSettings(IConfigurationSource? source, Func<string, bool> fileExists, IConfigurationTelemetry telemetry)
+            : this(new Raw(source ?? NullConfigurationSource.Instance, telemetry), fileExists, telemetry)
+        {
+        }
+
+        internal ExporterSettings(Raw rawSettings, Func<string, bool> fileExists, IConfigurationTelemetry telemetry)
         {
             _fileExists = fileExists;
             _telemetry = telemetry;
+            RawSettings = rawSettings;
 
             ValidationWarnings = new List<string>();
 
-            source ??= NullConfigurationSource.Instance;
+            var traceSettings = GetTraceTransport(
+                agentUri: rawSettings.TraceAgentUri,
+                tracesPipeName: rawSettings.TracesPipeName,
+                agentHost: rawSettings.TraceAgentHost,
+                agentPort: rawSettings.TraceAgentPort,
+                tracesUnixDomainSocketPath: rawSettings.TracesUnixDomainSocketPath);
 
-            // Get values from the config
-            var config = new ConfigurationBuilder(source, telemetry);
-            var traceAgentUrl = config.WithKeys(ConfigurationKeys.AgentUri).AsString();
-            var tracesPipeName = config.WithKeys(ConfigurationKeys.TracesPipeName).AsString();
-            var tracesUnixDomainSocketPath = config.WithKeys(ConfigurationKeys.TracesUnixDomainSocketPath).AsString();
-
-            var agentHost = config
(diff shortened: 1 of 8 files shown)


andrewlock:
@codex review

andrewlock:
@codex review

bouwkast (approved):
LGTM 

I've assumed most of the code was just a move with the equality / Raw stuff being new which looked fine to me (and the new tests)

How often does the new big equality check run? I don't think it is very often right? Thinking in terms of performance that you mentioned

andrewlock:
> LGTM
> 
> I've assumed most of the code was just a move with the equality / Raw stuff being new which looked fine to me (and the new tests)

Thanks, yes, I meant to add comments to the PR about that and totally forgot, sorry 🤦‍♂️ 

> How often does the new big equality check run? I don't think it is very often right? Thinking in terms of performance that you mentioned

As of _this_ PR it doesn't really ever run, but in subsequent PRs it runs every time you do configuration in code, or every time there's a dynamic/remote config update. Overall, that should be relatively rare, and there's a lot else going on, so it's not hot path. Obviously we should optimise what we can, but it's not critical path 🙂

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -84,52 +84,47 @@ internal ExporterSettings(IConfigurationSource? source, IConfigurationTelemetry
...
+
+        internal ExporterSettings(Raw rawSettings, Func<string, bool> fileExists, IConfigurationTelemetry telemetry)
         {
             _fileExists = fileExists;
             _telemetry = telemetry;
+            RawSettings = rawSettings;
 
             ValidationWarnings = new List<string>();
 
-            source ??= NullConfigurationSource.Instance;
```
andrewlock, line 93: This config is all just moved as-is to the `Raw` nested type

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -136,178 +118,14 @@ internal TracerSettings(IConfigurationSource? source, IConfigurationTelemetry te
                 AzureAppServiceMetadata = new ImmutableAzureAppServiceSettings(source, _telemetry);
             }
 
-            var otelTags = config
```
andrewlock, line 139: This is all moved as-is to `MutableSettings`

lucaspimentel (approved):

PR #7530, opened 2025-09-18
Move `DefaultServiceName` to `MutableSettings`

andrewlock:
## Summary of changes

- Expose `DefaultServiceName` on `MutableSettings` instead of on `TracerManager`
- Expose `MutableSettings` on `PerTraceSettings`

## Reason for change

The `DefaultServiceName` depends on `ServiceName`, which can change at runtime, so `MutableSettings` seems like the best place for it (eventually we will only have a single `TracerManager` and `TracerSettings` per lifetime

> Note that this also solves an existing edge-case bug when customers do config in code and already-finished traces are serialized with the "incorrect" default service name.

## Implementation details

Mostly commit-by-commit but:

- Move the "fallback application name" calculation to a helper class
- Expose the fallback application name on `TracerSettings`
  - Created as a `Lazy<>` because this calculation can be kind of expensive, and isn't necessary if the customer specifies a service name.
  - Exposed on `TracerSettings` because it doesn't change
- Expose `DefaultServiceName` on `MutableSettings` instead of `TracerManager`.
- Update usages to point to the new location

Additionally, there are some places where I believe we were "incorrectly" using the `DD_SERVICE` value, and ignoring the real fallback name. 
- [ ] Service Discovery's `StoreTracerMetadata` - @anna-git can you confirm if I'm correct that this _should_ be using the "calculated" service name as a fallback?
- [ ] `TraceExporterConfiguration` - @ganeshnj, same question, can you confirm that we should be passing the "calculated" service name, not just the "explictly set" service name?

One additional aspect I think we should consider: 
- Currently we're calling `NormalizeService(serviceName)` in a few places. Is there any reason we shouldn't be _always_ normalizing the service name?
- e.g. why don't we normalize `DefaultServiceName` automatically?

## Test coverage

Covered by existing tests generally

## Other details

Part of a config stack


andrewlock:
@codex review

tracer/src/Datadog.Trace/LibDatadog/ServiceDiscovery/ServiceDiscoveryHelper.cs
```diff
@@ -36,7 +36,7 @@ internal static StoreMetadataResult StoreTracerMetadata(TracerSettings tracerSet
                     TracerConstants.Language,
                     TracerConstants.ThreePartVersion,
                     Environment.MachineName,
-                    tracerSettings.ServiceName,
+                    tracerSettings.MutableSettings.DefaultServiceName,
```
anna-git, line 39: so this is called only once at start, I'm having trouble understanding what the difference is here if you have any practical examples 😅 
I though that ServiceName was the actual computed service name and that it's what we want, but I'm not sure what could be the difference between calculated / default service name actually 🤯

anna-git: image

from [the rfc](https://docs.google.com/document/d/1kcW6BLdYxXeTSUz31cBqoqfW1Jjs0IDljfKeUfIRQp4/edit?tab=t.0#heading=h.zgll08erod9d)
🤔

andrewlock: > I though that ServiceName was the actual computed service name and that it's what we want, but I'm not sure what could be the difference between calculated / default service name actually 🤯

Nope, `ServiceName` is _only_ what the customer provides in `DD_SERVICE` 😄  So it's null if they don't set that variable

`DefaultServiceName` is the calculated name that gives a fallback, e.g. `MyApp` if it's `MyApp.exe`

It sounds like `DefaultServiceName` _is_ what you want to me 🙂

anna-git: yep all good then thanks for fixing it I got confused! wondering if `defaultServiceName` should be named like `FinalServiceName` or `CalculatedServiceName` and service name maybe `ProvidedServiceName` but it's a detail

andrewlock: Yeah, I agree, renaming them to make it clear makes a lot of sense - I'll leave that to a later PR though, thanks!

dudikeleti (approved):

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -32,12 +33,10 @@ namespace Datadog.Trace.Configuration
     public record TracerSettings
     {
         private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor<TracerSettings>();
-        private static readonly HashSet<string> DefaultExperimentalFeatures = new HashSet<string>()
-        {
-            "DD_TAGS"
-        };
+        private static readonly HashSet<string> DefaultExperimentalFeatures = ["DD_TAGS"];
```
tonyredondo, line 36: ❤️

PR #7531, opened 2025-09-18
Hide problematic public APIs using `[EditorBrowsable]`

andrewlock:
## Summary of changes

Hides some of the public APIs exposed in Datadog.Trace that we don't think customer should be using.

## Reason for change

There are several configuration APIs that customers can call using configuration in code which are problematic:

- `new TracerSettings()` - ignores all env vars and default sources. Customers should generally be calling `TracerSettings.FromDefaultSources()`, it's unlikely that they really _want_ this behavior. It also makes our code much more complex.
- `new TracerSettings(bool useDefaultSources)` - as above
- `TracerSettings.set_StatsComputationEnabled` - changing this at runtime is problematic, particularly with the move to libdatadog data pipeline, and at some point soon we'll basically be ignoring it 
- `TracerSettings.set_AgentUri` - changing this at runtime means everything has to reinitialialize, and customers likely end up with incorrect values for some period of time, as well as increased startup times
- `TracerSettings.Exporter.set_AgentUri` - as above

We would _like_ to obsolete and remove these properties, but that takes two major versions to do (one to obsolete them, one to remove them). This at least ensures that _new_ customers shouldn't use these APIs.

## Implementation details

Add `[Browsable(false)]` and `[EditorBrowsable(EditorBrowsableState.Never)]` to the APIs.

## Test coverage

There's no change in functionality, so nothing really to test. Updated the public APIs. This requires an additional in-box reference for .NET Core 3.1 + .NET 6, but I think that's fine.

## Other details

### DIFF
--- a/tracer/src/Datadog.Trace.Manual/Configuration/ExporterSettings.cs
+++ b/tracer/src/Datadog.Trace.Manual/Configuration/ExporterSettings.cs
@@ -24,6 +24,11 @@ public sealed class ExporterSettings
     /// Gets or sets the Uri where the Tracer can connect to the Agent.
     /// Default is <c>"http://localhost:8126"</c>.
     /// </summary>
+    /// <remarks>As of tracer version 3.27.0, this property cannot be used to set the
+    /// agent URI. You must instead use a static configuration source such
+    /// as environment variables or datadog.json to set the value instead. This
+    /// property will be marked obsolete and removed in a future version of Datadog.Trace.
+    /// </remarks>
     [Obsolete("This property is obsolete and will be removed in a future version. To set the AgentUri, use the TracerSettings.AgentUri property")]
     public Uri AgentUri
     {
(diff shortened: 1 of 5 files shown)


tracer/src/Datadog.Trace.Manual/Configuration/ExporterSettings.cs
```diff
@@ -26,6 +28,8 @@ internal ExporterSettings(TracerSettings tracerSettings)
     public Uri AgentUri
     {
         get => _tracerSettings.AgentUri;
+        [Browsable(false)]
```
lucaspimentel, line 36: Should we add a comment to the xml docs, too? "DON'T USE THIS!!!"
Someone might actually see it in their IntelliSense pop-up.

andrewlock: > Should we add a comment to the xml docs, too? "DON'T USE THIS!!!"
> Someone might actually see it in their IntelliSense pop-up.

Well, the point is that they _don't_ see it in their IntelliSense pop-up 😄 But I agree we could/should

andrewlock: Done

PR #7534, opened 2025-09-18
Add a Continuous Profiler availability check

andrewlock:
## Summary of changes

Add a checker to decide whether the profiler is available

## Reason for change

We are going to be (and are currently) making P/Invoke calls on startup to call the profiler. If the ClrProfiler is not attached, or the CP file is not available (and so P/Invokes are not re-written) then this can cause additional delays.

## Implementation details

Add explicit checks for
- Unsupported environments
- Has the profiler set the "profiler available" env var (only works on Windows right now)
- Are we in known unsupported environments (lambda and Azure functions outside of AAS extension)
- Is the ClrProfiler attached

## Test coverage

If everything still works, this _should_ work

## Other details

A blocker for #7287

### DIFF
--- a/tracer/src/Datadog.Trace/ContinuousProfiler/ProfilerAvailabilityHelper.cs
+++ b/tracer/src/Datadog.Trace/ContinuousProfiler/ProfilerAvailabilityHelper.cs
@@ -18,7 +18,18 @@ internal static class ProfilerAvailabilityHelper
     // We should add or remove conditions from here as our deployment requirements change.
     // Longer term, we'd like to be able to pass this information from the native side to the managed side, but
     // today that only works on Windows (hence the early short-circuit).
-    private static readonly Lazy<bool> ProfilerIsAvailable = new(() =>
+    private static readonly Lazy<bool> ProfilerIsAvailable = new(() => GetIsContinuousProfilerAvailable(EnvironmentHelpersNoLogging.IsClrProfilerAttachedSafe));
+
+    /// <summary>
+    /// Gets a value indicating whether returns true if the continuous profiler _should_ be available
+    /// </summary>
+    public static bool IsContinuousProfilerAvailable => ProfilerIsAvailable.Value;
+
+    // Internal for testing
+    internal static bool IsContinuousProfilerAvailable_TestingOnly(Func<bool> isClrProfilerAttached)
+        => GetIsContinuousProfilerAvailable(isClrProfilerAttached);
+
+    private static bool GetIsContinuousProfilerAvailable(Func<bool> isClrProfilerAttached)
     {
         // Profiler is not available on ARM(64)
         var fd = FrameworkDescription.Instance;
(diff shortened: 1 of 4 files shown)


tracer/src/Datadog.Trace/ContinuousProfiler/ProfilerStatus.cs
```diff
@@ -63,6 +65,12 @@ private void EnsureNativeIsIntialized()
 
                 _isInitialized = true;
 
+                if (!ProfilerAvailabilityHelper.IsContinuousProfilerAvailable)
+                {
+                    Log.Debug("The continuous profiler is not available in this environment.");
```
chrisnas: Don't we want a log all the time because it could help quickly investigating possible issues without asking for debug logs?

chrisnas (approved):
LGTM with just a log level question

tracer/src/Datadog.Trace/ContinuousProfiler/ProfilerAvailabilityHelper.cs
```diff
@@ -0,0 +1,58 @@
...
+        // on Windows at the moment. We assume that the CLR profiler must be attached in this scenario
+        if (fd.IsWindows() && !string.IsNullOrEmpty(EnvironmentHelpers.GetEnvironmentVariable("DD_INTERNAL_PROFILING_NATIVE_ENGINE_PATH")))
+        {
+            return true;
+        }
+
+        // Now we're into fuzzy territory. The CP is not available in some environments
+        // - AWS Lambda
+        // - Azure Functions where the site extension is _not_ used
+        var isUnsupported = EnvironmentHelpers.IsAwsLambda()
```
anna-git, line 52: is it ok if it tries logging here? as this ends up [there](https://github.com/DataDog/dd-trace-dotnet/blob/d0d795bd1aa817e41d9ce555a2747c1eb3d5f642/tracer/src/Datadog.Trace/Util/EnvironmentHelpers.cs#L67)

 otherwise there's this [beautiful solution](https://github.com/DataDog/dd-trace-dotnet/blob/303c4f9d3dc4e4ac35ccd1094a7f7e254663ebce/tracer/src/Datadog.Trace/Util/EnvironmentHelpersNoLogging.cs#L41) 😅

tracer/src/Datadog.Trace/ContinuousProfiler/ProfilerStatus.cs
andrewlock: Yeah, I was torn, but it probably makes sense 👍

andrewlock:
```suggestion
                    Log.Information("The continuous profiler is not available in this environment.");
```

tracer/src/Datadog.Trace/ContinuousProfiler/ProfilerAvailabilityHelper.cs
andrewlock: Yeah, I originally used your solution, but I can't use `IsServerless()` as it excludes too much. Also, this isn't used in the critical config loading paths, so I think it's ok for us to log if we _have_ to 🙂

anna-git: ok :) also, you probably know already, FrameworkDescription has a static logger field 😅

PR #7536, opened 2025-09-18
[ER Default-On] DEBUG-4406 Support multi config mering with priorities

dudikeleti:
## Summary of changes
Implements priority-based merging for APM_TRACING config with determenistic order:
Priority Order: Service+Env (4) > Service (3) > Env (2) > Wildcard (1) > Org (0)

## Reason for change
When multiple configs are sent, the "last one wins." This causes non-deterministic behavior across tracers and inconsistent feature enablement

## Test coverage
ApmTracingConfigMergerTests.cs


tracer/src/Datadog.Trace/RemoteConfigurationManagement/RcmCapabilitiesIndices.cs
```diff
@@ -97,6 +97,8 @@ internal static class RcmCapabilitiesIndices
 
         public static readonly BigInteger AsmTraceTaggingRules = Create(43);
 
+        public static readonly BigInteger CapabilityAsmExtendedDataCollection = Create(44);
```
anna-git, line 100: how is this one related with APM_TRACING?

tracer/src/Datadog.Trace/Configuration/DynamicConfigurationManager.cs
```diff
@@ -47,6 +49,7 @@ public void Start()
                 _subscriptionManager.SetCapability(RcmCapabilitiesIndices.ApmTracingSampleRate, true);     // 12
                 _subscriptionManager.SetCapability(RcmCapabilitiesIndices.ApmTracingTracingEnabled, true); // 19
                 _subscriptionManager.SetCapability(RcmCapabilitiesIndices.ApmTracingSampleRules, true);    // 29
+                _subscriptionManager.SetCapability(RcmCapabilitiesIndices.ApmTracingMulticonfig, true);    // 44
```
anna-git:
```suggestion
                _subscriptionManager.SetCapability(RcmCapabilitiesIndices.ApmTracingMulticonfig, true);    // 45
```

tracer/src/Datadog.Trace/Configuration/ApmTracingConfig.cs
```diff
@@ -0,0 +1,184 @@
...
+            return serviceMatches && envMatches;
+        }
+
+        /// <summary>
+        /// Merges this configuration with another, giving priority to the higher priority config
+        /// </summary>
+        public ApmTracingConfig MergeWith(ApmTracingConfig other)
+        {
+            var higherPriority = this.Priority >= other.Priority ? this : other;
+            var lowerPriority = this.Priority >= other.Priority ? other : this;
```
anna-git, line 83: nit
```suggestion
            var (higherPriority, lowerPriority) = this.Priority >= other.Priority ? (this, other) : (other, this);
```

tracer/src/Datadog.Trace/Configuration/DynamicConfigurationManager.cs
```diff
@@ -114,65 +117,86 @@ private static void OnConfigurationChanged(ConfigurationBuilder settings)
...
 
-                for (int i = 0; i < apmLibrary.Count; i++)
-                {
-                    result[i] = ApplyDetails.FromError(apmLibrary[i].Path.Path, ex.ToString());
-                }
+        private void ApplyMergedConfiguration()
+        {
+            // Get current service/environment for filtering
+            var currentSettings = Tracer.Instance.Settings;
+            var serviceName = currentSettings.ServiceName ?? "unknown";
```
anna-git: do we want the `DefaultServiceName` here or the provided service name through `DD_SERVICE`?  cc @andrewlock

tracer/src/Datadog.Trace/Configuration/DynamicConfigurationManager.cs
```diff
@@ -114,65 +117,86 @@ private static void OnConfigurationChanged(ConfigurationBuilder settings)
...
+            var environment = currentSettings.Environment ?? "unknown";
 
-                return result;
-            }
+            var mergedConfigJson = ApmTracingConfigMerger.MergeConfigurations(
```
anna-git: in `MergeConfigurations` we deserialize and reserialize something that's gonna be deserialized again within `DynamicConfigConfigurationSource`. Could we not integrate the merge logic inside DynamicConfigConfiguratioSource directly? like around here https://github.com/DataDog/dd-trace-dotnet/blob/7e24fb283e48468c1febdbb78b6d12ef942043f2/tracer/src/Datadog.Trace/Configuration/ConfigurationSources/DynamicConfigConfigurationSource.cs#L43

this way we could deserialize once and treat the configurations directly

tracer/src/Datadog.Trace/Configuration/ApmTracingConfig.cs
```diff
@@ -0,0 +1,184 @@
...
+        private static LibConfig MergeLibConfigs(LibConfig higher, LibConfig lower)
+        {
+            return new LibConfig
+            {
+                TracingEnabled = higher.TracingEnabled ?? lower.TracingEnabled,
+                LogInjectionEnabled = higher.LogInjectionEnabled ?? lower.LogInjectionEnabled,
+                TracingSamplingRate = higher.TracingSamplingRate ?? lower.TracingSamplingRate,
+                TracingSamplingRules = higher.TracingSamplingRules ?? lower.TracingSamplingRules,
+                TracingHeaderTags = higher.TracingHeaderTags ?? lower.TracingHeaderTags,
+                TracingTags = higher.TracingTags ?? lower.TracingTags,
```
bouwkast, line 101: It isn't clear to me why we don't also merge the other properties: `DebugEnabled`, `RuntimeMetricsEnabled`, `ServiceMapping`, `DataStreamsEnabled`, `SpanSamplingRules`

The Java implementation merges all of them from what I can tell https://github.com/DataDog/dd-trace-java/blob/274d1448c251f13a88b72e7438dfb8a2c8474213/dd-trace-core/src/main/java/datadog/trace/core/TracingConfigPoller.java#L415

tracer/src/Datadog.Trace/Configuration/ApmTracingConfigMerger.cs
```diff
@@ -0,0 +1,99 @@
...
+                return "{\"lib_config\":{}}";
+            }
+
+            applicableConfigs.Sort((a, b) =>
+            {
+                var priorityComparison = b.Priority.CompareTo(a.Priority); // Descending
+                return priorityComparison != 0 ? priorityComparison : string.Compare(a.ConfigId, b.ConfigId, StringComparison.Ordinal); // Ascending
+            });
+
+            var mergedConfig = applicableConfigs.Aggregate((current, next) => current.MergeWith(next));
```
bouwkast: I find this is a bit difficult to follow in my opinion and would be beneficial to just show that it is expected to do in a comment

tracer/src/Datadog.Trace/Configuration/ApmTracingConfigMerger.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            foreach (var config in configs)
+            {
+                try
+                {
+                    var jsonContent = Encoding.UTF8.GetString(config.Contents);
+                    var configData = ParseConfiguration(config.Path.Id, jsonContent);
+
+                    // Filter immediately during parsing
+                    if (configData?.Matches(serviceName, environment) == true)
```
bouwkast, line 52: I don't see this in the Java implementation? https://github.com/DataDog/dd-trace-java/pull/9360

I do see it in the Python implementation- https://github.com/DataDog/dd-trace-py/pull/14364/files

Maybe I'm not seeing that logic in the Java implementation

tracer/src/Datadog.Trace/Configuration/ApmTracingConfig.cs
```diff
@@ -0,0 +1,184 @@
...
+                var hasCluster = ClusterTarget != null;
+
+                return new ConfigurationTarget(hasService, hasEnv, hasCluster) switch
+                {
+                    (true, true, _) => 5, // Service+env (highest priority)
+                    (true, false, _) => 4, // Service only
+                    (false, true, _) => 3, // Env only
+                    (false, false, true) => 2, // Cluster
+                    (false, false, false) => 1 // Org level
+                };
```
bouwkast: It appears the Python implementation has more granularity than this 

https://github.com/DataDog/dd-trace-py/blob/8ed7a3e93e4b892566b1c80f44346665890681a4/ddtrace/internal/remoteconfig/products/apm_tracing.py#L34-L40


```python
    # Precedence ordering goes from most specific to least specific:
    # 1. Service, 2. Env, 3. Cluster target, 4. Wildcard `*`
    return (
        ((service is not None and service != "*") << 2)
        | ((env is not None and env != "*") << 1)
        | (cluster_target is not None) << 0
    )
```
It seems they are taking into account each possible combination, should we do that too?

tracer/src/Datadog.Trace/Configuration/ApmTracingConfig.cs
dudikeleti: I like it too but does not work in .net461 (am I missing something?)

tracer/src/Datadog.Trace/RemoteConfigurationManagement/RcmCapabilitiesIndices.cs
dudikeleti: It's not. I had to do this to make [test](https://github.com/DataDog/dd-trace-dotnet/blob/c6bd674a30577738e28c2ae85ed74476fa8e768e/tracer/test/Datadog.Trace.Tests/Configuration/RcmCapabilitiesTests.cs#L19) pass. The bit number is taken from another tracer implementation. If this is wrong, I'll handle it differently.

tracer/src/Datadog.Trace/Configuration/ApmTracingConfig.cs
dudikeleti: Good catch. I've added those properties later and forgot to update that code. Fixed in 20fe7257ed99147aea95c407163a544803573a81

tracer/src/Datadog.Trace/Configuration/ApmTracingConfig.cs
dudikeleti: Sure, thanks for pointing out. Fixed in a6023fe9ff08ea455fbcfe8e44a7c6957dfb205f

tracer/src/Datadog.Trace/Configuration/ApmTracingConfigMerger.cs
dudikeleti: I changed it to the current behaviour (as Java does) until we have an agreement on what what we should do. Fixed in 1599052dd1c0725a7b999add5abee25fe8c41f31

tracer/src/Datadog.Trace/Configuration/ApmTracingConfig.cs
lucaspimentel: I think this syntax uses `ValueTuple` which was added in net47.

tracer/src/Datadog.Trace/Configuration/ApmTracingConfigMerger.cs
dudikeleti: Fixed in d50b9ded5a9af40949cf32e1a251e508484b01c4

tracer/src/Datadog.Trace/Configuration/DynamicConfigurationManager.cs
dudikeleti: Please see updated solution 66af9cfb627d5f77bdc2db6f0492fcba41b388fa
I Can integrate it into DynamicConfigConfigurationSource but I'm afraid that would be too much for configuratioin sousrce. Anyway let me know WDYT

tracer/src/Datadog.Trace/Configuration/ApmTracingConfig.cs
anna-git: ah yes true 😓 , I meant also not testing twice `this.Priority >= other.Priority` but it's very nit

tracer/src/Datadog.Trace/Configuration/DynamicConfigurationManager.cs
dudikeleti: Fixed in d29f9166e07d6903bdb19d82afe73b56134c1acb

bouwkast (approved):
👍

PR #7537, opened 2025-09-18
[Azure Functions] Add Nuke targets to build local nuget for testing

lucaspimentel:
## Summary of changes

Adds a Nuke target to download `Datadog.Trace.Bundle` and extract it into the local bundle home directory.

## Reason for change

This is useful for building `Datadog.Trace.Bundle` or `Datadog.AzureFunctions` nupkg packages locally for testing, and I do it manually too often. Automate all the things?

## Implementation details

Add Nuke target `DownloadBundleNugetFromBuild`:
- download `Datadog.Trace.Bundle` from an Azure DevOps build
- extracts the files into `tracer/src/Datadog.Trace.Bundle/home`

Usage:
```shell
# download a fresh copy of the bundle files from Azure DevOps
.\tracer\build.ps1 DownloadBundleNugetFromBuild --BuildId 187203
```

## Test coverage

Tested locally. Not used in production builds.

### DIFF
--- a/tracer/build/_build/Build.Utilities.cs
+++ b/tracer/build/_build/Build.Utilities.cs
@@ -493,12 +493,9 @@ partial class Build
                     }
                 });
 
-       Target UpdateAzureFunctionsNugetFromBuild => _ => _
-        .Description("Downloads Datadog.AzureFunctions package from Azure DevOps, " +
-                     "updates the bundle contents with a local build of Datadog.Trace.dll, " +
-                     "and rebuilds the package")
+       Target DownloadBundleNugetFromBuild => _ => _
+        .Description("Downloads Datadog.Trace.Bundle package from Azure DevOps and extracts it to the local bundle home directory")
         .Requires(() => BuildId)
-        .Triggers(BuildAzureFunctionsNuget)
         .Executes(async () =>
         {
             if (!int.TryParse(BuildId, out var buildNumber))
@@ -506,17 +503,15 @@ partial class Build
                 throw new InvalidParametersException("BuildId should be an int");
             }
 
-            const string artifactName = "azurefunctions-nuget-package";
+            const string artifactName = "bundle-nuget-package";
 
-            var tempRoot = TemporaryDirectory / "azurefunctions-local-build";
+            var tempRoot = TemporaryDirectory / "bundle-nuget";
             var downloadDirectory = tempRoot / "download";
             var packageExtractionDirectory = tempRoot / "package";
-            var publishDirectory = tempRoot / "publish";
 
             EnsureCleanDirectory(tempRoot);
             EnsureExistingDirectory(downloadDirectory);
             EnsureExistingDirectory(packageExtractionDirectory);
-            EnsureExistingDirectory(publishDirectory);
 
             using var connection = new VssConnection(
                 new Uri(AzureDevopsOrganisation),
@@ -533,11 +528,11 @@ partial class Build
 
             var artifactDirectory = downloadDirectory / artifact.Name;
 
-            var packageFile = artifactDirectory.GlobFiles("Datadog.AzureFunctions*.nupkg").FirstOrDefault();
+            var packageFile = artifactDirectory.GlobFiles("Datadog.Trace.Bundle.*.nupkg").FirstOrDefault();
 
             if (packageFile is null)
             {
-                throw new Exception($"Datadog.AzureFunctions package was not found in artifact '{artifact.Name}'.");
+                throw new Exception($"Datadog.Trace.Bundle package was not found in artifact '{artifact.Name}'.");
             }
 
             EnsureCleanDirectory(packageExtractionDirectory);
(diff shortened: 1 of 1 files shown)


lucaspimentel:
@codex review

tracer/build/_build/Build.Utilities.cs
```diff
@@ -493,6 +493,97 @@ bool IsTestApplication(Project x)
...
+                throw new Exception($"Could not locate datadog content folder in extracted package at '{packageExtractionDirectory}'.");
+            }
+
+            EnsureCleanDirectory(BundleHomeDirectory);
+            CopyDirectoryRecursively(contentDirectory, BundleHomeDirectory, DirectoryExistsPolicy.Merge, FileExistsPolicy.Overwrite);
+        });
+
+       Target UpdateAzureFunctionsNugetFromBuild => _ => _
+        .Description("Updates the bundle home contents with local builds of Datadog.Trace.dll, and rebuilds the Datadog.AzureFunctions package")
+        .DependsOn(DownloadBundleNugetFromBuild)
```
lucaspimentel: TODO: change this so it does run `DownloadBundleNugetFromBuild` every time. I want to run `DownloadBundleNugetFromBuild` manually when I need to refresh from a CI build, but otherwise I'll run `UpdateAzureFunctionsNugetFromBuild` multiple time from a single download.

tracer/build/_build/Build.Utilities.cs
```diff
@@ -493,6 +493,97 @@ bool IsTestApplication(Project x)
...
+            }
+
+            EnsureCleanDirectory(BundleHomeDirectory);
+            CopyDirectoryRecursively(contentDirectory, BundleHomeDirectory, DirectoryExistsPolicy.Merge, FileExistsPolicy.Overwrite);
+        });
+
+       Target UpdateAzureFunctionsNugetFromBuild => _ => _
+        .Description("Updates the bundle home contents with local builds of Datadog.Trace.dll, and rebuilds the Datadog.AzureFunctions package")
+        .After(DownloadBundleNugetFromBuild)
+        .Triggers(BuildAzureFunctionsNuget)
```
andrewlock: We generally try to avoid `Triggers` because it makes it even harder to understand the already complex web than we currently have 🙁 

I would also recommend _against_ having a completely separate way of building Datadog.Trace here. Could this target be rewritten as the following?

```csharp
Target UpdateAzureFunctionsNugetFromBuild => _ => _
        .Description("Updates the bundle home contents with local builds of Datadog.Trace.dll, and rebuilds the Datadog.AzureFunctions package")
        .After(DownloadBundleNugetFromBuild)
        .DependsOn(Restore, CompileManagedSrc, PublishManagedTracer);
```

(You would need to make sure `DownloadBundleNugetFromBuild` has a `Before(PublishManagedTracer)` to ensure it downloads before the publish)

lucaspimentel: On second thought, I'll only add `DownloadBundleNugetFromBuild` for now and remove this `UpdateAzureFunctionsNugetFromBuild`. Downloading the artifact and extracting it is the more annoying part.

I can always add this second target later if needed.

tracer/build/_build/Build.Utilities.cs
```diff
@@ -493,6 +493,62 @@ bool IsTestApplication(Project x)
                     }
                 });
 
+       Target DownloadBundleNugetFromBuild => _ => _
+        .Description("Downloads Datadog.Trace.Bundle package from Azure DevOps and extracts it to the local bundle home directory." +
+                     " Useful for building Datadog.Trace.Bundle or Datadog.AzureFunctions nupkg packages locally.")
```
bouwkast, line 498: nit I think it changed to just be this?

```suggestion
                     " Useful for building the Datadog.AzureFunctions nupkg package locally.")
```

bouwkast (approved):

tracer/build/_build/Build.Utilities.cs
lucaspimentel: This new target downloads `Datadog.Trace.Bundle.nupkg` and extracts the files where they are expected to build either of those two nuget packages with `BuildBundleNuget` or `BuildAzureFunctionsNuget`.

My focus is building `Datadog.AzureFunctions`, not the bundle, but I'm trying to keep this target generic enough that it can be used for either one.

lucaspimentel: in other words, you can use either
```
# new target added in this PR
.\tracer\build.ps1 DownloadBundleNugetFromBuild --BuildId 187203

...do stuff to replace some bundle files...

# build new bundle nuget
.\tracer\build.ps1 BuildBundleNuget
```
or
```
# new target added in this PR
.\tracer\build.ps1 DownloadBundleNugetFromBuild --BuildId 187203

...do stuff to replace some bundle files...

# build new azfunc nuget
.\tracer\build.ps1 BuildAzureFunctionsNuget
```

Issue #6749, opened 2025-03-07
[FEATURE]: Add support to filtering traces


ffissore:
### Framework or Library Name

Any

### Describe the feature you'd like

In languages such as python, it's easy to [filter out traces](https://docs.datadoghq.com/tracing/guide/ignoring_apm_resources/?tab=kuberneteshelm&code-lang=python) before they reach the agent. The .NET library doesn't support filtering traces: can it be added to the roadmap?

Filtering traces is crucial when you're monitoring your endpoints overall latencies and you have a `/healthcheck` endpoint that is screwing up your average. Such endpoint has micro-seconds lantencies, while your other endpoints are in the millis.
This is even worse while developing new services. "app" endpoints receive little traffic, while the healthcheck one gets a constant rate of calls. Anomaly monitors from the "dev" environment will trigger for the wrong reason, making it harder for the team to evaluate them.

link04:
Hello @ffissore, you are correct that the .NET Tracer doesn't have a Post Processing option like [Python does](https://docs.datadoghq.com/tracing/guide/ignoring_apm_resources/?tab=kuberneteshelm&code-lang=python#tracer-configuration-options), but when it comes to managing which traces are sent to Datadog, you have a different option, for example the one below.

You can [Configure the Agent to Ignore Specific Resources](https://docs.datadoghq.com/tracing/guide/ignoring_apm_resources/?tab=kuberneteshelm&code-lang=python#ignoring-based-on-resources): You can set the `DD_APM_IGNORE_RESOURCES` environment variable or update the `ignore_resources` setting in the Agent's datadog.yaml configuration file. This allows the Agent to exclude traces from specified resources based on their names.
```  
ignore_resources: ["(GET|POST) /healthcheck","API::NotesController#index"]
```

If your use case requires a different type of approach please respond with the details so we can look into it. Thanks!

ffissore:
Hi @link04, I've read about that option and it would solve our problem.
The issue with it is that datadog agents are managed by an entirely different team in my org, so applying changes there takes much longer.
Having a postprocessing/filtering option at the library level would give us more freedom and reduce friction.

For now, we have worked around the issue by filtering out the health check endpoints in our monitors.

andrewlock:
Given the workaround would solve your problem, and we don't have plans to expose a programmatic way to do this filtering at the moment, I'll close this issue for now. Thanks!

PR #7540, opened 2025-09-20
Add root span filter support to Tracer

sorenhansendk:
## Summary of changes

- Added `Tracer.AddRootSpanFilter()` in TracerManager to register per-root-span drop predicates before traces reach the agent
- Exposed the API from the manual tracer, plus a calltarget hook (AddRootSpanFilterIntegration) so automatic instrumentation forwards registrations to the automatic tracer

## Reason for change

- Applications need an in-process hook to suppress traces with unwanted root-span resources (e.g., health checks), matching the capability already offered in other languages

## Implementation details

- `TracerManager` now keeps a thread-safe list of root-span filters and evaluates them once per completed trace before writing to the agent
- `Tracer.AddRootSpanFilter()` records public API usage then delegates to a new AddRootSpanFilterInternal() used by the calltarget integration
- Manual tracer gained a no-op implementation (with argument validation) so automatic instrumentation can safely bind, and a new InstrumentMethod-decorated integration handles forwarding to the automatic tracer across target frameworks

## Test coverage

Added two unit tests:

- `RootSpanFilter_DropsTraceWhenPredicateReturnsTrue`
- `RootSpanFilter_AllowsTraceWhenPredicateReturnsFalse`

To confirm new predicates drop/allow traces as expected

## Other details

Fixes: https://github.com/DataDog/dd-trace-dotnet/issues/6749


andrewlock:
Hi @sorenhansendk, thanks for the contribution 🙏 

In general, the provided approach is something we're hesitant to include at this time, as it introduces constraints in the tracing implementation, has inconsistent behavior in partial-flush scenarios, and can also have inadvertent interactions with the sampling implementation. Today, you can achieve a similar goal using the sampling rule techniques described in https://github.com/DataDog/dd-trace-dotnet/issues/6749.

We may well look at exposing behaviour similar to OpenTelemetry's SpanProcessors in the future, but at the moment this isn't something we're looking to include I'm afraid.

PR #7541, opened 2025-09-22
Remove the guards around M.D.SqlClient V4 in tests

vandonr:
## Summary of changes

there was a piece of code skipping some tests in v4 and above of the library. Those were added in #2077 referencing https://github.com/dotnet/SqlClient/issues/1390
However, this check has stayed in place ever since, stopping tests from being run on newer versions, even though the issue says that the issue was fixed in 4.0.1 if I read correctly. Since we don't run tests on v4.0.0 (the lowest 4 we run on is 4.0.6), we shouldn't need to check for this anymore.
This might have been kept around because of an other issue where the sample app encountered an authentication issue trying to connect to the server in docker because the security certificate was not proper. Adding `TrustServerCertificate` on the connection string fixes this, and should allow us to run all tests always.

## Reason for change

better test coverage for SqlClient

## Test coverage

ran a pipeline testing all package versions for this test here: https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=187543

### DIFF
--- a/.github/workflows/verify_app_trimming_changes_are_persisted.yml
+++ b/.github/workflows/verify_app_trimming_changes_are_persisted.yml
@@ -26,7 +26,7 @@ jobs:
         run: Get-ChildItem –Path ".\tracer\src\Datadog.Trace.Trimming\build\Datadog.Trace.Trimming.xml" -Recurse -File | Remove-Item
 
       - name: "Regenerating Datadog.Trace.Trimming.xml"
-        run: .\tracer\build.ps1 BuildTracerHome CreateTrimmingFile
+        run: .\tracer\build.ps1 CreateRequiredDirectories Restore CompileManagedLoader CompileManagedSrc CreateTrimmingFile
 
       - name: "Verify no changes in Datadog.Trace.Trimming.xml"
         run: |


tracer/test/test-applications/integrations/Samples.Microsoft.Data.SqlClient/Program.cs
```diff
@@ -17,17 +17,9 @@ private static async Task<int> Main()
 
             using (var connection = OpenConnection(typeof(SqlConnection)))
             {
-                if (connection is null)
-                {
-                    Console.WriteLine("No connection could be established. Exiting with skip code (13)");
-                    return 13;
-                }
```
andrewlock, line 24: I wonder if we should keep this 🤔 I guess it depends if there's any _actual_ flake (instead of this "fake" flake)

andrewlock (approved):
Thanks LGTM, unless we see any "real" flake, in which case we'll likely want to reinstate the `exit code 13`

tracer/test/test-applications/integrations/Samples.Microsoft.Data.SqlClient/Program.cs
vandonr: ah yes, I was wondering. We can try without it, and if we see that it flakes too often, we can put it back. We should be able to check this in the "test optimization" product on DD.

andrewlock: Yep. We have a weekly dashboard that reports which we check and [we'll spot it in there](https://app.datadoghq.com/dashboard/47b-bv3-qtw/dd-trace-dotnet-weekly-ci-report?fromUser=false&refresh_mode=sliding&from_ts=1756713408360&to_ts=1757318208360&live=true). cc @NachoEchevarria just for visibility, in case you spot it showing up 🙂

PR #7543, opened 2025-09-22
Fix ad-hoc use of `TracerSettings` to use `MutableSettings` where appropriate

andrewlock:
## Summary of changes

Fix usages of `Tracer.Instance.Settings` to use `Tracer.Instance.CurrentTracerSettings.Settings` where appropriate

## Reason for change

This PR "fixes" the places that were previously grabbing the environment/service etc from `TracerSettings` to use the `MutableSettings` exposed via `CurrentTracerSettings` instead, as the location where these settings will ultimately exist.

This is effectively still just a refactoring, but prepares for the point where these settings aren't exposed on `TracerSettings` at all. The updates in this PR are for cases where you _don't_ have long-lived services, and rather need to do ad-hoc `Tracer.Instance` grabbing of the setting values in a global context. Note too that many of these places _could_ be updated in the future to subscribe to changes if that provides performance benefits. Also note that I elected not to change most calls to `IsIntegrationEnabled()` etc in this PR as there are hundreds of locations. The follow up PR handles that

Also found a few cases that were incorrectly assuming that these values cannot change. Marked them with 'TODO: Subscribe to changes in settings'

## Implementation details

- Mostly find and replace to use `CurrentTracerSettings.Settings`
- Occasional extraction of a variable where it makes sense to avoid repeated access
- Functionally identical currently (where `MutableSettings` is replaced on `TracerSettings`) but will be a required change once we stop replacing `TracerManager`.

## Test coverage

Covered by existing tests

## Other details

Part of a config stack


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Logging/Log4Net/LogsInjection/AppenderAttachedImplIntegration.cs
```diff
@@ -39,12 +39,13 @@ internal static CallTargetState OnMethodBegin<TTarget, TLoggingEvent>(TTarget in
...
 
-            if (tracer.Settings.LogsInjectionEnabled &&
+            var mutableSettings = tracer.CurrentTraceSettings.Settings;
+            if (mutableSettings.LogsInjectionEnabled &&
                 !loggingEvent.Properties.Contains(CorrelationIdentifier.ServiceKey))
             {
-                loggingEvent.Properties[CorrelationIdentifier.ServiceKey] = tracer.DefaultServiceName ?? string.Empty;
-                loggingEvent.Properties[CorrelationIdentifier.VersionKey] = tracer.Settings.ServiceVersion ?? string.Empty;
-                loggingEvent.Properties[CorrelationIdentifier.EnvKey] = tracer.Settings.Environment ?? string.Empty;
+                loggingEvent.Properties[CorrelationIdentifier.ServiceKey] = mutableSettings.DefaultServiceName;
```
bouwkast, line 46: Are we omitting `?? string.Empty` because it is never `null`?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Logging/NLog/LogsInjection/DiagnosticContextHelper.cs
```diff
@@ -118,26 +118,27 @@ private static KeyValuePair<string, object>[] CreateEntriesList(Tracer tracer, o
...
                 LogContext.TryGetValues(context, out var traceId, out var spanId, tracer.Settings.TraceId128BitLoggingEnabled))
             {
                 hasTraceIds = true;
 
                 return
                 [
-                    new KeyValuePair<string, object>(CorrelationIdentifier.ServiceKey, tracer.DefaultServiceName ?? string.Empty),
-                    new KeyValuePair<string, object>(CorrelationIdentifier.VersionKey, tracer.Settings.ServiceVersion ?? string.Empty),
-                    new KeyValuePair<string, object>(CorrelationIdentifier.EnvKey, tracer.Settings.Environment ?? string.Empty),
+                    new KeyValuePair<string, object>(CorrelationIdentifier.ServiceKey, mutableSettings.DefaultServiceName),
```
bouwkast, line 129: copy/paste comment
Are we omitting ?? string.Empty because it is never null?

Same below too
It must be never `null` 😅

tracer/src/Datadog.Trace/Configuration/GitMetadataTagsProvider.cs
```diff
@@ -21,13 +21,18 @@ internal class GitMetadataTagsProvider : IGitMetadataTagsProvider
...
+    private readonly string? _gitRepositoryUrl;
     private readonly IScopeManager _scopeManager;
     private GitMetadata? _cachedGitTags = null;
     private int _tryCount = 0;
 
-    public GitMetadataTagsProvider(TracerSettings immutableTracerSettings, IScopeManager scopeManager, ITelemetryController telemetry)
+    public GitMetadataTagsProvider(TracerSettings immutableTracerSettings, MutableSettings settings, IScopeManager scopeManager, ITelemetryController telemetry)
     {
         _immutableTracerSettings = immutableTracerSettings;
+        // These never change, even though they are exposed on MutableSettings, so we can safely grab them once here
```
bouwkast, line 33: What if they do change (or is the "never" a "can't")? Will that break anything here if they change?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Logging/Log4Net/LogsInjection/AppenderAttachedImplIntegration.cs
andrewlock: Yup, exactly that

tracer/src/Datadog.Trace/Configuration/GitMetadataTagsProvider.cs
andrewlock: The "never" here is a "can't" because there's no mechanism for them to change. So why are they in the mutable settings, you ask? Well...

- `DD_TAGS` _can_ be used to set some other env vars. That includes service, env, and also these git tags
- We strip those values from the tags initially, and set them on the "target" properties
- `DD_TAGS` can _also_ be set in code and via remote config, therefore it needs to live on `MutableSettings`
- _However_, setting the _other_ properties _via_ `DD_TAGS` is _not_ permitted in code or via remote config (we strip the disallowed values)
- Therefore, git commit sha etc never change - they can be set either as tags directly, or using `DD_TAGS`, or via other static sources - but they cannot ever actually change

Hope that makes sense, it's kind of annoying they have to live there 😅

tonyredondo (approved):

lucaspimentel (approved):

NachoEchevarria (approved):
LGTM. Thanks!

PR #7550, opened 2025-09-23
Protect against UriFormatException

NachoEchevarria:
## Summary of changes

We are getting the [following error](https://app.datadoghq.com/error-tracking?query=source%3Adotnet%20-%2AAppSec%2A%20-%2Adebugger%2A&refresh_mode=sliding&source=all&sp=%5B%7B%22p%22%3A%7B%22issueId%22%3A%22ca471d9c-3adf-11f0-9488-da7ad0900002%22%7D%2C%22i%22%3A%22error-tracking-issue%22%7D%5D&view=spans&from_ts=1758274900427&to_ts=1758289300427&live=true) in error tracking:

```
Error : Datadog ASP.NET HttpModule instrumentation error
System.UriFormatException
at System.Uri.CreateThis(String uri, Boolean dontEscape, UriKind uriKind)
at System.Web.Util.UriUtil.BuildUriImpl(String scheme, String serverName, String port, String path, String queryString, Boolean useLegacyRequestUrlGeneration)
at System.Web.HttpRequest.BuildUrl(Func`1 pathAccessor)
at System.Web.HttpRequest.get_Url()
at Datadog.Trace.AspNet.TracingHttpModule.OnBeginRequest(Object sender, EventArgs eventArgs)
```

The error was not easy to reproduce since the request has to be initially correct so it actually reaches the server.

After some tests, the error could be reproduced by adding a new module in our sample page. This module modifies the request by adding a new server name, which ends up modifying the host name of the url.

```
public class ForwardedHostBreakModule : IHttpModule
{
    public void Init(HttpApplication app)
    {
        app.BeginRequest += (s, e) =>
        {
            var ctx = ((HttpApplication)s).Context;

            // Customer-likely path: If a proxy sent X-Forwarded-Host, copy it to SERVER_NAME
            var fwd = ctx.Request.Headers["X-Forwarded-Host"];
            if (!string.IsNullOrEmpty(fwd))
            {
                // This mutates what System.Web uses to assemble Request.Url
                ctx.Request.ServerVariables.Set("SERVER_NAME", fwd);
            }
        };
    }
    public void Dispose() { }
}
```

And in web.config:

```
  <system.webServer>
    <modules runAllManagedModulesForAllRequests="true">
      <add name="ForwardedHostBreakModule" type="ForwardedHostBreakModule" />
      </modules>
  </system.webServer>
</configuration>
```

If we send this request containing this header and value to a sample page containing the previous module, we will get the same exception that is shown in error tracking:

X-Forwarded-Host: localhost:4444ide

## Test coverage

A new integration tests could be added by replicating the previous method. Not really fond of adding this overhead to our CI and sample apps in this case, since it's a corner case not so easy to reproduce.

### DIFF
--- a/tracer/src/Datadog.Trace/Util/Http/HttpRequestExtensions.Framework.cs
+++ b/tracer/src/Datadog.Trace/Util/Http/HttpRequestExtensions.Framework.cs
@@ -8,6 +8,8 @@
 using System.Web;
 using Datadog.Trace.ClrProfiler.AutoInstrumentation.AspNet;
 
+#nullable enable
+
 namespace Datadog.Trace.Util.Http
 {
     internal static partial class HttpRequestExtensions
@@ -36,13 +38,15 @@ namespace Datadog.Trace.Util.Http
         /// <seealso cref="Configuration.ConfigurationKeys.FeatureFlags.BypassHttpRequestUrlCachingEnabled"/>
         internal static string GetUrlForSpan(this HttpRequest request, QueryStringManager queryStringManager, bool bypassHttpRequestUrl)
         {
-            if (bypassHttpRequestUrl)
+            var url = bypassHttpRequestUrl ? RequestDataHelper.BuildUrl(request) : RequestDataHelper.GetUrl(request);
+
+            if (url is not null)
             {
-                return HttpRequestUtils.GetUrl(RequestDataHelper.BuildUrl(request), queryStringManager);
+                return HttpRequestUtils.GetUrl(url, queryStringManager);
             }
             else
             {
-                return HttpRequestUtils.GetUrl(RequestDataHelper.GetUrl(request), queryStringManager);
+                return string.Empty;
             }
         }
     }
--- a/tracer/src/Datadog.Trace/Util/Http/HttpRequestExtensions.Framework.cs
+++ b/tracer/src/Datadog.Trace/Util/Http/HttpRequestExtensions.Framework.cs
@@ -36,7 +36,7 @@ namespace Datadog.Trace.Util.Http
         /// <param name="bypassHttpRequestUrl">Whether or not to call access HttpRequest.Url (which caches the URL in HttpRequest).</param>
         /// <returns>The retrieved Url.</returns>
         /// <seealso cref="Configuration.ConfigurationKeys.FeatureFlags.BypassHttpRequestUrlCachingEnabled"/>
-        internal static string GetUrlForSpan(this HttpRequest request, QueryStringManager queryStringManager, bool bypassHttpRequestUrl)
+        internal static string? GetUrlForSpan(this HttpRequest request, QueryStringManager queryStringManager, bool bypassHttpRequestUrl)
         {
             var url = bypassHttpRequestUrl ? RequestDataHelper.BuildUrl(request) : RequestDataHelper.GetUrl(request);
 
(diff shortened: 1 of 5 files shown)


tracer/src/Datadog.Trace/Util/Http/HttpRequestExtensions.Framework.cs
```diff
@@ -36,13 +38,15 @@ internal static string GetUrlForSpan(this HttpRequestBase request, QueryStringMa
...
+
+            if (url is not null)
             {
-                return HttpRequestUtils.GetUrl(RequestDataHelper.BuildUrl(request), queryStringManager);
+                return HttpRequestUtils.GetUrl(url, queryStringManager);
             }
             else
             {
-                return HttpRequestUtils.GetUrl(RequestDataHelper.GetUrl(request), queryStringManager);
+                return string.Empty;
```
andrewlock: I wonder if this method should actually return `null` (and we should update the callsites to handle it)?

My reason being that the expectation is that this method returns a "valid" URL, but `string.Empty` isn't valid. WDYT?

tracer/src/Datadog.Trace/Util/Http/HttpRequestUtils.cs
```diff
@@ -34,6 +36,6 @@ internal static string GetUrl(string scheme, string host, int? port, string path
             return $"{scheme}://{GetNormalizedHost(host)}{(port.HasValue ? $":{port}" : string.Empty)}{pathBase}{path}";
         }
 
-        internal static string GetNormalizedHost(string host) => string.IsNullOrEmpty(host) ? NoHostSpecified : host;
+        internal static string GetNormalizedHost(string? host) => host is null || host.Length == 0 ? NoHostSpecified : host;
```
andrewlock: Thanks for updating the nullability requirements here 👍 

IsNullOrEmpty is optimised so is generally preferred - to fix the nullability issue we have a helper, `StringUtil` 🙂 

```suggestion
        internal static string GetNormalizedHost(string? host) => StringUtil.IsNullOrEmpty(host) ? NoHostSpecified : host;
```

andrewlock (approved):
Nice work! A couple of minor suggestions.

In terms of testing, if you wanted, you could potentially add the module to an _existing_ sample, and control whether it does anything via an env variable. I'm generally OK with us not explicitly testing it as it's an edge case, my only tiny concern is accidental regressions

tracer/src/Datadog.Trace/Util/Http/HttpRequestUtils.cs
NachoEchevarria: Cool. Thanks!

tracer/src/Datadog.Trace/Util/Http/HttpRequestExtensions.Framework.cs
NachoEchevarria: Sure. Taking a closer look, returning null would be more aligned with what we do with other similar tags. Updated. Thanks!

NachoEchevarria:
> Nice work! A couple of minor suggestions.
> 
> In terms of testing, if you wanted, you could potentially add the module to an _existing_ sample, and control whether it does anything via an env variable. I'm generally OK with us not explicitly testing it as it's an edge case, my only tiny concern is accidental regressions

Actually, I found a way to unit test this, so we can test it with less overhead. I have added the test.

NachoEchevarria:
Thanks for the feedback and review!

PR #7559, opened 2025-09-24
[Profiler] Add Index to samples and build index vector for libdatadog

chrisnas:
## Summary of changes

Add index to samples 

## Reason for change
Libdatadog will update the API to create and send .pprof. In that context, it is needed to map offsets in values vector to specific profiles used to create .pprof or OTel format.

## Implementation details
Propagate the index from the providers to the samples down to the profile exporter.

## Test coverage
Add SampleTypeProvider tests

### DIFF
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native/Profile.cpp
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native/Profile.cpp
@@ -184,6 +184,7 @@ libdatadog::profile_unique_ptr CreateProfile(std::vector<SampleValueType> const&
 
     // TODO: create a vector<int32> containing the indexes of the valueTypes
     std::vector<int32_t> indexes;
+    indexes.reserve(valueTypes.size());
 
     for (auto const& type : valueTypes)
     {
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native/RawCpuSample.h
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native/RawCpuSample.h
@@ -11,7 +11,7 @@
 class RawCpuSample : public RawSample
 {
 public:
-    RawCpuSample() = default;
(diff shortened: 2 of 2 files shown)


profiler/src/ProfilerEngine/Datadog.Profiler.Native/Profile.cpp
```diff
@@ -162,9 +162,13 @@ libdatadog::profile_unique_ptr CreateProfile(std::vector<SampleValueType> const&
     std::vector<ddog_prof_ValueType> samplesTypes;
     samplesTypes.reserve(valueTypes.size());
 
+    // TODO: create a vector<int32> containing the indexes of the valueTypes
+    std::vector<int32_t> indexes;
+
```
gleocadie, line 188: you can call `reserve` to pre-allocate slots for indexes

profiler/src/ProfilerEngine/Datadog.Profiler.Native/RawCpuSample.h
```diff
@@ -11,7 +11,7 @@
 class RawCpuSample : public RawSample
 {
 public:
-    RawCpuSample() noexcept = default;
+    RawCpuSample() = default;
```
gleocadie: why removing the noexcept ?

profiler/src/ProfilerEngine/Datadog.Profiler.Native/Profile.cpp
chrisnas: ✅ done

profiler/src/ProfilerEngine/Datadog.Profiler.Native/RawCpuSample.h
chrisnas: ✅ Just incomplete revert  :^)

PR #7561, opened 2025-09-25
Fix log directory location for managed code in Windows nano server

andrewlock:
## Summary of changes

- Add a fallback for determining log location in Windows nanoserver in managed code
- ~Add smoke tests for nano server~ more of a pain than I would like so will do these separately

## Reason for change

We determine the default log location in managed code (On Windows) by calling:

```csharp
Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData)
```

However, on Windows NanoServer, this returns `""`, which means we end up logging to a _relative_ path. 

This PR adds a fallback to ensure we always log to `C:\ProgramData` by default if this check fails.

This is a known issue:
- https://github.com/dotnet/runtime/issues/22690
- https://github.com/dotnet/runtime/issues/21430
- https://github.com/dotnet/runtime/pull/109673

## Implementation details

- Add a fallback to checking the `ProgramData` env var (as in https://github.com/dotnet/runtime/pull/109673)
- Add a final explicit fallback to `C:\ProgramData` (to match how the native side works)

## Test coverage

~Added a couple of smoke tests for nanoserver. Might move these to a separate PR if they prove too much of a pain~ They were

## Other details

Incidental finding in an escalation

### DIFF
--- a/tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/StartupLogger.cs
+++ b/tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/StartupLogger.cs
@@ -84,7 +84,18 @@ namespace Datadog.Trace.ClrProfiler.Managed.Loader
                 {
                     if (Environment.OSVersion.Platform == PlatformID.Win32NT)
                     {
-                        var windowsDefaultDirectory = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData), @"Datadog .NET Tracer", "logs");
+                        // On Nano Server, this returns "", so we fallback to reading from the env var set in the base image instead
+                        // - https://github.com/dotnet/runtime/issues/22690
+                        // - https://github.com/dotnet/runtime/issues/21430
+                        // - https://github.com/dotnet/runtime/pull/109673
+                        // If _that_ fails, we just hard code it to "C:\ProgramData", which is what the native components do anyway
+                        var programData = Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData);
+                        if (string.IsNullOrEmpty(programData))
+                        {
+                            programData = Environment.GetEnvironmentVariable("ProgramData");
+                        }
+
+                        var windowsDefaultDirectory = Path.Combine(programData ?? @"C:\ProgramData", "Datadog .NET Tracer", "logs");
                         logDirectory = windowsDefaultDirectory;
                     }
                     else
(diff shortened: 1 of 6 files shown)


tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/StartupLogger.cs
```diff
@@ -84,7 +84,22 @@ private static string GetLogDirectory()
...
+                        // - https://github.com/dotnet/runtime/issues/21430
+                        // - https://github.com/dotnet/runtime/pull/109673
+                        // If _that_ fails, we just hard code it to "C:\ProgramData", which is what the native components do anyway
+                        var programData = Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData);
+                        if (string.IsNullOrEmpty(programData))
+                        {
+                            programData = Environment.GetEnvironmentVariable("ProgramData");
+                            if (string.IsNullOrEmpty(programData))
+                            {
+                                programData = @"C:\ProgramData";
```
tonyredondo, line 98: is it ok to assume the C drive? I gues for 99% of cases but 🤷🏻

tracer/src/Datadog.Trace/Logging/Internal/DatadogLoggingFactory.cs
```diff
@@ -205,7 +205,22 @@ private static string GetDefaultLogDirectory(IConfigurationSource source, IConfi
...
+                // - https://github.com/dotnet/runtime/issues/21430
+                // - https://github.com/dotnet/runtime/pull/109673
+                // If _that_ fails, we just hard code it to "C:\ProgramData", which is what the native components do anyway
+                var programData = Environment.GetFolderPath(Environment.SpecialFolder.CommonApplicationData);
+                if (string.IsNullOrEmpty(programData))
+                {
+                    programData = Environment.GetEnvironmentVariable("ProgramData");
+                    if (string.IsNullOrEmpty(programData))
+                    {
+                        programData = @"C:\ProgramData";
```
tonyredondo, line 219: same comment

tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/StartupLogger.cs
andrewlock: > is it ok to assume the C drive? I gues for 99% of cases but 🤷🏻

Yeah, I had the same thought, but it's what we do on the _native_ side, so the answer is _probably_ yes? We could try getting the `SystemDrive` env var, but if `ProgramData` failed, I'm guessing `SystemDrive` would fail too, so we'll probably end up in the same situation either way? Ultimately, if we can't create the folder, we write to a temp path instead so, meh?

andrewlock: As discussed, we could look at transferring more data between native-> managed, but will push that to a separate PR (and we need to "fix" that communication first)

bouwkast (approved):

PR #7566, opened 2025-09-25
Output MSBuild errors when publishing

NachoEchevarria:
## Summary of changes

We saw the following error in our CI:

```
01:59:31 [INF] > "C:\Program Files\dotnet\dotnet.exe" publish c:\mnt\tracer\src\Datadog.Trace.MSBuild\Datadog.Trace.MSBuild.csproj --configuration Release --framework net461 --no-restore --no-build --output c:\mnt\shared\bin\monitoring-home\net461 /property:Platform=AnyCPU
01:59:34 [ERR] MSBUILD : error MSB4166: Child node "3" exited prematurely. Shutting down. Diagnostic information may be found in files in "C:\Users\ContainerAdministrator\AppData\Local\Temp\MSBuildTemp\" and will be named MSBuild_*.failure.txt. This location can be changed by setting the MSBUILDDEBUGPATH environment variable to a different directory.
01:59:34 [ERR] Target PublishManagedTracer has thrown an exception
```

In order to diagnose the actual error generated from MSBUILD, we are checking for these MSBuild_*.failure.txt files and printing them into the logs.

Potentially, these logs could be generated in any MSBUILD call. Since we have only seen this in PublishManagedTracer, in this PR we are only checking for error files in that task, but it could easily be added to any other compilation or publishing task.

## Test coverage

We can force an intentional crash by using this target file and using these properties:  /p:CrashDuringBuild=true /p:CustomBeforeMicrosoftCommonTargets="Crash.targets"

```
<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <UsingTask TaskName="CrashNow"
             TaskFactory="RoslynCodeTaskFactory"
             AssemblyFile="$(MSBuildToolsPath)\Microsoft.Build.Tasks.Core.dll">
    <Task>
      <Using Namespace="System" />
      <Code Type="Fragment" Language="cs"><![CDATA[
        Environment.FailFast("Intentional crash for MSBuild failure-file test");
        return true; // not reached
      ]]></Code>
    </Task>
  </UsingTask>

  <!-- Fire before Build/Publish, but only when asked AND only for the target project -->
  <Target Name="CrashIfEnabled"
          BeforeTargets="Build;Publish;CoreCompile"
          Condition="'$(CrashDuringBuild)'=='true' and '$(MSBuildProjectName)'=='Datadog.Trace'">
    <Message Text="Crashing MSBuild worker for $(MSBuildProjectName)..." Importance="high" />
    <CrashNow />
  </Target>
</Project>
```


tracer/build/_build/MSBuildLogHelper.cs
```diff
@@ -0,0 +1,63 @@
...
+using System.Linq;
+using System.Text;
+using Nuke.Common;
+using Nuke.Common.IO;
+
+using Logger = Serilog.Log;
+
+internal static class MSBuildLogHelper
+{
+    static internal AbsolutePath MsbuildDebugPath => NukeBuild.RootDirectory / "logs" / "msbuild";
```
andrewlock: nit, I think we should put this in the `build_data` directory
```suggestion
    static internal AbsolutePath MsbuildDebugPath => TestLogsDirectory / "msbuild";
```

NachoEchevarria: Sure!

PR #7568, opened 2025-09-25
Lpimentel/remove tracer home dependency

lucaspimentel:
## Other details

Based on #7567.

### DIFF
--- a/tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetCore.cs
+++ b/tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetCore.cs
@@ -8,6 +8,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Reflection;
+using System.Runtime.InteropServices;
 
 namespace Datadog.Trace.ClrProfiler.Managed.Loader
 {
@@ -32,12 +33,18 @@ namespace Datadog.Trace.ClrProfiler.Managed.Loader
                 tracerFrameworkDirectory = version.Major >= 6 ? "net6.0" : "netcoreapp3.1";
             }
 
-            var tracerHomeDirectory = ReadEnvironmentVariable("DD_DOTNET_TRACER_HOME") ?? string.Empty;
+            var tracerHomeInfo = GetTracerHomeInfo();
+            if (tracerHomeInfo is null)
+            {
+                return null;
+            }
+
+            var tracerHomeDirectory = tracerHomeInfo.Value.Path;
             var fullPath = Path.GetFullPath(Path.Combine(tracerHomeDirectory, tracerFrameworkDirectory));
 
             if (!Directory.Exists(fullPath))
             {
-                StartupLogger.Log($"The tracer home directory cannot be found at '{fullPath}', based on the DD_DOTNET_TRACER_HOME value '{tracerHomeDirectory}' and current directory {Environment.CurrentDirectory}");
+                StartupLogger.Log($"The tracer home directory cannot be found at '{fullPath}', based on {tracerHomeInfo.Value.Description} and current directory {Environment.CurrentDirectory}");
                 return null;
             }
 
@@ -106,6 +113,34 @@ namespace Datadog.Trace.ClrProfiler.Managed.Loader
             // The file doesn't exist in the Home folder.
             StartupLogger.Debug("Assembly not found in path: {0}", path);
             return null;
+        }
+
+        private static IEnumerable<string> EnumerateCoreClrProfilerPathEnvironmentVariables()
+        {
+            var archSpecific = GetCoreClrProfilerPathForCurrentArchitecture();
+            if (!string.IsNullOrEmpty(archSpecific))
+            {
+                yield return archSpecific;
+            }
+
+            yield return "CORECLR_PROFILER_PATH";
+        }
+
+        private static string GetCoreClrProfilerPathForCurrentArchitecture()
+        {
+            switch (RuntimeInformation.ProcessArchitecture)
+            {
+                case Architecture.X64:
+                    return "CORECLR_PROFILER_PATH_64";
+                case Architecture.X86:
+                    return "CORECLR_PROFILER_PATH_32";
+                case Architecture.Arm64:
+                    return "CORECLR_PROFILER_PATH_ARM64";
+                case Architecture.Arm:
+                    return "CORECLR_PROFILER_PATH_ARM";
+                default:
+                    return null;
+            }
         }
 
         private static bool IsDatadogAssembly(string path, out Assembly cachedAssembly)
--- a/tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetCore.cs
+++ b/tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetCore.cs
@@ -4,6 +4,9 @@
 // </copyright>
 
 #if NETCOREAPP
+
+#nullable enable
+
 using System;
 using System.Collections.Generic;
 using System.IO;
@@ -17,39 +20,49 @@ namespace Datadog.Trace.ClrProfiler.Managed.Loader
     /// </summary>
     public partial class Startup
     {
-        private static CachedAssembly[] _assemblies;
+        private static CachedAssembly[]? _assemblies;
 
         internal static System.Runtime.Loader.AssemblyLoadContext DependencyLoadContext { get; } = new ManagedProfilerAssemblyLoadContext();
 
-        private static string ResolveManagedProfilerDirectory()
+        private static string? ResolveManagedProfilerDirectory()
         {
-            string tracerFrameworkDirectory = "netstandard2.0";
+            var tracerHomeInfo = GetTracerHomePath();
 
-            var version = Environment.Version;
-
-            // Old versions of .net core have a major version of 4
-            if ((version.Major == 3 && version.Minor >= 1) || version.Major >= 5)
-            {
-                tracerFrameworkDirectory = version.Major >= 6 ? "net6.0" : "netcoreapp3.1";
-            }
-
-            var tracerHomeInfo = GetTracerHomeInfo();
             if (tracerHomeInfo is null)
             {
                 return null;
             }
 
-            var tracerHomeDirectory = tracerHomeInfo.Value.Path;
-            var fullPath = Path.GetFullPath(Path.Combine(tracerHomeDirectory, tracerFrameworkDirectory));
+            var version = Environment.Version;
+            string managedLibrariesDirectory;
+
+            if (version.Major >= 6)
+            {
+                // version > 6.0
+                managedLibrariesDirectory = "net6.0";
+            }
+            else if (version is { Major: 3, Minor: >= 1 } || version.Major == 5)
+            {
+                // version is 3.1 or 5.0
+                managedLibrariesDirectory = "netcoreapp3.1";
+            }
+            else
+            {
+                // version < 3.1 (note: previous versions of .NET Core had major version 4)
+                managedLibrariesDirectory = "netstandard2.0";
+            }
+
+            var fullPath = Path.Combine(Path.GetFullPath(tracerHomeInfo), managedLibrariesDirectory);
 
             if (!Directory.Exists(fullPath))
             {
-                StartupLogger.Log($"The tracer home directory cannot be found at '{fullPath}', based on {tracerHomeInfo.Value.Description} and current directory {Environment.CurrentDirectory}");
+                StartupLogger.Log($"Tracer home directory not found at '{fullPath}'");
                 return null;
             }
 
-            // We use the List/Array approach due the number of files in the tracer home folder (7 in netstandard, 2 netcoreapp3.1+)
+            // We use the List/Array approach due to the number of files in the tracer home folder (7 in netstandard, 2 netcoreapp3.1+)
             var assemblies = new List<CachedAssembly>();
+
             foreach (var file in Directory.EnumerateFiles(fullPath, "*.dll", SearchOption.TopDirectoryOnly))
             {
                 assemblies.Add(new CachedAssembly(file, null));
@@ -61,12 +74,12 @@ namespace Datadog.Trace.ClrProfiler.Managed.Loader
             return fullPath;
         }
 
-        private static Assembly AssemblyResolve_ManagedProfilerDependencies(object sender, ResolveEventArgs args)
+        private static Assembly? AssemblyResolve_ManagedProfilerDependencies(object sender, ResolveEventArgs args)
         {
             return ResolveAssembly(args.Name);
         }
 
-        private static Assembly ResolveAssembly(string name)
+        private static Assembly? ResolveAssembly(string name)
         {
             var assemblyName = new AssemblyName(name);
 
@@ -74,7 +87,7 @@ namespace Datadog.Trace.ClrProfiler.Managed.Loader
             // to enter the AssemblyResolve event when searching for resources
             // in its satellite assemblies. This seems to have been fixed in
             // .NET Core in the 2.0 servicing branch, so we should not see this
-            // occur, but guard against it anyways. If we do see it, exit early
+            // occur but guard against it anyway. If we do see it, exit early
             // so we don't cause infinite recursion.
             if (string.Equals(assemblyName.Name, "System.Private.CoreLib.resources", StringComparison.OrdinalIgnoreCase) ||
                 string.Equals(assemblyName.Name, "System.Net.Http", StringComparison.OrdinalIgnoreCase))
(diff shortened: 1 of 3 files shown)


tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.cs
```diff
@@ -175,6 +185,71 @@ private static void TryInvokeManagedMethod(string typeName, string methodName, s
...
+                {
+                    StartupLogger.Log("Unable to determine tracer home directory from {0}={1}", envVarName, profilerPath);
+                    return null;
+                }
+
+                // if the directory name is one of these, go one level higher
+                List<string> architectures = ["win-x64", "win-x86", "linux-x64", "linux-arm64", "linux-musl-x64", "linux-musl-arm64", "osx"];
+
+                if (architectures.Contains(directory.Name))
+                {
```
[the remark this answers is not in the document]
lucaspimentel: @codex add "osx-arm64" and "osx-x64" to this list

tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.cs
```diff
@@ -84,18 +87,25 @@ static Startup()
                 }
 #endif
 
-                var runInAas = ReadBooleanEnvironmentVariable(AzureAppServicesKey, false);
-                if (runInAas)
+                string typeName;
+                string methodName;
```
andrewlock: nit: I see what you were doing here, but I think we should move the declaration down to their initialization and just have 2 separate sets of vars. Mutable long-lived vars are harder to follow.

It would actually also be more performant because you can declare them as consts and the compiler will inline them instead of having to do interpolation etc.

tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetCore.cs
```diff
@@ -108,15 +128,37 @@ private static Assembly ResolveAssembly(string name)
...
+                                       _ => null
+                                   };
+
+            if (!string.IsNullOrEmpty(archSpecificPath))
+            {
+                yield return archSpecificPath!;
+            }
+
+            yield return "CORECLR_PROFILER_PATH";
+        }
```
andrewlock: It's a tiny optimisation, but we prob don't want the overhead of an iterator here, seeing as this is startup blocking. Personally I'd say we just inline this at the callsite

tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetCore.cs
```diff
@@ -108,15 +128,37 @@ private static Assembly ResolveAssembly(string name)
...
+
+        private static bool IsDatadogAssembly(string path, out Assembly? cachedAssembly)
         {
-            for (var i = 0; i < _assemblies.Length; i++)
+            if (_assemblies is not null)
```
andrewlock: We can (and probably should) restructure `Startup` so that `_assemblies` is non-nullable, which avoids all these superfluous checks (we _know_ it can't be null here in practice)

tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetFramework.cs
```diff
@@ -60,25 +70,39 @@ private static Assembly ResolveAssembly(string name)
...
+                StartupLogger.Debug("Assembly loaded: {0}", assembly.FullName);
                 return assembly;
             }
 
-            StartupLogger.Debug("Assembly not found in path: '{0}'", path);
+            StartupLogger.Debug("Assembly not found in path: {0}", path);
             return null;
         }
+
+        private static IEnumerable<string> EnumerateProfilerPathEnvironmentVariables()
```
andrewlock: same argument, I'd rather inline this than have the itterator implementation personally

tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.cs
```diff
@@ -175,6 +185,71 @@ private static void TryInvokeManagedMethod(string typeName, string methodName, s
...
+                var directory = Directory.GetParent(profilerPath);
+
+                if (directory is null)
+                {
+                    StartupLogger.Log("Unable to determine tracer home directory from {0}={1}", envVarName, profilerPath);
+                    return null;
+                }
+
+                // if the directory name is one of these, go one level higher
+                List<string> architectures = ["win-x64", "win-x86", "linux-x64", "linux-arm64", "linux-musl-x64", "linux-musl-arm64", "osx", "osx-arm64", "osx-x64"];
```
andrewlock: 1. this list should probably be a static readonly list at the top of the class instead of buried here
2. `osx`? Is that a thing?! 😅 
3. It'll be easy to miss adding a new arch here, I wonder how we can avoid that? 🤔

andrewlock (approved):
LGTM, thanks! I just think we need

1. Potential some perf tweaks
2. Tests (unit and smoke/integration)

tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.cs
lucaspimentel: > this list should probably be a static readonly list at the top of the class instead of buried here

Agreed. I'll fix it. (and I blame AI)

> `osx`? Is that a thing?! 

Yes, in `Datadog.Trace.Bundle`.

> It'll be easy to miss adding a new arch here

Yeah, I had the same concern. A previous version would try all combinations of OS and arch, but that seems a bit wasteful since we know we don't support many of those (like `win-arm64` or `linux-x86`).

On the other hand, we'll notice quickly if something is missing because the tracer won't work?

tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.cs
lucaspimentel: fixed

tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetCore.cs
lucaspimentel: Fixed, but not by inlining at the call site. There are different versions in `Startup.NetCore.cs` vs `Startup.NetFramework.cs`. Instead, I split this into two methods, one for the arch-specific env var name, and another for the generic one (no arch). This way we avoid allocating either a collection or this iterator.

tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetFramework.cs
lucaspimentel: refactored

tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.cs
lucaspimentel: > this list should probably be a static readonly list at the top of the class instead of buried here

fixed

tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.cs
```diff
@@ -21,6 +22,19 @@ public partial class Startup
...
+            "win-x64",
+            "win-x86",
+            "linux-x64",
+            "linux-arm64",
+            "linux-musl-x64",
+            "linux-musl-arm64",
+            "osx",
+            "osx-arm64",
+            "osx-x64"
+        };
```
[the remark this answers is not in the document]
lucaspimentel: @codex does dd-trace-dotnet support win-arm64?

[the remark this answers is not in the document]
lucaspimentel: Too slow, I already pushed those same changes 😅

tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetCore.cs
lucaspimentel: At a quick glance I couldn't make it non-nullable without defaulting to `= []`, which allocates an array (only the first time, but this only runs once). Since we _know_ they can't be null, I suppressed the warnings with `!` instead.

tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.cs
```diff
@@ -21,6 +22,20 @@ public partial class Startup
         private const string AzureAppServicesSiteExtensionKey = "DD_AZURE_APP_SERVICES"; // only set when using the AAS site extension
         private const string TracerHomePathKey = "DD_DOTNET_TRACER_HOME";
 
+        private static readonly HashSet<string> ArchitectureDirectories = new(StringComparer.OrdinalIgnoreCase)
+        {
+            "win-x64",
+            "win-x86",
+            "win-arm64",
```
andrewlock: We don't ship this today, I guess we should leave it out? 🤷‍♂️

tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.cs
```diff
@@ -125,13 +139,83 @@ static Startup()
...
 
-                // If the logger fails, throw the original exception. The profiler emits code to log it.
+                // If the logger fails, throw the original exception. The native library emits code to log it.
                 throw;
             }
         }
 
         internal static string? ManagedProfilerDirectory { get; }
 
+        internal static string? GetTracerHomePath(IEnvironmentVariableProvider envVars)
```
andrewlock: nit: this will box the `IEnvironmentVariableProvider`, we should probably use the generic with constraint trick?

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/SmokeTests/SmokeTestBase.cs
```diff
@@ -29,7 +29,11 @@ protected SmokeTestBase(
                 this.GetType(),
                 output,
                 samplesDirectory: "test/test-applications/regression",
-                prependSamplesToAppName: false);
+                prependSamplesToAppName: false)
+            {
+                // Don't set DD_DOTNET_TRACER_HOME in smoke tests to verify the fallback logic works
```
andrewlock, line 34: These are different smoke tests than the ones I meant 😅 But that's probably fine 🤷‍♂️

andrewlock (approved):
Thanks!

tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.cs
lucaspimentel: fixed, thanks

tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.cs
lucaspimentel: future proofing!

j/k. removed.

lucaspimentel:
TODO: Check for usage of `DD_DOTNET_TRACER_HOME` outside the managed loader. Not a blocker for this PR. For now this is a hidden/internal fallback and all docs and tooling still tell users to set `DD_DOTNET_TRACER_HOME`.

---- 

Edit: 🤖✨ found these usages:

1. `Datadog.Trace\AppSec\Waf\Initialization\LibraryLocationHelper.cs:95` - Reads the env var to locate the WAF library
2. `Datadog.Trace.Tools.dd_dotnet\Utils.cs:124` - Sets the env var when launching processes
3. `Datadog.Trace.Tools.dd_dotnet\Checks\ProcessBasicCheck.cs:125,127,141` - Checks if the env var is set and valid
4. `Datadog.Trace.Tools.Runner\Utils.cs:592` - Sets the env var when launching processes
5. `Datadog.Trace.Coverage.collector\CoverageSettings.cs:46` - Reads the env var for coverage settings
6. `Datadog.Trace.Coverage.collector\CoverageCollector.cs:139` - Error message referencing the env var
7. `Datadog.Trace.BenchmarkDotNet\DatadogDiagnoser.cs:135,232` - Reads and sets the env var for benchmarks
8. `Datadog.FleetInstaller\TracerValues.cs:28,48` - Sets the env var in environment variable dictionaries
9. `Datadog.Tracer.Native\environment_variables.h` - Native code constant definition

----

Edit: 🤖✨ made further analysis:

Summary of Impact When `DD_DOTNET_TRACER_HOME` is Not Set

  1. AppSec WAF Library Location (`LibraryLocationHelper.cs:95`)

  - Impact: Graceful fallback ✅
  - Behavior: If not set, it skips this source and tries other methods to locate the WAF library:
    - First checks `DD_INTERNAL_TRACE_NATIVE_ENGINE_PATH` (set by native loader)
    - Then checks profiler path environment variables (`CORECLR_PROFILER_PATH`, etc.)
    - Finally checks `AppDomain.CurrentDomain.BaseDirectory`
  - Result: AppSec can still work without it

  2. `dd_dotnet` Tool (`Utils.cs:124`)

  - Impact: Sets it regardless ✅
  - Behavior: The tool computes `tracerHome` from either:
    - User-specified `--tracer-home` parameter
    - Or defaults to `Path.Combine(AppContext.BaseDirectory, "..")` (line 81)
  - Result: Always sets the env var when launching processes; doesn't read the existing value

  3. `dd_dotnet` Process Check (`ProcessBasicCheck.cs:141`)

  - Impact: Validation fails ⚠️
  - Behavior: Explicitly checks if the env var is set and displays warning: `Environment variable not set: DD_DOTNET_TRACER_HOME`
  - Result: Diagnostic check reports failure (ok = false)

  4. Trace Runner Tool (`Utils.cs:592`)

  - Impact: Sets it regardless ✅
  - Behavior: Similar to `dd_dotnet`, computes `tracerHome` from either:
    - User-specified `tracerHome` parameter
    - Or defaults to `AppContext.BaseDirectory` parent (line 505 shows error if not found)
  - Result: Always sets the env var when launching processes

  5. Coverage Collector (`CoverageCollector.cs:139`)

  - Impact: Coverage disabled ❌
  - Behavior: Logs error: `Tracer home (DD_DOTNET_TRACER_HOME environment variable) is not defined or folder doesn't exist, coverage has been disabled.`
  - Result: Returns false and coverage collection is completely disabled

  6. `BenchmarkDotNet` Integration (`DatadogDiagnoser.cs:135,232`)

  - Impact: Graceful fallback ✅
  - Behavior: Tries multiple locations in order:
    a. `DD_DOTNET_TRACER_HOME` env var
    b. `{WorkingDirectory}/datadog`
    c. `{BenchmarkAssemblyLocation}/datadog`
    d. (Optional) relative path from benchmark assembly
  - Result: Can still find monitoring home from other locations; sets the env var when launching benchmark processes

  7. Fleet Installer (`TracerValues.cs:28,48`)

  - Impact: Sets it as part of configuration ✅
  - Behavior: The `TracerValues` constructor requires `tracerHomeDirectory` parameter, which it then includes in the environment variable dictionaries
  - Result: Always includes it in the env vars it configures for IIS and global installations

  Conclusion

  Components that REQUIRE it:
  - ❌ Coverage Collector - completely disables coverage if not set

  Components that VALIDATE it (but may not strictly require it):
  - ⚠️ `dd_dotnet` Process Check - reports diagnostic failure

  Components that DON'T need it (have fallbacks or set it themselves):
  - ✅ AppSec WAF - has multiple fallback mechanisms
  - ✅ `dd_dotnet` Tool - computes and sets it when launching processes
  - ✅ Trace Runner Tool - computes and sets it when launching processes
  - ✅ `BenchmarkDotNet` - has multiple fallback locations
  - ✅ Fleet Installer - receives it as constructor parameter and sets it

  The most critical dependency is the Coverage Collector (`Datadog.Trace.Coverage.collector`), which will completely disable coverage functionality without it.

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/InstrumentationTests.cs
```diff
@@ -251,6 +251,21 @@ public async Task WhenUsingPathWithDotsInInTracerHome_InstrumentsApp()
...
         }
 
+        [SkippableFact]
+        [Trait("RunOnWindows", "True")]
+        public async Task WhenOmittingTracerHome_InstrumentsApp()
+        {
+            SetLogDirectory();
+
+            // DD_DOTNET_TRACER_HOME is not set, so the tracer should derive it from the profiler path
+            Output.WriteLine("DD_DOTNET_TRACER_HOME not set, relying on profiler path environment variables");
```
bouwkast, line 265: Should we / Could we check to assert that it isn't set and fail the test if it is set? Wondering about chance of it getting picked up from other tests, but not really sure if that can happen in these.

bouwkast (approved):
👍

link04 (approved):

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/InstrumentationTests.cs
lucaspimentel: Good idea. I added checks to verify that `DD_DOTNET_TRACER_HOME` is not set in the system env var or in `EnvironmentHelper.CustomEnvironmentVariables`.

PR #7569, opened 2025-09-26
[AAP] Enable by default API Security Endpoints collection

e-n-0:
## Summary of changes

This PR enable by default the feature to collect all endpoints at startup.
See #6733 #7317 for implementation.


tracer/src/Datadog.Trace/AppSec/SecuritySettings.cs
```diff
@@ -132,7 +132,7 @@ public SecuritySettings(IConfigurationSource? source, IConfigurationTelemetry te
                                            .Value;
 
             ApiSecurityEndpointCollectionEnabled = config.WithKeys(ConfigurationKeys.AppSec.ApiSecurityEndpointCollectionEnabled)
-                                           .AsBool(false);
+                                           .AsBool(true);
```
andrewlock, line 135: If this is only enabled when ASM is enabled, then it's your call, as discussed

robertpi (approved):
Seems this will only be active when appsec is active - which is what we where asked to do.

PR #7570, opened 2025-09-26
Improve execution benchmark comment post

andrewlock:
## Summary of changes

Improve the report so it doesn't take as much real-estate

## Reason for change

The charts currently look a bit borked, and taking up that extra space seems unnecessary. I think we could do it with more direct comparisons (probably tabular). The metrics timeit outputs directly would be useful to see, particularly if we just highlight that there _are_ issues, and show the results side-by-side

### DIFF
--- a/tracer/build/_build/ExecutionTimeComparison/CompareExecutionTime.cs
+++ b/tracer/build/_build/ExecutionTimeComparison/CompareExecutionTime.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Reflection;
 using System.Text;
 using System.Text.Json.Nodes;
 using Nuke.Common;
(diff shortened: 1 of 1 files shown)


andrewlock:
> Didn't _really_ look at the code but the new comment looks good to me unfortunate that it has so many regressions still as it is noisy but not as noisy as before :)

Yeah, it's really flagging a lot of regressions there @bouwkast  and a lot of those "regressions" are in the baseline scenarios which, you know aren't regressions, that's just variability 😅 I'm going to double check the numbers in the tables are _actually_ the values we think should be there by checking the source json, but then fix the regression calculation. Based on the graphs there's only _actually_ a regression in the "bailout" scenarios here. That's obviously _still_ noisy, but it's closer to correct...

As for _why_ there's apparent regressions here... 🤷‍♂️ 😬

andrewlock:
OK, fixed the incorrect regression calculations @bouwkast, and compressed the format even further 🙂

PR #7578, opened 2025-09-29
Revert "Revert "[Profiler] Make `timer_create`-based CPU profiler default (#7322)""

gleocadie:
## Summary of changes

Reverts DataDog/dd-trace-dotnet#7427 and supersedes DataDog/dd-trace-dotnet#7444

## Reason for change

Another try to make `timer_create`-based CPU profiler the default. What's the difference between this PR and the last one(s): 
- The RingBuffer is now long-lived and does not have its lifecycle tied to the provider.

## Implementation details
- Only register managed threads that are assigned to native threads.
- Have the ring buffer as a field of the CorProfilerCallback class so it can outlive the customer(s).

## Other details

https://datadoghq.atlassian.net/browse/PROF-11986

### DIFF
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/ProfilerSignalManager.cpp
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/ProfilerSignalManager.cpp
@@ -93,6 +93,22 @@ bool ProfilerSignalManager::UnRegisterHandler()
     return true;
 }
 
+bool ProfilerSignalManager::IgnoreSignal() {
+    struct sigaction sampleAction;
+    sampleAction.sa_handler = SIG_IGN;
+    sigemptyset(&sampleAction.sa_mask);
+    sampleAction.sa_flags = 0;
+
+    int32_t result = sigaction(_signalToSend, &sampleAction, nullptr);
+    if (result != 0)
+    {
+        Log::Error("ProfilerSignalManager::IgnoreSignal: Failed mark ", strsignal(_signalToSend), " as ignored. Reason: ",
+                   strerror(errno), ".");
+        return false;
+    }
+    return true;
+}
+
 void ProfilerSignalManager::SetSignal(int32_t signal)
 {
     _signalToSend = signal;
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
@@ -71,6 +71,30 @@
 
 #include <cmath>
 
+void LogServiceStart(bool success, const char* name )
+{
+    if (success)
+    {
+        Log::Info(name, " started successfully.");
+    }
+    else
+    {
+        Log::Info(name, " failed to start. This service might have been started earlier.");
+    }
+}
+
+void LogServiceStop(bool success, const char* name )
+{
+    if (success)
+    {
+        Log::Info(name, " stopped successfully.");
+    }
+    else
+    {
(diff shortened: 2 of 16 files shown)


profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/ProfilerSignalManager.cpp
```diff
@@ -93,6 +93,22 @@ bool ProfilerSignalManager::UnRegisterHandler()
...
+bool ProfilerSignalManager::IgnoreSignal() {
+    struct sigaction sampleAction;
+    sampleAction.sa_handler = SIG_IGN;
+    sigemptyset(&sampleAction.sa_mask);
+    sampleAction.sa_flags = 0;
+
+    int32_t result = sigaction(_signalToSend, &sampleAction, nullptr);
+    if (result != 0)
+    {
+        Log::Error("ProfilerSignalManager::IgnoreSignal: Failed mark ", strsignal(_signalToSend), " as ignored. Reason: ",
```
chrisnas, line 105: In case of unexplained failure (because this should not fail), we risk to fill up the logs + stderr: maybe switch to debug or only lon 1 over 1000

profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
```diff
@@ -794,17 +819,22 @@ bool CorProfilerCallback::StartServices()
...
-            Log::Info(name, " failed to start. This service might have been started earlier.");
-        }
+        LogServiceStart(success, name);
         result &= success;
     }
 
+#ifdef LINUX
+    // We cannot add the timer_create-based CPU profiler to the _services list
+    // we have to control the Stop
+    // If we fail to stop, we mustn't release the memory associated to it
```
chrisnas:
```suggestion
    // If we fail to stop, we should not release the memory associated to it
```

profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
```diff
@@ -1599,7 +1622,16 @@ HRESULT STDMETHODCALLTYPE CorProfilerCallback::Shutdown()
 #ifdef LINUX
```
chrisnas: if we want to stop the services in creation reverse order, maybe we should stop this one first.

profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
```diff
@@ -1599,7 +1622,16 @@ HRESULT STDMETHODCALLTYPE CorProfilerCallback::Shutdown()
...
     if (_pCpuProfiler != nullptr)
     {
-        _pCpuProfiler->Stop();
+        // if we failed at stopping the time_create-based CPU profiler,
+        // it's safer to not release the memory.
+        // Otherwise, we might crash the application.
+        // Reason: one thread could be executing the signal handler and accessing some field
+        auto stopped = _pCpuProfiler->Stop();
+        LogServiceStop(stopped, _pCpuProfiler->GetName());
+        if (stopped)
```
chrisnas: Why don't we ALWAYS set it to nullptr?

profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/ProfilerSignalManager.cpp
chrisnas: Sorry... done only once

chrisnas (approved):

PR #7586, opened 2025-09-30
[DO NOT MERGE][2/x] feat(ssi): IIS injection can be disabled

dmehala:
## Summary of changes
The policy engine now supports rules for excluding application pools from instrumentation

### DIFF
--- a/shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
+++ b/shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
@@ -324,7 +324,7 @@ namespace datadog::shared::nativeloader
 
         // Workload Selection
 #ifdef BUILD_WITH_WORKLOAD_SELECTION
-        if (IsSingleStepInstrumentation() && !IsRunningOnIIS())
+        if (IsSingleStepInstrumentation())
         {
           enum class InjectionStatus : uint8_t {
             ALLOW,
@@ -332,8 +332,16 @@ namespace datadog::shared::nativeloader
             UNKNOWN
           } injection_status = InjectionStatus::UNKNOWN;
 
+          const bool isIis = IsRunningOnIIS();
+          auto processName = ToString(process_name);
+          if (isIis) {
+            if (auto maybe_application_pool = GetApplicationPool()) {
+                processName = ToString(*maybe_application_pool);
+            }
+          }
+
           plcs::init();
-          plcs::set_params(plcs::StringEvaluator::PROCESS_EXE, ToString(process_name));
+          plcs::set_params(plcs::StringEvaluator::PROCESS_EXE, processName);
           plcs::set_params(plcs::StringEvaluator::RUNTIME_LANGUAGE, "dotnet");
 
           // TODO: Remove once the policy engine will be able to log the context.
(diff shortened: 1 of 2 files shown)


shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -344,27 +352,32 @@ namespace datadog::shared::nativeloader
...
+                return std::nullopt;
             }
 
+            injection_status = (eval_result == plcs::Result::TRUE) ? InjectionStatus::DENY : InjectionStatus::UNKNOWN;
+            return std::nullopt;
+          });
+
+          auto maybe_error = plcs::evaluate_buffer_from_file(GetPoliciesPath());
+            if (maybe_error) {
+            Log::Error("CorProfiler::Initialize: An error occured while evaluating workload selection (reason: ", *maybe_error, ")");
```
bmermet, line 372: nit: aren't we missing identation here?

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -344,27 +352,32 @@ namespace datadog::shared::nativeloader
...
 
-          const auto wls_file = GetPoliciesPath();
-          if (fs::exists(wls_file))
-          {
-            auto maybe_error = plcs::evaluate_buffer_from_file(wls_file);
-            if (maybe_error) {
-              Log::Error("CorProfiler::Initialize: An error occured while evaluating workload selection (reason: ", *maybe_error, ")");
-              return E_FAIL;
+          plcs::register_action(plcs::Action::INJECT_DENY, [&injection_status](plcs::Result eval_result, const std::vector<const char*>&, const char* desc) -> std::optional<plcs::Error> {
+            if (injection_status == InjectionStatus::DENY)
```
bmermet, line 361: We should update the status only if it's unknown, the first rule that fires a Deny or an Allow wins.

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -344,27 +352,32 @@ namespace datadog::shared::nativeloader
               {
```
bmermet, line 352: Idem here, we should update only if the status is Unknown

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -344,27 +352,32 @@ namespace datadog::shared::nativeloader
...
+
+          auto maybe_error = plcs::evaluate_buffer_from_file(GetPoliciesPath());
+            if (maybe_error) {
+            Log::Error("CorProfiler::Initialize: An error occured while evaluating workload selection (reason: ", *maybe_error, ")");
+            return E_FAIL;
+          }
+
+          if (isIis) {
+            if (injection_status == InjectionStatus::DENY) {
+              return CORPROF_E_PROFILER_CANCEL_ACTIVATION;
```
bmermet, line 378: We can have the log "CorProfiler::Initialize: Instrumentation denied due to workload selection." here, it's even more true than for the case below where deny is the default behavior.

PR #7591, opened 2025-09-30
Partial Flush enablement

pierotibou:
## Summary of changes

Enable Partial flushing by default to align with other tracers

## Reason for change

Align with other tracers.
More info in https://docs.google.com/document/d/1ivPw_6croUQsbYAsdTMUv8zwPZOPnPPzFBUYUDvelmg/edit?tab=t.0#heading=h.y9vxsdoezrvw

## Implementation details
Just changed the default value. Didn't change the amount of spans for now

## Test coverage
Already covered


tracer/test/Datadog.Trace.Tests/Configuration/TracerSettingsTests.cs
```diff
@@ -967,7 +967,7 @@ public void OtlpHeadersParsing(string primaryValue, string fallbackValue, string
         }
 
         [Theory]
-        [MemberData(nameof(BooleanTestCases), false)]
+        [MemberData(nameof(BooleanTestCases), true)]
```
lucaspimentel, line 970: Hmmm, this test is suspicious. Doesn't the second argument mean we _always_ expect `settings.PartialFlushEnabled` to be `true` (or `false` before this PR), regardless of the value of `DD_TRACE_PARTIAL_FLUSH_ENABLED`?

PR #7593, opened 2025-09-30
fix(ci): recycle stale ServicePoint connections

tonyredondo:
## Summary of changes

  - Harden the CI Visibility TestOptimizationClient on .NET Framework by reconfiguring the target ServicePoint before every send to aggressively recycle idle keep-alives.
  - When a retry observes a keep-alive–style WebException, immediately close the ServicePoint’s connection group so the next attempt opens a fresh socket.

  ## Reason for change

  - Production logs showed System.Net.WebException: The underlying connection was closed during CI Visibility uploads; stale keep-alive sockets from the load balancer were being reused across retries, triggering customer-facing errors.

  ## Implementation details

  - Configure each ServicePoint with short lease/idle timeouts, disable Expect100Continue and Nagle delays, and raise the connection limit to reduce the chance of reusing dead connections.
  - On keep-alive/connection-closed failures, explicitly close the ServicePoint connection group before rethrowing so our retry loop forces a new socket.

  ## Test coverage

Hard to reproduce.

  ## Other details

  - Scoped to the net48 path; Core transports remain unchanged for now. Debug-level logs were extended to record connection-group resets when they occur.

### DIFF
--- a/tracer/src/Datadog.Trace/Ci/Net/TestOptimizationClient.cs
+++ b/tracer/src/Datadog.Trace/Ci/Net/TestOptimizationClient.cs
@@ -133,6 +133,52 @@ internal sealed partial class TestOptimizationClient : ITestOptimizationClient
         return customConfiguration;
     }
 
+#if NETFRAMEWORK
+    private static ServicePoint ConfigureServicePoint(Uri baseUri)
+    {
+        var sp = ServicePointManager.FindServicePoint(baseUri);
+
+        // Kill “zombie” keep-alives proactively
+        sp.ConnectionLeaseTimeout = (int)TimeSpan.FromSeconds(25).TotalMilliseconds; // < typical 30s LB idle
+        sp.MaxIdleTime = (int)TimeSpan.FromSeconds(20).TotalMilliseconds; // drop idle sockets quickly
+
+        // Reduce proxy quirks / latency micro-hiccups
+        sp.Expect100Continue = false;  // per-ServicePoint (not global)
+        sp.UseNagleAlgorithm = false;  // optional: lowers small-write delays
+
+        // Keep enough lanes open to avoid request queueing
+        sp.ConnectionLimit = Math.Max(64, Environment.ProcessorCount * 8);
+        return sp;
+    }
+
+    private static void CloseConnectionGroupOnFailure(Uri baseUri, Exception exception)
+    {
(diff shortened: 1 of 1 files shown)


tracer/src/Datadog.Trace/Ci/Net/TestOptimizationClient.cs
```diff
@@ -133,6 +133,52 @@ public static ITestOptimizationClient CreateCached(string workingDirectory, ITes
...
+        sp.MaxIdleTime = (int)TimeSpan.FromSeconds(20).TotalMilliseconds; // drop idle sockets quickly
+
+        // Reduce proxy quirks / latency micro-hiccups
+        sp.Expect100Continue = false;  // per-ServicePoint (not global)
+        sp.UseNagleAlgorithm = false;  // optional: lowers small-write delays
+
+        // Keep enough lanes open to avoid request queueing
+        sp.ConnectionLimit = Math.Max(64, Environment.ProcessorCount * 8);
+        return sp;
+    }
```
bouwkast, line 166: I don't see how this is causing it to actually recycle the connection before each send?

Wouldn't we need to do something like you have in `CloseConnectionGroupOnFailure` in place of calling `ConfigureServicePoint` on each send?

```csharp
                    var servicePoint = ServicePointManager.FindServicePoint(baseUri);
                    servicePoint?.CloseConnectionGroup(string.Empty);
```

tonyredondo: You're right, the intention of that code is just change the defaults configured for that service point in order to invalidate the connections sooner, there's no way to check the connection before trying it and receiving the exception, and that's the part handled by the CloseConnectionGroupOnFailure. I'll try to add more comments about the reasoning for this.

tonyredondo: I've added a big chunk of comments here: [add comments.](https://github.com/DataDog/dd-trace-dotnet/pull/7593/commits/c1c7aacca0b915e72d6839bcf099fff5581102a1)

bouwkast: Ahh okay that helps a lot thanks!

bouwkast (approved):

PR #7594, opened 2025-09-30
managed loader: refactor for testability, add unit tests

lucaspimentel:
## Summary of changes

Refactored the managed loader (`Datadog.Trace.ClrProfiler.Managed.Loader`) to improve testability and added comprehensive unit tests.

## Reason for change

The managed loader code was difficult to test due to tight coupling with environment variables and filesystem access. This refactoring enables better test coverage and makes the code more maintainable.

## Implementation details

- **New abstractions**: 
  - Added `IEnvironmentVariableProvider` and `EnvironmentVariableProvider` to abstract environment variable access, enabling dependency injection for tests
  - Implemented as `readonly struct` with generic type parameters to avoid heap allocations and boxing
- **Refactored startup logic**:
  - Moved TFM directory computation into dedicated methods that can be tested independently
  - Separated directory existence checks from path computation
  - Improved error messages and logging consistency
- **Code improvements**:
  - Added nullable reference type annotations throughout
  - Standardized Boolean environment variable parsing to match tracer conventions
  - Cleaned up comments and improved code clarity

## Test coverage

Added 5 new test files with coverage of:
- Environment variable reading and Boolean parsing (`EnvironmentVariableProviderTests`)
- Log directory resolution across platforms (`StartupLoggerTests`)
- TFM directory computation for .NET Core (`StartupNetCoreTests`)
- TFM directory computation for .NET Framework (`StartupNetFrameworkTests`)
- Edge cases and error conditions
- Added `MockEnvironmentVariableProvider` for test isolation

## Other details

This refactoring is preparatory work for #7568 which makes `DD_DOTNET_TRACER_HOME` optional, but the changes were hard to test.

### DIFF
--- a/tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetCore.cs
+++ b/tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetCore.cs
@@ -55,7 +55,7 @@ namespace Datadog.Trace.ClrProfiler.Managed.Loader
                     assemblies.Add(new CachedAssembly(file, null));
                 }
 
-                _assemblies = assemblies.ToArray();
+                _assemblies = [..assemblies];
                 StartupLogger.Debug("Total number of assemblies: {0}", _assemblies.Length);
             }
 


tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetCore.cs
```diff
@@ -76,9 +84,8 @@ private static Assembly ResolveAssembly(string name)
             }
 
             // WARNING: Logs must not be added _before_ we check for the above bail-out conditions
-            StartupLogger.Debug("Assembly Resolve event received for: {0}", name);
             var path = Path.Combine(ManagedProfilerDirectory, $"{assemblyName.Name}.dll");
-            StartupLogger.Debug("Looking for: {0}", path);
+            StartupLogger.Debug("Assembly Resolve event received for: {0}. Looking for: {1}", name, path);
```
tonyredondo: Path.Combine can throw. It shouldn't but it can, it would be better to write the log before the Path, just to get more info in that case.

tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetCore.cs
```diff
@@ -90,15 +97,16 @@ private static Assembly ResolveAssembly(string name)
...
-                // 2) The AssemblyVersion is lower than the version used by Datadog.Trace, the assembly will fail to load
+                // 2) If the AssemblyVersion is lower than the version used by Datadog.Trace, the assembly will fail to load
                 //    and invoke this resolve event. It must be loaded in a separate AssemblyLoadContext since the application will only
-                //    load the originally referenced version
-                StartupLogger.Debug("Loading {0} with DependencyLoadContext.LoadFromAssemblyPath", path);
-                var assembly = DependencyLoadContext.LoadFromAssemblyPath(path); // Load unresolved framework and third-party dependencies into a custom Assembly Load Context
+                //    load the originally referenced version.
+                StartupLogger.Debug("Calling DependencyLoadContext.LoadFromAssemblyPath(\"{0}\")", path);
+                var assembly = DependencyLoadContext.LoadFromAssemblyPath(path); // Load unresolved framework and third-party dependencies into a custom AssemblyLoadContext
+                StartupLogger.Debug("Assembly loaded: {0}", assembly.FullName);
```
tonyredondo: Do we really need this? `assembly.FullName` is not cached and internally is a QCall. https://source.dot.net/#System.Private.CoreLib/src/System/Reflection/RuntimeAssembly.cs,142

If is not necessary let's remove it.

tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/StartupLogger.cs
```diff
@@ -11,14 +13,34 @@ namespace Datadog.Trace.ClrProfiler.Managed.Loader
...
-        private static readonly string LogDirectory = GetLogDirectory();
-        private static readonly string StartupLogFilePath = SetStartupLogFilePath();
+        private static readonly bool DebugEnabled;
+        private static readonly string? StartupLogFilePath;
         private static readonly object PadLock = new();
 
-        public static void Log(string message, params object[] args)
+        static StartupLogger()
+        {
+            var envVars = new EnvironmentVariableProvider(logErrors: false);
```
tonyredondo, line 24: I know is not the same instance than before, but we are moving from a zero allocation code to having two different instances of an EnvironmentVariableProvider. Can we improve this? maybe we don't need the allocations for the interface at all, and can be donde with some generic constraints + struct magic.

lucaspimentel: I thought about that, but this only happens once per process, and `EnvironmentVariableProvider` itself is tiny (one bool). But yeah, I can certainly do that.

lucaspimentel: Done!

tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetCore.cs
lucaspimentel: Flipped them around so we log before `Path.Combine()`

tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetCore.cs
lucaspimentel: This was just there from before. Removed, thanks!

tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetCore.cs
```diff
@@ -16,58 +19,63 @@ namespace Datadog.Trace.ClrProfiler.Managed.Loader
...
+            if (Directory.Exists(fullPath))
+            {
+                // We use the List/Array approach due to the number of files in the tracer home folder (7 in netstandard, 2 netcoreapp3.1+)
+                var assemblies = new List<CachedAssembly>();
+                foreach (var file in Directory.EnumerateFiles(fullPath, "*.dll", SearchOption.TopDirectoryOnly))
+                {
+                    assemblies.Add(new CachedAssembly(file, null));
+                }
+
+                _assemblies = assemblies.ToArray();
```
andrewlock: nit:
```suggestion
                _assemblies = [..assemblies];
```

tracer/src/Datadog.Trace.ClrProfiler.Managed.Loader/Startup.NetFramework.cs
```diff
@@ -16,20 +18,12 @@ namespace Datadog.Trace.ClrProfiler.Managed.Loader
...
-            var tracerHomeDirectory = ReadEnvironmentVariable("DD_DOTNET_TRACER_HOME") ?? string.Empty;
-            var fullPath = Path.GetFullPath(Path.Combine(tracerHomeDirectory, "net461"));
-            if (!Directory.Exists(fullPath))
-            {
-                StartupLogger.Log($"The tracer home directory cannot be found at '{fullPath}', based on the DD_DOTNET_TRACER_HOME value '{tracerHomeDirectory}' and current directory {Environment.CurrentDirectory}");
-                return null;
-            }
-
-            return fullPath;
+            return Path.Combine(Path.GetFullPath(tracerHomeDirectory), "net461");
```
andrewlock, line 23: Do we definitely want to drop the `Directory.Exists` check here?

lucaspimentel: The call to `Directory.Exists()` is now consolidated in `Startup.cs`, instead of having one each in `Startup.NetCore.cs` and `Startup.NetFramework.cs`. It happens right after this `ComputeTfmDirectory()` returns.

This refactoring makes the platform-specific `ComputeTfmDirectory()` methods responsible only for computing the path, while the shared `Startup.cs` handles validation. It also makes the code more testable since `ComputeTfmDirectory()` is now a pure function that can be tested independently.

PR #7614, opened 2025-10-03
Capture Exception when sending metrics

NachoEchevarria:
## Summary of changes

We are getting some [test errors](https://app.datadoghq.com/ci/test/runs?query=test_level%3Atest%20%40test.service%3Add-trace-dotnet%20%40test.status%3Afail%20%40git.branch%3Amaster%20%40test.full_name%3A%2ACouch%2A&agg_m=count&agg_m_source=base&agg_t=count&currentTab=overview&eventStack=AwAAAZmlUDfVUlPy2QAAABhBWm1sWG5pQUFBQkRidXJUT3YtMjlsbEgAAAAkZjE5OWE1NTEtYjcxYi00ZTNjLWE4MjQtMzk3OWJkMWZmNWFlAAHSfw&fromUser=false&index=citest&testId=AwAAAZmlUDfVUlPy2QAAABhBWm1sWG5pQUFBQkRidXJUT3YtMjlsbEgAAAAkZjE5OWE1NTEtYjcxYi00ZTNjLWE4MjQtMzk3OWJkMWZmNWFlAAHSfw&trace=AwAAAZmlUDfVUlPy2QAAABhBWm1sWG5pQUFBQkRidXJUT3YtMjlsbEgAAAAkZjE5OWE1NTEtYjcxYi00ZTNjLWE4MjQtMzk3OWJkMWZmNWFlAAHSfw&start=1758882011475&end=1759486811475&paused=false) because the flaky attributte sends some telemetry reporting the flaky test failures and, when the telemetry call fails, en exception is thrown and the test itself fails as well.

```
System.Net.Http.HttpRequestException: The SSL connection could not be established, see inner exception.
---- System.IO.IOException : Received an unexpected EOF or 0 bytes from the transport stream.
```

This error can make any test using telemetry fail if the telemetry endpoint does not respond, for instance, so it can potentially affect many tests.

### DIFF
--- a/tracer/test/Datadog.Trace.TestHelpers.AutoInstrumentation/ErrorHelpers.cs
+++ b/tracer/test/Datadog.Trace.TestHelpers.AutoInstrumentation/ErrorHelpers.cs
@@ -10,6 +10,7 @@ using System.Net.Http;
 using System.Reflection;
 using System.Text;
 using System.Threading.Tasks;
+using Datadog.Trace.ClrProfiler.AutoInstrumentation.AWS.SDK;
 using Datadog.Trace.Tagging;
 using Datadog.Trace.TestHelpers.AutoInstrumentation;
 using Newtonsoft.Json.Linq;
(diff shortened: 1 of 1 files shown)


tracer/test/Datadog.Trace.TestHelpers.AutoInstrumentation/ErrorHelpers.cs
```diff
@@ -10,6 +10,7 @@
 using System.Reflection;
 using System.Text;
 using System.Threading.Tasks;
+using Datadog.Trace.ClrProfiler.AutoInstrumentation.AWS.SDK;
```
bouwkast:
```suggestion
```

PR #7616, opened 2025-10-03
Add configuration option to disable agent discovery service

lucaspimentel:
## Summary

Adds `DD_TRACE_DISCOVERY_ENABLED` environment variable (default: `true`) to allow disabling the agent discovery service. When disabled, the tracer uses `NullDiscoveryService` instead of querying the agent for available endpoints.

## Motivation

In certain environments, such as Azure Functions with the Rust agent, the discovery endpoint is not available. Currently, the tracer always attempts to query the discovery service, which results in unnecessary network calls and potential errors in these scenarios.

## Changes

- Added `DD_TRACE_DISCOVERY_ENABLED` configuration key to `ConfigurationKeys.cs`
- Added `DiscoveryServiceEnabled` property to `TracerSettings` (defaults to `true`)
- Updated `TracerManagerFactory.GetDiscoveryService()` to return `NullDiscoveryService.Instance` when disabled
- Changed `GetDiscoveryService()` visibility from `protected` to `internal virtual` for consistency with override in `TestOptimizationTracerManagerFactory`
- Added unit test `DiscoveryServiceCanBeDisabled` to verify both enabled and disabled states

## Test Plan

- [x] Unit test verifies setting is read correctly
- [x] Unit test verifies `DiscoveryService` is returned when enabled
- [x] Unit test verifies `NullDiscoveryService.Instance` is returned when disabled
- [ ] Manual testing in Azure Functions environment with Rust agent

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tracer/src/Datadog.Trace/Ci/TestOptimizationTracerManagerFactory.cs
+++ b/tracer/src/Datadog.Trace/Ci/TestOptimizationTracerManagerFactory.cs
@@ -108,7 +108,7 @@ namespace Datadog.Trace.Ci
             return new ApmAgentWriter(settings, updateSampleRates, discoveryService, traceBufferSize);
         }
 
-        protected override IDiscoveryService GetDiscoveryService(TracerSettings settings)
+        internal override IDiscoveryService GetDiscoveryService(TracerSettings settings)
             => _testOptimizationTracerManagement.DiscoveryService;
     }
 }
--- a/tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
+++ b/tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
@@ -586,6 +586,14 @@ namespace Datadog.Trace.Configuration
         public const string ApplicationMonitoringConfigFileEnabled = "DD_APPLICATION_MONITORING_CONFIG_FILE_ENABLED";
 
         /// <summary>
+        /// Configuration key to disable the agent discovery service.
+        /// When disabled, the tracer will not query the agent for available endpoints.
+        /// This is useful in environments where the discovery endpoint is not available (e.g., Azure Functions with Rust agent).
+        /// Default value is true (discovery service enabled).
+        /// </summary>
+        public const string DiscoveryServiceEnabled = "DD_TRACE_DISCOVERY_ENABLED";
+
+        /// <summary>
         /// String constants for CI Visibility configuration keys.
         /// </summary>
         public static class CIVisibility
--- a/tracer/src/Datadog.Trace/TracerManagerFactory.cs
+++ b/tracer/src/Datadog.Trace/TracerManagerFactory.cs
@@ -462,9 +462,6 @@ namespace Datadog.Trace
             }
         }
 
-        protected virtual IDiscoveryService GetDiscoveryService(TracerSettings settings)
-            => DiscoveryService.Create(settings.Exporter);
-
         internal static IDogStatsd CreateDogStatsdClient(TracerSettings settings, string serviceName, List<string> constantTags, string prefix = null)
         {
             try
@@ -537,6 +534,11 @@ namespace Datadog.Trace
 
             return CreateDogStatsdClient(settings, serviceName, constantTags);
         }
+
+        internal virtual IDiscoveryService GetDiscoveryService(TracerSettings settings)
+            => settings.DiscoveryServiceEnabled
+                   ? DiscoveryService.Create(settings.Exporter)
+                   : NullDiscoveryService.Instance;
 
         /// <summary>
         /// Gets an "application name" for the executing application by looking at
--- a/tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
+++ b/tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
@@ -372,6 +372,10 @@ namespace Datadog.Trace.Configuration
                                              .WithKeys(ConfigurationKeys.AzureServiceBusBatchLinksEnabled)
                                              .AsBool(defaultValue: true);
 
+            DiscoveryServiceEnabled = config
+                                     .WithKeys(ConfigurationKeys.DiscoveryServiceEnabled)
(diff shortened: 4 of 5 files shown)


tracer/test/Datadog.Trace.Tests/Telemetry/config_norm_rules.json
```diff
@@ -309,6 +309,7 @@
   "dbm_propagation_mode": "dbm_propagation_mode",
   "trace.remove_root_span_laravel_queue": "trace_remove_root_span_laravel_queue_enabled",
   "trace.remove_autoinstrumentation_orphans": "trace_remove_auto_instrumentation_orphans_enabled",
+  "DD_DISCOVERY_SERVICE_ENABLED": "discovery_service_enabled",
```
dudikeleti: Do we need to add the new config to [config_norm_rules.json in dd-go](https://github.com/DataDog/dd-go/blob/2f7f1b882d0b77295a6d8b3e2f86ee9d11f5eb52/trace/apps/tracer-telemetry-intake/telemetry-payload/static/config_norm_rules.json)?

BTW, I noticed a config with a similar name - `agent_discovery_enabled`. I’m not sure we use it or what it was originally meant for. Could it be intended for something similar to what you’re doing?

lucaspimentel: > Do we need to add the new config to `config_norm_rules.json`

Yes!

> I noticed a config with a similar name - `agent_discovery_enabled`

Maybe from another tracer, since this is file is shared. I'll look into it, thanks.

lucaspimentel: >> I noticed a config with a similar name - `agent_discovery_enabled`
>
> Maybe from another tracer, since this is file is shared. I'll look into it, thanks.

As far a I can tell, the closest thing to `agent_discovery_enabled` is a `DD_PROCESS_AGENT_DISCOVERY_ENABLED` config in the agent, which is unrelated.

lucaspimentel:
~`dd-trace-java` went with `DD_TRACE_SERVICE_DISCOVERY_ENABLED`.~
~See https://github.com/DataDog/dd-trace-java/pull/9705.~

Sorry, different feature. "Service discovery" vs "discovery service". I changed the name to `DD_AGENT_FEATURE_POLLING_ENABLED` to avoid confusion.

mcculls:
> dd-trace-java went with DD_TRACE_SERVICE_DISCOVERY_ENABLED.
See https://github.com/DataDog/dd-trace-java/pull/9705.

we can change it to `DD_SERVICE_DISCOVERY_ENABLED` in `dd-trace-java` if we want to align on that name

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -585,6 +585,14 @@ internal static partial class ConfigurationKeys
         /// </summary>
         public const string ApplicationMonitoringConfigFileEnabled = "DD_APPLICATION_MONITORING_CONFIG_FILE_ENABLED";
 
+        /// <summary>
+        /// Configuration key to disable the agent discovery service.
+        /// When disabled, the tracer will not query the agent for available endpoints.
+        /// This is useful in environments where the discovery endpoint is not available (e.g., Azure Functions with Rust agent).
+        /// Default value is true (discovery service enabled).
+        /// </summary>
+        public const string DiscoveryServiceEnabled = "DD_DISCOVERY_SERVICE_ENABLED";
```
lucaspimentel: Change the name to `DD_AGENT_FEATURE_POLLING_ENABLED `.

"Discovery service" is confusing because "service discovery" is a completely unrelated feature.

lucaspimentel: Done

tracer/src/Datadog.Trace/TracerManagerFactory.cs
```diff
@@ -507,6 +504,11 @@ internal static IDogStatsd CreateDogStatsdClient(TracerSettings settings, string
             }
         }
 
+        internal virtual IDiscoveryService GetDiscoveryService(TracerSettings settings)
```
andrewlock, line 508: nit: 
```suggestion
        // internal for testing
        internal virtual IDiscoveryService GetDiscoveryService(TracerSettings settings)
```

tracer/src/Datadog.Trace/Ci/TestOptimizationTracerManagerFactory.cs
```diff
@@ -107,7 +107,7 @@ protected override IAgentWriter GetAgentWriter(TracerSettings settings, IDogStat
             return new ApmAgentWriter(settings, updateSampleRates, discoveryService, traceBufferSize);
         }
 
-        protected override IDiscoveryService GetDiscoveryService(TracerSettings settings)
+        internal override IDiscoveryService GetDiscoveryService(TracerSettings settings)
             => _testOptimizationTracerManagement.DiscoveryService;
```
andrewlock, line 111: Should TestOptimization also respect this setting? If so we prob need to update it there too

tracer/src/Datadog.Trace/TracerManagerFactory.cs
lucaspimentel: thanks

tracer/src/Datadog.Trace/Ci/TestOptimizationTracerManagerFactory.cs
lucaspimentel: `TestOptimizationTracerManagerFactory` always uses `NullDiscoveryService.Instance`.

https://github.com/DataDog/dd-trace-dotnet/blob/0cb18766c69c887169f81a89322fdb59e46ac459/tracer/src/Datadog.Trace/Ci/TestOptimizationTracerManagement.cs#L75-L81

PR #7620, opened 2025-10-06
[Tracer] Add Azure Event Hubs instrumentation 

pablomartinezbernardo:
## Summary of changes

New integration for Azure Event Hubs configurable by:
* `DD_TRACE_AZUREEVENTHUBS_ENABLED` (default `true`) enables/disables the integration
* `DD_TRACE_AZURE_EVENTHUBS_BATCH_LINKS_ENABLED` (default `true`) enables/disables Span Linking and `TryAdd` spans

## Integration points

* `Azure.Messaging.EventHubs.Producer.EventDataBatch.TryAdd` that adds an `EventData` to a batch. Creates a span and stores context.
* `Azure.Messaging.EventHubs.Producer.EventHubProducerClient.SendAsync(EventDataBatch)` sends a batch. Creates a span extracting context for Span Links.
* `Azure.Messaging.EventHubs.Producer.EventHubProducerClient.SendAsync(IEnumerable<EventData>)` sends a collection of `EventData`. Creates a span.
* `Azure.Core.Shared.MessagingClientDiagnostics.InstrumentMessage` (In the `Azure.Messaging.EventHubs` assembly) function to instrument specific messages. Adds context to messages.
* `Azure.Messaging.EventHubs.Amqp.AmqpConsumer.ReceiveAsync` receives all messages, single entry point for all reception. Creates span and links.
* Added context extraction for Event Hubs triggers

## Test coverage

Integration tests

## Other details

Some refactoring has been made to the Service Bus integration to better reuse components. I also added previously missing calls to `IntegrationGeneratedSpan` in Service Bus. Unlike the previous Service Bus PRs, these are all changes in a single PR.


lucaspimentel:
@codex review

tracer/missing-nullability-files.csv
```diff
@@ -187,6 +187,7 @@ src/Datadog.Trace/Tagging/AwsKinesisTags.cs
 src/Datadog.Trace/Tagging/AwsSdkTags.cs
 src/Datadog.Trace/Tagging/AwsSnsTags.cs
 src/Datadog.Trace/Tagging/AwsSqsTags.cs
+src/Datadog.Trace/Tagging/AzureEventHubsTags.cs
```
lucaspimentel: When adding new files, always use `#nullable enable` to prevent adding to this list. We want this list to get smaller over time until it's empty.

tracer/src/Datadog.Trace/Tagging/AzureEventHubsTags.cs
```diff
@@ -0,0 +1,119 @@
...
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2 License.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
+// </copyright>
+
+using System;
+using Datadog.Trace.Configuration;
+using Datadog.Trace.ExtensionMethods;
+using Datadog.Trace.SourceGenerators;
+
+#pragma warning disable SA1402 // File must contain single type
```
lucaspimentel, line 12: New files should use `#nullable enable` until we have migrated everything and we can enable it solution-wide.

lucaspimentel:
> `DD_TRACE_AZUREEVENTHUBS_ENABLED` (default `false`) enables/disables the integration

I don't see `DD_TRACE_AZUREEVENTHUBS_ENABLED`. I assume it's enabled by default?

tracer/build/PackageVersionsGeneratorDefinitions.json
```diff
@@ -889,6 +875,17 @@
       "7.*.*"
     ]
   },
+  {
+    "IntegrationName": "AzureEventHubs",
+    "SampleProjectName": "Samples.AzureEventHubs",
+    "NugetPackageSearchName": "Azure.Messaging.EventHubs",
+    "MinVersion": "5.11.0",
+    "MaxVersionExclusive": "5.13.0",
```
bouwkast:
```suggestion
    "MaxVersionExclusive": "6.0.0",
```

Otherwise we'd stop testing right away basically whereas we'd likely want to test all 5.11.0+

pablomartinezbernardo: Thanks! Changed it for SB as well

pablomartinezbernardo:
> Still going through it, quite a large PR
> 
> Why is this disabled by default?

Because of a misunderstanding on previous talks. Changed it to enabled by default along with SB

pablomartinezbernardo:
> LGTM! Just one question - will spans be generated for `EventHubBufferedProducerClient.EnqueueEventAsync`? It looks like this is another option users have to send events
> 
> https://learn.microsoft.com/en-us/dotnet/api/azure.messaging.eventhubs.producer.eventhubbufferedproducerclient?view=azure-dotnet

Yes! Being tested in [TestEventHubsBufferedProducerAsync](https://github.com/DataDog/dd-trace-dotnet/pull/7620/files#diff-72de00cfe172a7a55bc018c17138f3cb8a5b4c335dbe7d55c5237ae3125321c1R244), it is actually how you use the buffered client. No specific instrumentation needed, it uses the non-buffered producer internally.

PR #7622, opened 2025-10-06
Fix profiler_execution_benchmarks flakiness

NachoEchevarria:
## Summary of changes

We are getting the following flaky error when running the profiler_execution_benchmarks:

`Scenario: Profiler_garbagecollections Cmd: Samples.Computer01 --scenario 12 --iterations 4000 --param 2 Warming up .. Duration: 2 sec 426 ms Run .An error occurred while running TimeItSharp: System.IO.EndOfStreamException: Unable to read beyond the end of the stream. at void System.ThrowHelper.ThrowEndOfFileException() at double System.IO.BinaryReader.ReadDouble() at async Task<DataPoint> TimeItSharp.Common.ScenarioProcessor.RunCommandAsync(int index, Scenario scenario, TimeItPhase phase, int executionId, Cancellation*** cancellation***) in /_/src/TimeItSharp.Common/ScenarioProcessor.cs:858 at void System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()`

This error comes from TimeItSharp and is a known bug that will be addressed in future versions of the library. In the meantime, retrying the task will prevent it from failing.

### DIFF
--- a/.azure-pipelines/ultimate-pipeline.yml
+++ b/.azure-pipelines/ultimate-pipeline.yml
@@ -4536,6 +4536,7 @@ stages:
         DD_PROFILING_AGENTLESS: 1
         DD_CIVISIBILITY_LOGS_ENABLED: 0
         DD_CIVISIBILITY_ITR_ENABLED: 0
+      retryCountOnTaskFailure: 3
 
     - powershell: |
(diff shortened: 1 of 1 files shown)


gleocadie (approved):
LGTM

PR #7623, opened 2025-10-06
Add aas.site.kind and aas.site.type tags to agentless logs

lucaspimentel:
## Summary of changes

Add `aas.site.kind` and `aas.site.type` tags to logs submitted via Direct Log Submission (agentless logging) in Azure App Services and Azure Functions.

## Reason for change

The Serverless UI needs these tags to correctly differentiate between Azure Web Apps and Azure Function Apps. Currently, some Azure Function Apps are incorrectly displayed as mis-configured Azure Web Apps because the logs lack distinguishing tags.

Related Slack discussion: https://dd.slack.com/archives/C0880PRMWG3/p1757605468066169

## Implementation details

Modified `LogFormatter.EnrichTagsWithAasMetadata()` to include two additional tags from `ImmutableAzureAppServiceSettings`:
- `aas.site.kind`: The kind of application instance (e.g., "functionapp", "app", "functionapp_linux")
- `aas.site.type`: The type of application instance (e.g., "function", "app")

These tags are added to the `ddtags` property alongside the existing `aas.resource.id` tag, and are automatically populated from Azure environment variables like `FUNCTIONS_WORKER_RUNTIME` and `FUNCTIONS_EXTENSION_VERSION`.

## Test coverage

Updated `LogFormatterTests.WritesLogFormatCorrectly` to validate that the new tags are included when running in Azure App Services. All 128 test variations pass.

## Other details

Fixes APMSVLS-195


tracer/src/Datadog.Trace/Logging/DirectSubmission/Formatting/LogFormatter.cs
```diff
@@ -76,26 +75,69 @@ private static string StringifyGlobalTags(IReadOnlyDictionary<string, string> gl
...
+
+            return url;
+        }
+
+        private static string EnrichTagsWithAasMetadata(string globalTags, ImmutableAzureAppServiceSettings? aasSettings)
         {
             if (aasSettings is null)
             {
-                return string.IsNullOrEmpty(globalTags) ? null : globalTags;
+                return globalTags;
```
bouwkast, line 107: I guess this is likely fine, but we used to return `null` if `globalTags` is a `string.Empty`, but without going into the code I don't think if that is even important. 

Basically before:

- if `globalTags` is `string.Empty` -> return `null`

After

- if `globalTags` is `string.Empty` -> return `string.Empty`

bouwkast (approved):

tracer/src/Datadog.Trace/Logging/DirectSubmission/Formatting/LogFormatter.cs
```diff
@@ -63,7 +62,7 @@ public LogFormatter(
 
         internal delegate LogPropertyRenderingDetails FormatDelegate<T>(JsonTextWriter writer, in T state);
 
-        internal string? Tags { get; private set; }
+        private string? Tags { get; set; }
```
bouwkast: Looks like this needs to be `internal`


```suggestion
        internal string? Tags { get; set; }
```

tracer/src/Datadog.Trace/Logging/DirectSubmission/Formatting/LogFormatter.cs
lucaspimentel: Yeah, I was trying to clean that up because it was weird and confusing. 😅  Now `globalTags` is not nullable.

tracer/src/Datadog.Trace/Logging/DirectSubmission/Formatting/LogFormatter.cs
lucaspimentel: I thought I fixed that .... thanks.

tracer/src/Datadog.Trace/Logging/DirectSubmission/Formatting/LogFormatter.cs
bouwkast: Is it `ddtags":""`?

Because now it looks like we always append it to the logs which I don't think is ideal, we'd likely want to omit it if it isn't present

lucaspimentel: Yup, thanks. I had to update the previous for `Tags is not null` to `!string.IsNullOrEmpty(Tags)`. Yay, tests 🎉

tracer/src/Datadog.Trace/Logging/DirectSubmission/Formatting/LogFormatter.cs
```diff
@@ -76,26 +75,69 @@ private static string StringifyGlobalTags(IReadOnlyDictionary<string, string> gl
...
+                  .Append(TagSeparator);
             }
 
             // remove final joiner
             sb.Remove(sb.Length - 1, length: 1);
             return StringBuilderCache.GetStringAndRelease(sb);
         }
 
-        private string? EnrichTagsWithAasMetadata(string globalTags, ImmutableAzureAppServiceSettings? aasSettings)
+        private static string RemoveScheme(string url)
```
andrewlock, line 88: How dare you. You changed a beautiful switch expression to this. You monster. 😛

tracer/src/Datadog.Trace/Logging/DirectSubmission/Formatting/LogFormatter.cs
```diff
@@ -76,26 +75,69 @@ private static string StringifyGlobalTags(IReadOnlyDictionary<string, string> gl
...
+                  .Append(KeyValueTagSeparator)
+                  .Append(aasSettings.SiteKind)
+                  .Append(TagSeparator);
+            }
+
+            // remove final joiner
+            sb.Remove(sb.Length - 1, length: 1);
+            var aasTags = StringBuilderCache.GetStringAndRelease(sb);
+
+            return string.IsNullOrEmpty(globalTags) ? aasTags : $"{aasTags}{TagSeparator}{globalTags}";
```
andrewlock: nit: given we're already using the string builder cache here, we may as well use it to build the final string too?
```suggestion
            if (!string.IsNullOrEmpty(globalTags))
            {
                sb.Append(TagSeparator)
                  .Append(globalTags);
            }

            return StringBuilderCache.GetStringAndRelease(sb);
```

tracer/src/Datadog.Trace/Logging/DirectSubmission/Formatting/LogFormatter.cs
```diff
@@ -317,7 +349,7 @@ internal void FormatLog<T>(
 
             EnrichTagsStringWithGitMetadata();
 
-            if (Tags is not null && !renderingDetails.HasRenderedTags)
+            if (!string.IsNullOrEmpty(Tags) && !renderingDetails.HasRenderedTags)
```
andrewlock: nit: StringUtil has correct nullability checks 🙂 
```suggestion
            if (!StringUtil.IsNullOrEmpty(Tags) && !renderingDetails.HasRenderedTags)
```

lucaspimentel: Nice, thanks!

tracer/src/Datadog.Trace/Logging/DirectSubmission/Formatting/LogFormatter.cs
lucaspimentel: makes sense, thanks

tracer/src/Datadog.Trace/Logging/DirectSubmission/Formatting/LogFormatter.cs
lucaspimentel: frankly, I think this is abusing the syntax and it not "beautiful" 😅 

```
{ } when url.StartsWith("https://" ...
{ } when url.StartsWith("http://" ...
_ => url
```

also, I think it's adding unnecessary null checks (but I haven't looked at the generated IL to confirm)

level of care low, so I'll change it back if you want 😄

andrewlock: > frankly, I think this is abusing the syntax and it not "beautiful" 😅

Agreed to disagree 😉 

> also, I think it's adding unnecessary null checks (but I haven't looked at the generated IL to confirm)

It does add a null check, yes:

```csharp
private string RemoveScheme(string url)
{
    if (url != null)
    {
        if (url.StartsWith("https://", StringComparison.OrdinalIgnoreCase))
        {
            return url.Substring("https://".Length);
        }
        if (url.StartsWith("http://", StringComparison.OrdinalIgnoreCase))
        {
            return url.Substring("http://".Length);
        }
    }
    return url;
}
```

but using `_` instead of `{ }` removes that, and then the two are identical, but the switch expression is 6 lines instead of 12 😉 

> level of care low, so I'll change it back if you want 😄

Same, I mostly assumed it was AI changing random things for no benefit just to make PRs noisier 😉

lucaspimentel: > _I mostly assumed it was AI_

It's meeee

lucaspimentel: > but using `_` instead of `{ }` removes that, and then the two are identical

Ooh, I like that better, good to know. I was gonna change it now (with `_`), but it's already merged. Sorry 😝

PR #7628, opened 2025-10-06
[Azure Functions] Fix span parenting in ASP.NET Core integration

lucaspimentel:
## Summary of changes

Fixes incorrect span parenting in isolated Azure Functions when using ASP.NET Core integration for HTTP triggers. Worker process spans are now correctly parented to the ASP.NET Core request span, instead of being incorrectly parented to the root host span.

## Reason for change

When using isolated Azure Functions with ASP.NET Core integration and HTTP proxying enabled, spans created in the worker process were being parented to the wrong span, causing disconnected or incorrectly structured traces. This made it difficult to understand the complete request flow and latency attribution.

**Current (incorrect) behavior:**
```
ROOT: azure_functions.invoke: GET /api/httptest [HOST]
  ├─ http.request: GET localhost:40521/api/HttpTest [HOST → WORKER]
  └─ azure_functions.invoke: Http HttpTest [WORKER] ❌ wrong parent
      └─ (worker child spans)
```

**Fixed (correct) behavior:**
```
ROOT: azure_functions.invoke: GET /api/httptest [HOST]
  └─ http.request: GET localhost:40521/api/HttpTest [HOST → WORKER]
      └─ aspnet_core.request [WORKER]
          └─ azure_functions.invoke: Http HttpTest [WORKER] ✅ correct parent
              └─ (worker child spans)
```

## Implementation details

The root cause was that `AsyncLocal` context doesn't flow correctly through Azure Functions middleware, causing the worker's `azure_functions.invoke` span to have no local parent. The instrumentation would then fall back to extracting trace context from gRPC message headers, which contained **stale context** (the host's root span context), resulting in incorrect parenting.

**The Fix**: Use `HttpContext.Items` as an explicit bridge to pass scope between ASP.NET Core and Azure Functions middleware layers:

1. **Store scope in HttpContext.Items** (`AspNetCoreHttpRequestHandler.cs:159-171`)
   - After creating the `aspnet_core.request` scope, store it in `HttpContext.Items[HttpContextActiveScopeKey]`
   - Only done when running in Azure Functions isolated worker

2. **Skip stale gRPC header extraction** (`AzureFunctionsCommon.cs:243-259`)
   - Detect ASP.NET Core integration by checking for `"HttpRequestContext"` key in `FunctionContext.Items`
   - Skip extracting trace context from gRPC message headers (which contain stale host root span context)
   - Only extract headers in non-ASP.NET Core mode (timer triggers, non-proxying HTTP triggers)

3. **Retrieve scope from `HttpContext.Items`** (`AzureFunctionsCommon.cs:287-367`)
   - When `tracer.InternalActiveScope` is null (AsyncLocal didn't flow), call `GetAspNetCoreScope()`
   - Get HttpContext from `FunctionContext.Items["HttpRequestContext"]` (set by `FunctionsHttpProxyingMiddleware`)
   - Get scope from `HttpContext.Items[HttpContextActiveScopeKey]`
   - Use retrieved scope as parent if found, otherwise fall back to extracted context or create root

4. **Added Items property to `IFunctionContext`** (`IFunctionContext.cs:21`)
   - Added `IDictionary<object, object>? Items { get; }` for duck-typed access to FunctionContext.Items

This preserves existing behavior for non-proxying scenarios (timer triggers, non-ASP.NET Core HTTP triggers) while fixing the proxying case.

## Test coverage

Covered by existing tests in `AzureFunctionsTests.cs`.

Fixed test snapshots to reflect correct span counts and hierarchy.

### Before

image

https://dd-dotnet.datadoghq.com/apm/trace/69129102000000009bd07f7872769a84

### After

image

https://dd-dotnet.datadoghq.com/apm/trace/69150c5500000000cb655a7e34732dd6

More recent examples with these changes:
- https://dd-dotnet.datadoghq.com/apm/trace/697be6d40000000000710c25e093bbc4
- https://dd-dotnet.datadoghq.com/apm/trace/6994c3a200000000008c4ab320c9c4b7

## Other details

Fixes APMSVLS-58


tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
```diff
@@ -27,8 +27,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="$(MSBuildThisFileDirectory)..\Datadog.Trace.Manual\Datadog.Trace.Manual.csproj" />
-    <ProjectReference Include="$(MSBuildThisFileDirectory)..\Datadog.Trace.Annotations\Datadog.Trace.Annotations.csproj" />
+    <PackageReference Include="Datadog.Trace" Version="*"/>
+    <PackageReference Include="Datadog.Trace.Annotations" Version="*" />
```
lucaspimentel: TODO: revert this.

Temporary change for testing (need to reference public nuget packages).

lucaspimentel:
@codex review

tracer/src/Datadog.Trace/PlatformHelpers/AspNetCoreHttpRequestHandler.cs
```diff
@@ -153,6 +151,21 @@ private Scope StartAspNetCorePipelineScope(Tracer tracer, Security security, Ias
             httpContext.Items[HttpContextTrackingKey] = new RequestTrackingFeature(originalPath, scope, proxyContext?.Scope);
 #endif
 
+            if (EnvironmentHelpers.IsAzureFunctions())
```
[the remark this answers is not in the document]
lucaspimentel: We are aware of this issue. I'm doing this for _all_ the Azure Functions environment variable probes in a separate PR. Thanks.

lucaspimentel:
@codex review

lucaspimentel:
@claude review (does this work?)

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -311,23 +371,65 @@ _ when type.StartsWith("eventGrid", StringComparison.OrdinalIgnoreCase) => "Even
...
+                    }
+                }
+                else
+                {
+                    Log.Debug("Azure Functions span creation: Could not retrieve AspNetCore scope from HttpContext.Items");
+                }
+            }
+            catch (Exception ex)
+            {
+                Log.Debug(ex, "Azure Functions span creation: Error retrieving AspNetCore scope from HttpContext.Items");
```
bouwkast: Should this be a Error?
```suggestion
                Log.Error(ex, "Azure Functions span creation: Error retrieving AspNetCore scope from HttpContext.Items");
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -269,36 +290,75 @@ _ when type.StartsWith("eventGrid", StringComparison.OrdinalIgnoreCase) => "Even
...
+                if (aspNetCoreScope != null && activeScope == aspNetCoreScope)
                 {
-                    // shouldn't be hit, but better safe than sorry
-                    scope = tracer.StartActiveInternal(OperationName);
-
+                    // The ASP.NET Core span is already active - don't create a new span,
+                    // just update the existing root span's tags to make it a "serverless" span.
+                    // This matches the behavior for isolated worker scenario where we detect
+                    // an existing scope and update it rather than creating a duplicate.
+                    scope = activeScope;
```
bouwkast: 😆 Clicked on this line as I just wanted to check if setting and then returning the scope here is safe / correct as I think it just changes the potential timing of the span closing?

Seems like the AI is thinking the same

tracer/test/snapshots/AzureFunctionsTests.Isolated.V4.AspNetCore.verified.txt
```diff
@@ -2,24 +2,102 @@
   {
     TraceId: Id_1,
     SpanId: Id_2,
+    Name: aspnet_core.request,
+    Resource: GET /api/badrequest,
+    Service: AzureFunctionsAllTriggers,
+    Type: web,
+    ParentId: Id_3,
```
bouwkast, line 9: This is expected correct that `Id_3` is the other process?

tracer/test/snapshots/AzureFunctionsTests.Isolated.V4.AspNetCore.verified.txt
```diff
@@ -33,60 +111,143 @@
...
+      aas.site.type: function,
+      aspnet_core.endpoint: Exception,
+      aspnet_core.route: api/exception,
+      component: aspnet_core,
       env: integration_tests,
-      http-client-handler-type: System.Net.Http.HttpClientHandler,
       http.method: GET,
+      http.request.headers.host: localhost:00000,
+      http.route: api/exception,
       http.status_code: 200,
```
bouwkast: I was just looking at before/after of the snapshots, this is the "new" `ASP.NET Core` span, but I think that this status code should be 500 right?

bouwkast (approved):
Just a few questions but LGTM thanks!

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -238,8 +240,27 @@ _ when type.StartsWith("eventGrid", StringComparison.OrdinalIgnoreCase) => "Even
...
+                            if (isAspNetCoreIntegration)
+                            {
+                                // Skip extraction - will rely on HttpContext.Items bridge or create root span
+                                Log.Debug("Skipping header extraction - HTTP trigger with ASP.NET Core integration detected (HTTP proxying mode)");
+                            }
+                            else
+                            {
+                                // Only extract from gRPC message when NOT using ASP.NET Core integration
+                                extractedContext = ExtractPropagatedContextFromHttp(functionContext, entry.Key as string).MergeBaggageInto(Baggage.Current);
+                                Log.Debug("Extracted trace context from gRPC message (non-ASP.NET Core mode)");
```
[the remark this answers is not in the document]
lucaspimentel: This code path only runs when ASP.NET Core integration is detected (`isAspNetCoreIntegration` is true on line 266), so the scenario where ASP.NET Core is disabled doesn't apply here. The gRPC headers are stale in this mode (they contain the host's root span context), so skipping extraction is correct. Added a comment explaining why.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -253,39 +273,75 @@ _ when type.StartsWith("eventGrid", StringComparison.OrdinalIgnoreCase) => "Even
...
+                // Check if the ASP.NET Core scope is already active
+                if (aspNetCoreScope != null && activeScope == aspNetCoreScope)
                 {
-                    // shouldn't be hit, but better safe than sorry
-                    scope = tracer.StartActiveInternal(OperationName);
+                    // The ASP.NET Core span is already active - don't create a new span,
+                    // just update the existing root span's tags to make it a "serverless" span.
+                    // This matches the behavior for isolated worker scenario where we detect
+                    // an existing scope and update it rather than creating a duplicate.
+                    scope = activeScope;
```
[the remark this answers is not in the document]
lucaspimentel: Fixed. The borrowed ASP.NET Core scope is no longer returned as calltarget state.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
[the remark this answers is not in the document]
lucaspimentel: Fixed. The borrowed ASP.NET Core scope is no longer returned as calltarget state.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
lucaspimentel: Fixed.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
lucaspimentel: Fixed.

tracer/test/snapshots/AzureFunctionsTests.Isolated.V4.AspNetCore.verified.txt
lucaspimentel: Good catch.

🤖 says:

The host's `http.request` (client) span correctly shows `http.status_code: 500`, but the worker's `aspnet_core.request` span shows 200. This is a timing issue - our ASP.NET Core instrumentation reads `HttpResponse.StatusCode` when the span closes, but at that point the Functions exception handler middleware hasn't converted it to 500 yet. This is a pre-existing issue with how the ASP.NET Core span captures status codes in Azure Functions, not specific to this PR. I'll track it separately.

lucaspimentel: I'm not sure that ai is correct. I'll look into this further before merging.

jcstorms1 (approved):

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/Isolated/IFunctionContext.cs
```diff
@@ -18,6 +18,8 @@ internal interface IFunctionContext
     FunctionDefinitionStruct FunctionDefinition { get; }
 
     IEnumerable<KeyValuePair<Type, object?>>? Features { get; }
+
+    IDictionary<object, object>? Items { get; }
```
andrewlock, line 22: Should this be `object?`?
```suggestion
    IDictionary<object, object?>? Items { get; }
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/Isolated/IHttpContextItems.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Azure.Functions;
+
+/// <summary>
+/// Duck type for Microsoft.AspNetCore.Http.HttpContext,
+/// used to avoid a hard assembly reference to Microsoft.AspNetCore.Http.Abstractions
+/// which may not be available in non-ASP.NET Core Azure Functions workers.
+/// </summary>
+internal interface IHttpContextItems
+{
+    IDictionary<object, object> Items { get; }
```
andrewlock, line 20: Should this be 
```suggestion
    IDictionary<object, object?> Items { get; }
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -15,24 +15,26 @@
...
+using Datadog.Trace.Vendors.Serilog.Events;
 
 #nullable enable
 
 namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Azure.Functions
 {
     internal static class AzureFunctionsCommon
     {
-        public const string IntegrationName = nameof(Configuration.IntegrationId.AzureFunctions);
+        private const string HttpRequestContextKey = "HttpRequestContext";
```
andrewlock, line 33: nit: might want a comment showing where this is defined - I thought it was internal, but I think it's a framework thing?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -311,23 +373,65 @@ _ when type.StartsWith("eventGrid", StringComparison.OrdinalIgnoreCase) => "Even
...
+
+            // AsyncLocal context didn't flow - try to get parent scope from HttpContext.Items
+            // This happens in Azure Functions isolated worker where middleware breaks AsyncLocal flow
+            Scope? parentScope = null;
+
+            try
+            {
+                if (functionContext.Items != null &&
+                    functionContext.Items.TryGetValue(HttpRequestContextKey, out var httpContextObj) &&
+                    httpContextObj.TryDuckCast<IHttpContextItems>(out var httpContext) &&
```
andrewlock, line 389: Can `httpContextObj` be null here (but the _key_ is present)?

tracer/src/Datadog.Trace/Tagging/AzureFunctionsTags.cs
```diff
@@ -51,10 +50,20 @@ internal static void SetRootSpanTags(
                 tags.SetTag(Tags.InstrumentationName, ComponentName);
             }
 
-            tags.SetTag(ShortNameTagName, shortName);
-            tags.SetTag(FullNameTagName, fullName);
-            tags.SetTag(BindingSourceTagName, bindingSource);
-            tags.SetTag(TriggerTypeTagName, triggerType);
+            if (tags is AzureFunctionsTags azureFunctionsTags)
```
andrewlock, line 53: Nice 👍

andrewlock (approved):
LGTM, apart from a couple of nits: and the existing question about the snapshots

tracer/test/snapshots/AzureFunctionsTests.Isolated.V4.AspNetCore.verified.txt
lucaspimentel: Looked into this. Root cause: the `aspnet_core.request` span's 500 status is normally set by `OnHostingUnhandledException`, which listens to the ASP.NET Core `Microsoft.AspNetCore.Hosting.UnhandledException` diagnostic event. In the Isolated worker the worker's `FunctionExecutionMiddleware` catches the exception before it escapes to ASP.NET Core, so that event never fires. At span close we fall back to `httpContext.Response.StatusCode`, which still reads 200 because the worker hasn't written its 500 response onto the ASP.NET Core pipeline yet. This is pre-existing behavior, not a regression from this PR. Tracking as a separate follow-up.

🤖

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/Isolated/IFunctionContext.cs
lucaspimentel: Fixed

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/Isolated/IHttpContextItems.cs
lucaspimentel: Fixed

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
lucaspimentel: Good catch. Added comment with links.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
lucaspimentel: Yes, but `TryDuckCast()` handles nulls so we don't need a extra check.


edit: actually, I think I'll add it anyway. I don't like when extension methods handle null. Looks weird.

tracer/src/Datadog.Trace/PlatformHelpers/AspNetCoreHttpRequestHandler.cs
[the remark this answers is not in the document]
lucaspimentel: This was done.

tracer/test/snapshots/AzureFunctionsTests.Isolated.V4.AspNetCore.verified.txt
lucaspimentel: This is technically pre-existing behavior, but we were not using this span before and now we are, so I need to fix this.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
lucaspimentel: added a few null checks, thanks

tracer/test/snapshots/AzureFunctionsTests.Isolated.V4.AspNetCore.verified.txt
lucaspimentel: I did a little refactoring to try to capture the exception and add it to the asp.net core span.

lucaspimentel: Snapshot fixed.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -269,36 +306,72 @@ _ when type.StartsWith("eventGrid", StringComparison.OrdinalIgnoreCase) => "Even
...
+                        rootSpan.Tags,
+                        shortName: tags.ShortName,
+                        fullName: tags.FullName,
                         bindingSource: rootSpan.Tags is AzureFunctionsTags t ? t.BindingSource : null,
-                        triggerType: triggerType);
+                        triggerType: tags.TriggerType);
+
+                    rootSpan.Type = SpanType; // "serverless"
+                    rootSpan.ResourceName = $"{tags.TriggerType} {tags.ShortName}";
+                }
```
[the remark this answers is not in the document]
lucaspimentel: We intentionally don't want to create a second span here. It introduces unnecessary noise to the trace.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -254,8 +269,30 @@ _ when type.StartsWith("eventGrid", StringComparison.OrdinalIgnoreCase) => "Even
...
+                                // The gRPC message headers contain STALE context from the host process
+                                // (the host's root span), so extracting them would create an incorrect parent.
+                                // Instead, we rely on the HttpContext.Items bridge to get the ASP.NET Core scope.
+                                Log.Debug("Skipping header extraction - HTTP trigger with ASP.NET Core integration detected (HTTP proxying mode)");
+                            }
+                            else
+                            {
+                                // Only extract from gRPC message when NOT using ASP.NET Core integration
+                                extractedContext = ExtractPropagatedContextFromHttp(functionContext, entry.Key as string).MergeBaggageInto(Baggage.Current);
+                                Log.Debug("Extracted trace context from gRPC message (non-ASP.NET Core mode)");
```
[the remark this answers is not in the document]
lucaspimentel: AFAIK this can't happen in practice, but I agree it's better to keep the spans in the same trace even if it's the wrong parent. This would fallback to the current state of things, so it won't be _worse_. Accepting commit.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
andrewlock: Yeah, for the record, my concern was that if `httpContextObj` was `null`, `httpContext` would be `null`, which would cause `httpContext.Items` to throw 😃

PR #7629, opened 2025-10-07
Increase Docker run step resiliency for Profiler integration tests

NachoEchevarria:
This change sets retryCountForRunCommand: 3 in the run-in-docker.yml template invocation for the BuildAndRunProfilerCpuLimitTests step of the profilesr integration tests (Linux).

We’ve seen rare transient Docker daemon errors on hosted agents during container creation:

`unable to start unit "...scope": Message recipient disconnected from message bus without replying`

These failures typically return exit code 125 and are not related to our build logic or test code. By enabling up to three automatic retries at the task level, the pipeline can recover from occasional runc / systemd / dbus cgroup creation hiccups without manual reruns.

### DIFF
--- a/.azure-pipelines/ultimate-pipeline.yml
+++ b/.azure-pipelines/ultimate-pipeline.yml
@@ -2771,6 +2771,7 @@ stages:
         useNativeSdkVersion: $(useNativeSdkVersion)
         command: "BuildProfilerSamples"
         apiKey: $(DD_LOGGER_DD_API_KEY)
+        retryCountForRunCommand: 3
 
     - template: steps/run-in-docker.yml
       parameters:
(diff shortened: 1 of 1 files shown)


bouwkast (approved):

.azure-pipelines/ultimate-pipeline.yml
```diff
@@ -2779,6 +2780,7 @@ stages:
         command: "BuildAndRunProfilerCpuLimitTests"
         extraArgs: "--cpus 2 --env CONTAINER_CPUS=1"
         apiKey: $(DD_LOGGER_DD_API_KEY)
+        retryCountForRunCommand: 3
```
andrewlock, line 2783: nit: we generally try to _only_ re-run the build stages instead of re-running all the integration tests, because if an integration test fails we'll end up running it 3 times, which can end up causing timeouts overall...

Can/should we split this into separate "BuildProfilerCpuLimitTests" (with retries) and `RunProfilerCpuLimitTests`?

PR #7632, opened 2025-10-08
Small OpenTelemetry Metrics fixes and optimizations

zacharycmontoya:
## Summary of changes
Adds small optimizations to the processing / exporting of OpenTelemetry Metrics

## Reason for change
I identified some small improvements we could make.

## Implementation details
Makes two improvements:
- Adds more comprehensive type-checking for the `IsLongType` check
- Removes allocations for the anonymous method that creates HttpRequests for OTLP exports

## Test coverage
Existing tests should enforce correctness

## Other details
N/A


tracer/src/Datadog.Trace/OpenTelemetry/Metrics/MetricStreamIdentity.cs
```diff
@@ -35,8 +35,11 @@ public MetricStreamIdentity(Instrument instrument, InstrumentType instrumentType
             InstrumentType = instrumentType;
             IsHistogram = instrumentType == InstrumentType.Histogram;
 
-            IsLongType = instrument.GetType().IsGenericType &&
-                         instrument.GetType().GetGenericArguments()[0] == typeof(long);
+            Type? instrumentGenericType = instrument.GetType().IsGenericType ? instrument.GetType().GetGenericArguments()[0] : null;
+            IsLongType = instrumentGenericType == typeof(long) ||
```
andrewlock, line 39: do we need to support unsigned versions too?

zacharycmontoya: Nope! The OpenTelemetry .NET SDK does not, so I'm operating under the assumption that we do not need to either. I believe that's based on the protobuf type being a [sfixed64 type](https://github.com/open-telemetry/opentelemetry-proto/blob/5b94857a80582a7aad5be8161c80a3d0a954c4b2/opentelemetry/proto/metrics/v1/metrics.proto#L419C5-L419C25) which maps to a `long` type in C# according to the [proto 3 language guide](https://protobuf.dev/programming-guides/proto3/), since it's a signed 64-bit int

PR #7638, opened 2025-10-09
[OTEL] Moves OTelMetrics Under OpenTelemetry.Metrics

link04:
## Summary of changes
Moving `OTelMetrics` Under `OpenTelemetry.Metrics`.

## Reason for change
To align more with Otel and have better structure as we continue to add suppport for OpenTelemetry APIs.

### DIFF
--- a/tracer/src/Datadog.Trace/OpenTelemetry/Metrics/AggregationTemporality.cs
+++ b/tracer/src/Datadog.Trace/OpenTelemetry/Metrics/AggregationTemporality.cs
@@ -13,7 +13,7 @@ namespace Datadog.Trace.OpenTelemetry.Metrics;
 /// Represents the aggregation temporality of a metric.
 /// Values match the OTLP protobuf specification.
 /// </summary>
-public enum AggregationTemporality
+internal enum AggregationTemporality
 {
     /// <summary>
     /// Delta temporality, representing changes since the last measurement.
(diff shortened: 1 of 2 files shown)


tracer/src/Datadog.Trace/OTelMetrics/AggregationTemporality.cs
bouwkast, line 16: Just realized this is `public` 

Hmmmm I don't think it should have been

tracer/src/Datadog.Trace/OpenTelemetry/Metrics/AggregationTemporality.cs
bouwkast: Just realizing that this is `public` I don't think it should have been.

Breaking change to fix it technically.
Breaking change to update the namespace.

tracer/src/Datadog.Trace/OTelMetrics/AggregationTemporality.cs
bouwkast: Ignore this one I lost the comment and had to retype it 😆

tracer/src/Datadog.Trace/OpenTelemetry/Metrics/AggregationTemporality.cs
zacharycmontoya: While these are technically public, they're not exposed to the user since this is not in the `Manual` assembly. That's likely why it got by

bouwkast: Ahh yes

bouwkast: @link04 can we change this to `internal`?

bouwkast: We should make note of it in the next release notes for transparency, but probably fine :)

link04: Ah my bad, just fixed it in the lastest commit: https://github.com/DataDog/dd-trace-dotnet/pull/7638/commits/0ff8ff4c4e815c71950886278acba25c7dd97482

PR #7645, opened 2025-10-13
Compress telemetry payload

NachoEchevarria:
## Summary of changes

This PR enables by default the telemetry compression. This feature allows to compress the telemetry payloads in oder to send smaller messages.

The variable DD_INSTRUMENTATION_TELEMETRY_COMPRESSION_METHOD controls the compression method. Now, it only accepts "gzip" or "none". By default, this variable is set as "gzip".

A PR has been created in dd-go to add this new configration value: https://github.com/DataDog/dd-go/pull/204882

## Test coverage

This feature has been tested by sending telemetry to the UI both with agent and in agentless mode.


tracer/src/Datadog.Trace/Telemetry/Transports/JsonTelemetryTransport.cs
```diff
@@ -51,9 +57,13 @@ public async Task<TelemetryPushResult> PushTelemetry(TelemetryData data)
...
                 var serializedData = SerializeTelemetry(data);
                 var bytes = Encoding.UTF8.GetBytes(serializedData);
-
                 var request = _requestFactory.Create(_endpoint);
 
+                if (_telemetryGzipCompressionEnabled)
+                {
+                    bytes = CompressGzip(bytes);
+                }
+
```
andrewlock, line 67: It's unfortunate that we do so much allocation here 🤔 We serialize the data to a string, _then_ convert it to a byte array and _then_ gzip it again. I think we can avoid at least _some_ of this allocation by moving the gzip compression into `SerializeTelemetry()` e.g. something like this:

```csharp
internal static byte[] SerializeTelemetryWithGzip<T>(T data)
{
    using var memStream = new MemoryStream();
    using (var zipStream = new GZipStream(memStream, CompressionMode.Compress, true))
    {
        using var streamWriter = new StreamWriter(zipStream);
        using var jsonWriter = new JsonTextWriter(streamWriter);
        var serializer = new JsonSerializer { NullValueHandling = NullValueHandling.Ignore, ContractResolver = new DefaultContractResolver { NamingStrategy = new SnakeCaseNamingStrategy(), }, Formatting = Formatting.None };

        serializer.Serialize(jsonWriter, data);
    }

    return memStream.ToArray();
}
``` 

However, I think there's _another_ option, in that we could potentially do "chunked" uploads, like we do [for the tracer flare](https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/src/Datadog.Trace/Logging/TracerFlare/TracerFlareApi.cs#L55). That would mean we don't have to do _any_ up-front allocation, which should reduce overhead even further 🤔

NachoEchevarria: I have implemented your solution. Thanks!

tracer/src/Datadog.Trace/Telemetry/Transports/JsonTelemetryTransport.cs
```diff
@@ -73,35 +85,36 @@ public async Task<TelemetryPushResult> PushTelemetry(TelemetryData data)
...
 
                 TelemetryFactory.Metrics.RecordCountTelemetryApiRequests(endpointMetricTag);
-                using var response = await request.PostAsync(new ArraySegment<byte>(bytes), "application/json").ConfigureAwait(false);
+
+                using var response = await request.PostAsync(new ArraySegment<byte>(bytes), "application/json", _telemetryGzipCompressionEnabled ? "gzip" : null).ConfigureAwait(false);
                 TelemetryFactory.Metrics.RecordCountTelemetryApiResponses(endpointMetricTag, response.GetTelemetryStatusCodeMetricTag());
                 if (response.StatusCode is >= 200 and < 300)
                 {
-                    Log.Debug("Telemetry sent successfully");
+                    Log.Debug("telemetry sent successfully. Compression {Compression}", _telemetryCompressionMethod);
```
andrewlock:
```suggestion
                    Log.Debug("Telemetry sent successfully. Compression {Compression}", _telemetryCompressionMethod);
```

tracer/test/Datadog.Trace.IntegrationTests/TelemetryTransportTests.cs
```diff
@@ -22,6 +22,8 @@ namespace Datadog.Trace.IntegrationTests
 {
     public class TelemetryTransportTests
     {
+        private const string _gzipCompression = "gzip";
+        private const string _noCompression = "none";
```
andrewlock: nit: IDEs will say to use Pascal case for these
```suggestion
        private const string GzipCompression = "gzip";
        private const string NoCompression = "none";
```

tracer/test/Datadog.Trace.TestHelpers/MockTelemetryAgent.cs
```diff
@@ -129,25 +130,29 @@ public void Dispose()
...
+            static TelemetryData DeserializeV2(Stream inputStream, string requestType, bool compressed)
             {
                 if (!TelemetryConverter.V2Serializers.TryGetValue(requestType, out var serializer))
                 {
                     throw new Exception($"Unknown V2 telemetry request type {requestType}");
                 }
 
-                using var sr = new StreamReader(inputStream);
-                var text = sr.ReadToEnd();
-                var tr = new StringReader(text);
```
andrewlock, line 149: I wonder why we did this before, instead of what you did in this PR? 🤔 Seems odd that we would want to do that 😄 Maybe it was for easier debugging 🤷‍♂️

tracer/src/Datadog.Trace/Telemetry/Transports/AgentTelemetryTransport.cs
```diff
@@ -12,8 +12,8 @@ namespace Datadog.Trace.Telemetry.Transports;
 
 internal class AgentTelemetryTransport : JsonTelemetryTransport
 {
-    public AgentTelemetryTransport(IApiRequestFactory requestFactory, bool debugEnabled)
-        : base(requestFactory, debugEnabled)
+    public AgentTelemetryTransport(IApiRequestFactory requestFactory, bool debugEnabled, string telemetryCompressionMethod)
```
bouwkast, line 15: Is there a minimum agent version for this? 
I think telemetry skips the agent though so maybe not a concern?

tracer/src/Datadog.Trace/Telemetry/Transports/JsonTelemetryTransport.cs
```diff
@@ -73,35 +85,36 @@ public async Task<TelemetryPushResult> PushTelemetry(TelemetryData data)
                 }
 
                 TelemetryFactory.Metrics.RecordCountTelemetryApiRequests(endpointMetricTag);
-                using var response = await request.PostAsync(new ArraySegment<byte>(bytes), "application/json").ConfigureAwait(false);
+
+                using var response = await request.PostAsync(new ArraySegment<byte>(bytes), "application/json", _telemetryGzipCompressionEnabled ? "gzip" : null).ConfigureAwait(false);
```
bouwkast, line 89: Is there a possibility that we get rejected due to compressed payloads and would need to fallback to uncompressed?

tracer/src/Datadog.Trace/Telemetry/Transports/AgentTelemetryTransport.cs
NachoEchevarria: According to the agent team, compressed telemetry is supported from version 7.34 (Dec 2021). Actually, the agent is like a proxy and does not do anything with the headers or data but sending it directly to the backend. Before this version, the endpoint did not exist in the agent, so I think that we are safe with the compression. We already checked before this PR for agents without the endpoint.

NachoEchevarria:
Thanks for the feedback and reviews!

PR #7650, opened 2025-10-14
Split BuildAndRunProfilerCpuLimitTests

NachoEchevarria:
## Summary of changes

This PR splits the target BuildAndRunProfilerCpuLimitTests into BuildProfilerCpuLimitTests and RunProfilerCpuLimitTests. We want to retry the build of the test projects to avoid flaky errors but not when we run the tests to avoid timeouts in case of a failing test.

While target BuildAndRunProfilerCpuLimitTests is no longer used in the pipeline, it has been left for local testing.


tracer/test/Datadog.Trace.Tests/ContinuousProfiler/ProfilerAvailabilityHelperTests.cs
```diff
@@ -29,7 +29,7 @@ public class ProfilerAvailabilityHelperTests
     [SkippableFact]
     public void IsContinuousProfilerAvailable_OnUnsupportedPlatforms_ReturnsFalse()
     {
-        // Skip on platforms that it's available
+        // Skip on platforms that it's available.
```
NachoEchevarria, line 32: Dummy change to launch the profiler tests in this PR

tracer/build/_build/Build.Profiler.Steps.cs
```diff
@@ -328,13 +328,43 @@ partial class Build
...
-        .Description("Run the profiler container tests")
+        .Description("Builds and runs the profiler container tests")
         .Requires(() => IsLinux && !IsArm64)
         .Executes(() =>
         {
             BuildAndRunProfilerIntegrationTestsInternal("(Category=CpuLimitTest)");
         });
 
+    Target RunProfilerCpuLimitTests => _ => _
+        .After(BuildProfilerSamples)
```
andrewlock:
```suggestion
        .After(BuildProfilerSamples, BuildProfilerCpuLimitTests)
```

PR #7651, opened 2025-10-14
[Tracer] Add new process tags

vandonr:
## Summary of changes

this is the first PR introducing process tags to dotnet, see [RFC](https://docs.google.com/document/d/1AFdLUuVk70i0bJd5335-RxqsvwAV9ovAqcO2z5mEMbA)
The goal is to pass as much data as possible to the backend to enable customers to set the "correct" names for their services.

## Reason for change

overarching product need

## Implementation details

 - didn't add the config to MutableSettings because I don't think it's something that needs to support live updates ?

### DIFF
--- a/tracer/src/Datadog.Trace/ProcessTags.cs
+++ b/tracer/src/Datadog.Trace/ProcessTags.cs
@@ -11,6 +11,7 @@ using System.IO;
 using System.Text;
 using Datadog.Trace.Configuration;
 using Datadog.Trace.Processors;
+using Datadog.Trace.Util;
 
 namespace Datadog.Trace;
 
@@ -49,7 +50,7 @@ internal static class ProcessTags
         };
 
         // then normalize values and put all tags in a string
-        var serializedTags = new StringBuilder();
+        var serializedTags = StringBuilderCache.Acquire();
         foreach (var kvp in tags)
         {
             if (!string.IsNullOrEmpty(kvp.Value))
@@ -59,7 +60,7 @@ internal static class ProcessTags
         }
 
         serializedTags.Remove(serializedTags.Length - 1, length: 1); // remove last comma
-        return serializedTags.ToString();
+        return StringBuilderCache.GetStringAndRelease(serializedTags);
     }
 
     private static string NormalizeTagValue(string tagValue)


tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -0,0 +1,79 @@
...
+    public static string SerializedTags
+    {
+        get => _lazySerializedTags.Value;
+    }
+
+    private static Dictionary<string, string> LoadBaseTags()
+    {
+        var tags = new Dictionary<string, string>();
+
+        if (!Tracer.Instance.Settings.PropagateProcessTags)
```
vandonr: is this the appropriate way to get settings here ?

tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            // do not collect anything when disabled
+            return tags;
```
andrewlock: nit, if you return `null` instead you can avoid the allocations of the above dictionary in this path

tracer/src/Datadog.Trace/ProcessTags.cs
andrewlock: In short - probably not 😄 

- Accessing global statics like this makes the code it untestable
- It can contribute to confusing flake in general as you're initializing an "in process" tracer object (with all the associated background processes etc)


I see a couple of alternatives:
- Make `SerializedTags()` a method and pass in the settings object (or just a boolean, that's all you need)
- Just bypass calling this type entirely if the setting shows it's disabled? If we cache the serialized result anyway, then that's prob a bit cleaner?

tracer/test/Datadog.Trace.Tests/ProcessTagsTests.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    public ProcessTagsTests()
+    {
+        ProcessTags.ResetForTests();
+    }
+
+    [Fact]
+    public void EmptyIfDisabled()
+    {
+        Tracer.Configure(TracerSettings.Create(new Dictionary<string, object>()));
```
andrewlock: You really don't want to couple this basic testing to creating the _full_ `Tracer` instance etc IMO. There's really 3 different sets of tests that should be separated out IMO:

- Tests that `ProcessTags` generates the correct values in general (here)
- Tests that the _setting_ has the correct value (should be in `TracerSettingsTests`)
- Tests that the formatted message pack payload etc includes the process tags - you _can_ do that with a full `Tracer`  instance but you should never be calling `Tracer.Configure()` - use `TracerHelper.Create()` instead if you need to (and make sure to dispose it)

tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -0,0 +1,68 @@
...
+    public static string SerializedTags
+    {
+        get => _lazySerializedTags.Value;
+    }
+
+    private static Dictionary<string, string> LoadBaseTags()
+    {
+        var tags = new Dictionary<string, string>();
+
+        var entrypointFullName = Assembly.GetEntryAssembly()?.EntryPoint?.DeclaringType?.FullName;
```
raphaelgavache: if I'm not mistaken there are at least 2 different `entrypoint.type` in dotnet, a binary or a .dll
similarly in java the split is done with entrypoint.type:jar_name/main_class

vandonr: yes, I was thinking of adding those dotnet specific details in an other PR, where we could discuss the specifics.
I can see several entry point types : 
 - Console app
 - GUI (winforms/WPF)
 - ASP net webapp
 - windows service
 - Azure function
 - triggered by another library ? (like tests ?)
 - Xamarin mobile app ?

Not sure the distinction exe/dll is super relevant, because a dll can behave the same way as an exe when launched with `dotnet run`.

andrewlock: Most importantly, `Assembly.GetEntryAssembly()` [doesn't work for web apps](https://stackoverflow.com/questions/4277692/getentryassembly-for-web-applications). We have a helper to try to grab this: https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/src/Datadog.Trace/Configuration/EntryAssemblyLocator.cs#L16

But an important part to understand and work around is that this might _not_ be available on app startup... At some "random" later point (i.e. during a reques), we _eventually_ will  be able to access the "real" entrypoint for the application.

We already have to work around this currently by retrying multiple times if it fails the first 100 times; we probably will need to do something similar here: https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/src/Datadog.Trace/Configuration/GitMetadataTagsProvider.cs#L150-L154

I'm not sure what the implications of this are for the process tags feature in general, but it ultimately means that we probably need to add an instrumentation to make sure we trigger loading the entry assembly at the right point (inside a request for owin/aspnet apps, or just before flushing the first span for everything else).

tracer/src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
```diff
@@ -603,6 +604,19 @@ private int WriteTags(ref byte[] bytes, int offset, in SpanModel model, ITagProc
                 }
             }
 
+            // Process tags will be sent only once per trace
+            if (model.IsFirstSpanInChunk && model.TraceChunk.ShouldPropagateProcessTags)
```
raphaelgavache: don't forget to apply this only to the first span in payload (1st chunk only), like in these implementations
- https://github.com/DataDog/dd-trace-java/pull/9657
- https://github.com/DataDog/dd-trace-go/pull/4010

tracer/src/Datadog.Trace/ProcessTags.cs
vandonr: it _should_ happen just before the first flush, because we query this in SpanMessagePackFormatter, so I'm not too worried about it not being available at app startup.
Where it could become trouble is if the value is flaky because of that. If the customer writes a rule to set the service name based on this value, but it's only here in 50% of the spans (or any % that is != 100), that no good. In that case, one could even argue it's better to not have the value at all (to not tempt users into sin)

lucaspimentel:
@codex review

tracer/src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
```diff
@@ -603,6 +604,19 @@ private int WriteTags(ref byte[] bytes, int offset, in SpanModel model, ITagProc
...
+            {
+                var processTags = ProcessTags.SerializedTags;
+                var processTagsRawBytes = MessagePackStringCache.GetProcessTagsBytes(processTags);
+                if (!string.IsNullOrEmpty(processTags))
+                {
+                    count++;
+                    offset += MessagePackBinary.WriteStringBytes(ref bytes, offset, _processTagsNameBytes);
+                    offset += MessagePackBinary.WriteRaw(ref bytes, offset, processTagsRawBytes);
+                }
+            }
```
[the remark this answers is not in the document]
lucaspimentel: > The comment says “Process tags will be sent only once per trace”, but...

The comment is the issue 😅 it's once per trace _payload_.

tracer/src/Datadog.Trace/Agent/MessagePack/MessagePackStringCache.cs
```diff
@@ -34,6 +34,7 @@ internal static class MessagePackStringCache
 
     // NOTE: all of these can be cached in SpanMessagePackFormatter as static byte[]
     // since they never change over the lifetime of a process
+    private static CachedBytes _processTags;
```
lucaspimentel: See the comment above 😅 

> NOTE: all of these can be cached in SpanMessagePackFormatter as `static byte[]` since they never change over the lifetime of a process

See `_languageValueBytes` and `_runtimeIdValueBytes` in `SpanMessagePackFormatter` for examples.

lucaspimentel: Some of the ones below should follow _that_ pattern and not be here in`MessagePackStringCache` using `CachedBytes`.

tracer/src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
```diff
@@ -75,6 +75,7 @@ internal class SpanMessagePackFormatter : IMessagePackFormatter<TraceChunkModel>
         private readonly byte[] _originNameBytes = StringEncoding.UTF8.GetBytes(Trace.Tags.Origin);
         private readonly byte[] _lastParentIdBytes = StringEncoding.UTF8.GetBytes(Trace.Tags.LastParentId);
         private readonly byte[] _baseServiceNameBytes = StringEncoding.UTF8.GetBytes(Trace.Tags.BaseService);
+        private readonly byte[] _processTagsNameBytes = StringEncoding.UTF8.GetBytes(Tags.ProcessTags);
```
lucaspimentel: Since the _value_ doesn't change, you can cache the bytes right here. Something like:

```suggestion
        private readonly byte[] _processTagsNameBytes = StringEncoding.UTF8.GetBytes(Tags.ProcessTags);
        private readonly byte[] _processTagsValueBytes = StringEncoding.UTF8.GetBytes(ProcessTags.SerializedTags);
```

No need to use `MessagePackStringCache`.

tracer/src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
```diff
@@ -603,6 +604,19 @@ private int WriteTags(ref byte[] bytes, int offset, in SpanModel model, ITagProc
                 }
             }
 
+            // Process tags will be sent only once per trace
```
lucaspimentel:
```suggestion
            // Process tags will be sent only once per trace payload
```

tracer/src/Datadog.Trace/Agent/MessagePack/TraceChunkModel.cs
```diff
@@ -200,9 +207,10 @@ public SpanModel GetSpanModel(int spanIndex)
         return new SpanModel(
             span,
             this,
-            isLocalRoot: isLocalRoot,
-            isChunkOrphan: isChunkOrphan,
-            isFirstSpanInChunk: isFirstSpan);
+            isLocalRoot,
+            isChunkOrphan,
+            isFirstSpan,
+            IsFirstChunkInBuffer);
```
lucaspimentel: You don't need to pass this to the span. In the formatter, you can access `model.TraceChunk.IsFirstChunkInBuffer` directly.

tracer/src/Datadog.Trace/Agent/MessagePack/SpanModel.cs
```diff
@@ -31,17 +31,21 @@ internal readonly struct SpanModel
 
     public readonly bool IsFirstSpanInChunk;
 
+    public readonly bool IsInFirstChunkOfBuffer;
```
lucaspimentel: If this is the same as `TraceChunk.IsFirstChunkInBuffer`, then we don't need to add it. The span has a reference to the trace chunk so you can access `span.TraceChunk.IsFirstChunkInBuffer`.

tracer/src/Datadog.Trace/Agent/MessagePack/TraceChunkModel.cs
```diff
@@ -33,6 +33,8 @@ internal readonly struct TraceChunkModel
 
     public readonly int? SamplingPriority = null;
 
+    public readonly bool IsFirstChunkInBuffer = false;
```
lucaspimentel: Should we call this `IsFirstChunkInPayload`?

tracer/src/Datadog.Trace/Agent/SpanBuffer.cs
```diff
@@ -91,7 +91,7 @@ public WriteStatus TryWrite(ArraySegment<Span> spans, ref byte[] temporaryBuffer
                 // to get the other values we need (sampling priority, origin, trace tags, etc) for now.
                 // the idea is that as we refactor further, we can pass more than just the spans,
                 // and these values can come directly from the trace context.
-                var traceChunk = new TraceChunkModel(spans, samplingPriority);
+                var traceChunk = new TraceChunkModel(spans, samplingPriority, TraceCount == 0);
```
lucaspimentel: for readability
```suggestion
                var traceChunk = new TraceChunkModel(spans, samplingPriority, isFirstChunkInBuffer: TraceCount == 0);
```

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
```diff
@@ -168,6 +168,11 @@ internal static partial class ConfigurationKeys
         /// <seealso cref="TracerSettings.HeaderTags"/>
         public const string GrpcTags = "DD_TRACE_GRPC_TAGS";
 
+        /// <summary>
+        /// Propagate the process tags in every supported payload
+        /// </summary>
+        public const string PropagateProcessTags = "DD_EXPERIMENTAL_PROPAGATE_PROCESS_TAGS_ENABLED";
```
lucaspimentel, line 174: Why use the `DD_EXPERIMENTAL_` prefix if it's enabled by default?

tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -0,0 +1,68 @@
...
+        return tags;
+    }
+
+    /// <summary>
+    /// From the full path of a directory, get the name of the leaf directory.
+    /// </summary>
+    private static string GetLastPathSegment(string directoryPath)
+    {
+        // Path.GetFileName returns an empty string if the path ends with a '/'.
+        // We could use Path.TrimEndingDirectorySeparator instead of the trim here, but it's not available on .netframework
```
lucaspimentel:
```suggestion
        // We could use Path.TrimEndingDirectorySeparator instead of the trim here, but it's not available on .NET Framework
```

tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -0,0 +1,68 @@
...
+        return Path.GetFileName(directoryPath.TrimEnd('\\', '/'));
+    }
+
+    private static string GetSerializedTags()
+    {
+        return string.Join(",", LoadBaseTags().OrderBy(kv => kv.Key).Select(kv => $"{kv.Key}:{NormalizeTagValue(kv.Value)}"));
```
lucaspimentel: We should avoid using LINQ (`OrderBy()`, `Select()`). It will allocate delegates and iterators on the heap.

tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /// <summary>
+    /// From the full path of a directory, get the name of the leaf directory.
+    /// </summary>
+    private static string GetLastPathSegment(string directoryPath)
+    {
+        // Path.GetFileName returns an empty string if the path ends with a '/'.
+        // We could use Path.TrimEndingDirectorySeparator instead of the trim here, but it's not available on .netframework
+        return Path.GetFileName(directoryPath.TrimEnd('\\', '/'));
```
lucaspimentel: This allocates for the `params char[]` (even in .NET 6, surprisingly).
Since paths will normally only have either `/` or `\`, but not both, it might be better to do
```csharp
directoryPath.TrimEnd('\\').TrimEnd('/')
```

`TrimEnd()` itself returns the same string without allocating if there's nothing to trim.

tracer/src/Datadog.Trace/Tags.cs
```diff
@@ -506,6 +506,12 @@ public static partial class Tags
         /// </summary>
         internal const string RuntimeFamily = "_dd.runtime_family";
 
+        /// <summary>
+        /// contains a serialized list of process tags, that can be used in the backend for service renaming
```
lucaspimentel:
```suggestion
        /// Contains a serialized list of process tags, that can be used in the backend for service renaming.
```

tracer/src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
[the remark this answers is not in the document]
vandonr: ah yes 🤦 

if this is the main issue it found, I'd say it's pretty good :D

tracer/src/Datadog.Trace/Agent/MessagePack/TraceChunkModel.cs
vandonr: yes why not

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
vandonr: hmm, it's not supposed to be enabled by default, at least not for now.
see https://github.com/DataDog/dd-trace-dotnet/blob/cc4b720462fb97eb9da59676a4fc075a6c3c1195/tracer/src/Datadog.Trace/Configuration/TracerSettings.cs#L105-L108

tracer/src/Datadog.Trace/Agent/MessagePack/MessagePackStringCache.cs
vandonr: we can fix that in a different PR I guess

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
lucaspimentel: I see `.AsBool(false)`, but also `|| ExperimentalFeaturesEnabled.Contains(...)` which will be `true`.

tracer/src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
```diff
@@ -68,6 +68,12 @@ internal class SpanMessagePackFormatter : IMessagePackFormatter<TraceChunkModel>
         private readonly byte[] _runtimeIdNameBytes = StringEncoding.UTF8.GetBytes(Trace.Tags.RuntimeId);
         private readonly byte[] _runtimeIdValueBytes = StringEncoding.UTF8.GetBytes(Tracer.RuntimeId);
 
+        // using a Lazy here to make sure we don't compute the value of the process tags too early in the life of the app,
+        // some values may neeb a bit of time to be accessible.
```
lucaspimentel:
```suggestion
        // some values may need a bit of time to be accessible.
```

tracer/src/Datadog.Trace/Agent/MessagePack/TraceChunkModel.cs
```diff
@@ -200,9 +207,9 @@ public SpanModel GetSpanModel(int spanIndex)
         return new SpanModel(
             span,
             this,
-            isLocalRoot: isLocalRoot,
-            isChunkOrphan: isChunkOrphan,
-            isFirstSpanInChunk: isFirstSpan);
+            isLocalRoot,
+            isChunkOrphan,
+            isFirstSpan);
```
lucaspimentel: We like having the parameter names sometimes to avoid getting arguments mixed up. Since they are all bools, it would be easy to switch the order around the the code will still compile. This way as least any errors are more visible when reading the code.

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -33,7 +33,7 @@ namespace Datadog.Trace.Configuration
     public record TracerSettings
     {
         private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor<TracerSettings>();
-        private static readonly HashSet<string> DefaultExperimentalFeatures = ["DD_TAGS"];
+        private static readonly HashSet<string> DefaultExperimentalFeatures = ["DD_TAGS", ConfigurationKeys.PropagateProcessTags];
```
lucaspimentel: Doesn't this enable process tags by default?

tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -0,0 +1,78 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Text;
+using Datadog.Trace.Processors;
+using Datadog.Trace.VendoredMicrosoftCode.System.Collections.Immutable;
+using Datadog.Trace.Vendors.Newtonsoft.Json.Utilities;
```
lucaspimentel: Not used.

```suggestion
using System.Reflection;
using System.Text;
using Datadog.Trace.Processors;
```

tracer/src/Datadog.Trace/Agent/MessagePack/TraceChunkModel.cs
```diff
@@ -63,15 +65,18 @@ internal readonly struct TraceChunkModel
...
     public readonly bool IsApmEnabled = true;
 
     /// <summary>
     ///     Initializes a new instance of the <see cref="TraceChunkModel"/> struct.
     /// </summary>
     /// <param name="spans">The spans that will be within this <see cref="TraceChunkModel"/>.</param>
     /// <param name="samplingPriority">Optional sampling priority to override the <see cref="TraceContext"/> sampling priority.</param>
-    public TraceChunkModel(in ArraySegment<Span> spans, int? samplingPriority = null)
-        : this(spans, TraceContext.GetTraceContext(spans), samplingPriority)
+    /// <param name="isFirstChunkInPayload">marks if this is the first chunk being written to the buffer that then gets sent to the agent</param>
```
lucaspimentel:
```suggestion
    /// <param name="isFirstChunkInPayload">Indicates if this is the first trace chunk being written to the output buffer.</param>
```

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
lucaspimentel: Oh wait. I was getting `DefaultExperimentalFeatures` and `ExperimentalFeaturesEnabled` mixed up.

lucaspimentel: Still, why `EXPERIMENTAL`?

tracer/test/Datadog.Trace.Tests/Agent/MessagePack/SpanMessagePackFormatterTests.cs
```diff
@@ -126,6 +127,49 @@ public void SerializeSpans()
...
+    [InlineData(true, false)]
+    [InlineData(false, true)]
+    public void ProcessTags_Serialization(bool enabled, bool firstChunk)
+    {
+        var formatter = SpanFormatterResolver.Instance.GetFormatter<TraceChunkModel>();
+
+        var settings = new TracerSettings(new NameValueConfigurationSource(new NameValueCollection { { ConfigurationKeys.PropagateProcessTags, enabled ? "true" : "false" } }));
+        var tracer = TracerHelper.Create(settings);
+
+        var parentContext = new SpanContext(new TraceId(Upper: 0, Lower: 1), spanId: 2, (int)SamplingPriority.UserKeep, "ServiceName1", "origin1");
```
lucaspimentel: We should not create standalone spans like this without a `TraceContext` anymore. We expect all spans to belong to a trace and this can mess things up. If you see a test doing this, it's an old test that we haven't updated yet 😅

tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -0,0 +1,78 @@
...
+    private static readonly Lazy<string> LazySerializedTags = new(GetSerializedTags);
+
+    public static string SerializedTags
+    {
+        get => LazySerializedTags.Value;
+    }
+
+    private static SortedDictionary<string, string> LoadBaseTags()
+    {
+        var tags = new SortedDictionary<string, string>();
```
lucaspimentel: This could be `List<KeyValuePair<<string, string>>`, which does less working during inserts and allocates less data structures on the heap. Since we are adding the tags manually, we can ensure they 1) they are sorted and 2) they have unique keys, without the extra overhead of a `SortedDictionary`.

lucaspimentel: In fact, we can concatenate each key/value into a `StringBuilder` right here without creating any collection at all. Something like:

```csharp
var serializedTags = StringBuilderCache.Acquire();
var entrypointFullName = Assembly.GetEntryAssembly()?.EntryPoint?.DeclaringType?.FullName;

if (!StringUtil.IsNullOrEmpty(entrypointFullName))
{
    serializedTags.Append($"{EntrypointName}:{NormalizeTagValue(entrypointFullName)},");
}

serializedTags.Append($"{EntrypointBasedir}:{NormalizeTagValue(GetLastPathSegment(AppContext.BaseDirectory))},");

// workdir can be changed by the code, but we consider that capturing the value when this is called is good enough
serializedTags.Append($"{EntrypointWorkdir}:{NormalizeTagValue(GetLastPathSegment(Environment.CurrentDirectory))}"); // don't include last comma

return StringBuilderCache.GetStringAndRelease(serializedTags);
```

You'll have to add the tags in the right order if you want them sorted.

Note also the use of `StringBuilderCache` instead of `StringBuilder` (we reuse builders to reduce allocations), and `StringUtil.IsNullOrEmpty()` instead of `string.IsNullOrEmpty()` (older versions of .NET don't have the nullability annotations).

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
vandonr: idk, I guess the idea is that since we introduce it bits by bits and not as a single huge PR, it'll not work at 100% until fully done, so it's considered experimental, and at the end we switch the default value to true.
I took the flag value from the RFC.

tracer/src/Datadog.Trace/Agent/MessagePack/TraceChunkModel.cs
vandonr: Hmm yes, I didn't do that myself tbh, it's the IDE auto-formatting. Idk if we can edit the rules to match the situation you describe.

tracer/test/Datadog.Trace.Tests/Agent/MessagePack/SpanMessagePackFormatterTests.cs
vandonr: hmm, I took example on the other tests in that file...
would this be better ?
```suggestion
        var parentContext = new SpanContext(parent: null, new TraceContext(tracer), "ServiceName1");
```

tracer/src/Datadog.Trace/ProcessTags.cs
vandonr: hmm, yes fair. The SortedDict ensures that it's sorted without having to organize the code manually, and also ensures the tags are only present once. But this should be coverable with tests.
(it's something that's done only once in the lifetime of the app though, so Idk if we need to be that careful about performance.)

vandonr: not super fan of the fact that the calls to `NormalizeTagValue` have to be repeated for each value in your proposed code, this makes it easy to forget it, and actually hard to cover in tests. We can put everything in one method and replace the dict with a list, but I'd prefer keeping the normalization in a loop.

tracer/src/Datadog.Trace/Configuration/ConfigurationKeys.cs
lucaspimentel: > we introduce it bits by bits and not as a single huge PR, it'll not work at 100% until fully done

Ahh, that makes sense. Thanks!

tracer/test/Datadog.Trace.Tests/Agent/MessagePack/SpanMessagePackFormatterTests.cs
lucaspimentel: Something like this (ai-generated code snippet based on real usage patterns), where the `MockApi` let's you inspect the serialized spans:

```csharp
[Fact]
public async Task MyTest()
{
    // Setup: Custom settings + MockApi to capture serialized traces
    var mockApi = new MockApi();
    var settings = new TracerSettings(myConfigSource);
    var agentWriter = new AgentWriter(mockApi, statsAggregator: null, statsd: null, automaticFlush: false);
    await using var tracer = TracerHelper.Create(settings, agentWriter, sampler: null, scopeManager: null, statsd: null);

    // Use tracer to generate spans
    using (tracer.StartActiveInternal("root"))
    {
      // ... operations ...
    }

    // Flush and verify serialization
    await tracer.FlushAsync();
    var traceChunks = mockApi.Wait();
    var deserializedSpan = traceChunks.Single().Single();

    // Assert on serialized data
    deserializedSpan.GetTag("some-tag").Should().Be("expected-value");
}
```

Also, consider moving this test somewhere else, maybe to `tracer/test/Datadog.Trace.IntegrationTests/Tagging/ProcessTagsSerializationTests.cs` (next to `AASTagsTests.cs`). This file has general serialization tests or test for cross-cutting features like 128-bit trace ids, not for specific features like process tags. Just trying to keep things organized 😅

tracer/src/Datadog.Trace/ProcessTags.cs
lucaspimentel: That's fair. In this repository performance often has to take priority over other engineering principles like readability or maintainability, but this is only happening once per payload, so it's not as bad as something in the hot path (every request or every span).

Maybe we can find a middle-ground, where we list the sorted keys manually (to avoid the unnecessary `SortedDict` allocations), but still use a loop to concatenate them into a `StringBuilder`. Something like this?

```csharp
// this still allocates an array, which we can avoid by not using a loop, but it's much better than SortedDictionary
var tags = new KeyValuePair<string, string>[]
{
    // keep these sorted!
    // key1, value1
    // key2, value2
    // key3, value3
};

var serializedTags = StringBuilderCache.Acquire();

foreach (var tag in tags)
{
    if (!StringUtil.IsNullOrEmpty(tags.Value))
    {
        serializedTags.Append($"{tag.Key}:{NormalizeTagValue(tag.Value)},");
    }
}

serializedTags.Remove(serializedTags.Length - 1, length: 1); // remove last comma
return StringBuilderCache.GetStringAndRelease(serializedTags);
```

vandonr: it's not even once per payload, it's once a bit after init, because then the value is cached in the `Lazy<>` and never recomputed.

lucaspimentel: I still think `SortedDictionary` is overkill for this. And performance is important even in code that only runs once. Startup perf can even be as important as "hotpath" performance in short-lived apps like serverless functions.

tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -0,0 +1,71 @@
...
+        var tags = new List<KeyValuePair<string, string?>>
+        {
+            new(EntrypointBasedir, GetLastPathSegment(AppContext.BaseDirectory)),
+            new(EntrypointName, EntryAssemblyLocator.GetEntryAssembly()?.EntryPoint?.DeclaringType?.FullName),
+            // workdir can be changed by the code, but we consider that capturing the value when this is called is good enough
+            new(EntrypointWorkdir, GetLastPathSegment(Environment.CurrentDirectory))
+        };
+
+        // then normalize values and put all tags in a string
+        var serializedTags = new StringBuilder();
```
lucaspimentel: Use `StringBuilderCache` instead of creating a new `StringBuilder`. We reuse `StringBuilder` instances to reduce heap allocations. See the code snippets in my previous comments for example usage.

vandonr: ah yes I was pretty sure I had done this, sorry for making you repeat yourself

tracer/src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
```diff
@@ -603,6 +609,18 @@ private int WriteTags(ref byte[] bytes, int offset, in SpanModel model, ITagProc
                 }
             }
 
+            // Process tags will be sent only once per buffer/payload (one payload can contain many chunks from different traces)
+            if (model.IsFirstSpanInChunk && model.TraceChunk.IsFirstChunkInPayload && model.TraceChunk.ShouldPropagateProcessTags)
```
andrewlock, line 613: Are process tags supposed to be added to CI Visibility payloads? Because they have their own formatters if so

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -102,6 +102,11 @@ internal TracerSettings(IConfigurationSource? source, IConfigurationTelemetry te
                         string s => new HashSet<string>(s.Split([','], StringSplitOptions.RemoveEmptyEntries)),
                     };
 
+            PropagateProcessTags = config
+                                       .WithKeys(ConfigurationKeys.PropagateProcessTags)
+                                       .AsBool(false)
+                                || ExperimentalFeaturesEnabled.Contains(ConfigurationKeys.PropagateProcessTags);
```
andrewlock, line 108: Is this really correct? 🤔 If the customer explicitly sets `DD_EXPERIMENTAL_PROPAGATE_PROCESS_TAGS_ENABLED=false`, but they happen to have "all" for experimental features, then they have no way to disable the feature? Should it actually be that the _default_ value is controlled by that? 

```suggestion
                                       .AsBool(
                                       ExperimentalFeaturesEnabled.Contains(ConfigurationKeys.PropagateProcessTags);
```

tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    /// <summary>
+    /// From the full path of a directory, get the name of the leaf directory.
+    /// </summary>
+    private static string GetLastPathSegment(string directoryPath)
+    {
+        // Path.GetFileName returns an empty string if the path ends with a '/'.
+        // We could use Path.TrimEndingDirectorySeparator instead of the trim here, but it's not available on .NET Framework
+        return Path.GetFileName(directoryPath.TrimEnd('\\').TrimEnd('/'));
```
andrewlock, line 38: This is going to allocate multiple times, we can use ReadOnlySpan on netcore3.1+ to reduce the allocations

```csharp
#if NETCOREAPP3_1_OR_GREATER
        return Path.GetFileName(directoryPath.AsSpan().TrimEnd('\\').TrimEnd('/'));
#else
        return Path.GetFileName(directoryPath.TrimEnd('\\').TrimEnd('/'));
#endif
```

tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -0,0 +1,72 @@
...
+    private static string GetSerializedTags()
+    {
+        // ⚠️ make sure entries are added in alphabetical order of keys
+        var tags = new List<KeyValuePair<string, string?>>
+        {
+            new(EntrypointBasedir, GetLastPathSegment(AppContext.BaseDirectory)),
+            new(EntrypointName, EntryAssemblyLocator.GetEntryAssembly()?.EntryPoint?.DeclaringType?.FullName),
+            // workdir can be changed by the code, but we consider that capturing the value when this is called is good enough
+            new(EntrypointWorkdir, GetLastPathSegment(Environment.CurrentDirectory))
+        };
```
andrewlock, line 50: Collection expressions can be more efficient for the same code:

```csharp
List<KeyValuePair<string, string?>> tags =
[
    new(EntrypointBasedir, GetLastPathSegment(AppContext.BaseDirectory)),
    new(EntrypointName, EntryAssemblyLocator.GetEntryAssembly()?.EntryPoint?.DeclaringType?.FullName),
    // workdir can be changed by the code, but we consider that capturing the value when this is called is good enough
    new(EntrypointWorkdir, GetLastPathSegment(Environment.CurrentDirectory))
];
```

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
vandonr: I think this is never going to happen, but you are right

tracer/src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
vandonr: hmm 🤔  that's a good question, does CI visibility have a notion of service name too ? (if yes yes, if no no)

tracer/src/Datadog.Trace/ProcessTags.cs
vandonr: I'd say the same thing as as said to Lucas above, sick trick, but do we really want to start having different code path for different fwk in there, given that this code runs only once ?

tracer/src/Datadog.Trace/ProcessTags.cs
vandonr: wow I'd have guessed they resolved to exactly the same code underneath

andrewlock: It depends 😄 https://andrewlock.net/behind-the-scenes-of-collection-expressions-part-2-exploring-the-generated-code-list-and-fallback-cases/#optimizing-listt

tracer/src/Datadog.Trace/ProcessTags.cs
andrewlock: Yeah, that's fair. I would but meh 😄

tracer/src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
vandonr: confirmed that we don't need them on CI visibility

andrewlock: Ci Vis does have a notion of service name, yes, but I guess this isn't an issue 🤷‍♂️

PR #7656, opened 2025-10-14
[OTEL] Switches OTLP Tests To dd-apm-test-agent Instead of MockTracerAgent

link04:
## Summary of changes
Making use of the [dd-apm-test-agent](https://github.com/DataDog/dd-apm-test-agent) container to enable OTLP metrics support for GRPC and HTTP, then gets then as JSON the container for snapshot validation.

## Reason for change
To easily and more reliably test GRPC and ongoing OTLP support and validation in the .NET tracer without needing to build GRPC servers or change our agent for each new signal we plan to support. 

## Implementation details
Added the container to run for the integrations tests in the corresponding host name making use of either port `4317` or 4318`:

1. Pings `test/session/clear` to clear the session at the start of the tests
2. Pings `/test/session/metrics` endpoint for the generated metrics

## Test coverage
Updated the snapshots format but the same 4 should pass even for GRPC now that added to run for OTEL.

## Other details

[APMAPI-1678](https://datadoghq.atlassian.net/browse/APMAPI-1678
)

[APMAPI-1678]: https://datadoghq.atlassian.net/browse/APMAPI-1678?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docker-compose.windows.yml
+++ b/docker-compose.windows.yml
@@ -32,8 +32,8 @@ services:
     - ./artifacts/build_data/snapshots:c:/debug_snapshots
     ports:
     - "8126"
-    - "4317"
-    - "4318"
+    - "4317:4317"
+    - "4318:4318"
     environment:
     - SNAPSHOT_CI=1
       # api-security attrs are unfortunately ignored because gzip compression generates different bytes per platform windows/linux


zacharycmontoya (approved):
LGTM!

tracer/test/test-applications/integrations/Samples.OpenTelemetrySdk/Samples.OpenTelemetrySdk.csproj
```diff
@@ -3,7 +3,7 @@
     <!-- System.Runtime.CompilerServices.Unsafe doesn't support netcoreapp2.1-->
     <!-- https://andrewlock.net/stop-lying-about-netstandard-2-support/-->
     <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings>
-    <ApiVersion Condition="'$(ApiVersion)' == ''">1.12.0</ApiVersion>
+    <ApiVersion Condition="'$(ApiVersion)' == ''">1.13.1</ApiVersion>
```
zacharycmontoya, line 7: Now these tests need a Docker dependency in order to run

```suggestion
        <RequiresDockerDependency>true</RequiresDockerDependency>
        <ApiVersion Condition="'$(ApiVersion)' == ''">1.13.1</ApiVersion>
```

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/OpenTelemetrySdkTests.cs
```diff
@@ -213,10 +215,11 @@ public async Task IntegrationDisabled(string packageVersion)
         [SkippableTheory]
         [Trait("Category", "EndToEnd")]
         [Trait("RunOnWindows", "True")]
+        [Trait("RequiresDockerDependency", "true")]
```
zacharycmontoya, line 218: You will need to remove the `[Trait("RunOnWindows", "True")]` so that we do not run it on Windows CI. Theoretically we can run Windows containers in CI, but I don't know that we've done so yet so even if there's a Windows container version of dd-apm-test-agent, this could be unreliable

tracer/test/test-applications/integrations/Samples.OpenTelemetrySdk/Samples.OpenTelemetrySdk.csproj
zacharycmontoya: And then you'll need to re-run the `GeneratePackageVersions` target to update the matrix of sample apps with this

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/OpenTelemetrySdkTests.cs
zacharycmontoya, line 218: The tests also failed on Debian/Alpine so perhaps we don't have an arm64 image of the dd-apm-test-agent? If that's the case, we should indicate that we can't run this on arm

```suggestion
        [Trait("RequiresDockerDependency", "true")]
        [Trait("Category", "ArmUnsupported")]
```

bouwkast (approved):

bouwkast:
As we are seeing flake on the Windows tests related to the docker image I've made https://datadoghq.atlassian.net/browse/APMLP-559

@link04 has this https://datadoghq.atlassian.net/browse/APMAPI-1688 to get the tests re-enabled

PR #7657, opened 2025-10-15
[crashtracker/profiler] Notify profiler the application is crashing

gleocadie:
## Summary of changes

Nth attempt to prevent the profiler from collecting callstacks when the application is crashing.

## Reason for change

We have this issue for a while and this appeared in some crash reports lately. The profiler is trying to collect callstacks while the app is crashing. This can lead to a crash on a crash and appearance of zombie processes.
```
syscall (sysdeps/unix/sysv/linux/x86_64/syscall.S:38)
_write_validate (/project/obj/libunwind-prefix/src/libunwind/src/mi/Gaddress_validator.c:140)
access_mem (/project/obj/libunwind-prefix/src/libunwind/src/x86_64/Ginit.c:89)
dwarf_get (/project/obj/libunwind-prefix/src/libunwind/include/tdep-x86_64/libunwind_i.h:205)
unw_backtrace2 (/project/obj/libunwind-prefix/src/libunwind/src/mi/backtrace.c:111)
LinuxStackFramesCollector::CollectStackWithBacktrace2(void*) (/project/profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/LinuxStackFramesCollector.cpp:285)
LinuxStackFramesCollector::CollectStackSampleSignalHandler(int, siginfo_t*, void*) (/project/profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/LinuxStackFramesCollector.cpp)
ProfilerSignalManager::SignalHandler(int, siginfo_t*, void*) (/project/profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/ProfilerSignalManager.cpp:183)
__restore_rt
__GI___wait4 (sysdeps/unix/sysv/linux/wait4.c:30)
PROCCreateCrashDump(std::vector<char const*, std::allocator<char const*> >&, char*, int, bool) (/__w/1/s/src/coreclr/pal/src/thread/process.cpp:2545)
PROCCreateCrashDumpIfEnabled (/__w/1/s/src/coreclr/pal/src/thread/process.cpp)
PROCAbort (/__w/1/s/src/coreclr/pal/src/thread/process.cpp:2793)
PROCEndProcess(void*, unsigned int, int) (/__w/1/s/src/coreclr/pal/src/thread/process.cpp:1355)
UnwindManagedExceptionPass1(PAL_SEHException&, _CONTEXT*) (/__w/1/s/src/coreclr/vm/exceptionhandling.cpp)
DispatchManagedException(PAL_SEHException&, bool) (/__w/1/s/src/coreclr/vm/exceptionhandling.cpp)
IL_Throw(Object*) (/__w/1/s/src/coreclr/vm/jithelpers.cpp)
```

When the application is crashing, the .NET runtime calls `fork()`. Calling this function leads to the creation of a new process (child process). This process will have a copy of the parent address space.
Then the .NET runtime, in the child process, calls `execve()` to run `createdump`. We wrapped `execve()` in order to call our own tool which will create a crash report and send it to our backend.
In our implementation of `execve()`, we set a flag to notify the profiler to not collect callstack because the application is crashing.

The current implementation does not work because we use an `atomic_int` which will be copied and when the child process set flag, it will be its own version of the flag. The parent process won't be able to see it changed. We need a shared memory so the parent and child processes can communicate.

## Implementation details

- create a shared memory region (using mmap)

## Test coverage

We should not see issue caused by the profiler in our CI when crashes happen
## Other details

@kevingosse , I finally found it :)

### DIFF
--- a/profiler/src/ProfilerEngine/Datadog.Linux.ApiWrapper/functions_to_wrap.c
+++ b/profiler/src/ProfilerEngine/Datadog.Linux.ApiWrapper/functions_to_wrap.c
@@ -1,14 +1,15 @@
 #define _GNU_SOURCE
 #include <dlfcn.h>
 #include <link.h>
+#include <pthread.h>
 #include <signal.h>
 #include <stddef.h>
 #include <stdio.h>
 #include <string.h>
-#include <stdlib.h>
-#include <unistd.h>
 #include <stdatomic.h>
-#include <pthread.h>
+#include <stdlib.h>
+#include <sys/mman.h>
+#include <unistd.h>
 
 #include "common.h"
 
@@ -58,13 +59,28 @@ enum FUNCTION_ID
 // counters: one byte per function
 __thread unsigned long long functions_entered_counter = 0;
 
+// This variable is used to indicate to the profiler that the application is crashing
+// and it should not collect samples while the app is crashing.
+// At crash time, the .NET runtime calls fork() to create a child process which will
+// be in charge of collecting a crash dump (by calling execve()) while the parent is waiting
+// for the child to finish.
+// By calling fork(), the child process and the parent process will have their own address space,
+// which means that the child process won't be able to modify the parent process's variables.
+// We need a way to enable communication between the child and parent processes.
+// This is done by creating a shared memory region and use it as a flag to indicate that
+// the application is crashing.
+// This variable will be a pointer to that shared memory region.
 __attribute__((visibility("hidden")))
-atomic_int is_app_crashing = 0;
+int* is_app_crashing = NULL;
 
 // this function is called by the profiler
 unsigned long long dd_inside_wrapped_functions()
 {
-    return functions_entered_counter + is_app_crashing;
+    int app_is_crashing = 0;
+    if (is_app_crashing != NULL) {
+        app_is_crashing = *is_app_crashing;
+    }
+    return functions_entered_counter + app_is_crashing;
 }
 
 #if defined(__aarch64__)
@@ -467,7 +483,9 @@ int execve(const char* pathname, char* const argv[], char* const envp[])
         return __real_execve(pathname, argv, envp);
     }
 
-    is_app_crashing = 1;
+    if (is_app_crashing != NULL) {
+        *is_app_crashing = 1;
+    }
     // Execute the alternative crash handler, and prepend "createdump" to the arguments
 
     // Count the number of arguments (the list ends with a null pointer)
@@ -671,6 +689,13 @@ static void init()
     __real_pthread_setattr_default_np = __dd_dlsym(RTLD_NEXT, "pthread_setattr_default_np");
     //__real_fork = __dd_dlsym(RTLD_NEXT, "fork");
 #endif
+    // if we failed at allocating memory for the shared variable
+    // the parent process won't be notified that the app is crashing.
+    is_app_crashing = mmap(NULL, sizeof(int), PROT_READ | PROT_WRITE,
+                           MAP_SHARED | MAP_ANONYMOUS, -1, 0);
+    if (is_app_crashing != MAP_FAILED) {
+        *is_app_crashing = 0; // Initialize flag
+    }
 }
 
 static void check_init()


andrewlock (approved):
ooooh, nice one!

profiler/src/ProfilerEngine/Datadog.Linux.ApiWrapper/functions_to_wrap.c
```diff
@@ -58,13 +59,28 @@ enum FUNCTION_ID
...
+// for the child to finish.
+// By calling fork(), the child process and the parent process will have their own address space,
+// which means that the child process won't be able to modify the parent process's variables.
+// We need a way to enable communication between the child and parent processes.
+// This is done by creating a shared memory region and use it as a flag to indicate that
+// the application is crashing.
+// This variable will be a pointer to that shared memory region.
 __attribute__((visibility("hidden")))
-atomic_int is_app_crashing = 0;
+int* is_app_crashing = NULL;
```
chrisnas, line 74: I would rename the variable to **p_is_app_crashing** to make it more explicit that it is a pointer

profiler/src/ProfilerEngine/Datadog.Linux.ApiWrapper/functions_to_wrap.c
```diff
@@ -671,6 +689,13 @@ static void init()
     __real_pthread_setattr_default_np = __dd_dlsym(RTLD_NEXT, "pthread_setattr_default_np");
     //__real_fork = __dd_dlsym(RTLD_NEXT, "fork");
 #endif
+    // if we failed at allocating memory for the shared variable
+    // the parent process won't be notified that the app is crashing.
+    is_app_crashing = mmap(NULL, sizeof(int), PROT_READ | PROT_WRITE,
+                           MAP_SHARED | MAP_ANONYMOUS, -1, 0);
+    if (is_app_crashing != MAP_FAILED) {
```
chrisnas, line 696: Should we at least log a message because if it fails (even if it should not happen), we're doomed?

profiler/src/ProfilerEngine/Datadog.Linux.ApiWrapper/functions_to_wrap.c
```diff
@@ -671,6 +689,13 @@ static void init()
     __real_pthread_setattr_default_np = __dd_dlsym(RTLD_NEXT, "pthread_setattr_default_np");
     //__real_fork = __dd_dlsym(RTLD_NEXT, "fork");
 #endif
+    // if we failed at allocating memory for the shared variable
+    // the parent process won't be notified that the app is crashing.
+    is_app_crashing = mmap(NULL, sizeof(int), PROT_READ | PROT_WRITE,
```
chrisnas, line 694: I don't know for linux but on Windows, objects need to be named in order to be shared between processes.
Just to be sure that I understand how this is supposed to work and why it does not matter for Linux:
1. the parent code calls this init() function
2. the forked child does not but inherits the value of this variable from its parent
3. the forked child changes the value pointed to so that the parent could also see that change when calling dd_inside_wrapped_functions()

chrisnas (approved):
LGTM with maybe a renaming

profiler/src/ProfilerEngine/Datadog.Linux.ApiWrapper/functions_to_wrap.c
gleocadie: nop, at this point nothing is ready. That said, there is no logging from this library.

profiler/src/ProfilerEngine/Datadog.Linux.ApiWrapper/functions_to_wrap.c
gleocadie: as defined in the call to mmap, it's an anonymous area.
> Just to be sure that I understand how this is supposed to work and why it does not matter for Linux:

I do not understand what you mean by `it does not matter for Linux`. What does not matter for Linux ?

kevingosse:
Good job but also wtf

PR #7663, opened 2025-10-16
[Test Optimization] MsTest support updates

tonyredondo:
## Summary of changes

  - Extend the MSTest package matrix (generator definitions, props, and generated test enumerations) to track new 4.x releases and keep the integration samples in sync.
  - Update the MSTest V2 calltarget instrumentation to support the async ExecuteAsync path introduced in MSTest 4.x, including new duck-typed interfaces, retry-state handling, and regenerated instrumentation metadata/calltarget tables.
  - Refresh the sample MSTest test projects to default to 4.x, add async test method attribute overrides, and enable additional CI diagnostics for the corresponding integration tests.

## Reason for change

  - MSTest 4.x introduced asynchronous execution APIs that bypassed our existing instrumentation and were not yet part of the supported version set; without these updates we miss spans/metadata for customers adopting the new framework.

## Implementation details

  - Bumped the MSTest package version ranges (up to <5.0.0) and added explicit 4.x entries across generator inputs, props, and generated C# enumerations used by the integration test harness.
  - Introduced new duck-typing interfaces for MSTest 4.x, split the calltarget implementation to share logic between Execute and ExecuteAsync, and regenerated instrumentation/calltarget metadata files to cover the new methods.
  - Adjusted the sample MSTest projects and tests (conditional compilation with MSTEST_ASYNC, async overrides, environment configuration) so the integration suite exercises the new async pipeline.

## Test coverage

  - Current tests now runs on the new versions.

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Testing/MsTestV2/UnitTestDiscovererSendTestCasesIntegration.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Testing/MsTestV2/UnitTestDiscovererSendTestCasesIntegration.cs
@@ -6,7 +6,6 @@
 
 using System.Collections;
 using System.ComponentModel;
-using System.Linq;
 using Datadog.Trace.ClrProfiler.CallTarget;
 
 namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Testing.MsTestV2;
@@ -29,8 +28,14 @@ public class UnitTestDiscovererSendTestCasesIntegration
 {
     internal static CallTargetState OnMethodBegin<TTarget, TDiscoverySink, TDiscoveryContext, TLogger>(TTarget instance, ref string? source, ref IEnumerable? testElements, ref TDiscoverySink? discoverySink, ref TDiscoveryContext? discoveryContext, ref TLogger? logger)
     {
-        if (testElements?.Cast<object>().Count() is { } count)
+        if (testElements is not null)
         {
+            var count = 0;
+            foreach (var ele in testElements)
+            {
+                count++;
+            }
+
             MsTestIntegration.AddTotalTestCases(count);
         }
 
@@ -58,8 +63,14 @@ public class UnitTestDiscovererSendTestCasesIntegrationV4
 {
     internal static CallTargetState OnMethodBegin<TTarget, TDiscoverySink, TDiscoveryContext, TLogger>(TTarget instance, ref IEnumerable? testElements, ref TDiscoverySink? discoverySink, ref TDiscoveryContext? discoveryContext, ref TLogger? logger)
     {
-        if (testElements?.Cast<object>().Count() is { } count)
+        if (testElements is not null)
         {
+            var count = 0;
+            foreach (var ele in testElements)
+            {
+                count++;
+            }
+
             MsTestIntegration.AddTotalTestCases(count);
         }
 


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Testing/MsTestV2/TestMethodAttributeExecuteIntegration.cs
```diff
@@ -41,27 +43,97 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Testing.MsTestV2;
...
 
-        return new CallTargetState(null, new TestRunnerState(testMethod, MsTestIntegration.OnMethodBegin(testMethod, testMethod.Type, isRetry: false)));
+        var testMethodProxy = (ITestMethod?)testMethod.DuckAs<ITestMethodV4>() ?? testMethod.DuckAs<ITestMethodV3>();
+        if (testMethodProxy is null)
+        {
+            DuckTypeException.Throw("Failed to duck type the test method instance to ITestMethodV3 or ITestMethodV4.");
+        }
+
+        var testRunnerState = new TestRunnerState(testMethodProxy, MsTestIntegration.OnMethodBegin(testMethodProxy, testMethodProxy.Type, isRetry: false));
+        return new CallTargetState(Tracer.Instance.InternalActiveScope, testRunnerState);
```
andrewlock, line 128: You previously were returning `null` as the scope here, and you already grabbed it and confirmed it was _not_ a test scope previously, so this seems like a behaviour change. Is it intentional? 🤔

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Testing/MsTestV2/UnitTestDiscovererSendTestCasesIntegration.cs
```diff
@@ -37,3 +37,32 @@ internal static CallTargetState OnMethodBegin<TTarget, TDiscoverySink, TDiscover
...
+#pragma warning disable SA1402
+public class UnitTestDiscovererSendTestCasesIntegrationV4
+#pragma warning restore SA1402
+{
+    internal static CallTargetState OnMethodBegin<TTarget, TDiscoverySink, TDiscoveryContext, TLogger>(TTarget instance, ref IEnumerable? testElements, ref TDiscoverySink? discoverySink, ref TDiscoveryContext? discoveryContext, ref TLogger? logger)
+    {
+        if (testElements?.Cast<object>().Count() is { } count)
+        {
+            MsTestIntegration.AddTotalTestCases(count);
+        }
```
andrewlock: Cast feels a bit janky, would this do? Do you prefer it?
```suggestion
        if (testElements is not null)
        {
            var cnt = 0;
            foreach (var ele in testElements)
            {
                cnt++;
            }
            
            MsTestIntegration.AddTotalTestCases(count);
        }
```

tracer/test/test-applications/integrations/Samples.MSTestTests/Samples.MSTestTests.csproj
```diff
@@ -3,25 +3,31 @@
...
     <!-- Required to build multiple projects with the same Configuration|Platform -->
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <GenerateTargetFrameworkAttribute>false</GenerateTargetFrameworkAttribute>
     <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
   </PropertyGroup>
-
+  
+  <PropertyGroup Condition="$(ApiVersion.StartsWith('4.'))">
+    <DefineConstants>$(DefineConstants);MSTEST_ASYNC</DefineConstants>
+    <SuppressTfmSupportBuildErrors>true</SuppressTfmSupportBuildErrors>
```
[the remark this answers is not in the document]
andrewlock: Lies

tracer/test/test-applications/integrations/Samples.MSTestTests/TestSuite.cs
```diff
@@ -130,27 +131,57 @@ public void CustomMultipleResultsTestMethodAttributeTest()
...
     }
+#endif
 }
 
 public class MyCustomRenameTestMethodAttribute : TestMethodAttribute
 {
+#if MSTEST_ASYNC
+    public override async Task<TestResult[]> ExecuteAsync(ITestMethod testMethod)
+    {
+        var results = await base.ExecuteAsync(testMethod).ConfigureAwait(false);
```
andrewlock, line 158: Silly question - I assume MSTest doesn't have a sync context right (e.g. like XUnit _does_)

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Testing/MsTestV2/TestMethodAttributeExecuteIntegration.cs
tonyredondo: Totally, this code was in a sync integration before with the OnMethodEnd call closing the span and restoring the scope. Now due to the latest change we had to move this as an async counterpart meaning that we now have `OnAsyncMethodEnd` the issue with this is that the OnMethodBegin change the active scope with the new Span, but we need a way to restore it synchronously (because the async version contains a copy on write version of the scope, any restoration in that execution context doesn't affect the previous execution context), that's also the reason we call `IntegrationOptions.RestoreScopeFromAsyncExecution(in state);` in the Sync version. We don't need that line in the Async version because the CallTarget machinery is aware of this issue and that restoration is done automatically for all async integrations.

tracer/test/test-applications/integrations/Samples.MSTestTests/TestSuite.cs
tonyredondo: Not yet, I'm waiting for 4.0.2 for them to do that change 🤣

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Testing/MsTestV2/UnitTestDiscovererSendTestCasesIntegration.cs
tonyredondo: done in [changes from review](https://github.com/DataDog/dd-trace-dotnet/pull/7663/commits/dd5f0ab6c38571ef9af8e6990faa9bc2ff4cea9d)

PR #7666, opened 2025-10-16
[OTEL] Vendoring OtlpGrpcExportClient And Enabling OTLP Metrics gRPC Tests

link04:
## Summary of changes

Added gRPC protocol support to the OTLP metrics exporter by vendoring OpenTelemetry's gRPC transport client. The implementation now supports `grpc` which also was updated to be the default protocol value.

## Reason for change

Customers need gRPC support for OTLP metrics export to maintain compatibility with OpenTelemetry ecosystem standards and their existing infrastructure that may require or prefer gRPC transport.

## Implementation details

**Vendored dependencies:**
- Vendored `OpenTelemetry.Exporter.OpenTelemetryProtocol` v1.13.1(lastest ATM with included fixes) gRPC transport client (`OtlpGrpcExportClient` and related helpers)
- Created `OpenTelemetryStubs.cs` with minimal stub implementations to avoid vendoring the entire OpenTelemetry SDK:
  - `Guard` and `UriExtensions` from `OpenTelemetry.Internal` (Would need to discard deletion after usng vendoring command)
  - `OtlpExporterOptions` (minimal configuration class)
  - `OtlpExportProtocol` enum (kept in stub due to namespace issues with the original being in parent namespace `OpenTelemetry.Exporter`)
  - `OpenTelemetryProtocolExporterEventSource` (stub EventSource for logging)

**Core changes:**
- Updated `OtlpExporter.cs` to instantiate and use the vendored `OtlpGrpcExportClient` when protocol is `Grpc`
- Added 5-byte gRPC message frame prefix (compression flag + message length) to protobuf payloads for gRPC transport
- Modified `OtlpMetricsSerializer.cs` to optionally reserve bytes at the start of the buffer for the gRPC frame header

**Vendoring infrastructure:**
- Added `AddOpenTelemetryUsings` transform to inject common `using` directives needed by vendored OTel files
- Configured exclusions to vendor only the gRPC transport client, not the full OTLP exporter

## Test coverage

- Updated `OpenTelemetrySdkTests.SubmitsOtlpMetrics` to test both `http/protobuf` and `grpc` protocols
- gRPC tests use the `dd-apm-test-agent` container
- All tests validate metrics are correctly exported and received by the respective agents

## DLL File Size Difference

After building both `master` branch and this branch locally then comparing file size I got:
|Target Framework | Master (bytes) | Feature (bytes) | Difference (bytes) | Difference (%) | Impact |
|------------------|----------------|-----------------|-------------------|----------------|--------|
| net6.0 | 8,252,416 | 8,270,336 | **+17,920** | +0.22% | +17.5 KB |
| netcoreapp3.1 | 8,173,056 | 8,190,976 | **+17,920** | +0.22% | +17.5 KB |

## Other Details
[APMAPI-1679](https://datadoghq.atlassian.net/browse/APMAPI-1679)

[APMAPI-1679]: https://datadoghq.atlassian.net/browse/APMAPI-1679?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tracer/src/Datadog.Trace/OpenTelemetry/Metrics/OtlpExporter.cs
+++ b/tracer/src/Datadog.Trace/OpenTelemetry/Metrics/OtlpExporter.cs
@@ -190,20 +190,20 @@ namespace Datadog.Trace.OpenTelemetry.Metrics
         {
             try
             {
-                // Serialize metrics to protobuf format
-                // For gRPC, reserve 5 bytes at the start for the frame header (added later)
-                // For HTTP, start at position 0
-                // Note: JSON serialization is not yet implemented, so we always use protobuf ATM
-                var startPosition = _protocol == Configuration.OtlpProtocol.Grpc ? 5 : 0;
-                var otlpPayload = _serializer.SerializeMetrics(metrics, startPosition);
-
-                return _protocol switch
+                // OTEL-style: serialize per protocol so we never leak the gRPC 5-byte reservation into HTTP.
+                var task = _protocol switch
                 {
-                    Configuration.OtlpProtocol.HttpProtobuf => await SendHttpProtobufRequest(otlpPayload).ConfigureAwait(false),
-                    Configuration.OtlpProtocol.HttpJson => await SendHttpJsonRequest(otlpPayload).ConfigureAwait(false),
-                    Configuration.OtlpProtocol.Grpc => await SendGrpcRequest(otlpPayload).ConfigureAwait(false),
-                    _ => await SendHttpProtobufRequest(otlpPayload).ConfigureAwait(false)
+                    Configuration.OtlpProtocol.Grpc =>
+                        SendGrpcRequest(_serializer.SerializeMetrics(metrics, startPosition: 5)),
+                    Configuration.OtlpProtocol.HttpProtobuf =>
+                        SendHttpProtobufRequest(_serializer.SerializeMetrics(metrics, startPosition: 0)),
+                    Configuration.OtlpProtocol.HttpJson =>
+                        SendHttpJsonRequest(_serializer.SerializeMetrics(metrics, startPosition: 0)),
+                    _ =>
+                        SendHttpProtobufRequest(_serializer.SerializeMetrics(metrics, startPosition: 0))
                 };
+
+                return await task.ConfigureAwait(false);
             }
             catch (Exception ex)
             {
@@ -246,7 +246,27 @@ namespace Datadog.Trace.OpenTelemetry.Metrics
             if (_grpcClient is null)
             {
                 Log.Warning("GRPC selected but gRPC client is not initialized; falling back to HTTP/protobuf.");
-                return await SendHttpProtobufRequest(otlpPayload).ConfigureAwait(false);
+                // otlpPayload was serialized with startPosition=5 => slice off the 5-byte reservation.
+                return await SendWithRetry(
+                        otlpPayload,
+                        (p, endpoint, headers) =>
+                        {
+                            var content = new ByteArrayContent(p, 5, p.Length - 5);
+                            content.Headers.ContentType = new MediaTypeHeaderValue("application/x-protobuf");
+
+                            var httpRequest = new HttpRequestMessage(HttpMethod.Post, endpoint)
+                            {
+                                Content = content
+                            };
+
+                            foreach (var header in headers)
+                            {
+                                httpRequest.Headers.TryAddWithoutValidation(header.Key, header.Value);
+                            }
+
+                            return httpRequest;
+                        })
+                    .ConfigureAwait(false);
             }
 
             try
--- a/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/OpenTelemetrySdkTests.cs
+++ b/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/OpenTelemetrySdkTests.cs
@@ -214,6 +214,7 @@ namespace Datadog.Trace.ClrProfiler.IntegrationTests
 
 #if NET6_0_OR_GREATER
         [SkippableTheory]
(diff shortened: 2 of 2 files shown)


tracer/src/Datadog.Trace/Vendors/OpenTelemetry.Exporter.OpenTelemetryProtocol/OpenTelemetryStubs.cs
zacharycmontoya, line 1: Since this is hand-authored code, can you move this to a different directory? Perhaps inside the folder `Datadog.Trace/OpenTelemetry`? Or we could even split up the files for better discoverability by creating individual `.cs` files per class and placing them in a `Datadog.Trace/OpenTelemetry/Stubs` directory

tracer/src/Datadog.Trace/OpenTelemetry/Metrics/OtlpExporter.cs
```diff
@@ -226,8 +243,38 @@ static HttpRequestMessage CreateHttpProtobufRequest(byte[] payload, Uri endpoint
...
+                // bytes 5+: the actual protobuf payload (already serialized at position 5)
+                otlpPayload[0] = 0; // No compression
+
+                // Write message length in big-endian format (payload length - 5 header bytes)
+                var dataLength = otlpPayload.Length - 5;
+                System.Buffers.Binary.BinaryPrimitives.WriteUInt32BigEndian(
+                    new System.Span<byte>(otlpPayload, 1, 4),
+                    (uint)dataLength);
+
+                var resp = await Task.Run(() => _grpcClient.SendExportRequest(
```
zacharycmontoya, line 286: Note for future: We should consider making this entire pipeline synchronous, especially since we're going to rely on this vendored client code that wants to do it synchronously. That will make it so we don't have to allocate to create closures, like we're doing here ^

zacharycmontoya (approved):
LGTM with some small comments. ~Can you also document in the PR description the size difference of `Datadog.Trace.dll` before/after? I guess that it isn't much but it would be good for reviewers to be aware of.~ Thanks!

EDIT: You already included the size, nice!!

tracer/dependabot/Datadog.Dependabot.Vendors.csproj
```diff
@@ -27,11 +28,8 @@
     <!-- https://www.nuget.org/packages/dnlib/3.4.0 -->
     <PackageReference Include="dnlib" Version="3.4.0" />
 
-    <!-- https://www.nuget.org/packages/Datadog.Sketches/1.0.0 -->
-    <PackageReference Include="Datadog.Sketches" Version="1.0.0" />
-
-    <!-- https://www.nuget.org/packages/IndieSystem.Text.RegularExpressions/0.6 -->
-    <PackageReference Include="IndieSystem.Text.RegularExpressions" Version="0.6" />
+    <!-- https://www.nuget.org/packages/Datadog.Sketches/main -->
+    <PackageReference Include="Datadog.Sketches" Version="main" />
```
bouwkast, line 32: `Version="main"`

Okay I think this one is on me
I don't any action here, but I may have messed up the code that generates the value for this specific instance

bouwkast: I honestly can't remember what I did here but I think we can ignore this

tracer/src/Datadog.Trace.Trimming/build/Datadog.Trace.Trimming.xml
```diff
@@ -911,6 +916,7 @@
       <type fullname="System.IO.BinaryReader" />
       <type fullname="System.IO.BinaryWriter" />
       <type fullname="System.IO.EndOfStreamException" />
+      <type fullname="System.IO.InvalidDataException" />
```
bouwkast, line 919: I guess I don't know if we care that this is here twice? Probably fine, other instance is from `System.Runtime` this one is `System.Runtime.Extensions`

docker-compose.yml
```diff
@@ -445,6 +445,7 @@ services:
       - COUCHBASE_HOST=couchbase
       - COUCHBASE_PORT=8091
       - ASB_CONNECTION_STRING=Endpoint=sb://azureservicebus-emulator:5672;SharedAccessKeyName=RootManageSharedAccessKey;SharedAccessKey=SAS_KEY_VALUE;UseDevelopmentEmulator=true;
+      - TEST_AGENT_HOST=test-agent
```
bouwkast, line 448: Is this what caused the snapshots to change?

bouwkast (approved):
Nice thanks!

Only question is why the snapshots changed, which I'm thinking is due to the different test agent?

docker-compose.yml
link04: Yesss instead of getting them from the Mock Agent, the test Agent returns JSON instead so they look different but same amount of metrics with same data within, we used the changes from the version below:

https://github.com/DataDog/dd-apm-test-agent/releases/tag/v1.32.0

tracer/dependabot/Datadog.Dependabot.Vendors.csproj
link04: Yeah I wasn't sure why but since it was the command that generated it I assumed it was good.

tracer/src/Datadog.Trace/Vendors/OpenTelemetry.Exporter.OpenTelemetryProtocol/OpenTelemetryStubs.cs
link04: Yeah I was also feeling about having this file crated by us and with the warnings, thank you for the idea!

PR #7668, opened 2025-10-17
[DO NOT MERGE] workload selection

dmehala:
Based on: https://github.com/DataDog/dd-trace-dotnet/pull/7183

[INPLAT-753]

[INPLAT-753]: https://datadoghq.atlassian.net/browse/INPLAT-753?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/shared/src/Datadog.Trace.ClrProfiler.Native/util.h
+++ b/shared/src/Datadog.Trace.ClrProfiler.Native/util.h
@@ -46,7 +46,7 @@ static fs::path GetPoliciesPath()
         program_data_path = WStr(R"(C:\ProgramData)");
     }
 
-    return program_data_path / "Datadog" / "workload_selection.fb";
+    return program_data_path / "Datadog" / "user-wls-policy.bin";
 }
 #endif
 


shared/src/Datadog.Trace.ClrProfiler.Native/util.h
```diff
@@ -26,13 +29,26 @@ static fs::path GetCurrentModuleFolderPath()
...
+static fs::path GetPoliciesPath()
+{
+    fs::path program_data_path = shared::GetEnvironmentValue(WStr("PROGRAMDATA"));
+
+    if (program_data_path.empty())
+    {
+        program_data_path = WStr(R"(C:\ProgramData)");
+    }
+
+    return program_data_path / "Datadog" / "protected" / "workload_selection.fb";
```
dmehala: seems like IIS can't access due to permission. @bmermet any idea how we could fix the permission issue or should we change the location of this file?

bmermet: How are the permissions set currently?
I think we can set the same permissions than we set for APM libraries, IIS is able to load the .NET tracer with it: https://github.com/DataDog/datadog-agent/blob/main/pkg/fleet/installer/paths/installer_paths_windows.go#L540

shared/src/Datadog.Trace.ClrProfiler.Native/workload_selection_impl.cpp
```diff
@@ -0,0 +1,147 @@
...
+            return "";
+        }
+
+        auto dll_path = fs::path(::shared::ToString(tokenized_command_line[1]));
+        return dll_path.filename().string();
+    }
+
+} // namespace
+
+bool wls()
```
andrewlock: nit: can we expand this acronym please, e.g. something like `is_workload_allowed()`

shared/src/Datadog.Trace.ClrProfiler.Native/workload_selection_impl.cpp
```diff
@@ -0,0 +1,147 @@
...
+{
+    const auto process_name = ::shared::ToString(::shared::GetCurrentProcessName());
```
andrewlock: nit: we read this already in cor_profiler, we can probably just pass it in?

shared/src/Datadog.Trace.ClrProfiler.Native/util.h
```diff
@@ -78,12 +94,44 @@ static fs::path GetConfigurationFilePath()
...
+{
+    if (const auto& app_pool_id = ::shared::GetEnvironmentValue(environment::azure_app_services_app_pool_id);
+        !app_pool_id.empty())
+    {
+        return app_pool_id;
+    }
+
+    // Try to infer the Application Pool from the command line. w3wp.exe (IIS Worker Process)
+    // can be started with an Application Pool by using `-ap` argument.
+    const auto [_, argv] = ::shared::GetCurrentProcessCommandLine();
```
andrewlock, line 119: We already grab this in cor_profiler, maybe we can just pass it in?

shared/src/Datadog.Trace.ClrProfiler.Native/workload_selection_impl.cpp
```diff
@@ -0,0 +1,147 @@
...
+        application_pool = std::move(::shared::ToString(*maybe_application_pool));
+    }
+
+    const auto dotnet_dll = GetDotnetDll();
+
+    plcs_eval_ctx_init();
+    plcs_eval_ctx_set_str_eval_param(PLCS_STR_EVAL_RUNTIME_LANGUAGE, "dotnet");
+    plcs_eval_ctx_set_str_eval_param(PLCS_STR_EVAL_PROCESS_EXE, process_name.c_str());
+    plcs_eval_ctx_set_str_eval_param(PLCS_STR_EVAL_IIS_APPLICATION_POOL, application_pool.c_str());
+    plcs_eval_ctx_set_str_eval_param(PLCS_STR_EVAL_RUNTIME_ENTRY_POINT_FILE, dotnet_dll.c_str());
```
andrewlock, line 120: "Entry point file" is not necessarily correct for IIS, just an FYI 🤔 

Also this will miss cases like `dotnet exec /some/dll`... As long as this doesn't _need_ to be exhaustive that's ok 🤷‍♂️

shared/src/Datadog.Trace.ClrProfiler.Native/workload_selection_impl.cpp
```diff
@@ -0,0 +1,147 @@
...
+    plcs_eval_ctx_register_action(onInjectionDeny, PLCS_ACTION_INJECT_DENY);
+    plcs_eval_ctx_register_action(onInjectionAllow, PLCS_ACTION_INJECT_ALLOW);
+
+    injection_status = InjectionStatus::UNKNOWN;
+    auto maybe_error = evaluatePolicies();
+    if (maybe_error)
+    {
+        Log::Error(__func__, ": An error occured while evaluating workload selection policies (errno: ", *maybe_error,
+                   ")");
+        return false;
```
dmehala:
```suggestion
        return true;
```

shared/src/Datadog.Trace.ClrProfiler.Native/workload_selection_impl.cpp
dmehala:
```suggestion
bool is_workload_allowed()
```

shared/src/Datadog.Trace.ClrProfiler.Native/workload_selection_impl.h
```diff
@@ -0,0 +1,15 @@
...
+{
+#include <dd/policies/eval_ctx.h>
+#include <dd/policies/policies.h>
+}
+
+namespace datadog::shared::nativeloader
+{
+
+/// TBD
+bool wls();
```
dmehala:
```suggestion
bool is_workload_allowed();
```

shared/src/Datadog.Trace.ClrProfiler.Native/workload_selection_noop.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+namespace datadog::shared::nativeloader
+{
+
+inline bool wls()
```
dmehala:
```suggestion
inline bool is_workload_allowed()
```

shared/src/Datadog.Trace.ClrProfiler.Native/cor_profiler.cpp
```diff
@@ -316,6 +317,15 @@ namespace datadog::shared::nativeloader
             return E_FAIL;
         }
 
+        if (IsSingleStepInstrumentation())
+        {
+            Log::Info("CorProfiler::Initialize: Evaluating workload selection.");
+            if (wls() == false)
```
dmehala:
```suggestion
            if (is_workload_allowed() == false)
```

shared/src/Datadog.Trace.ClrProfiler.Native/workload_selection_impl.cpp
dmehala: addressed in 4f895fd81b7c5b78d2638e72bc2b72f7c090ea56

shared/src/Datadog.Trace.ClrProfiler.Native/workload_selection_impl.cpp
dmehala: addressed in 813b52d67546743034aa3c60f731503d9a44291e

shared/src/Datadog.Trace.ClrProfiler.Native/util.h
dmehala: addressed in 813b52d67546743034aa3c60f731503d9a44291e

shared/src/Datadog.Trace.ClrProfiler.Native/util.h
```diff
@@ -26,13 +29,26 @@ static fs::path GetCurrentModuleFolderPath()
...
-    if (dladdr((void*)GetCurrentModuleFolderPath, &info))
+    if (dladdr((void*) GetCurrentModuleFolderPath, &info))
     {
         return fs::path(info.dli_fname).remove_filename();
     }
 #endif
     return {};
 }
+#ifdef _WIN32
+static fs::path GetPoliciesPath()
```
andrewlock, line 40: Just a thought - I'd suggest that we only keep this calculation as a "fallback" path, and instead that we set this path via an env variable in fleet installer? That way if, for whatever reason, we need to move where the agent puts this, we're still backwards compatible? WDYT @bmermet ?

shared/test/Datadog.Trace.ClrProfiler.Native.Tests/workload_selection_test.cpp
andrewlock, line 1: These are cool, but I'd love if we had a couple of integration tests too, to test the end-to-end. That wouldn't be too hard to do if we make the location of the flat buffers file configurable via an env var. We do a bunch of simple "did this instrument, did it write a log file" tests here: https://github.com/DataDog/dd-trace-dotnet/blob/b59dea4bdabe3869af50745c23cd2b7f61b1de86/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/InstrumentationTests.cs, for example. We could just add a couple of basic ones there too? Happy to help with that if you need it - just let me know 🙂

shared/src/Datadog.Trace.ClrProfiler.Native/util.h
bmermet: How will it help maintaining backward compatibility?  Since,`FleetInstaller` and the native profiler are shipped together in the OCI, it seems to me it doesn't make a big difference if the path is computed in either of them.

shared/src/Datadog.Trace.ClrProfiler.Native/util.h
```diff
@@ -26,13 +29,26 @@ static fs::path GetCurrentModuleFolderPath()
...
+{
+    fs::path program_data_path = shared::GetEnvironmentValue(WStr("PROGRAMDATA"));
+
+    if (program_data_path.empty())
+    {
+        program_data_path = WStr(R"(C:\ProgramData)");
+    }
+
+    return program_data_path / "Datadog" / "workload_selection.fb";
```
bmermet: To be consistent with what's done on linux, I think this should be named: `user-wls-policy.bin`

shared/src/Datadog.Trace.ClrProfiler.Native/util.h
andrewlock: My thinking was that this would be the fleet installer (not the .NET `FleetInstaller.exe` 😅 ) that sets the path. Because the location of the flatbuffers file is essentially managed completely independently, by the agent(?), so what happens if we need to put it somewhere else, or rename it or etc? A new version of the agent/installer might need to change the paths, say for security reasons? 

As an aside, purely for integration testing, it makes this actually testable in CI 😄 So I'd advocate for an env var override either way tbh, even if we want to have the fallback be the "normal" case

shared/src/Datadog.Trace.ClrProfiler.Native/util.h
dmehala: addressed in https://github.com/DataDog/dd-policy-engine/pull/33

shared/src/Datadog.Trace.ClrProfiler.Native/util.h
dmehala: addressed in f8ad90b94b69519ee8e73011f16f574994c3166a

shared/src/Datadog.Trace.ClrProfiler.Native/util.h
bmermet: Sorry I forgot to reply 🙈

Having this set by the `datadog-installer` would make the feature dependant on a new agent release, which we are trying to avoid. On linux the path is hardcoded in the injector so we are making it part of the contract with the agent.

I don't see a reason not to allow overriding it with an environment variable though.

PR #7678, opened 2025-10-20
Use RuntimeHelpers.GetHashCode in IndexObject

bouwkast:
## Summary of changes

This changes to use a means to get the hash code of an object to handle `null`, `object.GetHashCode` and a fallback to use RuntimeHelpers `GetHashCode` for when `object.GetHashCode` fails.

## Reason for change

My guess is that `GetHashCode` is overridden for the given object and we probably should avoid that?

Spotted the following in Error Tracking:

```
Error : PropagationModuleImpl.PropagateWholeResultWhenInputTainted exception
System.NullReferenceException
   at REDACTED
   at Datadog.Trace.Iast.DefaultTaintedMap.IndexObject(Object objectStored)
   at Datadog.Trace.Iast.DefaultTaintedMap.Get(Object objectToFind)
   at Datadog.Trace.Iast.Propagation.PropagationModuleImpl.PropagateWholeResultWhenInputTainted(String result, Object input1, Object input2, Object input3)
```

## Implementation details

Swapped from one to the other.

## Test coverage

🤷 Didn't try to reproduce

### DIFF
--- a/tracer/src/Datadog.Trace/Iast/DefaultTaintedMap.cs
+++ b/tracer/src/Datadog.Trace/Iast/DefaultTaintedMap.cs
@@ -228,7 +228,7 @@ internal class DefaultTaintedMap : ITaintedMap
 
     private int IndexObject(object objectStored)
     {
-        return Index(PositiveHashCode(objectStored.GetHashCode()));
+        return Index(PositiveHashCode(IastUtils.IdentityHashCode(objectStored)));
     }
 
     private int PositiveHashCode(int hash)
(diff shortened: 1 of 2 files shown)


andrewlock:
@bouwkast, wondering why you closed this one? There was a bug in the implementation (highlighted by codex) that would explain why the tests failed, if that's the reason?

bouwkast:
> @bouwkast, wondering why you closed this one? There was a bug in the implementation (highlighted by codex) that would explain why the tests failed, if that's the reason?

I'm not confident that what I'm doing here makes sense is mainly why I closed it.
I pushed up commits to do that though, but I don't really know if this `Runtime` version is the same as the other 🤷 as I had to just remove the old way.

bouwkast:
@codex review

bouwkast:
@codex review

bouwkast:
@codex review

PR #7680, opened 2025-10-21
[OTEL] OTLP Logs API Support

link04:
## Summary of changes
This PR implements OTLP Logs export for ILogger as part of the OpenTelemetry Logs RFC Key additions:

- Custom OTLP Serializer: Protobuf writer using vendored OpenTelemetry utilities (no SDK dependency)
- Multi-Protocol Support: gRPC transports and HTTP/protobuf
- Auto-Instrumentation: Zero-touch enablement via DD_LOGS_OTEL_ENABLED=true
- Configurations: Full support for OTEL_EXPORTER_OTLP_* environment variables

## Reason for change
The [RFC] OpenTelemetry Logs API support enables customers using OpenTelemetry standards to migrate to Datadog SDK with minimal friction while allowing customers to continue to receive their logs as expected.

## Implementation details
✅ Core Components:
- `OtlpLogsSerializer` - Custom protobuf writer using vendored OTLP field constants (no external dependencies)
- `OtlpExporter` - Multi-protocol exporter (HTTP/protobuf and gRPC) with configurable endpoint/headers/timeout
- `OtlpSubmissionLogSink` - Batching sink that integrates with existing Direct Log Submission pipeline
- `LogPoint` - Internal data model representing OTLP log records with trace/span IDs and attributes

✅ Vendored OpenTelemetry Components:
- `ProtobufSerializer` - Low-level protobuf writer
- `ProtobufOtlpLogFieldNumberConstants` - Official OTLP field numbers
- `OtlpHttpExportClient` - HTTP/protobuf transport client

✅ Integration:
- `DirectSubmissionLogger` auto-instrumentation detects `OtlpSubmissionLogSink`and creates `LogPoint` objects
- When OTLP logs are enabled, they take precedence over Datadog Direct Log Submission to send logs over OTLP instead.

## Test coverage

Unit Tests:
- OtlpSinkTests - Validates batching, trace context injection, attribute handling, severity mapping
- TracerSettingsTests - Configuration parsing (protocol, endpoint, headers, timeout fallbacks)

Integration Tests:
- OpenTelemetrySdkTests.SubmitsOtlpLogs - End-to-end validation with snapshot testing
- Verifies OTLP payload structure and resource attributes
- Confirms trace-log correlation with active spans

### DIFF
--- a/.azure-pipelines/ultimate-pipeline.yml
+++ b/.azure-pipelines/ultimate-pipeline.yml
@@ -25,6 +25,7 @@ pr:
   paths:
     exclude:
       - .azure-pipelines/noop-pipeline.yml
+      - .gitignore
       - .github/
       - .claude/
       - .devcontainer/
@@ -134,7 +135,7 @@ variables:
   DD_LOGGER_DD_TAGS: test.configuration.job:$(System.JobDisplayName)
   DD_LOGGER_ENABLED: true
   DD_COLLECTOR_CPU_USAGE: true
-  ToolVersion: 3.29.0
+  ToolVersion: 3.30.0
   # .NET SDK performance optimization variables
   DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1
   DOTNET_CLI_TELEMETRY_OPTOUT: 1
@@ -1683,11 +1684,13 @@ stages:
       displayName: 'Initialize LocalDB'
       workingDirectory: $(Build.Repository.LocalPath)
 
-    - powershell: |
-        mkdir -Force ./artifacts/build_data/snapshots
-        docker compose -f docker-compose.windows.yml run --rm start-test-agent.windows
-      displayName: Start test agent
-      retryCountOnTaskFailure: 3
+#    TODO: Re-enable test-agent.windows once VM images are updated (hitting rate limits for docker)
+#    Tests with [Trait("RequiresDockerDependency", "true")] are skipped below
+#    - powershell: |
+#        mkdir -Force ./artifacts/build_data/snapshots
+#        docker compose -f docker-compose.windows.yml run --rm start-test-agent.windows
+#      displayName: Start test agent
+#      retryCountOnTaskFailure: 3
 
     - script: tracer\build.cmd CompileTrimmingSamples BuildWindowsIntegrationTests BuildWindowsRegressionTests -Framework $(framework) --code-coverage-enabled $(CodeCoverageEnabled)
       displayName: Build integration tests
@@ -1706,10 +1709,11 @@ stages:
         Filter: $(IntegrationTestFilter)
         SampleName: $(IntegrationTestSampleName)
 
-    - script: docker compose -f docker-compose.windows.yml down
-      displayName: Stop test agent
-      condition: succeededOrFailed()
-      continueOnError: true
+#    TODO: Re-enable test-agent.windows once VM images are updated (hitting rate limits for docker)
+#    - script: docker compose -f docker-compose.windows.yml down
+#      displayName: Stop test agent
+#      condition: succeededOrFailed()
+#      continueOnError: true
 
     - task: PublishTestResults@2
       displayName: publish test results
--- a/.gitignore
+++ b/.gitignore
@@ -389,6 +389,9 @@ datadog-coverage-*
 # nuke
 .nuke/build.schema.json
 
+# Claude Code
+.claude/settings.local.json
+
 # tools/dumps
 tools/dumps/
 !tools/dumps/README_ALPINE_ARM64.md
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -226,3 +226,9 @@ dsm_throughput:
   trigger:
     include: .gitlab/benchmarks/dsm-throughput.yml
   allow_failure: true
+  rules:
+    - if: '$CI_PIPELINE_SOURCE == "push" && $CI_COMMIT_BRANCH == "master"'
+      when: always
+    - if: '$CI_PIPELINE_SOURCE == "schedule" && $BENCHMARK_RUN == "true"'
+      when: always
+    - when: manual
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -31,6 +31,77 @@
 
 
 
+
+## [Release 3.29.0](https://github.com/DataDog/dd-trace-dotnet/releases/tag/v3.29.0)
+
+## Summary
+
+- [OTEL] Metrics API Support - Metrics Reader & OTLP Exporter (#7514)
+- [Azure Functions] Add aas.site.kind tag to agentless logs (https://github.com/DataDog/dd-trace-dotnet/pull/7623)
+- [Azure Functions] Skip instrumenting orchestrator functions (#7640)
+- Various other fixes and improvements
+
+## Changes
+
+### Tracer
+* Rebuild `MutableSettings` on dynamic config changes (#7525)
+* Move `DefaultServiceName` to `MutableSettings` (#7530)
+* Simplify `PerTraceSettings.GetServiceName()` (#7532)
+* [PDB] Handle exception during opening a pdb file (#7573)
+
+### CI Visibility
+* [TestOptimization] Catch IOException on the Git folder discovery (#7658)
+
+### Continuous Profiler
+* [crashtracker/profiler] Notify profiler the application is crashing (#7657)
+
+### Debugger
+* [Dynamic Instrumentation] DEBUG-3695 Return correct RC apply details (#6890)
+* [Dynamic Instrumentation] Skip some debugger tests on .NET 8 or greater (#7631)
+* Minor fix in DynamicConfigurationManager update process (#7664)
+* [Dynamic Instrumentation] DEBUG 3794 Stabilize snapshot enumerable serializing (#6910)
+* [Dynamic Instrumentation] DEBUG-3959 Guard against undefined value expression (#7008)
+* [Debugger Default-On] DEBUG-4341 Forward snapshots to debugger intake (#7388)
+* [Dynamic Instrumentation] Ensure DI metric probe does not send automatic telemetry (#7621)
+* [Dynamic Instrumentation] Remove reference types value property from snapshot (#7635)
+
+### Serverless
+* [Azure Functions] Add `aas.site.kind` tag to agentless logs (#7623)
+*  [Azure Functions] Skip instrumenting orchestrator functions (#7640)
+
+### Miscellaneous
+* [OTEL] Metrics API Support - Metrics Reader & OTLP Exporter (#7514)
+* [internal docs] Add CI troubleshooting documentation (#7634)
+* [OTEL] Moves OTelMetrics Under OpenTelemetry.Metrics (#7638)
+* Minor refactor and added tests for `DynamicConfigurationManager` (#7652)
+* [OTEL] Vendoring OtlpGrpcExportClient And Enabling OTLP Metrics gRPC Tests (#7666)
+* [Code Origin] Add debug logs to code origin for entry span (#7612)
+* [Code Origin] Performance: Add Assembly Filter Caching for Code Origin (#7667)
+* Small OpenTelemetry Metrics fixes and optimizations (#7632)
+
+### Build / Test
+* Optimize `AGENTS.md` with contextual load guidance (#7625)
+* Bump timeit to v0.4.6 (#7647)
+* Improve and simplify the microbenchmarks CI setup (#7613)
+* Capture Exception when sending metrics (#7614)
+* [Test Package Versions Bump] Updating package versions (#7617)
+* Fix profiler_execution_benchmarks flakiness (#7622)
+* Fix the code freeze action (#7626)
+* Revert "[SINT-4157] sanitise ultimate-pipeline.yml" (#7627)
+* Increase Docker run step resiliency for Profiler integration tests (#7629)
+* [SINT-4157] sanitise ultimate-pipeline.yml (#7630)
+* [Azure Functions] Add `Build-AzureFunctionsNuget.ps1` helper script for local developement (#7633)
+* Update `CODEOWNERS` for consistent ownership of managed loader files (#7643)
+* [Test Package Versions Bump] Updating package versions (#7644)
+* Clone the prod dd-trace-dotnet microbenchmark branch for microbenchmarks (#7646)
+* Fix CleanObjFiles job (#7648)
+* [Test Package Versions Bump] Updating package versions (#7655)
+* [OTEL] Switches OTLP Tests To dd-apm-test-agent Instead of MockTracerAgent (#7656)
+* Fix nuget installer smoke tests (#7619)
+
+
+[Changes since 3.28.0](https://github.com/DataDog/dd-trace-dotnet/compare/v3.28.0...v3.29.0)
+
 ## [Release 3.28.0](https://github.com/DataDog/dd-trace-dotnet/releases/tag/v3.28.0)
 
 ## Summary
--- a/profiler/src/ProfilerEngine/Datadog.Linux.ApiWrapper/CMakeLists.txt
+++ b/profiler/src/ProfilerEngine/Datadog.Linux.ApiWrapper/CMakeLists.txt
@@ -2,7 +2,7 @@
 # Project definition
 # ******************************************************
 
-project("Datadog.Linux.ApiWrapper" VERSION 3.29.0)
+project("Datadog.Linux.ApiWrapper" VERSION 3.30.0)
 
 # ******************************************************
 # Compiler options
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/CMakeLists.txt
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/CMakeLists.txt
@@ -2,7 +2,7 @@
 # Project definition
 # ******************************************************
 
-project("Datadog.Profiler.Native.Linux" VERSION 3.29.0)
+project("Datadog.Profiler.Native.Linux" VERSION 3.30.0)
 
 option(RUN_ASAN "Build with Clang Undefined-Behavior Sanitizer" OFF)
 option(RUN_UBSAN "Build with Clang Undefined-Behavior Sanitizer" OFF)
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native.Windows/Resource.rc
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native.Windows/Resource.rc
@@ -62,8 +62,8 @@ END
 
 // ------- version info -------------------------------------------------------
 VS_VERSION_INFO VERSIONINFO
-FILEVERSION             3,29,0,0
-PRODUCTVERSION          3,29,0,0
+FILEVERSION             3,30,0,0
+PRODUCTVERSION          3,30,0,0
 FILEFLAGSMASK           VS_FF_PRERELEASE
 FILEOS                  VOS_NT
 FILETYPE                VFT_DLL
@@ -74,12 +74,12 @@ BEGIN
         BEGIN
             VALUE "CompanyName", "Datadog"
             VALUE "FileDescription", "Continuous Profiler for .NET Applications"
-            VALUE "FileVersion", "3.29.0.0"
+            VALUE "FileVersion", "3.30.0.0"
             VALUE "InternalName", "Native Profiler Engine"
             VALUE "LegalCopyright", "(c) Datadog 2020-2022"
             VALUE "OriginalFilename", "Datadog.Profiler.Native.dll"
             VALUE "ProductName", "Continuous Profiler for .NET Applications"
-            VALUE "ProductVersion", "3.29.0.0"
+            VALUE "ProductVersion", "3.30.0.0"
         END
     END
     BLOCK "VarFileInfo"
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native/dd_profiler_version.h
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native/dd_profiler_version.h
@@ -3,4 +3,4 @@
 
 #pragma once
 
-constexpr auto PROFILER_VERSION = "3.29.0";
+constexpr auto PROFILER_VERSION = "3.30.0";
--- a/profiler/src/ProfilerEngine/ProductVersion.props
+++ b/profiler/src/ProfilerEngine/ProductVersion.props
@@ -5,7 +5,7 @@
 
   <!-- * * * * * * * * * * * INPUTS. Update this section EVERY time the component is shipped/released! * * * * * * * * * * *                -->
   <PropertyGroup>
-    <ProductVersion>3.29.0</ProductVersion>
+    <ProductVersion>3.30.0</ProductVersion>
   </PropertyGroup>
   <!-- * * * * * * * * * * * END OF INPUTS.  * * * * * * * * * * *                                                                          -->
 
--- a/shared/src/Datadog.Trace.ClrProfiler.Native/CMakeLists.txt
+++ b/shared/src/Datadog.Trace.ClrProfiler.Native/CMakeLists.txt
@@ -12,7 +12,7 @@ endif()
 # Project definition
 # ******************************************************
 
-project("Datadog.Trace.ClrProfiler.Native" VERSION 3.29.0)
+project("Datadog.Trace.ClrProfiler.Native" VERSION 3.30.0)
 
 if (UNIVERSAL)
    find_package(GlibcCompat REQUIRED)
--- a/shared/src/Datadog.Trace.ClrProfiler.Native/Resource.rc
+++ b/shared/src/Datadog.Trace.ClrProfiler.Native/Resource.rc
@@ -57,8 +57,8 @@ VS_VERSION_INFO VERSIONINFO
 #else
  FILEFLAGS 0x0L
 #endif
- FILEVERSION             3,29,0,0
- PRODUCTVERSION          3,29,0,0
+ FILEVERSION             3,30,0,0
+ PRODUCTVERSION          3,30,0,0
  FILEOS                  VOS_NT
  FILETYPE                VFT_DLL
 BEGIN
@@ -68,12 +68,12 @@ BEGIN
         BEGIN
             VALUE "CompanyName", "Datadog"
             VALUE "FileDescription", "Native loader for Datadog .NET APM"
-            VALUE "FileVersion", "3.29.0.0"
+            VALUE "FileVersion", "3.30.0.0"
             VALUE "InternalName", "Native loader"
             VALUE "LegalCopyright", "(c) Datadog 2020-2022"
             VALUE "OriginalFilename", "Datadog.Trace.ClrProfiler.Native.dll"
             VALUE "ProductName", "Native loader for Datadog .NET APM"
-            VALUE "ProductVersion", "3.29.0.0"
+            VALUE "ProductVersion", "3.30.0.0"
         END
     END
     BLOCK "VarFileInfo"
--- a/shared/src/msi-installer/WindowsInstaller.wixproj
+++ b/shared/src/msi-installer/WindowsInstaller.wixproj
@@ -17,9 +17,9 @@
     <IntermediateOutputPath>obj\$(Configuration)\$(Platform)\</IntermediateOutputPath>
     <SuppressPdbOutput>True</SuppressPdbOutput>
     <DefineSolutionProperties>false</DefineSolutionProperties>
-    <OutputName>datadog-dotnet-apm-3.29.0-$(Platform)</OutputName> <!-- -The regex recognizes this line -->
+    <OutputName>datadog-dotnet-apm-3.30.0-$(Platform)</OutputName> <!-- -The regex recognizes this line -->
     <MonitoringHomeDirectory Condition="'$(MonitoringHomeDirectory)' == ''">$(MSBuildThisFileDirectory)..\..\bin\monitoring-home</MonitoringHomeDirectory>
-    <DefineConstants>InstallerVersion=3.29.0;MonitoringHomeDirectory=$(MonitoringHomeDirectory);</DefineConstants>
+    <DefineConstants>InstallerVersion=3.30.0;MonitoringHomeDirectory=$(MonitoringHomeDirectory);</DefineConstants>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|x86' ">
     <DefineConstants>$(DefineConstants);Debug</DefineConstants>
--- a/shared/src/native-src/version.h
+++ b/shared/src/native-src/version.h
@@ -1,3 +1,3 @@
 #pragma once
 
-constexpr auto PROFILER_VERSION = "3.29.0";
+constexpr auto PROFILER_VERSION = "3.30.0";
--- a/tracer/build/_build/Build.cs
+++ b/tracer/build/_build/Build.cs
@@ -65,7 +65,7 @@ partial class Build : NukeBuild
     const int LatestMajorVersion = 3;
 
     [Parameter("The current version of the source and build")]
-    readonly string Version = "3.29.0";
+    readonly string Version = "3.30.0";
 
     [Parameter("Whether the current build version is a prerelease(for packaging purposes)")]
     readonly bool IsPrerelease = false;
--- a/tracer/build/artifacts/dd-dotnet.sh
+++ b/tracer/build/artifacts/dd-dotnet.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-TRACER_VERSION="3.29.0"
+TRACER_VERSION="3.30.0"
 
 # Get the directory of the script
 DIR=$(dirname "$(readlink -f "$0")")
--- a/tracer/samples/AutomaticTraceIdInjection/Log4NetExample/Log4NetExample.csproj
+++ b/tracer/samples/AutomaticTraceIdInjection/Log4NetExample/Log4NetExample.csproj
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Datadog.Trace.Bundle" Version="3.28.0" />
+    <PackageReference Include="Datadog.Trace.Bundle" Version="3.29.0" />
     <PackageReference Include="log4net" Version="2.0.12" />
     <PackageReference Include="log4net.Ext.Json" Version="2.0.8.3" />
   </ItemGroup>
--- a/tracer/samples/AutomaticTraceIdInjection/MicrosoftExtensionsExample/MicrosoftExtensionsExample.csproj
+++ b/tracer/samples/AutomaticTraceIdInjection/MicrosoftExtensionsExample/MicrosoftExtensionsExample.csproj
@@ -6,7 +6,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Datadog.Trace.Bundle" Version="3.28.0" />
+    <PackageReference Include="Datadog.Trace.Bundle" Version="3.29.0" />
     <PackageReference Include="Microsoft.Extensions.Hosting" Version="7.0.1" />
     <PackageReference Include="Serilog.Extensions.Logging" Version="3.1.0" />
     <PackageReference Include="Serilog.Sinks.File" Version="5.0.0" />
--- a/tracer/samples/AutomaticTraceIdInjection/NLog40Example/NLog40Example.csproj
+++ b/tracer/samples/AutomaticTraceIdInjection/NLog40Example/NLog40Example.csproj
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Datadog.Trace.Bundle" Version="3.28.0" />
+    <PackageReference Include="Datadog.Trace.Bundle" Version="3.29.0" />
     <PackageReference Include="NLog" Version="4.0.0" />
   </ItemGroup>
 
--- a/tracer/samples/AutomaticTraceIdInjection/NLog45Example/NLog45Example.csproj
+++ b/tracer/samples/AutomaticTraceIdInjection/NLog45Example/NLog45Example.csproj
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Datadog.Trace.Bundle" Version="3.28.0" />
+    <PackageReference Include="Datadog.Trace.Bundle" Version="3.29.0" />
     <PackageReference Include="NLog" Version="4.5.11" />
   </ItemGroup>
 
--- a/tracer/samples/AutomaticTraceIdInjection/NLog46Example/NLog46Example.csproj
+++ b/tracer/samples/AutomaticTraceIdInjection/NLog46Example/NLog46Example.csproj
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Datadog.Trace.Bundle" Version="3.28.0" />
+    <PackageReference Include="Datadog.Trace.Bundle" Version="3.29.0" />
     <PackageReference Include="NLog" Version="4.6.7" />
   </ItemGroup>
 
--- a/tracer/samples/AutomaticTraceIdInjection/SerilogExample/SerilogExample.csproj
+++ b/tracer/samples/AutomaticTraceIdInjection/SerilogExample/SerilogExample.csproj
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Datadog.Trace.Bundle" Version="3.28.0" />
+    <PackageReference Include="Datadog.Trace.Bundle" Version="3.29.0" />
     <PackageReference Include="Serilog" Version="2.9.0" />
     <PackageReference Include="Serilog.Formatting.Compact" Version="1.1.0" />
     <PackageReference Include="Serilog.Sinks.File" Version="4.1.0" />
--- a/tracer/samples/ConsoleApp/Alpine3.10.dockerfile
+++ b/tracer/samples/ConsoleApp/Alpine3.10.dockerfile
@@ -16,7 +16,7 @@ COPY --from=build /app/out .
 
 # Set up Datadog APM
 RUN apk --no-cache update && apk add curl
-ARG TRACER_VERSION=3.28.0
+ARG TRACER_VERSION=3.29.0
 RUN mkdir -p /var/log/datadog
 RUN mkdir -p /opt/datadog
 RUN curl -L https://github.com/DataDog/dd-trace-dotnet/releases/download/v${TRACER_VERSION}/datadog-dotnet-apm-${TRACER_VERSION}-musl.tar.gz \
--- a/tracer/samples/ConsoleApp/Alpine3.9.dockerfile
+++ b/tracer/samples/ConsoleApp/Alpine3.9.dockerfile
@@ -16,7 +16,7 @@ COPY --from=build /app/out .
 
 # Set up Datadog APM
 RUN apk --no-cache update && apk add curl
-ARG TRACER_VERSION=3.28.0
+ARG TRACER_VERSION=3.29.0
 RUN mkdir -p /var/log/datadog
 RUN mkdir -p /opt/datadog
 RUN curl -L https://github.com/DataDog/dd-trace-dotnet/releases/download/v${TRACER_VERSION}/datadog-dotnet-apm-${TRACER_VERSION}-musl.tar.gz \
--- a/tracer/samples/ConsoleApp/Debian.dockerfile
+++ b/tracer/samples/ConsoleApp/Debian.dockerfile
@@ -15,7 +15,7 @@ WORKDIR /app
 COPY --from=build /app/out .
 
 # Set up Datadog APM
-ARG TRACER_VERSION=3.28.0
+ARG TRACER_VERSION=3.29.0
 RUN mkdir -p /var/log/datadog
 RUN mkdir -p /opt/datadog
 RUN curl -LO https://github.com/DataDog/dd-trace-dotnet/releases/download/v${TRACER_VERSION}/datadog-dotnet-apm_${TRACER_VERSION}_amd64.deb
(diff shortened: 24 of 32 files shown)


tracer/missing-nullability-files.csv
```diff
@@ -332,6 +332,7 @@ src/Datadog.Trace/Iast/Analyzers/UserStringInterop.cs
 src/Datadog.Trace/Iast/Aspects/DebugAspects.cs
 src/Datadog.Trace/Iast/Dataflow/AspectFilter.cs
 src/Datadog.Trace/Iast/Dataflow/AspectType.cs
+src/Datadog.Trace/OpenTelemetry/Stubs/OtlpSignalType.cs
```
andrewlock: We should add `#nullable enable` to this files

tracer/src/Datadog.Trace.Trimming/build/Datadog.Trace.Trimming.xml
```diff
@@ -237,6 +237,10 @@
       <type fullname="System.Diagnostics.DebuggerTypeProxyAttribute" />
    </assembly>
    <assembly fullname="System.Diagnostics.DiagnosticSource">
+      <type fullname="System.Diagnostics.Activity" />
+      <type fullname="System.Diagnostics.ActivityIdFormat" />
+      <type fullname="System.Diagnostics.ActivitySpanId" />
+      <type fullname="System.Diagnostics.ActivityTraceId" />
```
andrewlock, line 243: I'm kind of surprised we don't reference these already, and it suggests to me we _probably_ are doing something wrong here, but I don't know enough about it off hand 🤔 I expect @zacharycmontoya can shed light

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Logging/ILogger/DirectSubmission/DirectSubmissionLogger.cs
```diff
@@ -57,26 +62,40 @@ public void Log<TState>(int logLevel, object eventId, TState state, Exception? e
...
+                TelemetryFactory.Metrics.RecordCountDirectLogLogs(MetricTags.IntegrationName.ILogger);
+                _sink.EnqueueLog(log);
+                return;
+            }
 
-            TelemetryFactory.Metrics.RecordCountDirectLogLogs(MetricTags.IntegrationName.ILogger);
-            _sink.EnqueueLog(log);
+#if NETCOREAPP3_1_OR_GREATER
+            // OTLP logs: use structured data directly
+            if (_sink is Datadog.Trace.Logging.DirectSubmission.Sink.OtlpSubmissionLogSink)
```
andrewlock: As discussed offline, I think we can improve this a little by adding another abstraction and/or moving some things around

- Creating a new type, e.g. something like ILogCreator
- Create two different version, that just are for creating the `LoggerDirectSubmissionLogEvent`, `OtelLogCreator` and `DatadogLogCreator` (or something)
- Inject the correct one into `DirectSubmissionLogger` when you create it.

That would remove the need for the `if` and for the `#if NETCOREAPP3_1_OR_GREATER` in this code.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Logging/ILogger/DirectSubmission/Formatting/OtlpLogEventBuilder.cs
```diff
@@ -0,0 +1,96 @@
...
+                attributes["EventId"] = eventId.GetHashCode();
+            }
+
+            return attributes ?? new Dictionary<string, object?>();
+        }
+
+        private static (ActivityTraceId? TraceId, ActivitySpanId? SpanId, int Flags) ExtractTraceContext()
+        {
+            var activity = System.Diagnostics.Activity.Current;
+            if (activity != null && activity.IdFormat == ActivityIdFormat.W3C)
```
andrewlock: As discussed, is this def correct 🤔 Do we end up with the logs being correctly correlated to the _datadog_ spans in this case? I ask because I think we _sometimes_ "override" the span/trace ID from the activities when we're creating our "mirror" spans. Again prob a question for @zacharycmontoya

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Logging/ILogger/DirectSubmission/LoggerDirectSubmissionLogEvent.cs
```diff
@@ -19,6 +23,10 @@ public LoggerDirectSubmissionLogEvent(string? serializedEvent)
             _serializedEvent = serializedEvent;
         }
 
+#if NETCOREAPP3_1_OR_GREATER
+        public LogPoint? OtlpLog { get; set; }
+#endif
+
```
andrewlock, line 29: Hmmm, ok, this is making me wonder if we should restructure this entirely 🤔 

_Currently_ you're using the `LoggerDirectSubmissionLogEvent` to "smuggle" this `LogPoint` into the sink. But _really_ `LogPoint` and `LoggerDirectSubmissionLogEvent` are kind of the two different _types_ of event I think?

Which makes me wonder if we should refactor the abstractions in general, but looking at it right now, I'm not exactly sure _how_ to make that change/introduce that seam. I think the answer is _probably_ that it should be in `DirectSubmissionLogger` - we should have two different types, and they should just use two different implementations. But again, I think that's something we can safely improve/refactor later

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -264,6 +264,69 @@ not null when string.Equals(x, "lowmemory", StringComparison.OrdinalIgnoreCase)
...
+                            .WithKeys(ConfigurationKeys.OpenTelemetry.ExporterOtlpLogsTimeoutMs, ConfigurationKeys.OpenTelemetry.ExporterOtlpTimeoutMs)
+                            .AsInt32(defaultValue: 10_000);
+
+            var otelLogsExporter = config
+                    .WithKeys(ConfigurationKeys.OpenTelemetry.LogsExporter);
+
+            OtelLogsExporterEnabled = string.Equals(otelLogsExporter.AsString(defaultValue: "otlp"), "otlp", StringComparison.OrdinalIgnoreCase);
+
+            var otelLogsExporterResult = otelLogsExporter
+               .AsBoolResult(
```
andrewlock, line 321: nit: calling `AsString` and then `AsBool` will record the telemetry twice. It's prob better to keep `otelLogsExporterResult` as-is, and then use:

```csharp
OtelLogsExporterEnabled = otelLogsExporterResult.ConfigurationResult is { IsValid: true, Result: true };
```

tracer/src/Datadog.Trace/Logging/DirectSubmission/DirectLogSubmissionSettings.cs
```diff
@@ -140,42 +154,51 @@ public DirectLogSubmissionSettings(IConfigurationSource? source, IConfigurationT
                 }
             }
 
-            var isEnabled = enabledIntegrations is not null;
+            // For OTLP logs, automatically enable ILogger integration
```
andrewlock: Note that this doesn't _disable_ the other integrations (Serilog etc) if they're there... should it?

tracer/src/Datadog.Trace/Logging/DirectSubmission/Sink/OtlpSubmissionLogSink.cs
```diff
@@ -0,0 +1,83 @@
...
+                return true;
+            }
+
+            try
+            {
+                if (_otlpExporter == null)
+                {
+                    _otlpExporter = new OtlpExporter(Tracer.Instance.Settings);
+                }
+
```
andrewlock: We shouldn't call `Tracer.Instance.Settings` - this should be passed in as a constructor argument and initialized there

```suggestion
```

tracer/src/Datadog.Trace/Logging/DirectSubmission/Sink/OtlpSubmissionLogSink.cs
```diff
@@ -0,0 +1,83 @@
...
+{
+    internal class OtlpSubmissionLogSink : BatchingSink<DirectSubmissionLogEvent>, IDirectSubmissionLogSink
+    {
+        private readonly IDatadogLogger _logger = DatadogLogging.GetLoggerFor<OtlpSubmissionLogSink>();
+        private OtlpExporter? _otlpExporter;
+
+        public OtlpSubmissionLogSink(BatchingSinkOptions sinkOptions)
+            : base(sinkOptions, null)
+        {
+        }
```
andrewlock:
```suggestion
        private readonly OtlpExporter _otlpExporter;

        public OtlpSubmissionLogSink(BatchingSinkOptions sinkOptions, TracerSettings settings)
            : base(sinkOptions, null)
        {
            _otlpExporter = new OtlpExporter(settings)
        }
```

tracer/src/Datadog.Trace/OpenTelemetry/Logs/LogPoint.cs
```diff
@@ -0,0 +1,133 @@
...
+            // Add trace context if available
+#if NETCOREAPP3_1_OR_GREATER
+            var activity = System.Diagnostics.Activity.Current;
+            if (activity != null)
+            {
+                logPoint.TraceId = activity.TraceId;
+                logPoint.SpanId = activity.SpanId;
+                logPoint.Flags = activity.Recorded ? 1 : 0;
+            }
+#endif
```
andrewlock: We're already grabbing this elsewhere too, we almost certainly shouldn't be grabbing in both places (and as mentioned above, I'm not 100% they're correct with DD spans 😅 )

tracer/src/Datadog.Trace/OpenTelemetry/Logs/OtlpExporter.cs
```diff
@@ -0,0 +1,261 @@
...
+        /// Exports a batch of logs using OTLP protocol.
+        /// This method implements the LogExporter base class Export operation.
+        /// </summary>
+        /// <param name="logs">Batch of logs to export</param>
+        /// <returns>ExportResult indicating success or failure</returns>
+        public ExportResult Export(IReadOnlyList<LogPoint> logs)
+        {
+            // For backward compatibility, call the async version synchronously
+            return ExportAsync(logs).GetAwaiter().GetResult();
+        }
```
andrewlock: 😱 No, don't do that, this is sync-over-async 😅 Just delete this method, we don't need it

```suggestion
```

tracer/src/Datadog.Trace/OpenTelemetry/Stubs/OpenTelemetryProtocolExporterEventSource.cs
```diff
@@ -78,6 +78,14 @@ public void RequestTimedOut(params object?[]? args)
         public void GrpcRetryDelayParsingFailed(params object?[]? args)
         {
         }
+
+        public void BufferExceededMaxSize(string signalType, int bufferSize)
```
zacharycmontoya, line 82: This file will be good for implementing telemetry metrics, though this is a central place for all signals not just Logs or not just Metrics 🤔

tracer/src/Datadog.Trace/OpenTelemetry/Logs/OtlpExporter.cs
```diff
@@ -0,0 +1,261 @@
...
+        /// Shuts down the exporter and ensures all pending exports complete.
+        /// </summary>
+        /// <param name="timeoutMilliseconds">Maximum time to wait for shutdown</param>
+        /// <returns>True if shutdown completed successfully</returns>
+        public bool Shutdown(int timeoutMilliseconds)
+        {
+            try
+            {
+                _httpClient.CancelPendingRequests();
+                _httpClient.Dispose();
```
andrewlock: nit: We should probably check that we can't call shutdown _while_ an existing send is ongoing. The reason being, we could end up with exceptions in the _send_ path instead of here if it tries to use a disposed client. That probably doesn't _really_ matter, it's just something to bear in mind

tracer/src/Datadog.Trace/OpenTelemetry/Logs/OtlpExporter.cs
```diff
@@ -0,0 +1,261 @@
...
+                Log.Debug("OTLP Log Exporter shutdown completed.");
+                return true;
+            }
+            catch (Exception ex)
+            {
+                Log.Error(ex, "Error during OTLP Log Exporter shutdown.");
```
andrewlock: I'm guessing `CancelPendingRequests` may throw an exception. In which case, we may want to catch those exceptions explicitly, or otherwise downgrade this to warning. Unless we never _Should_ have ongoing requests, in which case we're probably fine

tracer/src/Datadog.Trace/OpenTelemetry/Logs/OtlpExporter.cs
```diff
@@ -0,0 +1,261 @@
...
+            catch (Exception ex)
+            {
+                Log.Error(ex, "Error sending OTLP request.");
+                return false;
+            }
+        }
+
+        private async Task<bool> SendHttpJsonRequest(byte[] otlpPayload)
+        {
+            Log.Warning("HTTP/JSON protocol is not yet implemented for logs, falling back to HTTP/protobuf");
```
andrewlock: Do we need this warning? 🤔 We should probably just update our settings to not allow using http/json if we don't support this

tracer/src/Datadog.Trace/OpenTelemetry/Logs/OtlpExporter.cs
```diff
@@ -0,0 +1,261 @@
...
+            return await SendHttpProtobufRequest(otlpPayload).ConfigureAwait(false);
+        }
+
+        private async Task<bool> SendHttpProtobufRequest(byte[] otlpPayload)
+        {
+            if (_httpExportClient is null)
+            {
+                Log.Warning("HTTP/Protobuf selected but HTTP client is not initialized; cannot send logs.");
+                return false;
+            }
```
andrewlock: This is another smell IMO. We should refactor this type so we _can't_ be in this "invalid" state

tracer/src/Datadog.Trace/OpenTelemetry/Logs/OtlpExporter.cs
```diff
@@ -0,0 +1,261 @@
...
+
+            try
+            {
+                var resp = await Task.Run(() => _httpExportClient.SendExportRequest(
+                                            otlpPayload,
+                                            otlpPayload.Length,
+                                            default))
+                                    .ConfigureAwait(false);
+                return resp.Success;
```
andrewlock: We shouldn't push this into a background task, this is class AI silliness 😅 
```suggestion
                var resp = await _httpExportClient.SendExportRequest(
                                            otlpPayload,
                                            otlpPayload.Length,
                                            default)
                                    .ConfigureAwait(false);
                return resp.Success;
```

tracer/src/Datadog.Trace/OpenTelemetry/Logs/OtlpExporter.cs
```diff
@@ -0,0 +1,261 @@
...
+            }
+        }
+
+        private async Task<bool> SendGrpcRequest(byte[] otlpPayload)
+        {
+            if (_grpcClient is null)
+            {
+                Log.Warning("GRPC selected but gRPC client is not initialized; falling back to HTTP/protobuf.");
+                return await SendHttpProtobufRequest(otlpPayload).ConfigureAwait(false);
+            }
```
andrewlock: as above

tracer/src/Datadog.Trace/OpenTelemetry/Logs/OtlpExporter.cs
```diff
@@ -0,0 +1,261 @@
...
+                var dataLength = otlpPayload.Length - 5;
+                System.Buffers.Binary.BinaryPrimitives.WriteUInt32BigEndian(
+                    new System.Span<byte>(otlpPayload, 1, 4),
+                    (uint)dataLength);
+
+                var resp = await Task.Run(() => _grpcClient.SendExportRequest(
+                                            otlpPayload,
+                                            otlpPayload.Length,
+                                            default))
+                                    .ConfigureAwait(false);
```
andrewlock:
```suggestion
                var resp = await _grpcClient.SendExportRequest(
                                            otlpPayload,
                                            otlpPayload.Length,
                                            default)
                                    .ConfigureAwait(false);
```

tracer/src/Datadog.Trace/OpenTelemetry/Logs/OtlpLogsSerializer.cs
```diff
@@ -0,0 +1,278 @@
...
+            {
+                writePosition = WriteResourceAttribute(buffer, writePosition, "service.version", settings.ServiceVersion!);
+            }
+
+            if (!StringUtil.IsNullOrEmpty(settings.Environment))
+            {
+                writePosition = WriteResourceAttribute(buffer, writePosition, "deployment.environment", settings.Environment!);
+            }
+
+            if (settings.GlobalTags.Count > 0)
```
andrewlock: PSA, these can change at runtime so we will need to handle that soon with `MutableSettings`

tracer/src/Datadog.Trace/OpenTelemetry/Logs/OtlpLogsSerializer.cs
```diff
@@ -0,0 +1,278 @@
...
+            writePosition = ProtobufSerializer.WriteEnumWithTag(buffer, writePosition, LogRecord_Severity_Number, log.GetSeverityNumber());
+            writePosition = ProtobufSerializer.WriteStringWithTag(buffer, writePosition, LogRecord_Severity_Text, log.GetSeverityText());
+            writePosition = WriteLogRecordBody(buffer, writePosition, log.Message);
+
+            foreach (var attr in log.Attributes)
+            {
+                writePosition = WriteKeyValueAttribute(buffer, writePosition, attr.Key, attr.Value?.ToString() ?? string.Empty);
+            }
+
+            if (log.TraceId.HasValue && !log.TraceId.Value.ToString().Equals("00000000000000000000000000000000", StringComparison.OrdinalIgnoreCase))
```
andrewlock: This seems janky 😅

tracer/src/Datadog.Trace/OpenTelemetry/Logs/OtlpLogsSerializer.cs
```diff
@@ -0,0 +1,278 @@
...
+            {
+                writePosition = ProtobufSerializer.WriteTag(buffer, writePosition, LogRecord_Trace_Id, ProtobufWireType.LEN);
+                writePosition = ProtobufSerializer.WriteLength(buffer, writePosition, TraceIdSize);
+                writePosition = WriteTraceId(buffer, writePosition, log.TraceId.Value);
+            }
+
+            if (log.SpanId.HasValue && !log.SpanId.Value.ToString().Equals("0000000000000000", StringComparison.OrdinalIgnoreCase))
```
andrewlock: same here 😄

tracer/src/Datadog.Trace/OpenTelemetry/Logs/OtlpLogsSerializer.cs
```diff
@@ -0,0 +1,278 @@
...
+            writePosition = ProtobufSerializer.WriteStringWithTag(buffer, writePosition, AnyValue_String_Value, value);
+
+            ProtobufSerializer.WriteReservedLength(buffer, valueLengthPosition, writePosition - (valueLengthPosition + ReserveSizeForLength));
+
+            ProtobufSerializer.WriteReservedLength(buffer, attributeLengthPosition, writePosition - (attributeLengthPosition + ReserveSizeForLength));
+
+            return writePosition;
+        }
+
+        private static int WriteTraceId(byte[] buffer, int writePosition, System.Diagnostics.ActivityTraceId traceId)
```
andrewlock: Really feels like there should be better option with this, but...🤔

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -264,6 +264,69 @@ not null when string.Equals(x, "lowmemory", StringComparison.OrdinalIgnoreCase)
...
+                                  },
+                                  validator: null);
+
+            OtlpLogsEndpoint = config
+                .WithKeys(ConfigurationKeys.OpenTelemetry.ExporterOtlpLogsEndpoint)
+                .GetAs(
+                    defaultValue: new DefaultResult<Uri>(
+                        result: OtlpLogsProtocol switch
+                        {
+                            OtlpProtocol.Grpc => OtlpEndpoint,
```
zacharycmontoya, line 299: There's a small bug that can happen: The port for the `OtlpEndpoint` is being set based on `OtlpMetricsProtocol` config even though this is the log signal, so if the protocols are different between metrics and logs we'd be sending them with the wrong protocol. Priority is low because this situation is likely to be rare, but we could address that in this PR

tracer/src/Datadog.Trace/Logging/DirectSubmission/DirectLogSubmissionSettings.cs
```diff
@@ -28,6 +28,13 @@ internal class DirectLogSubmissionSettings
         internal const int DefaultBatchSizeLimit = 1000;
         internal const int DefaultQueueSizeLimit = 100_000;
         internal const int DefaultBatchPeriodSeconds = 2;
+
+        // OpenTelemetry OTLP logs defaults (OTEL_BLRP_* environment variables)
+        internal const DirectSubmissionLogLevel OtlpMinimumLevel = DirectSubmissionLogLevel.Verbose;
```
zacharycmontoya: This seems too verbose. Is there a source that says we should emit Verbose-level logs by default for OTLP rather than Information-level logs?

tracer/src/Datadog.Trace/Logging/DirectSubmission/DirectLogSubmissionSettings.cs
```diff
@@ -67,10 +74,13 @@ public DirectLogSubmissionSettings(IConfigurationSource? source, IConfigurationT
                                          },
                                          validator: x => !string.IsNullOrEmpty(x));
 
+            var defaultMinLevel = otlpLogsEnabled ? OtlpMinimumLevel : DefaultMinimumLevel;
+            var defaultMinLevelName = otlpLogsEnabled ? nameof(OtlpMinimumLevel) : nameof(DefaultMinimumLevel);
```
zacharycmontoya: This is incorrect, for example `nameof(DefaultMinimumLevel)` will return `"DefaultMinimumLevel"` whereas I think you meant to return the value "Information" as we were doing before. This suggestion handles both cases

```suggestion
            var defaultMinLevelName = defaultMinLevel.ToString();
```

tracer/src/Datadog.Trace/Logging/DirectSubmission/DirectLogSubmissionSettings.cs
```diff
@@ -28,6 +28,13 @@ internal class DirectLogSubmissionSettings
         internal const int DefaultBatchSizeLimit = 1000;
         internal const int DefaultQueueSizeLimit = 100_000;
         internal const int DefaultBatchPeriodSeconds = 2;
+
+        // OpenTelemetry OTLP logs defaults (OTEL_BLRP_* environment variables)
+        internal const DirectSubmissionLogLevel OtlpMinimumLevel = DirectSubmissionLogLevel.Verbose;
+        internal const int OtlpBatchSizeLimit = 512;        // OTEL_BLRP_MAX_EXPORT_BATCH_SIZE
+        internal const int OtlpQueueSizeLimit = 2048;       // OTEL_BLRP_MAX_QUEUE_SIZE
+        internal const int OtlpBatchPeriodSeconds = 1;      // OTEL_BLRP_SCHEDULE_DELAY
```
zacharycmontoya, line 37: Note for later: OTEL_BLRP_SCHEDULE_DELAY is in milliseconds, whereas our previous configuration was in seconds. We will want to look into updating the previous config to accept milliseconds so these two settings are more interoperable

tracer/src/Datadog.Trace/Logging/DirectSubmission/DirectLogSubmissionSettings.cs
```diff
@@ -82,19 +92,23 @@ public DirectLogSubmissionSettings(IConfigurationSource? source, IConfigurationT
 
             GlobalTags = new ReadOnlyDictionary<string, string>(globalTags ?? []);
 
+            var actualBatchSize = otlpLogsEnabled ? OtlpBatchSizeLimit : DefaultBatchSizeLimit;
```
zacharycmontoya: super small nit: I think it would be more clear to rename `actualBatchSize` to `defaultBatchSize` (same for the others), while renaming the constant to `DatadogDefaultBatchSizeLimit`

tracer/src/Datadog.Trace.Trimming/build/Datadog.Trace.Trimming.xml
zacharycmontoya: In the metrics PR Maximo recently did, there was no use of Activity. In this PR we are accessing the current trace-id and span-id from `Activity.Current` (since this is running only on netcoreapp3.1+) and this is the first time we're calling it directly. I too was surprised seeing this.

My advice is to use our own active Trace ID rather than the Activity one, or at least prefer our own Trace ID then fall back to Activity, since we're likely using our tracing integration as well

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Logging/ILogger/DirectSubmission/Formatting/OtlpLogEventBuilder.cs
zacharycmontoya: Yep, I think we should prefer our own Datadog ID's but we can fallback to Activity if our tracing isn't enabled

tracer/src/Datadog.Trace/Logging/DirectSubmission/DirectLogSubmissionSettings.cs
zacharycmontoya: I'm actually unsure if we should be setting ILogger on by default because if, for example, a customer uses ILogger+Serilog, then maybe they want the Serilog ones to win out. I think it's probably better if the user sees duplicate information short-term so they can go and disable whichever source they want...but then ILogger is not disable-able 🤔

tracer/src/Datadog.Trace/OpenTelemetry/Logs/OtlpExporter.cs
zacharycmontoya: +1

tracer/src/Datadog.Trace/Logging/DirectSubmission/DirectLogSubmissionSettings.cs
link04: I ran my test and was missing the most verbose one hence why I used the highest level, currently customers could use `DD_TRACE_LOG_MINIMUM_LEVEL` but I wanted to avoid that so we would need to look into adding a specific env var they can use(like the one we spoke about they dont have yet for ,NET), that said I can keep it as Information by default like OTEL does just not sure if we want customers setting `DD_TRACE_LOG_MINIMUM_LEVEL` or I can follow up in another PR about this.

tracer/src/Datadog.Trace/Logging/DirectSubmission/DirectLogSubmissionManager.cs
```diff
@@ -40,6 +40,14 @@ public static DirectLogSubmissionManager Create(
             IGitMetadataTagsProvider gitMetadataTagsProvider)
         {
             var formatter = new LogFormatter(settings, directLogSettings, azureAppServiceSettings, serviceName, env, serviceVersion, gitMetadataTagsProvider);
+
+#if NETCOREAPP3_1_OR_GREATER
+            if (settings.OpenTelemetryLogsEnabled is true && settings.OtelLogsExporterEnabled is true)
```
zacharycmontoya: In TracerSettings.cs you already reassign settings.OpenTelemetryLogsEnabled like this: `OpenTelemetryLogsEnabled = OpenTelemetryLogsEnabled is true && OtelLogsExporterEnabled is true;`. So you shouldn't need to do it again.

```suggestion
            if (settings.OpenTelemetryLogsEnabled)
```

tracer/src/Datadog.Trace/Logging/DirectSubmission/DirectLogSubmissionSettings.cs
zacharycmontoya: Let's set as Information by default, and we can look into `DD_TRACE_LOG_MINIMUM_LEVEL` later

tracer/src/Datadog.Trace/Logging/DirectSubmission/DirectLogSubmissionSettings.cs
```diff
@@ -140,42 +154,51 @@ public DirectLogSubmissionSettings(IConfigurationSource? source, IConfigurationT
...
+
             if (!AzureFunctionsHostEnabled && EnvironmentHelpers.IsRunningInAzureFunctionsHost())
             {
-                // If we are in `dotnet-isolated` Azure Functions host customers may want to disable direct log submission
-                // because the host will re-log logs coming from the worker process causing duplicate logs.
                 isEnabled = false;
             }
 
-            if (string.IsNullOrWhiteSpace(Host))
+            if (!otlpLogsEnabled)
```
zacharycmontoya, line 179: Are there any validations we need to do for OTLP Logs?

tracer/src/Datadog.Trace/Logging/DirectSubmission/DirectLogSubmissionManager.cs
link04: Lol yeah missed it after the change ooops

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Logging/ILogger/DirectSubmission/Formatting/OtlpLogEventBuilder.cs
```diff
@@ -0,0 +1,96 @@
...
+            }
+
+            // Add EventId if present
+            if (eventId.GetHashCode() != 0)
+            {
+                attributes ??= new Dictionary<string, object?>();
+                attributes["EventId"] = eventId.GetHashCode();
+            }
+
+            return attributes ?? new Dictionary<string, object?>();
```
zacharycmontoya: super small nit: You're going to create a new `Dictionary<string, object?>` no matter what, so you may as well initialize the local `attributes` variable with a new one instead of `null`

tracer/src/Datadog.Trace/Logging/DirectSubmission/Sink/OtlpSubmissionLogSink.cs
zacharycmontoya: +1

tracer/src/Datadog.Trace/Logging/DirectSubmission/DirectLogSubmissionSettings.cs
link04: Since we don't require a specific Site or API and have defaults instead I think we should be good with what we have within TracerSettings.cs

tracer/src/Datadog.Trace/Logging/DirectSubmission/DirectLogSubmissionSettings.cs
zacharycmontoya: I think for the purposes of this PR we can leave this code as-is

tracer/src/Datadog.Trace/OpenTelemetry/Logs/LogPoint.cs
zacharycmontoya: +1

My opinion is that the trace context information be passed in from the caller, either in the constructor or just updating the properties. Let's not call AsyncLocal state in here

tracer/src/Datadog.Trace/OpenTelemetry/Logs/OtlpLogsSerializer.cs
zacharycmontoya: I recommended changing to using our Datadog trace-ids/span-ids, but if we were to keep using the `ActivityTraceId` type, I think you'd want this comparison

```suggestion
            if (log.TraceId != default(ActivityTraceId))
```

tracer/src/Datadog.Trace/Logging/DirectSubmission/DirectLogSubmissionSettings.cs
link04: So the RFC option 3  is explicitly for built in loggers so I think we should disable everything else ATM? 
So I am thinking we should disable, then come back and look into supporting all loggers. WDYT?

tracer/src/Datadog.Trace/OpenTelemetry/Logs/OtlpLogsSerializer.cs
zacharycmontoya: Again, this may be obsolete if we switch away from ActivityTraceId, but the OTel .NET SDK does [this](https://github.com/open-telemetry/opentelemetry-dotnet/blob/cdbbf435ff3c2177c86c6ccd310f756cebfb493d/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/Serializer/ProtobufOtlpTraceSerializer.cs#L248-L253):

```csharp
        var traceBytes = new Span<byte>(buffer, position, TraceIdSize);
        activityTraceId.CopyTo(traceBytes);
        return position + TraceIdSize;
```

tracer/src/Datadog.Trace/OpenTelemetry/Logs/OtlpExporter.cs
zacharycmontoya: The problem is that the Http and gRPC exports are _synchronous_ so if we want to maintain a higher-level async interface, then I think we'll need this

zacharycmontoya: I do wonder if we could make this synchronous all the way up, but I'd worry about difficulties of eventually migrating the HTTP code to .NET Framework

zacharycmontoya: Oh no, I don't like this at all 😱 From the vendored HTTP export code, it could be sent synchronously or asynchronously: https://github.com/open-telemetry/opentelemetry-dotnet/blob/7bd113780baf2a2382459d930189c61b5973adfa/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/ExportClient/OtlpExportClient.cs#L117

zacharycmontoya: Ok we definitely have some follow-up work there...

tracer/test/test-applications/integrations/Samples.OpenTelemetrySdk/Program.cs
```diff
@@ -45,16 +49,32 @@ public static async Task Main(string[] args)
...
+#if OTEL_1_5
+        using var loggerFactory = CustomLoggerFactoryBuilderExtensions
+            .AddOtlpExporterIfEnvironmentVariablePresent();
+#endif
+
         _tracer = tracerProvider.GetTracer(serviceName); // The version is omitted so the ActivitySource.Version / otel.library.version is not set
         var _otherLibraryTracer = tracerProvider.GetTracer(otherLibraryName, version: otherLibraryVersion);
 
+#if OTEL_1_5
+        var logger = loggerFactory.CreateLogger(string.Empty);
```
zacharycmontoya: nit: Can you add a name for the logger? I expect users to always do so

tracer/test/test-applications/integrations/Samples.OpenTelemetrySdk/Program.cs
```diff
@@ -67,13 +87,23 @@ public static async Task Main(string[] args)
             using (otherSpan = _otherLibraryTracer.StartActiveSpan("Response"))
             {
                 PrintSpanStartedInformation(otherSpan);
+                
+#if OTEL_1_5
+                logger.LogInformation("Response from other library");
```
zacharycmontoya: If you're trying to test "from another library", you should get a logger with a different name. And then we can see that the OTLP logs scope changes

tracer/test/snapshots/OpenTelemetrySdkTests.SubmitsOtlpLogs.verified.txt
```diff
@@ -0,0 +1,133 @@
...
+            {
+              "key": "telemetry.sdk.version",
+              "value": {
+                "string_value": "sdk-version"
+              }
+            },
+            {
+              "key": "service.name",
+              "value": {
+                "string_value": "unknown_service:dotnet"
```
zacharycmontoya: I think you'll need to have a separate OTel vs DD snapshot since the `service.name` (among others) may be different

tracer/test/Datadog.Trace.Tests/Logging/DirectSubmission/Sink/OtlpSinkTests.cs
```diff
@@ -0,0 +1,291 @@
...
+        private static readonly TimeSpan TinyWait = TimeSpan.FromMilliseconds(50);
+
+        [Fact]
+        public void SinkSendsLogsToOtlpExporter()
+        {
+            var mutex = new ManualResetEventSlim();
+            var capturedLogs = new List<LogPoint>();
+
+            var options = new BatchingSinkOptions(batchSizeLimit: 2, queueLimit: DefaultQueueLimit, period: TinyWait);
+            var sink = new TestOtlpSink(options, logs =>
```
zacharycmontoya: Rather than create a new sink to test, I think the most effective way to test this would be to use the exact same class that is going to be used at runtime, which is the `OtlpSubmissionLogSink` object. Perhaps you can pass an `OtlpExporter` into the `OtlpSubmissionLogSink` constructor, so this test code could be controlled by creating a test `OtlpExporter` class rather than a test `OtlpSubmissionLogSink` class

zacharycmontoya: And the test `OtlpExporter` class could take the delegate that sets the mutex. But the idea is that the code path executed by the `OtlpSubmissionLogSink` to send data to its OtlpExporter is tested unchanged

tracer/src/Datadog.Trace/Logging/DirectSubmission/DirectLogSubmissionSettings.cs
zacharycmontoya: Yes disabling everything else should be fine, but let's make sure to write an Information level debug log that we're doing so

zacharycmontoya (commented):
Looking through the PR, I left some comments throughout but the most important for me are the comments regarding testing. I'd like to resolve those and other parts can be refined later

tracer/src/Datadog.Trace/OpenTelemetry/Logs/LogPoint.cs
link04: Thank you for the check! I've updated this in the latest commit.

tracer/test/snapshots/OpenTelemetrySdkTests.SubmitsOtlpLogs.verified.txt
link04: I've left one for now will update in a new PR with different tags and resource and different combinations. Otherwise I could addd the same service name, but I've added multiple scopes and fixed how they show up.

link04: Just saw your commend about testing so will add the service names and update the snapshot.

tracer/src/Datadog.Trace/Logging/DirectSubmission/Sink/OtlpSubmissionLogSink.cs
```diff
@@ -4,80 +4,73 @@
 // </copyright>
 
 #if NETCOREAPP3_1_OR_GREATER
-#nullable enable
```
zacharycmontoya: You should put this back in. We want to enable nullable reference types as much as possible in the project

tracer/test/Datadog.Trace.Tests/Logging/DirectSubmission/Sink/OtlpSinkTests.cs
link04: Attempt to solve this in this commit: https://github.com/DataDog/dd-trace-dotnet/pull/7680/commits/eeeb189cd62300c5f318b164401fa1050e949ad3

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Logging/ILogger/DirectSubmission/DirectSubmissionLoggerProvider.cs
```diff
@@ -75,7 +83,6 @@ public void Dispose()
         [DuckReverseMethod(ParameterTypeNames = new[] { "Microsoft.Extensions.Logging.IExternalScopeProvider, Microsoft.Extensions.Logging.Abstractions" })]
         public void SetScopeProvider(IExternalScopeProvider scopeProvider)
         {
-            _scopeProvider = scopeProvider;
```
zacharycmontoya, line 78: Why the change here? Changing the way scopes work seems out of scope for this PR

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Logging/ILogger/DirectSubmission/Formatting/OtlpLogEventBuilder.cs
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        private static (TraceId TraceId, ulong SpanId, int Flags) ExtractTraceContext()
+        {
+            // Prefer Datadog's trace context for accurate correlation with DD spans
+            // Fallback to Activity if Datadog tracing is not active
+            var ddSpan = Tracer.Instance.ActiveScope?.Span as Span;
+            if (ddSpan != null)
+            {
+                return (ddSpan.TraceId128, ddSpan.SpanId, 1);
```
zacharycmontoya: The flags that must be returned are trace context flags. Currently, the only bit that is expected is the last bit, which says whether the trace has been sampled or not. This should be roughly the calculation, with some additional using statements

```suggestion
                return (ddSpan.TraceId128, ddSpan.SpanId, SamplingPriorityValues.IsKeep(ddSpan.Context.SamplingPriority) ? 1 : 0);
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Logging/ILogger/DirectSubmission/DirectSubmissionLoggerProvider.cs
```diff
@@ -40,10 +38,20 @@ internal DirectSubmissionLoggerProvider(
             IExternalScopeProvider? scopeProvider)
         {
             _sink = sink;
-            _formatter = formatter;
             _minimumLogLevel = minimumLogLevel;
             _createLoggerFunc = CreateLoggerImplementation;
-            _scopeProvider = scopeProvider;
+
+#if NETCOREAPP3_1_OR_GREATER
```
zacharycmontoya, line 45: This is kind of hard to follow with the `#endif` trying to come right after the else. Can you do rewrite this into full `#if NETCOREAPP3_1_OR_GREATER`, `#else`, and `#endif` blocks, with code duplication as needed?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Logging/ILogger/DirectSubmission/DirectSubmissionLoggerProvider.cs
zacharycmontoya: Unless there's a really compelling reason, let's leave the `_scopeProvider` field and its usages unchanged, otherwise we may introduce unintentional regressions

link04: I did so based on: https://github.com/DataDog/dd-trace-dotnet/pull/7680/commits/15f3c5dff8efb0b7065c2881fe0e5fd795ef1c99#r2448702004

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Logging/ILogger/DirectSubmission/DirectSubmissionLoggerProvider.cs
```diff
@@ -40,10 +39,19 @@ internal DirectSubmissionLoggerProvider(
...
+            if (sink is OtlpSubmissionLogSink)
+            {
+                _logEventCreator = new OtelLogEventCreator();
```
zacharycmontoya, line 48: Last thing, can you handle the scope provider the same way that you're handling it in our standard Sink? Unless you have a specific reason for not propagating scope changes, let's keep the handling consistent

```suggestion
                _logEventCreator = new OtelLogEventCreator(scopeProvider);
```

zacharycmontoya: Oh I guess we're not using the scopeProvider? It seems like something we may need to do in the future, but for now I suppose we can leave this as is. We should definitely test how this is handled when using the OpenTelemetry SDK and its OTLP exporter, but for now we can leave this as out-of-scope

link04: So currently we don't make use of the scopeProvider, I tried adding it and the logs differed from OTEL so I want to connect on how we should use that data and the effect on OTEL customers.

tracer/src/Datadog.Trace/OpenTelemetry/Logs/OtlpExporter.cs
andrewlock: Yikes, good catch 😱 Yeah we should definitely remove that sync-over-async!

PR #7681, opened 2025-10-21
[Profiler] Generate heap snapshot

chrisnas:
## Summary of changes
Allow the generation of heap snapshots that are sent to the backend

## Reason for change
For memory leak workflow, we would like to compare the surviving objects count to identify the leaky ones. These heap snapshots will be transformed into class histograms to look like what is generated by Java.

## Implementation details
The .NET runtime allows the creation of "gcdump" thanks to dotnet-gcdump tool. Behind the scene, an induced GC is triggered when an event pipe session is created with the right keyword/verbosity. Undocumented events are then emitted by the runtime at the end of the GC - read https://chnasarre.medium.com/net-gcdump-internals-fcce5d327be7?source=friends_link&sk=3225ff119458adafc0e6935951fcc323 for more details.

## Test coverage
Test added as needed

### DIFF
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native/NativeThreadList.cpp
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native/NativeThreadList.cpp
@@ -45,5 +45,7 @@ bool NativeThreadList::RegisterThread(uint32_t tid)
 bool NativeThreadList::Contains(uint32_t tid) const
 {
     std::lock_guard<std::mutex> lock(_mutex);
-    return _nativeThreadIds.contains(tid);
+
+    // NOTE: contains() does not seem to be available when compiled on ARM
+    return _nativeThreadIds.find(tid) != _nativeThreadIds.end();
 }


chrisnas:
> This is just configuration, right? There's no implementation of taking the snapshot here?

Work in progress yes  :^)
My last commit seems to generate a "class histogram" as expected by our backend when conditions are fullfiled

profiler/src/ProfilerEngine/Datadog.Profiler.Native/HeapSnapshotManager.h
```diff
@@ -0,0 +1,183 @@
...
+    std::shared_ptr<ProxyMetric> _heapSnapshotTotalSizeMetric;
+
+    ICorProfilerInfo12* _pCorProfilerInfo;
+    IFrameStore* _pFrameStore;
+    IThreadsCpuManager* _pThreadsCpuManager;
+    INativeThreadList* _pNativeThreadList;
+
+    std::unique_ptr<std::thread> _pLoopThread;
+    DWORD _loopThreadOsId;
+    volatile bool _shutdownRequested = false;
```
gleocadie, line 148: use `std::atomic` instead. 
`volatile` does not guarantee atomicity, or memory ordering (not like in C#)

profiler/src/ProfilerEngine/Datadog.Profiler.Native/HeapSnapshotManager.cpp
```diff
@@ -0,0 +1,483 @@
...
+        return static_cast<double>(_totalSize);
+    });
+
+
+    _pCorProfilerInfo->AddRef();
+}
+
+HeapSnapshotManager::~HeapSnapshotManager()
+{
+    StopImpl();
```
gleocadie, line 67: shouldn't we call `Stop` instead of `StopImpl`.
`Stop` will modify the state of the service but not `StopImpl`. 
In this case it's not big, because it lives as long as the process lives, but just to be clean

profiler/src/ProfilerEngine/Datadog.Profiler.Native/HeapSnapshotManager.cpp
```diff
@@ -0,0 +1,483 @@
...
+{
+    _runtimeSessionKeywords = keywords;
+    _runtimeSessionVerbosity = verbosity;
+}
+
+bool HeapSnapshotManager::StartImpl()
+{
+    _pLoopThread = std::make_unique<std::thread>([this] {
+        OpSysTools::SetNativeThreadName(ThreadName);
+        MainLoop();
```
gleocadie, line 87: not in this PR, but I wonder if we should not check if the thread is correctly started before returning `true`.
If the thread failed to start, we should return false.
This should be done in all other classes that have the same pattern.

profiler/src/ProfilerEngine/Datadog.Profiler.Native/HeapSnapshotManager.cpp
```diff
@@ -0,0 +1,483 @@
...
+    if (_shouldStartHeapDump)
+    {
+        _shouldStartHeapDump = false;
+        StartGCDump();
+    }
+}
+
+std::string HeapSnapshotManager::GetAndClearHeapSnapshotText()
+{
+    // TODO: this should be protected by a lock because both the dedicated thread and the exporter thread
```
gleocadie: yes

profiler/src/ProfilerEngine/Datadog.Profiler.Native/HeapSnapshotManager.cpp
```diff
@@ -0,0 +1,483 @@
...
+        _totalSize += size;
+
+        // TODO: we should not be called from different threads so no need to lock
+        auto entry = _classHistogram.find(pNodes[i].TypeID);
+        if (entry == _classHistogram.end())
+        {
+            std::string className;
+            if (_pFrameStore->GetTypeName(static_cast<ClassID>(pNodes[i].TypeID), className))
+            {
+                ClassHistogramEntry histogramEntry(className);
```
gleocadie: you could move `className`

profiler/src/ProfilerEngine/Datadog.Profiler.Native/HeapSnapshotManager.h
```diff
@@ -0,0 +1,183 @@
...
+class IThreadsCpuManager;
+class INativeThreadList;
+
+using namespace std::chrono_literals;
+
+
+class ClassHistogramEntry
+{
+public:
+    ClassHistogramEntry(std::string& className)
```
gleocadie: use `std::string` instead of `std::string&`, you could directly move `className`  into `ClassName`

profiler/src/ProfilerEngine/Datadog.Profiler.Native/HeapSnapshotManager.cpp
```diff
@@ -0,0 +1,483 @@
...
+void HeapSnapshotManager::MainLoopIteration()
+{
+    if (_shouldCleanupHeapDumpSession)
+    {
+        // close the session + start/stop the fake session to reset the keywords/verbosity
+        _shouldCleanupHeapDumpSession = false;
+        CleanupSession();
+    }
+    else
+    if (_shouldStartHeapDump)
```
gleocadie: `_shouldStartHeadDump`  should be `std::atomic<bool>` since it can be modified/read by multiple threads

profiler/src/ProfilerEngine/Datadog.Profiler.Native/HeapSnapshotManager.cpp
```diff
@@ -0,0 +1,483 @@
...
+    std::cout << "OnBulkNodes #" << index << " x" << count  << std::endl;
+#endif
+
+    _objectCount += count;
+    for (size_t i = 0; i < count; i++)
+    {
+        auto size = pNodes[i].Size;
+        _totalSize += size;
+
+        // TODO: we should not be called from different threads so no need to lock
```
gleocadie: guaranteed ?

profiler/src/ProfilerEngine/Datadog.Profiler.Native/HeapSnapshotManager.cpp
```diff
@@ -0,0 +1,483 @@
...
+            Log::Error("Unknown Exception in HeapSnapshotManager::MainLoop.");
+        }
+    }
+
+    Log::Debug("HeapSnapshotManager::MainLoop has ended.");
+}
+
+void HeapSnapshotManager::MainLoopIteration()
+{
+    if (_shouldCleanupHeapDumpSession)
```
gleocadie: `_shouldCleanupHeapDumpSession`  should be a `std::atomic<bool>`

profiler/src/ProfilerEngine/Datadog.Profiler.Native/HeapSnapshotManager.cpp
```diff
@@ -0,0 +1,483 @@
...
+void HeapSnapshotManager::OnGarbageCollectionStart(
+    std::chrono::nanoseconds timestamp,
+    int32_t number,
+    uint32_t generation,
+    GCReason reason,
+    GCType type)
+{
+    // waiting for the first induced foreground gen2 collection corresponding to the one
+    // triggered by the session creation
+    if (_isHeapDumpInProgress && (_inducedGCNumber == -1))
```
gleocadie: `_isHeapDumpInProgress` should be `std::atomic<bool>`

gleocadie: I suspect `OnGarbageCollectionStart` and `OnGarbageCollectionEnd` being called by the same thread, so `_inducedGCNumber` can be `std::int32_t` 
But if those functions can be called from different threads, `_inducedGCNumber` should be `std::atomic<std::int32_t>`

profiler/src/ProfilerEngine/Datadog.Profiler.Native/IHeapSnapshotManager.h
```diff
@@ -0,0 +1,14 @@
...
+#pragma once
+
+#include <string>
+
+class IHeapSnapshotManager
+{
+public:
+    virtual void SetRuntimeSessionParameters(uint64_t keywords, uint32_t verbosity) = 0;
+    virtual std::string GetAndClearHeapSnapshotText() = 0;
+    virtual ~IHeapSnapshotManager() = default;
```
gleocadie: nit: can you move the dtor just after `public:` please

profiler/src/ProfilerEngine/Datadog.Profiler.Native/INativeThreadList.h
```diff
@@ -0,0 +1,13 @@
...
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2022 Datadog, Inc.
+
+#pragma once
+
+#include <cstdint>
+
+class INativeThreadList
+{
+public:
+    virtual bool RegisterThread(uint32_t tid) = 0;
```
gleocadie, line 13: add a virtual dtor with ` = default`

profiler/src/ProfilerEngine/Datadog.Profiler.Native/NativeThreadList.cpp
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    Log::Debug("NativeThreadList: thread 0x", std::hex, tid, std::dec, " was already registered.");
+    return false;
+}
+
+bool NativeThreadList::Contains(uint32_t tid) const
+{
+    std::lock_guard<std::mutex> lock(_mutex);
+    return _nativeThreadIds.find(tid) != _nativeThreadIds.end();
```
gleocadie: you may be able to use the `contains`  function instead

profiler/src/ProfilerEngine/Datadog.Profiler.Native/NativeThreadList.h
```diff
@@ -0,0 +1,42 @@
...
+private:
+    NativeThreadList(const NativeThreadList&) = delete;
+    NativeThreadList& operator=(const NativeThreadList&) = delete;
+
+public:
+    // Inherited via IService
+    const char* GetName() override;
+
+    // Inherited via INativeThreadList
+    bool RegisterThread(uint32_t tid);
```
gleocadie: add `override` for `RegisterThread`

profiler/src/ProfilerEngine/Datadog.Profiler.Native/ProfileExporter.cpp
```diff
@@ -700,6 +723,26 @@ std::string ProfileExporter::CreateMetricsFileContent() const
...
+    std::stringstream builder;
+
+    // TODO: just for tests, the heap snapshot manager returns a string instead of a reference to an unorderedmap
+    auto heapSnapshot = _heapSnapshotManager->GetAndClearHeapSnapshotText();
+    if (!heapSnapshot.empty())
+    {
+        builder << heapSnapshot;
+    }
+
+    return builder.str();
```
gleocadie: nit: you rewrite this so the `builder` is created only if `heapSnapshot` is not empty()

profiler/test/Datadog.Profiler.IntegrationTests/Helpers/HeapSnapshotHelper.cs
```diff
@@ -0,0 +1,28 @@
+using System.Linq;
```
gleocadie, line 6: can you add the header pls

profiler/src/ProfilerEngine/Datadog.Profiler.Native/HeapSnapshotManager.h
chrisnas: Interesting... I've used the exact same definition as what is done in StackSamplerLoop
Maybe we should change in both places

profiler/src/ProfilerEngine/Datadog.Profiler.Native/HeapSnapshotManager.cpp
chrisnas: I'm not sure because the destructor is the last chance of cleanup and the service lifetime stop should have been called before. I've followed what the StackSamplerLoop is doing because they both have created threads that need to be stopped.

profiler/src/ProfilerEngine/Datadog.Profiler.Native/HeapSnapshotManager.cpp
chrisnas: Interesting question... Not sure to find a reason why a thread would not be able to be started by the OS.
Let's discuss that pattern and decide what to do in another PR

profiler/src/ProfilerEngine/Datadog.Profiler.Native/HeapSnapshotManager.cpp
chrisnas: ✅ done

profiler/src/ProfilerEngine/Datadog.Profiler.Native/HeapSnapshotManager.cpp
chrisnas: ✅ done

profiler/src/ProfilerEngine/Datadog.Profiler.Native/HeapSnapshotManager.cpp
chrisnas: ✅ updated to run under a lock

profiler/src/ProfilerEngine/Datadog.Profiler.Native/HeapSnapshotManager.cpp
chrisnas: ✅ done

profiler/src/ProfilerEngine/Datadog.Profiler.Native/HeapSnapshotManager.cpp
chrisnas: ✅ done

profiler/src/ProfilerEngine/Datadog.Profiler.Native/HeapSnapshotManager.cpp
chrisnas: ✅ done

chrisnas: ✅ done

profiler/src/ProfilerEngine/Datadog.Profiler.Native/IHeapSnapshotManager.h
chrisnas: ✅ done

profiler/src/ProfilerEngine/Datadog.Profiler.Native/INativeThreadList.h
chrisnas: ✅ done

profiler/src/ProfilerEngine/Datadog.Profiler.Native/NativeThreadList.cpp
chrisnas: ✅ done

profiler/src/ProfilerEngine/Datadog.Profiler.Native/NativeThreadList.h
chrisnas: ✅ done

profiler/src/ProfilerEngine/Datadog.Profiler.Native/ProfileExporter.cpp
chrisnas: ✅ done
I'm not 100% sure of this part where the manager has to know how to serialize the class histogram instead of letting the exporter be responsible for the format. Looks good for a preview

profiler/test/Datadog.Profiler.IntegrationTests/Helpers/HeapSnapshotHelper.cs
chrisnas: ✅ good catch

PR #7686, opened 2025-10-21
Run dsm-throughput macrobenchmarks on master push & schedule

robcarlan-datadog:
## Summary of changes
One of our pipeline scripts started failing because we are failing to check out a tag. We should run dsm-throughput in the same manner as existing macrobenchmarks.

## Reason for change
Set up scripts was failing to check out a tag, because we ran this on a tag pipeline in the release.


bouwkast (approved):
Thanks!

PR #7687, opened 2025-10-21
[tracing] add support for `DiagnosticSource` (and Quartz) in .NET Framework

chojomok:
## Summary of changes                                                                                                                                                           
  Adds `.NET Framework` support for the `DiagnosticManager` / `DiagnosticObserver` infrastructure and uses it to enhance Quartz span metadata on both Framework and non-Framework
  targets.                                                                                                                                                                        
                                                                                                                                                                                  
  ## Reason for change                                                                                                                                                            
  The `DiagnosticObserver` class was previously gated behind `#if !NETFRAMEWORK`, limiting tracing integrations that depend on it to modern .NET only. Quartz is the first   
   
  ## Implementation details                                                                                                                                                       
                                                                       
  ### DiagnosticManager — Framework support
  On `!NETFRAMEWORK`, the existing path is unchanged: `DiagnosticListener.AllListeners.Subscribe(new DiagnosticListenerObserver(this))`.
                                                                                                                                                                                  
  On `NETFRAMEWORK`, `DiagnosticListener` isn't available at compile time (it ships as a NuGet package, not part of the BCL). Two problems had to be solved:
                                                                                                                                                                                  
  1. **Accessing the static `AllListeners` property** — solved with reflection to locate the `DiagnosticListener` type and read its static property at runtime.                   
  2. **Generic invariance** — `AllListeners.Subscribe` expects `IObserver<DiagnosticListener>` (the real type). A new `FrameworkDiagnosticListenerObserver` reverse duck type
  (`[DuckReverseMethod]` + `.DuckImplement()`) generates a proxy at runtime that satisfies the exact generic interface, forwarding each `OnNext` call back to `DiagnosticManager`.
                                                                       
  ### DiagnosticObserver — duck typed DiagnosticListener                                                                                                                          
  `SubscribeIfMatch` was updated to accept `IDiagnosticListener` (a new duck type interface) instead of the concrete `DiagnosticListener`, removing the compile-time dependency on
   the type across the whole observer hierarchy.                                                                                                                                  
   
  ### Quartz                                                                                                                                                                      
  - `QuartzDiagnosticObserver` is registered in `StartDiagnosticManager` unconditionally (both platforms).
  - `QuartzCommon` was enhanced to set the `component: quartz` tag and activity kind on the pre-`IActivity5` fallback path (< .NET 5), bringing its span metadata in line with
  newer runtimes.                                                                                                                                                                 
   
  ### Cleanup                                                                                                                                                                     
  - Consolidated the duplicate `#if !NETFRAMEWORK` / `#else` split of `StartDiagnosticManager` in `Instrumentation.cs` into a single method with an inline directive.
  - Removed an unnecessary `#if NETFRAMEWORK` block in `ActivityListener.cs` that duplicated `CreateDiagnosticSourceListenerInstance` — the IL emit approach works on Framework   
  too.                                                                                                                                                                            
                                                                                                                                                                                  
  ## Test coverage                                                                                                                                                                
  - New snapshot `QuartzTestsV3NETFRAMEWORK.verified.txt` covering Quartz on .NET Framework.
  - Updated `QuartzTestsV3NETCOREAPP3X.verified.txt` to include the `component: quartz` tag now set on the pre-`IActivity5` path.
  
                                                                                                                                                                                    
## Other details                                                                                                                                                                  
<!-- Fixes #{issue} -->                                                                                                                                                           
                                                                                                                                                                                  
                                                                                                                                                                                  
<!--  ⚠️  Note:                                                                                                                                                                    
                                                                                                                                                                                  
Where possible, please obtain 2 approvals prior to merging. Unless CODEOWNERS specifies otherwise, for external teams it is typically best to have one review from a team member, 
 and one review from apm-dotnet. Trivial changes do not require 2 reviews.                                                                                                        
                                                                                                                                                                                  
MergeQueue is NOT enabled in this repository. If you have write access to the repo, the PR has 1-2 approvals (see above), and all of the required checks have passed, you can use 
 the Squash and Merge button to merge the PR. If you don't have write access, or you need help, reach out in the #apm-dotnet channel in Slack.                                    
-->

### DIFF
--- a/.azure-pipelines/ultimate-pipeline.yml
+++ b/.azure-pipelines/ultimate-pipeline.yml
@@ -146,7 +146,7 @@ variables:
   DD_LOGGER_DD_TAGS: test.configuration.job:$(System.JobDisplayName)
   DD_LOGGER_ENABLED: true
   DD_COLLECTOR_CPU_USAGE: true
-  ToolVersion: 3.40.0
+  ToolVersion: 3.41.0
   # .NET SDK performance optimization variables
   DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1
   DOTNET_CLI_TELEMETRY_OPTOUT: 1
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -42,6 +42,87 @@
 
 
 
+
+## [Release 3.40.0](https://github.com/DataDog/dd-trace-dotnet/releases/tag/v3.40.0)
+
+## Summary
+
+* [Runtime Metrics] Enables Runtime Metrics by Default for .NET 6+ (#8267)
+* [Tracing] Add interoperability between the Datadog Baggage API and the OpenTelemetry Baggage API (#7921)
+* Add support for Microsoft.Data.SqlClient 7.x.x (#8346)
+
+## Changes
+
+### Tracer
+* Refactor log directory resolution into smaller methods, add tests (#7255)
+* Adjust Log levels in OltpExporter (#8133)
+* [Metrics] Fixing sync-over-async patterns in DogStatsD Client Fork (#8265)
+* Add _dd.svc_src meta tag to track service name source (#8274)
+* Add process tags to service discovery (#8295)
+* Add "manual" service source (#8302)
+* ServiceNameSource for inferred proxy (#8309)
+
+### CI Visibility
+* [CI Visibility] honor EFD retry mode over ATR (#8288)
+
+### Continuous Profiler
+* [Profiler] Add unit tests for ManagedCodeCache (#8307)
+
+### Debugger
+* [Dynamic Instrumentation] DEBUG-5317 fix snapshot stackoverflow in collection (#8310)
+* [Dynamic Instrumentation] Add detailed diagnostics for line probe resolution (#8316)
+
+### Fixes
+* More remote config performance improvements (#8241)
+* WCF cleanup and trace context extraction (#8263)
+
+### Data Streams Monitoring
+* Support adding kafka_cluster_id for Confluent.Kafka (#7702)
+* [DSM] Include container tags hash in node hash calculation (#8163)
+
+### Miscellaneous
+* [Tracing] Add interoperability between the Datadog Baggage API and the OpenTelemetry Baggage API (#7921)
+* [Process Tags] Explicit whether the service name is the default one (#8106)
+* Add an `IArrayPool<char>` implementation for vendored Newtonsoft.JSON (#8228)
+* Use `IArrayPool` in more places for JSON (#8236)
+* Remote configuration tests and performance improvements (#8237)
+* Reduce payload size sent by agent in remote config (#8238)
+* [Agent Skill] Rename to `analyze-azdo-build`, add retry support, and more (#8247)
+* [Metrics] Send Diagnostics GC Pause Time as Counter instead of Timer (#8266)
+* [Runtime Metrics] Enables Runtime Metrics by Default for .NET 6+ (#8267)
+* Rename telemetry `GetData` methods to `GetIncrementalData` (#8269)
+* Remove unnecessary use of duck typing in HTTP integrations (#8275)
+* Exclude Otel for MongoDb 3.7 (#8279)
+* Detect whether an app has been trimmed (#8281)
+* [Agent Docs] Clean up `AGENTS.md`: fix stale reference, consolidate structure (#8290)
+* Add some documentation on running system-tests (#8298)
+* Update local smoke testing docs (#8311)
+* [Tracer] Add Component Tag to Quartz Instrumentation (#8315)
+* [AI Agents] Remove plugin settings from shared Claude Code config (#8343)
+* Add support for Microsoft.Data.SqlClient 7.x.x (#8346)
+* Attempt to fix macOS runner native build (#8348)
+
+### Build / Test
+* (CI) Migrating benchmark-serverless CI job to short-lived tokens (#8219)
+* Update Wix to 5.x.x (#8268)
+* Allow generating a snapshot of the MSI contents (#8270)
+* Move running of smoke tests to Nuke (#8271)
+* Enable performance analyzers for Datadog.Trace (#8276)
+* [Test Package Versions Bump] Updating package versions (#8278)
+* Retry ChromeDriver startup in Selenium tests (#8284)
+* Fix Datadog.Trace.Annotations dependency version in Datadog.AzureFunctions nuspec (#8285)
+* Ban SetVersion/SetVersionPrefix/SetVersionSuffix in Nuke Build (#8289)
+* Only allow creating hotfix branches from a tag (#8297)
+* Bump test dependencies and add test.final_status to benchmark exporter (#8301)
+* Update `FeatureFlagsEvaluator` and Debugger `ToUUID` to not violate `CA1850` (#8303)
+* [Test Package Versions Bump] Updating package versions (#8306)
+* Relax concurrent TagsList stress test (#8317)
+* Replace our custom GitHub App Token with `dd-octo-sts` (#8318)
+* Fix `GeneratePackageVersions` so that it correctly respects `IncludePackages` and `ExcludePackages` (#8340)
+
+
+[Changes since 3.39.0](https://github.com/DataDog/dd-trace-dotnet/compare/v3.39.0...v3.40.0)
+
 ## [Release 3.39.0](https://github.com/DataDog/dd-trace-dotnet/releases/tag/v3.39.0)
 
 ## Summary
--- a/profiler/src/ProfilerEngine/Datadog.Linux.ApiWrapper/CMakeLists.txt
+++ b/profiler/src/ProfilerEngine/Datadog.Linux.ApiWrapper/CMakeLists.txt
@@ -2,7 +2,7 @@
 # Project definition
 # ******************************************************
 
-project("Datadog.Linux.ApiWrapper" VERSION 3.40.0)
+project("Datadog.Linux.ApiWrapper" VERSION 3.41.0)
 
 # ******************************************************
 # Compiler options
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/CMakeLists.txt
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/CMakeLists.txt
@@ -2,7 +2,7 @@
 # Project definition
 # ******************************************************
 
-project("Datadog.Profiler.Native.Linux" VERSION 3.40.0)
+project("Datadog.Profiler.Native.Linux" VERSION 3.41.0)
 
 option(RUN_ASAN "Build with Clang Undefined-Behavior Sanitizer" OFF)
 option(RUN_UBSAN "Build with Clang Undefined-Behavior Sanitizer" OFF)
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native.Windows/Resource.rc
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native.Windows/Resource.rc
@@ -62,8 +62,8 @@ END
 
 // ------- version info -------------------------------------------------------
 VS_VERSION_INFO VERSIONINFO
-FILEVERSION             3,40,0,0
-PRODUCTVERSION          3,40,0,0
+FILEVERSION             3,41,0,0
+PRODUCTVERSION          3,41,0,0
 FILEFLAGSMASK           VS_FF_PRERELEASE
 FILEOS                  VOS_NT
 FILETYPE                VFT_DLL
@@ -74,12 +74,12 @@ BEGIN
         BEGIN
             VALUE "CompanyName", "Datadog"
             VALUE "FileDescription", "Continuous Profiler for .NET Applications"
-            VALUE "FileVersion", "3.40.0.0"
+            VALUE "FileVersion", "3.41.0.0"
             VALUE "InternalName", "Native Profiler Engine"
             VALUE "LegalCopyright", "(c) Datadog 2020-2022"
             VALUE "OriginalFilename", "Datadog.Profiler.Native.dll"
             VALUE "ProductName", "Continuous Profiler for .NET Applications"
-            VALUE "ProductVersion", "3.40.0.0"
+            VALUE "ProductVersion", "3.41.0.0"
         END
     END
     BLOCK "VarFileInfo"
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native/dd_profiler_version.h
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native/dd_profiler_version.h
@@ -3,4 +3,4 @@
 
 #pragma once
 
-constexpr auto PROFILER_VERSION = "3.40.0";
+constexpr auto PROFILER_VERSION = "3.41.0";
--- a/profiler/src/ProfilerEngine/ProductVersion.props
+++ b/profiler/src/ProfilerEngine/ProductVersion.props
@@ -5,7 +5,7 @@
 
   <!-- * * * * * * * * * * * INPUTS. Update this section EVERY time the component is shipped/released! * * * * * * * * * * *                -->
   <PropertyGroup>
-    <ProductVersion>3.40.0</ProductVersion>
+    <ProductVersion>3.41.0</ProductVersion>
   </PropertyGroup>
   <!-- * * * * * * * * * * * END OF INPUTS.  * * * * * * * * * * *                                                                          -->
 
--- a/shared/src/Datadog.Trace.ClrProfiler.Native/CMakeLists.txt
+++ b/shared/src/Datadog.Trace.ClrProfiler.Native/CMakeLists.txt
@@ -12,7 +12,7 @@ endif()
 # Project definition
 # ******************************************************
 
-project("Datadog.Trace.ClrProfiler.Native" VERSION 3.40.0)
+project("Datadog.Trace.ClrProfiler.Native" VERSION 3.41.0)
 
 if (UNIVERSAL)
    find_package(GlibcCompat REQUIRED)
--- a/shared/src/Datadog.Trace.ClrProfiler.Native/Resource.rc
+++ b/shared/src/Datadog.Trace.ClrProfiler.Native/Resource.rc
@@ -57,8 +57,8 @@ VS_VERSION_INFO VERSIONINFO
 #else
  FILEFLAGS 0x0L
 #endif
- FILEVERSION             3,40,0,0
- PRODUCTVERSION          3,40,0,0
+ FILEVERSION             3,41,0,0
+ PRODUCTVERSION          3,41,0,0
  FILEOS                  VOS_NT
  FILETYPE                VFT_DLL
 BEGIN
@@ -68,12 +68,12 @@ BEGIN
         BEGIN
             VALUE "CompanyName", "Datadog"
             VALUE "FileDescription", "Native loader for Datadog .NET APM"
-            VALUE "FileVersion", "3.40.0.0"
+            VALUE "FileVersion", "3.41.0.0"
             VALUE "InternalName", "Native loader"
             VALUE "LegalCopyright", "(c) Datadog 2020-2022"
             VALUE "OriginalFilename", "Datadog.Trace.ClrProfiler.Native.dll"
             VALUE "ProductName", "Native loader for Datadog .NET APM"
-            VALUE "ProductVersion", "3.40.0.0"
+            VALUE "ProductVersion", "3.41.0.0"
         END
     END
     BLOCK "VarFileInfo"
--- a/shared/src/msi-installer/WindowsInstaller.wixproj
+++ b/shared/src/msi-installer/WindowsInstaller.wixproj
@@ -7,9 +7,9 @@
     <MsiOutputPath Condition="'$(MsiOutputPath)' == ''">bin\$(Configuration)\$(Platform)\</MsiOutputPath>
     <OutputPath>$(MsiOutputPath)</OutputPath>
     <IntermediateOutputPath>obj\$(Configuration)\$(Platform)\</IntermediateOutputPath>
-    <OutputName>datadog-dotnet-apm-3.40.0-$(Platform)</OutputName> <!-- -The regex recognizes this line -->
+    <OutputName>datadog-dotnet-apm-3.41.0-$(Platform)</OutputName> <!-- -The regex recognizes this line -->
     <MonitoringHomeDirectory Condition="'$(MonitoringHomeDirectory)' == ''">$(MSBuildThisFileDirectory)..\..\bin\monitoring-home</MonitoringHomeDirectory>
-    <DefineConstants>InstallerVersion=3.40.0;MonitoringHomeDirectory=$(MonitoringHomeDirectory);</DefineConstants>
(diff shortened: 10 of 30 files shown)


tracer/src/Datadog.Trace/DiagnosticListeners/DiagnosticManager.cs
```diff
@@ -3,23 +3,27 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
 // </copyright>
 
-#if !NETFRAMEWORK
```
bouwkast, line 6: Hmm I think this is introducing `System.Diagnostics.DiagnosticSource` as a reference which I don't think we should do.

chojomok: System.Diagnostics.DiagnosticSource isn't actually being used. We're using reflection to access it.

tracer/src/Datadog.Trace/DiagnosticListeners/DiagnosticListenerObserverFactory.cs
```diff
@@ -0,0 +1,168 @@
...
+using System;
+using System.Reflection;
+using System.Reflection.Emit;
+using Datadog.Trace.DuckTyping;
+using Datadog.Trace.Logging;
+
+namespace Datadog.Trace.DiagnosticListeners
+{
+    /// <summary>
+    /// This is code written by Cursor to do the reflection needed for DiagnosticListener.
```
lucaspimentel: Not a blocker, but I don't think we need to starting adding comments everywhere we generated code with AI assistance.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Quartz/QuartzCommon.cs
```diff
@@ -41,7 +41,7 @@ internal static void SetActivityKind(IActivity5 activity)
             ActivityListener.SetActivityKind(activity, GetActivityKind(activity));
         }
 
-        internal static void EnhanceActivityMetadata(IActivity5 activity)
+        internal static void EnhanceActivityMetadata5(IActivity5 activity)
```
lucaspimentel: I would call this one `EnhanceActivity5Metadata()` since it's enhancing the metadata on an `IActivity5`.

But also, does it really need to be renamed, or can we just have two overloads (same name, different parameters)? (not a big deal either way)
```csharp
EnhanceActivityMetadata(IActivity5 activity)
EnhanceActivityMetadata(IActivity activity)
```

chojomok: I like the overload idea!

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Quartz/QuartzCommon.cs
```diff
@@ -67,5 +83,47 @@ internal static void AddException(object exceptionArg, IActivity activity)
...
+                case "Quartz.Job.Execute.Start":
+                case "Quartz.Job.Veto.Start":
+                    var activity = ActivityListener.GetCurrentActivity();
+                    if (activity is IActivity5 activity5)
+                    {
+                        SetActivityKind(activity5);
+                    }
+                    else
+                    {
+                        Log.Debug("The activity was not Activity5 (Less than .NET 5.0). Unable to set span kind.");
```
lucaspimentel: I don't think `IActivity5` is specific to .NET 5.0 (you can reference a different version of the nuget package), so maybe remove the `(Less than .NET 5.0)` part.

zacharycmontoya: +1. You could note something like "The loaded System.Diagnostics.Activity type does not have a Kind property.".

tracer/test/snapshots/QuartzTestsV3NETFRAMEWORK.verified.txt
zacharycmontoya, line 1: Note for reviewers: The only difference between the `tracer/test/snapshots/QuartzTestsV3NETCOREAPP3X.verified.txt` and `tracer/test/snapshots/QuartzTestsV3NETFRAMEWORK.verified.txt` files that will be checked in is the stack trace content stored in the `error.stack` tag. Everything else is identical

chojomok: Thank you for the call out Zach! 
I've discussed adding some scrubbing so the snapshots can be the same. Please let me know if this something we'd like to do.

tracer/test/snapshots/QuartzTestsV3NETCOREAPP3X.verified.txt
zacharycmontoya, line 1: Note for reviewers: The only difference between the new `tracer/test/snapshots/QuartzTestsV3NETCOREAPP3X.verified.txt` snapshot and the existing `tracer/test/snapshots/QuartzTestsV3.verified.txt` (all newer runtimes running against Quartz v3) is that the NETCOREAPP3X snapshot does _not_ have a `span.kind: internal` tag since System.Diagnostics.Activity does not have a Kind property before .NET 5. The rest of the span contents are identical.

tracer/src/Datadog.Trace/DiagnosticListeners/DiagnosticManager.cs
```diff
@@ -3,23 +3,27 @@
...
+using Datadog.Trace.DiagnosticListeners.DuckTypes;
+using Datadog.Trace.DuckTyping;
 using Datadog.Trace.Logging;
 using Datadog.Trace.Util;
 using Datadog.Trace.Vendors.Serilog.Events;
 
 namespace Datadog.Trace.DiagnosticListeners
 {
-    internal sealed class DiagnosticManager : IDiagnosticManager, IObserver<DiagnosticListener>, IDisposable
+    internal sealed class DiagnosticManager : IDiagnosticManager, IObserver<IDiagnosticListener>, IDisposable
```
zacharycmontoya: I have one refactor I'd like for you to apply: Using reflection you are now creating a type that implements `IObserver<DiagnosticListener>` so we no longer need to implement this interface. We should remove this to reduce confusion.

In addition to this code suggestion, you will also need to:
- Remove `void IObserver<IDiagnosticListener>.OnCompleted()`
- Remove `void IObserver<IDiagnosticListener>.OnError(Exception error)`
- Rename `void IObserver<IDiagnosticListener>.OnNext(IDiagnosticListener listener)` to `public void OnNext(IDiagnosticListener listener)`
- On line 181, change `((IObserver<IDiagnosticListener>)this).OnNext(listener);` to `OnNext(listener);`

```suggestion
    internal sealed class DiagnosticManager : IDiagnosticManager, IDisposable
```

tracer/test/snapshots/QuartzTestsV3NETFRAMEWORK.verified.txt
zacharycmontoya: Up to you! If this test becomes flaky then it will definitely be a priority to scrub it, otherwise I don't think it's necessary for this PR

tracer/src/Datadog.Trace.Trimming/build/Datadog.Trace.Trimming.xml
```diff
@@ -244,7 +244,6 @@
       <type fullname="System.Diagnostics.ActivityIdFormat" />
       <type fullname="System.Diagnostics.ActivitySpanId" />
       <type fullname="System.Diagnostics.ActivityTraceId" />
-      <type fullname="System.Diagnostics.DiagnosticListener" />
```
andrewlock, line 262: This automatic removal is actually a problem, as it means the type could be trimmed from the target application, even though we need it 🤔 Overall, I think that can be "fixed" by not relying on reflection in .NET Core (which i think we shouldn't do anyway)

tracer/src/Datadog.Trace/DiagnosticListeners/DiagnosticListenerObserverFactory.cs
andrewlock: Agreed, but I'm not sure we need this type at all 🤔 As far as I can see we already have almost the exact same code in `ActivityListener` - could/should we unify those?

tracer/src/Datadog.Trace/DiagnosticListeners/DiagnosticManager.cs
```diff
@@ -40,19 +44,76 @@ public void Start()
             if (_allListenersSubscription == null)
             {
                 Log.Debug("Starting DiagnosticListener.AllListeners subscription");
-                _allListenersSubscription = DiagnosticListener.AllListeners.Subscribe(this);
+
+                try
+                {
+                    // Get the DiagnosticListener type
```
andrewlock: I don't think we should be using reflection here in .NET Core + .NET Standard when these types are already available 🤔 Can we `#if` this so that we only take the expensive path in .NET Framework, and we use the optimized version elsewhere?

chojomok: @zacharycmontoya and I worked on adding the directive to use the builtin class instead of doing reflection from !NETFRAMEWORK.

chojomok: I'm going to look into make the factory class more generic so it can be used for ActivityListener.

tracer/src/Datadog.Trace.Trimming/build/Datadog.Trace.Trimming.xml
chojomok: this has been reverted now with adding the directive to use the built in class for !NETFRAMEWORK.

tracer/src/Datadog.Trace/DiagnosticListeners/DiagnosticManager.cs
chojomok: resolving this since we made the changes together.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Quartz/QuartzCommon.cs
chojomok: I updated the debug log, ~~resolving this!~~ - actually a bit more wording is needed for the log to be useful let me rethink it

tracer/test/snapshots/QuartzTestsV3NETFRAMEWORK.verified.txt
chojomok: I'll just leave it as is until we see the test becoming flaky

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Quartz/QuartzCommon.cs
chojomok: @lucaspimentel I added in the overload method

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Quartz/QuartzCommon.cs
```diff
@@ -54,6 +54,24 @@ internal static void EnhanceActivityMetadata(IActivity5 activity)
             activity.DisplayName = CreateResourceName(activity.DisplayName, jobName);
         }
 
+        internal static void EnhanceActivityMetadata(IActivity activity)
+        {
```
lucaspimentel, line 70: Technically, you can still force the compiler to call this method even with an `IActivity5`, so we should probably add a check for that first to be safe. Something like 
```csharp
internal static void EnhanceActivityMetadata(IActivity activity)
{
    if (activity is IActivity5 activity5)
    {
        EnhanceActivityMetadata(activity5);
        return;
    }
...
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Quartz/QuartzCommon.cs
lucaspimentel: We try to avoid using LINQ (`FirstOrDefault()` here). Although it makes the code more concise and readable, it incurs a few heap allocations that we can avoid by iterating manually. Something like:

```csharp
string? jobName = null;

foreach (var tag in activity.Tags)
{
    if (tag.Key == "job.name")
    {
        jobName = tag.Value;
        break;
    }
}
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Quartz/QuartzCommon.cs
```diff
@@ -54,6 +54,24 @@ internal static void EnhanceActivityMetadata(IActivity5 activity)
...
+            if (activity.OperationName is null)
+            {
+                return;
+            }
+
+            activity.AddTag("operation.name", activity.OperationName);
+            var jobName = activity.Tags.FirstOrDefault(kv => kv.Key == "job.name").Value ?? string.Empty;
```
lucaspimentel: Same comment as above about using LINQ (`FirstOrDefault()`).

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Quartz/QuartzCommon.cs
chojomok: I didn't realize it was memory intensive. Makes sense to consider lowering the memory usage of the tracer as much as possible to reduce our overhead.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Quartz/QuartzCommon.cs
chojomok: This is because Activity5 implements Activity right? 
Added this in!

tracer/src/Datadog.Trace/DiagnosticListeners/DiagnosticManager.cs
chojomok: @andrewlock - made a factory class that generates the IL for a static and a instance Subscribe call. There is some reusability in the no-op methods needed in the reflection that I move to methods. 
Please let me know your feedback!

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Quartz/QuartzCommon.cs
lucaspimentel: > This is because Activity5 implements Activity right?

Yup! You can do something like this to force the compiler to call the overload you want:
```csharp
IActivity5 activity5 = GetActivity5();

// stick activity5 into an IActivity variable.
// still the same object, but accessed as if it was IActivity, which is legal because IActivity5 implements IActivity
IActivity activity = (IActivity)activity5;

// this calls the IActivity5 overload
EnhanceActivityMetadata(activity5);

// but this calls the IActivity overload, even though it's the same object.
// the check we added above is in case this happens.
EnhanceActivityMetadata(activity);
```

lucaspimentel: Another way of doing this could have been implementing a single `EnhanceActivityMetadata(IActivity)` (no overloads) and choosing the code path inside:
```csharp
internal static void EnhanceActivityMetadata(IActivity activity)
{
    if (activity is IActivity5 activity5)
    {
        // do IActivity5 things
    }
    else
    {
        // do IActivity things
    }
}
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Quartz/QuartzCommon.cs
lucaspimentel: It's not terrible, but in the tracer every allocation counts!

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Quartz/QuartzCommon.cs
```diff
@@ -44,14 +43,58 @@ internal static void SetActivityKind(IActivity5 activity)
...
+
+            string? jobName = null;
+            foreach (var tag in activity.Tags)
+            {
+                if (tag.Key == "job.name")
+                {
+                    jobName = tag.Value;
+                    break;
+                }
+            }
```
andrewlock: A couple of performance nits here (that I've only found after diving deep on this stuff recently 😅)

- In `IActivity5`, you should use the `TagObjects` property, not `Tags`.
- When enumerating these objects, you should use the new `ActivityEnumerationHelper.EnumerateTagObjects` helper (as we do now in `OtlpHelpers`)

Unfortunately, I made that helper a bit _too_ specific, so I need to tweak it before you can use it here I think 😅  I can look into that in a subsequent PR 🙂

tracer/src/Datadog.Trace/DiagnosticListeners/DiagnosticManager.cs
andrewlock: I'm still struggling to get my head around all this, but I'm still not sure we need the factory, and I feel like we could use a reverse duck type here (which removes a lot of the custom code/dynamic method stuff) using `DuckImplement`?🤔 

e.g. you could implement a simple type that does what you want, which uses duck types. The implementation would look something like this (mostly the same as the .NET Core version)

```csharp
internal sealed class FrameworkDiagnosticListenerObserver
{
    private readonly DiagnosticManager _diagnosticManager;
    public DiagnosticListenerObserver(DiagnosticManager diagnosticManager)
    {
        _diagnosticManager = diagnosticManager;
    }

    public void OnCompleted()
    {
    }

    public void OnError(Exception error)
    {
    }

    
    public void OnNext(IDiagnosticListener value) // 👈 Note using duck type here
    {
        _diagnosticManager.OnDiagnosticListenerNext(value);
    }
}
```

and then you would need to get the type to implement, which is IObserver<DiagnosticListener>. So you would do something like this:

```csharp
// This bit is what you already have:
var diagnosticListenerType = Type.GetType("System.Diagnostics.DiagnosticListener, System.Diagnostics.DiagnosticSource");
if (diagnosticListenerType == null)
{
    Log.Warning("Unable to find DiagnosticListener type");
    return;
}

// This gets the IObserver<DiagnosticListener> type
var iObserverType = typeof(IObserver<>).MakeGenericType(diagnosticListenerType);
```

and _then_ you create the duck type using this:

```csharp
var observer = new FrameworkDiagnosticListenerObserver(this);
var implementation = observer.DuckImplement(iObserverType);
// 👆 this is now implements `IObserver<DiagnosticListener>`
```

And now you have an implementation that you can do what you need to with _or_ the underlying `observer` implementation, depending on what you need? 

To be clear, I haven't stepped through the next bits to see if this is actually useful, but I feel like it's _probably_ doing effectively the same as the `DiagnosticObserverFactory` just with existing tested approaches 😄

chojomok: I'm in favor of using already implemented patterns. I'll try to rewrite and update ya'll with any blockers I come across.

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
```diff
@@ -519,7 +518,19 @@ private static bool SkipAspNetCoreDiagnosticObserver()
             return EnvironmentHelpers.IsAzureFunctions();
         }
 
-#endif // #if !NETFRAMEWORK
+#else
+        private static void StartDiagnosticManager()
```
zacharycmontoya: I think you can consolidate this even further since all the lines of code in this NETFRAMEWORK-only method are already used above in the !NETFRAMEWORK version of `StartDiagnosticManager`

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Quartz/QuartzCommon.cs
chojomok: waited long enough for this to become available! 🕺🏾

chojomok:
Need to update the PR summary as we're doing reverse ducktyping

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
chojomok: todo: test consolidating the code block with https://github.com/DataDog/dd-trace-dotnet/blob/b6dc5f639b3af9baf633f492d7e89fd4b0f26dfc/tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs#L482C1-L498C1

docs/development/TracerDebugging.md
andrewlock, line 1: @lucaspimentel Do we already have this information anywhere? 🤔

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Quartz/QuartzCommon.cs
```diff
@@ -47,14 +46,59 @@ internal static void EnhanceActivityMetadata(IActivity5 activity)
         {
             activity.AddTag("operation.name", activity.DisplayName);
             activity.AddTag(Tags.InstrumentationName, ComponentName);
-            var jobName = activity.Tags.FirstOrDefault(kv => kv.Key == "job.name").Value ?? string.Empty;
+
+            string? jobName = null;
```
andrewlock, line 50: We should probably rework this to be allocation free (the `foreach ` will allocate) but that requires a bunch of refactoring to the `ActivityEnumerationHelper` so I'm happy to take a look in a follow up PR 😄

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
```diff
@@ -519,7 +518,19 @@ private static bool SkipAspNetCoreDiagnosticObserver()
...
+        {
+            var observers = new List<DiagnosticObserver>
```
andrewlock: Braindump:

There's a somewhat interesting knock-on effect of this work... Many integrations change their behavior when there's a diagnostic manager enabled and attached. In particular, this can have performance impacts, and in some cases change real behaviour.

I was considering whether we should check if the quartz integration is actually enabled before enabling this, but given that we can toggle that at runtime, _and_ there's a global switch for disabling the diagnostic source integration entirely, I don't think there's anything more we need to do.

andrewlock (approved):
Great work! 🎉

docs/development/TracerDebugging.md
lucaspimentel: I don't think so. At least not so explicitly.

docs/development/TracerDebugging.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Optional Variables
+- `DD_TRACE_DEBUG=1` - Enable verbose debug logging
+- `DD_TRACE_LOG_DIRECTORY=/path/to/logs` - Log output directory
+
+## Related Documentation
+
+- [AutomaticInstrumentation.md](AutomaticInstrumentation.md) - Creating and testing integrations
+- [DuckTyping.md](DuckTyping.md) - Duck typing patterns for instrumentation
+- [../CONTRIBUTING.md](../CONTRIBUTING.md) - General contribution guidelines
```
lucaspimentel: I don't think we need this one.
```suggestion
```

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
chojomok: done

PR #7695, opened 2025-10-22
Create `SettingsManager` for managing mutable settings and ExporterSettings

andrewlock:
## Summary of changes

- Introduces `SettingsManager` responsible for managing `MutableSettings` and `ExporterSettings`

## Reason for change

We need to be notified about runtime changes to settings (i.e. config in code or remote config) but don't want to tear down the world and rebuild every time. `SettingsManager` is responsible for handling this. Consumers subscribe to changes and can be notified about updates.

This is a first step which just introduces the type, but doesn't force users to consume changes or remove the current places settings are exposed. Instead, it just encapsulates the changes.

## Implementation details

- Introduce `SettingsManager`
- Move code duplicated in `DynamicConfigurationManager` and `ConfigureIntegration` into `SettingsManager`
- Create a new instance of `SettingsManager` (and maintain it throughout the app lifetime)
- Subscribe to changes one time in `TracerManager` to do the "full rebuild"
  - This is a stop gap before we use it "properly" and stop exposing the settings on `TracerSettings`

## Test coverage

- Mostly a refactor so covered by integration tests
- Unit tests for `SettingsManager` functionality

## Other details

Part of a config stack

- https://github.com/DataDog/dd-trace-dotnet/pull/7695 👈

### DIFF
--- a/tracer/src/Datadog.Trace/Configuration/SettingsManager.cs
+++ b/tracer/src/Datadog.Trace/Configuration/SettingsManager.cs
@@ -24,6 +24,11 @@ public partial record TracerSettings
     {
         private readonly TracerSettings _tracerSettings = tracerSettings;
         private readonly List<SettingChangeSubscription> _subscribers = [];
+
+        private IConfigurationSource _dynamicConfigurationSource = NullConfigurationSource.Instance;
+        private ManualInstrumentationConfigurationSourceBase _manualConfigurationSource =
+            new ManualInstrumentationConfigurationSource(new Dictionary<string, object?>(), useDefaultSources: true);
+
         private SettingChanges? _latest;
 
         /// <summary>
@@ -75,11 +80,40 @@ public partial record TracerSettings
         /// Regenerate the application's new <see cref="MutableSettings"/> and <see cref="ExporterSettings"/>
         /// based on runtime configuration sources.
         /// </summary>
-        /// <param name="dynamicConfigSource">An <see cref="IConfigurationSource"/> for dynamic config via remote config</param>
-        /// <param name="manualSource">An <see cref="IConfigurationSource"/> for manual configuration (in code)</param>
+        /// <param name="manualSource">An <see cref="IConfigurationSource"/> containing the new settings created by manual configuration (in code)</param>
         /// <param name="centralTelemetry">The central <see cref="IConfigurationTelemetry"/> to report config telemetry updates to</param>
         /// <returns>True if changes were detected and consumers were updated, false otherwise</returns>
-        public bool UpdateSettings(
+        public bool UpdateManualConfigurationSettings(
+            ManualInstrumentationConfigurationSourceBase manualSource,
+            IConfigurationTelemetry centralTelemetry)
+        {
+            // we lock this whole method so that we can't conflict with UpdateDynamicConfigurationSettings calls too
+            lock (_subscribers)
+            {
+                _manualConfigurationSource = manualSource;
+                return UpdateSettings(_dynamicConfigurationSource, manualSource, centralTelemetry);
+            }
+        }
+
+        /// <summary>
+        /// Regenerate the application's new <see cref="MutableSettings"/> and <see cref="ExporterSettings"/>
+        /// based on runtime configuration sources.
+        /// </summary>
+        /// <param name="dynamicConfigSource">An <see cref="IConfigurationSource"/> for dynamic config via remote config</param>
+        /// <param name="centralTelemetry">The central <see cref="IConfigurationTelemetry"/> to report config telemetry updates to</param>
+        /// <returns>True if changes were detected and consumers were updated, false otherwise</returns>
+        public bool UpdateDynamicConfigurationSettings(
+            IConfigurationSource dynamicConfigSource,
+            IConfigurationTelemetry centralTelemetry)
+        {
+            lock (_subscribers)
+            {
+                _dynamicConfigurationSource = dynamicConfigSource;
+                return UpdateSettings(dynamicConfigSource, _manualConfigurationSource, centralTelemetry);
+            }
+        }
+
+        private bool UpdateSettings(
             IConfigurationSource dynamicConfigSource,
             ManualInstrumentationConfigurationSourceBase manualSource,
             IConfigurationTelemetry centralTelemetry)
@@ -153,25 +187,17 @@ public partial record TracerSettings
 
         private void NotifySubscribers(SettingChanges settings)
         {
-            // Strictly, for safety, we only need to lock in the subscribers list access. However,
-            // there's nothing to prevent NotifySubscribers being called concurrently,
-            // which could result in weird out-of-order notifications for customers. So for simplicity
-            // we just lock the whole method to ensure serialized updates.
+            _latest = settings;
 
-            lock (_subscribers)
+            foreach (var subscriber in _subscribers)
             {
-                Volatile.Write(ref _latest, settings);
-
-                foreach (var subscriber in _subscribers)
+                try
                 {
-                    try
-                    {
-                        subscriber.Notify(settings);
-                    }
-                    catch (Exception ex)
-                    {
-                        Log.Error(ex, "Error notifying subscriber of MutableSettings change");
-                    }
+                    subscriber.Notify(settings);
+                }
+                catch (Exception ex)
+                {
+                    Log.Error(ex, "Error notifying subscriber of MutableSettings change");
                 }
             }
         }
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/ManualInstrumentation/Tracer/ConfigureIntegration.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/ManualInstrumentation/Tracer/ConfigureIntegration.cs
@@ -73,12 +73,7 @@ public class ConfigureIntegration
                 ? new ManualInstrumentationLegacyConfigurationSource(values, isFromDefaults)
                 : new ManualInstrumentationConfigurationSource(values, isFromDefaults);
 
-        // We need to save this immediately, even if there's no manifest changes in the final settings
-        // so that it can be picked up by other configuration updaters, e.g. remote config
-        GlobalConfigurationSource.UpdateManualConfigurationSource(manualConfig);
-        var dynamicConfig = GlobalConfigurationSource.DynamicConfigurationSource;
-
-        var wasUpdated = Datadog.Trace.Tracer.Instance.Settings.Manager.UpdateSettings(dynamicConfig, manualConfig, TelemetryFactory.Config);
+        var wasUpdated = Datadog.Trace.Tracer.Instance.Settings.Manager.UpdateManualConfigurationSettings(manualConfig, TelemetryFactory.Config);
         if (wasUpdated)
         {
             Log.Information("Setting updates made via configuration in code were applied");
--- a/tracer/src/Datadog.Trace/Configuration/ConfigurationSources/GlobalConfigurationSource.cs
+++ b/tracer/src/Datadog.Trace/Configuration/ConfigurationSources/GlobalConfigurationSource.cs
@@ -22,19 +22,12 @@ namespace Datadog.Trace.Configuration;
 /// </summary>
 internal class GlobalConfigurationSource
 {
-    private static IConfigurationSource _dynamicConfigConfigurationSource = NullConfigurationSource.Instance;
-    private static ManualInstrumentationConfigurationSourceBase _manualConfigurationSource = new ManualInstrumentationConfigurationSource(new Dictionary<string, object?>(), useDefaultSources: true);
-
     /// <summary>
     /// Gets the configuration source instance.
     /// </summary>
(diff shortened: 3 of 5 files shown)


andrewlock:
@codex review

tracer/src/Datadog.Trace/Configuration/SettingsManager.cs
```diff
@@ -0,0 +1,220 @@
...
+            // which could result in weird out-of-order notifications for customers. So for simplicity
+            // we just lock the whole method to ensure serialized updates.
+
+            lock (_subscribers)
+            {
+                var subscribers = _subscribers;
+                Volatile.Write(ref _latest, settings);
+
+                foreach (var subscriber in subscribers)
+                {
```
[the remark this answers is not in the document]
andrewlock: That's not true I think 🤔 `_subscribers` is only accessed in a `lock`, but that wouldn't save us if everything is happening on the same thread. What saves us is that `Trace.Tracer.Configure` _doesn't_ call `tracerSettings.Manager.SubscribeToChanges` more than once. It calls it in the `OneTimeSetup()` method only

andrewlock (commented):
**Issue: Race condition in SettingsManager.SubscribeToChanges**

**Location**: `SettingsManager.cs:56-68`

The `_latest` field is read outside the lock, which can cause double notifications or out-of-order notifications:

```csharp
lock (_subscribers)
{
    _subscribers.Add(subscription);
}

if (Volatile.Read(ref _latest) is { } currentConfig)  // Outside lock!
{
    callback(currentConfig);
}
```

**Scenario**: 
1. Thread A: adds subscriber, unlocks, reads `_latest` (gets value X)
2. Thread B: `NotifySubscribers` locks, writes new value Y to `_latest`, notifies all subscribers including the new one with Y
3. Thread A: notifies with X

Result: The new subscriber gets notified twice (once with Y from Thread B, once with X from Thread A) and out-of-order (Y before X).

**Recommended Fix**: Move the `_latest` check inside the lock:
```csharp
SettingChanges? currentConfig;
lock (_subscribers)
{
    _subscribers.Add(subscription);
    currentConfig = Volatile.Read(ref _latest);
}

if (currentConfig is not null)
{
    try { callback(currentConfig); }
    catch (Exception ex) { /* ... */ }
}
```

---
*Review by Claude Code*

tracer/src/Datadog.Trace/Configuration/SettingsManager.cs
```diff
@@ -0,0 +1,220 @@
...
+            lock (_subscribers)
+            {
+                _subscribers.Add(subscription);
+
+                if (Volatile.Read(ref _latest) is { } currentConfig)
+                {
+                    try
+                    {
+                        // If we already have updates, call this immediately
+                        callback(currentConfig);
```
lucaspimentel, line 67: If a callback in `NotifySubscribers()` tries to call `SubscribeToChanges()`, we'll have a deadlock since they both lock on `_subscribers`. Should we call this callback _outside_ of the lock?

lucaspimentel: Also if the callback is slow, the lock will hold up other threads trying to subscribe.

tracer/src/Datadog.Trace/Configuration/SettingsManager.cs
```diff
@@ -0,0 +1,220 @@
...
+        /// <param name="centralTelemetry">The central <see cref="IConfigurationTelemetry"/> to report config telemetry updates to</param>
+        /// <returns>True if changes were detected and consumers were updated, false otherwise</returns>
+        public bool UpdateSettings(
+            IConfigurationSource dynamicConfigSource,
+            ManualInstrumentationConfigurationSourceBase manualSource,
+            IConfigurationTelemetry centralTelemetry)
+        {
+            if (BuildNewSettings(dynamicConfigSource, manualSource, centralTelemetry) is { } newSettings)
+            {
+                NotifySubscribers(newSettings);
```
lucaspimentel, line 123: If two threads call `UpdateSettings()`, could there be a race condition where `NotifySubscribers()` is also called twice, since both can see the same `_latest` value in `BuildNewSettings()` below?

Ignore if you already considered this and decided that was acceptable.

tracer/src/Datadog.Trace/Configuration/SettingsManager.cs
```diff
@@ -0,0 +1,220 @@
...
+            // there's nothing to prevent NotifySubscribers being called concurrently,
+            // which could result in weird out-of-order notifications for customers. So for simplicity
+            // we just lock the whole method to ensure serialized updates.
+
+            lock (_subscribers)
+            {
+                var subscribers = _subscribers;
+                Volatile.Write(ref _latest, settings);
+
+                foreach (var subscriber in subscribers)
```
lucaspimentel: What's the use of local variable `subscribers` instead of iterating `_subscribers` directly?

tracer/src/Datadog.Trace/Configuration/SettingsManager.cs
```diff
@@ -0,0 +1,220 @@
...
+        /// static sources. To subscribe to updates to these settings, from code or remote config, call <see cref="SubscribeToChanges"/>.
+        /// </summary>
+        public ExporterSettings InitialExporterSettings { get; } = initialExporter;
+
+        /// <summary>
+        /// Subscribe to changes in <see cref="MutableSettings"/> and/or <see cref="ExporterSettings"/>.
+        /// Called whenever these settings change. If the settings have already changed when <see cref="SubscribeToChanges"/>
+        /// is called, <paramref name="callback"/> is invoked immediately with the latest configuration.
+        /// Also note that calling <see cref="SubscribeToChanges"/> twice with the same callback
+        /// will invoke the callback twice.
```
lucaspimentel:
```suggestion
        /// Subscribe to changes in <see cref="MutableSettings"/> and/or <see cref="ExporterSettings"/>.
        /// <paramref name="callback"/> is called whenever these settings change. If the settings have already changed when <see cref="SubscribeToChanges"/>
        /// is called, <paramref name="callback"/> is synchronously invoked immediately with the latest configuration.
        /// Also note that calling <see cref="SubscribeToChanges"/> twice with the same callback
        /// will invoke the callback twice. Callbacks should complete quickly to avoid blocking other operations.
```

tracer/src/Datadog.Trace/Configuration/SettingsManager.cs
andrewlock: > Should we call this callback outside of the lock?

Ironically, this is what I had originally, however AI pointed out that this can lead to a race condition where subscribers don't ever see the latest config (or see an outdated config) if an update happens while they're subscribing, which I considered to be v. bad. 

> If a callback in NotifySubscribers() tries to call SubscribeToChanges(), we'll have a deadlock since they both lock on _subscribers

`lock` is reentrant, so actually this _won't_ block if you call `SubscribeToChanges()` from inside the callback. However, then you're in race condition territory, so we should never do that. I think it's probably ok to just document that in the config, as there's frankly no good reason to do it that I've seen, WDYT?

andrewlock: > Also if the callback is slow, the lock will hold up other threads trying to subscribe.

This is true, but I think it's acceptable personally

tracer/src/Datadog.Trace/Configuration/SettingsManager.cs
andrewlock: I think this is just left over from when I wasn't doing a big lock in Subscribe() 🤔 I think we can just remove it

tracer/src/Datadog.Trace/Configuration/SettingsManager.cs
andrewlock: Hmmm, yeah, good catch 🤔 This could happen if a customer calls `Tracer.Configure()` with manual instrumentation at the same time as remote config does a dynamic config update. I'd always expect two distinct calls in that case, and I think there would be races _elsewhere_, before this is called that I need to 👀 again (despite considering this at the time 😅)

Maybe I should stop trying to be cute and should just lock this whole `UpdateSettings` call... I was a bit concerned about doing that because it's relatively expensive, but I think otherwise we're just going to be fighting edge cases...

lucaspimentel: To be fair, calling `NotifySubscribers()` twice with the same settings is not the worst thing in the world.

tracer/src/Datadog.Trace/Configuration/SettingsManager.cs
lucaspimentel: Wait, I think I meant "is called synchronously" to be later, in the "If the settings have already changed ... callback is invoked immediately" part. Or it's always synchronous?

tracer/src/Datadog.Trace/Configuration/SettingsManager.cs
lucaspimentel: > ok to just document that in the config
> ...
> it's acceptable personally

Sounds good. Just checking 😄

andrewlock: Also, just for prosperity, in the "default" case, the callback won't be called, because `changes` should always be null during the initial subscribe. That assumes that:

- Every service subscribes on initial "app startup"
- We don't start the remote config service until _after_ the initial subscribe

This probably _will_ be true in general, and I'll add some refactorings to make it specifically more likely, so we're mostly catering to edge cases.

tracer/src/Datadog.Trace/Configuration/SettingsManager.cs
andrewlock: Yeah, my (minor) concern is more around a mis-match of settings, if manual and remote config is happening at the same time. I think, currently, in that scenario, you could end up with a weird situation where you get the following:

- Initially: no manual config, no remote config
- Manual config and remote config happens at the same time, with conflicting values
  - Manual updates the "global" manual config source
  - Remote updates the "global" remote config source
  - This happens concurrently, and neither "sees" the update changes
-  Consumers get two updates:
  - One with the new manual config and the _old_ remote config
  - One with the _old_ manual config and the new remote config
  - Neither is correct 😅 

So I think we need some locks around the global config updates... I'll add those in (technically not related to this PR, but makes sense to include here)

tracer/src/Datadog.Trace/Configuration/SettingsManager.cs
andrewlock: Yeah, "called synchronously whenever these settings change" doesn't really make sense I think 😅  The callback _is_ called synchronously, but it's called synchronously from the "updating thread", so it's asynchronous from the PoV of the person calling `SubscribeToChanges()`.

Like you say, "the settings have already changed ... callback is invoked immediately" _is_ called synchronously, so I've tweaked the wording

tracer/src/Datadog.Trace/Configuration/SettingsManager.cs
andrewlock:
```suggestion
                Volatile.Write(ref _latest, settings);

                foreach (var subscriber in _subscribers)
```

tracer/src/Datadog.Trace/Configuration/SettingsManager.cs
andrewlock: I reworked this quite a lot:

- Instead of storing the manual and dynamic config sources as static globals, store them on `SettingsManager`
  - This is practically equivalent, because we only create a single `TracerSettings` instance in production, but it makes things much easier to reason about and manage
- Have specific methods for "update manual" and "update dynamic config". That way we can lock inside these methods and avoid the specific scenario I described above
- Locking  around the whole update process avoids any future edge cases we might hit. They're very much _edge_ cases, but given these updates should be relatively rare, I think playing it safest makes sense. There's not much to be gained by trying to reduce how long the lock is held.

Thanks for the 👀!

tracer/src/Datadog.Trace/Configuration/SettingsManager.cs
```diff
@@ -0,0 +1,249 @@
...
+
+            return new SettingChanges(updatedMutableSettings, updatedExporterSettings, currentMutable, currentExporter);
+        }
+
+        private void NotifySubscribers(SettingChanges settings)
+        {
+            // Strictly, for safety, we only need to lock in the subscribers list access. However,
+            // there's nothing to prevent NotifySubscribers being called concurrently,
+            // which could result in weird out-of-order notifications for customers. So for simplicity
+            // we just lock the whole method to ensure serialized updates.
```
bouwkast: `So for simplicity we just lock the whole method to ensure serialized updates.`

Is that referring to this function as there isn't a lock here?

andrewlock: 😅 There _was_ a lock here 😅 But I moved the lock further up to resolve other edge cases. Now we lock in 

- `UpdateManualConfigurationSettings`
- `UpdateDynamicConfigurationSettings`

Which then call this method. I'll update the comment to make that clear (actually, I'll remove it entirely, it doesn't make sense now.

(Also noticed that we don't need the `Volatile.Write` now either - `lock` inserts the required memory barriers anyway

PR #7689, opened 2025-10-21
[ConfigRegistry] 4/6 Aliases handling and analyzers to prevent using any string key

anna-git:
## Context

Part of **Configuration Inversion (Step 4)** - Stack progress:
1. [#7548](https://github.com/DataDog/dd-trace-dotnet/pull/7548) - Add GitLab step and JSON configuration file
2. [#7688](https://github.com/DataDog/dd-trace-dotnet/pull/7688) - Cleanup configuration / platform keys + analyzers  
3. [#7698](https://github.com/DataDog/dd-trace-dotnet/pull/7698) - Source generator for ConfigurationKeys
4. **→ [#7689](https://github.com/DataDog/dd-trace-dotnet/pull/7689) - Aliases handling and analyzers (this PR)**
5. [#7931](https://github.com/DataDog/dd-trace-dotnet/pull/7931) - Replace manual ConfigurationKeys with generated version
6. #[7932](https://github.com/DataDog/dd-trace-dotnet/pull/7932)- Forbid use of System.Environment methods and adapt everywhere 

## Summary

Adds source generator for configuration key aliases, integrates alias resolution into `ConfigurationBuilder`, and adds Roslyn analyzers to enforce proper configuration key usage.

## Changes

**Alias Source Generator:**
- `ConfigKeyAliasesSwitcherGenerator` reads aliases from `supported-configurations.json`
- Auto-generates switch statements to resolve primary keys from aliases
- Generated for all target frameworks (net461, netstandard2.0, netcoreapp3.1, net6.0)

**ConfigurationBuilder Improvements:**
- Integrated alias resolution directly into `ConfigurationBuilder.WithKeys()`
- Removed manual fallback overloads throughout codebase
- Added `GetKeyWithAlias()` method to `IConfigurationSource` interface

**IntegrationSettings Special Handling:**
- Updated to use pattern-based key construction (e.g., `DD_TRACE_{INTEGRATION}_ENABLED`)
- Simplified configuration reading by leveraging alias system

**Roslyn Analyzers:**
- **DD0007**: `PlatformKeysAnalyzer` - Enforces use of `PlatformKeys` for external platform environment variables
[...]

PR #7688, opened 2025-10-21
[Config registry] 2/5 Reorganize keys, split config keys and platform keys

anna-git:
## Context

Part of **Configuration Inversion (Step 2)** - Stack progress:
1. [#7548](https://github.com/DataDog/dd-trace-dotnet/pull/7548) - Add GitLab step and JSON configuration file
2. **→ [#7688](https://github.com/DataDog/dd-trace-dotnet/pull/7688) - [Config registry] Reorganize keys, split config keys and platform keys(this PR)**
3. [#7698](https://github.com/DataDog/dd-trace-dotnet/pull/7698) - Generate ConfigurationKeys with source generator
4. [#7689](https://github.com/DataDog/dd-trace-dotnet/pull/7689) - Aliases handling and analyzers
5. [#7931](https://github.com/DataDog/dd-trace-dotnet/pull/7931) - Replace manual ConfigurationKeys by generated ones in the whole solution
## Summary

Separates platform-specific environment variables into `PlatformKeys` class and adds Roslyn analyzers to enforce proper usage of configuration keys throughout the codebase.

## Changes

**Platform Keys Separation:**
- Created `PlatformKeys` class for external platform environment variables (AWS, Azure, GCP, Kubernetes, CLR Profiler), those that don't start with DD_ and come from other platforms.
- Updated all references throughout codebase to use `PlatformKeys` for platform-specific variables
- Keeps `ConfigurationKeys` focused only on Datadog configuration

**Roslyn Analyzers:**
- **DD0007**: Platform key analyzer - Enforces use of `PlatformKeys` for external platform environment variables
- **DD0008**: ConfigurationBuilder analyzer - Enforces use of `ConfigurationKeys` or `PlatformKeys` constants in `ConfigurationBuilder.WithKeys()` calls
- Prevents hardcoded strings and ensures type safety

**Configuration Mapping:**
- Added `configuration_keys_mapping.json` to map environment variable names to their `ConfigurationKeys` constant names, so that we avoid renaming hundreds of variables according to automatically source generated deducted ones.
- Enables preservation of original constant names during refactoring
- Supports source generator integration

**Code Updates:**
- Updated `TracerSettings`, test files, and profiler code to use `PlatformKeys`
- Replaced manual `ConfigurationKeys` with generated version from `supported-configurations.json`
- Added missing `DD_TRACE_LOG_PATH` to configuration registry

## Motivation

Ensures clean separation between Datadog and platform configuration by:
- Preventing accidental mixing of platform and Datadog keys
[...]

PR #7548, opened 2025-09-23
[Config Registry] 1/5 Add gitlab step and json configuration file and yaml for doc

anna-git:
## Context

Part of **Configuration Inversion (Step 1)** - Stack progress:
1. **→ [#7548](https://github.com/DataDog/dd-trace-dotnet/pull/7548) - Add GitLab step and JSON configuration file (this PR)**
2. [#7688](https://github.com/DataDog/dd-trace-dotnet/pull/7688) - Reorganize keys, split config keys and platform keys
3. [#7698](https://github.com/DataDog/dd-trace-dotnet/pull/7698) - Generate ConfigurationKeys with source generator
4. [#7689](https://github.com/DataDog/dd-trace-dotnet/pull/7689) - Aliases handling and analyzers
5. [#7931](https://github.com/DataDog/dd-trace-dotnet/pull/7931) - Replace manual ConfigurationKeys by generated ones in the whole solution 

## Summary

Adds centralized configuration registry (`supported-configurations.json` + [supported-configurations-docs.yaml](cci:7://file:///Users/anna.yafi/go/src/github.com/DataDog/dd-trace-dotnet3/tracer/src/Datadog.Trace/Configuration/supported-configurations-docs.yaml:0:0-0:0)) containing ~280 configuration keys with automated GitLab CI validation.

## Changes

**Configuration Files:**
- `supported-configurations.json` - Machine-readable registry of all configuration keys with versions, aliases, and deprecations
- [supported-configurations-docs.yaml](cci:7://file:///Users/anna.yafi/go/src/github.com/DataDog/dd-trace-dotnet3/tracer/src/Datadog.Trace/Configuration/supported-configurations-docs.yaml:0:0-0:0) - Human-readable documentation for each configuration key

**CI/CD:**
- Added `validate_supported_configurations_local_file` step to `.gitlab-ci.yml`
- Updated `.gitlab/one-pipeline.locked.yaml` to reference validation template

**File Structure:**
```json
{
  "supportedConfigurations": { "DD_TRACE_ENABLED": { "version": ["A"] }, ... },
  "aliases": { "DD_AGENT_HOST": ["DD_TRACE_AGENT_HOSTNAME", ...], ... },
  "deprecations": { "DD_TRACE_ANALYTICS_ENABLED": "Deprecated - ...", ... }
}
```

## Motivation

Provides single source of truth for configuration keys to enable:
- Automated documentation validation
- Cross-tracer compatibility checks
- Configuration migration tooling
- Automated doc generation

[...]

PR #7698, opened 2025-10-22
[ConfigRegistry] 3/5 Generate ConfigurationKeys with source generator

anna-git:
## Context

Part of **Configuration Inversion (Step 3)** - Stack progress:
1. [#7548](https://github.com/DataDog/dd-trace-dotnet/pull/7548) - Add GitLab step and JSON configuration file
2. [#7688](https://github.com/DataDog/dd-trace-dotnet/pull/7688) - Cleanup configuration / platform keys + analyzers
3. **→ [#7698](https://github.com/DataDog/dd-trace-dotnet/pull/7698) - Generate ConfigurationKeys with source generator (this PR)**
4. [#7689 ](https://github.com/DataDog/dd-trace-dotnet/pull/7689) - Aliases handling via source generator

## Summary

Adds source generator to auto-generate `ConfigurationKeys` from `supported-configurations.json` with name mapping to preserve existing constant names.

## Changes

**Source Generator:**
- [ConfigurationKeysGenerator](cci:7://file:///var/folders/4j/q1f1tvc503g_thnq6p79cp500000gp/T/SourceGeneratedDocuments/D0E9C9C62821E2CB6FADE080/Datadog.Trace.SourceGenerators/ConfigurationKeysGenerator:0:0-0:0) reads `supported-configurations.json` and [supported-configurations-docs.yaml](cci:7://file:///Users/anna.yafi/go/src/github.com/DataDog/dd-trace-dotnet3/tracer/src/Datadog.Trace/Configuration/supported-configurations-docs.yaml:0:0-0:0)
- Auto-generates `ConfigurationKeys` with nested product classes (AppSec, CIVisibility, OpenTelemetry, etc.)
- Generates XML documentation and `[Obsolete]` attributes from JSON

**Configuration Mapping:**
- `configuration_keys_mapping.json` maps env vars to original constant names (e.g., `DD_TRACE_ENABLED` → `TraceEnabled`)
- Avoids refactoring hundreds of references across codebase

**Updates:**
- Added missing `DD_TRACE_ACTIVITY_LISTENER_ENABLED` key
- Committed temporary [ConfigurationKeys2.g.cs](cci:7://file:///Users/anna.yafi/go/src/github.com/DataDog/dd-trace-dotnet3/tracer/src/Datadog.Trace/Generated/net461/Datadog.Trace.SourceGenerators/ConfigurationKeysGenerator/ConfigurationKeys2.g.cs:0:0-0:0) files for validation

## Motivation

Makes `supported-configurations.json` the single source of truth while preserving existing constant names to avoid massive refactoring.

## Test Coverage

- Added `ConfigurationKeysGeneratorTests.cs` with comprehensive unit tests
- Tests validate generator output, XML documentation generation, and obsolete attribute application
- Tests cover nested product classes, deprecations, and YAML documentation integration
- All existing tests pass without modification

## Related Work

Builds on #7548 (configuration registry) and #7688 (PlatformKeys separation + analyzers).

### DIFF
--- a/tracer/src/Datadog.Trace/Generated/net461/Datadog.Trace.SourceGenerators/ConfigurationKeysGenerator/ConfigurationKeys2.Telemetry.g.cs
+++ b/tracer/src/Datadog.Trace/Generated/net461/Datadog.Trace.SourceGenerators/ConfigurationKeysGenerator/ConfigurationKeys2.Telemetry.g.cs
@@ -30,6 +30,12 @@ internal static partial class ConfigurationKeys2
         public const string AgentlessEnabled = "DD_INSTRUMENTATION_TELEMETRY_AGENTLESS_ENABLED";
 
         /// <summary>
+        /// Configuration key to allow telemetry compression.
+        /// <see cref="Datadog.Trace.Telemetry.TelemetrySettings.CompressionMethod"/>
+        /// </summary>
+        public const string TelemetryCompressionMethod = "DD_INSTRUMENTATION_TELEMETRY_COMPRESSION_METHOD";
+
+        /// <summary>
         /// Configuration key for enabling or disabling internal telemetry.
         /// Default value is <c>true</c> (enabled).
         /// </summary>
--- a/tracer/src/Datadog.Trace/Configuration/configuration_keys_mapping.json
+++ b/tracer/src/Datadog.Trace/Configuration/configuration_keys_mapping.json
@@ -17,6 +17,10 @@
       "const_name": "AgentHost"
     },
     {
+      "env_var": "DD_AGENT_FEATURE_POLLING_ENABLED",
+      "const_name": "AgentFeaturePollingEnabled"
+    },
+    {
       "env_var": "DD_API_KEY",
       "const_name": "ApiKey"
     },
@@ -463,6 +467,10 @@
     {
       "env_var": "DD_INSTRUMENTATION_TELEMETRY_AGENT_PROXY_ENABLED",
       "const_name": "AgentProxyEnabled"
+    },
+    {
+      "env_var": "DD_INSTRUMENTATION_TELEMETRY_COMPRESSION_METHOD",
+      "const_name": "TelemetryCompressionMethod"
     },
     {
       "env_var": "DD_INSTRUMENTATION_TELEMETRY_ENABLED",
--- a/tracer/src/Datadog.Trace/Configuration/supported-configurations-docs.yaml
+++ b/tracer/src/Datadog.Trace/Configuration/supported-configurations-docs.yaml
@@ -16,6 +16,11 @@ DD_AGENT_HOST: |
   Default value is "localhost".
   <seealso cref="Datadog.Trace.Configuration.ExporterSettings.AgentUri"/>
 
+DD_AGENT_FEATURE_POLLING_ENABLED: |
+  Configuration key to disable polling the /info endpoint in the trace agent for feature discovery.
+  Default value is true (polling enabled).
+  <seealso cref="TracerSettings.AgentFeaturePollingEnabled"/>
+
 DD_API_KEY: |
   Configuration key for setting the API key, used by the Agent.
 
@@ -426,6 +431,10 @@ DD_INSTRUMENTATION_TELEMETRY_AGENT_PROXY_ENABLED: |
   Configuration key for sending telemetry via agent proxy. If enabled, sends telemetry
   via agent proxy. Enabled by default. If disabled, or agent is not available, telemetry
   is sent to agentless endpoint, based on <see cref="Datadog.Trace.Configuration.ConfigurationKeys.CIVisibility.AgentlessEnabled"/> setting.
+
+DD_INSTRUMENTATION_TELEMETRY_COMPRESSION_METHOD: |
+  Configuration key to allow telemetry compression.
+  <see cref="Datadog.Trace.Telemetry.TelemetrySettings.CompressionMethod"/>
 
 DD_INSTRUMENTATION_TELEMETRY_ENABLED: |
   Configuration key for enabling or disabling internal telemetry.
--- a/tracer/src/Datadog.Trace/Configuration/supported-configurations.json
+++ b/tracer/src/Datadog.Trace/Configuration/supported-configurations.json
@@ -23,6 +23,11 @@
         "A"
       ]
     },
+    "DD_AGENT_FEATURE_POLLING_ENABLED": {
+      "version": [
+        "A"
+      ]
+    },
     "DD_API_KEY": {
       "version": [
         "A"
@@ -661,6 +666,12 @@
       "product": "Telemetry"
     },
     "DD_INSTRUMENTATION_TELEMETRY_AGENT_PROXY_ENABLED": {
+      "version": [
+        "A"
+      ],
+      "product": "Telemetry"
+    },
+    "DD_INSTRUMENTATION_TELEMETRY_COMPRESSION_METHOD": {
       "version": [
         "A"
       ],
--- a/tracer/src/Datadog.Trace/Generated/net461/Datadog.Trace.SourceGenerators/ConfigurationKeysGenerator/ConfigurationKeys2.g.cs
+++ b/tracer/src/Datadog.Trace/Generated/net461/Datadog.Trace.SourceGenerators/ConfigurationKeysGenerator/ConfigurationKeys2.g.cs
@@ -26,6 +26,13 @@ internal static partial class ConfigurationKeys2
     public const string StatsComputationInterval = "_DD_TRACE_STATS_COMPUTATION_INTERVAL";
 
     /// <summary>
+    /// Configuration key to disable polling the /info endpoint in the trace agent for feature discovery.
+    /// Default value is true (polling enabled).
+    /// </summary>
+    /// <seealso cref="TracerSettings.AgentFeaturePollingEnabled"/>
+    public const string AgentFeaturePollingEnabled = "DD_AGENT_FEATURE_POLLING_ENABLED";
+
+    /// <summary>
     /// Configuration key for the Agent host where the Tracer can send traces.
     /// Overridden by <see cref="Datadog.Trace.Configuration.ExporterSettings.AgentUri"/> if present.
     /// Default value is "localhost".
--- a/tracer/src/Datadog.Trace/Generated/net6.0/Datadog.Trace.SourceGenerators/ConfigurationKeysGenerator/ConfigurationKeys2.Telemetry.g.cs
+++ b/tracer/src/Datadog.Trace/Generated/net6.0/Datadog.Trace.SourceGenerators/ConfigurationKeysGenerator/ConfigurationKeys2.Telemetry.g.cs
@@ -30,6 +30,12 @@ internal static partial class ConfigurationKeys2
         public const string AgentlessEnabled = "DD_INSTRUMENTATION_TELEMETRY_AGENTLESS_ENABLED";
 
         /// <summary>
+        /// Configuration key to allow telemetry compression.
+        /// <see cref="Datadog.Trace.Telemetry.TelemetrySettings.CompressionMethod"/>
+        /// </summary>
+        public const string TelemetryCompressionMethod = "DD_INSTRUMENTATION_TELEMETRY_COMPRESSION_METHOD";
+
+        /// <summary>
         /// Configuration key for enabling or disabling internal telemetry.
         /// Default value is <c>true</c> (enabled).
         /// </summary>
--- a/tracer/src/Datadog.Trace/Generated/net6.0/Datadog.Trace.SourceGenerators/ConfigurationKeysGenerator/ConfigurationKeys2.g.cs
+++ b/tracer/src/Datadog.Trace/Generated/net6.0/Datadog.Trace.SourceGenerators/ConfigurationKeysGenerator/ConfigurationKeys2.g.cs
@@ -26,6 +26,13 @@ internal static partial class ConfigurationKeys2
     public const string StatsComputationInterval = "_DD_TRACE_STATS_COMPUTATION_INTERVAL";
 
     /// <summary>
+    /// Configuration key to disable polling the /info endpoint in the trace agent for feature discovery.
+    /// Default value is true (polling enabled).
+    /// </summary>
+    /// <seealso cref="TracerSettings.AgentFeaturePollingEnabled"/>
+    public const string AgentFeaturePollingEnabled = "DD_AGENT_FEATURE_POLLING_ENABLED";
+
+    /// <summary>
     /// Configuration key for the Agent host where the Tracer can send traces.
     /// Overridden by <see cref="Datadog.Trace.Configuration.ExporterSettings.AgentUri"/> if present.
     /// Default value is "localhost".
--- a/tracer/src/Datadog.Trace/Generated/netcoreapp3.1/Datadog.Trace.SourceGenerators/ConfigurationKeysGenerator/ConfigurationKeys2.Telemetry.g.cs
+++ b/tracer/src/Datadog.Trace/Generated/netcoreapp3.1/Datadog.Trace.SourceGenerators/ConfigurationKeysGenerator/ConfigurationKeys2.Telemetry.g.cs
@@ -30,6 +30,12 @@ internal static partial class ConfigurationKeys2
         public const string AgentlessEnabled = "DD_INSTRUMENTATION_TELEMETRY_AGENTLESS_ENABLED";
 
         /// <summary>
+        /// Configuration key to allow telemetry compression.
+        /// <see cref="Datadog.Trace.Telemetry.TelemetrySettings.CompressionMethod"/>
+        /// </summary>
+        public const string TelemetryCompressionMethod = "DD_INSTRUMENTATION_TELEMETRY_COMPRESSION_METHOD";
+
+        /// <summary>
         /// Configuration key for enabling or disabling internal telemetry.
         /// Default value is <c>true</c> (enabled).
         /// </summary>
--- a/tracer/src/Datadog.Trace/Generated/netcoreapp3.1/Datadog.Trace.SourceGenerators/ConfigurationKeysGenerator/ConfigurationKeys2.g.cs
+++ b/tracer/src/Datadog.Trace/Generated/netcoreapp3.1/Datadog.Trace.SourceGenerators/ConfigurationKeysGenerator/ConfigurationKeys2.g.cs
@@ -26,6 +26,13 @@ internal static partial class ConfigurationKeys2
     public const string StatsComputationInterval = "_DD_TRACE_STATS_COMPUTATION_INTERVAL";
 
     /// <summary>
+    /// Configuration key to disable polling the /info endpoint in the trace agent for feature discovery.
+    /// Default value is true (polling enabled).
+    /// </summary>
+    /// <seealso cref="TracerSettings.AgentFeaturePollingEnabled"/>
+    public const string AgentFeaturePollingEnabled = "DD_AGENT_FEATURE_POLLING_ENABLED";
+
+    /// <summary>
     /// Configuration key for the Agent host where the Tracer can send traces.
     /// Overridden by <see cref="Datadog.Trace.Configuration.ExporterSettings.AgentUri"/> if present.
     /// Default value is "localhost".
--- a/tracer/src/Datadog.Trace/Generated/netstandard2.0/Datadog.Trace.SourceGenerators/ConfigurationKeysGenerator/ConfigurationKeys2.Telemetry.g.cs
+++ b/tracer/src/Datadog.Trace/Generated/netstandard2.0/Datadog.Trace.SourceGenerators/ConfigurationKeysGenerator/ConfigurationKeys2.Telemetry.g.cs
@@ -30,6 +30,12 @@ internal static partial class ConfigurationKeys2
         public const string AgentlessEnabled = "DD_INSTRUMENTATION_TELEMETRY_AGENTLESS_ENABLED";
 
         /// <summary>
+        /// Configuration key to allow telemetry compression.
+        /// <see cref="Datadog.Trace.Telemetry.TelemetrySettings.CompressionMethod"/>
+        /// </summary>
+        public const string TelemetryCompressionMethod = "DD_INSTRUMENTATION_TELEMETRY_COMPRESSION_METHOD";
+
+        /// <summary>
         /// Configuration key for enabling or disabling internal telemetry.
         /// Default value is <c>true</c> (enabled).
         /// </summary>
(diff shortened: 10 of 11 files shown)


lucaspimentel:
@codex review

tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigurationKeysGenerator.cs
```diff
@@ -0,0 +1,959 @@
...
+            return new Result<ConfigurationData>(
+                null!,
+                new EquatableArray<DiagnosticInfo>(
+                [
+                    CreateDiagnosticInfo("DDSG0007", "JSON parse error", $"Error parsing supported-configurations.json: {ex.Message}", DiagnosticSeverity.Error)
+                ]));
+        }
+    }
+
+    private static Dictionary<string, string> ParseDeprecations(string deprecationsJson)
```
lucaspimentel: Are we parsing JSON manually in `ParseDeprecations()`, `ParseMappingFile()`, and `ParseConfigurationEntries()`? Why not use a well-established JSON serializer like `System.Text.Json`?

tracer/src/Datadog.Trace.SourceGenerators/Configuration/JsonReader.cs
```diff
@@ -0,0 +1,198 @@
...
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Datadog.Trace.SourceGenerators.Configuration;
+
+/// <summary>
+/// Json reader as we can't really use vendored libraries here, for framework compatibility reasons
```
lucaspimentel: Is this used an run time? Or only at build time to generate code?

If it's at run time, why can't se use the vendored `Newtonsoft.Json`? Or, if build time, we can't use `System.Text.Json`?

tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigurationKeysGenerator.cs
```diff
@@ -0,0 +1,672 @@
...
+                              .Where(static file => Path.GetFileName(file.Path).Equals(SupportedConfigurationsDocsFileName, StringComparison.OrdinalIgnoreCase))
+                              .WithTrackingName(TrackingNames.ConfigurationKeysGenYamlAdditionalText);
+
+        // Get the configuration_keys_mapping.json file (optional)
+        var mappingFile = context.AdditionalTextsProvider
+                                 .Where(static file => Path.GetFileName(file.Path).Equals(ConfigurationKeysMappingFileName, StringComparison.OrdinalIgnoreCase))
+                                 .WithTrackingName(TrackingNames.ConfigurationKeysGenMappingAdditionalText);
+
+        // Combine all files
+        var combinedFiles = jsonFile.Collect().Combine(yamlFile.Collect()).Combine(mappingFile.Collect());
```
andrewlock: Instead of the multiple stages here you _could_ do something like this:

```csharp
var Files = context.AdditionalTextsProvider
                              .Where(static file => Path.GetFileName(file.Path).Equals(SupportedConfigurationsFileName, StringComparison.OrdinalIgnoreCase)
|| // yaml
|| // mapping
)
                              .WithTrackingName(TrackingNames.AllFiles)
.Collect()
.Select(allFiles => 
{
// find and check for the files here
}
```

tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigurationKeysGenerator.cs
```diff
@@ -0,0 +1,672 @@
...
+                                          {
+                                              var ((jsonFiles, yamlFiles), mappingFiles) = files;
+
+                                              if (jsonFiles.Length == 0)
+                                              {
+                                                  return new Result<(string Json, string? Yaml, string? Mapping)>(
+                                                      (string.Empty, null, null),
+                                                      new EquatableArray<DiagnosticInfo>(
+                                                      [
+                                                          CreateDiagnosticInfo("DDSG0005", "Configuration file not found", $"The file '{SupportedConfigurationsFileName}' was not found. Make sure the supported-configurations.json file exists and is included as an AdditionalFile.", DiagnosticSeverity.Error)
```
andrewlock: Just FYI, best practices for source generators now is to _not_ emit diagnostics, because locations can change somewhat dynamically. Don't think it's an issue here though because you don't actually have any associated locations here, and we shouldn't ever hit this path anyway

tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigurationKeysGenerator.cs
```diff
@@ -0,0 +1,672 @@
...
+                                              string? yamlContent = null;
+                                              if (yamlFiles.Length > 0)
+                                              {
+                                                  var yamlResult = ExtractConfigurationContent(yamlFiles[0], ct);
+                                                  if (yamlResult.Errors.Count == 0)
+                                                  {
+                                                      yamlContent = yamlResult.Value;
+                                                  }
+
+                                                  // YAML is optional, so we don't fail if it has errors
```
andrewlock: Optional, but we already found it, so we should probably bail if we can't _read_ it?

tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigurationKeysGenerator.cs
```diff
@@ -0,0 +1,672 @@
...
+        sb.Append(Constants.FileHeader);
+        sb.Append(Constants.ConfigurationGeneratorComment);
+    }
+
+    private static string GenerateProductPartialClass(string product, List<KeyValuePair<string, ConfigEntry>> entries, Dictionary<string, ConstantNameMapping>? nameMapping)
+    {
+        var sb = new StringBuilder();
+
+        AppendFileHeader(sb);
+        sb.AppendLine($"namespace {Namespace};");
```
andrewlock: nit: never use string interpolation with `StringBuilder`, just use `Append()`. Otherwise you're doing extra allocations. Applies to all the `$""` cases in here 🙂 

Also, always use `';'` instead of `";"` when adding single characters:

```suggestion
        sb.Append("namespace ").Append(Namespace).AppendLine(';');
```

tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigurationKeysGenerator.cs
```diff
@@ -0,0 +1,672 @@
...
+            case "Tracer":
+                return new[] { "Trace" };
+            case "CiVisibility":
+                return new[] { "Civisibility" };
+            case "OpenTelemetry":
+                return new[] { "Otel" };
+            default:
+                return new[] { productName! };
+        }
+    }
```
andrewlock: nit:


```suggestion
    private static string[] ProductNameEquivalents(string? productName)
        => productName switch
        {
            null or "" => [ string.Empty ],
            "AppSec" => [ "Appsec" ],
            "Tracer" => [ "Trace" ],
            "CiVisibility" => [ "Civisibility" ],
            "OpenTelemetry" => [ "Otel" ],
            var x => [ x ],
        }
```

tracer/src/Datadog.Trace.SourceGenerators/Datadog.Trace.SourceGenerators.csproj
```diff
@@ -12,6 +12,7 @@
   <ItemGroup>
     <PackageReference Include="System.Reflection.Emit" Version="4.7.0" />
     <PackageReference Include="System.Reflection.Emit.Lightweight" Version="4.7.0" />
+    <PackageReference Include="System.Text.Json" Version="8.0.5" PrivateAssets="all" />
```
andrewlock: Why 8.0.5? 🤔 Given we build with the .NET 10 SDK, I think you should use .NET 10? Also, you should check if this actually works in Visual Studio, because I worry it _won't_ (VS builds using the .NET FX MSBuild by default). 

e.g. the docs explain how his is a PITA 😅 https://github.com/dotnet/roslyn/blob/main/docs/features/incremental-generators.cookbook.md#use-functionality-from-nuget-packages

I wonder if we can just remove `PrivateAssets="all"`? 🤔

tracer/src/Datadog.Trace.SourceGenerators/Helpers/YamlReader.cs
```diff
@@ -0,0 +1,142 @@
...
+    internal static class YamlReader
+    {
+        /// <summary>
+        /// Parses a YAML file containing configuration key documentation.
+        /// Expects format: KEY: | followed by multi-line documentation
+        /// </summary>
+        public static Dictionary<string, string> ParseDocumentation(string yamlContent)
+        {
+            var result = new Dictionary<string, string>();
+            var lines = yamlContent.Replace("\r\n", "\n").Split('\n');
```
andrewlock: tiny nit, if this code ends up running a lot, that's a bunch of extra allocations, maybe we could make it a span first and use that? If it's only going to run once on project load, then it's less of an issue (just makes initial loads slower)

tracer/src/Datadog.Trace/Datadog.Trace.csproj
```diff
@@ -156,4 +156,23 @@
...
+    </AdditionalFiles>
+  </ItemGroup>
+
+  <ItemGroup>
+    <Folder Include="Generated\net461\Datadog.Trace.SourceGenerators\ConfigurationKeysGenerator\" />
+    <Folder Include="Generated\net6.0\Datadog.Trace.SourceGenerators\ConfigurationKeysGenerator\" />
+    <Folder Include="Generated\netcoreapp3.1\Datadog.Trace.SourceGenerators\ConfigurationKeysGenerator\" />
+    <Folder Include="Generated\netstandard2.0\Datadog.Trace.SourceGenerators\ConfigurationKeysGenerator\" />
+  </ItemGroup>
+
```
andrewlock: nit: we don't need these:
```suggestion
```

tracer/test/Datadog.Trace.SourceGenerators.Tests/ConfigurationKeysGeneratorTests.cs
andrewlock, line 1: Nice work on these tests 👍

tracer/src/Datadog.Trace.SourceGenerators/Configuration/JsonReader.cs
anna-git: Good point changed it to use `System.Text.Json` !

tracer/src/Datadog.Trace/Datadog.Trace.csproj
anna-git: fixed in [6ccb01e](https://github.com/DataDog/dd-trace-dotnet/pull/7698/commits/6ccb01ef401444eb38dfaee62024d1ccacb000a1)

tracer/src/Datadog.Trace.SourceGenerators/Helpers/YamlReader.cs
anna-git: changed in [6ccb01e](https://github.com/DataDog/dd-trace-dotnet/pull/7698/commits/6ccb01ef401444eb38dfaee62024d1ccacb000a1)

tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigurationKeysGenerator.cs
anna-git: changed in [6ccb01e](https://github.com/DataDog/dd-trace-dotnet/pull/7698/commits/6ccb01ef401444eb38dfaee62024d1ccacb000a1)

tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigurationKeysGenerator.cs
anna-git: changed in [6ccb01e](https://github.com/DataDog/dd-trace-dotnet/pull/7698/commits/6ccb01ef401444eb38dfaee62024d1ccacb000a1)

tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigurationKeysGenerator.cs
anna-git: yep, changed in [6ccb01e](https://github.com/DataDog/dd-trace-dotnet/pull/7698/commits/6ccb01ef401444eb38dfaee62024d1ccacb000a1)

tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigurationKeysGenerator.cs
anna-git: good to know thanks!

tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigurationKeysGenerator.cs
anna-git: changed in [ed3a6bc](https://github.com/DataDog/dd-trace-dotnet/pull/7698/commits/ed3a6bc11707177482b0004027504ea30cd16280)

anna-git:
> Nice, LGTM.
> 
> The generation work is somewhat heavy so I was just randomly wondering if we should try to add some sort of caching layer to not-regenerate everything on the first project load if nothing has changed, but I'm not sure exactly _how_ to do that or anything right now, so just mentioning it as food for thought 😄

like keeping a file somewhere containing a hash of the 3 files? and comparing the hash on load? 
Maybe if we see build is too long, we could try adding it later to cut down some time?

tracer/src/Datadog.Trace.SourceGenerators/Datadog.Trace.SourceGenerators.csproj
anna-git: tested on Lucas Visual studio and it seemed to work fine 🤔

andrewlock: Interesting, OK, it's fine then 😄

andrewlock:
> > Nice, LGTM.
> > The generation work is somewhat heavy so I was just randomly wondering if we should try to add some sort of caching layer to not-regenerate everything on the first project load if nothing has changed, but I'm not sure exactly _how_ to do that or anything right now, so just mentioning it as food for thought 😄
> 
> like keeping a file somewhere containing a hash of the 3 files? and comparing the hash on load? Maybe if we see build is too long, we could try adding it later to cut down some time?

Yeah, something like that 🙂 Honestly it was only a random idea, and in general you shouldn't write files from source generators (they have analyzers to make sure you don't) so it would mean adding something somewhere we _could_ read it. Not something I've ever tried or anything, no idea if it would be worth it, and absolutely not suggesting you should do it, just something I might R&D one day 😄

tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigurationKeysGenerator.cs
```diff
@@ -0,0 +1,767 @@
...
+    private static Result<ConfigurationData> MergeResults(
+        Result<ConfigurationData> jsonResult,
+        Result<ParsedYamlDocs> yamlResult,
+        Result<ParsedMappingData> mappingResult)
+    {
+        var diagnostics = new List<DiagnosticInfo>(jsonResult.Errors);
+        diagnostics.AddRange(yamlResult.Errors);
+        diagnostics.AddRange(mappingResult.Errors);
+
+        if (jsonResult.Value is null)
```
lucaspimentel, line 218: `jsonResult.Value` can never be null?

tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigurationKeysGenerator.cs
```diff
@@ -0,0 +1,767 @@
...
+    private static void Execute(SourceProductionContext context, Result<ConfigurationData> result)
+    {
+        // Report any diagnostics
+        foreach (var diagnostic in result.Errors)
+        {
+            context.ReportDiagnostic(Diagnostic.Create(diagnostic.Descriptor, diagnostic.Location?.ToLocation()));
+        }
+
+        // Generate source code even if there are errors (use empty configuration as fallback)
+        var configData = result.Value ?? new ConfigurationData(new Dictionary<string, ConfigEntry>(), null);
```
lucaspimentel, line 109: `result.Value` can never be null?

tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigurationKeysGenerator.cs
```diff
@@ -0,0 +1,767 @@
...
+        {
+            var docLines = entry.Documentation.Split(["\r\n", "\n"], StringSplitOptions.None);
+            sb.AppendLine($"{indent}/// <summary>");
+            var seeAlsoLines = new List<string>();
+            foreach (var line in docLines)
+            {
+                var trimmedLine = line.TrimStart();
+
+                // Check if line contains seealso tag (self-closing /> or closing </seealso>) as we need to extract it
+                if (trimmedLine.StartsWith("<seealso") && (trimmedLine.Contains("/>") || trimmedLine.Contains("</seealso>")))
```
lucaspimentel:
```suggestion
                if (trimmedLine.StartsWith("<seealso", StringComparison.Ordinal) && (trimmedLine.Contains("/>") || trimmedLine.Contains("</seealso>")))
```

tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigurationKeysGenerator.cs
```diff
@@ -0,0 +1,767 @@
...
+        // Convert to PascalCase
+        var parts = name.Split('_');
+        var pascalName = string.Concat(parts.Select(p => p.Length > 0 ? char.ToUpper(p[0]) + p.Substring(1).ToLower() : string.Empty));
+
+        // Strip product prefix if the const name starts with it
+        if (productNames != null)
+        {
+            foreach (var productName in productNames)
+            {
+                if (pascalName!.Length > productName.Length &&
```
lucaspimentel: `pascalName` can't be null?

```suggestion
                if (pascalName.Length > productName.Length &&
```

tracer/src/Datadog.Trace/Generated/net461/Datadog.Trace.SourceGenerators/ConfigurationKeysGenerator/ConfigurationKeys2.DataStreamsMonitoring.g.cs
```diff
@@ -0,0 +1,31 @@
...
+// NOTE: If you remove keys/products from the JSON, run 'dotnet clean' and remove old generated files.
+namespace Datadog.Trace.Configuration;
+
+internal static partial class ConfigurationKeys2
+{
+    internal static class DataStreamsMonitoring
+    {
+        /// <summary>
+        /// Enables data streams monitoring support
+        /// <see cref="Datadog.Trace.Configuration.TracerSettings.IsDataStreamsMonitoringEnabled"/>
```
zacharycmontoya: Semantically I think the `<see>` section should come after the `<summary>` section

tracer/src/Datadog.Trace/Generated/net461/Datadog.Trace.SourceGenerators/ConfigurationKeysGenerator/ConfigurationKeys2.Telemetry.g.cs
```diff
@@ -0,0 +1,81 @@
...
+        /// <summary>
+        /// Configuration key for whether telemetry metrics should be sent.
+        /// <see cref="Datadog.Trace.Telemetry.TelemetrySettings.MetricsEnabled"/>
+        /// </summary>
+        public const string MetricsEnabled = "DD_TELEMETRY_METRICS_ENABLED";
+
+        /// <summary>
+        /// Configuration key to enable or disable the ActivityListener.
+        /// </summary>
+        public const string ActivityListenerEnabled = "DD_TRACE_ACTIVITY_LISTENER_ENABLED";
```
zacharycmontoya: Note for follow-up: This key logically belongs with the `OpenTelemetry` set of environment variables

tracer/test/Datadog.Trace.SourceGenerators.Tests/ConfigurationKeysGeneratorTests.cs
```diff
@@ -0,0 +1,503 @@
...
+            [],
+            [],
+            [("supported-configurations.json", invalidJson)],
+            assertOutput: false);
+
+        diagnostics.Should().Contain(d => d.Id == "DDSG0002" || d.Id == "DDSG0007");
+    }
+
+    [Fact]
+    public void SortsEntriesAlphabetically()
```
zacharycmontoya: Small nit to be even more precise

```suggestion
    public void SortsEntriesAlphabeticallyByEnvironmentVariable()
```

zacharycmontoya (approved):
LGTM. I actually reviewed this only looking at the generated files and tests, and by that measure these changes look great 👍🏼

tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigurationKeysGenerator.cs
anna-git: I think it could be given that MergeResults return a null wrapped if the json doc is not present.. but this shouldn't happen...

tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigurationKeysGenerator.cs
anna-git: it could be if the json array is empty.. but this would certainly be a bug..

tracer/test/Datadog.Trace.SourceGenerators.Tests/ConfigurationKeysGeneratorTests.cs
anna-git: yes! changed in [85deee3](https://github.com/DataDog/dd-trace-dotnet/pull/7698/commits/85deee33871decb344dd8d29e4f4de7f1744fc57)

tracer/src/Datadog.Trace/Generated/net461/Datadog.Trace.SourceGenerators/ConfigurationKeysGenerator/ConfigurationKeys2.Telemetry.g.cs
anna-git: thanks, changed in [8776673](https://github.com/DataDog/dd-trace-dotnet/pull/7698/commits/87766739e5a7ea1841c8926de56e11dc76799477)

tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigurationKeysGenerator.cs
anna-git: yes, changed in 85deee3

tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigurationKeysGenerator.cs
anna-git: yes changed in [28e3698](https://github.com/DataDog/dd-trace-dotnet/pull/7698/commits/28e3698a53ab295ef3c7d7320cc92343dd7c247a)

PR #7702, opened 2025-10-23
Support adding kafka_cluster_id for Confluent.Kafka

robcarlan-datadog:
## Summary of changes

- Adds kafka_cluster_id as a tag for APM Spans and DSM checkpoints/offsets. (Note that **DSM is enabled by default**). Without this, we cannot differentiate between reported offsets on the same topic but across different clusters, which leads us to wildly incorrect metric values. (For example, a prod cluster with consume/produce offsets at 1000,1001 should have an offset lag of 1. If there is also a matching dev cluster with offsets at 0 and 1, then we can calculate the offset lag as 1000).
- In most tracer libraries (python, java, node) we support tracking `kafka_cluster_id`. Without this, metrics become incorrect when there are the same topics across different clusters (i.e. dev/prod environments).  
- It's available for Confluent.Kafka 2.3.0 and above only - roughly half of orgs that use this library according to telemetry. (For DSM where this is most useful, we can add a recommendation to our docs to use this version or above.)

**Note:** I drafted an alternative approach calling librdkafka directly [here](https://github.com/DataDog/dd-trace-dotnet/pull/8264). That gives the cluster id without any external API call or version dependency (It's available and unchanged from librdkafka 1.0.0), but calls into unmanaged code

## Reason for change

This functionality exists in other tracers:
1. Java (doesn't block, intercepts the metadata response to enrich cluster id going forwards, see [here](https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/instrumentation/kafka/kafka-clients-3.8/src/main/java17/datadog/trace/instrumentation/kafka_clients38/ProducerAdvice.java#L44) and [here](https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/instrumentation/kafka/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/MetadataInstrumentation.java#L82))

2. Node (blocks, see [here](https://github.com/DataDog/dd-trace-js/blob/master/packages/datadog-instrumentations/src/kafkajs.js#L234) and [here](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/contrib/internal/kafka/patch.py#L183))

3. Python (blocks with a 1 second timeout, see [here](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/contrib/internal/kafka/patch.py#L360) and [here](https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/contrib/internal/kafka/patch.py#L183))

## Implementation details

Constructs the Kafka Admin API client and queries for the cluster id directly on consumer/producer startup. 
We cache this by bootstrap servers, so we expect to make this API call once.

## Other details
These DSM metrics will now tag by kafka_cluster_id
Screenshot 2026-02-25 at 12 42 00 pm
Added to spans (I checked this on produce too, I just don't have a screenshot of it)
Screenshot 2026-02-24 at 5 21 59 pm


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -364,6 +406,82 @@ internal static void TryInjectHeaders<TTopicPartitionMarker, TMessage>(
...
+            if (string.IsNullOrEmpty(bootstrapServers))
+            {
+                return null;
+            }
+
+            try
+            {
+                _isGettingClusterId = true;
+
+                var configType = Type.GetType("Confluent.Kafka.AdminClientConfig, Confluent.Kafka");
```
robcarlan-datadog: afaik this is the only way to get this object, and the builder below

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -364,6 +405,89 @@ internal static void TryInjectHeaders<TTopicPartitionMarker, TMessage>(
...
+                    return null;
+                }
+
+                var handle = clientHandle.Handle;
+                if (handle is null)
+                {
+                    return null;
+                }
+
+                var builderType = Type.GetType("Confluent.Kafka.DependentAdminClientBuilder, Confluent.Kafka");
```
robcarlan-datadog: `DependentAdminClientBuilder` resuses existing handle from existing producer/consumer

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -364,6 +405,89 @@ internal static void TryInjectHeaders<TTopicPartitionMarker, TMessage>(
...
+                    return clusterId;
+                }
+                finally
+                {
+                    adminClient.Dispose();
+                }
+            }
+            catch (Exception ex)
+            {
+                Log.Debug(ex, "Error extracting cluster_id from Kafka metadata");
```
robcarlan-datadog: This will appear for anyone using a version of Confluent.Kafka too low: I assume that's ok for debug level

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/IDescribeClusterResult.cs
```diff
@@ -0,0 +1,16 @@
...
+#nullable enable
+
+namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Kafka;
+
+/// <summary>
+/// Duck Type for Confluent.Kafka.Admin.DescribeClusterResult
+/// </summary>
+internal interface IDescribeClusterResult
+{
+    string ClusterId { get; }
```
andrewlock:
```suggestion
    string? ClusterId { get; }
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/IClientHandle.cs
```diff
@@ -0,0 +1,16 @@
...
+#nullable enable
+
+namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Kafka;
+
+/// <summary>
+/// Duck Type for Confluent.Kafka.IClient
+/// </summary>
+internal interface IClientHandle
+{
+    object Handle { get; }
```
andrewlock: Unlikely, but possible, so we should play it safe:
```suggestion
    object? Handle { get; }
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/IAdminClientBuilder.cs
```diff
@@ -0,0 +1,16 @@
...
+#nullable enable
+
+namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Kafka;
+
+/// <summary>
+/// Duck Type for Confluent.Kafka.AdminClientBuilder and DependentAdminClientBuilder
+/// </summary>
+internal interface IAdminClientBuilder
+{
+    object Build();
```
andrewlock: I think we can return a chained duck type here? 🤔 
```suggestion
    IAdminClient Build();
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaConsumerCommitAllIntegration.cs
```diff
@@ -32,14 +32,18 @@ internal static CallTargetReturn<TResponse> OnMethodEnd<TTarget, TResponse>(TTar
...
             {
                 var item = response[i];
-                dataStreams.TrackBacklog(
-                    $"consumer_group:{groupId},partition:{item.Partition.Value},topic:{item.Topic},type:kafka_commit",
-                    item.Offset.Value);
+                var backlogTags = $"consumer_group:{groupId},partition:{item.Partition.Value},topic:{item.Topic},type:kafka_commit";
+                if (!string.IsNullOrEmpty(clusterId))
+                {
+                    backlogTags = $"kafka_cluster_id:{clusterId},{backlogTags}";
+                }
```
andrewlock: This will do extra allocation, it's better to do it all in one, even though it gives duplication.

Side note, I thought these tags needed to be in strict alphabetical order to give a correct hash? 🤔 

```suggestion
                var backlogTags = StringUtil.IsNullOrEmpty(clusterId)
                    ? $"consumer_group:{groupId},partition:{item.Partition.Value},topic:{item.Topic},type:kafka_commit"
                    : $"consumer_group:{groupId},kafka_cluster_id:{clusterId},partition:{item.Partition.Value},topic:{item.Topic},type:kafka_commit";
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaConsumerCommitIntegration.cs
```diff
@@ -37,15 +37,19 @@ internal static CallTargetReturn OnMethodEnd<TTarget>(TTarget instance, Exceptio
...
                 if (offset.TryDuckCast<ITopicPartitionOffset>(out var item))
                 {
-                    dataStreams.TrackBacklog(
-                        $"consumer_group:{groupId},partition:{item.Partition.Value},topic:{item.Topic},type:kafka_commit",
-                        item.Offset.Value);
+                    var backlogTags = $"consumer_group:{groupId},partition:{item.Partition.Value},topic:{item.Topic},type:kafka_commit";
+                    if (!string.IsNullOrEmpty(clusterId))
+                    {
+                        backlogTags = $"kafka_cluster_id:{clusterId},{backlogTags}";
+                    }
```
andrewlock: As above. Also note using StringUtil.IsNullOrEmpty avoids nullability warnings in .NET FX/.NET Standard

```suggestion
                    var backlogTags = StringUtil.IsNullOrEmpty(clusterId)
                        ? $"consumer_group:{groupId},partition:{item.Partition.Value},topic:{item.Topic},type:kafka_commit"
                        : $"consumer_group:{groupId},kafka_cluster_id:{clusterId},partition:{item.Partition.Value},topic:{item.Topic},type:kafka_commit";
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaConsumerConstructorIntegration.cs
```diff
@@ -63,9 +63,7 @@ internal static CallTargetState OnMethodBegin<TTarget, TConsumerBuilder>(TTarget
             // Only config setting "group.id" is required, so assert that the value is non-null before adding to the ConsumerGroup cache
             if (groupId is not null)
             {
-                // Save the map between this consumer and a consumer group
-                ConsumerCache.SetConsumerGroup(instance, groupId, bootstrapServers);
-                return new CallTargetState(scope: null, state: instance);
+                return new CallTargetState(scope: null, state: new string[] { groupId, bootstrapServers });
```
andrewlock: nit: instead of an array, use a named helper type (defined below)
```suggestion
                return new CallTargetState(scope: null, state: new Config(groupId, bootstrapServers));
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaConsumerConstructorIntegration.cs
```diff
@@ -63,9 +63,7 @@ internal static CallTargetState OnMethodBegin<TTarget, TConsumerBuilder>(TTarget
             // Only config setting "group.id" is required, so assert that the value is non-null before adding to the ConsumerGroup cache
             if (groupId is not null)
             {
-                // Save the map between this consumer and a consumer group
-                ConsumerCache.SetConsumerGroup(instance, groupId, bootstrapServers);
```
andrewlock, line 67: I'm assuming the constructor _never_ does something in its constructor that we would need to track? 🤔

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaConsumerConstructorIntegration.cs
andrewlock, line 90:
```suggestion
    private sealed class Config(string groupId, string bootstrapServers)
    {
        public string GroupId { get; } = groupId;

        public string BootstrapServers{ get; } = bootstrapServers;
    }
}
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaConsumerConstructorIntegration.cs
```diff
@@ -74,11 +72,12 @@ internal static CallTargetState OnMethodBegin<TTarget, TConsumerBuilder>(TTarget
...
-        // This method is called in the Consumer constructor, so if we have an exception
-        // the consumer won't be created, so no point recording it.
-        if (exception is not null && state is { State: { } consumer })
+        if (exception is null && state is { State: string[] { Length: 2 } config })
         {
-            ConsumerCache.RemoveConsumerGroup(consumer);
+            var groupId = config[0];
+            var bootstrapServers = config[1];
+            var clusterId = KafkaHelper.GetClusterId(bootstrapServers, instance) ?? string.Empty;
+            ConsumerCache.SetConsumerGroup(instance, groupId, bootstrapServers, clusterId);
```
andrewlock:
```suggestion
        if (exception is null && state is { State: Config config })
        {
            var clusterId = KafkaHelper.GetClusterId(config.BootstrapServers, instance) ?? string.Empty;
            ConsumerCache.SetConsumerGroup(instance, config.GroupId, config.BootstrapServers, clusterId);
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -244,9 +256,19 @@ private static long GetMessageSize<T>(T message)
                 {
                     // TODO: we could pool these arrays to reduce allocations
                     // NOTE: the tags must be sorted in alphabetical order
```
andrewlock, line 259: Oh, huh, maybe this _isn't_ required in the commit all integration? 🤔 Would seem strange though

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaConsumerConstructorIntegration.cs
```diff
@@ -74,11 +72,12 @@ internal static CallTargetState OnMethodBegin<TTarget, TConsumerBuilder>(TTarget
...
     {
-        // This method is called in the Consumer constructor, so if we have an exception
-        // the consumer won't be created, so no point recording it.
-        if (exception is not null && state is { State: { } consumer })
+        if (exception is null && state is { State: string[] { Length: 2 } config })
         {
-            ConsumerCache.RemoveConsumerGroup(consumer);
+            var groupId = config[0];
+            var bootstrapServers = config[1];
+            var clusterId = KafkaHelper.GetClusterId(bootstrapServers, instance) ?? string.Empty;
```
andrewlock: Is using `` actually correct in this case? Or should it be omitted from the tags entirely if we can't grab it? 🤔 

If `""` is correct, maybe `KafkaHelper.GetClusterId()` should return that instead of `null`?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -339,9 +363,26 @@ internal static void TryInjectHeaders<TTopicPartitionMarker, TMessage>(
...
+                        edgeTags = string.IsNullOrEmpty(topic)
+                                       ? new[] { "direction:out", $"kafka_cluster_id:{producerClusterId}", "type:kafka" }
+                                       : new[] { "direction:out", $"kafka_cluster_id:{producerClusterId}", $"topic:{topic}", "type:kafka" };
+                    }
+                    else
+                    {
+                        edgeTags = string.IsNullOrEmpty(topic)
                                        ? DefaultProduceEdgeTags
-                                       : ["direction:out", $"topic:{topic}", "type:kafka"];
+                                       : new[] { "direction:out", $"topic:{topic}", "type:kafka" };
```
andrewlock: nit: use collection expressions
```suggestion
                        edgeTags = string.IsNullOrEmpty(topic)
                                       ? ["direction:out", $"kafka_cluster_id:{producerClusterId}", "type:kafka"]
                                       : ["direction:out", $"kafka_cluster_id:{producerClusterId}", $"topic:{topic}", "type:kafka"];
                    }
                    else
                    {
                        edgeTags = string.IsNullOrEmpty(topic)
? DefaultProduceEdgeTags
                                       : ["direction:out", $"topic:{topic}", "type:kafka"];
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -364,6 +405,89 @@ internal static void TryInjectHeaders<TTopicPartitionMarker, TMessage>(
...
+            }
+
+            if (ClusterIdCache.TryGetValue(bootstrapServers!, out var cached))
+            {
+                return cached;
+            }
+
+            try
+            {
+                if (!clientInstance.TryDuckCast<IClientHandle>(out var clientHandle))
```
andrewlock: Is there any reason the `TryDuckCast` _should_ fail? i.e. is there any scenario we call this where the `clientInstance` will _not_ be an `IClientHandle`? If we always _expect_ it to work, we should use `DuckCast<T>`.

We should also potentially _not_ catch exceptions caused by trying (and failing) to duck type - the reason being that our integrations automatically disable themselves when they detect a duck typing exception. In this case, we would try to duck type every time, hurting performance from repeatedly throwing exceptions which will never pass 🙂 

Just things to think about!

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
andrewlock: Every time we call `GetClusterId`, we will do this reflection, create a new instance of the builder, etc, etc. Could we cache all this, hang onto the admin client long-term (instead of repeatedly rebuilding etc)? Or is that not worth it, given we're already caching against the `bootstrapServers` 🤔 

Might be worth adding a debug log in here so that we can catch cases where this is erroneously happening a lot...

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -364,6 +405,89 @@ internal static void TryInjectHeaders<TTopicPartitionMarker, TMessage>(
...
         }
 
+        internal static string? GetClusterId(string? bootstrapServers, object clientInstance)
+        {
+            if (string.IsNullOrEmpty(bootstrapServers))
+            {
+                return null;
+            }
+
+            if (ClusterIdCache.TryGetValue(bootstrapServers!, out var cached))
```
andrewlock: nit: if you use `StringUtil` instead, you can get rid of the `!`


```suggestion
            if (StringUtil.IsNullOrEmpty(bootstrapServers))
            {
                return null;
            }

            if (ClusterIdCache.TryGetValue(bootstrapServers, out var cached))
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -364,6 +405,89 @@ internal static void TryInjectHeaders<TTopicPartitionMarker, TMessage>(
...
+                if (adminClientObj is null)
+                {
+                    return null;
+                }
+
+                if (!adminClientObj.TryDuckCast<IAdminClient>(out var adminClient))
+                {
+                    (adminClientObj as IDisposable)?.Dispose();
+                    return null;
+                }
```
andrewlock: I think with the duckchaining suggested in `IAdminClientBuilder`, and switching to `DuckCast<>`, this becomes
```suggestion
                var builder = Activator.CreateInstance(builderType, new object[] { handle });
                using var adminClientObj = builder.DuckCast<IAdminClientBuilder>().Build();
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -364,6 +405,89 @@ internal static void TryInjectHeaders<TTopicPartitionMarker, TMessage>(
...
+                try
+                {
+                    var clusterId = DescribeClusterWithTimeout(adminClient);
+                    ClusterIdCache.TryAdd(bootstrapServers!, clusterId);
+                    return clusterId;
+                }
+                finally
+                {
+                    adminClient.Dispose();
+                }
```
andrewlock: With the `using` declaration above, and if you use `StringUtil`, this becomes
```suggestion
                var clusterId = DescribeClusterWithTimeout(adminClient);
                ClusterIdCache.TryAdd(bootstrapServers, clusterId);
                return clusterId;
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -364,6 +405,89 @@ internal static void TryInjectHeaders<TTopicPartitionMarker, TMessage>(
...
+
+        private static string? DescribeClusterWithTimeout(IAdminClient adminClient)
+        {
+            object? options = null;
+            var optionsType = Type.GetType("Confluent.Kafka.Admin.DescribeClusterOptions, Confluent.Kafka");
+            if (optionsType is not null)
+            {
+                options = Activator.CreateInstance(optionsType);
+                var requestTimeoutProp = optionsType.GetProperty("RequestTimeout");
+                requestTimeoutProp?.SetValue(options, TimeSpan.FromSeconds(2));
```
andrewlock: We should probably create an `IDescribeClusterOptions` duck type to do this instead of using the "raw" reflection

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -364,6 +405,89 @@ internal static void TryInjectHeaders<TTopicPartitionMarker, TMessage>(
...
+                Log.Debug(ex, "Error extracting cluster_id from Kafka metadata");
+                return null;
+            }
+        }
+
+        private static string? DescribeClusterWithTimeout(IAdminClient adminClient)
+        {
+            object? options = null;
+            var optionsType = Type.GetType("Confluent.Kafka.Admin.DescribeClusterOptions, Confluent.Kafka");
+            if (optionsType is not null)
```
andrewlock: IMO, if it _is_ `null`, there's a big issue, so we should probably log an error and bail here, otherwise we can't set the timeout and we could end up stuck. Given there's no good reason we _should_ fail, we should also log and `Error` (so it ends up in our error tracking, and we can investigate). Same goes for anywhere we unexpectedly fail in the `DescribeCluster` I would have said 🤔

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
andrewlock: Every time we call `GetClusterId`, we will do this reflection, create a new instance of the builder, etc, etc. Could we cache all this, hang onto the admin client long-term (instead of repeatedly rebuilding etc)? Or is that not worth it, given we're already caching against the `bootstrapServers` 🤔 

Might be worth adding a debug log in this method so that we can catch cases where this is erroneously happening a lot...

andrewlock: Additionally, it's generally better to "anchor" the type lookup to a specific assembly, rather than just passing it as a string, as that can fail in unexpected ways 😄 Generally, the "best" way is to pass a "marker" type as a generic parameter on the method, e.g.

```csharp
        internal static string? GetClusterId<TInstance>(string? bootstrapServers, TInstance clientInstance)
```

so that you can do `typeof(TInstance).Assembly.GetType("...")`, but if we're going to call this method with many different instance _types_, and _Rarely_ hit this path, then maybe leaving it as `object` and doing a `GetType()` is an easier/better option:

```suggestion
                var builderType = clientInstance.GetType().Assembly.GetType("Confluent.Kafka.DependentAdminClientBuilder");
```

However, that assumes that the types are defined in the same assembly, though I'm not _sure_ that's the case here 🤔. If not, we should pass a different marker type that definitely _is_ in the assembly...

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -364,6 +405,89 @@ internal static void TryInjectHeaders<TTopicPartitionMarker, TMessage>(
...
+            {
+                Log.Debug(ex, "Error extracting cluster_id from Kafka metadata");
+                return null;
+            }
+        }
+
+        private static string? DescribeClusterWithTimeout(IAdminClient adminClient)
+        {
+            object? options = null;
+            var optionsType = Type.GetType("Confluent.Kafka.Admin.DescribeClusterOptions, Confluent.Kafka");
```
andrewlock: Same comment as above about using a marker type to get the assembly to search

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -364,6 +405,89 @@ internal static void TryInjectHeaders<TTopicPartitionMarker, TMessage>(
...
+            if (optionsType is not null)
+            {
+                options = Activator.CreateInstance(optionsType);
+                var requestTimeoutProp = optionsType.GetProperty("RequestTimeout");
+                requestTimeoutProp?.SetValue(options, TimeSpan.FromSeconds(2));
+            }
+
+            var duckTask = adminClient.DescribeClusterAsync(options);
+            var describeResult = duckTask.GetAwaiter().GetResult();
+            return describeResult?.ClusterId;
```
andrewlock, line 466: The `GetAwaiter().GetResult()` pattern is _better_ than using `.Result` but is still "bad" as it's sync-over-async, and can technically cause deadlocks...

We have `SafeWait` and `SafeGetResult` methods for handling this, but it won't work for your duck typing patterns here, so you might need to do it "manually" yourself.

(It's ugly as hell, but this is trying to avoid any boxing of struct duck types by using the generic with constraint trick)

```suggestion
            var describeResult = SafeGetResult<IDuckTypeTask<IDescribeClusterResult>, IDescribeClusterResult>(duckTask);
            return describeResult?.ClusterId;
            
            static TResult? SafeGetResult<TTask, TResult>(TTask task)
                where TTask : IDuckTypeTask<TResult>
                where TResult : IDescribeClusterResult
            {
                if (task.IsCompletedSuccessfully)
                {
                    return task.Result;
                }
    
                var originalContext = SynchronizationContext.Current;
                try
                {
                    // Set the synchronization context to null to avoid deadlocks.
                    SynchronizationContext.SetSynchronizationContext(null);
    
                    // Wait synchronously for the task to complete.
                    return task.GetAwaiter().GetResult();
                }
                finally
                {
                    // Restore the original synchronization context.
                    SynchronizationContext.SetSynchronizationContext(originalContext);
                }
            }
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaProduceAsyncIntegration.cs
```diff
@@ -100,9 +101,13 @@ internal static TResponse OnAsyncMethodEnd<TTarget, TResponse>(TTarget instance,
                     var dataStreams = Tracer.Instance.TracerManager.DataStreamsManager;
                     if (dataStreams.IsEnabled)
                     {
-                        dataStreams.TrackBacklog(
-                            $"partition:{deliveryResult.Partition.Value},topic:{deliveryResult.Topic},type:kafka_produce",
-                            deliveryResult.Offset.Value);
+                        var backlogTags = $"partition:{deliveryResult.Partition.Value},topic:{deliveryResult.Topic},type:kafka_produce";
+                        if (tags.ClusterId != null)
+                        {
+                            backlogTags = $"kafka_cluster_id:{tags.ClusterId},{backlogTags}";
```
andrewlock: See previous notes about avoiding an extra allocation here

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaProduceSyncDeliveryHandlerIntegration.cs
```diff
@@ -147,9 +147,13 @@ internal static Action<TDeliveryReport> WrapAction<TDeliveryReport>(Action<TDeli
                         if (!isError)
                         {
                             var dataStreams = Tracer.Instance.TracerManager.DataStreamsManager;
-                            dataStreams.TrackBacklog(
-                                $"partition:{report.Partition.Value},topic:{report.Topic},type:kafka_produce",
-                                report.Offset.Value);
+                            var backlogTags = $"partition:{report.Partition.Value},topic:{report.Topic},type:kafka_produce";
+                            if (tags.ClusterId != null)
+                            {
+                                backlogTags = $"kafka_cluster_id:{tags.ClusterId},{backlogTags}";
```
andrewlock: See previous notes about avoiding an extra allocation here

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/OffsetsCommittedCallbacks.cs
```diff
@@ -32,12 +32,23 @@ public void OnDelegateEnd(object? sender, Exception? exception, object? state)
...
+                }
+
                 for (var i = 0; i < committedOffsets?.Offsets.Count; i++)
                 {
                     var item = committedOffsets.Offsets[i];
-                    dataStreams.TrackBacklog(
-                        $"consumer_group:{GroupId},partition:{item.Partition.Value},topic:{item.Topic},type:kafka_commit",
-                        item.Offset.Value);
+                    var backlogTags = $"consumer_group:{GroupId},partition:{item.Partition.Value},topic:{item.Topic},type:kafka_commit";
+                    if (!string.IsNullOrEmpty(clusterId))
```
andrewlock: See previous notes about avoiding an extra allocation here

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/KafkaTests.cs
```diff
@@ -157,6 +157,14 @@ public async Task SubmitsTraces(string packageVersion, string metadataSchemaVers
                     x => x.Tags[Tags.KafkaConsumerGroup] == "Samples.Kafka.AutoCommitConsumer1"
                       || x.Tags[Tags.KafkaConsumerGroup] == "Samples.Kafka.ManualCommitConsumer2");
 
+            // cluster_id is only available with Confluent.Kafka 2.0+ (DescribeClusterAsync)
```
andrewlock: Right... I forgot this 🤔 I think we should _potentially_ split our integrations to account for this, because otherwise we're going to be doing expensive work which we _know_ is going to fail when we try to describe the cluster.

There's two ways to do that, depending on how reliably we can detect what version we're in from the KafkaHelper:

- Extract the common code out and create a duplicate of the integration
  - One integration targets 1.x.x only, and does the existing stuff
  - The other integration targets 2.x.x+, and includes the describe cluster work
- Pro-actively try to grab the `DescribeClusterOptions` type - if it doesn't exist, there's no point trying to build and duck type an `AdminClient`, so we can (and should) just bail at that point. What's more, we know that can _never_ work, so we should probably cache a `null`/`""` value for the bootstrap servers (and potentially bypass that cache entirely - just _always_ return `null` if that lookup of the `Type` has failed...)

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaConsumerConstructorIntegration.cs
zacharycmontoya: +1 to this question. It seems like you're using the entry `string.Empty` to indicate that yes you've done the lookup but no there's not a valid cluster ID we can use. In this case, I suppose it makes sense and I would also recommend updated `KafkaHelper.GetClusterId()` for consistency

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -212,10 +218,16 @@ private static long GetMessageSize<T>(T message)
                     tags.Offset = offset.ToString();
                 }
 
-                if (ConsumerCache.TryGetConsumerGroup(consumer, out var groupId, out var bootstrapServers))
+                var consumerClusterId = string.Empty;
+                if (ConsumerCache.TryGetConsumerGroup(consumer, out var groupId, out var bootstrapServers, out var clusterId))
```
zacharycmontoya: I think you can avoid having two variables and consolidate to one:
```suggestion
                if (ConsumerCache.TryGetConsumerGroup(consumer, out var groupId, out var bootstrapServers, out var consumerClusterId))
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -339,9 +363,26 @@ internal static void TryInjectHeaders<TTopicPartitionMarker, TMessage>(
 
                 if (dataStreamsManager.IsEnabled)
                 {
-                    var edgeTags = string.IsNullOrEmpty(topic)
+                    var producerClusterId = string.Empty;
+                    if (ProducerCache.TryGetProducer(producer, out _, out var clusterId) && !string.IsNullOrEmpty(clusterId))
+                    {
+                        producerClusterId = clusterId;
+                    }
```
zacharycmontoya: Similar feedback here for the producer side: I think you can consolidate down to one variable since you are already going to check `string.IsNullOrEmpty` anyways

```suggestion
                    ProducerCache.TryGetProducer(producer, out _, out var producerClusterId);
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
zacharycmontoya: +1 ugly as hell, +1 we should definitely do this.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaConsumerConstructorIntegration.cs
robcarlan-datadog: As far as I know..

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
robcarlan-datadog: The ordering specifically? (I think you were right about needing the ordering of tags in general and I updated those)

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
robcarlan-datadog: Done! 😄

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/KafkaTests.cs
robcarlan-datadog: I like the second approach in terms of simplicity, I'll do that. Flagging an early null/"" result in that case seems like a good thing to do

robcarlan-datadog: (My original comment was wrong and the version requirement is actually > 2.3.0)

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -365,6 +400,100 @@ internal static void TryInjectHeaders<TTopicPartitionMarker, TMessage>(
...
+            }
+
+            if (ClusterIdCache.TryGetValue(bootstrapServers, out var cached))
+            {
+                return cached;
+            }
+
+            try
+            {
+                var kafkaAssembly = clientInstance.GetType().Assembly;
```
andrewlock, line 419: 🤔 I'm just thinking... does this definitely work correctly? The `clientInstance` will be a Consumer/Producer right? Do customers ever _derive_ from these types? I _think_ they're always `internal` and therefore can't be derived, rigth? in which case this is fine 🙂

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -365,6 +400,100 @@ internal static void TryInjectHeaders<TTopicPartitionMarker, TMessage>(
...
+                    Log.Debug("Confluent.Kafka.Admin.DescribeClusterOptions not found; cluster_id tag requires Confluent.Kafka >= 2.3.0");
+                    ClusterIdCache.TryAdd(bootstrapServers, string.Empty);
+                    return string.Empty;
+                }
+
+                var builderType = kafkaAssembly.GetType("Confluent.Kafka.DependentAdminClientBuilder");
+                if (builderType is null)
+                {
+                    Log.Error("Failed to find Confluent.Kafka.DependentAdminClientBuilder in the Confluent.Kafka assembly");
+                    return null;
```
andrewlock: I think these should probably add to the cache and return `string.Empty` too, otherwise we're going to keep trying to do this, and flood with Error logs, even though it will never suceed? 🤔

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -365,6 +400,100 @@ internal static void TryInjectHeaders<TTopicPartitionMarker, TMessage>(
...
+                }
+
+                var clientHandle = clientInstance.DuckCast<IClientHandle>();
+                var handle = clientHandle.Handle;
+                if (handle is null)
+                {
+                    Log.Error("Kafka client handle is null; unable to extract cluster_id");
+                    return null;
```
andrewlock, line 443: This is an interesting one - I guess in _this_ case we probably _Shouldn't_ return `""`, because an error here is related to a specific client instance. The handle _shouldn't_ ever be `null` I think though right, so we don't expect this to ever _actually_ ping us

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
andrewlock: Now that we're explicitly handling the "too low" case, maybe we should make this an `Error` log again so we get error tracking?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -365,6 +400,100 @@ internal static void TryInjectHeaders<TTopicPartitionMarker, TMessage>(
...
+            catch (Exception ex)
+            {
+                Log.Debug(ex, "Error extracting cluster_id from Kafka metadata");
+                return null;
+            }
+        }
+
+        private static string? DescribeClusterWithTimeout(IAdminClient adminClient)
+        {
+            var optionsType = adminClient.Instance?.GetType().Assembly.GetType("Confluent.Kafka.Admin.DescribeClusterOptions");
```
andrewlock: nit: you can pass the `Assembly` in as an arg from the caller, seeing as you've already grabbed it 🙂 (It _is_ the same assembly I believe?)

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
```diff
@@ -365,6 +400,100 @@ internal static void TryInjectHeaders<TTopicPartitionMarker, TMessage>(
...
+            if (optionsType is null)
+            {
+                Log.Error("Failed to find Confluent.Kafka.Admin.DescribeClusterOptions in the Confluent.Kafka assembly");
+                return null;
```
andrewlock: This returns null, which we then cache in the `ClusterIdCache` - should we actually return `string.Empty` in this case?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
robcarlan-datadog: Yes! I originally did this to keep the signature simpler but it's probably better to avoid doing the duplicate work

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
robcarlan-datadog: Actually I can do one better here, we checked for the same type in the caller so I can pass the nuderlying type through rather than the `Assembly`, and know that it won't be null

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
robcarlan-datadog: Good point, added 😄

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
robcarlan-datadog: That's a good point, yes they are internal: https://github.com/confluentinc/confluent-kafka-dotnet/blob/master/src/Confluent.Kafka/Consumer.cs & https://github.com/confluentinc/confluent-kafka-dotnet/blob/master/src/Confluent.Kafka/Producer.cs

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Kafka/KafkaHelper.cs
zacharycmontoya: +1

PR #7705, opened 2025-10-24
[Profiler] Don't take tracer stable config notification when kill switch detected

chrisnas:
## Summary of changes
Discard tracer notification of Stable Configuration

## Reason for change
Avoid unexpected restart of the services

## Implementation details
Check kill switch also in the P/Invoke code

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
@@ -74,8 +74,6 @@ namespace Datadog.Trace.ClrProfiler
                 return;
             }
 
-            var tracer = Tracer.Instance;
-            var tracerSettings = tracer.Settings;
             var profilerSettings = Profiler.Instance.Settings;
             if (!profilerSettings.IsManagedActivationEnabled)
             {
@@ -84,6 +82,8 @@ namespace Datadog.Trace.ClrProfiler
             }
 
             Log.Debug("Setting Stable Configuration in Continuous Profiler native library.");
+            var tracer = Tracer.Instance;
+            var tracerSettings = tracer.Settings;
             NativeInterop.SharedConfig config = new NativeInterop.SharedConfig
             {
                 ProfilingEnabled = profilerSettings.ProfilerState switch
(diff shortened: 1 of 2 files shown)


tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
```diff
@@ -74,11 +74,16 @@ private static void PropagateStableConfiguration()
...
             var tracer = Tracer.Instance;
             var tracerSettings = tracer.Settings;
             var profilerSettings = Profiler.Instance.Settings;
+            if (!profilerSettings.IsManagedActivationEnabled)
+            {
+                Log.Debug("Set Stable Configuration in Continuous Profiler native library is disabled.");
+                return;
+            }
 
+            Log.Debug("Setting Stable Configuration in Continuous Profiler native library.");
```
andrewlock, line 84: nit: may as well move the grabbing of tracerSettings etc till afterwards too
```suggestion
            var profilerSettings = Profiler.Instance.Settings;
            if (!profilerSettings.IsManagedActivationEnabled)
            {
                Log.Debug("Set Stable Configuration in Continuous Profiler native library is disabled.");
                return;
            }

            Log.Debug("Setting Stable Configuration in Continuous Profiler native library.");
            var tracer = Tracer.Instance;
            var tracerSettings = tracer.Settings;
```

tracer/src/Datadog.Trace/ContinuousProfiler/ProfilerSettings.cs
```diff
@@ -13,6 +13,8 @@ namespace Datadog.Trace.ContinuousProfiler;
 
 internal class ProfilerSettings
 {
+    private bool _isManagedActivationEnabled = false;
```
andrewlock: nit:
```suggestion
    private readonly bool _isManagedActivationEnabled = false;
```

andrewlock (approved):
Thanks!

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
chrisnas: ✅ done

tracer/src/Datadog.Trace/ContinuousProfiler/ProfilerSettings.cs
chrisnas: ✅ done

lucaspimentel:
@codex review

gleocadie (approved):
LGTM

PR #4298, opened 2023-06-21
[Profiler] Propagate git repository url and commit hash from tracer to profiler

gleocadie:
## Summary of changes

Propagate `git.repository_url` and `git.commit.sha` from the Tracer to the Profiler.

## Reason for change

The goal of this PR is to allow Source Code integration on profiles and share the git metadata information between the Tracer and the Profiler (Hard things have already been done, thanks @OmerRaviv  ;) let's just reuse it)

## Implementation details

* When the Tracer extracts git metadata, calls (PInvoke) the Profiler to share the information (if the feature is enabled)
* If the feature is disabled on the Tracer side, let's just fallback on Environment variables
* Add `git.repository_url` and `git.commit.sha` to the profile.

## Test coverage
* Add integration tests
* Add unit tests.

PR #7715, opened 2025-10-27
Add process tags to profiles

vandonr:
## Summary of changes

Write [process tags](https://docs.google.com/document/d/1AFdLUuVk70i0bJd5335-RxqsvwAV9ovAqcO2z5mEMbA) to profiles

## Reason for change

[OLDAIDM-701]
[AIDM-195]

## Implementation details

Most of the C++ code is AI generated, fed with PR #4298 

## Other details

[OLDAIDM-701]: https://datadoghq.atlassian.net/browse/OLDAIDM-701?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/profiler/test/Datadog.Profiler.IntegrationTests/ProcessTagsTest.cs
+++ b/profiler/test/Datadog.Profiler.IntegrationTests/ProcessTagsTest.cs
@@ -20,7 +20,7 @@ namespace Datadog.Profiler.IntegrationTests.ProcessTags
     public class ProcessTagsTest
     {
         private const string Scenario = "--scenario 2";
-        private static readonly Regex ProcessTagsPattern = new("process_tags:(?<process_tags>[^\\r\\n\"]+)", RegexOptions.Compiled | RegexOptions.Multiline | RegexOptions.IgnoreCase);
+        private static readonly Regex ProcessTagsPattern = new("\"process_tags\":\"(?<process_tags>[^\"]+)", RegexOptions.Compiled | RegexOptions.Multiline | RegexOptions.IgnoreCase);
 
         private readonly ITestOutputHelper _output;
 
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native/AgentProxy.hpp
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native/AgentProxy.hpp
@@ -159,7 +159,7 @@ private:
                 &_exporter, encodedProfile,
                 to_compress_files_view, uncompressed_files_view,
                 static_cast<ddog_Vec_Tag const*>(*tags._impl),
-                pMetadata, pInfo, pProcessTags);
+                pProcessTags, pMetadata, pInfo);
 
         if (requestResult.tag == DDOG_PROF_REQUEST_RESULT_ERR_HANDLE_REQUEST)
         {


tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -68,4 +77,20 @@ private static string NormalizeTagValue(string tagValue)
         // which we don't want because we use it as a key/value separator.
         return TraceUtil.NormalizeTag(tagValue).Replace(oldChar: ':', newChar: '_');
     }
+
+    private static void PropagateProcessTagsToProfiler(string values)
+    {
+        try
+        {
+            // Avoid P/Invoke if the profiler is not ready
+            if (Profiler.Instance.Status.IsProfilerReady && !string.IsNullOrEmpty(values))
```
vandonr: not too sure about this. Because of where it's called, there is only 1 shot at this, so if it's called too early (if that's a thing), then we would never propagate the tags. I wonder if I shouldn't call this method in another place where I could be more sure that the profiler has had time to get ready.

andrewlock: IMO we should almost certainly not be creating another interop call here. We should be adding the process tags to the existing interop calls we make that propagate the config to the profiler. I suspect the correct place is the existing PInvoke into the profiler that happens in `Instrumentation`.

cc @chrisnas

vandonr: ah well good point

vandonr: I'm thinking I should do it here:
https://github.com/DataDog/dd-trace-dotnet/blob/2f5de7cd246188c79dea4c4f282403f6655d1a3e/tracer/src/Datadog.Trace/TracerManagerFactory.cs#L79
the reason I did it this way initially is that I wrote this code when the lazy was still in place, but now it doesn't make sense anymore.

andrewlock: That's called every time the customer makes a change, which can _technically_ can happen a lot. Given the process tags can't change, I think it makes more sense to do it here? 

https://github.com/DataDog/dd-trace-dotnet/blob/b902cd8f691b0ac1a39bd850d18d27de8fdfed64/tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs#L64

andrewlock: Also, bear in mind you can't just change/remove the `SetApplicationInfoForAppDomain` native call on the native side, because v2 apps can call into it. You _can_ change the one in `Instrumentation` though I believe, because it was only added in v3, and there's no version conflict there 😉

vandonr: > Given the process tags can't change, I think it makes more sense to do it here?

hmm, yeah true. Though process tags can't change _for now_ but I don't know if we want to rely on this notion too much, because I see a close future where it'll be like "actually getting the entrypoint name for IIS is important, we should write it to the process tags when we have it" or something like that

vandonr: > because v2 apps can call into it

wait you can have an app with the tracer in v2 and the profiler in v3 ?

vandonr: I figured I could always add a path to edit the value later

profiler/test/Datadog.Profiler.IntegrationTests/ProcessTagsTest.cs
```diff
@@ -0,0 +1,151 @@
...
+using MessagePack;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Datadog.Profiler.IntegrationTests.DebugInfo
+{
+    public class ProcessTagsTest
+    {
+        private const string Scenario = "--scenario 2";
+        private static readonly Regex ProcessTagsPattern = new("_dd\\.tags\\.process:(?<process_tags>[^,\"]+)", RegexOptions.Compiled | RegexOptions.Multiline | RegexOptions.IgnoreCase);
```
wantsui: Looks like Java added it as "process_tags" without the `_dd.tags.process": https://github.com/DataDog/dd-trace-java/pull/8780/files

vandonr: good catch, when I wrote this I assumed we'd use the same syntax as for spans, but it is not the case !

tracer/src/Datadog.Trace/ProcessTags.cs
andrewlock: > wait you can have an app with the tracer in v2 and the profiler in v3 ?

We can have an app with v2 of the tracer _and_ v3 of the tracer in it 😢

vandonr:
putting it back in draft because I just saw with Gregory that we need to put the tag in a dedicated field that requires a change in libdatadog, an issue that java and go didn't have.

PR #7716, opened 2025-10-27
Revert "[Tracer] Add Azure Event Hubs instrumentation "

bouwkast:
Reverts DataDog/dd-trace-dotnet#7620

This pulls in new Docker containers that we don't have on the VMs and it appears to have broken CI


bouwkast:
Looks to be unrelated actually and may be due to the RC 2 update

PR #7717, opened 2025-10-27
SSR Implementation for .NET

rithikanarayan:


tracer/src/Datadog.Trace/Tagging/AwsDynamoDbTags.cs
```diff
@@ -28,10 +28,10 @@ internal partial class AwsDynamoDbV1Tags : AwsDynamoDbTags
         // However, this can still be set from ITags.SetTag so the user can
         // customize the value if they wish.
         [Tag(Trace.Tags.PeerService)]
```
lucaspimentel: I think we need to remove this `[Tag(Trace.Tags.PeerService)]` for every derived class, since it's already included in the base class `AwsSdkTags`.

PR #7719, opened 2025-10-28
Protect against `claimsPrincipal` being `null`

bouwkast:
## Summary of changes

We are getting `NullReferenceException` in `HttpContextSetUser`, guessing the `claimsPrincipal` can be `null`.

## Reason for change

```
System.NullReferenceException
   at Datadog.Trace.ClrProfiler.AutoInstrumentation.AspNetCore.UserEvents.HttpContextSetUser.OnMethodBegin[TTarget](TTarget instance, ClaimsPrincipal& claimsPrincipal)
   at Microsoft.AspNetCore.Http.DefaultHttpContext.set_User(ClaimsPrincipal value)
```

## Implementation details

Added `claimsPrincipal is not null`

## Test coverage

None 🤷

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AspNetCore/UserEvents/HttpContextSetUser.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AspNetCore/UserEvents/HttpContextSetUser.cs
@@ -43,7 +43,7 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.AspNetCore.UserEvents
 
         private const string HttpContextExtensionsTypeName = "Microsoft.AspNetCore.Http.DefaultHttpContext";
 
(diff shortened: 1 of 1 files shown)


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AspNetCore/UserEvents/HttpContextSetUser.cs
andrewlock: If this can be null, we should probably mark it as such
```suggestion
        internal static CallTargetState OnMethodBegin<TTarget>(TTarget instance, ref ClaimsPrincipal? claimsPrincipal)
```

bouwkast: Ah good point

I don't know why but it won't let me commit this suggestion

bouwkast: Weird had to do it on a different page

PR #7720, opened 2025-10-28
Check if `httpContext.Response` is not `null`

bouwkast:
## Summary of changes

Saw a couple of of `NullReferenceException` from calling things in ` HttpContext.Response.` so checking if `Response` is not `null`

## Reason for change

```
Error : Event Exception: {EventName}
System.NullReferenceException
   at Microsoft.AspNetCore.Http.DefaultHttpResponse.get_StatusCode()
   at Datadog.Trace.PlatformHelpers.AspNetCoreHttpRequestHandler.StopAspNetCorePipelineScope(Tracer tracer, Security security, Scope rootScope, HttpContext httpContext)
   at Datadog.Trace.DiagnosticListeners.AspNetCoreDiagnosticObserver.OnHostingHttpRequestInStop(Object arg)
   at Datadog.Trace.DiagnosticListeners.AspNetCoreDiagnosticObserver.OnNext(String eventName, Object arg)
   at Datadog.Trace.DiagnosticListeners.DiagnosticObserver.System.IObserver<System.Collections.Generic.KeyValuePair<System.String,System.Object>>.OnNext(KeyValuePair`2 value)
```

```
Error : Event Exception: {EventName}
System.NullReferenceException
   at Datadog.Trace.PlatformHelpers.AspNetCoreHttpRequestHandler.StopAspNetCorePipelineScope(Tracer tracer, Security security, Scope rootScope, HttpContext httpContext)
   at Datadog.Trace.DiagnosticListeners.AspNetCoreDiagnosticObserver.OnHostingHttpRequestInStop(Object arg)
   at Datadog.Trace.DiagnosticListeners.DiagnosticObserver.System.IObserver<System.Collections.Generic.KeyValuePair<System.String,System.Object>>.OnNext(KeyValuePair`2 value)
```

```
Error : Event Exception: {EventName}
System.NullReferenceException
   at Datadog.Trace.DiagnosticListeners.AspNetCoreDiagnosticObserver.OnHostingHttpRequestInStop(Object arg)
   at Datadog.Trace.DiagnosticListeners.DiagnosticObserver.System.IObserver<System.Collections.Generic.KeyValuePair<System.String,System.Object>>.OnNext(KeyValuePair`2 value)
```

## Implementation details

Added some null checks / protections

## Test coverage

Looked at https://github.com/DataDog/dd-trace-dotnet/blob/5f33ec08a4789ebf123c5ea1b5430ccadd38b7ce/tracer/test/Datadog.Trace.Tests/PlatformHelpers/AspNetCoreHttpRequestHandlerTests.cs#L15, but it doesn't look to be super straightforward to test.

### DIFF
--- a/tracer/src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
+++ b/tracer/src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
@@ -713,8 +713,9 @@ namespace Datadog.Trace.DiagnosticListeners
                 return;
             }
 
-            if (arg.DuckCast<HttpRequestInStopStruct>().HttpContext is { } httpContext
-             && httpContext.Items[AspNetCoreHttpRequestHandler.HttpContextTrackingKey] is AspNetCoreHttpRequestHandler.RequestTrackingFeature { RootScope: { } rootScope })
+            if (arg.TryDuckCast<HttpRequestInStopStruct>(out var stopStruct)
+             && stopStruct.HttpContext is { } httpContext
+             && httpContext.Features.Get<AspNetCoreHttpRequestHandler.RequestTrackingFeature>() is { RootScope: { } rootScope })
             {
                 AspNetCoreRequestHandler.StopAspNetCorePipelineScope(_tracer, _security, rootScope, httpContext);
             }
--- a/tracer/src/Datadog.Trace/AppSec/Coordinator/SecurityCoordinatorHelpers.Core.cs
+++ b/tracer/src/Datadog.Trace/AppSec/Coordinator/SecurityCoordinatorHelpers.Core.cs
@@ -44,7 +44,7 @@ internal static class SecurityCoordinatorHelpers
             if (security.AppsecEnabled && CoreHttpContextStore.Instance.Get() is { } httpContext)
             {
                 var transport = new SecurityCoordinator.HttpTransport(httpContext);
-                if (!transport.IsBlocked)
+                if (!transport.IsBlocked && httpContext.Response is not null)
                 {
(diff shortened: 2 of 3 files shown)


tracer/src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
```diff
@@ -754,7 +754,8 @@ private void OnHostingHttpRequestInStop(object arg)
                 return;
             }
 
-            if (arg.DuckCast<HttpRequestInStopStruct>().HttpContext is { } httpContext
+            if (arg.TryDuckCast<HttpRequestInStopStruct>(out var stopStruct)
+             && stopStruct.HttpContext is { } httpContext
```
andrewlock: I don't think you need to switch to `TryDuckCast` here - if `arg` is `null`, that's a pass-through for `DuckCast<T>`? 🤔 
```suggestion
            if (arg.DuckCast<HttpRequestInStopStruct>()?.HttpContext is { } httpContext
```

bouwkast: Ah I wasn't sure if `DuckCast` would throw but I'll do that thanks!

bouwkast:
> It makes me so sad we have to handle this. `HttpContext.Response` is non-`null` [by definition](https://github.com/dotnet/aspnetcore/blob/main/src/Http/Http.Abstractions/src/HttpContext.cs#L32) 🙁

Yeah I'm not sure exactly how this comes about? 

Should we opt to accept a level of exceptions here as it was quite low volume?

NachoEchevarria (approved):

PR #7721, opened 2025-10-28
Enforce that you can't update exporter settings to point to UDS on Windows

andrewlock:
## Summary of changes

Enforces that you can't _change_ the `AgentUri` to be a UDS Uri if you're on Windows

## Reason for change

The trace exporter doesn't work with UDS on Windows, so we have a check in `TracerSettings` that disables the pipeline if we find this scenario. However, user's can still _change_ the agent URI at runtime in code (😭).

We currently assume that the data pipeline won't be toggled at runtime (we _do_ allow for reconfiguring it in general, but not for completely removing or reintroducing). Changing this to allow the scenario would be a pain, so instead this PR blocks you from setting a UDS URI in code if you're on Windows.

The good news is that as far as I can tell, noone does this today, so while _technically_ it could be considered a breaking change, I think it's ok.

## Implementation details

- Throw an `ArgumentException` in the Datadog.Trace.Manual library, if you're on Windows (or .NET FX) and you try to set a UDS agent URI (using the same "detection" we do in `ExporterSettings`.
- Add a check in the Instrumentation of `Tracer.Configure()` to make sure it hasn't slipped through. This could happen if a customer was using an old version of the Datadog.Trace NuGet package with a newer version of auto instrumentation.

Note that this adds two additional framework references for .NET Core 3.1+, to check if we're on Windows.

## Test coverage

Added an extra step to the manual instrumentation integration test to confirm we throw

## Other details

Part of a config stack

- https://github.com/DataDog/dd-trace-dotnet/pull/7721 👈

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/ManualInstrumentation/Tracer/ConfigureIntegration.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/ManualInstrumentation/Tracer/ConfigureIntegration.cs
@@ -55,7 +55,7 @@ public class ConfigureIntegration
         if (FrameworkDescription.Instance.IsWindows()
          && values.TryGetValue(TracerSettingKeyConstants.AgentUriKey, out var raw)
          && raw is Uri uri
-         && uri.ToString().StartsWith(ExporterSettings.UnixDomainSocketPrefix, StringComparison.OrdinalIgnoreCase))
+         && uri.OriginalString.StartsWith(ExporterSettings.UnixDomainSocketPrefix, StringComparison.OrdinalIgnoreCase))
         {
             InvalidConfigurationException.Throw(
                 $"Error changing AgentUri. " +


andrewlock (commented):
## Code Review by Claude Code

I've reviewed this PR and overall it looks like a solid defensive fix to prevent runtime configuration issues with UDS on Windows. Here are my findings:

### ✅ Strengths

1. **Dual-layer validation**: Good defense-in-depth approach with validation in both the manual instrumentation library and auto-instrumentation layer
2. **Clear error messages**: Users get helpful guidance on how to work around the limitation
3. **Good test coverage**: The manual instrumentation test properly validates the exception is thrown on Windows
4. **Consistent exception handling**: Using `CallTargetBubbleUpException` ensures proper exception propagation

### 🔍 Issues Found

#### 1. Inconsistent URI checking between files

**Location**: `TracerSettings.cs:406` vs `ConfigureIntegration.cs:58`

In `TracerSettings.cs`, you use:
```csharp
value.OriginalString.StartsWith("unix://", StringComparison.OrdinalIgnoreCase)
```

But in `ConfigureIntegration.cs`, you use:
```csharp
uri.ToString().StartsWith(ExporterSettings.UnixDomainSocketPrefix, StringComparison.OrdinalIgnoreCase)
```

**Recommendation**: Use `OriginalString` consistently (as in TracerSettings) since it represents what the user actually provided. Also consider extracting the "unix://" prefix to a shared constant.

#### 2. Inconsistent error messages

The error messages are slightly different between the two validation points:
- TracerSettings: "Error setting {nameof(AgentUri)} to '{value}'. {nameof(AgentUri)} can not be set to..."
- ConfigureIntegration: "Error changing AgentUri. AgentUri can not be set to..."

**Recommendation**: Consider using a shared error message or at least ensuring they're semantically equivalent for consistency.

#### 3. .NET Framework behavior might be too restrictive

**Location**: `TracerSettings.cs:402-405`

On .NET Framework, the code blocks UDS URIs without checking the OS. While Windows is the primary platform for .NET Framework, this technically blocks UDS even if somehow running on a non-Windows OS with .NET Framework.

**Question**: Is this intentional? If .NET Framework only runs on Windows in your supported scenarios, this is fine. Otherwise, you might want to add OS detection for .NET Framework too.

---

**Review performed by Claude Code** 🤖

This PR prevents a runtime configuration issue and the implementation is sound. The issues noted are minor consistency/robustness improvements rather than critical bugs.

tracer/src/Datadog.Trace.Manual/Configuration/TracerSettings.cs
```diff
@@ -397,7 +397,23 @@ public Uri AgentUri
...
-        set => _agentUri = _agentUri.Override(value);
+        set
+        {
+            if (
+#if !NETFRAMEWORK
+                System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform(System.Runtime.InteropServices.OSPlatform.Windows) &&
+#endif
+                value.OriginalString.StartsWith("unix://", StringComparison.OrdinalIgnoreCase))
+            {
+                throw new ArgumentException(
```
lucaspimentel, line 408: Breaking change? I assume it's ok because it's better than the current behavior?

lucaspimentel, line 408: Should this use the throw helper instead of using `throw` directly?

lucaspimentel: Oh, not available in `Datadog.Trace.Manual`?

andrewlock: Yeah, from the description 😄 

> The good news is that as far as I can tell, noone does this today, so while technically it could be considered a breaking change, I think it's ok.

andrewlock: Yeah, this is throwing an `ArgumentException`, not our special `CallTargetBubbleUpException` that we have to make sure to throw in our integrations. `ArgumentException` seems like the correct option to me here?

lucaspimentel: > from the description

I see what you did there. 🤦🏽‍♂️

lucaspimentel: I meant the `ThrowHelper` helper methods we use to throw exceptions to improve the chances that a method can be inlined by avoiding a direct `throw`. [source](https://github.com/DataDog/dd-trace-dotnet/blob/05622b1cca4b9d28819e4785791292491fed4b8c/tracer/src/Datadog.Trace/Util/ThrowHelper.cs#L36-L44)

andrewlock: Oh right, you meant those ones, yeah, meh 😄 especially given we mark everything as NoInlining 😅

PR #7722, opened 2025-10-28
Add helper `DictionaryExtensions.SequenceEqual`

andrewlock:
## Summary of changes

Add a helper for comparing `ReadOnlyDictionary<>` instances

## Reason for change

As part of the config work, we need to detect if tags have changed when customers do a manual/remote config update. This helper makes it easy

## Implementation details

Added a `SequenceEqual` extension method.

Note that I used `SequenceEqual` because it _already_ exists in _System.Linq_, but I could see an argument that it's too easy to use the wrong one, and instead we could use a different name? `IsSameAs(other)`?

Also, I only wrote this for `ReadOnlyDictionary<>` because that's all we need, it's what we use for all our setting dictionaries, and it will be (a tiny bit) faster than making it `IDictionary<>`, but happy to change if people feel strongly.

## Test coverage

Added unit tests

## Other details

Part of a config stack

- https://github.com/DataDog/dd-trace-dotnet/pull/7722 👈

### DIFF
--- a/tracer/src/Datadog.Trace/ExtensionMethods/DictionaryExtensions.cs
+++ b/tracer/src/Datadog.Trace/ExtensionMethods/DictionaryExtensions.cs
@@ -141,7 +141,7 @@ namespace Datadog.Trace.ExtensionMethods
             ReadOnlyDictionary<string, string> dict2,
             StringComparison valueComparison = StringComparison.Ordinal)
         {
-            if (dict1 is null & dict2 is null)
+            if (dict1 is null && dict2 is null)
             {
                 return true;
             }


tracer/src/Datadog.Trace/ExtensionMethods/DictionaryExtensions.cs
```diff
@@ -128,5 +129,42 @@ public static TV Get<TK, TV>(this Dictionary<TK, TV> map, TK key, Func<TK, TV> c
...
             return newVal;
         }
+
+        /// <summary>
+        /// Checks if two dictionaries contain the same keys and values.
+        /// Note that this method assumes the two dictionaries use the same
+        /// <see cref="IEqualityComparer"/>; using different comparers in each
+        /// dictionary is not supported.
+        /// </summary>
+        public static bool SequenceEqual(
```
lucaspimentel, line 139: The name `SequenceEqual` feels weird to me because it implies that it checks that the _order_ of the items is the same (for example in a `List<T>` or `T[]`), but items are not ordered in most dictionaries.

but level of care is low

lucaspimentel: Never mind.

https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.sequenceequal

> `true` if the two source sequences are of equal length and their corresponding elements are equal according to the default equality comparer for their type; otherwise, `false`.

andrewlock: To be fair, I'm not a big fan of `SequenceEqual()` either, I only used it because of the prior art. Happy to take suggestions: e.g. `IsSameAs()`? 🤷‍♂️

tracer/src/Datadog.Trace/ExtensionMethods/DictionaryExtensions.cs
```diff
@@ -128,5 +129,42 @@ public static TV Get<TK, TV>(this Dictionary<TK, TV> map, TK key, Func<TK, TV> c
...
+            this ReadOnlyDictionary<string, string> dict1,
+            ReadOnlyDictionary<string, string> dict2,
+            StringComparison valueComparison = StringComparison.Ordinal)
+        {
+            if (dict1 is null & dict2 is null)
```
bouwkast:
```suggestion
            if (dict1 is null && dict2 is null)
```

Unsure if there is a reason for the bitwise(logical?) AND?

andrewlock: typo 😅

tracer/src/Datadog.Trace/ExtensionMethods/DictionaryExtensions.cs
```diff
@@ -128,5 +129,42 @@ public static TV Get<TK, TV>(this Dictionary<TK, TV> map, TK key, Func<TK, TV> c
...
+
+        /// <summary>
+        /// Checks if two dictionaries contain the same keys and values.
+        /// Note that this method assumes the two dictionaries use the same
+        /// <see cref="IEqualityComparer"/>; using different comparers in each
+        /// dictionary is not supported.
+        /// </summary>
+        public static bool SequenceEqual(
+            this ReadOnlyDictionary<string, string> dict1,
+            ReadOnlyDictionary<string, string> dict2,
```
tonyredondo, line 141: any reason to not doing it in a Generic way?

andrewlock: Right now, mostly because it's unnecessary, so extra abstraction doesn't seem worth it 🤷‍♂️

andrewlock: Oops, auto merge just went ahead with it 😓

PR #7723, opened 2025-10-28
Remove properties from `TracerSettings` which can change at runtime

andrewlock:
## Summary of changes

Instead of exposing settings like `ServiceVerion` and `Environment` on `TracerSettings`, only expose them on `MutableSettings`

## Reason for change

This is all part of the refactoring towards having `TracerSettings` being an immutable settings object, that's created once on app startup, and fixed for the lifetime. `ServiceVerion` and `Environment` currently delegate to the `MutableSettings` object stored on `TracerSettings`, but are going to move that elsewhere shortly so that consumers can subscribe to changes

## Implementation details

For this PR, it's just a case of changing `TracerSettings.XXX` => `TracerSettings.Mutable.XXX`. A future PR will then subsequently "fix" these usages properly. This is just a small step to be able to remove the mutable properties from `TracerSettings`

## Test coverage

Just a refactoring, covered by existing tests.

## Other details

Part of a config stack

- https://github.com/DataDog/dd-trace-dotnet/pull/7723 👈

### DIFF
--- a/tracer/test/benchmarks/Benchmarks.Trace/Asm/EmptyDatadogTracer.cs
+++ b/tracer/test/benchmarks/Benchmarks.Trace/Asm/EmptyDatadogTracer.cs
@@ -26,7 +26,7 @@ namespace Benchmarks.Trace.Asm
         {
             Settings = new(new NullConfigurationSource());
             _gitMetadata = new NullGitMetadataProvider();
-            _perTraceSettings = new PerTraceSettings(null, null, null, Settings.InitialMutableSettings);
+            _perTraceSettings = new PerTraceSettings(null, null, null, Settings.Manager.InitialMutableSettings);
         }
 
         public string DefaultServiceName => "My Service Name";


andrewlock (commented):
**Automated code review by Claude Code**

This PR successfully refactors TracerSettings to remove mutable properties and require explicit access through MutableSettings. The changes are generally consistent and well-executed. I've identified a few specific issues below for consideration.

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -734,9 +734,10 @@ not null when string.Equals(value, "otlp", StringComparison.OrdinalIgnoreCase) =
             // We create a lazy here because this is kind of expensive, and we want to avoid calling it if we can
             _fallbackApplicationName = new(() => ApplicationNameHelpers.GetFallbackApplicationName(this));
 
-            InitialMutableSettings = MutableSettings.CreateInitialMutableSettings(source, telemetry, errorLog, this);
-            MutableSettings = InitialMutableSettings;
-            Manager = new(this, InitialMutableSettings, Exporter);
+            // Move the creation of these settings inside SettingsManager?
```
andrewlock, line 730: This comment suggests future work but doesn't have a tracking issue. Consider either:
1. Removing the comment if this is the intended final state
2. Converting it to a TODO with a JIRA ticket reference (like the other PR description links)
3. Creating a follow-up issue if this refactoring should happen

*Automated comment by Claude Code*

tracer/src/Datadog.Trace.BenchmarkDotNet/DatadogDiagnoser.cs
```diff
@@ -216,12 +216,12 @@ private static void SetEnvironmentVariables(DiagnoserActionParameters parameters
 
         if (!environment.TryGetValue(ConfigurationKeys.Environment, out _))
         {
-            environment[ConfigurationKeys.Environment] = tracer.Settings.Environment;
+            environment[ConfigurationKeys.Environment] = tracer.CurrentTraceSettings.Settings.Environment;
```
andrewlock, line 219: The access pattern `tracer.CurrentTraceSettings.Settings.Environment` is quite verbose (3 levels of property access). While technically correct (since `CurrentTraceSettings.Settings` returns `MutableSettings`), consider if there's a simpler way to expose these commonly-used properties, perhaps through a helper method or property on Tracer.

This same pattern appears on line 224 for ServiceVersion.

*Automated comment by Claude Code*

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -747,47 +748,24 @@ not null when string.Equals(value, "otlp", StringComparison.OrdinalIgnoreCase) =
 
         internal IConfigurationTelemetry Telemetry => _telemetry;
 
-        internal MutableSettings InitialMutableSettings { get; }
-
         internal MutableSettings MutableSettings { get; init; }
```
andrewlock, line 746: Minor observation: `MutableSettings` is marked with `init` accessor but is being set in the constructor. This is valid C# but consider whether this property could be moved to a constructor parameter with property initializer syntax for clarity, or if the `init` is necessary for the `with` expression support.

*Automated comment by Claude Code*

andrewlock (commented):
**Automated code review by Claude Code**

This PR successfully refactors TracerSettings to remove mutable properties and require explicit access through MutableSettings. The changes are generally consistent and well-executed.

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
andrewlock, line 730: This comment suggests future work but doesn't have a tracking issue. Consider either:
1. Removing the comment if this is the intended final state
2. Converting it to a TODO with a JIRA ticket reference (like the other PR description links)
3. Creating a follow-up issue if this refactoring should happen

*Automated comment by Claude Code*

tracer/src/Datadog.Trace.BenchmarkDotNet/DatadogDiagnoser.cs
andrewlock, line 219: The access pattern `tracer.CurrentTraceSettings.Settings.Environment` is quite verbose (3 levels of property access). While technically correct (since `CurrentTraceSettings.Settings` returns `MutableSettings`), consider if there's a simpler way to expose these commonly-used properties, perhaps through a helper method or property on Tracer.

This same pattern appears on line 224 for ServiceVersion.

*Automated comment by Claude Code*

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
andrewlock, line 746: Minor observation: `MutableSettings` is marked with `init` accessor but is being set in the constructor. This is valid C# but consider whether this property could be moved to a constructor parameter with property initializer syntax for clarity, or if the `init` is necessary for the `with` expression support.

*Automated comment by Claude Code*

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
andrewlock: meh, it's more a question for other people, future refactoring, not a big deal

tracer/src/Datadog.Trace.BenchmarkDotNet/DatadogDiagnoser.cs
andrewlock: Yeah, this bugs me a bit, but it's something we can improve later if needs be. We could (for example) expose these as extension methods directly on `Tracer` instances which automatically does the right thing for these one-off accesses. The thing is, this is _not_ the same as accessing the "initial" settings _or_ the "subscriber" approach, in that it reacts as necessary but is consistent _within_ traces

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
andrewlock: This is addressed in the subsequent PR, so it's fine

lucaspimentel (approved):

zacharycmontoya (approved):

e-n-0 (approved):

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -727,9 +727,10 @@ not null when string.Equals(value, "otlp", StringComparison.OrdinalIgnoreCase) =
             // We create a lazy here because this is kind of expensive, and we want to avoid calling it if we can
             _fallbackApplicationName = new(() => ApplicationNameHelpers.GetFallbackApplicationName(this));
 
-            InitialMutableSettings = MutableSettings.CreateInitialMutableSettings(source, telemetry, errorLog, this);
-            MutableSettings = InitialMutableSettings;
-            Manager = new(this, InitialMutableSettings, Exporter);
+            // Move the creation of these settings inside SettingsManager?
+            var initialMutableSettings = MutableSettings.CreateInitialMutableSettings(source, telemetry, errorLog, this);
+            Manager = new(this, initialMutableSettings, Exporter);
+            MutableSettings = initialMutableSettings;
```
bouwkast, line 733: Was the order of operations here changed on purpose? `Manager`  / `MutableSettings` got flipped around but I don't see anything really so just wondering

andrewlock: No, no reason, I think that was just accidental in amongst all the refactoring 😅 It's moot in the next PR where we remove `MutableSettings` entirely 😅

tonyredondo (approved):

PR #7724, opened 2025-10-28
Handle changes to `MutableSettings` and `ExporterSettings` without rebuilding

andrewlock:
## Summary of changes

- This is the big one
- Update services to dynamically update when mutable settings or exporter settings change
- Stop rebuilding everything when there's manual/remote configuration

## Reason for change

This is the "endpoint" that we've been heading for - services only being disposed/rebuilt at the end of the app, and otherwise only rebuilding the _necessary_ parts. For example - we don't need to tear down all the API factories when a customer changes a global tag via remote config; they only need to change if the `ExporterSettings` change.

The hope is that overall this reduces the overhead of using configuration in code and/or remote configuration, while also reducing the number of issues due to managing disposal of services.

## Implementation details

Overall, this PR is kind of a pain. Moving from the "rebuild everything" to "reconfigure each service" couldn't be done piecemeal, so this is the one-shot PR. What's more, different services need different patterns (though we can probably consolidate some of them, this has taken a _lot_ of work and I likely changed patterns unnecessarily in some places).

In general, there's a couple of patterns:

- CI Vis doesn't let you change settings at runtime, so it _never_ needs to respond to changes. It always just uses the "initial" settings
- Debugger _today_ doesn't respond to changes at runtime (except its own dynamic config), so for now we ignore Debugger too as it's not really a regression. I hope we can fix this soon though.
- I've introduced the concept of `Managed*` versions of some services
  - These services generally "wrap" the existing type, delegating access to the underlying service, and handling settings changes
- Many services only care about a sub-set of mutable settings, so they only update if they need to
- Somewhat annoyingly, setting updates occur on a background thread, so we need to be careful about thread safety. Where necessary (most places) I've made sure access to a now-mutable service is done using `Volatile.Read()` (to ensure changes are visible) and are generally cached to a local variable (as the underlying field may be updated in the background).

## Test coverage

In the vast majority of places, this should be covered by existing tests

I plan to add some additional integration tests around reconfiguring and a bunch of manual testing to make sure I'm confident.

## Other details

I strongly recommend reviewing commit-by-commit. They're generally self-contained, and hopefully simple enough to understand one commit at a time.

Part of a config stack

- https://github.com/DataDog/dd-trace-dotnet/pull/7724 👈

This isn't the final PR in the stack, as there will be a bunch of cleaning up to do, but it's the final "implementation" PR

### DIFF
--- a/tracer/src/Datadog.Trace/Configuration/MutableSettings.cs
+++ b/tracer/src/Datadog.Trace/Configuration/MutableSettings.cs
@@ -1074,10 +1074,10 @@ internal sealed class MutableSettings : IEquatable<MutableSettings>
     /// by excluding all the default sources. Effectively gives all the settings their default
     /// values. Should only be used with the manual instrumentation source
     /// </summary>
-    public static MutableSettings CreateWithoutDefaultSources(TracerSettings tracerSettings)
+    public static MutableSettings CreateWithoutDefaultSources(TracerSettings tracerSettings, ConfigurationTelemetry telemetry)
         => CreateInitialMutableSettings(
             NullConfigurationSource.Instance,
-            new ConfigurationTelemetry(),
+            telemetry,
             new OverrideErrorLog(),
             tracerSettings);
 
--- a/tracer/src/Datadog.Trace/Configuration/SettingsManager.cs
+++ b/tracer/src/Datadog.Trace/Configuration/SettingsManager.cs
@@ -7,41 +7,56 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Threading;
 using Datadog.Trace.Configuration.ConfigurationSources;
 using Datadog.Trace.Configuration.ConfigurationSources.Telemetry;
 using Datadog.Trace.Configuration.Telemetry;
-using Datadog.Trace.Logging;
 
 namespace Datadog.Trace.Configuration;
 
 public partial record TracerSettings
 {
-    internal class SettingsManager(
-        TracerSettings tracerSettings,
-        MutableSettings initialMutable,
-        ExporterSettings initialExporter)
+    internal class SettingsManager
     {
-        private readonly TracerSettings _tracerSettings = tracerSettings;
+        private readonly TracerSettings _tracerSettings;
+        private readonly ConfigurationTelemetry _initialTelemetry;
         private readonly List<SettingChangeSubscription> _subscribers = [];
 
         private IConfigurationSource _dynamicConfigurationSource = NullConfigurationSource.Instance;
         private ManualInstrumentationConfigurationSourceBase _manualConfigurationSource =
             new ManualInstrumentationConfigurationSource(new Dictionary<string, object?>(), useDefaultSources: true);
 
+        // We delay creating these, as we likely won't need them
+        private ConfigurationTelemetry? _noDefaultSettingsTelemetry;
+        private MutableSettings? _noDefaultSourcesSettings;
+
         private SettingChanges? _latest;
+
+        public SettingsManager(IConfigurationSource source, TracerSettings tracerSettings, IConfigurationTelemetry telemetry, OverrideErrorLog errorLog)
+        {
+            // We record the telemetry for the initial settings in a dedicated ConfigurationTelemetry,
+            // because we need to be able to reapply this configuration on dynamic config updates
+            // We don't re-record error logs, so we just use the built-in for that
+            var initialTelemetry = new ConfigurationTelemetry();
+            InitialMutableSettings = MutableSettings.CreateInitialMutableSettings(source, initialTelemetry, errorLog, tracerSettings);
+            InitialExporterSettings = new ExporterSettings(source, initialTelemetry);
+            _tracerSettings = tracerSettings;
+            _initialTelemetry = initialTelemetry;
+            initialTelemetry.CopyTo(telemetry);
+        }
 
         /// <summary>
         /// Gets the initial <see cref="MutableSettings"/>. On app startup, these will be the values read from
         /// static sources. To subscribe to updates to these settings, from code or remote config, call <see cref="SubscribeToChanges"/>.
         /// </summary>
-        public MutableSettings InitialMutableSettings { get; } = initialMutable;
+        public MutableSettings InitialMutableSettings { get; }
 
         /// <summary>
         /// Gets the initial <see cref="ExporterSettings"/>. On app startup, these will be the values read from
         /// static sources. To subscribe to updates to these settings, from code or remote config, call <see cref="SubscribeToChanges"/>.
         /// </summary>
-        public ExporterSettings InitialExporterSettings { get; } = initialExporter;
+        public ExporterSettings InitialExporterSettings { get; }
 
         /// <summary>
         /// Subscribe to changes in <see cref="MutableSettings"/> and/or <see cref="ExporterSettings"/>.
@@ -133,13 +148,37 @@ public partial record TracerSettings
             ManualInstrumentationConfigurationSourceBase manualSource,
             IConfigurationTelemetry telemetry)
         {
-            var initialSettings = manualSource.UseDefaultSources
-                                      ? InitialMutableSettings
-                                      : MutableSettings.CreateWithoutDefaultSources(_tracerSettings);
+            // Set the correct default telemetry and initial settings depending
+            // on whether the manual config source explicitly disables using the default sources
+            ConfigurationTelemetry defaultTelemetry;
+            MutableSettings initialSettings;
+            if (manualSource.UseDefaultSources)
+            {
+                defaultTelemetry = _initialTelemetry;
+                initialSettings = InitialMutableSettings;
+            }
+            else
+            {
+                // We only need to initialize the "no default sources" settings once
+                // and we don't want to initialize them if we don't _need_ to
+                // so lazy-initialize here
+                if (_noDefaultSourcesSettings is null || _noDefaultSettingsTelemetry is null)
+                {
+                    InitialiseNoDefaultSourceSettings();
+                }
+
+                defaultTelemetry = _noDefaultSettingsTelemetry;
+                initialSettings = _noDefaultSourcesSettings;
+            }
 
             var current = _latest;
             var currentMutable = current?.UpdatedMutable ?? current?.PreviousMutable ?? InitialMutableSettings;
             var currentExporter = current?.UpdatedExporter ?? current?.PreviousExporter ?? InitialExporterSettings;
+
+            // we create a temporary ConfigurationTelemetry object to hold the changes to settings
+            // if nothing is actually written, and nothing changes compared to the default, then we
+            // don't need to report it to the provided telemetry
+            var tempTelemetry = new ConfigurationTelemetry();
 
             var overrideErrorLog = new OverrideErrorLog();
             var newMutableSettings = MutableSettings.CreateUpdatedMutableSettings(
@@ -147,7 +186,7 @@ public partial record TracerSettings
                 manualSource,
                 initialSettings,
                 _tracerSettings,
-                telemetry,
+                tempTelemetry,
                 overrideErrorLog); // TODO: We'll later report these
 
             // The only exporter setting we currently _allow_ to change is the AgentUri, but if that does change,
@@ -159,7 +198,7 @@ public partial record TracerSettings
             var newRawExporterSettings = ExporterSettings.Raw.CreateUpdatedFromManualConfig(
                 currentExporter.RawSettings,
                 manualSource,
-                telemetry,
+                tempTelemetry,
                 manualSource.UseDefaultSources);
 
             var isSameMutableSettings = currentMutable.Equals(newMutableSettings);
@@ -171,11 +210,31 @@ public partial record TracerSettings
                 return null;
             }
 
+            // we have changes, so we need to report them
+            // First record the "default"/fallback values, then record the "new" values
+            defaultTelemetry.CopyTo(telemetry);
+            tempTelemetry.CopyTo(telemetry);
+
             Log.Information("Notifying consumers of new settings");
             var updatedMutableSettings = isSameMutableSettings ? null : newMutableSettings;
             var updatedExporterSettings = isSameExporterSettings ? null : new ExporterSettings(newRawExporterSettings, telemetry);
 
             return new SettingChanges(updatedMutableSettings, updatedExporterSettings, currentMutable, currentExporter);
+        }
+
+        [MemberNotNull(nameof(_noDefaultSettingsTelemetry))]
+        [MemberNotNull(nameof(_noDefaultSourcesSettings))]
+        private void InitialiseNoDefaultSourceSettings()
+        {
+            if (_noDefaultSourcesSettings is not null
+             && _noDefaultSettingsTelemetry is not null)
+            {
+                return;
+            }
+
+            var telemetry = new ConfigurationTelemetry();
+            _noDefaultSettingsTelemetry = telemetry;
+            _noDefaultSourcesSettings = MutableSettings.CreateWithoutDefaultSources(_tracerSettings, telemetry);
         }
 
         private void NotifySubscribers(SettingChanges settings)
--- a/tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
+++ b/tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
@@ -125,8 +125,6 @@ namespace Datadog.Trace.Configuration
                                        .AsBoolResult()
                                        .OverrideWith(in otelActivityListenerEnabled, ErrorLog, defaultValue: false);
 
-            var exporter = new ExporterSettings(source, _telemetry);
-
             PeerServiceTagsEnabled = config
                .WithKeys(ConfigurationKeys.PeerServiceDefaultsEnabled)
                .AsBool(defaultValue: false);
@@ -334,66 +332,6 @@ namespace Datadog.Trace.Configuration
                                     .AsBool(defaultValue: false);
 
             OpenTelemetryLogsEnabled = OpenTelemetryLogsEnabled && OtelLogsExporterEnabled;
-
-            DataPipelineEnabled = config
-                                  .WithKeys(ConfigurationKeys.TraceDataPipelineEnabled)
-                                  .AsBool(defaultValue: EnvironmentHelpers.IsUsingAzureAppServicesSiteExtension() && !EnvironmentHelpers.IsAzureFunctions());
-
-            if (DataPipelineEnabled)
-            {
-                // Due to missing quantization and obfuscation in native side, we can't enable the native trace exporter
-                // as it may lead to different stats results than the managed one.
-                if (StatsComputationEnabled)
-                {
-                    DataPipelineEnabled = false;
-                    Log.Warning(
-                        $"{ConfigurationKeys.TraceDataPipelineEnabled} is enabled, but {ConfigurationKeys.StatsComputationEnabled} is enabled. Disabling data pipeline.");
-                    _telemetry.Record(ConfigurationKeys.TraceDataPipelineEnabled, false, ConfigurationOrigins.Calculated);
-                }
-
-                // Windows supports UnixDomainSocket https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/
-                // but tokio hasn't added support for it yet https://github.com/tokio-rs/tokio/issues/2201
-                // There's an issue here, in that technically a user can initially be configured to send over TCP/named pipes,
-                // and so we allow and enable the datapipeline. Later, they could configure the app in code to send over UDS.
-                // This is a problem, as we currently don't support toggling the data pipeline at runtime, so we explicitly block
-                // this scenario in the public API.
-                if (exporter.TracesTransport == TracesTransportType.UnixDomainSocket && FrameworkDescription.Instance.IsWindows())
-                {
-                    DataPipelineEnabled = false;
-                    Log.Warning(
-                        $"{ConfigurationKeys.TraceDataPipelineEnabled} is enabled, but TracesTransport is set to UnixDomainSocket which is not supported on Windows. Disabling data pipeline.");
-                    _telemetry.Record(ConfigurationKeys.TraceDataPipelineEnabled, false, ConfigurationOrigins.Calculated);
-                }
-
-                if (!isLibDatadogAvailable.IsAvailable)
-                {
-                    DataPipelineEnabled = false;
-                    if (isLibDatadogAvailable.Exception is not null)
-                    {
-                        Log.Warning(
-                            isLibDatadogAvailable.Exception,
-                            $"{ConfigurationKeys.TraceDataPipelineEnabled} is enabled, but libdatadog is not available. Disabling data pipeline.");
-                    }
-                    else
-                    {
-                        Log.Warning(
-                            $"{ConfigurationKeys.TraceDataPipelineEnabled} is enabled, but libdatadog is not available. Disabling data pipeline.");
-                    }
-
-                    _telemetry.Record(ConfigurationKeys.TraceDataPipelineEnabled, false, ConfigurationOrigins.Calculated);
-                }
-
-                // SSI already utilizes libdatadog. To prevent unexpected behavior,
-                // we proactively disable the data pipeline when SSI is enabled. Theoretically, this should not cause any issues,
-                // but as a precaution, we are taking a conservative approach during the initial rollout phase.
-                if (!string.IsNullOrEmpty(EnvironmentHelpers.GetEnvironmentVariable("DD_INJECTION_ENABLED")))
-                {
-                    DataPipelineEnabled = false;
-                    Log.Warning(
-                        $"{ConfigurationKeys.TraceDataPipelineEnabled} is enabled, but SSI is enabled. Disabling data pipeline.");
-                    _telemetry.Record(ConfigurationKeys.TraceDataPipelineEnabled, false, ConfigurationOrigins.Calculated);
-                }
-            }
 
             // We should also be writing telemetry for OTEL_LOGS_EXPORTER similar to OTEL_METRICS_EXPORTER, but we don't have a corresponding Datadog config
             // When we do, we can insert that here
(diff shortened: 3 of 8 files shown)


tracer/src/Datadog.Trace/DogStatsd/StatsdManager.cs
```diff
@@ -0,0 +1,143 @@
...
+    internal static bool HasImpactingChanges(TracerSettings.SettingsManager.SettingChanges changes)
+    {
+        var hasChanges = changes.UpdatedExporter is not null // relying on this to only be non null if _anything_ changed
+                      || (changes.UpdatedMutable is { } updated
+                       && !(
+                               string.Equals(updated.Environment, changes.PreviousMutable.Environment, StringComparison.Ordinal)
+                            && string.Equals(updated.ServiceVersion, changes.PreviousMutable.ServiceVersion, StringComparison.Ordinal)
+                               // The service name comparison isn't _strictly_ correct, because we normalize it further, but this is probably good enough
+                            && string.Equals(updated.DefaultServiceName, changes.PreviousMutable.DefaultServiceName, StringComparison.OrdinalIgnoreCase)
+                            && updated.GlobalTags.SequenceEqual(changes.PreviousMutable.GlobalTags)));
```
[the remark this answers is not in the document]
andrewlock: That's not a Linq extension, it's the method added in https://github.com/DataDog/dd-trace-dotnet/pull/7722

tracer/src/Datadog.Trace/Agent/Api.cs
```diff
@@ -76,6 +80,12 @@ private enum SendResult
             Failed_DontRetry,
         }
 
+        [MemberNotNull(nameof(_statsd))]
+        public void ToggleTracerHealthMetrics(bool enabled)
+        {
+            _statsd = enabled ? _originalStatsd : NoOpStatsd.Instance;
```
andrewlock: Thread safety issue: The assignment to `_statsd` should use `Volatile.Write(ref _statsd, enabled ? _originalStatsd : NoOpStatsd.Instance)` or `Interlocked.Exchange` to ensure proper memory visibility across threads. Without a memory barrier, other threads may not see the updated value.

Since this method is called from a settings change callback (which runs on a background thread), and `_statsd` is accessed from other threads, proper synchronization is critical.

*Automated comment by Claude Code*

tracer/src/Datadog.Trace/Agent/ManagedApi.cs
```diff
@@ -0,0 +1,59 @@
...
+    private Api _api;
+
+    public ManagedApi(
+        TracerSettings.SettingsManager settings,
+        IDogStatsd statsd,
+        Action<Dictionary<string, float>> updateSampleRates,
+        bool partialFlushEnabled)
+    {
+        UpdateApi(settings.InitialExporterSettings, settings.InitialMutableSettings.TracerMetricsEnabled);
+        settings.SubscribeToChanges(changes =>
```
andrewlock, line 34: Memory leak: The `ManagedApi` class subscribes to settings changes but never disposes the subscription. Consider implementing `IDisposable` and disposing the subscription, or document why this is intentional (e.g., if the lifetime matches the application lifetime).

Pattern to follow:
```csharp
private readonly IDisposable _settingsSubscription;

public void Dispose()
{
    _settingsSubscription?.Dispose();
}
```

*Automated comment by Claude Code*

tracer/src/Datadog.Trace/LibDatadog/DataPipeline/ManagedTraceExporter.cs
```diff
@@ -0,0 +1,174 @@
...
+            {
+                var exporter = CreateTraceExporter(
+                    settings,
+                    changes.UpdatedMutable ?? changes.PreviousMutable,
+                    changes.UpdatedExporter ?? changes.PreviousExporter,
+                    updateSampleRates,
+                    telemetrySettings);
+
+                var previous = Interlocked.Exchange(ref _current, exporter);
+                previous?.Dispose();
```
andrewlock, line 60: Potential use-after-dispose: The previous exporter is disposed immediately after exchange on line 55. However, there could be in-flight operations still using the old exporter. Consider either:

1. Deferring disposal with a delay (similar to `StatsdManager.cs:67-72`)
2. Implementing a graceful shutdown mechanism
3. Documenting why immediate disposal is safe here (e.g., if the API methods handle null gracefully)

The pattern in StatsdManager provides a good reference for safer disposal.

*Automated comment by Claude Code*

tracer/src/Datadog.Trace/DogStatsd/StatsdManager.cs
```diff
@@ -0,0 +1,143 @@
...
+            if (!HasImpactingChanges(c))
+            {
+                return;
+            }
+
+            IDogStatsd previous;
+
+            lock (_lock)
+            {
+                if (Volatile.Read(ref _isDisposed))
```
andrewlock: Potential race condition: The check `Volatile.Read(ref _isDisposed)` on line 50 is inside the lock, but then `_current` is used outside the lock on line 56. If disposal occurs between releasing the lock and calling `StatsdFactory.CreateDogStatsdClient`, the new instance could be created after disposal.

Consider restructuring to either:
1. Keep the entire operation inside the lock
2. Return early before the lock if disposed
3. Re-check disposal after updating `_current` but before exiting the lock

*Automated comment by Claude Code*

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
```diff
@@ -27,22 +28,38 @@ internal class DataStreamsManager
     private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor<DataStreamsManager>();
     private static readonly AsyncLocal<PathwayContext?> LastConsumePathway = new(); // saves the context on consume checkpointing only
     private readonly ConcurrentDictionary<string, RateLimiter> _schemaRateLimiters = new();
-    private readonly NodeHashBase _nodeHashBase;
+    private readonly IDisposable _updateSubscription;
```
andrewlock, line 30: Memory leak: The `DataStreamsManager` subscribes to settings changes via `_updateSubscription` but never disposes of it. The `DataStreamsManager` class should implement `IDisposable` and dispose the subscription in the `Dispose` method.

```csharp
public void Dispose()
{
    _updateSubscription?.Dispose();
    _writer?.Dispose(); // if applicable
}
```

*Automated comment by Claude Code*

tracer/src/Datadog.Trace/Agent/DiscoveryService/DiscoveryService.cs
```diff
@@ -32,16 +32,35 @@ internal class DiscoveryService : IDiscoveryService
...
+        public DiscoveryService(
+            TracerSettings.SettingsManager settings,
+            TimeSpan tcpTimeout,
+            int initialRetryDelayMs,
+            int maxRetryDelayMs,
+            int recheckIntervalMs)
+            : this(CreateApiRequestFactory(settings.InitialExporterSettings, tcpTimeout), initialRetryDelayMs, maxRetryDelayMs, recheckIntervalMs)
+        {
+            // Create as a "managed" service that can update the request factory
+            settings.SubscribeToChanges(changes =>
```
andrewlock: Memory leak: The `DiscoveryService` constructor subscribes to settings changes but the subscription is not stored or disposed. Consider storing the subscription as a field and disposing it in `DisposeAsync`.

```csharp
private readonly IDisposable? _settingsSubscription;

public DiscoveryService(TracerSettings.SettingsManager settings, ...)
{
    // ...
    _settingsSubscription = settings.SubscribeToChanges(changes => { ... });
}

public Task DisposeAsync()
{
    _settingsSubscription?.Dispose();
    // ... existing disposal logic
}
```

*Automated comment by Claude Code*

tracer/src/Datadog.Trace/Sampling/ManagedTraceSampler.cs
```diff
@@ -0,0 +1,160 @@
...
+                var rates = Volatile.Read(ref _defaultSampleRates);
+                if (rates is not null)
+                {
+                    newSampler.SetDefaultSampleRates(rates);
+                }
+
+                Interlocked.Exchange(ref _current, newSampler);
+
+                // There's a small edge case here where the agent rates are being updated at the same time as we're
+                // updating the sampler, so check and just update them again if necessary
```
andrewlock: Race condition with double-check pattern: Lines 48-52 implement a double-check to handle concurrent updates to `_defaultSampleRates`. However, this pattern has a subtle issue: between line 45 (Exchange) and line 49 (Volatile.Read), another thread could update `_defaultSampleRates` twice, and only the second update would be applied.

While this is unlikely to cause serious issues (since it's just applying the same rates again), consider whether the complexity of the double-check is justified, or if a simpler approach (accepting that very rare concurrent updates might miss one setting) would be clearer.

*Automated comment by Claude Code*

tracer/src/Datadog.Trace/Agent/ManagedApi.cs
andrewlock: Yeah, there's a bunch of these - in _general_ the lifetime of all subscribers _will_ match the application lifetime. Will document it for now, but maybe we should consider not even bothering to have "unsubscribe" as an option? 🤔

tracer/src/Datadog.Trace/LibDatadog/DataPipeline/ManagedTraceExporter.cs
andrewlock: Yeah, I agonized a bit about this one. Will document my conclusion

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
andrewlock: There's a `DisposeAsync()` method that's called. That's good enough

tracer/src/Datadog.Trace/DogStatsd/StatsdManager.cs
andrewlock: > but then _current is used outside the lock on line 56

No it's not? 🤔 In fact, realised we don't need the volatile if we're inside the lock, so removed those calls

tracer/src/Datadog.Trace/Agent/DiscoveryService/DiscoveryService.cs
andrewlock: It _does_ make sense to fix this one, because `DiscoveryService` doesn't _necessarily_ live as long as the application

tracer/src/Datadog.Trace/Sampling/ManagedTraceSampler.cs
andrewlock: Yeah, it's not worth the complexity for the rare updates here I think. Will switch to `lock` instead.

tracer/src/Datadog.Trace/Agent/Api.cs
andrewlock: Fixed

andrewlock:
# Code Review: PR #7724 - Handle changes to MutableSettings and ExporterSettings without rebuilding

## Overview

This PR implements dynamic reconfiguration of tracer services in response to runtime settings changes (from manual configuration or remote config), eliminating the need to tear down and rebuild the entire tracer infrastructure. The changes introduce "managed" versions of services that can respond to specific setting changes without full disposal/reconstruction.

**Key Changes:**
- Introduce `IStatsdManager` abstraction with lease-based access pattern
- Add `ManagedApi` and update `DiscoveryService` to handle exporter setting changes
- Update `AgentWriter`, `Api`, `StatsAggregator`, and related services to subscribe to relevant setting changes
- Replace `IDogStatsd` with `IStatsdManager` throughout
- Add thread-safe patterns using `Volatile.Read()` for accessing mutable state

---

## Code Quality & Architecture

### Strengths ✓

1. **Clear separation of concerns**: Managed vs unmanaged services (e.g., `DiscoveryService.CreateManaged()` vs `CreateUnmanaged()`)
2. **Selective subscription pattern**: Services only subscribe to changes they care about
3. **Backward compatibility**: CI Vis and Debugger paths maintain existing behavior
4. **Thread safety**: Consistent use of `Volatile.Read()` for cross-thread visibility
5. **Lease pattern**: The `IStatsdManager.TryGetClientLease()` pattern is clean and prevents access to disposed/stale clients

### Areas for Improvement ⚠️

1. **Thread safety complexity**: Multiple patterns for handling concurrency could be consolidated
2. **Memory management**: No explicit disposal of old `Api` instances in `ManagedApi`
3. **Subscription lifecycle**: Some subscriptions (e.g., in `ManagedApi`) are never disposed
4. **Pattern inconsistency**: Different services use different update mechanisms

---

## Specific Issues & Suggestions

### 1. Memory Leak Risk in ManagedApi.cs:52

```csharp
var api = new Api(apiRequestFactory, statsd, updateSampleRates, partialFlushEnabled, healthMetricsEnabled);
Interlocked.Exchange(ref _api!, api);
```

**Issue**: Old `Api` instances are replaced but never explicitly disposed. The `Api` class doesn't implement `IDisposable`, but it holds resources like `IApiRequestFactory`.

**Suggestion**: Consider implementing disposal or documenting why it's safe to replace without cleanup.

### 2. Race Condition Pattern in AgentWriter.cs:138-161

```csharp
if (Volatile.Read(ref _traceMetricsEnabled))
{
    using var lease = _statsd.TryGetClientLease();
    if (lease.Client is { } statsd)
    {
        statsd.Increment(TracerMetricNames.Queue.EnqueuedTraces);
        statsd.Increment(TracerMetricNames.Queue.EnqueuedSpans, trace.Count);
    }
}
```

**Issue**: `_traceMetricsEnabled` could change between the check and the lease acquisition. While benign here, this pattern appears in multiple locations.

**Suggestion**: Consider caching the `Volatile.Read()` result to a local variable for consistency:
```csharp
var metricsEnabled = Volatile.Read(ref _traceMetricsEnabled);
if (metricsEnabled)
{
    using var lease = _statsd.TryGetClientLease();
    // ...
}
```

### 3. Inconsistent Pattern in Api.cs:314-325

In `Api.SendTracesAsyncImpl`, the lease is acquired conditionally:
```csharp
using var lease = Volatile.Read(ref _healthMetricsEnabled) ? _statsd.TryGetClientLease() : default;
var healthStats = _healthMetricsEnabled ? lease.Client : null;
```

**Issue**: Uses `_healthMetricsEnabled` without `Volatile.Read()` in the second line, but with it in the first. Also reads the field twice.

**Suggestion**:
```csharp
var healthMetricsEnabled = Volatile.Read(ref _healthMetricsEnabled);
using var lease = healthMetricsEnabled ? _statsd.TryGetClientLease() : default;
var healthStats = lease.Client;
```

### 4. ClientStatsPayload Record Pattern (ClientStatsPayload.cs:14-25)

```csharp
private AppSettings _settings = CreateSettings(settings);
// ...
public void UpdateDetails(MutableSettings settings)
    => Interlocked.Exchange(ref _settings, CreateSettings(settings));
```

**Issue**: While functionally correct, using `Interlocked.Exchange` on a record field is unconventional. Records are typically immutable by design.

**Suggestion**: Document why this pattern was chosen, or consider using a class instead of a record.

### 5. Missing Null Check in AgentWriter.cs:73-79

```csharp
settings.Manager.SubscribeToChanges(changes =>
{
    if (changes.UpdatedMutable is { } mutable
     && mutable.TracerMetricsEnabled != changes.PreviousMutable.TracerMetricsEnabled)
    {
        // ...
    }
});
```

**Issue**: If `changes.PreviousMutable` is null, accessing `TracerMetricsEnabled` could throw (though unlikely in practice).

**Suggestion**: Add null-safety check or document that `PreviousMutable` is never null.

---

## Testing Considerations

### Missing Test Coverage

1. **Concurrency tests**: No evidence of tests for race conditions during setting updates
2. **Subscription cleanup**: Tests for memory leaks from subscriptions
3. **Service reconfiguration**: Integration tests validating services continue working after settings change
4. **Lease pattern edge cases**: What happens if settings change mid-lease?

### Recommended Test Scenarios

1. Update `ExporterSettings` while traces are being sent
2. Toggle `TracerMetricsEnabled` rapidly to test race conditions
3. Update multiple settings simultaneously
4. Verify old service instances are eventually GC'd

---

## Performance Implications

### Positive Impacts ✓
- Eliminates overhead of tearing down/rebuilding entire infrastructure
- Reduces GC pressure by reusing services

### Potential Concerns ⚠️
- `Volatile.Read()` has memory barrier overhead (minimal on modern CPUs)
- Multiple subscription callbacks on hot paths (e.g., every trace write)
- Lease pattern adds indirection (negligible but measurable)

**Recommendation**: Add performance benchmarks comparing before/after settings changes.

---

## Security Considerations

No major security concerns identified. The changes are primarily architectural and don't introduce new external inputs or trust boundaries.

---

## Overall Assessment

**Rating: Approve with Minor Changes**

This is a solid architectural improvement that significantly reduces the overhead of runtime configuration changes. The implementation is generally well-thought-out with good separation between managed and unmanaged paths.

### Must-Fix Before Merge
1. Document memory management strategy for replaced `Api` instances
2. Fix inconsistent use of `Volatile.Read()` in `Api.cs:314-325`
3. Add null-safety checks in subscription lambdas

### Should-Fix Soon After
1. Add concurrency tests
2. Consolidate thread-safety patterns
3. Add XML documentation to new public APIs
4. Performance benchmarks

### Nice-to-Have
1. Consider implementing `IDisposable` on `Api` if resources need explicit cleanup
2. Add `[DebuggerDisplay]` attributes to aid debugging
3. Consistent nullable reference context across all files

The PR successfully achieves its stated goal of eliminating unnecessary service rebuilds while maintaining backward compatibility. The commit-by-commit structure mentioned in the description should make review easier.

---

*Review generated by Claude Code*

tracer/src/Datadog.Trace/Agent/Api.cs
```diff
@@ -298,10 +311,12 @@ private async Task<SendResult> SendTracesAsyncImpl(IApiRequest request, bool fin
 
             try
             {
+                using var lease = Volatile.Read(ref _healthMetricsEnabled) ? _statsd.TryGetClientLease() : default;
```
andrewlock: **Thread Safety Concern: Missing volatile read before lease acquisition**

In this method, you're checking `_healthMetricsEnabled` without a volatile read before getting the client lease, but then using `Volatile.Read()` inside the ternary. This creates a potential race condition where the health metrics setting could be disabled between the two checks, leading to unnecessary lease acquisition.

Suggested fix:
```csharp
var healthMetricsEnabled = Volatile.Read(ref _healthMetricsEnabled);
using var lease = healthMetricsEnabled ? _statsd.TryGetClientLease() : default;
var healthStats = healthMetricsEnabled ? lease.Client : null;
```

This ensures consistent state throughout the method and avoids the double-check pattern that could cause issues.

🤖 Claude Code

tracer/src/Datadog.Trace/Agent/DiscoveryService/DiscoveryService.cs
```diff
@@ -32,15 +32,35 @@ internal class DiscoveryService : IDiscoveryService
...
+            int recheckIntervalMs)
+            : this(CreateApiRequestFactory(settings.InitialExporterSettings, tcpTimeout), initialRetryDelayMs, maxRetryDelayMs, recheckIntervalMs)
+        {
+            // Create as a "managed" service that can update the request factory
+            _settingSubscription = settings.SubscribeToChanges(changes =>
+            {
+                if (changes.UpdatedExporter is { } exporter)
+                {
+                    var newFactory = CreateApiRequestFactory(exporter, tcpTimeout);
+                    Interlocked.Exchange(ref _apiRequestFactory!, newFactory);
```
andrewlock, line 63: **Potential NullReferenceException: Missing null check after Exchange**

After `Interlocked.Exchange(ref _apiRequestFactory!, newFactory)`, the old factory is not disposed. While this may be intentional for thread safety reasons, there's no guarantee that `_apiRequestFactory` is initialized before the subscription callback fires.

The constructor calls `CreateApiRequestFactory()` after setting up the subscription, which means the callback could theoretically fire before `_apiRequestFactory` is initialized (though unlikely in practice).

Consider either:
1. Initialize `_apiRequestFactory` before creating the subscription
2. Add a null-check in the subscription callback
3. Dispose the old factory if it's safe to do so

🤖 Claude Code

tracer/src/Datadog.Trace/Agent/AgentWriter.cs
```diff
@@ -138,10 +151,14 @@ public void WriteTrace(ArraySegment<Span> trace)
                 }
             }
 
-            if (_statsd != null)
+            if (Volatile.Read(ref _traceMetricsEnabled))
```
andrewlock, line 154: **Thread Safety Issue: TOCTOU (Time-of-Check-Time-of-Use) vulnerability**

This pattern has a classic TOCTOU race condition:

1. Line 154: Check `Volatile.Read(ref _traceMetricsEnabled)`
2. Line 156: Get client lease
3. Line 157: Null-check the client

Between steps 1 and 2, another thread could disable metrics and dispose the statsd client, potentially causing issues when `TryGetClientLease()` is called on a disposed/null client.

Suggested fix - cache the volatile read:
```csharp
var metricsEnabled = Volatile.Read(ref _traceMetricsEnabled);
if (metricsEnabled)
{
    using var lease = _statsd.TryGetClientLease();
    if (lease.Client is { } statsd)
    {
        statsd.Increment(TracerMetricNames.Queue.EnqueuedTraces);
        statsd.Increment(TracerMetricNames.Queue.EnqueuedSpans, trace.Count);
    }
}
```

This pattern appears multiple times in AgentWriter.cs (lines 154, 334, 536) and should be fixed consistently.

🤖 Claude Code

tracer/src/Datadog.Trace/Agent/ManagedApi.cs
```diff
@@ -0,0 +1,61 @@
...
+                _api.ToggleTracerHealthMetrics(mutable.TracerMetricsEnabled);
+            }
+        });
+
+        [MemberNotNull(nameof(_api))]
+        void UpdateApi(ExporterSettings exporterSettings, bool healthMetricsEnabled)
+        {
+            var apiRequestFactory = TracesTransportStrategy.Get(exporterSettings);
+            var api = new Api(apiRequestFactory, statsd, updateSampleRates, partialFlushEnabled, healthMetricsEnabled);
+            Interlocked.Exchange(ref _api!, api);
```
andrewlock, line 52: **Potential Resource Leak: No disposal of old `_api` instances**

In `UpdateApi()`, you're using `Interlocked.Exchange(ref _api!, api)` to swap the API instance, but the old instance is never disposed. Over time, if settings change frequently, this could lead to resource leaks (open connections, tasks, etc.).

The `Api` class doesn't implement `IDisposable`, so this might be by design, but it's worth confirming that the old instances are properly garbage collected and don't hold onto resources.

Consider either:
1. Making `Api` disposable and disposing the old instance
2. Adding a comment explaining why disposal isn't necessary
3. Implementing a proper cleanup mechanism

🤖 Claude Code

tracer/src/Datadog.Trace/Agent/AgentWriter.cs
```diff
@@ -65,17 +65,28 @@ internal class AgentWriter : IAgentWriter
 
         private long _droppedTraces;
 
-        public AgentWriter(IApi api, IStatsAggregator statsAggregator, IDogStatsd statsd, TracerSettings settings)
-            : this(api, statsAggregator, statsd, maxBufferSize: settings.TraceBufferSize, batchInterval: settings.TraceBatchInterval, apmTracingEnabled: settings.ApmTracingEnabled)
+        private bool _traceMetricsEnabled;
```
andrewlock, line 68: **Missing Initialization: `_traceMetricsEnabled` used before assignment**

The field `_traceMetricsEnabled` is declared but not initialized. The first constructor (starting line 70) calls the internal constructor and sets up a subscription that immediately references `_traceMetricsEnabled` via `Volatile.Write()`.

While the subscription callback won't fire until after construction completes, it's better to explicitly initialize the field to avoid potential issues and make the intent clear.

Add initialization:
```csharp
private bool _traceMetricsEnabled = false;
```

🤖 Claude Code

andrewlock:
## Review Summary

This is a significant refactoring that transitions from rebuilding all services when settings change to dynamically updating only the necessary components. The overall architecture is sound and the commit-by-commit approach makes the changes manageable.

### Key Findings

**Critical Issues:**
1. **Thread Safety Concerns**: Multiple TOCTOU (Time-of-Check-Time-of-Use) race conditions in  where  checks are not cached before use, potentially causing issues if settings change mid-execution.

2. **Resource Management**: The new managed services (, ) swap underlying instances without disposing old ones. While this may be intentional for thread safety, it risks resource leaks if settings change frequently.

**Moderate Issues:**
3. **Field Initialization**: The  field in  lacks explicit initialization, relying on default bool values.

4. **Inconsistent Volatile Access**: In , the  field changed from readonly to mutable, but initial access in  doesn't use volatile reads consistently.

The patterns introduced here (subscription-based updates, volatile reads, managed wrappers) are appropriate for the goal, but need tightening around thread safety guarantees. I recommend reviewing all  usage to ensure values are cached locally before use, and documenting the lifecycle expectations for swapped service instances.

Review by 🤖 Claude Code

tracer/src/Datadog.Trace/Agent/DiscoveryService/DiscoveryService.cs
andrewlock: I think this is nonsense:
1. the factory doesn't need to be disposed (it doesn't implement IDisposable)
2. This is only called _after_ the other constructor, so _apiRequestFactory is not null when this is called

tracer/src/Datadog.Trace/Agent/AgentWriter.cs
andrewlock: I think this is fine 🤔 The `metricsEnabled` check is just to avoid calling `TryGetClientLease()` when we don't need it. The StatsDManager won't give out an already disposed lease, so there's nothing to worry about threading-wise I believe. 

Also the suggested code is functionally identical

tracer/src/Datadog.Trace/Agent/ManagedApi.cs
andrewlock: There's nothing disposable, none of this stuff is disposable

tracer/src/Datadog.Trace/Agent/AgentWriter.cs
andrewlock: That's automatic in C#

tracer/src/Datadog.Trace/Agent/ManagedApi.cs
```diff
@@ -0,0 +1,61 @@
...
+                UpdateApi(exporter, mutable.TracerMetricsEnabled);
+            }
+            else if (changes.UpdatedMutable is { } mutable && mutable.TracerMetricsEnabled != changes.PreviousMutable.TracerMetricsEnabled)
+            {
+                _api.ToggleTracerHealthMetrics(mutable.TracerMetricsEnabled);
+            }
+        });
+
+        [MemberNotNull(nameof(_api))]
+        void UpdateApi(ExporterSettings exporterSettings, bool healthMetricsEnabled)
```
lucaspimentel, line 48:
```suggestion
        static void UpdateApi(ExporterSettings exporterSettings, bool healthMetricsEnabled)
```

andrewlock: This can't be static, it mutates `_api` (as indicated by [MemberNotNull] 🙂)

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
```diff
@@ -200,7 +217,9 @@ public void InjectPathwayContextAsBase64String<TCarrier>(PathwayContext? context
             var edgeStartNs = previousContext == null && timeInQueueMs > 0 ? nowNs - (timeInQueueMs * 1_000_000) : nowNs;
             var pathwayStartNs = previousContext?.PathwayStart ?? edgeStartNs;
 
-            var nodeHash = HashHelper.CalculateNodeHash(_nodeHashBase, edgeTags);
+            // Don't blame me, blame the fact we can't do Volatile.Read with a ulong in .NET FX...
+            var nodeHashBase = new NodeHashBase(unchecked((ulong)Volatile.Read(ref Unsafe.As<ulong, long>(ref _nodeHashBase))));
```
bouwkast: 😅 

So uh what exactly is going on here?
It is a `struct` so can't use InterLocked.CompareExchange but we also need to convert it from ulong to long?

tracer/src/Datadog.Trace/Agent/ManagedApi.cs
lucaspimentel: Weird, before commenting, I actually checked out the code and added `static`. I guess I was just looking for the IDE's red squiggles and didn't actually _build_ the code 😅

lucaspimentel: Nope, yep. I see them squiggles now.

tracer/test/snapshots/OpenTelemetrySdkTests.SubmitsOtlpMetrics_up_to_1_5_0.NET_6_DD.verified.txt
```diff
@@ -25,7 +25,7 @@
             {
               "key": "service.name",
               "value": {
-                "string_value": "unknown_service:dotnet"
+                "string_value": "Samples.OpenTelemetrySdk"
```
bouwkast, line 28: Hmm I do remember running into this previously with `unknown_service` I think I did something to override this like so a long time ago, unsure how / why this changed though now 🤷

andrewlock: Yeah, I'm going to defer to @link04 on this one 😅 My guess is that there were/are some race conditions here. I think this is now "correct" to be fair 😄

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
andrewlock: _so_, yeah, this is a mess 😅 

- We want to cache the value of `NodeHashBase`
- `NodeHashBase` depends on some mutable settings, e.g. service name, so we need to be able to _replace_ `NodeHashBase`.
- We need to use `Interlocked.Exchange()` to make sure we do the swap atomically.
  - You _can't_ do `Interlocked.Exchange()` with a struct (for technical reasons) 🙁 
- Luckily, `NodeHashBase` is just a thin wrapper around a `ulong`, which is 64 bit and _can_ be used with  `Interlocked.Exchange` 🎉 
  - _Except_, < .NET 5, you _can't_ do `Interlocked.Exchange(ulong)` because the overloads don't exist 😭 
- The "solution" for _updating_ is (you can see this in the `UpdateNodeHash` method):
    - Calculate the new `NodeHashBase`
    - Grab the `ulong` from the `NodeHashBase`
    - Do an unchecked cast of the `ulong` as a `long`
    - Do an `Interlocked.Exchange(ref long)`, treating the stored `ulong _nodeHashBase` as a `long` for the purposes of the exchange
- The solution for _reading_ is:
  - Do a `Volatile.Read(ref long)`, treating the stored `ulong _nodeHashBase` as a `long` for the purposes of the read
    - Do an unchecked case of the `long` as a `ulong`,
    - Create the new `NodeHashBase`

Yes, it's a PITA. 

All that said, I just realised that if we store the `_nodeHashBase` as an `int`, we can avoid all the `Unsafe.As` reinterpretation, and just stick with the unchecked casts, which simplifies this a bit! 🙂

tonyredondo (approved):

PR #7733, opened 2025-10-29
Split integration tests in groups

pablomartinezbernardo:
## Summary of changes
* Extract integration test running logic to template
* Split `DockerTest` job in two arbitrary groups
* Categorize tests with a new `DockerGroup` trait indicating which group they belong to
* Categorize docker containers using [service profiles](https://docs.docker.com/compose/how-tos/profiles/)

## Reason for change

The number of containers in a single job was starting to be problematic for startup. This approach splits images in two groups (attempting to keep them similarly weighted) easily extensible to more groups.

### DIFF
--- a/.azure-pipelines/ultimate-pipeline.yml
+++ b/.azure-pipelines/ultimate-pipeline.yml
@@ -2216,7 +2216,7 @@ stages:
   jobs:
   - template: steps/update-github-status-jobs.yml
     parameters:
-      jobs: [Test, DockerTestGroup1, DockerTestGroup2, Serverless]
+      jobs: [Test, DockerTest, Serverless]
 
   - job: Test
     timeoutInMinutes: 60 #default value


.azure-pipelines/steps/run-docker-integration-tests.yml
andrewlock, line 1: Instead of extracting this into a separate file, we can instead update the `integration_tests_linux_docker_matrix` test matrix to have a new groupNumber variable. That will mean we can easily keep this all in the main file.

I think keeping it all together here is important, because there's a lot of "find and replace" that often happens to the pipeline when we refactor things, and this will likely get forgotten. By adding the variable to the generation matrix we're not going to be duplicating anything, so I think it's fine 🙂

bouwkast (approved):
I didn't review _super_ closely but it looks right to me

Thank you very much!

.azure-pipelines/steps/run-docker-integration-tests.yml
pablomartinezbernardo: Also results in a cleaner diff, so I'm all for it

PR #7744, opened 2025-10-30
Stop using `Tracer.Instance` if we don't need to

andrewlock:
## Summary of changes

Updates a couple of places where we're calling `Tracer.Instance` where we don't need to

## Reason for change

In one of my other PRs I accidentally broke something that should _only_ have affected integration tests, but a bunch of unit tests broke. It highlighted where they were using `Tracer.Instance` and setting the global tracer for tests. In other places, it revealed that code that tests that _looked_ independent of other tests really wasn't...

Calling `Tracer.Instance` potentially does a _lot_ of work, as it initializes the tracer. It's also hard to follow if we're making static calls out in places we don't need to. So just pass through the settings we want instead.

## Implementation details

Instead of calling `Tracer.Instance.Settings`, use the value of `Tracer` or `TracerSettings` that's already available wherever possible. It makes the tests cleaner too.

## Test coverage

Same coverage, just a bit cleaner 

## Other details

Included as part of the config stack, just because I already refactored some of this code, and can't be bothered to faff with merge conflicts:

- https://github.com/DataDog/dd-trace-dotnet/pull/7744 👈

### DIFF
--- a/tracer/test/Datadog.Trace.Tests/DataStreamsMonitoring/DataStreamsContextPropagatorTests.cs
+++ b/tracer/test/Datadog.Trace.Tests/DataStreamsMonitoring/DataStreamsContextPropagatorTests.cs
@@ -15,8 +15,9 @@ namespace Datadog.Trace.Tests.DataStreamsMonitoring
 {
     public class DataStreamsContextPropagatorTests
     {
-        [Fact]
-        public void CanRoundTripPathwayContext()
+        [Theory]
+        [CombinatorialData]
+        public void CanRoundTripPathwayContext(bool isDataStreamsLegacyHeadersEnabled)
         {
             var oneMs = TimeSpan.FromMilliseconds(1);
             var headers = new TestHeadersCollection();
@@ -25,7 +26,7 @@ namespace Datadog.Trace.Tests.DataStreamsMonitoring
                 DateTimeOffset.UtcNow.AddSeconds(-5).ToUnixTimeNanoseconds(),
                 DateTimeOffset.UtcNow.ToUnixTimeNanoseconds());
 
-            DataStreamsContextPropagator.Instance.Inject(context, headers);
+            DataStreamsContextPropagator.Instance.Inject(context, headers, isDataStreamsLegacyHeadersEnabled);
 
             var extracted = DataStreamsContextPropagator.Instance.Extract(headers);
 
@@ -86,8 +87,9 @@ namespace Datadog.Trace.Tests.DataStreamsMonitoring
             extractedContext.Value.EdgeStart.Should().NotBe(binaryContext.EdgeStart);
         }
 
-        [Fact]
-        public void InjectedHeaders_HaveCorrectFormat()
+        [Theory]
+        [CombinatorialData]
+        public void InjectedHeaders_HaveCorrectFormat(bool isDataStreamsLegacyHeadersEnabled)
         {
             var headers = new TestHeadersCollection();
             var context = new PathwayContext(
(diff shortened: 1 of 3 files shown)


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AdoNet/DbScopeFactory.cs
```diff
@@ -245,7 +246,7 @@ static Cache()
             {
                 CommandType = typeof(TCommand);
 
-                if (TryGetIntegrationDetails(CommandType.FullName, out var integrationId, out var dbTypeName))
+                if (TryGetIntegrationDetails(Tracer.Instance.Settings.DisabledAdoNetCommandTypes, CommandType.FullName, out var integrationId, out var dbTypeName))
```
andrewlock, line 249: This one is actually kind of annoying. As it's in a static cctor, we can't pass it in anywhere, which means there's _one_ `DbScopeFactoryTests` test which will use this and end up initializing the global tracer. Oh well.

bouwkast: :sorry: pretty sure this was me

tracer/test/Datadog.Trace.Tests/DataStreamsMonitoring/DataStreamsContextPropagatorTests.cs
```diff
@@ -15,8 +15,9 @@ namespace Datadog.Trace.Tests.DataStreamsMonitoring
 {
     public class DataStreamsContextPropagatorTests
     {
-        [Fact]
-        public void CanRoundTripPathwayContext()
+        [Theory]
+        [CombinatorialData]
```
bouwkast, line 19: 😃

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AdoNet/DbScopeFactory.cs
zacharycmontoya: Hmm. I understand that we want to minimize work but I feel like we should calculate all DbCommands regardless of config (i.e. no dependency here) and then we do the config check on the hot path, which allows this setting to be hot-reloadable or "config at runtime" ready

zacharycmontoya: Concretely, I'm suggesting that we remove the `TryGetIntegrationDetails` call in the static constructor of `Cache<TCommand>` and refactor the other call in `CreateDbCommandScope` a bit. Perhaps in a separate PR

andrewlock: I can see arguments both ways tbh. 

The only downside to @zacharycmontoya's  proposal would be more work to do for _custom_ DbCommands (ones that we don't explicitly know about), which are _also_ marked as disabled initially, and which customers _have_ marked as disabled. The question I would have is whether that's ok - e.g. if they're disabled because we're erroring out literally running `TryGetIntegrationDetails`, and _that's_ why they disabled it, then that could be a problem. If it's just an optimization, then yes, it probably is fine.

That said, this actually is technically _already_ hot-reloadable. The `DisabledAdoNetCommandTypes` setting passed in here is used to decide whether to populate the cache for custom types, but if we _don't_ populate the cache, then we run `TryGetIntegrationDetails()` and check `DisabledAdoNetCommandTypes` at runtime _anyway_. So if this changes, then the result changes, it will just never be cached in that scenario.

So on that basis, I think removing the `TryGetIntegrationDetails` has a small amount of risk (if customer disabled adonet due to errors - not sure if that's really possible, but I can believe it), means a small bit more work (calculating values which will never be used), and doesn't change how hot-reloadable we are _in general_ (though if we made it hot reloadable, we'd have to rethink how caching works). The plus side is it removes the static access to Tracer.Instance which is better for tests.

So yeah, I'm torn 🤷‍♂️ I'd say separate PR either way, given the questions, as the current behaviour in this PR is identical to the existing. But it's worth thinking about

PR #7746, opened 2025-10-31
Removing V1 tags class from AWS integrations

rithikanarayan:
## Summary of changes

- Removed classes Aws{Service}V1Tags (such as AWSS3V1Tags) from Datadog.Trace.Tagging. 
- Removed accompanying usages of the classes in MessagingSchema and unit tests. 
- Added PeerService and PeerServiceSource tags to AwsSDKTags.
- Deletion of Aws{Service}V1 snapshots and removal of V1 schema from Aws managed service integration tests

## Reason for change
Cleanup of unused V1 code. Preparation for follow up PR addressing Serverless Service Representation for AWS Managed Services, which requires the V0 schema for these services to have access to the peer service and peer service source tags to hardcode their values in serverless environments. 

## Implementation details

- Added PeerService and PeerServiceSource tags in Datadog.Trace/src/Datadog.Trace/Tagging/AwsSdkTags.cs
- Removed classes AwsEventBridgeV1Tags, AwsKinesisV1Tags, AwsS3V1Tags, AwsSnsV1Tags, AwsSqsV1Tags, AwsStepFunctionsV1Tags. 
- Edited MessagingSchema and DatabaseSchema to not use the deleted classes in the CreateAws{Service}Tags() functions
- Did same as above for MessagingSchema unit tests
- Deleted V1 snapshots and any uses of V1 in the integration tests for AWS managed services.

### DIFF
--- a/tracer/test/snapshots/AwsEventBridgeTests.NetCore.SchemaV0.verified.txt
+++ b/tracer/test/snapshots/AwsEventBridgeTests.NetCore.SchemaV0.verified.txt
@@ -1,4 +1,4 @@
-﻿[
+[
   {
     TraceId: Id_1,
     SpanId: Id_2,
@@ -64,7 +64,7 @@
       language: dotnet,
       rulename: MyEventBus,
       runtime-id: Guid_1,
-      span.kind: client,
+      span.kind: producer,
       _dd.base_service: Samples.AWS.EventBridge
     },
     Metrics: {
--- a/tracer/test/snapshots/AwsSnsTests.NetCore.SchemaV0.pre3.7.3.verified.txt
+++ b/tracer/test/snapshots/AwsSnsTests.NetCore.SchemaV0.pre3.7.3.verified.txt
@@ -65,7 +65,7 @@
       http.url: http://localhost:00000/,
       language: dotnet,
       runtime-id: Guid_1,
-      span.kind: client,
+      span.kind: producer,
       topicname: MyTopic,
       _dd.base_service: Samples.AWS.SimpleNotificationService
     },
@@ -96,7 +96,7 @@
       http.url: http://localhost:00000/,
       language: dotnet,
       runtime-id: Guid_1,
-      span.kind: client,
+      span.kind: producer,
       topicname: MyTopic,
       _dd.base_service: Samples.AWS.SimpleNotificationService
     },
--- a/tracer/test/snapshots/AwsSnsTests.NetCore.SchemaV0.verified.txt
+++ b/tracer/test/snapshots/AwsSnsTests.NetCore.SchemaV0.verified.txt
@@ -65,7 +65,7 @@
       http.url: http://localhost:00000/,
       language: dotnet,
       runtime-id: Guid_1,
-      span.kind: client,
+      span.kind: producer,
       topicname: MyTopic,
       _dd.base_service: Samples.AWS.SimpleNotificationService
     },
@@ -96,7 +96,7 @@
       http.url: http://localhost:00000/,
       language: dotnet,
       runtime-id: Guid_1,
-      span.kind: client,
+      span.kind: producer,
       topicname: MyTopic,
       _dd.base_service: Samples.AWS.SimpleNotificationService
     },
--- a/tracer/test/snapshots/AwsSqsTests.NetCore.SchemaV0.pre3_7_300.verified.txt
+++ b/tracer/test/snapshots/AwsSqsTests.NetCore.SchemaV0.pre3_7_300.verified.txt
@@ -228,7 +228,7 @@
       language: dotnet,
       queuename: MyAsyncSQSQueue,
       runtime-id: Guid_1,
-      span.kind: client,
+      span.kind: producer,
       _dd.base_service: Samples.AWS.SQS
     },
     Metrics: {
@@ -259,7 +259,7 @@
       language: dotnet,
       queuename: MyAsyncSQSQueue,
       runtime-id: Guid_1,
-      span.kind: client,
+      span.kind: consumer,
       _dd.base_service: Samples.AWS.SQS
     },
     Metrics: {
@@ -321,7 +321,7 @@
       language: dotnet,
       queuename: MyAsyncSQSQueue2,
       runtime-id: Guid_1,
-      span.kind: client,
+      span.kind: producer,
       _dd.base_service: Samples.AWS.SQS
     },
     Metrics: {
@@ -352,7 +352,7 @@
       language: dotnet,
       queuename: MyAsyncSQSQueue2,
       runtime-id: Guid_1,
-      span.kind: client,
+      span.kind: consumer,
       _dd.base_service: Samples.AWS.SQS
     },
     Metrics: {
@@ -414,7 +414,7 @@
       language: dotnet,
       queuename: MyAsyncSQSQueue,
       runtime-id: Guid_1,
-      span.kind: client,
+      span.kind: producer,
       _dd.base_service: Samples.AWS.SQS
     },
     Metrics: {
@@ -445,7 +445,7 @@
       language: dotnet,
       queuename: MyAsyncSQSQueue,
       runtime-id: Guid_1,
-      span.kind: client,
+      span.kind: consumer,
       _dd.base_service: Samples.AWS.SQS
     },
     Metrics: {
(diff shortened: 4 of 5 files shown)


tracer/src/Datadog.Trace/Tagging/AwsSdkTags.cs
```diff
@@ -37,6 +37,12 @@ internal abstract partial class AwsSdkTags : InstrumentationTags, IHasStatusCode
         [Tag(Trace.Tags.AwsService)]
         public string Service { get; set; }
 
+        [Tag(Trace.Tags.PeerService)]
+        public string PeerService { get; set; }
+
+        [Tag(Trace.Tags.PeerServiceSource)]
+        public string PeerServiceSource { get; set; }
+
```
pablomartinezbernardo, line 45: If we are planning to set them in future PRs, let's wait until that logic is ready before actually adding the tags empty

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AWS/AwsDynamoDbTests.cs
```diff
@@ -28,16 +28,16 @@ public AwsDynamoDbTests(ITestOutputHelper output)
...
         public override Result ValidateIntegrationSpan(MockSpan span, string metadataSchemaVersion) => span.IsAwsDynamoDb(metadataSchemaVersion);
 
         [SkippableTheory]
         [MemberData(nameof(GetEnabledConfig))]
         [Trait("Category", "EndToEnd")]
-        public async Task SubmitsTraces(string packageVersion, string metadataSchemaVersion)
+        public async Task SubmitsTraces(string packageVersion)
         {
+            string metadataSchemaVersion = "v0";
             SetEnvironmentVariable("DD_TRACE_SPAN_ATTRIBUTE_SCHEMA", metadataSchemaVersion);
```
pablomartinezbernardo: Let's skip setting `DD_TRACE_SPAN_ATTRIBUTE_SCHEMA` given that it now does nothing for these tests.

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AWS/AwsDynamoDbTests.cs
```diff
@@ -28,16 +28,16 @@ public AwsDynamoDbTests(ITestOutputHelper output)
...
             var isExternalSpan = metadataSchemaVersion == "v0";
             var clientSpanServiceName = isExternalSpan ? $"{EnvironmentHelper.FullSampleName}-aws-dynamodb" : EnvironmentHelper.FullSampleName;
```
pablomartinezbernardo: We can clean these up now

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AWS/AwsDynamoDbTests.cs
pablomartinezbernardo: Can we just completely get rid of the `metadataSchemaVersion` variable? Is there a way to `ValidateIntegrationSpans` that doesn't make us call this "v0"? Don't really like calling the snapshots v0 because this is now a mix between v0 and v1

tracer/test/Datadog.Trace.Tests/Configuration/Schema/MessagingSchemaTests.cs
pablomartinezbernardo: Can we skip these tests altogether now? I don't think they are testing much anymore, maybe you were planning on repurposing them in future PRs?

tracer/src/Datadog.Trace/Tagging/AwsSdkTags.cs
lucaspimentel: fwiw, if the tag has no value (`null` or empty string), we don't serialize it into MessagePack when sending to the agent, so this is pretty harmless.

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AWS/AwsDynamoDbTests.cs
rithikanarayan: `ValidateIntegationSpans` is a shared function between many tests, including those that still have V1 schemas, so for readability and consistency between these tests I think we should keep the `metadataSchemaVersion`. I have removed its uses where possible (AWS-specific functions).

tracer/src/Datadog.Trace/Tagging/AwsSdkTags.cs
lucaspimentel: For context, this work was going to be a single PR, but it was growing too big so @rithikanarayan decided to split it into two PRs. She will be setting these tags in the next PR.

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AWS/AwsDynamoDbTests.cs
```diff
@@ -28,19 +28,17 @@ public AwsDynamoDbTests(ITestOutputHelper output)
...
         [SkippableTheory]
         [MemberData(nameof(GetEnabledConfig))]
         [Trait("Category", "EndToEnd")]
-        public async Task SubmitsTraces(string packageVersion, string metadataSchemaVersion)
+        public async Task SubmitsTraces(string packageVersion)
         {
-            SetEnvironmentVariable("DD_TRACE_SPAN_ATTRIBUTE_SCHEMA", metadataSchemaVersion);
-            var isExternalSpan = metadataSchemaVersion == "v0";
-            var clientSpanServiceName = isExternalSpan ? $"{EnvironmentHelper.FullSampleName}-aws-dynamodb" : EnvironmentHelper.FullSampleName;
+            string metadataSchemaVersion = "v0";
```
lucaspimentel: Just a small nit-pick, when a local variable never changes, we like to make it a constant. It makes it more obvious to readers, and can avoid bugs where someone changes the value later on (the compiler won't let you if it's `const`).
```suggestion
            const string metadataSchemaVersion = "v0";
```

lucaspimentel: It doesn't matter much here since this a test, but I think that following guidelines everywhere is a good practice 😅

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AWS/AwsDynamoDbTests.cs
```diff
@@ -58,7 +56,7 @@ public async Task SubmitsTraces(string packageVersion, string metadataSchemaVers
                     span => span.Tags.TryGetValue("component", out var component) && component == "aws-sdk");
 
                 dynamoDbSpans.Should().NotBeEmpty();
-                ValidateIntegrationSpans(dynamoDbSpans, metadataSchemaVersion, expectedServiceName: clientSpanServiceName, isExternalSpan);
+                ValidateIntegrationSpans(dynamoDbSpans, metadataSchemaVersion, expectedServiceName: clientSpanServiceName, true);
```
lucaspimentel: In C# you can optionally include the parameter name. It can be used to provide arguments out of order or to skip optional parameters, or, in this case, just to document which value you are passing to make the code more readable. It's especially useful when passing a Boolean ("`true` what??") 😅 

```suggestion
                ValidateIntegrationSpans(dynamoDbSpans, metadataSchemaVersion, expectedServiceName: clientSpanServiceName, isExternalSpan: true);
```

lucaspimentel (approved):
I left small coding-style comments, but feel free to do those in the next PR.

tracer/src/Datadog.Trace/Configuration/Schema/MessagingSchema.cs
```diff
@@ -78,41 +78,17 @@ public MsmqTags CreateMsmqTags(string spanKind)
                 _ => new MsmqV1Tags(spanKind),
             };
 
-        public AwsS3Tags CreateAwsS3Tags(string spanKind) => _version switch
-        {
-            SchemaVersion.V0 when !_peerServiceTagsEnabled => new AwsS3Tags(),
-            _ => new AwsS3V1Tags(spanKind),
-        };
+        public AwsS3Tags CreateAwsS3Tags(string spanKind) => new AwsS3Tags();
```
lucaspimentel: These types all have constructors that take a `spanKind`, so we should pass that long.

```suggestion
        public AwsS3Tags CreateAwsS3Tags(string spanKind) => new AwsS3Tags(spanKind);
```

You already did this with `AwsKinesisTags` and `AwsStepFunctionsTags` below.

PR #7747, opened 2025-10-31
Add fix for whitespace in exporter settings and metrics URL

andrewlock:
## Summary of changes

- Trim whitespace in the exporter settings values before trying to use them
- Fix a bug where having a whitespace in the `DD_DOGSTATSD_SOCKET` variable would cause incorrect (and invalid) UDS config

## Reason for change

This manifested in a different test in a branch, where an empty string passed in `DD_DOGSTATSD_SOCKET` was causing us to use UDS for metrics even through it's not available and not _really_ set.

## Implementation details

- `Trim()` the string variables (I don't think there's a good reason not to?)
- As we're trimming, we can switch to `IsNullOrEmpty` instead of `IsWhitespace`
- Treat empty `DD_DOGSTATSD_SOCKET` as not set

Alternatively, we could not trim, stick to using `IsWhitespace`, and just update the check I fixed to use `IsWhitespace` 🤷‍♂️ 

## Test coverage

Added a couple of unit tests. Only one test is for this specific issue, the others were just part of my investigation, and seemed reasonable.

## Other details

Discovered as part of https://github.com/DataDog/dd-trace-dotnet/pull/7724 and https://datadoghq.atlassian.net/browse/LANGPLAT-819

### DIFF
--- a/tracer/src/Datadog.Trace/Configuration/ExporterSettings.Shared.cs
+++ b/tracer/src/Datadog.Trace/Configuration/ExporterSettings.Shared.cs
@@ -46,7 +46,7 @@ namespace Datadog.Trace.Configuration
             // Check the parameters in order of precedence
             // For some cases, we allow falling back on another configuration (eg invalid url as the application will need to be restarted to fix it anyway).
             // For other cases (eg a configured unix domain socket path not found), we don't fallback as the problem could be fixed outside the application.
-            if (!string.IsNullOrWhiteSpace(agentUri))
+            if (!string.IsNullOrEmpty(agentUri))
             {
                 if (TryGetAgentUriAndTransport(agentUri!, origin, out var settings))
                 {
@@ -54,7 +54,7 @@ namespace Datadog.Trace.Configuration
                 }
             }
 
-            if (!string.IsNullOrWhiteSpace(tracesPipeName))
+            if (!string.IsNullOrEmpty(tracesPipeName))
             {
                 RecordTraceTransport(nameof(TracesTransportType.WindowsNamedPipe), origin);
 
@@ -73,7 +73,7 @@ namespace Datadog.Trace.Configuration
 
             // This property shouldn't have been introduced. We need to remove it as part of 3.0
             // But while it's here, we need to handle it properly
-            if (!string.IsNullOrWhiteSpace(tracesUnixDomainSocketPath))
+            if (!string.IsNullOrEmpty(tracesUnixDomainSocketPath))
             {
 #if NETCOREAPP3_1_OR_GREATER
                 if (TryGetAgentUriAndTransport(UnixDomainSocketPrefix + tracesUnixDomainSocketPath, origin, out var settings))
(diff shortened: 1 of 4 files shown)


tracer/src/Datadog.Trace/Configuration/ExporterSettings.Shared.cs
```diff
@@ -46,15 +46,15 @@ private TraceTransportSettings GetTraceTransport(string? agentUri, string? trace
             // Check the parameters in order of precedence
             // For some cases, we allow falling back on another configuration (eg invalid url as the application will need to be restarted to fix it anyway).
             // For other cases (eg a configured unix domain socket path not found), we don't fallback as the problem could be fixed outside the application.
-            if (!string.IsNullOrWhiteSpace(agentUri))
+            if (!string.IsNullOrEmpty(agentUri))
```
andrewlock, line 49: If we're trimming, we only need to check for empty

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -273,11 +273,11 @@ private MetricsTransportSettings ConfigureMetricsTransport(string? metricsUrl, s
                     portSource: dogStatsDPortSource,
                     out settings);
             }
-            else if (!string.IsNullOrWhiteSpace(metricsPipeName))
+            else if (!string.IsNullOrEmpty(metricsPipeName))
             {
                 settings = new MetricsTransportSettings(TransportType.NamedPipe, PipeName: metricsPipeName);
             }
-            else if (metricsUnixDomainSocketPath != null)
+            else if (!string.IsNullOrEmpty(metricsUnixDomainSocketPath))
```
andrewlock, line 280: This fixes the actual bug

tracer/test/Datadog.Trace.Tests/Configuration/ExporterSettingsTests.cs
```diff
@@ -92,6 +92,43 @@ public void Metrics_DogStatsdUrl_UdsOnLinux()
...
+            AssertUdsIsConfigured(config, socket);
+            AssertMetricsUdpIsConfigured(config);
+            config.ValidationWarnings.Should().BeEmpty();
+        }
+
+        [Theory]
+        [InlineData(null)]
+        [InlineData("")]
+        [InlineData(" ")]
+        public void Metrics_UdsTraceAgent_EmptyDogStatsdSocket_UsesDefaultUdp(string metricsSocket)
```
andrewlock, line 124: This is the test that fails before the fix

tracer/test/Datadog.Trace.Tests/Configuration/ExporterSettingsTests.cs
```diff
@@ -104,7 +141,7 @@ public void Metrics_DogStatsdUrl_UdsUnsupported_UsesDefaultUdpOnWindows()
         }
 
         [Fact]
-        public void Metrics_DogStatsdUrl_UdsUnsupported_UsesDefaultUdsOnLinux()
+        public void Metrics_DogStatsdUrl_UdsUnsupported_UsesDefaultUdpOnLinux()
```
andrewlock, line 144: typo in the test name

PR #6683, opened 2025-02-18
Setup CosmosDB Linux Docker container to run tests on CI

bouwkast:
## Summary of changes

This adds a `cosmosdb-emulator` to the `docker-compose.yml` so that we can run the tests on CI.

## Reason for change

For each update to the CosmosDB NuGet I manually test it out to see if our instrumentation still works, I'd like to not have to do that.

## Implementation details

Followed: https://learn.microsoft.com/en-us/azure/cosmos-db/how-to-develop-emulator?pivots=api-nosql&tabs=docker-linux%2Ccsharp

and

https://learn.microsoft.com/en-us/azure/cosmos-db/emulator

## Test coverage

Now there are tests in CI again (hopefully)
Ran locally with the latest version and seemed good

## Other details

```
cosmosdb-emulator  | This is an evaluation version.  There are [88] days left in the evaluation period.
cosmosdb-emulator  | 2.14.21.0 (1c783d8f)
cosmosdb-emulator  | Copyright (C) Microsoft Corporation. All rights reserved.
```

We only have 88 days until I need to run them manually again 😭 

The solution to this is to apparently delete the container image and pull it back down.

`docker compose down`
`docker rmi mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator:latest`
`docker compose up cosmosdb-emulator`

Unsure if the days will actually go down though 😅

PR #7752, opened 2025-11-03
Cosmos DB partial integration testing

pablomartinezbernardo:
## Summary of changes

Integration test Cosmos DB instrumentation with [mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator:vnext-preview](https://learn.microsoft.com/en-us/azure/cosmos-db/emulator-linux).

Due to incomplete support in the emulator, only item queries can be currently tested.

## Reason for change

No integration tests exist for Cosmos DB. This PR changes it using the under development Linux-based emulator. [Compatibility](https://learn.microsoft.com/en-us/azure/cosmos-db/emulator-linux#feature-support) is limited, but it supports items operations, which are arguably used the most. The emulator will also help development of future Cosmos DB support.

## Implementation details

Keep current test to keep supporting full manual testing, while doing specific testing for CI.

### DIFF
--- a/tracer/test/test-applications/integrations/Samples.CosmosDb.Vnext/App.config
+++ b/tracer/test/test-applications/integrations/Samples.CosmosDb.Vnext/App.config
@@ -4,6 +4,10 @@
         <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6.1" />
     </startup>
   <appSettings>
+    <!--
+        This is the well-known Cosmos DB emulator key, safe for public repositories.
+        See https://learn.microsoft.com/en-us/azure/cosmos-db/emulator
+    -->
     <add key="PrimaryKey" value="C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw==" />
   </appSettings>
 </configuration>


tracer/build/_build/Build.Steps.cs
```diff
@@ -1835,7 +1835,7 @@ string GetFilter()
 
                 var filter = (string.IsNullOrWhiteSpace(Filter), IsWin) switch
                 {
-                    (false, _) => $"({Filter}){dockerFilter}{armFilter}",
+                    (false, _) => $"({Filter})&(SkipInCI!=True){dockerFilter}{armFilter}",
```
pablomartinezbernardo, line 1838: After adding [IntegrationTestFilter: DockerGroup=$(dockerGroup)
](https://github.com/DataDog/dd-trace-dotnet/blob/30d2142e021d34f690f3b73865a9be82024b9585/.azure-pipelines/ultimate-pipeline.yml#L2305) to the jobs, we started falling through this case, which ignored `SkipInCI`

tracer/build/PackageVersionsGeneratorDefinitions.json
```diff
@@ -675,10 +675,10 @@
     "IntegrationName": "CosmosDb",
     "SampleProjectName": "Samples.CosmosDb",
     "NugetPackageSearchName": "Microsoft.Azure.Cosmos",
-    "MinVersion": "3.6.0",
+    "MinVersion": "3.12.0",
```
pablomartinezbernardo, line 693: We need to use `HttpClientFactory` when constructing the SDK client for use with the emulator, which has a self-signed certificate that .net fails to validate by default. `3.12.0` is the first version that supports `HttpClientFactory`

tracer/test/test-applications/integrations/Samples.CosmosDb/Program.cs
```diff
@@ -104,9 +144,6 @@ private async Task CreateDatabaseAsync()
             // Create a new database
             database = await cosmosClient.CreateDatabaseIfNotExistsAsync(databaseId);
             Console.WriteLine($"{DateTime.Now:o}: Created Database: {database.Id}\n");
-
-            var user = await database.CreateUserAsync("user");
-            Console.WriteLine($"{DateTime.Now:o}: Created user: {user.Resource.Id}\n");
```
pablomartinezbernardo, line 109: Emulator doesn't support creating users

tracer/test/test-applications/integrations/Samples.CosmosDb/Program.cs
```diff
@@ -66,13 +74,36 @@ public static async Task Main(string[] args)
...
-                new CosmosClientOptions() 
-                { 
-                    ApplicationName = "CosmosDBDotnetQuickstart", 
+            var clientOptions =
+                new CosmosClientOptions()
+                {
+                    ApplicationName = "CosmosDBDotnetQuickstart",
                     RequestTimeout = TimeSpan.FromMinutes(10),
-                    OpenTcpConnectionTimeout = TimeSpan.FromMinutes(1),
+                    ConnectionMode = ConnectionMode.Gateway,
```
pablomartinezbernardo: Emulator only supports gateway mode

tracer/test/test-applications/integrations/Samples.CosmosDb/Program.cs
```diff
@@ -66,13 +74,36 @@ public static async Task Main(string[] args)
...
+                    LimitToEndpoint = true,
```
pablomartinezbernardo: https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.cosmos.cosmosclientoptions.limittoendpoint?view=azure-dotnet

Otherwise the SDK will attempt to connect to `127.0.0.1` after the first request

bouwkast (commented):
Unsure how I feel about this, one reason why I closed https://github.com/DataDog/dd-trace-dotnet/pull/6683 was because the docker container couldn't run all of the necessary commands.

I think this is running into the same / similar issue where only a small subset of the functionality of CosmosDb works within the docker container. 

I think we should maybe consider to go back to running the Cosmos Emulator (Windows) in CI as opposed to this

pablomartinezbernardo:
> I think this is running into the same / similar issue where only a small subset of the functionality of CosmosDb works within the docker container.

It is a subset, but I would argue against it being small. The only opeartions that we are missing are queries over databases, containers and users. Operations on items, which are the most common scenarios, are supported. Some complex operations on items are not supported, but that becomes irrelevant, as testing the simpler cases completely tests our instrumentation.

> I think we should maybe consider to go back to running the Cosmos Emulator (Windows) in CI as opposed to this

To defend your case... it is not only a Windows emulator, it can be run on Linux. Just not on arm64. It also looks like it will be heavier than the linux-based one:

```
docker images mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator                                                                                                                                                                                           
REPOSITORY                                               TAG             IMAGE ID       CREATED        SIZE
mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator   latest          bbf28177ceb0   2 months ago   3.06GB
mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator   vnext-preview   efe42b6f3e32   3 months ago   900MB
```

It's a valid option as well, the con is that there's no way to run it on arm machines (Not even with QEMU), which means
* No local testing for mac users
* No arm CI, as we are running dependencies on arm as well

Let's hear more opinions about this!

bouwkast:
This is the emulator that I was mentioning (non-Docker) https://learn.microsoft.com/en-us/azure/cosmos-db/emulator-release-notes - but yeah your limitations still apply.

We used to run this in CI, but it proved to be too flakey.

Could we do something where we split between the old way of testing and the new way of testing -> 

```
{
    "IntegrationName": "CosmosDb",
    "SampleProjectName": "Samples.CosmosDb",
    "NugetPackageSearchName": "Microsoft.Azure.Cosmos",
    "MinVersion": "3.6.0",
    "MaxVersionExclusive": "4.0.0",
    "SpecificVersions": [
      "3.6.*",
      "3.*.*"
    ],
    "VersionConditions": [{
      "MinVersion": "3.29.0",
      "ExcludeTargetFrameworks": ["netcoreapp2.1", "netcoreapp3.0"]
    }]
  },
  {
    "IntegrationName": "CosmosDb_Emulator",
    "SampleProjectName": "Samples.CosmosDb",
    "NugetPackageSearchName": "Microsoft.Azure.Cosmos",
    "MinVersion": "3.12.0",
    "MaxVersionExclusive": "4.0.0",
    "SpecificVersions": [
      "3.12.*",
      "3.*.*"
    ],
    "VersionConditions": [{
      "MinVersion": "3.29.0",
      "ExcludeTargetFrameworks": ["netcoreapp2.1", "netcoreapp3.0"]
    }]
```

docker-compose.yml
```diff
@@ -65,6 +65,10 @@ services:
     - ACCEPT_EULA=Y
     - SA_PASSWORD=Strong!Passw0rd
 
+  cosmosdb-emulator_arm64:
+    image: mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator:vnext-preview
+    command: ["--protocol", "https"]
+
```
bouwkast, line 71: Will require a VM update after merging to not slow down CI for too long - I'd think that we could probably do so around .NET 10 GA release

bouwkast: 😢 
We will have to update the VMs for this, we can merge it and not update the VMs but there will be some additional pipeline time hit. 

I guess fine, I'll make another task to update them again after this gets merged.

tracer/test/test-applications/integrations/Samples.CosmosDb.Vnext/App.config
```diff
@@ -0,0 +1,9 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<configuration>
+    <startup> 
+        <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6.1" />
+    </startup>
+  <appSettings>
+    <add key="PrimaryKey" value="C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw==" />
```
lucaspimentel, line 11: Maybe add a comment here so other know where this came from. I'm sure it will get pinged by some security software somewhere 😅 
```suggestion
    <!--
        This is the well-known Cosmos DB emulator key, safe for public repositories.
        See https://learn.microsoft.com/en-us/azure/cosmos-db/emulator
    -->
    <add key="PrimaryKey" value="C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw==" />
```

PR #7754, opened 2025-11-03
Improve exception redaction to handle Aggregate Exceptions better

andrewlock:
## Summary of changes

Improve handling of telemetry redaction of `AggregateException`

## Reason for change

While investigating something else, I noticed that we were not including all the inner exceptions when recording `AggregateException`. If there's more than one exception previously we were only including the first exception.

## Implementation details

Update `ExceptionRedactor` to include all the inner exceptions of an `AggregateException`

## Test coverage

Added a couple of unit tests

### DIFF
--- a/tracer/test/Datadog.Trace.Tests/Logging/ExceptionRedactorTests.cs
+++ b/tracer/test/Datadog.Trace.Tests/Logging/ExceptionRedactorTests.cs
@@ -102,6 +102,31 @@ public class ExceptionRedactorTests
     }
 
     [Fact]
+    public async Task Redact_FlattensAggregateExceptions_WhenAggregateExceptionWithMultipleExceptions()
+    {
+        var tcs = new TaskCompletionSource<Exception>();
+
+        var task1 = Task.Run(ThrowException);
+        var task2 = Task.Run(ThrowException);
+
+        // Create AggregateException
+        var task3 = Task.WhenAll(task1, task2);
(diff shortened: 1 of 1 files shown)


tracer/test/Datadog.Trace.Tests/Logging/ExceptionRedactorTests.cs
```diff
@@ -67,6 +68,39 @@ public void Redact_IncludesInnerExceptionWhenHaveStackFrames()
...
+                .ContinueWith(t => { tcs.TrySetResult(t.Exception); }, TaskContinuationOptions.OnlyOnFaulted);
+
+        var ex = await tcs.Task;
+
+        var redacted = ExceptionRedactor.Redact(ex);
+
+        redacted.Should().StartWith("System.AggregateException (Multiple Exceptions) ---> System.Exception" + Environment.NewLine);
+        redacted.Should().Contain("---> System.Exception", Exactly.Twice());
+    }
+
```
NachoEchevarria, line 103: NIT: While flatten should do the job, is it worth to test nested exceptions?

NachoEchevarria (approved):
Nice!

tracer/test/Datadog.Trace.Tests/Logging/ExceptionRedactorTests.cs
andrewlock: Added 👍

PR #7768, opened 2025-11-04
Fix (more) EventHubs flaky batch tests

pablomartinezbernardo:
## Summary of changes

Same issue with flaky tests as #7734, I missed more usages of the function

> The test expectations were incorrect, as ReadEventsFromPartitionAsync may call ReceiveAsync [several times depending on timing](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Azure.Messaging.EventHubs/src/Consumer/EventHubConsumerClient.cs#L520).


andrewlock (approved):
Perfect, thank you!

PR #7775, opened 2025-11-04
Add process tags to DSM

vandonr:
## Summary of changes

 - add a hash of process tags to the base hash of pathway checkpoint hashes
 - write the process tags in DSM payloads(so that the backend can resolve the right service name)

Since this makes us query the process tags during the tracer initialization, I removed the Lazy mechanism that I had put in place previously, because there would be nothing lazy about it anymore.
This means that values that are available only later in the app life (like IIS entrypoint for instance) are not accessible.

still controlled by the same feature flag as in #7651 

This is the dotnet equivalent of https://github.com/DataDog/dd-trace-java/pull/8836

## Reason for change

[AIDM-195]

## Other details

### DIFF
--- a/tracer/src/Datadog.Trace/ProcessTags.cs
+++ b/tracer/src/Datadog.Trace/ProcessTags.cs
@@ -20,12 +20,7 @@ internal static class ProcessTags
     public const string EntrypointBasedir = "entrypoint.basedir";
     public const string EntrypointWorkdir = "entrypoint.workdir";
 
-    private static readonly Lazy<string> LazySerializedTags = new(GetSerializedTags);
-
-    public static string SerializedTags
-    {
-        get => LazySerializedTags.Value;
-    }
+    public static readonly string SerializedTags = GetSerializedTags();
 
     /// <summary>
     /// From the full path of a directory, get the name of the leaf directory.
@@ -43,7 +38,7 @@ internal static class ProcessTags
         List<KeyValuePair<string, string?>> tags =
         [
             new(EntrypointBasedir, GetLastPathSegment(AppContext.BaseDirectory)),
-            new(EntrypointName, EntryAssemblyLocator.GetEntryAssembly()?.EntryPoint?.DeclaringType?.FullName),
+            new(EntrypointName, GetEntryPointName()),
             // workdir can be changed by the code, but we consider that capturing the value when this is called is good enough
             new(EntrypointWorkdir, GetLastPathSegment(Environment.CurrentDirectory))
         ];
@@ -60,6 +55,11 @@ internal static class ProcessTags
 
         serializedTags.Remove(serializedTags.Length - 1, length: 1); // remove last comma
         return StringBuilderCache.GetStringAndRelease(serializedTags);
+    }
+
+    private static string? GetEntryPointName()
+    {
+        return EntryAssemblyLocator.GetEntryAssembly()?.EntryPoint?.DeclaringType?.FullName;
     }
 
     private static string NormalizeTagValue(string tagValue)


tracer/src/Datadog.Trace/DataStreamsMonitoring/Aggregation/DataStreamsMessagePackFormatter.cs
```diff
@@ -94,13 +97,14 @@ private static long GetProductsMask(TracerSettings tracerSettings, ProfilerSetti
 
         public int Serialize(Stream stream, long bucketDurationNs, List<SerializableStatsBucket> statsBuckets, List<SerializableBacklogBucket> backlogsBuckets)
         {
+            var withProcessTags = _writeProcessTags && !string.IsNullOrEmpty(ProcessTags.SerializedTags);
             var bytesWritten = 0;
 
             // Should be in sync with Java
-            // https://github.com/DataDog/dd-trace-java/blob/a4b7a7b177709e6bdfd9261904cb9a777e4febbe/dd-trace-core/src/main/java/datadog/trace/core/datastreams/MsgPackDatastreamsPayloadWriter.java#L35
+            // https://github.com/DataDog/dd-trace-java/blob/master/dd-trace-core/src/main/java/datadog/trace/core/datastreams/MsgPackDatastreamsPayloadWriter.java
```
andrewlock, line 104: nit: don't use `master` for github links, because they break 😅 
```suggestion
            // https://github.com/DataDog/dd-trace-java/blob/8db72c0988adb1010c1f2d6b028b93b10cd10ff2/dd-trace-core/src/main/java/datadog/trace/core/datastreams/MsgPackDatastreamsPayloadWriter.java
```

tracer/src/Datadog.Trace/DataStreamsMonitoring/Aggregation/DataStreamsMessagePackFormatter.cs
```diff
@@ -194,6 +198,12 @@ public int Serialize(Stream stream, long bucketDurationNs, List<SerializableStat
             bytesWritten += MessagePackBinary.WriteStringBytes(stream, _productMaskBytes);
             bytesWritten += MessagePackBinary.WriteInt64(stream, _productMask);
 
+            if (withProcessTags)
+            {
+                bytesWritten += MessagePackBinary.WriteStringBytes(stream, _processTagsBytes);
+                bytesWritten += MessagePackBinary.WriteString(stream, ProcessTags.SerializedTags);
```
andrewlock, line 204: nit: not a fan of the fact that these are static globals, global state is the enemy of testing. Could/should we just pass it in as a parameter?

tracer/src/Datadog.Trace/DataStreamsMonitoring/Hashes/HashHelper.cs
```diff
@@ -32,6 +32,11 @@ public static NodeHashBase CalculateNodeHashBase(string service, string? env, st
             hash = FnvHash64.GenerateHash(primaryTag!, HashVersion, hash);
         }
 
+        if (!string.IsNullOrEmpty(processTags))
+        {
+            hash = FnvHash64.GenerateHash(processTags!, HashVersion, hash);
```
andrewlock: nit: if you use `StringUtil.IsNullOrEmpty` you don't need the null-forgiving to keep .NET Framework happy 🙂 (I think you need to add an extra `using` though)
```suggestion
        if (!StringUtil.IsNullOrEmpty(processTags))
        {
            hash = FnvHash64.GenerateHash(processTags, HashVersion, hash);
```

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
```diff
@@ -59,7 +60,7 @@ public static DataStreamsManager Create(
                          ? DataStreamsWriter.Create(settings, profilerSettings, discoveryService, defaultServiceName)
                          : null;
 
-        return new DataStreamsManager(settings.Environment, defaultServiceName, writer, settings.IsDataStreamsMonitoringInDefaultState);
+        return new DataStreamsManager(settings.Environment, defaultServiceName, writer, settings.IsDataStreamsMonitoringInDefaultState, settings.PropagateProcessTags ? ProcessTags.SerializedTags : null);
```
andrewlock, line 63: This calls `ProcessTags.SerializedTags`, which realizes the tags. Given that the `DataStreamsManager` is created in the app initialization (_before_ a request has arrived), I think this completely breaks the "lazy" aspect of the serialized tags? 🤔

tracer/src/Datadog.Trace/DataStreamsMonitoring/Aggregation/DataStreamsMessagePackFormatter.cs
vandonr: ah hmm well, yes they do if the file moves, but also a link to a fixed sha loses its relevance as time passes (for instance, here it was pointing to code that had changed a lot). This could be especially problematic for a comment that says "keep in sync with ...".

tracer/src/Datadog.Trace/DataStreamsMonitoring/Aggregation/DataStreamsMessagePackFormatter.cs
vandonr: hmm yes we can, but at some point we'd query them from the static global anyway. I don't know what would be the alternative for something that's supposed to be a singleton.

tracer/src/Datadog.Trace/DataStreamsMonitoring/Hashes/HashHelper.cs
vandonr: ah good to know thanks, I'll do that

vandonr:
> I'll leave the DSM team to confirm the correctness (compared to what we need + other implementations) but I think there's a fundamental issue which will break process tags in general for IIS apps.
> 
> The `DataStreamsManager` grabs the process tags immediately, in order to calculate the `NodeHashBase`, which triggers resolution of the tags. But this can happen too early and cause you to get the wrong process tags, particularly in asp.net.
> 
> I'm not sure what the best solution is here. You _could_ make the nodehashbase calculation lazy too, but that still doesn't stop you grabbing it "too early". And presumably it's not acceptable for the process tags to ever change, as it would break all the DSM hashes. So you're stuck with DSM (which is on by default now) potentially breaking all your other process tags calculations 😕

yes I know, but I don't see a way out of this. I'm thinking about just giving up on application entrypoint for IIS, since it seems every time I add process tags somewhere in the tracer it creates more issues because it can be requested too early.
Since process tags are not a "nice to have" but something mandatory as soon as customer starts to use them to build their service name, we cannot do any retry like we do in other places.
As you say, I can make this lazy, but there is still no guarantee. I was thinking of just "letting it rip", i.e. we try to get the entry point, and if we don't, well we don't.
It's more something I need to check with the service renaming team than with DSM.

tracer/src/Datadog.Trace/DataStreamsMonitoring/Aggregation/DataStreamsMessagePackFormatter.cs
andrewlock: It's probably not a big deal here, but there's a "correct" answer:

> I don't know what would be the alternative for something that's supposed to be a singleton.

The answer is to only have a thin global static at the "outside" of the app. Everything else uses dependency inversion to inject the values in. Makes everything inherently testable, and removes the implicit dependencies between components you get otherwise.

andrewlock (approved):

PR #7777, opened 2025-11-04
Avoid creating new tasks on wait

kr-igor:
## Summary of changes
Avoid creating extra tasks/timer. DSM is not use a dedicated background thread.

## Reason for change
See [this](https://datadoghq.atlassian.net/browse/APMS-17551) ticket.

## Implementation details
Based on the information we have it seems like the main reason for extra overhead comes from the  "Finalizer" thread which calls `CloseHandle` too frequently. It's not clear what causes the kernel resource allocation/disposal and why it only happens in customer's environment.

### DIFF
--- a/tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
+++ b/tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
@@ -290,7 +290,7 @@ internal class DataStreamsWriter : IDataStreamsWriter
             var currentFlushTcs = Volatile.Read(ref _currentFlushTcs);
 
             ProcessQueue(flushRequested == 1, currentFlushTcs);
-            Thread.Sleep(5);
+            Thread.Sleep(20);
         }
 
         _tcsBackgroundThread.TrySetResult(true);


tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
```diff
@@ -313,13 +315,8 @@ private async Task ProcessQueueLoopAsync()
                 continue;
             }
 
-            // The logic is copied from https://github.com/dotnet/runtime/blob/main/src/libraries/Common/tests/System/Threading/Tasks/TaskTimeoutExtensions.cs#L26
-            // and modified to avoid dealing with exceptions
-            var tcs = new TaskCompletionSource<bool>();
-            using (new Timer(s => ((TaskCompletionSource<bool>)s!).SetResult(true), tcs, _waitTimeSpan, Timeout.InfiniteTimeSpan))
-            {
-                await Task.WhenAny(_processExit.Task, tcs.Task).ConfigureAwait(false);
```
bouwkast, line 321: We aren't responding / reacting to `_processExit.Task` anymore but I don't think a big deal

bouwkast (commented):
LGTM seems like it should be a lot less allocations now.

Were you able to reproduce this or benchmark at all or is this just an educated guess?

bouwkast (approved):

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
andrewlock: > We aren't responding / reacting to _processExit.Task anymore but I don't think a big deal

Yeah, given it's only a 15ms wait, I think that's fine, but it does make me wonder... a 15ms wait period makes this is a pretty fast spinning loop, no? 😕For context, we flush traces every _1000ms_ (1s)... On low-core (1) machines would this not potentially be taking up a lot of thread/CPU time?

What's more, the `_resetEvent.Wait()` that's never triggered is effectively a `Thread.Sleep()`. And doing a `Thread.Sleep` on a thread pool thread is generally not a good idea, as it blocks anything else doing work. @tonyredondo did a bunch of work to remove all the `Thread.Sleep()` from our async code for exactly this reason. If we _do_ want to dedicate a thread to this, I think we need to turn this loop into a sync loop and create a dedicated long-running thread, but that might be more work than we'd like.

Overall, I don't know if we have any evidence that the TCS and Tasks here are the source of the perf issue yet? and I wonder if it's the potentially small value of `_waitTimeSpan` that's causing the issue due to this loop spinning too fast? 🤔 Have we considered alternatives like increasing the buffering wait time? Or taking an approach similar to the agent writer where we have a longer buffering period, with dedicated flushing when the buffers get to a certain "fullness"? It'll be a trade-off obviously, but I'm not sure which is best

kr-igor:
>Overall, I don't know if we have any evidence that the TCS and Tasks here are the source of the perf issue yet?

@andrewlock  From the `.nettrace` it seems like a bunch of OS resources get created / destroyed in this loop. I was not able to reproduce this in any of my setups. Tried various environments / thread pool configurations. Maybe this is related to the underlying OS architecture and differences in actual thread scheduling in low resource scenarios. 

Screenshot 2025-11-06 at 9 09 23 AM

My guess is that it comes from [this](https://github.com/dotnet/runtime/blob/12b8d8387495b9552e85f3bd374e7fa5929a857f/src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/Task.cs#L1637) code. Tasks wait due to thread pool starvation, causing `ManualResetEventSlim` to allocate a wait handle [here](https://github.com/dotnet/runtime/blob/53af66a48215f0b7833dce499343eccae1dc9dff/src/libraries/System.Private.CoreLib/src/System/Threading/ManualResetEventSlim.cs#L217). 

I didn't know that `_resetEvent.wait` is equivalent of thread.sleep, my assumption was that is uses some Spin logic under the hood. I'll dig a bit more and will try to find a better way.

andrewlock:
> I didn't know that _resetEvent.wait is equivalent of thread.sleep

To be clear, I haven't looked into what it's _actually_ like, but it's fundamentally not releasing the thread which is what I was getting at 🙂

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
```diff
@@ -255,8 +241,7 @@ private async Task WriteToApiAsync()
             // This flushes on the same thread as the processing loop
             var data = new ArraySegment<byte>(_serializationBuffer.GetBuffer(), offset: 0, (int)_serializationBuffer.Length);
 
-            var success = await _api.SendAsync(data).ConfigureAwait(false);
-
+            var success = _api.SendAsync(data).ConfigureAwait(false).GetAwaiter().GetResult();
```
andrewlock, line 244: Definitely can't do `GetAwaiter().GetResult();` 🙁

PR #7780, opened 2025-11-05
Add process tags to telemetry

vandonr:
## Summary of changes

add processtags to the body of telemetry requests, as part of the application data.

equivalent of https://github.com/DataDog/dd-trace-java/pull/8837

## Reason for change

[AIDM-195]

## Implementation details

pretty straightforward.

## Test coverage

relies on snapshots

## Other details

### DIFF
--- a/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/Helpers/TelemetryHelperTests.cs
+++ b/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/Helpers/TelemetryHelperTests.cs
@@ -40,7 +40,8 @@ public class TelemetryHelperTests
             runtimeName: "dotnet",
             runtimeVersion: "7.0.1",
             commitSha: "testCommitSha",
-            repositoryUrl: "testRepositoryUrl");
+            repositoryUrl: "testRepositoryUrl",
+            processTags: "entrypoint.basedir:Users,entrypoint.workdir:Downloads");
         _host = new HostTelemetryData("MY_HOST", "Windows", "x64");
         _output = output;
     }


vandonr:
> I think that has the same problem as #7775 - i.e. it will immediately cause us to grab the process tags, and these will "lock in" the wrong values in IIS

less so though, no ? Because it'd only grab them when the first telemetry payload is sent, which is not _right_ as the app/tracer starts

andrewlock:
> which is not _right_ as the app/tracer starts

No, it's _right_ as the app/tracer starts 😅 i.e. the `app-started` request

vandonr:
> No, it's _right_ as the app/tracer starts 😅 i.e. the `app-started` request

ah yes, so we are even more f*cked with this, which is exactly what I said in the other PR, every new place where I add them makes the problem worse ^^

tracer/src/Datadog.Trace/Telemetry/Collectors/ApplicationTelemetryCollector.cs
```diff
@@ -30,6 +30,12 @@ public void RecordTracerSettings(
             Interlocked.Exchange(ref _gitMetadata, gitMetadata);
         }
 
+        string? processTags = null;
+        if (tracerSettings.PropagateProcessTags && !string.IsNullOrEmpty(ProcessTags.SerializedTags))
```
andrewlock, line 34: Get your filthy static global state out of my lovely testable code 😂 😛

vandonr: I will, later :p

PR #7781, opened 2025-11-05
Avoid exceptions in ProcessBasicChecksTests

NachoEchevarria:
## Summary of changes

We are getting[ the following error](https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=190458&view=logs&j=9ae0d641-16d0-5fe4-2a78-25e7d3a739c7&t=18795609-41b3-5076-f153-54e821578c6a) in master:

```
04:54:49 [DBG] Datadog.Trace.Tools.dd_dotnet.ArtifactTests: STARTED: Datadog.Trace.Tools.dd_dotnet.ArtifactTests.Checks.ProcessBasicChecksTests.WorkingWithContinuousProfiler(1, True)[0m
04:54:49 [DBG] Datadog.Trace.Tools.dd_dotnet.ArtifactTests: STARTED: Datadog.Trace.Tools.dd_dotnet.ArtifactTests.Checks.AgentConnectivityCheckTests.DetectVersionNamedPipes()
...
04:54:51 [DBG]   Failed Datadog.Trace.Tools.dd_dotnet.ArtifactTests.Checks.ProcessBasicChecksTests.WorkingWithContinuousProfiler(enabled: "1", ssiInjectionEnabled: True) [390 ms]
04:54:51 [DBG]   Error Message:
04:54:51 [DBG]    System.ComponentModel.Win32Exception : Access is denied
04:54:51 [DBG]   Stack Trace:
04:54:51 [DBG]      at System.Diagnostics.Process.StartWithCreateProcess(ProcessStartInfo startInfo)
04:54:51 [DBG]    at System.Diagnostics.Process.Start()
04:54:51 [DBG]    at System.Diagnostics.Process.Start(ProcessStartInfo startInfo)
04:54:51 [DBG]    at Datadog.Trace.Tools.dd_dotnet.ArtifactTests.ConsoleTestHelper.<StartConsole>d__5.MoveNext()
04:54:51 [DBG] --- End of stack trace from previous location where exception was thrown ---
04:54:51 [DBG]    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
04:54:51 [DBG]    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
04:54:51 [DBG]    at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd(Task task)
04:54:51 [DBG]    at Datadog.Trace.Tools.dd_dotnet.ArtifactTests.Checks.ProcessBasicChecksTests.<WorkingWithContinuousProfiler>d__13.MoveNext() in D:\a\_work\1\s\tracer\test\Datadog.Trace.Tools.dd_dotnet.ArtifactTests\Checks\ProcessBasicChecksTests.cs:line 244
04:54:51 [DBG] --- End of stack trace from previous location where exception was thrown ---
04:54:51 [DBG]    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
04:54:51 [DBG]    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
04:54:51 [DBG]    at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd(Task task)
04:54:51 [DBG] --- End of stack trace from previous location where exception was thrown ---
04:54:51 [DBG]    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
04:54:51 [DBG]    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
04:54:51 [DBG] --- End of stack trace from previous location where exception was thrown ---
04:54:51 [DBG]    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
04:54:51 [DBG]    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
04:54:51 [DBG]   Standard Output Messages:
04:54:51 [DBG]  Platform: X86
04:54:51 [DBG]  TargetPlatform: X86
04:54:51 [DBG]  Configuration: Release
04:54:51 [DBG]  TargetFramework: net48
04:54:51 [DBG]  .NET Core: False
04:54:51 [DBG]  Native Loader DLL: D:\a\_work\1\s\shared\bin\monitoring-home\win-x86\Datadog.Trace.ClrProfiler.Native.dll
04:54:51 [DBG]  Searching for CorFlags.exe in C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin
04:54:51 [DBG]  CorFlags.exe found at C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.8 Tools\x64\CorFlags.exe
04:54:51 [DBG]  Updating Samples.Console.exe using /32BITREQ+
```

 The "Access is denied" error happens when trying to start the process. This suggests that:

  1. The WorkingWithContinuousProfiler test is running at the same time as AgentConnectivityCheckTests which uses the same console app
  2. Each test case calls StartConsole, which calls PrepareSampleApp, which calls ProfilerHelper.SetCorFlags on line 46
  of ConsoleTestHelper.cs
  3. SetCorFlags modifies the same executable file in-place using CorFlags.exe
  4. One test thread is still running CorFlags.exe to modify the executable
  5. While CorFlags.exe has the file locked for writing
  6. Another test thread completes SetCorFlags() and tries to start that same executable
  7. Windows denies access because the file is still locked by CorFlags.exe from the other thread

  So the lock needs to protect not just the CorFlags modification, but also ensure the CorFlags.exe process has fully
  released the file lock before another thread can use that executable.

This error have been observed under Windows, net4, x86, which makes sense based on the ProfilerHelper.SetCorFlags conditions:

```
    protected (string Executable, string Args) PrepareSampleApp(EnvironmentHelper environmentHelper)
    {
        var sampleAppPath = environmentHelper.GetSampleApplicationPath();
        var executable = EnvironmentHelper.IsCoreClr() ? environmentHelper.GetSampleExecutionSource() : sampleAppPath;
        var args = EnvironmentHelper.IsCoreClr() ? sampleAppPath : string.Empty;

        if (EnvironmentTools.IsWindows()
         && !EnvironmentHelper.IsCoreClr()
         && !EnvironmentTools.IsTestTarget64BitProcess())
        {
            ProfilerHelper.SetCorFlags(executable, Output, !EnvironmentTools.IsTestTarget64BitProcess());
        }

        return (executable, args);
    }
```

### DIFF
--- a/tracer/test/Datadog.Trace.TestHelpers/ProfilerHelper.cs
+++ b/tracer/test/Datadog.Trace.TestHelpers/ProfilerHelper.cs
@@ -150,11 +150,18 @@ namespace Datadog.Trace.TestHelpers
                     UseShellExecute = false
                 };
 
-                var executedSuccessfully = Process.Start(opts).WaitForExit(20_000);
-
-                if (!executedSuccessfully)
+                using (var process = Process.Start(opts))
                 {
-                    throw new Exception($"Error setting CorFlags.exe {Path.GetFileName(executable)} {setBit}");
+                    if (process == null)
+                    {
+                        throw new Exception("Failed to start CorFlags process.");
+                    }
+
+                    var exited = process.WaitForExit(20_000);
+                    if (!exited)
+                    {
+                        throw new Exception($"Error setting CorFlags.exe {Path.GetFileName(executable)} {setBit}");
+                    }
                 }
 
                 _corFlagsApplied.TryAdd(executablePath, true);


tracer/test/Datadog.Trace.TestHelpers/ProfilerHelper.cs
```diff
@@ -89,53 +92,68 @@ public static async Task<Process> StartProcessWithProfiler(
 
         public static void SetCorFlags(string executable, ITestOutputHelper output, bool require32Bit)
         {
-            var corFlagsExe = _corFlagsExe;
-            var setBit = require32Bit ? "/32BITREQ+" : "/32BITREQ-";
-            if (string.IsNullOrEmpty(corFlagsExe))
+            // Use the absolute path as the key to track which executables have been modified
+            var executablePath = Path.GetFullPath(executable);
+            var lockObj = _corFlagsLocks.GetOrAdd(executablePath, _ => new object());
```
andrewlock: nit: I think it's fine to have a single lock object that's used when _applying_ the flags tbh, and to have a fast-path _outside_ the lock that checks if they're already applied.

e.g.

```csharp
private readonly object _lockObj = new();

//...

var executablePath = Path.GetFullPath(executable);
if (_corFlagsApplied.TryGetValue(executablePath, out _))
{
    return;
}

// not applied
lock (_lockObj)
{
   // check again to be safe
  if (_corFlagsApplied.TryGetValue(executablePath, out _))
  {
      return;
  }

  // do the check
}
``` 

That way we don't need a lock per executable, and we avoid taking the lock (and having that contention) every time we start an app? wdyt?

Another tiny thing is I wonder if we need to check that we've set CorFlags _correctly_ i.e. 32 bit or 64 bit. I don't _think_ so, because I _think_ we only set that based on whether we run the _whole_ test suite as 32bit or 64bit, is that correct?

NachoEchevarria: A single lock is more simple and the performance impact would be negligible. I will apply that. Thanks!
Yes, we don't need to check the corflags because the whole test suite would run either in x86 or x64

tracer/test/Datadog.Trace.TestHelpers/ProfilerHelper.cs
```diff
@@ -19,6 +20,8 @@ public class ProfilerHelper
                                                                      .Concat(new[] { "dotnet", "iisexpress" })
                                                                      .ToArray();
 
+        private static readonly object _lockObj = new();
+        private static readonly ConcurrentDictionary<string, bool> _corFlagsApplied = new ConcurrentDictionary<string, bool>(StringComparer.OrdinalIgnoreCase);
```
andrewlock, line 24: nit: _should_ we ignore case 🤔 On Linux for example there could be two different samples with the same name, just different cases 😅 I don't think it's a problem in reality though

NachoEchevarria: Actually, we only change cor vars in Windows x86 net4, so I would keep it

tracer/test/Datadog.Trace.TestHelpers/ProfilerHelper.cs
```diff
@@ -89,53 +92,72 @@ public static async Task<Process> StartProcessWithProfiler(
...
+                    UseShellExecute = false
+                };
 
-            output?.WriteLine($"Updating {Path.GetFileName(executable)} using {setBit}");
-            var opts = new ProcessStartInfo(corFlagsExe, $"\"{executable}\" {setBit} /force")
-            {
-                CreateNoWindow = true,
-                UseShellExecute = false
-            };
+                var executedSuccessfully = Process.Start(opts).WaitForExit(20_000);
```
bouwkast: nit: I think we should dispose of the `Process` from `Process.Start`?

It was already like this though so 🤷

NachoEchevarria: Good point! I have modified the code. Even thought it was already like that, it's better to dispose it. Thanks!

NachoEchevarria:

PR #7785, opened 2025-11-06
Fix null character termination in CString abstraction.

hoolioh:
## Summary of changes
Add null termination byte for CString abstraction.

## Reason for change
Testing the new libdatadog release I noticed some errors when Cstrings were passed to the process discovery native API.

## Implementation details

Make extra space for null character and add the termination character at the end of the buffer.

## Test coverage

🙈


andrewlock (approved):
Thanks!

PR #7787, opened 2025-11-06
Remove unused `PublicApiUsage` members

andrewlock:
## Summary of changes

Deletes a bunch of enum members from `PublicApiUsage` which are unused

## Reason for change

These were unused or _couldn't_ logically be called in production code, so were just adding overhead as we allocate the array to record them.

Fixed one case where we _Should_ have been recording it, but weren't

## Implementation details

FindAllReferences on each member => if it's not used in Datadog.Trace, remove it

## Test coverage

Covered by existing

## Other details

Part of a stack

- https://github.com/DataDog/dd-trace-dotnet/pull/7786
- https://github.com/DataDog/dd-trace-dotnet/pull/7787 👈
- https://github.com/DataDog/dd-trace-dotnet/pull/7788

PR #7786, opened 2025-11-06
Remove more `[PublicApi]` usages and unused code

andrewlock:
## Summary of changes

- Remove `[PublicApi]` usages that aren't necessary any more
- Rename `DebugEnabledInternal` => `DebugEnabled`
- Delete some "public" methods that aren't, and are no longer necessary (as they're unused)

## Reason for change

Cleaning up

## Implementation details

Mostly <kbd>del</kbd>

## Test coverage

Covered by existing

## Other details

Part of a stack

- https://github.com/DataDog/dd-trace-dotnet/pull/7786 👈
- https://github.com/DataDog/dd-trace-dotnet/pull/7787
- https://github.com/DataDog/dd-trace-dotnet/pull/7788

PR #7788, opened 2025-11-06
Repurpose `PublicApiAnalyzer` as `InternalForTestingAnalyzer` 

andrewlock:
## Summary of changes

Repurposes the now-defunct `PublicApiAnalyzer` to an `InternalForTestingAnalyzer`

## Reason for change

We have various methods in `Datadog.Trace` that should never be called by production code, and are only exposed to make testing functionality easier. Currently we rely on method names and comments to avoid this. With this change, we can add an attribute instead, and using the method from Datadog.Trace etc becomes a build error.

## Implementation details

- Basically just renamed `PublicApi*` -> `InternalForTesting*`
- Decorated methods that were currently relying on comments/naming to avoid being called

## Test coverage

Refactored the analyzer tests, so should still be covered the same

## Other details

Note that there are a bunch of comments for methods `// internal for testing` which are a bit different. In those cases, we _do_ call it in Datadog.Trace, the comment is just explaining why the method isn't `private`. Those remain as-is, this attribute is purely if you have a "don't call this method in production code, use that other one instead" method 😄 

Part of a stack

- https://github.com/DataDog/dd-trace-dotnet/pull/7786
- https://github.com/DataDog/dd-trace-dotnet/pull/7787
- https://github.com/DataDog/dd-trace-dotnet/pull/7788 👈

### DIFF
--- a/tracer/src/Datadog.Trace.Tools.Runner/AnalyzeInstrumentationErrorsCommand.cs
+++ b/tracer/src/Datadog.Trace.Tools.Runner/AnalyzeInstrumentationErrorsCommand.cs
@@ -197,6 +197,8 @@ internal class AnalyzeInstrumentationErrorsCommand : CommandWithExamples
             logDirectory = ProcessConfiguration.GetProcessLogDirectory(pid.Value);
         }
 
+#pragma warning disable DD0002
         return logDirectory ?? Logging.DatadogLoggingFactory.GetLogDirectory(NullConfigurationTelemetry.Instance);
+#pragma warning restore DD0002
     }
 }


bouwkast (approved):
👍

lucaspimentel:
> Note that there are a bunch of comments for methods `// internal for testing` which are a bit different. In those cases, we do call it in Datadog.Trace, the comment is just explaining why the method isn't private.

I just assume anything that's `internal` means "could be private, but we call it from test code" (why else would they be `internal`?), so those `// internal for testing` comments seem redundant (but level of care low).

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/NameValueConfigurationSource.cs
```diff
@@ -21,7 +21,7 @@ internal class NameValueConfigurationSource : StringConfigurationSource
     {
         private readonly NameValueCollection _nameValueCollection;
 
-        // Internal for testing only
+        [InternalForTesting]
         public NameValueConfigurationSource(NameValueCollection nameValueCollection)
```
lucaspimentel, line 25: This is `public`, not `internal`.

tracer/src/Datadog.Trace/SpanContext.cs
```diff
@@ -63,12 +63,12 @@ public partial class SpanContext : ISpanContext, IReadOnlyDictionary<string, str
         /// Initializes a new instance of the <see cref="SpanContext"/> class
         /// from a propagated context. <see cref="Parent"/> will be null
         /// since this is a root context locally.
-        /// Internal for testing
         /// </summary>
         /// <param name="traceId">The propagated trace id.</param>
         /// <param name="spanId">The propagated span id.</param>
         /// <param name="samplingPriority">The propagated sampling priority.</param>
         /// <param name="serviceName">The service name to propagate to child spans.</param>
+        [InternalForTesting]
```
lucaspimentel: This is `public`, not `internal`.

lucaspimentel (approved):

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/NameValueConfigurationSource.cs
andrewlock: Yeah, but public/internal is irrelevant in Datadog.Trace world😛 Everything public could be internal and vice versa 😉 Maybe we should just have an analyzer that forces us to use `public` _instead_ of internal for consistency? Or vice versa? 🤷‍♂️

andrewlock:
> > Note that there are a bunch of comments for methods `// internal for testing` which are a bit different. In those cases, we do call it in Datadog.Trace, the comment is just explaining why the method isn't private.
> 
> I just assume anything that's `internal` means "could be private, but we call it from test code" (why else would they be `internal`?), so those `// internal for testing` comments seem redundant (but level of care low).

I think there's a difference. 

- Some code is `internal`/`public` because it _needs_ to be, and it _should_ be called by other consumers. This code isn't marked by any comments today, it's just "normal" public API surface.
- Some code is `internal`/`public` _only_ because we want to call it from test code (and it should _never_ be called in "normal" Datadog.Trace usages). These are the cases tracked by the analyzer in this PR today.
- Some code is `internal`/`public` because we want to call if from test code but we _also_ call it "internally". This is code that _Should_ be private, but isn't. This is code currently marked by `//internal for testing"
  - There is a "risk" that production code in Datadog.Trace calls these methods when they shouldn't today.

I might actually tweak this analyzer to account for both cases, so we can remove the comments, and be properly protected 🤔

andrewlock:
FYI @bouwkast @lucaspimentel, I tweaked the naming of the attribute `[TestingOnly]` to avoid the whole public/internal issue 😅 I also added a variation, `[TestingAndPrivateOnly]` so that we can get rid of most of the other `// internal for testing` comments too. In summary

- `[TestingOnly]`: Should not be called directly by our library code, only from test code.
- `[TestingAndPrivateOnly]`: Can be called from within the same type, or from test code, but not from other types.

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/NameValueConfigurationSource.cs
andrewlock: Fixed by changing the name of the attribute

tracer/src/Datadog.Trace/SpanContext.cs
andrewlock: Fixed by changing the name of the attribute

andrewlock:
> Datadog.Trace.AppSec.WafEncoding.SetPoolSize(int poolSize)
> Datadog.Trace.Debugger.Symbols.SymbolExtractor.GetClassSymbols(string typeToExtract)
> Datadog.Trace.Iast.DefaultTaintedMap.GetListValues()
> HashBasedDeduplication.HashBasedDeduplication(DateTime? currentTime = null)
> TraceSampler.GetRules()
> RedactedErrorLogCollector.EnqueueLog(LogMessageData log)
> TelemetryFactory.CreateFactory()

Great, thanks! I've found a bunch more too elsewhere too, so will add those as well. I'll do them in a separate PR though, to unblock adding the attribute to new PRs sooner rather than later 🙂

PR #5280, opened 2024-03-06
Allow skipping span generation in `ProcessStart` integration

andrewlock:
## Summary of changes

Adds a workaround to avoid recursive initialisation of the tracer

## Reason for change

When running on a Linux host, when we can't get the ContainerId using `ContainerMetadata`, we call `GetCgroupInodeInternal` which tries to get the Inode number by calling `stat --printf=%i {path}` using `Process.Start()`.

Unfortunately, we try to do this during tracer initialization, _before_ it's complete. The Process.Start instrumentation kicks in, and we hit a recursive code path that causes the `ContainerMetadata` to throw. 

## Implementation details

When we're starting a process, add a flag to the `Environment` collection, check for the flag in the integration, and bail out if we find it

## Test coverage

Coming soon, to a PR near you... I'm busy manually repro-ing right now 😅

PR #7789, opened 2025-11-06
Add fix for persistent `Process.Start` version-conflict issue

andrewlock:
## Summary of changes

- Adds a workaround for the version-conflict issue that occurs on app startup on Linux
- Allow doing call target modification of version conflict dll
- Fix a bug in CallTarget instrumentation where we get the assembly reference wrong

## Reason for change

As part of app startup, on _some_ linux distros, we shell out to `stat` to build the container tags/entity ID. We don't want to trace this `Process.Start()` call, so in https://github.com/DataDog/dd-trace-dotnet/pull/5280 we added a flag to skip instrumenting these calls. However, this fix relies on a `[ThreadStatic]` variable, and in version-conflict scenarios (2.x.x manual, 3.x.x automatic) we end up still instrumenting this call, which causes recursion in `Tracer` initialization and [errors](https://app.datadoghq.com/error-tracking?query=service%3Ainstrumentation-telemetry-data%20source%3Adotnet%20%40tracer_version%3A3.29.0.0%20-%40error.is_crash%3Atrue&et-side=activity&order=total_count&refresh_mode=sliding&source=all&sp=%5B%7B%22p%22%3A%7B%22issueId%22%3A%22cbce5fc2-3adf-11f0-a4be-da7ad0900002%22%7D%2C%22i%22%3A%22error-tracking-issue%22%7D%5D&view=spans&from_ts=1761134402500&to_ts=1762344002500&live=true).

> Note that since https://github.com/DataDog/dd-trace-dotnet/pull/7453 we don't do a process start at all, but that doesn't help in this situation, because it's the 2.x.x library that's doing the `Process.Start()`

## Implementation details

- Use standard call target instrumentation on the 2.x.x version of `Datadog.Trace` (i.e. version conflict only)
- Hook the `ProcessHelpers.StartWithDoNotTrace()` method, and set the 3.x.x `_doNotTrace` variable for the duration of the method call
- Tweak the Rejit handler so that we _do_ rejit/call target the Datadog.Trace 2.x.x module (but not the 3.x.x module)
- Fix a bug in the module builder which was incorrectly injecting a reference to the 2.x.x assembly instead of the 3.x.x assembly

## Test coverage

We were already working around this issue in our VersionConflict tests, so I removed the workaround, confirmed that the test failed, then made the fix, and confirmed the tests pass again.

## Other details

This will only help for customers using a manual version of 2.49.0+ (when we introduced the `StartWithDoNotTrace()` call). I think that's good enough support.

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/VersionConflict/ProcessHelpersStartWithDoNotTraceIntegration.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/VersionConflict/ProcessHelpersStartWithDoNotTraceIntegration.cs
@@ -22,7 +22,7 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.VersionConflict;
     MethodName = "StartWithDoNotTrace",
     ReturnTypeName = ClrNames.Process,
     ParameterTypeNames = ["System.Diagnostics.ProcessStartInfo", ClrNames.Bool],
-    MinimumVersion = "2.49.0",
+    MinimumVersion = "2.49.0", // We only introduced ProcessHelpers.StartWithDoNotTrace() in 2.49.0
     MaximumVersion = "2.*.*",
     IntegrationName = nameof(IntegrationId.DatadogTraceVersionConflict))]
 [Browsable(false)]
--- a/tracer/src/Datadog.Trace/Util/ProcessHelpers.cs
+++ b/tracer/src/Datadog.Trace/Util/ProcessHelpers.cs
@@ -70,7 +70,7 @@ namespace Datadog.Trace.Util
         /// Used to override the "do not trace" state for the current thread's call to <see cref="Process.Start()"/>.
         /// Prefer using <see cref="StartWithDoNotTrace"/> - this should only be used to work around version conflict scenarios
         /// </summary>
-        public static bool ForceDoNotTrace(bool doNotTrace) => _doNotTrace = doNotTrace;
+        public static void ForceDoNotTrace(bool doNotTrace) => _doNotTrace = doNotTrace;
 
         /// <summary>
         /// Run a command and get the standard output content as a string
--- a/tracer/src/Datadog.Tracer.Native/metadata_builder.cpp
+++ b/tracer/src/Datadog.Tracer.Native/metadata_builder.cpp
@@ -71,7 +71,7 @@ HRESULT MetadataBuilder::FindIntegrationTypeRef(const IntegrationDefinition& int
         // Name matches, now check the version, in case we have multiple versions of the assembly loaded
         // TODO: we should already have this data, so should probably just pass that data through,
         // but it's a bit of a pain, to thread that needle right now
-        const auto current_assembly_metadata = GetAssemblyImportMetadata(assembly_import_);
+        const auto& current_assembly_metadata = GetAssemblyImportMetadata(assembly_import_);
         if (current_assembly_metadata.version == integration_definition.integration_type.assembly.version)
         {
             // type is defined in this assembly (same name AND version)


tracer/src/Datadog.Tracer.Native/metadata_builder.cpp
```diff
@@ -64,14 +64,24 @@ HRESULT MetadataBuilder::FindIntegrationTypeRef(const IntegrationDefinition& int
 
     HRESULT hr;
     type_ref = mdTypeRefNil;
+    bool is_same_assembly = false;
 
     if (metadata_.assemblyName == integration_definition.integration_type.assembly.name)
```
andrewlock, line 69: Previously we were only checking for "is this the right assembly" by checking the assembly name, but we should check the version too. This is obviously an issue in the version conflict scenario, but _theoretically_ it's an issue in other cases where two versions of the assembly are loaded (I think)

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/VersionConflict/VersionConflict2xTests.cs
```diff
@@ -18,11 +18,6 @@ public class VersionConflict2xTests : TestHelper
         public VersionConflict2xTests(ITestOutputHelper output)
             : base("VersionConflict.2x", output)
         {
-            // We disable the process integration to prevent the `stat` process spans that
-            // _may_ be present depending on the underlying host system (cgroup v1/v2).
-            // The "do not trace" helper that normally blocks tracing these spans doesn't
-            // work in version conflict scenarios.
-            SetEnvironmentVariable("DD_TRACE_Process_ENABLED", "0");
```
andrewlock, line 25: Prior to the fix, removing this caused some of the the tests to fail, because we generated process spans when we shouldn't on linux (on some distros only)

tracer/src/Datadog.Tracer.Native/cor_profiler.cpp
andrewlock, line 1: The changes here look worse than they are - if you ignore whitespace it makes more sense. The change is just to set `perform_calltarget_instrumentation_on_version_conflict_assembly` when we're in version conflict, and then to _not_ skip rewriting in that case.

tracer/src/Datadog.Tracer.Native/metadata_builder.cpp
```diff
@@ -64,14 +64,24 @@ HRESULT MetadataBuilder::FindIntegrationTypeRef(const IntegrationDefinition& int
...
     {
-        // type is defined in this assembly
-        hr = metadata_emit_->DefineTypeRefByName(module_, integration_definition.integration_type.name.c_str(),
+        // Name matches, now check the version, in case we have multiple versions of the assembly loaded
+        // TODO: we should already have this data, so should probably just pass that data through,
+        // but it's a bit of a pain, to thread that needle right now
+        const auto current_assembly_metadata = GetAssemblyImportMetadata(assembly_import_);
```
lucaspimentel: We use `const auto&` everywhere else we call `GetAssemblyImportMetadata()` to avoid making a copy.


```suggestion
        const auto& current_assembly_metadata = GetAssemblyImportMetadata(assembly_import_);
```

tracer/src/Datadog.Trace/Util/ProcessHelpers.cs
```diff
@@ -61,11 +61,17 @@ public static void GetCurrentProcessInformation(out string processName, out stri
...
+        /// Gets the "do not trace" state for the current thread's call to <see cref="Process.Start()"/>
         /// </summary>
         /// <returns>True if the <see cref="Process.Start()"/> call should be traced, False if "do not trace" is set</returns>
         public static bool ShouldTraceProcessStart() => !_doNotTrace;
 
+        /// <summary>
+        /// Used to override the "do not trace" state for the current thread's call to <see cref="Process.Start()"/>.
+        /// Prefer using <see cref="StartWithDoNotTrace"/> - this should only be used to work around version conflict scenarios
+        /// </summary>
+        public static bool ForceDoNotTrace(bool doNotTrace) => _doNotTrace = doNotTrace;
```
lucaspimentel: This just returns the value of the `doNotTrace` parameter, which seems useless. And it looks like the return value is never used, anyway?

```suggestion
        public static void ForceDoNotTrace(bool doNotTrace) => _doNotTrace = doNotTrace;
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/VersionConflict/ProcessHelpersStartWithDoNotTraceIntegration.cs
```diff
@@ -0,0 +1,54 @@
...
+/// <summary>
+/// System.Diagnostics.Process Datadog.Trace.Util.ProcessHelpers::StartWithDoNotTrace(System.Diagnostics.ProcessStartInfo,System.Boolean) calltarget instrumentation
+/// </summary>
+[InstrumentMethod(
+    AssemblyName = "Datadog.Trace",
+    TypeName = "Datadog.Trace.Util.ProcessHelpers",
+    MethodName = "StartWithDoNotTrace",
+    ReturnTypeName = ClrNames.Process,
+    ParameterTypeNames = ["System.Diagnostics.ProcessStartInfo", ClrNames.Bool],
+    MinimumVersion = "2.49.0",
```
lucaspimentel: Consider adding a comments to explain why versions `2.49.0 <= x <= 2.*.*`

From your PR description:
> This will only help for customers using a manual version of 2.49.0+ (when we introduced the `StartWithDoNotTrace()` call).

tracer/src/Datadog.Trace/Util/ProcessHelpers.cs
andrewlock: Yep, copy paste error 👍

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/VersionConflict/ProcessHelpersStartWithDoNotTraceIntegration.cs
andrewlock:
```suggestion
    MinimumVersion = "2.49.0", // We only introduced ProcessHelpers.StartWithDoNotTrace() in 2.49.0
```

tracer/src/Datadog.Trace/Util/ProcessHelpers.cs
tonyredondo: this implementation only works because the integration is not for an async method (ThreadStatic).

That made me think for a general solution backed in the tracer Scope, maybe we could add a flag in the Scope that tell the tracer if a span can be on top of that scope or not. if the flag is true then no scope can be on top of the active one, and that by definition is a do not trace feature. If the flag is false then we can behave the same as we currently do. If we do it in this way then the flag can "travel" with the async flow. This is just an idea that I just had.

andrewlock: Yeah, I was exploring something a while back, similar to Java's "black hole spans". When I tried it initally, (in the v2 era) we ran into issues with out public API, but with v3, this should become much more achievable.

The `ThreadStatic` was absolutely a hack to work around the fact we couldn't do the feature we _wanted_ to 😅 It's actually a customer-facing feature too, because customers (for example) want to be able to disable tracing for certain _specific_ calls (and they can't always add do-not-trace headers for example). It's in the backlog 😄

PR #7790, opened 2025-11-06
Fix benchmark flakiness

NachoEchevarria:
## Summary of changes

This PR addresses [benchmark flakiness detected](https://github.com/DataDog/relenv-benchmarking-service/pull/91) by eliminating static constructors and moving all initialization to `[GlobalSetup]` methods, which BenchmarkDotNet explicitly excludes from timing measurements.

Benchmarks were experiencing flakiness due to:
1. **Static constructor timing issues**: Static constructors run at unpredictable times and their overhead is included in early benchmark measurements
2. **Incorrect static field usage**: Using `static` fields with `[GlobalSetup]` creates confusion since GlobalSetup runs once per benchmark method, not once per class
3. **Native code warmup**: Some benchmarks (especially WAF-related) need some warmup to stabilize native code paths

### Why Use Instance Fields for Runtime State?

`[GlobalSetup]` runs **once per benchmark method**, not once per class. Using instance fields for runtime state (clients, services, etc.):
1. **Ensures proper isolation** - Each benchmark gets a fresh class instance
2. **Prevents state leakage** - No shared state between different benchmark methods
3. **Matches BenchmarkDotNet semantics** - Clear intent that each benchmark is isolated

### Why Use `static readonly` for Test Data?

For pre-computed immutable test data (lambdas, test objects, constant values):
1. **Better compiler optimization** - `static readonly` fields can be inlined and optimized better than instance fields
2. **Single allocation** - Test data allocated once, shared across all benchmark runs
3. **Clear intent** - Signals that the data never changes and is shared safely

## Reason for change

  Benchmark variance analysis revealed high Coefficients of Variation (CV) ranging from 5.8% to 151%, indicating unstable and unreliable performance measurements.

### DIFF
--- a/tracer/test/benchmarks/Benchmarks.Trace/GraphQLBenchmark.cs
+++ b/tracer/test/benchmarks/Benchmarks.Trace/GraphQLBenchmark.cs
@@ -13,7 +13,7 @@ namespace Benchmarks.Trace
     [BenchmarkCategory(Constants.TracerCategory)]
     public class GraphQLBenchmark
     {
-        private Task<ExecutionResult> _result;
+        private readonly static Task<ExecutionResult> _result = Task.FromResult(new ExecutionResult { Value = 42 });
         private ExecutionContext _context;
         private GraphQLClient _client;
 
@@ -24,7 +24,6 @@ namespace Benchmarks.Trace
 
             Tracer.UnsafeSetTracerInstance(new Tracer(settings, new DummyAgentWriter(), null, null, null));
 
-            _result = Task.FromResult(new ExecutionResult { Value = 42 });
             _context = new ExecutionContext();
             _client = new GraphQLClient(_result);
         }
@@ -39,7 +38,7 @@ namespace Benchmarks.Trace
 
             return task.GetAwaiter().GetResult().Value;
 
-            Task<ExecutionResult> ExecuteAsyncImpl(ExecutionContext context) => _result;
+            static Task<ExecutionResult> ExecuteAsyncImpl(ExecutionContext context) => _result;
         }
 
         private class GraphQLClient : IExecutionStrategy
--- a/tracer/test/benchmarks/Benchmarks.Trace/HttpClientBenchmark.cs
+++ b/tracer/test/benchmarks/Benchmarks.Trace/HttpClientBenchmark.cs
@@ -15,7 +15,7 @@ namespace Benchmarks.Trace
     public class HttpClientBenchmark
     {
         private HttpRequestMessage _httpRequest;
-        private Task<HttpResponseMessage> _cachedResult;
+        private static readonly Task<HttpResponseMessage> _cachedResult = Task.FromResult(new HttpResponseMessage());
 
         [GlobalSetup]
         public void GlobalSetup()
@@ -25,7 +25,6 @@ namespace Benchmarks.Trace
             Tracer.UnsafeSetTracerInstance(new Tracer(settings, new DummyAgentWriter(), null, null, null));
 
             _httpRequest = new HttpRequestMessage { RequestUri = new Uri("http://datadoghq.com") };
-            _cachedResult = Task.FromResult(new HttpResponseMessage());
         }
 
         [Benchmark]
@@ -35,7 +34,7 @@ namespace Benchmarks.Trace
                 (this, _httpRequest, CancellationToken.None, &GetResult).GetAwaiter().GetResult();
             return "OK";
 
-            Task<HttpResponseMessage> GetResult(HttpRequestMessage request, CancellationToken cancellationToken) => _cachedResult;
+            static Task<HttpResponseMessage> GetResult(HttpRequestMessage request, CancellationToken cancellationToken) => _cachedResult;
         }
     }
 }
--- a/tracer/test/benchmarks/Benchmarks.Trace/RedisBenchmark.cs
+++ b/tracer/test/benchmarks/Benchmarks.Trace/RedisBenchmark.cs
@@ -40,7 +40,7 @@ namespace Benchmarks.Trace
             return CallTarget.Run<RedisNativeClientSendReceiveIntegration, RedisNativeClient, byte[][], Func<int>, Action<Func<int>>, bool, int>
                 (_client, _rawCommands, _fn, _completePipelineFn, true, &SendReceiveImpl);
 
-            int SendReceiveImpl(byte[][] cmdWithBinaryArgs, Func<int> fn, Action<Func<int>> completePipelineFn, bool sendWithoutRead) => fn();
+            static int SendReceiveImpl(byte[][] cmdWithBinaryArgs, Func<int> fn, Action<Func<int>> completePipelineFn, bool sendWithoutRead) => fn();
         }
     }
 }
--- a/tracer/test/benchmarks/Benchmarks.Trace/TraceAnnotationsBenchmark.cs
+++ b/tracer/test/benchmarks/Benchmarks.Trace/TraceAnnotationsBenchmark.cs
@@ -28,9 +28,6 @@ namespace Benchmarks.Trace
             var settings = TracerSettings.Create(new() { { ConfigurationKeys.StartupDiagnosticLogEnabled, false } });
 
             Tracer.UnsafeSetTracerInstance(new Tracer(settings, new DummyAgentWriter(), null, null, null));
-
-            // Warmup
-            RunOnMethodBegin();
         }
 
         [Benchmark]


tracer/test/benchmarks/Benchmarks.Trace/README_GUIDELINES.md
```diff
@@ -0,0 +1,109 @@
+# Benchmark Guidelines
```
NachoEchevarria, line 1: Guide generated with Claude code.

igoragoli:
Thanks for working on this, @NachoEchevarria!

Seeing [errors](https://gitlab.ddbuild.io/DataDog/apm-reliability/dd-trace-dotnet/-/jobs/1220076873#L2560) like these on the CI:
```
 [ERR] RunBenchmarks: C:\dd-trace-dotnet\tracer\test\benchmarks\Benchmarks.Trace\bin\Release\net6.0\2b73268d-661e-4034-861a-8c6b24014cbc\2b73268d-661e-4034-861a-8c6b24014cbc.notcs(592,33): error CS0103: The name 'GlobalSetup' does not exist in the current context [C:\dd-trace-dotnet\tracer\test\benchmarks\Benchmarks.Trace\bin\Release\net6.0\2b73268d-661e-4034-861a-8c6b24014cbc\BenchmarkDotNet.Autogenerated.csproj]
```

The benchmarks completes successfully, though.

Because of a rate limiting issue, the script failed to run the PR commenter[^1] and failed to upload results to the BP UI (which was done after the PR comment). I'm rue-running the benchmarks so we can get results on the BP UI and I can run the `/flakiness` endpoint.

That said, looks good to me, excited to see the results.

[^1]: I took note of that for our team, this shouldn't happen.

NachoEchevarria:
> Thanks for working on this, @NachoEchevarria!
> 
> Seeing [errors](https://gitlab.ddbuild.io/DataDog/apm-reliability/dd-trace-dotnet/-/jobs/1220076873#L2560) like these on the CI:
> 
> ```
>  [ERR] RunBenchmarks: C:\dd-trace-dotnet\tracer\test\benchmarks\Benchmarks.Trace\bin\Release\net6.0\2b73268d-661e-4034-861a-8c6b24014cbc\2b73268d-661e-4034-861a-8c6b24014cbc.notcs(592,33): error CS0103: The name 'GlobalSetup' does not exist in the current context [C:\dd-trace-dotnet\tracer\test\benchmarks\Benchmarks.Trace\bin\Release\net6.0\2b73268d-661e-4034-861a-8c6b24014cbc\BenchmarkDotNet.Autogenerated.csproj]
> ```
> 
> The benchmarks completes successfully, though.
> 
> Because of a rate limiting issue, the script failed to run the PR commenter[1](#user-content-fn-1-19706ce7ccdbeaeff7da97da74e5f53e) and failed to upload results to the BP UI (which was done after the PR comment). I'm rue-running the benchmarks so we can get results on the BP UI and I can run the `/flakiness` endpoint.
> 
> That said, looks good to me, excited to see the results.
> 
> ## Footnotes
> 1. I took note of that for our team, this shouldn't happen. [↩](#user-content-fnref-1-19706ce7ccdbeaeff7da97da74e5f53e)

Thanks! I will check the cause of the errors.

igoragoli:
I ran the flakiness check again for 0f5d7f6, looks like there were some big improvements!

E.g., `Benchmarks.Trace.CIVisibilityProtocolWriterBenchmark.WriteAndFlushEnrichedTraces net6.0` went from 0.478733 to 0.004474.

Full check: https://gist.github.com/igoragoli/806c821ba255e1a8fcad6538307ae323

igoragoli:
I simplified the `/flakiness` endpoint, so tests are slightly different.

Full check comparing this branch at 82bfad against master at 8603e6d: https://docs.google.com/spreadsheets/d/1LKKYSYefti9l4AVA1imZSssPes0aDuqkSUD7zfxlONw/edit?usp=sharing

tracer/test/benchmarks/Benchmarks.Trace/ActivityBenchmark.cs
```diff
@@ -40,6 +41,9 @@ static ActivityBenchmark()
         ActivitySource.AddActivityListener(activityListener);
 
         _duckSource = new Datadog.Trace.Activity.DuckTypes.ActivitySource { Name = _source.Name, Version = _source.Version ?? string.Empty };
+
+        // Warmup
+        StartStopWithChild();
```
andrewlock: nit: I don't think this is really necessary. By design, Benchmarkdotnet does a _lot_ of warmup runs before it starts recording. It doesn't hurt, but it does mildly incovenience adding a new benchmark, meh 🤷‍♂️

tracer/test/benchmarks/Benchmarks.Trace/Asm/AppSecEncodeBenchmark.cs
```diff
@@ -21,17 +21,22 @@ namespace Benchmarks.Trace.Asm;
...
 public class AppSecEncoderBenchmark
 {
-    private static readonly Encoder _encoder;
-    private static readonly EncoderLegacy _encoderLegacy;
-    private static readonly NestedMap _args;
+    private static readonly NestedMap _args = MakeNestedMap(20);
 
-    static AppSecEncoderBenchmark()
+    private Encoder _encoder;
+    private EncoderLegacy _encoderLegacy;
```
andrewlock: Out of interest, why make some of these non-static and keep others as static? 🤔 If you keep the static ones, isn't that the same issue you're trying to remove by getting rid of the static constructor?

Put another way, having a static initializer (like you have for `_args`) is the same as having a static initializer like you had before, so probably we you should either move _everything_ to non-statics, or not bother changing _anything_? 😄

tracer/test/benchmarks/Benchmarks.Trace/Asm/AppSecWafBenchmark.cs
```diff
@@ -52,7 +53,25 @@ static AppSecWafBenchmark()
         {
             throw new ArgumentException($"Waf could not initialize, error message is: {initResult.ErrorMessage}");
         }
-        Waf = initResult.Waf;
+        _waf = initResult.Waf;
+
+        // More aggressive warmup for native code paths (WAF library)
+        // Ensures JIT compilation completes and native context creation stabilizes
```
andrewlock: Benchmarkdotnet does all this for every benchmark as standard. In fact, it does it more carefully 😄

tracer/test/benchmarks/Benchmarks.Trace/README_GUIDELINES.md
```diff
@@ -0,0 +1,310 @@
...
+        _client = new MyClient();
+    }
+}
+```
+
+**✅ Do this:**
+```csharp
+public class MyBenchmark
+{
+    private static readonly Func<int> SimpleFunc = () => 42;  // OK: Static readonly for immutable data
```
andrewlock: I'm not sure on what basis we think this ok 🤷‍♂️ 😄

tracer/test/benchmarks/Benchmarks.Trace/README_GUIDELINES.md
```diff
@@ -0,0 +1,310 @@
...
+```
+
+**Why?**
+- `[GlobalSetup]` runs **once per benchmark method**, not once per class
+- Using instance fields ensures proper isolation between benchmark methods
+- Static constructors run at CLR-controlled times, potentially during measurement
+- BenchmarkDotNet explicitly excludes `[GlobalSetup]` from timing
+- See: https://github.com/dotnet/BenchmarkDotNet/issues/1304
+
+### 2. When to Use `static readonly` vs Instance Fields
```
andrewlock: I still think that _if_ static constructors are a problem, then _all_ static initializers should be a problem?

tracer/test/benchmarks/Benchmarks.Trace/Asm/AppSecEncodeBenchmark.cs
NachoEchevarria: The issue is not static fields per se, but static constructors for runtime initialization (not controlled by the benchmarking library, could run during measurement, etc).
I'm using static readonly for immutable data or mocks (like _args here) that will not cause conflicts between instances. They run before the type is used so they are predictable and can benefit from compiler optimizations.

tracer/test/benchmarks/Benchmarks.Trace/Asm/AppSecWafBenchmark.cs
NachoEchevarria: Yes, probably it is worth to leave warmup to the library. I will update the guide as well. Thanks!

tracer/test/benchmarks/Benchmarks.Trace/README_GUIDELINES.md
NachoEchevarria: After discussing, the approach would be to use instance fields by default and initialize them in GlobalSetup methods. I have updated the code and doc. Thanks!

tracer/test/benchmarks/Benchmarks.Trace/Asm/AppSecEncodeBenchmark.cs
NachoEchevarria: After discussing, the approach would be to use instance fields by default and initialize them in GlobalSetup methods. I have updated the code and doc. Thanks!

tracer/test/benchmarks/Benchmarks.Trace/ActivityBenchmark.cs
NachoEchevarria: Manual warmup codes have been removed. Thanks!

PR #7791, opened 2025-11-06
Add process tags to client-side stats

vandonr:
## Summary of changes

adding process tags at the ClientStatsPayload level.
See [proto](https://github.com/DataDog/datadog-agent/blob/main/pkg/proto/datadog/trace/stats.proto#L52-L55) in the agent

## Reason for change

AIDM-195

## Test coverage

piggybacked on existing tests

## Other details

I realized that dotnet doesn't write the service name at the top level, but only in the buckets. Doesn't seem to be a big issue, but there could be a win here by writing the service in the buckets only when there is an override.

### DIFF
--- a/tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporterConfiguration.cs
+++ b/tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporterConfiguration.cs
@@ -119,17 +119,6 @@ internal class TraceExporterConfiguration : SafeHandle
         }
     }
 
-    public string? ProcessTags
-    {
-        init
-        {
-            // TODO: uncomment when this is implemented in libdatadog
-            // using var processTags = new CharSlice(value);
-            // using var error = NativeInterop.Config.SetProcessTags(this, processTags);
-            // error.ThrowIfError();
-        }
-    }
-
     public TelemetryClientConfiguration? TelemetryClientConfiguration
     {
         init


vandonr:
@codex review

tracer/src/Datadog.Trace/Agent/StatsBuffer.cs
```diff
@@ -57,9 +57,15 @@ public void Reset()
             Start = DateTimeOffset.UtcNow.ToUnixTimeNanoseconds();
         }
 
-        public void Serialize(Stream stream, long bucketDuration)
+        public void Serialize(Stream stream, long bucketDuration, bool propagateProcessTags)
         {
-            MessagePackBinary.WriteMapHeader(stream, 8);
+            var count = 8;
+            if (propagateProcessTags && !string.IsNullOrEmpty(ProcessTags.SerializedTags))
```
andrewlock: nit: static access ca be relatively slow, particularly in .NET Framework, I would cache ProcessTags.SerializedTags to a variable instead of doing 3 accesses

tracer/src/Datadog.Trace/TracerManagerFactory.cs
```diff
@@ -441,7 +441,7 @@ internal static IApi GetApi(TracerSettings settings, IDogStatsd statsd, Action<D
                 }
             }
 
-            return new Api(apiRequestFactory, statsd, updateSampleRates, settings.PartialFlushEnabled);
+            return new Api(apiRequestFactory, statsd, updateSampleRates, settings.PartialFlushEnabled, settings.PropagateProcessTags);
```
andrewlock: You need to add support for this to the data-pipeline trace exporter above, too

tracer/test/Datadog.Trace.Tests/Agent/ApiTests.cs
```diff
@@ -38,7 +38,7 @@ public async Task SendTraceAsync_200OK_AllGood()
             factoryMock.Setup(x => x.GetEndpoint(It.Is<string>(s => s == TracesPath))).Returns(new Uri("http://localhost/traces"));
             factoryMock.Setup(x => x.GetEndpoint(It.Is<string>(s => s == StatsPath))).Returns(new Uri("http://localhost/stats"));
 
-            var api = new Api(apiRequestFactory: factoryMock.Object, statsd: null, updateSampleRates: null, partialFlushEnabled: false);
```
andrewlock, line 41: This is the only one you set `propagateProcessTags: true` for - is that intentional?

tracer/test/Datadog.Trace.Tests/Agent/StatsBufferTests.cs
```diff
@@ -46,12 +46,14 @@ public void Serialization()
...
             var stream = new MemoryStream();
-            buffer.Serialize(stream, expectedDuration);
+            buffer.Serialize(stream, expectedDuration, propagateProcessTags: true);
             var result = MessagePackSerializer.Deserialize<MockClientStatsPayload>(stream.ToArray());
 
             result.Hostname.Should().Be(payload.HostName);
             result.Env.Should().Be(payload.Environment);
             result.Version.Should().Be(payload.Version);
+            result.ProcessTags.Should().NotBeEmpty();
+            result.ProcessTags.Should().Contain(":");
```
andrewlock: Can we be any more specific?

You know _how_ you could be more specific? Inject the value instead of implicitly using global state :trollface: 

In all seriousness, you're _already_ injecting a boolean everywhere (enable/disable) - why not just inject the string to use instead? 🤷‍♂️ it's testable, faster (can store in readonly field, no need for multiple checks) 🤷‍♂️

tracer/test/benchmarks/Benchmarks.Trace/AgentWriterBenchmark.cs
```diff
@@ -32,7 +32,7 @@ static AgentWriterBenchmark()
             var sources = new CompositeConfigurationSource(new[] { overrides, GlobalConfigurationSource.Instance });
             var settings = new TracerSettings(sources);
 
-            var api = new Api(new FakeApiRequestFactory(settings.Exporter.AgentUri), statsd: null, updateSampleRates: null, partialFlushEnabled: false);
+            var api = new Api(new FakeApiRequestFactory(settings.Exporter.AgentUri), statsd: null, updateSampleRates: null, partialFlushEnabled: false, propagateProcessTags: true);
```
andrewlock: This will probably skew the benchmarks a little - it will be good to see the results, but we should be aware that we _expect_ them to be impacted. Also this is another case where injecting would be preferable, because then you're not measuring the processtags initialization as part of this benchmark

tracer/src/Datadog.Trace/Agent/StatsBuffer.cs
vandonr: TIL

tracer/src/Datadog.Trace/TracerManagerFactory.cs
vandonr: 🤔 TraceExporter does not do any serialization since the `SendStats` is a noop, so i don't need the setting there ?

tracer/test/Datadog.Trace.Tests/Agent/ApiTests.cs
vandonr: yes because this is the `200OK_AllGood` test, so I want things to still be 200 OK with the process tags in.
The other tests are about error handling, so I think it doesn't have much value to have process tags in there, since the assertions are on other things. But I could also set it to true everywhere on the basis that it shouldn't change anything to existing tests.

tracer/test/benchmarks/Benchmarks.Trace/AgentWriterBenchmark.cs
vandonr: yes true, it'll have an impact, I might have mentioned it in a clearer way. But since the end goal is to default to true, I think it makes sense to include it in the benchmark.

tracer/test/Datadog.Trace.Tests/Agent/StatsBufferTests.cs
vandonr: hmm, true, though I'd still need to get the static value somewhere, preferably not too far so that I don't have to inject it in many calls, but I could inject a bit further to make testing easier.

tracer/src/Datadog.Trace/TracerManagerFactory.cs
andrewlock: > 🤔 TraceExporter does not do any serialization since the SendStats is a noop, so i don't need the setting there ?

The trace exporter does serialization in libdatadog code, so you'll need to pass the process tags to the native component (created above) so that it can add them as necessary

vandonr: ah well that's interesting yes. I'll see how libdatadog need to support this.

tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporterConfiguration.cs
```diff
@@ -119,6 +119,17 @@ public string? Service
         }
     }
 
+    public string? ProcessTags
+    {
+        init
+        {
+            using var processTags = new CharSlice(value);
```
lucaspimentel: Is this line doing anything or can we comment it out as well?

tracer/src/Datadog.Trace/Agent/Api.cs
```diff
@@ -48,6 +49,7 @@ public Api(
             IDogStatsd statsd,
             Action<Dictionary<string, float>> updateSampleRates,
             bool partialFlushEnabled,
+            string processTags,
```
lucaspimentel: Adding process tags as `Api` state only so it can be added to client-side stats doesn't feel right.

Would it makes more sense to pass the process tags string into `SendStatsAsync()`, add it to the new `SendStatsState` there, which then passes it to `SendStatsAsyncImpl()` where you are using it?

tracer/src/Datadog.Trace/Agent/StatsBuffer.cs
```diff
@@ -57,9 +57,15 @@ public void Reset()
             Start = DateTimeOffset.UtcNow.ToUnixTimeNanoseconds();
         }
 
-        public void Serialize(Stream stream, long bucketDuration)
+        public void Serialize(Stream stream, long bucketDuration, string processTags)
```
lucaspimentel: This doesn't feel like the right place to add a `processTags` parameter. Would it be possible to add process tags to `ClientStatsPayload` and access it from the `_header` field like we do with `Hostname`, `Env`, and `Version`?

tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporterConfiguration.cs
vandonr: yes true we could comment this too

andrewlock: I would remove this entirely - to the casual observer (in TracerManagerFactory it _looks_ like it's implemented.

tracer/src/Datadog.Trace/Agent/Api.cs
andrewlock: I would have said it makes most sense being injected into `StatsBuffer` directly, likely by way of `ClientStatsPayload` 🤔 That way you don't have to thread it through everywhere, and it lives with other "stable" things like `HostName` etc?

tracer/src/Datadog.Trace/Agent/StatsBuffer.cs
vandonr: yeah that looks better

tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporterConfiguration.cs
vandonr: what do you mean, remove the property so that it's not misleading when reading TracerManagerFactory ?

andrewlock: Exactly 👍

tracer/src/Datadog.Trace/LibDatadog/DataPipeline/TraceExporterConfiguration.cs
```diff
@@ -119,6 +119,17 @@ public string? Service
...
+    {
+        init
+        {
+            // TODO: uncomment when this is implemented in libdatadog
+            // using var processTags = new CharSlice(value);
+            // using var error = NativeInterop.Config.SetProcessTags(this, processTags);
+            // error.ThrowIfError();
+        }
+    }
+
```
andrewlock: As discussed: 

```suggestion
```

tracer/src/Datadog.Trace/TracerManagerFactory.cs
```diff
@@ -423,6 +423,7 @@ internal static IApi GetApi(TracerSettings settings, IDogStatsd statsd, Action<D
                         Env = settings.Environment,
                         Version = settings.ServiceVersion,
                         Service = settings.MutableSettings.DefaultServiceName,
+                        ProcessTags = settings.PropagateProcessTags ? ProcessTags.SerializedTags : null,
```
andrewlock: If libdatadog doesn't support it, I think it's just better we don't do this for now (more chance we forget to update the implementation)
```suggestion
```

tracer/src/Datadog.Trace/Agent/Api.cs
lucaspimentel: or that, yeah 😅

PR #7796, opened 2025-11-10
Add fix for not re-reporting telemetry when there are "empty" dynamic config changes

andrewlock:
## Summary of changes

A fix for #7724 to handle telemetry reporting in dynamic config "reset" scenarios

## Reason for change

The system tests for #7724 were failing in some dynamic configuration scenarios. Specifically, the tests were sending remote config _without_ any configuration values "i.e. 'reset to use defaults'" and were waiting a telemetry update. However, we never sent it, because there was "no telemetry to record". 

Note that we _did_ correctly apply the new configuration, we just didn't report the telemetry correctly, primarily due to limitations in the telemetry protocol. This PR adds a fix for that, and will be merged into #7724.

## Implementation details

The solution is to "remember" the telemetry from the default mutable configuration values, _without_ any dynamic sources, and "replay" this telemetry when we update telemetry. This feels kind of hacky, but it's something I suspected we might need to do, and had been avoiding up to this point because we do a "full reconfigure" anyway.

## Test coverage

Added a specific unit test that mimics the behaviour of the system-test (i.e. an "empty" dynamic config response) and confirms the telemetry is recorded as expected

## Other details

Part of a config stack

- https://github.com/DataDog/dd-trace-dotnet/pull/7796 👈

Unlike other PRs in the stack, I'll merge this directly into #7724 to fix the tests there, just thought I'd keep this separate for easier reviewing

### DIFF
--- a/tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
+++ b/tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
@@ -125,8 +125,6 @@ namespace Datadog.Trace.Configuration
                                        .AsBoolResult()
                                        .OverrideWith(in otelActivityListenerEnabled, ErrorLog, defaultValue: false);
 
-            var exporter = new ExporterSettings(source, _telemetry);
-
             PeerServiceTagsEnabled = config
                .WithKeys(ConfigurationKeys.PeerServiceDefaultsEnabled)
                .AsBool(defaultValue: false);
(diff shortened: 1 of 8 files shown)


tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -335,66 +333,6 @@ not null when string.Equals(value, "otlp", StringComparison.OrdinalIgnoreCase) =
 
             OpenTelemetryLogsEnabled = OpenTelemetryLogsEnabled && OtelLogsExporterEnabled;
 
-            DataPipelineEnabled = config
```
andrewlock, line 338: This has all just moved to the end of the constructor, cut and paste

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -731,9 +669,70 @@ not null when string.Equals(value, "otlp", StringComparison.OrdinalIgnoreCase) =
...
+                    _telemetry.Record(ConfigurationKeys.TraceDataPipelineEnabled, false, ConfigurationOrigins.Calculated);
+                }
+
+                // Windows supports UnixDomainSocket https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/
+                // but tokio hasn't added support for it yet https://github.com/tokio-rs/tokio/issues/2201
+                // There's an issue here, in that technically a user can initially be configured to send over TCP/named pipes,
+                // and so we allow and enable the datapipeline. Later, they could configure the app in code to send over UDS.
+                // This is a problem, as we currently don't support toggling the data pipeline at runtime, so we explicitly block
+                // this scenario in the public API.
+                if (Manager.InitialExporterSettings.TracesTransport == TracesTransportType.UnixDomainSocket && FrameworkDescription.Instance.IsWindows())
```
andrewlock, line 699: This is the only change, we read the exporter settings that the `Manager` created, instead of creating the `new ExporterSettings()` and passing them _to_ the `Manager`. It's the same thing overall, but it means the `Manager` can record the initial telemetry for the exporter settings (so it can be replayed again later)

tracer/src/Datadog.Trace/RemoteConfigurationManagement/Transport/RemoteConfigurationApi.cs
```diff
@@ -6,6 +6,7 @@
 #nullable enable
 
 using System;
+using System.IO;
```
NachoEchevarria, line 9: NIT: Probably not needed

PR #7801, opened 2025-11-11
Set timeouts and improved logging in dotnet-tool Nuget smoke tests.

NachoEchevarria:
## Summary of changes

 Summary of changes

  - Add timeout to HttpClient in Worker.cs to prevent indefinite hangs
  - Add CancellationToken support to ReadAsStringAsync for .NET 5.0+
  - Add 2-minute timeout for application startup to prevent indefinite waiting
  
## Reason for change

 The dotnet-tool-nuget-smoke-tests CI job has been experiencing[ flaky timeouts](https://dev.azure.com/datadoghq/a51c4863-3eb4-4c5d-878a-58b41a049e4e/_apis/build/builds/190751/logs/21713), where the test hangs for the full
  15-minute timeout period before failing:

```
  2025-11-10T14:35:33.9771350Z Running: dotnet /app/AspNetCoreSmokeTest.dll
  2025-11-10T14:50:42.1702536Z ##[error]The task has timed out.
```

  The application starts but never completes, causing the entire CI pipeline to hang.

## Implementation details

  Previously, the HttpClient had no timeout configured, which could cause indefinite hangs when making the
  self-request to /api/values. Added a 30-second timeout:

  Added cancellation token support to ReadAsStringAsync() for .NET 5.0+, allowing the operation to be cancelled
  properly:

  Added a 2-minute timeout when waiting for the ApplicationStarted event. If the application fails to start (due to
  tracer/profiler initialization issues), the worker now:
  - Logs an error with diagnostic information
  - Sets the exit code to 1
  - Stops the application cleanly
  - Prevents hanging for the full 15-minute CI timeout

### DIFF
--- a/tracer/test/test-applications/regression/AspNetCoreSmokeTest/Worker.cs
+++ b/tracer/test/test-applications/regression/AspNetCoreSmokeTest/Worker.cs
@@ -34,8 +34,19 @@ namespace AspNetCoreSmokeTest
 
         protected override async Task ExecuteAsync(CancellationToken stoppingToken)
         {
+            var timeout = TimeSpan.FromMinutes(2);
+            var startTime = DateTimeOffset.UtcNow;
+
             while (!_appListening && !stoppingToken.IsCancellationRequested)
             {
+                if (DateTimeOffset.UtcNow - startTime > timeout)
+                {
+                    _logger.LogError("Timed out waiting for application to start after {Timeout}", timeout);
+                    Program.ExitCode = 1;
+                    _lifetime.StopApplication();
+                    return;
+                }
+
                 _logger.LogInformation("Waiting for app started handling requests");
                 await Task.Delay(100, stoppingToken);
             }


andrewlock (commented):
Hmmm, I understand the thinking here, but AFAICT this can only _increase_ flakiness, not reduce it? It will reduce the amount of time we "hang" for, but it risks increasing flake?

My big question is _why_ are we hanging, because it seems to be after the app has started, after we have attached (we have logs).

So either it's something to do with how the CI is trying to read from stdout (possible that it's blocking) or it's something on our side which is causing a hang (possible, and the scariest option)

NachoEchevarria:
> Hmmm, I understand the thinking here, but AFAICT this can only _increase_ flakiness, not reduce it? It will reduce the amount of time we "hang" for, but it risks increasing flake?
> 
> My big question is _why_ are we hanging, because it seems to be after the app has started, after we have attached (we have logs).
> 
> So either it's something to do with how the CI is trying to read from stdout (possible that it's blocking) or it's something on our side which is causing a hang (possible, and the scariest option)

This PR does not solve the flakiness but it would give us (luckilly) more information about the real issue by launching exceptions and provinding log messages, so we can work on it. The timeouts should be enough to avoid causing new flakiness by themselves.

andrewlock:
> > Hmmm, I understand the thinking here, but AFAICT this can only _increase_ flakiness, not reduce it? It will reduce the amount of time we "hang" for, but it risks increasing flake?
> > My big question is _why_ are we hanging, because it seems to be after the app has started, after we have attached (we have logs).
> > So either it's something to do with how the CI is trying to read from stdout (possible that it's blocking) or it's something on our side which is causing a hang (possible, and the scariest option)
> 
> This PR does not solve the flakiness but it would give us (luckilly) more information about the real issue by launching exceptions and provinding log messages, so we can work on it. The timeouts should be enough to avoid causing new flakiness by themselves.

I actually don't think it will...

Currently we get no logs on those tasks in CI. But that means we get no logs 😅 The app _always_ writes logs on startup, and as we don't have those, we won't have these either 🤷

PR #7805, opened 2025-11-12
[DO NOT MERGE] Build OCI with workload selection

dmehala:


shared/src/Datadog.Trace.ClrProfiler.Native/workload_selection_impl.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+    // We enable or disable instrumentation depending on the context:
+    //   - Windows IIS: instrumentation is ENABLED by default, only need to consider DENY rules.
+    //   - .NET: instrumentation is DISABLED by default, only consider ALLOW rules.
+    if (is_iis)
+    {
+        return injection_status == InjectionStatus::DENY ? false : true;
+    }
+
+    return injection_status == InjectionStatus::ALLOW;
```
dmehala:
```suggestion
    return injection_status == InjectionStatus::ALLOW ? true : false;
```

PR #7806, opened 2025-11-12
[tracing] Create and reparent span for Azure API Management

jcstorms1:
## Summary of changes
- New logic to identify Azure API Management headers
- Correctly creates an APIM span and reparents it as the parent for Isolated HTTP Azure Functions
- Refactored existing AWS proxy logic and tests to account for the addition of the Azure Proxy.
- Changes inferred proxy `span.kind` to `server` instead of `internal` to follow [this spec](https://docs.google.com/document/d/1lH9RCZLR6SyJmPAIRYBpv4sKRy54kUAWOMPuOAOpS2Q/edit?tab=t.0#heading=h.3vz8tvt4qzsx)

Note: This PR does not include functions using ASP.NET Core and will be added later

image

## Reason for change
This is a frequently requested feature for proper distributed tracing with Azure Functions using APIM

[APMSVLS-245](https://datadoghq.atlassian.net/browse/APMSVLS-245)

## Other details

[APMSVLS-245]: https://datadoghq.atlassian.net/browse/APMSVLS-245?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
@@ -293,12 +293,11 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Azure.Functions
                         fullName: functionContext.FunctionDefinition.EntryPoint,
                         bindingSource: rootSpan.Tags is AzureFunctionsTags t ? t.BindingSource : null,
                         triggerType: triggerType);
-                    // change root span's type to "serverless"
-                    scope.Root.Span.Type = SpanType;
-                    scope.Span.ResourceName = $"{triggerType} {functionName}";
-                    scope.Span.Type = SpanType;
                 }
 
+                scope.Root.Span.Type = SpanType;
+                scope.Span.ResourceName = $"{triggerType} {functionName}";
+                scope.Span.Type = SpanType;
                 tracer.TracerManager.Telemetry.IntegrationGeneratedSpan(IntegrationId);
             }
             catch (Exception ex)


tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj
```diff
@@ -31,7 +31,7 @@
     <ProjectReference Include="$(MSBuildThisFileDirectory)..\Datadog.Trace.Annotations\Datadog.Trace.Annotations.csproj" />
 
     <!-- Use PrivateAssets="none" to stop excluding content files by default -->
-    <PackageReference Include="Datadog.Serverless.Compat" Version="1.1.0" PrivateAssets="none" />
+    <PackageReference Include="Datadog.Serverless.Compat" Version="1.2.0" PrivateAssets="none" />
```
lucaspimentel, line 33: (reminder to revert this change later)

tracer/src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
```diff
@@ -654,7 +654,11 @@ private int WriteTags(ref byte[] bytes, int offset, in SpanModel model, ITagProc
 
             // AAS tags need to be set on any span for the backend to properly handle the billing.
             // That said, it's more intuitive to find it on the local root for the customer.
-            if (model.TraceChunk.IsRunningInAzureAppService && model.TraceChunk.AzureAppServiceSettings is { } azureAppServiceSettings)
+            // Skip adding AAS tags to inferred proxy spans as they represent infrastructure outside the AAS environment
+            var isInferredProxySpan = span.Tags.GetMetric(Metrics.InferredSpan) == 1.0;
```
lucaspimentel: (note to self to check for specialized tag type and access via property instead of `GetMetric()`)

tracer/src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
```diff
@@ -654,7 +654,11 @@ private int WriteTags(ref byte[] bytes, int offset, in SpanModel model, ITagProc
...
+            if (model.TraceChunk.IsRunningInAzureAppService &&
+                model.TraceChunk.AzureAppServiceSettings is { } azureAppServiceSettings &&
+                !isInferredProxySpan)
```
lucaspimentel: Currently we compute `isInferredProxySpan` always. Instead, we could do this check only if the other conditions are true so it doesn't add any overhead to apps not running in AAS. Something like:


```suggestion
            if (model.TraceChunk.IsRunningInAzureAppService &&
                model.TraceChunk.AzureAppServiceSettings is { } azureAppServiceSettings &&
                span.Tags.GetMetric(Metrics.InferredSpan) != 1.0)
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/Isolated/HttpHeadersCollection.cs
```diff
@@ -52,5 +53,18 @@ public void Remove(string name)
     {
         _headers.Remove(name);
     }
+
+    public override string ToString()
```
lucaspimentel: Is this only used for debugging? Might want to add a comment to explain that (since I would not consider this production-ready code 😅 ).

tracer/src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
lucaspimentel: Somewhere else you check if the span's operation name starts with `azure.` (for `azure.apim` I think). Should we try to be consistent with these checks, or are they fundamentally different?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -116,6 +118,10 @@ public static CallTargetState OnFunctionExecutionBegin<TTarget, TFunction>(TTarg
 
                 var functionName = instanceParam.FunctionDescriptor.ShortName;
 
+                // Check if there's an inferred proxy span (e.g., azure.apim) that we shouldn't overwrite
+                var isProxySpan = tracer.InternalActiveScope?.Root.Span.OperationName?.StartsWith("azure.", StringComparison.OrdinalIgnoreCase) == true ||
```
lucaspimentel: Should we check for specific operation names like `azure.apim` to avoid issues if there are other spans with operation names like `azure.*` or `aws.*` in the future?

Also, can we get away with a case-sensitive check (since we set the names ourselves and they are defined constants)?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -116,6 +118,10 @@ public static CallTargetState OnFunctionExecutionBegin<TTarget, TFunction>(TTarg
...
+                                  tracer.InternalActiveScope?.Root.Span.OperationName?.StartsWith("aws.", StringComparison.OrdinalIgnoreCase) == true;
```
lucaspimentel: You can get `tracer.InternalActiveScope?.Root.Span.OperationName` once here instead of twice. A little less work and easier to read.

```suggestion
                var rootName = tracer.InternalActiveScope?.Root.Span.OperationName;

                var isProxySpan = rootName?.StartsWith("azure.", StringComparison.OrdinalIgnoreCase) == true ||
                                  rootName?.StartsWith("aws.", StringComparison.OrdinalIgnoreCase) == true;
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -252,35 +286,45 @@ _ when type.StartsWith("eventGrid", StringComparison.OrdinalIgnoreCase) => "Even
...
                 var functionName = context.FunctionDefinition.Name;
 
-                var tags = new AzureFunctionsTags
-                {
-                    TriggerType = triggerType,
-                    ShortName = functionName,
-                    FullName = context.FunctionDefinition.EntryPoint,
-                };
+                // Check if there's an APIM proxy span that we shouldn't overwrite
+                var isProxySpan = tracer.InternalActiveScope?.Root.Span.OperationName?.StartsWith("azure.", StringComparison.OrdinalIgnoreCase) == true;
```
lucaspimentel: same comment as above about checking for exact operation name

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -454,6 +484,26 @@ private static bool AreAllSpanContextsIdentical(List<PropagationContext> context
...
+            where TFeature : struct
+        {
+            if (context.Features == null)
+            {
+                return null;
+            }
+
+            foreach (var kvp in context.Features)
+            {
+                if (kvp.Key.FullName?.Equals(featureTypeName) == true)
```
lucaspimentel: This is easier to read and does the same thing 😅 

```suggestion
                if (kvp.Key.FullName == featureTypeName)
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -454,6 +484,26 @@ private static bool AreAllSpanContextsIdentical(List<PropagationContext> context
...
+                {
+                    return kvp.Value?.TryDuckCast<TFeature>(out var feature) == true ? feature : null;
```
lucaspimentel, line 474: I _think_ we have a method that does this check for you? Something like:
```csharp
return kvp.Value?.DuckAs<TFeature>();
```

It's meant to be analogous to using `as` vs a cast a in C#.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/AwsApiGatewayExtractor.cs
```diff
@@ -22,26 +22,13 @@ internal sealed class AwsApiGatewayExtractor : IInferredProxyExtractor
...
-                return false;
-            }
-
-            // we need to first validate whether or not we have the header and whether or not it matches aws-apigateway
-            var proxyName = ParseUtility.ParseString(carrier, carrierGetter, InferredProxyHeaders.Name);
-            if (proxyName is null || !string.Equals(proxyName, ProxyName, StringComparison.OrdinalIgnoreCase))
-            {
-                Log.Debug("Invalid or missing {HeaderName} header with value {Value}", InferredProxyHeaders.Name, proxyName);
-                return false;
-            }
```
lucaspimentel, line 43: Did we move these checks somewhere else?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/AzureApiManagementExtractor.cs
```diff
@@ -0,0 +1,83 @@
...
+
+namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Proxy;
+
+/// <summary>
+/// Extracts proxy metadata from Azure API Management headers.
+/// </summary>
+internal sealed class AzureApiManagementExtractor : IInferredProxyExtractor
+{
+    // This is the expected value of the x-dd-proxy header
+    private const string ProxyName = "azure-apim";
```
lucaspimentel: I know it's in the comment above, but can we make it more clear in the field name that this is a header value, like `ProxyHeaderValue`?

lucaspimentel: Actually, this is already defined in `InferredProxySpanHelper.AzureProxyHeaderValue`.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/InferredProxySpanHelper.cs
```diff
@@ -32,6 +38,28 @@ internal static class InferredProxySpanHelper
...
+
+        if (string.IsNullOrEmpty(proxyName))
+        {
+            Log.Debug("Missing {HeaderName} header", InferredProxyHeaders.Name);
+            return null;
+        }
+
+        if (string.Equals(proxyName, AzureProxyHeaderValue, StringComparison.OrdinalIgnoreCase))
+        {
+            _azureCoordinator ??= new(new AzureApiManagementExtractor(), new AzureApiManagementSpanFactory());
```
lucaspimentel: Personally, I like having the type name if it's not obvious.
```suggestion
            _azureCoordinator ??= new InferredProxyCoordinator(new AzureApiManagementExtractor(), new AzureApiManagementSpanFactory());
```


I would only use `new()` if the type is already there, like:
```csharp
InferredProxyCoordinator foo = new();
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/InferredProxySpanHelper.cs
```diff
@@ -32,6 +38,28 @@ internal static class InferredProxySpanHelper
...
+            return _azureCoordinator.ExtractAndCreateScope(tracer, carrier, accessor, propagationContext);
+        }
+
+        if (string.Equals(proxyName, AwsProxyHeaderValue, StringComparison.OrdinalIgnoreCase))
+        {
+            _awsCoordinator ??= new(new AwsApiGatewayExtractor(), new AwsApiGatewaySpanFactory());
```
lucaspimentel: See comment above
```suggestion
            _awsCoordinator ??= new InferredProxyCoordinator(new AwsApiGatewayExtractor(), new AwsApiGatewaySpanFactory());
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/InferredProxySpanHelper.cs
```diff
@@ -32,6 +38,28 @@ internal static class InferredProxySpanHelper
...
+            return _awsCoordinator.ExtractAndCreateScope(tracer, carrier, accessor, propagationContext);
+        }
+
+        Log.Debug("Invalid {HeaderName} header with value {Value}", InferredProxyHeaders.Name, proxyName);
```
lucaspimentel: Slight tweak to make it more clear that it's the value that is invalid. Also, quotes.
```suggestion
        Log.Debug("Invalid \"{HeaderName}\" header value: \"{Value}\"", InferredProxyHeaders.Name, proxyName);
```

tracer/test/Datadog.Trace.ClrProfiler.Managed.Tests/AutoInstrumentation/Proxy/AwsApiGatewayExtractorTests.cs
```diff
@@ -4,48 +4,23 @@
...
 using Datadog.Trace.ClrProfiler.AutoInstrumentation.Proxy;
-using Datadog.Trace.Configuration;
 using Datadog.Trace.Headers;
 using Datadog.Trace.TestHelpers.TestTracer;
 using FluentAssertions;
 using Xunit;
 
 namespace Datadog.Trace.ClrProfiler.Managed.Tests.AutoInstrumentation.Proxy;
 
-public class AwsApiGatewayExtractorTests : IAsyncLifetime
```
lucaspimentel, line 19: Is `IAsyncLifetime` not needed anymore?

tracer/test/Datadog.Trace.ClrProfiler.Managed.Tests/AutoInstrumentation/Proxy/AwsApiGatewayExtractorTests.cs
```diff
@@ -92,34 +67,6 @@ public void TryExtract_WithMinimumValidHeaders_ReturnsTrue()
         data.Stage.Should().BeNull();
     }
 
-    [Theory]
-    [InlineData(null)]
-    [InlineData("")]
-    [InlineData("invalid")]
-    [InlineData("aws.apigateway")]
-    public void TryExtract_WithInvalidProxyName_ReturnsFalse(string proxyName)
```
lucaspimentel, line 100: Did these tests move elsewhere?

tracer/test/Datadog.Trace.ClrProfiler.Managed.Tests/AutoInstrumentation/Proxy/AwsApiGatewayExtractorTests.cs
jcstorms1: We do not need it anymore. The extractor no longer takes a tracer parameter which eliminated the need for the interface.

tracer/test/Datadog.Trace.ClrProfiler.Managed.Tests/AutoInstrumentation/Proxy/AwsApiGatewayExtractorTests.cs
jcstorms1: Yeah they were refactored in the `AwsProxyExtractorTests.cs` file.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/AwsApiGatewayExtractor.cs
jcstorms1: Yeah I refactored these into the new proxy tests.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
lucaspimentel: Turns out we can't use that because `TFeature` is a `struct`.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/Isolated/HttpHeadersCollection.cs
```diff
@@ -8,6 +8,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Net.Http.Headers;
+using System.Text;
```
lucaspimentel: We probably don't need this since there are no other changes in this file.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/AzureApiManagementExtractor.cs
```diff
@@ -0,0 +1,83 @@
...
+        start = default;
+
+        if (string.IsNullOrEmpty(startTime))
+        {
+            Log.Debug("Missing header '{HeaderName}'", InferredProxyHeaders.StartTime);
+            return false;
+        }
+
+        // Parse as ISO 8601 timestamp (e.g., "2025-12-03T14:21:01.1900116Z")
+        if (!DateTimeOffset.TryParse(startTime, out var parsedTime))
```
lucaspimentel: We should use `TryParseExact()` and provide an exact format instead of letting it try to guess the correct one.

tracer/test/Datadog.Trace.ClrProfiler.Managed.Tests/AutoInstrumentation/Proxy/ProxyTestHelpers.cs
```diff
@@ -43,4 +43,17 @@ internal static NameValueHeadersCollection CreateValidHeaders(string? start = nu
         headers.Set(InferredProxyHeaders.Stage, "prod");
         return headers;
     }
+
+    internal static NameValueHeadersCollection CreateValidAzureHeaders(string? start = null)
+    {
+        start ??= DateTimeOffset.UtcNow.ToUnixTimeMilliseconds().ToString(CultureInfo.InvariantCulture);
```
lucaspimentel: Is this header meant to be milliseconds from Unit epoch? One of the test (which is failing) says
```
Parse as ISO 8601 timestamp (e.g., "2025-12-03T14:21:01.1900116Z")
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/AwsApiGatewayExtractor.cs
pablomartinezbernardo: nit: Make this similar to AzureApiManagementExtractor so it grabs the value from InferredProxySpanHelper to avoid duplication.

Additionally maybe even consider skipping these `ProxyName` and `AzureApim` fields and reference `InferredProxySpanHelper` directly... Which leads me to think that I don't really like a helper containing constants, but I don't think creating a new constants file for these is worth it, WDYT?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/AzureApiManagementExtractor.cs
```diff
@@ -0,0 +1,62 @@
...
+            if (!InferredProxySpanHelper.GetStartTime(startTimeHeaderValue, out var startTime))
+            {
+                return false;
+            }
+
+            // the remaining headers aren't necessarily required
+            var domainName = ParseUtility.ParseString(carrier, carrierGetter, InferredProxyHeaders.Domain);
+            var httpMethod = ParseUtility.ParseString(carrier, carrierGetter, InferredProxyHeaders.HttpMethod);
+            var path = ParseUtility.ParseString(carrier, carrierGetter, InferredProxyHeaders.Path);
+            var region = ParseUtility.ParseString(carrier, carrierGetter, InferredProxyHeaders.Region);
```
pablomartinezbernardo, line 41: Unused, might be a bug? I guess the intention was for this to end somehow in InferredProxyData?

tracer/src/Datadog.Trace/Tagging/InferredProxyTags.cs
```diff
@@ -12,7 +12,7 @@ namespace Datadog.Trace.Tagging;
 internal sealed partial class InferredProxyTags : InstrumentationTags, IHasStatusCode
 {
     [Tag(Trace.Tags.SpanKind)]
-    public override string SpanKind => SpanKinds.Internal;
```
pablomartinezbernardo, line 15: I would add a note in the releases mentioning this change, as it might break something for customers. Do we have a process to remember doing this when releasing stuff?

tracer/src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
```diff
@@ -654,7 +654,10 @@ private int WriteTags(ref byte[] bytes, int offset, in SpanModel model, ITagProc
 
             // AAS tags need to be set on any span for the backend to properly handle the billing.
             // That said, it's more intuitive to find it on the local root for the customer.
-            if (model.TraceChunk.IsRunningInAzureAppService && model.TraceChunk.AzureAppServiceSettings is { } azureAppServiceSettings)
```
pablomartinezbernardo, line 657: I guess this will keep billing correctly because even if we have no AAS tags in the APIM spans, they will be somewhere else in the stack because there might be an azure function span later on, right? Is there any case where we might not attribute a trace to serverless correctly because of this (Some case where APIM is maybe used without any other serverless component)?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -116,7 +120,9 @@ public static CallTargetState OnFunctionExecutionBegin<TTarget, TFunction>(TTarg
                 }
 
                 var functionName = instanceParam.FunctionDescriptor.ShortName;
-
+                var rootSpanName = tracer.InternalActiveScope?.Root.Span.OperationName;
+                // Check if there's an inferred proxy span (e.g., azure.apim) that we shouldn't overwrite
+                var isProxySpan = rootSpanName?.Contains(AzureApim) == true;
```
pablomartinezbernardo: nit: Should we use string.Equals? Or is it actually possible that the string just contains `azure.apim`?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
pablomartinezbernardo, line 62: This function is starting to get scary complex 😅  I think I would put some effort into maybe refactoring it or testing it (because right now it's untested right?)

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -313,34 +359,34 @@ private static PropagationContext ExtractPropagatedContextFromHttp<T>(T context,
...
-                {
-                    if (keyValuePair.Key.FullName?.Equals("Microsoft.Azure.Functions.Worker.Context.Features.IFunctionBindingsFeature") == true)
-                    {
-                        feature = keyValuePair.Value;
-                        break;
-                    }
-                }
+                var bindingFeature = GetFeatureFromContext<T, GrpcBindingsFeatureStruct>(context, "Microsoft.Azure.Functions.Worker.Context.Features.IFunctionBindingsFeature");
 
-                if (feature is null || !feature.TryDuckCast<FunctionBindingsFeatureStruct>(out var bindingFeature))
```
pablomartinezbernardo, line 326: Is it guaranteed that this is always a `GrpcBindingsFeatureStruct`? Was there a reason previously for it to try casting to `FunctionBindingsFeatureStruct` instead?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -234,11 +251,30 @@ _ when type.StartsWith("cosmosDb", StringComparison.OrdinalIgnoreCase) => "Cosmo
...
                         _ => "Automatic",                                                                            // Automatic is the catch all for any triggers we don't explicitly handle
                     };
-
+                    // need to extract the headers from the context.
+                    // We currently only support httpTrigger, but other triggers may also propagate context,
+                    // e.g. Cosmos + ServiceBus, so we should handle those too
                     switch (triggerType)
                     {
                         case "Http":
-                            extractedContext = ExtractPropagatedContextFromHttp(functionContext, entry.Key as string).MergeBaggageInto(Baggage.Current);
```
pablomartinezbernardo, line 241: Is merging still being done?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -234,11 +251,30 @@ _ when type.StartsWith("cosmosDb", StringComparison.OrdinalIgnoreCase) => "Cosmo
...
+                            Log.Debug("These are the extracted headers {Headers}", extractedHeaders);
+                            InferredProxyScopePropagationContext? proxyContext = null;
+
+                            // Check if there's an active AspNetCore span
+                            var hasAspNetCoreSpan = tracer.InternalActiveScope is { } activeScope &&
+                                                    activeScope.Span.OperationName?.StartsWith("aspnet_core", StringComparison.OrdinalIgnoreCase) == true;
+
+                            if (!hasAspNetCoreSpan && tracer.Settings.InferredProxySpansEnabled && EnvironmentHelpers.IsRunningInAzureFunctionsHost() && extractedHeaders is { } headers)
+                            {
+                                proxyContext = InferredProxySpanHelper.ExtractAndCreateInferredProxyScope(tracer, new AzureHeadersCollectionAdapter(headers), extractedContext);
```
pablomartinezbernardo: The new span is created here, right? I might be missing, but I can't see an explicit dispose, when will the span be closed in that case?

tracer/src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
jcstorms1: Yeah we don't want these tags on the APIM span because it doesn't makes sense and they are inferred. The child function span has all the billing related tags. Also, the APIM span will only be created from the function integration at this time.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/AwsApiGatewayExtractor.cs
jcstorms1: >Additionally maybe even consider skipping these ProxyName and AzureApim fields and reference InferredProxySpanHelper directly... Which leads me to think that I don't really like a helper containing constants, but I don't think creating a new constants file for these is worth it, WDYT?

I thought about doing this, but the header uses a hyphen and the spans use dot notation.

lucaspimentel:
@codex review

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -234,11 +251,30 @@ _ when type.StartsWith("cosmosDb", StringComparison.OrdinalIgnoreCase) => "Cosmo
...
+                    // need to extract the headers from the context.
+                    // We currently only support httpTrigger, but other triggers may also propagate context,
+                    // e.g. Cosmos + ServiceBus, so we should handle those too
                     switch (triggerType)
                     {
                         case "Http":
-                            extractedContext = ExtractPropagatedContextFromHttp(functionContext, entry.Key as string).MergeBaggageInto(Baggage.Current);
+                            var (httpContext, extractedHeaders) = ExtractPropagatedContextandHeadersFromHttp(functionContext, entry.Key as string);
+                            extractedContext = httpContext;
+                            Log.Debug("These are the extracted headers {Headers}", extractedHeaders);
```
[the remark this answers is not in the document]
lucaspimentel: yup, we lost the `.MergeBaggageInto(Baggage.Current)`

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/AzureApiManagementExtractor.cs
```diff
@@ -0,0 +1,62 @@
...
+            data = new InferredProxyData(AzureApim, startTime, domainName, httpMethod, path, null);
+
+            if (Log.IsEnabled(LogEventLevel.Debug))
+            {
+                Log.Debug(
+                    "Successfully extracted proxy data: StartTime={StartTime}, Domain={Domain}, Method={Method}, Path={Path}, Stage={Stage}",
```
lucaspimentel: This is passing `region` as the `{Stage}` value. Consider fixing the log message to use `Region={Region}`.

lucaspimentel: or just remove it entirely since it's not used above
```
 data = new InferredProxyData(AzureApim, startTime, domainName, httpMethod, path, null);
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -298,7 +344,7 @@ _ when type.StartsWith("eventGrid", StringComparison.OrdinalIgnoreCase) => "Even
             return scope;
         }
 
-        private static PropagationContext ExtractPropagatedContextFromHttp<T>(T context, string? bindingName)
+        private static (PropagationContext Context, Dictionary<string, object>? Headers) ExtractPropagatedContextandHeadersFromHttp<T>(T context, string? bindingName)
```
lucaspimentel: small typo: `Contextand` ➡️ `ContextAnd`

```suggestion
        private static (PropagationContext Context, Dictionary<string, object>? Headers) ExtractPropagatedContextAndHeadersFromHttp<T>(T context, string? bindingName)
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -255,37 +291,47 @@ _ when type.StartsWith("eventGrid", StringComparison.OrdinalIgnoreCase) => "Even
...
                 }
 
-                // change root span's type to "serverless"
-                scope.Root.Span.Type = SpanType;
+                if (!isProxySpan)
+                {
+                    // change root span's type to "serverless"
+                    scope.Root.Span.Type = SpanType;
+                    scope.Span.ResourceName = $"{triggerType} {functionName}";
+                    scope.Span.Type = SpanType;
```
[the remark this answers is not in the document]
jcstorms1: The function spans are tagged correctly. This is blocking apim spans from being tagged with unwanted metadata.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -234,11 +251,30 @@ _ when type.StartsWith("cosmosDb", StringComparison.OrdinalIgnoreCase) => "Cosmo
...
+                            var hasAspNetCoreSpan = tracer.InternalActiveScope is { } activeScope &&
+                                                    activeScope.Span.OperationName?.StartsWith("aspnet_core", StringComparison.OrdinalIgnoreCase) == true;
+
+                            if (!hasAspNetCoreSpan && tracer.Settings.InferredProxySpansEnabled && EnvironmentHelpers.IsRunningInAzureFunctionsHost() && extractedHeaders is { } headers)
+                            {
+                                proxyContext = InferredProxySpanHelper.ExtractAndCreateInferredProxyScope(tracer, new AzureHeadersCollectionAdapter(headers), extractedContext);
+                                if (proxyContext != null)
+                                {
+                                    extractedContext = proxyContext.Value.Context;
+                                }
```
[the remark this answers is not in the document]
jcstorms1: It is properly handled https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/src/Datadog.Trace/AspNet/TracingHttpModule.cs#L289

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
jcstorms1: It is handled in https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/src/Datadog.Trace/AspNet/TracingHttpModule.cs like AWS proxy spans.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
jcstorms1: Discussed via a sync. Both structs are based on the same exact Azure feature. We decided to remove the Grpc binding in place of updating the original Function binding.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
jcstorms1: Agree with this, but we need to do this in a later PR.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -234,11 +251,29 @@ _ when type.StartsWith("cosmosDb", StringComparison.OrdinalIgnoreCase) => "Cosmo
...
                         case "Http":
-                            extractedContext = ExtractPropagatedContextFromHttp(functionContext, entry.Key as string).MergeBaggageInto(Baggage.Current);
+                            (extractedContext, var extractedHeaders) = ExtractPropagatedContextAndHeadersFromHttp(functionContext, entry.Key as string);
+                            InferredProxyScopePropagationContext? proxyContext = null;
+
+                            // Check if there's an active AspNetCore span
+                            var hasAspNetCoreSpan = tracer.InternalActiveScope is { } activeScope &&
+                                                    activeScope.Span.OperationName?.StartsWith("aspnet_core", StringComparison.OrdinalIgnoreCase) == true;
+
+                            if (!hasAspNetCoreSpan && tracer.Settings.InferredProxySpansEnabled && EnvironmentHelpers.IsRunningInAzureFunctionsHost() && extractedHeaders is { } headers)
```
pablomartinezbernardo: Is `EnvironmentHelpers.IsRunningInAzureFunctionsHost()` correct there? `CreateIsolatedFunctionScope` is called from `OnIsolatedFunctionBegin` which is [in the worker assembly](https://github.com/DataDog/dd-trace-dotnet/blob/13ff155bb7912b6f09e12f6a1c772e887ccdb00e/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/Isolated/FunctionExecutionMiddlewareInvokeIntegration.cs#L17). My reasoning is likely completely off here because otherwise apim spans wouldn't be created, but I can't see why 🤔 

Even if we postpone refactoring this class to not delay the feature, I would definitely add integration tests with snapshots for this feature, similar to the ones we have for aws.

jcstorms1: We only want to create a proxy span when it is a function host span, so this is the correct check. I believe this code handles the host and worker. Before we were having issues when not checking for the host span.

lucaspimentel: I think Pablo is right. This code in `CreateIsolatedFunctionScope()` is called from `FunctionExecutionMiddlewareInvokeIntegration`, which instruments the `Microsoft.Azure.Functions.Worker.Core` assembly, which runs in the worker process. You should need to check for `IsRunningInAzureFunctionsHost()` here (it should always be `false`, assuming it worked correctly).

lucaspimentel: In other words, I think you should be able to remove lines 267-274 since that code never runs.

pablomartinezbernardo: And if that code is not needed, I now think AzureFunctionsCommons diff can be made much simpler

tracer/src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
```diff
@@ -654,7 +654,11 @@ private int WriteTags(ref byte[] bytes, int offset, in SpanModel model, ITagProc
...
+            // Skip adding AAS tags to inferred proxy spans as they represent infrastructure outside the AAS environment
+            if (model.TraceChunk.IsRunningInAzureAppService &&
+                model.TraceChunk.AzureAppServiceSettings is { } azureAppServiceSettings &&
+                span.Tags is InferredProxyTags proxyTags &&
+                proxyTags.InferredSpan != 1.0)
             {
```
lucaspimentel, line 661: Oops, this was me. The `if` logic is wrong. What we want is

```suggestion
            if (model.TraceChunk.IsRunningInAzureAppService &&
                model.TraceChunk.AzureAppServiceSettings is { } azureAppServiceSettings &&
                !(span.Tags is InferredProxyTags proxyTags && proxyTags.InferredSpan == 1.0))
            {
```

lucaspimentel: We should have tests that fail with the code as I wrote it. If not, we should add some 😅

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/AzureApiManagementSpanFactory.cs
```diff
@@ -0,0 +1,60 @@
...
+internal sealed class AzureApiManagementSpanFactory : IInferredSpanFactory
+{
+    private const string OperationName = "azure.apim";
+    private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor<AzureApiManagementSpanFactory>();
+
+    public Scope? CreateSpan(Tracer tracer, InferredProxyData data, ISpanContext? parent = null)
+    {
+        try
+        {
+            if (data.Path is null)
```
lucaspimentel: Which of the fields in `data` are _really_ required? For example, couldn't we create the proxy span even without the URL path? Does the RFC/spec say anything about this? I ask because instead of returning null early here if `data.Path is null`, we could instead add the null check below to resolve the compiler error you had before. But which one is the desired behavior: no proxy span, or proxy span without the http url tag?

```diff
- var resourceUrl = UriHelpers.GetCleanUriPath(data.Path).ToLowerInvariant();
+ var resourceUrl = data.Path is null ?
+                       null :
+                       UriHelpers.GetCleanUriPath(data.Path).ToLowerInvariant();
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -255,37 +274,43 @@ _ when type.StartsWith("eventGrid", StringComparison.OrdinalIgnoreCase) => "Even
 
                 var functionName = functionContext.FunctionDefinition.Name;
 
-                var tags = new AzureFunctionsTags
-                {
-                    TriggerType = triggerType,
-                    ShortName = functionName,
-                    FullName = functionContext.FunctionDefinition.EntryPoint,
-                };
+                // Check if there's an APIM proxy span that we shouldn't overwrite
```
lucaspimentel: > Check if there's an APIM proxy span that we shouldn't overwrite

Are we doing this check anymore? It looks like `spanRootName` below is not used at all. Maybe this was part of the code you removed you were telling me about? If so, it appears that we can remove these two lines as well.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -116,7 +120,9 @@ public static CallTargetState OnFunctionExecutionBegin<TTarget, TFunction>(TTarg
                 }
 
                 var functionName = instanceParam.FunctionDescriptor.ShortName;
-
+                var rootSpanName = tracer.InternalActiveScope?.Root.Span.OperationName;
+                // Check if there's an inferred proxy span (e.g., azure.apim) that we shouldn't overwrite
+                var isProxySpan = rootSpanName?.Equals(AzureApim) == true;
```
lucaspimentel: This can be simplified
- no need for the `rootSpanName` local variable (low priority)
- prefer `==` over `Equals()` to compare strings unless you need to specify string comparison options (medium priority)

```suggestion
                // Check if there's an inferred proxy span (e.g., azure.apim) that we shouldn't overwrite
                var isProxySpan = tracer.InternalActiveScope?.Root.Span.OperationName == AzureApim;
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -159,17 +168,25 @@ public static CallTargetState OnFunctionExecutionBegin<TTarget, TFunction>(TTarg
...
+                            shortName: functionName,
+                            fullName: instanceParam.FunctionDescriptor.FullName,
+                            bindingSource: bindingSourceType.FullName,
+                            triggerType: triggerType);
+                    }
+                }
+
+                if (!isProxySpan)
+                {
+                    scope.Root.Span.Type = SpanType;
```
lucaspimentel, line 183: Is this logic correct? I understand not doing to the root span if it is a proxy span:
```
scope.Root.Span.Type = SpanType;
```

But the other two lines don't act on the root span:
```
scope.Span.ResourceName = $"{triggerType} {functionName}";
scope.Span.Type = SpanType;
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -234,7 +251,9 @@ _ when type.StartsWith("cosmosDb", StringComparison.OrdinalIgnoreCase) => "Cosmo
                         _ when type.StartsWith("eventGrid", StringComparison.OrdinalIgnoreCase) => "EventGrid",      // Microsoft.Azure.Functions.Worker.Extensions.EventGrid.CosmosDB
                         _ => "Automatic",                                                                            // Automatic is the catch all for any triggers we don't explicitly handle
                     };
-
+                    // need to extract the headers from the context.
+                    // We currently only support httpTrigger, but other triggers may also propagate context,
+                    // e.g. Cosmos + ServiceBus, so we should handle those too
```
lucaspimentel: This comment is outdated to I removed it in #8079. We're extracting context from ServiceBus and EventHub below. It looks like you're accidentally add the comment back (bad merge?) 😅

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/AzureApiManagementExtractor.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            var startTimeHeaderValue = ParseUtility.ParseString(carrier, carrierGetter, InferredProxyHeaders.StartTime);
+
+            // we also need to validate that we have the start time header otherwise we won't be able to create the span
+            if (!InferredProxySpanHelper.GetStartTime(startTimeHeaderValue, out var startTime))
+            {
+                return false;
+            }
+
+            // the remaining headers aren't necessarily required
```
lucaspimentel, line 37: > the remaining headers aren't necessarily required

But we just added a check for a null `path` and we don't create the proxy span 😅  Which is it?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/AzureApiManagementExtractor.cs
lucaspimentel: Un-resolving this comment since `region` is only used for the debug logs. If there is no "region" in Azure APIM like there is in AWS APIGW, we shouldn't even try to extract it from headers.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/InferredProxySpanHelper.cs
```diff
@@ -32,6 +38,56 @@ internal static class InferredProxySpanHelper
         PropagationContext propagationContext)
         where THeadersCollection : struct, IHeadersCollection
     {
-        return Coordinator.ExtractAndCreateScope(tracer, carrier, carrier.GetAccessor(), propagationContext);
+        var accessor = carrier.GetAccessor();
+        var proxyName = ParseUtility.ParseString(carrier, accessor, InferredProxyHeaders.Name);
+
+        if (string.IsNullOrEmpty(proxyName))
```
lucaspimentel: We should use `StringUtil.IsNullOrEmpty()` instead because it fixes the nullability checks in older .NET runtimes. (see [StringUtil.cs](https://github.com/DataDog/dd-trace-dotnet/blob/bbb12d24c5ad2a279e13b914e2c1a44d5feb40aa/tracer/src/Datadog.Trace/Util/StringUtil.cs))

(may need `using Datadog.Trace.Util` above)
```suggestion
        if (StringUtil.IsNullOrEmpty(proxyName))
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/InferredProxySpanHelper.cs
```diff
@@ -32,6 +38,56 @@ internal static class InferredProxySpanHelper
...
+
+        Log.Debug("Invalid \"{HeaderName}\" header value: \"{Value}\"", InferredProxyHeaders.Name, proxyName);
+        return null;
+    }
+
+    public static bool GetStartTime(string? startTime, out DateTimeOffset start)
+    {
+        start = default;
+
+        if (string.IsNullOrEmpty(startTime))
```
lucaspimentel:
```suggestion
        if (StringUtil.IsNullOrEmpty(startTime))
```

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AzureFunctionsTests.cs
```diff
@@ -6,8 +6,10 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Net.Http;
```
lucaspimentel: Where these new `using` directives needed? (seems suspect since there are no other changes in this file)

tracer/test/Datadog.Trace.ClrProfiler.Managed.Tests/AutoInstrumentation/Proxy/AwsApiGatewayExtractorTests.cs
lucaspimentel: thanks

tracer/test/Datadog.Trace.ClrProfiler.Managed.Tests/AutoInstrumentation/Proxy/AwsApiGatewayExtractorTests.cs
lucaspimentel: thanks

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/AzureApiManagementSpanFactory.cs
jcstorms1: I updated this to match how it is handled in AWS.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/AzureApiManagementExtractor.cs
jcstorms1: Changed the other behavior to an empty string.

tracer/src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
jcstorms1: The tests all failed!

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/AzureApiManagementExtractor.cs
jcstorms1: Region is a tag on the span.

lucaspimentel: Ahh, I missed it used in line 43 below. I thought we were passing `null` for the region`.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -254,38 +270,42 @@ _ when type.StartsWith("eventGrid", StringComparison.OrdinalIgnoreCase) => "Even
...
 
                 var functionName = functionContext.FunctionDefinition.Name;
-
-                var tags = new AzureFunctionsTags
-                {
-                    TriggerType = triggerType,
-                    ShortName = functionName,
-                    FullName = functionContext.FunctionDefinition.EntryPoint,
-                };
+                var spanRootName = tracer.InternalActiveScope?.Root.Span.OperationName;
```
lucaspimentel: This `spanRootName` is not used anywhere. Where you going to use it to detect the APIM span and skip adding the Azure Functions tags, like in line 124 above?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/AzureApiManagementSpanFactory.cs
```diff
@@ -0,0 +1,51 @@
...
+using Datadog.Trace.Util;
+
+namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Proxy;
+
+/// <summary>
+/// Creates spans representing requests handled by Azure API Management.
+/// </summary>
+internal sealed class AzureApiManagementSpanFactory : IInferredSpanFactory
+{
+    private const string OperationName = "azure.apim";
```
lucaspimentel: This constant is also defined in `AzureFunctionsCommon`. We should only define it once and reuse it to avoid them drifting over time. The one in `AzureFunctionsCommon` is public so you can already access it from this class.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -29,6 +32,7 @@ internal static class AzureFunctionsCommon
 
         public const string OperationName = "azure_functions.invoke";
         public const string SpanType = SpanTypes.Serverless;
+        public const string AzureApim = "azure.apim";
```
lucaspimentel: This constant is also defined in `AzureApiManagementSpanFactory`. We should only define it once and reuse it to avoid them drifting over time. (This one is public so `AzureApiManagementSpanFactory` can access it.)

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/AzureApiManagementExtractor.cs
```diff
@@ -0,0 +1,60 @@
...
+using Datadog.Trace.Vendors.Serilog.Events;
+
+namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Proxy;
+
+/// <summary>
+/// Extracts proxy metadata from Azure API Management headers.
+/// </summary>
+internal sealed class AzureApiManagementExtractor : IInferredProxyExtractor
+{
+    // This is the expected value of the x-dd-proxy header
```
lucaspimentel: This comment looks like it's in the wrong place. Maybe it was for a field that was removed?
```suggestion
```

tracer/test/Datadog.Trace.ClrProfiler.Managed.Tests/AutoInstrumentation/Proxy/InferredProxySpanHelperTests.cs
```diff
@@ -0,0 +1,177 @@
...
+
+namespace Datadog.Trace.ClrProfiler.Managed.Tests.AutoInstrumentation.Proxy;
+
+public class InferredProxySpanHelperTests
+{
+    private readonly Tracer _tracer;
+
+    public InferredProxySpanHelperTests()
+    {
+        _tracer = ProxyTestHelpers.GetMockTracer();
```
lucaspimentel, line 25: `AzureApiManagementSpanFactoryTests` implements `IAsyncLifetime` and
```csharp
public async Task DisposeAsync() => await _tracer.DisposeAsync();
```

Should we do that here to dispose the tracer? Is that a common pattern in tests?

tracer/src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
```diff
@@ -654,7 +654,10 @@ private int WriteTags(ref byte[] bytes, int offset, in SpanModel model, ITagProc
 
             // AAS tags need to be set on any span for the backend to properly handle the billing.
             // That said, it's more intuitive to find it on the local root for the customer.
-            if (model.TraceChunk.IsRunningInAzureAppService && model.TraceChunk.AzureAppServiceSettings is { } azureAppServiceSettings)
+            // Skip adding AAS tags to inferred proxy spans as they represent infrastructure outside the AAS environment
+            if (model.TraceChunk.IsRunningInAzureAppService &&
+                model.TraceChunk.AzureAppServiceSettings is { } azureAppServiceSettings &&
+                !(span.Tags is InferredProxyTags proxyTags && proxyTags.InferredSpan == 1.0))
```
andrewlock: nit: Mostly an FYI, feel free to ignore, but the pattern matching makes this neater IMO 😄 
```suggestion
                span.Tags is not InferredProxyTags { InferredSpan: 1.0 })
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -127,18 +132,21 @@ public static CallTargetState OnFunctionExecutionBegin<TTarget, TFunction>(TTarg
                     // and update it to be a "serverless" span.
                     var rootSpan = activeScope.Root.Span;
```
andrewlock: nit: you can move `rootSpan` inside the `if`

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -254,38 +270,40 @@ _ when type.StartsWith("eventGrid", StringComparison.OrdinalIgnoreCase) => "Even
...
-                        triggerType: triggerType);
+                    {
+                        var rootSpan = scope.Root.Span;
+                        AzureFunctionsTags.SetRootSpanTags(
+                            rootSpan,
+                            shortName: functionName,
+                            fullName: functionContext.FunctionDefinition.EntryPoint,
+                            bindingSource: rootSpan.Tags is AzureFunctionsTags t ? t.BindingSource : null,
+                            triggerType: triggerType);
+                    }
```
andrewlock: What's going on with these braces? 🤔 They're not actually doing anything AFAICT, so we should remove them, as they add confusion

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/AzureApiManagementSpanFactory.cs
```diff
@@ -0,0 +1,52 @@
...
+
+    public Scope? CreateSpan(Tracer tracer, InferredProxyData data, ISpanContext? parent = null)
+    {
+        try
+        {
+            var resourceUrl = data.Path is null ? string.Empty : UriHelpers.GetCleanUriPath(data.Path).ToLowerInvariant();
+
+            var tags = new InferredProxyTags
+            {
+                HttpMethod = data.HttpMethod,
```
andrewlock, line 32: Does this not need to be `ToUpperInvariant()`? If so, maybe `InferredProxyData` should handle that in it's constructor, so we don't have to call `ToUpperInvariant()` in multiple places?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/InferredProxySpanHelper.cs
```diff
@@ -32,6 +38,56 @@ internal static class InferredProxySpanHelper
         PropagationContext propagationContext)
         where THeadersCollection : struct, IHeadersCollection
     {
-        return Coordinator.ExtractAndCreateScope(tracer, carrier, carrier.GetAccessor(), propagationContext);
+        var accessor = carrier.GetAccessor();
```
andrewlock, line 46: You removed the `tracer?.Settings.InferredProxySpansEnabled` check that was previously in `AwsApiGatewayExtractor`... Should we check it here first and short circuit?
```suggestion
        if (!tracer.Settings.InferredProxySpansEnabled)
        {
            return null;
        }

        var accessor = carrier.GetAccessor();
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/InferredProxySpanHelper.cs
```diff
@@ -32,6 +38,56 @@ internal static class InferredProxySpanHelper
...
+            return _azureCoordinator.ExtractAndCreateScope(tracer, carrier, accessor, propagationContext);
+        }
+
+        if (string.Equals(proxyName, AwsProxyHeaderValue, StringComparison.OrdinalIgnoreCase))
+        {
+            _awsCoordinator ??= new InferredProxyCoordinator(new AwsApiGatewayExtractor(), new AwsApiGatewaySpanFactory());
+            return _awsCoordinator.ExtractAndCreateScope(tracer, carrier, accessor, propagationContext);
+        }
+
+        Log.Debug("Invalid \"{HeaderName}\" header value: \"{Value}\"", InferredProxyHeaders.Name, proxyName);
```
andrewlock: Is it _invalid_, or just "unknown"?

```suggestion
        Log.Debug("Unknown \"{HeaderName}\" header value: \"{Value}\"", InferredProxyHeaders.Name, proxyName);
```

tracer/src/Datadog.Trace/Tagging/InferredProxyTags.cs
andrewlock: 👍 As pointed out, this is potentially a breaking change - we should be able to verify whether it will break customers _before_ pushing this out. If it's not going to break anyone in practice, then we're ok pushing it, but otherwise we need to at least let customers know, or consider _not_ changing it.

tracer/test/Datadog.Trace.ClrProfiler.Managed.Tests/AutoInstrumentation/Proxy/AzureApiManagementSpanFactoryTests.cs
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    public Task InitializeAsync() => Task.CompletedTask;
+
+    public async Task DisposeAsync() => await _tracer.DisposeAsync();
+
+    [Fact]
+    public void CreateSpan_CreatesSpanWithCorrectProperties()
+    {
+        var startTime = DateTimeOffset.UtcNow;
```
andrewlock, line 24: nit: I know you just extracted this from the existing tests, but in general it's a lot easier to create an inline tracer instead of needing all the `IAsyncLifetime` crap (or fields if it comes to that) 😄 
```suggestion

public class AzureApiManagementSpanFactoryTests
{
    [Fact]
    public async Task CreateSpan_CreatesSpanWithCorrectProperties()
    {
        var factory = new AzureApiManagementSpanFactory();
        await using var tracer = ProxyTestHelpers.GetMockTracer();
        var startTime = DateTimeOffset.UtcNow;
```

andrewlock: Note the name change of `factory` and `tracer` 😅 

(Lucas edit: "Not" -> "Note")

tracer/test/Datadog.Trace.Tests/Agent/MessagePack/SpanMessagePackFormatterTests.cs
```diff
@@ -509,6 +510,120 @@ public async Task LastParentId_Tag()
...
+            { "WEBSITE_RESOURCE_GROUP", "SiteResourceGroup" },
+            { "WEBSITE_SITE_NAME", "SiteName" },
+            { "WEBSITE_OS", "windows" },
+            { "WEBSITE_INSTANCE_ID", "InstanceId" },
+            { "COMPUTERNAME", "InstanceName" },
+        };
+
+        var source = new NameValueConfigurationSource(collection);
+        var settings = new TracerSettings(source);
+        var tracer = TracerHelper.Create(settings);
```
andrewlock: You must always dispose these, otherwise they hang around and cause performance problems 
```suggestion
        await using var tracer = TracerHelper.Create(settings);
```

tracer/test/Datadog.Trace.Tests/Agent/MessagePack/SpanMessagePackFormatterTests.cs
```diff
@@ -509,6 +510,120 @@ public async Task LastParentId_Tag()
         tagValue0.Should().Be("0123456789abcdef");
     }
 
+    [Fact]
+    public void Serialize_InferredProxySpan_InAzureAppServices_DoesNotIncludeAasTags()
```
andrewlock:
```suggestion
    public async Task Serialize_InferredProxySpan_InAzureAppServices_DoesNotIncludeAasTags()
```

tracer/test/Datadog.Trace.Tests/Agent/MessagePack/SpanMessagePackFormatterTests.cs
```diff
@@ -509,6 +510,120 @@ public async Task LastParentId_Tag()
...
+        serializedSpan.Tags.Should().NotContainKey("aas.environment.runtime");
+        serializedSpan.Tags.Should().NotContainKey("aas.environment.extension_version");
+
+        // Verify it DOES have the inferred span metric
+        serializedSpan.Metrics.Should().ContainKey("_dd.inferred_span");
+        serializedSpan.Metrics["_dd.inferred_span"].Should().Be(1.0);
+    }
+
+    [Fact]
+    public void Serialize_NonProxySpan_InAzureAppServices_IncludesAasTags()
```
andrewlock:
```suggestion
    public async Task Serialize_NonProxySpan_InAzureAppServices_IncludesAasTags()
```

tracer/test/Datadog.Trace.Tests/Agent/MessagePack/SpanMessagePackFormatterTests.cs
```diff
@@ -509,6 +510,120 @@ public async Task LastParentId_Tag()
...
+            { "WEBSITE_RESOURCE_GROUP", "SiteResourceGroup" },
+            { "WEBSITE_SITE_NAME", "SiteName" },
+            { "WEBSITE_OS", "windows" },
+            { "WEBSITE_INSTANCE_ID", "InstanceId" },
+            { "COMPUTERNAME", "InstanceName" },
+        };
+
+        var source = new NameValueConfigurationSource(collection);
+        var settings = new TracerSettings(source);
+        var tracer = TracerHelper.Create(settings);
```
andrewlock:
```suggestion
        await using var tracer = TracerHelper.Create(settings);
```

tracer/src/Datadog.Trace/Tagging/InferredProxyTags.cs
andrewlock: After looking into this, it seems very unlikely any customers will be affected, and as we consider this change a bug fix essentially, we're just going to accept the potential break. We'll flag it in the release, but otherwise just make the break. If this adversely affects customers, we can always introduce an escape hatch to optionally reinstate the old `span.kind`

tracer/test/Datadog.Trace.ClrProfiler.Managed.Tests/AutoInstrumentation/Proxy/AzureApiManagementSpanFactoryTests.cs
jcstorms1: updated

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/AzureApiManagementSpanFactory.cs
jcstorms1: updated in `InferredProxyData`

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -286,15 +285,15 @@ _ when type.StartsWith("eventGrid", StringComparison.OrdinalIgnoreCase) => "Even
...
+                    AzureFunctionsTags.SetRootSpanTags(
+                        rootSpan,
+                        shortName: functionName,
+                        fullName: functionContext.FunctionDefinition.EntryPoint,
+                        bindingSource: rootSpan.Tags is AzureFunctionsTags t ? t.BindingSource : null,
+                        triggerType: triggerType);
+
                 }
 
                 {
```
andrewlock: Looks like this is another rogue set of braces😅

PR #7808, opened 2025-11-13
Add support for `MongoDB.Driver` v3.5+

andrewlock:
## Summary of changes

Add support for `MongoDB.Driver` v3.5+

## Reason for change

We didn't support 3.5. immediately, as it required using a new version of mongodb in CI - the existing version wouldn't work. We were running into capacity reasons, so we figured we would punt on testing it till later when we could update the VMs.

_Unfortunately_, it turns out 3.5 introduced a bunch of internal changes, which means instrumentation is broken for 3.5 🤦‍♂️ This adds support for it back.

## Implementation details

- Bump the version of mongo tested (this is already pulled on the VMs)
- Fix the integration

## Test coverage

Covered by existing.

I pinned testing of the latest 3.4.x too, but ditched some of the earlier 2.x. versions. We were testing 4 different 2.x.x versions, which seemed a bit overkill, seeing as they all hit the same code paths, and haven't changed in years.

## Other details

https://datadoghq.atlassian.net/browse/APMS-17799

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/MongoDb/IWireProtocol_3_5_ExecuteAsync_Integration.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/MongoDb/IWireProtocol_3_5_ExecuteAsync_Integration.cs
@@ -43,7 +43,7 @@ public class IWireProtocol_3_5_ExecuteAsync_Integration
         return new CallTargetState(scope);
     }
 
-    internal static TReturn? OnAsyncMethodEnd<TTarget, TReturn>(TTarget instance, TReturn? returnValue, Exception exception, in CallTargetState state)
+    internal static TReturn? OnAsyncMethodEnd<TTarget, TReturn>(TTarget instance, TReturn? returnValue, Exception? exception, in CallTargetState state)
     {
         var scope = state.Scope;
 
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/MongoDb/IWireProtocol_3_5_Execute_Integration.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/MongoDb/IWireProtocol_3_5_Execute_Integration.cs
@@ -35,7 +35,7 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.MongoDb;
 [EditorBrowsable(EditorBrowsableState.Never)]
 public class IWireProtocol_3_5_Execute_Integration
 {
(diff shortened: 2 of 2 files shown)


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/MongoDb/IWireProtocol_3_5_ExecuteAsync_Integration.cs
```diff
@@ -0,0 +1,54 @@
...
+    MinimumVersion = MongoDbIntegration.Major3Minor5,
+    MaximumVersion = MongoDbIntegration.Major3,
+    MethodName = "ExecuteAsync",
+    ParameterTypeNames = ["MongoDB.Driver.OperationContext", "MongoDB.Driver.Core.Connections.IConnection"],
+    ReturnTypeName = ClrNames.GenericTaskWithGenericClassParameter,
+    TypeNames = new[]
+    {
+        "MongoDB.Driver.Core.WireProtocol.CommandUsingQueryMessageWireProtocol`1",
+        "MongoDB.Driver.Core.WireProtocol.CommandUsingCommandMessageWireProtocol`1",
+        "MongoDB.Driver.Core.WireProtocol.CommandWireProtocol`1",
```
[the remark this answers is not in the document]
andrewlock: The sync methods targeting it are a mistake. These don't need to target the v2 assembly (MongoDB.Driver.Core) so there's no point in adding those assemblies. Will fix the sync version to remove the v2 assembly instead

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/MongoDb/IWireProtocol_3_5_Execute_Integration.cs
```diff
@@ -0,0 +1,64 @@
...
+        "MongoDB.Driver.Core.WireProtocol.CommandUsingCommandMessageWireProtocol`1",
+        "MongoDB.Driver.Core.WireProtocol.CommandWireProtocol`1",
+    })]
+#pragma warning restore SA1118
+// ReSharper disable once InconsistentNaming
+[Browsable(false)]
+[EditorBrowsable(EditorBrowsableState.Never)]
+public class IWireProtocol_3_5_Execute_Integration
+{
+    internal static CallTargetState OnMethodBegin<TTarget, TOperationContext, TConnection>(TTarget instance, TOperationContext operationContext, TConnection connection)
```
NachoEchevarria: NIT: Should operationContext be declared as nullable like in the ExecuteAsync integration? It's more a cosmetic thing since the parameter is not used.

andrewlock: Yeah, probably makes sense to be safe, thanks! 🙂

PR #7813, opened 2025-11-13
[OTEL] Removing OTLP Tests Flakiness

link04:
## Summary of changes
Attempt to hopefully remove flakes in SubmitsOtlpMetrics and SubmitsOtlpLogs tests.

## Implementation details
For metrics started calling the ForceFlush method: `meterProvider?.ForceFlush();`

Logs doesn't have that but I was using the wrong env var before so changed it to use the same default we use in the tracer implementation to hope dispose on time:
```
            - SetEnvironmentVariable("OTEL_LOG_EXPORT_INTERVAL", "1000");
            + SetEnvironmentVariable("OTEL_BLRP_SCHEDULE_DELAY", "1000");
```

If the above doesn't work maybe we would need to look into doing Simple Batch for logs instead: https://github.com/open-telemetry/opentelemetry-dotnet/issues/5102#issuecomment-1836570001

## Test coverage
Also removed the `[Flaky("New test agent seems to not always be ready", maxRetries: 3)]` so hoping they all stop flaking 🤞.


tracer/test/snapshots/OpenTelemetrySdkTests.SubmitsOtlpLogs.verified.txt
andrewlock, line 1: For reviewers: the only change in all these snapshots is the removal of the enclosing `[]`:

image

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/OpenTelemetrySdkTests.cs
```diff
@@ -335,7 +333,7 @@ public async Task SubmitsOtlpLogs(string packageVersion, string datadogLogsEnabl
             SetEnvironmentVariable("OTEL_LOGS_EXPORTER_ENABLED", otelLogsEnabled);
             SetEnvironmentVariable("OTEL_EXPORTER_OTLP_PROTOCOL", protocol);
             SetEnvironmentVariable("OTEL_EXPORTER_OTLP_ENDPOINT", $"http://{testAgentHost}:{otlpPort}");
-            SetEnvironmentVariable("OTEL_LOG_EXPORT_INTERVAL", "1000");
+            SetEnvironmentVariable("OTEL_BLRP_SCHEDULE_DELAY", "1000");
```
andrewlock, line 336: I think this is the default value, so I'm not sure this will change anything 🤔

tracer/test/test-applications/integrations/Samples.OpenTelemetrySdk/Program.cs
```diff
@@ -144,10 +144,10 @@ public static async Task Main(string[] args)
         }
 
 #if OTEL_1_2
+        meterProvider?.ForceFlush();
         meterProvider?.Dispose();
 #endif
 #if OTEL_1_9
-
         loggerFactory?.Dispose();
```
andrewlock, line 151: We're already disposing these, because they're created in `using` statements, so disposing them twice probably isn't correct 🤔 The question is why isn't disposing _always_ flushing the logs correctly in the otel-only case. This feels like a timeout/bug or something in the otel exporter, @zacharycmontoya? 🤔

PR #7814, opened 2025-11-13
[Logs] Fix OTLP Logs timestamp during serialization

zacharycmontoya:
## Summary of changes
Fixes the serialization of the OTLP Logs timestamps fields.

## Reason for change
Before this change, we would start with a UTC DateTime for the LogRecord object and then when we would encode it as a Unix epoch time we applied a "local->UTC time" conversion by using DateTime.ToUniversalTime(), which shifted the time. This resulted in OTLP logs from the PST timezone being reported as having a local time 7 hours in the future.

## Implementation details
- Change the `Timestamp` property of `Datadog.Trace.OpenTelemetry.Logs.LogPoint` to have type `DateTimeOffset` instead of `DateTime` so we're always dealing with a UTC time, and initialize the value to `DateTimeOffset.UtcNow`
- Remove the previous encoding of DateTime->Unix nanoseconds and replace it with a call to the existing extension method `TimeExtensions.ToUnixTimeNanoseconds`, which is already used for serializing timestamps for Metrics and Traces

## Test coverage
Add a regression test in `OpenTelemetrySdkTests.SubmitsOtlpLogs` by asserting that the timestamp in the OTLP log payload is in the expected range (application lifetime) before scrubbing the data and running the snapshot test.

## Other details
N/A

### DIFF
--- a/tracer/src/Datadog.Trace/ExtensionMethods/TimeExtensions.cs
+++ b/tracer/src/Datadog.Trace/ExtensionMethods/TimeExtensions.cs
@@ -12,16 +12,6 @@ namespace Datadog.Trace.ExtensionMethods
         /// <summary>
         /// Returns the number of nanoseconds that have elapsed since 1970-01-01T00:00:00.000Z.
         /// </summary>
-        /// <param name="dateTime">The value to get the number of elapsed nanoseconds for.</param>
-        /// <returns>The number of nanoseconds that have elapsed since 1970-01-01T00:00:00.000Z.</returns>
-        public static long ToUnixTimeNanoseconds(this DateTime dateTime)
-        {
-            return (dateTime.Ticks - TimeConstants.UnixEpochInTicks) * TimeConstants.NanoSecondsPerTick;
-        }
-
-        /// <summary>
-        /// Returns the number of nanoseconds that have elapsed since 1970-01-01T00:00:00.000Z.
-        /// </summary>
         /// <param name="dateTimeOffset">The value to get the number of elapsed nanoseconds for.</param>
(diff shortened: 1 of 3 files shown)


tracer/src/Datadog.Trace/ExtensionMethods/TimeExtensions.cs
```diff
@@ -9,6 +9,16 @@ namespace Datadog.Trace.ExtensionMethods
...
     internal static class TimeExtensions
     {
+        /// <summary>
+        /// Returns the number of nanoseconds that have elapsed since 1970-01-01T00:00:00.000Z.
+        /// </summary>
+        /// <param name="dateTime">The value to get the number of elapsed nanoseconds for.</param>
+        /// <returns>The number of nanoseconds that have elapsed since 1970-01-01T00:00:00.000Z.</returns>
+        public static long ToUnixTimeNanoseconds(this DateTime dateTime)
+        {
+            return (dateTime.Ticks - TimeConstants.UnixEpochInTicks) * TimeConstants.NanoSecondsPerTick;
```
lucaspimentel: Since the Unix epoch is defined as UTC, should we handle different `dateTime.Kind` values here? The `dateTime` could be a UTC kind or not, so it may need a conversion to UTC to get the right "unix time."

lucaspimentel: Or consider using `DateTimeOffset` instead to avoid the ambiguity.

zacharycmontoya: Good call. Changing to `DateTimeOffset` will help enforce this by design. Updated in 58e0760bae8f03648d552be86f5186c754e557be

PR #7818, opened 2025-11-14
Respect baggage in Azure messaging integrations

pablomartinezbernardo:
## Summary of changes

Respect baggage in Azure messaging instrumentations. Related to #7810 

## Reason for change

Baggage was not being extracted from headers

## Test coverage

Unit test `AzureFunctionsCommon.ExtractPropagatedContextFromMessaging` with baggage merging. `CreateIsolatedFunctionScope` would have been more complex to test.

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
@@ -403,7 +403,7 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Azure.Functions
                 }
 
                 bool areAllTheSame = extractedContexts.Count == 1 ||
-                                     (extractedContexts.Count > 1 && AreAllContextsIdentical(extractedContexts));
+                                     (extractedContexts.Count > 1 && AreAllSpanContextsIdentical(extractedContexts));
 
                 if (!areAllTheSame)
                 {
@@ -440,7 +440,8 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Azure.Functions
             }
         }
 
-        private static bool AreAllContextsIdentical(List<PropagationContext> contexts)
+        // Checks if all SpanContexts are identical (ignores baggage)
+        private static bool AreAllSpanContextsIdentical(List<PropagationContext> contexts)
         {
             if (contexts.Count <= 1)
             {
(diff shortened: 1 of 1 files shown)


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -443,17 +440,18 @@ private static bool TryParseJson<T>(object? jsonObj, [NotNullWhen(true)] out T?
...
                 return true;
             }
 
-            var first = contexts[0];
+            var first = contexts[0].SpanContext;
             return contexts.All(ctx =>
-                ctx.TraceId128.Equals(first.TraceId128) &&
-                ctx.SpanId == first.SpanId);
+                ctx.SpanContext != null &&
+                ctx.SpanContext.TraceId128.Equals(first!.TraceId128) &&
```
lucaspimentel: [nit-picking coding style] You can use the `==` operation here. It calls `TraceId.Equals(TraceId)` internally and is more consistent with the lines above and below.

```suggestion
                ctx.SpanContext.TraceId128 == first!.TraceId128 &&
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -443,17 +440,18 @@ private static bool TryParseJson<T>(object? jsonObj, [NotNullWhen(true)] out T?
             }
         }
 
-        private static bool AreAllContextsIdentical(List<SpanContext> contexts)
+        private static bool AreAllContextsIdentical(List<PropagationContext> contexts)
```
lucaspimentel: We should probably rename this method now because the name and signature imply it's comparing all the `PropagationContext` items, but it's only comparing the `SpanContext` property in each and ignoring baggage.

Maybe `AreAllSpanContextsIdentical()`? And add a comment to explain that it ignores baggage?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/Functions/AzureFunctionsCommon.cs
```diff
@@ -388,27 +389,28 @@ private static PropagationContext ExtractPropagatedContextFromMessaging<T>(T con
...
+                bool areAllTheSame = extractedContexts.Count == 1 ||
+                                     (extractedContexts.Count > 1 && AreAllSpanContextsIdentical(extractedContexts));
 
                 if (!areAllTheSame)
                 {
                     Log.Warning("Multiple different contexts found in messages. Using first context for parentship.");
                 }
 
-                return new PropagationContext(spanContexts[0], Baggage.Current, null);
+                return extractedContexts[0];
```
pablomartinezbernardo, line 413: This was actually the bug resulting in Baggage.Current merging into itself

PR #7821, opened 2025-11-14
Avoid allocating and preparing CI Visibility stuff in startup hot path

andrewlock:
## Summary of changes

Avoid allocating and preparing CI Visibility stuff in startup hot path

## Reason for change

Currently CI Visibility uses a bunch of lazy settings, but then we immediately instantiate those right at the start of the app to decide _if_ we should initialize CI Visibility 😅 

This just extracts out some of that work out to minimize the work if CI Vis is not enabled.

This shaves ~50ms off startup

## Implementation details

Make things lazy that can be, make things non-lazy that can't be!

Concretely:
- `Enabled` shouldn't be lazy because we check it immediately. Therefore:
  - Extract it from `TestOptimizationSettings` (as it doesn't need to be there anyway)
  - Make `InternalEnabled()` testable by moving to a helper (not done here, but could be)
  - Calculate it immediately in the ctor
- Make `CIEnvironmentVariables` calculation lazy, as we don't need it if we're not in CI Vis

## Test coverage

Same coverage right now.

### DIFF
--- a/tracer/src/Datadog.Trace/Ci/TestOptimization.cs
+++ b/tracer/src/Datadog.Trace/Ci/TestOptimization.cs
@@ -26,6 +26,7 @@ internal class TestOptimization : ITestOptimization
 
     private readonly Lazy<CIEnvironmentValues> _ciVariablesLazy;
     private int _firstInitialization = 1;
+    private TestOptimizationDetection.Enablement _enablement;
     private TestOptimizationSettings? _settings;
     private ITestOptimizationClient? _client;
     private Task? _additionalFeaturesTask;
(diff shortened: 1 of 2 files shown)


tracer/src/Datadog.Trace/Ci/TestOptimization.cs
```diff
@@ -427,72 +428,8 @@ public void Reset()
         _dynamicInstrumentationFeature = null;
     }
 
-    private bool InternalEnabled()
```
andrewlock, line 430: This method has moved to `TestOptimizationDetection.IsEnabled` instead

tracer/src/Datadog.Trace/Ci/Configuration/TestOptimizationSettings.cs
```diff
@@ -26,20 +26,8 @@ internal class TestOptimizationSettings
         public TestOptimizationSettings(IConfigurationSource source, IConfigurationTelemetry telemetry)
         {
             var config = new ConfigurationBuilder(source, telemetry);
-            Enabled = config.WithKeys(ConfigurationKeys.CIVisibility.Enabled).AsBool();
```
andrewlock, line 29: This has moved to `TestOptimizationDetection`

PR #7827, opened 2025-11-14
Fixing DSM flush logic & CPU usage

robcarlan-datadog:
## Summary of changes
Moves DSM flush work to a dedicated thread to improve thread pool performance. Also simplifies the flushing logic to more closely match the AgentWriter logic.

(See below for test results)

## Reason for change

A = DSM enabled after the fix
B = DSM enabled before the fix
Screenshot 2025-12-01 at 5 23 44 pm

DSM flush logic was previously running entirely on the thread pool, which includes the task to frequently move DSM checkpoints off of the buffer and add them to the backlog. This was causing thread contention shown in the above profile.

## Implementation details

Moves the DSM buffer flushing logic into it's own thread. Below is a profile after the fix (DSM disabled on the right).

Screenshot 2025-12-01 at 5 26 53 pm

## Test coverage

Mostly manual testing. I ran a set of three apps:
1. DSM enabled, after the fix
2. DSM disabled, after the fix
3. DSM enabled, before the fix

The test app just forwards a Kafka message every 0.5 seconds.

And run those via docker-compose with a restricted CPU count of 0.1 and 0.5 in the below screenshots. In both cases, DSM disabled is the lowest CPU usage, and the line just above is DSM enabled after the fix. The highest CPU is DSM enabled, prior to the fix. The performance is a lot better, but I'm not sure if the gap can be closed further.

Screenshot 2025-12-01 at 6 30 10 pm

Screenshot 2025-12-01 at 5 55 28 pm

### DIFF
--- a/tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
+++ b/tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
@@ -226,30 +226,28 @@ internal class DataStreamsWriter : IDataStreamsWriter
             return;
         }
 
-        if (!_flushSemaphore.Wait(TimeSpan.FromMilliseconds(100)))
+        if (await _flushSemaphore.WaitAsync(TimeSpan.FromMilliseconds(100)).ConfigureAwait(false))
         {
-            return;
-        }
-
-        try
-        {
-            while (_buffer.TryDequeue(out var statsPoint))
+            try
             {
-                _aggregator.Add(in statsPoint);
-            }
+                while (_buffer.TryDequeue(out var statsPoint))
+                {
+                    _aggregator.Add(in statsPoint);
+                }
 
-            while (_backlogBuffer.TryDequeue(out var backlogPoint))
-            {
-                _aggregator.AddBacklog(in backlogPoint);
+                while (_backlogBuffer.TryDequeue(out var backlogPoint))
+                {
+                    _aggregator.AddBacklog(in backlogPoint);
+                }
             }
-        }
-        catch (Exception ex)
-        {
-            Log.Error(ex, "An error occured while processing data streams buffers");
-        }
-        finally
-        {
-            _flushSemaphore.Release();
+            catch (Exception ex)
+            {
+                Log.Error(ex, "An error occured while processing data streams buffers");
+            }
+            finally
+            {
+                _flushSemaphore.Release();
+            }
         }
 
         await FlushAggregatorAsync().ConfigureAwait(false);


tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
```diff
@@ -285,41 +322,28 @@ private async Task ProcessQueueLoopAsync()
...
             {
-                if (isFinalFlush)
-                {
-                    return;
-                }
-
-                // do one more loop to make sure everything is flushed
-                RequestFlush();
-                isFinalFlush = true;
-                continue;
```
robcarlan-datadog, line 312: Might need to call the submit API inline here a final time

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
```diff
@@ -93,10 +92,10 @@ private void Initialize()
                 return;
             }
 
-            _processTask = Task.Run(ProcessQueueLoopAsync);
+            _processTask = Task.Factory.StartNew(ProcessQueueLoopAsync, TaskCreationOptions.LongRunning);
```
kr-igor: Can we remove Async, since the method is not async?

kr-igor (approved):

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
robcarlan-datadog: Will remove

tracer/src/Datadog.Trace.Trimming/build/Datadog.Trace.Trimming.xml
```diff
@@ -1068,6 +1068,7 @@
       <type fullname="System.OperationCanceledException" />
       <type fullname="System.Runtime.CompilerServices.AsyncTaskMethodBuilder" />
       <type fullname="System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1" />
+      <type fullname="System.Runtime.CompilerServices.AsyncVoidMethodBuilder" />
```
andrewlock: Is this actually required? 🤔 `async void` is normally something to be vehemently avoided, but I can't see where it's coming from here 🤔

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
```diff
@@ -93,10 +92,10 @@ private void Initialize()
...
             _processTask.ContinueWith(t => Log.Error(t.Exception, "Error in processing task"), TaskContinuationOptions.OnlyOnFaulted);
             _flushTimer = new Timer(
-                x => ((DataStreamsWriter)x!).RequestFlush(),
+                async x => await ((DataStreamsWriter)x!).FlushAsync().ConfigureAwait(false),
```
andrewlock: ah, this is the `async void`. Unfortunately, this is somewhat dangerous, because `Timer` doesn't understand `Task` 🙁 . I think we'll need to rework this with a "manual" async timer unfortunately. There's `PeriodicTimer` in .NET 6, but that doesn't really help us. In other parts of the code we have essentially "manually" implemented code similar to this - you start a task, and Task.Delay() the required amount (accounting for drift) and then fire the Flush.

Another option is to keep using the `Timer`, but to make it simply `RequestFlush` again, and then separately have a `FlushTaskLoopAsync` which waits for this signal. This is pretty much _exactly_ what we have in the `AgentWriter` at that point 😄

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
```diff
@@ -268,11 +247,22 @@ private async Task WriteToApiAsync()
...
 
-    private async Task ProcessQueueLoopAsync()
+    private void ProcessQueueLoopAsync()
     {
-        var isFinalFlush = false;
         while (true)
         {
+            Thread.Sleep(_waitTimeSpan);
+
+            if (!_flushSemaphore.Wait(_flushSemaphoreWaitTime))
```
andrewlock, line 323: Just to confirm, this basically avoids doing the aggregation while a flush is running, correct?

Are there any concerns that this could cause a "too large" build up of points here if (for example) the flush ends up timing out during the send (I'm not sure what your configured time out for the API flush is)?

tracer/src/Datadog.Trace.Trimming/build/Datadog.Trace.Trimming.xml
robcarlan-datadog: Fixed as part of below changes

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
robcarlan-datadog: Updated to create a new Task and run in the same way as AgentWriter does. Will quickly rerun the performance test now

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
robcarlan-datadog: Correct. Talked about this a bit offline, it's a known risk. We can definitely do something smarter such as double-buffering in the AgentWriter if we see that this starts becoming an issue.

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
```diff
@@ -184,52 +185,68 @@ private async Task FlushAndCloseAsync()
         {
             Log.Error("Could not flush all data streams stats before process exit");
         }
+
+        await FlushAsync().ConfigureAwait(false);
```
andrewlock, line 189: I think we _may_ need a more aggressive timeout here, similar to what you have for `completedTask` above, to handle the case where we can't flush to the agent. The reason being that this will block other components from shutting down and flushing, and this can be a bit of an issue for last data, but also for delaying shutdown in serverless envs (for example). 

We can also likely make the above timeout (processing task) of 20s much shorter, seeing as it really should be ~10ms max (as there's no flushing!)

robcarlan-datadog: Good call! I've updated both timeouts to 1 second

lucaspimentel: > delaying shutdown in serverless envs

Thanks! Can we also explicitly flush here in the AWS Lambda code? We already flush traces and DSM data with a configurable timeout.
https://github.com/DataDog/dd-trace-dotnet/blob/6347a3f7da20b04e5260a3585ae28c75f68ea7b6/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Lambda/LambdaCommon.cs#L84-L89

lucaspimentel: The reason we should flush at the end of a AWS Lambda invocation is that the Lambda runtime can freeze the function runtime (and the tracer inside it) at any time once the invocation is finished, so we want to ensure we flush all data before that happens. Otherwise the flush may not happen until the runtime is unfrozen again, which can be a long time, or never happen.

For example, if flushing only happens every 1 sec, and your Lambda function finishes in 100ms, that means the first flush won't happen for another 900ms after the invocation finishes. This flush could be delayed indefinitely, or never happen at all.

lucaspimentel: @andrewlock should we add one "flush everything" helper method that captures all of these? Not in this PR, just a nice helper we can add later and use in several places. And not just the serverless invocations. Don't we also have similar logic during app shutdown or on app domain close?

lucaspimentel: Sorry, ignore my comments above!

For some reason I thought this PR was for DBM, and I was like, "hey should DBM do this flushing thing in AWS Lambda that DSM is already doing?" 😅 

(and yes, I did think for a second, "why does DBM have anything to flush? I thought they just injected data into SQL query comments?")

andrewlock: > @andrewlock should we add one "flush everything" helper method that captures all of these? Not in this PR, just a nice helper we can add later and use in several places. And not just the serverless invocations. Don't we also have similar logic during app shutdown or on app domain close?

We certainly _could_ 🤔 I'm not sure it makes sense for everything, e.g. forcing telemetry flushing would mess up the stats, but given that it's basically only serverless/Lambda that has this issue today, don't we _Effecively_ already have a "centralised" place? It's the lambda handler code 😄

My only hesitation about moving it elsewhere is about making it _appear_ more generalised and "safe" than it _necessarily_ is? 🤔

lucaspimentel: Makes sense, thanks.

PR #7828, opened 2025-11-17
Publish native symbols in all DCs

gleocadie:
## Summary of changes

Publishing .NET APM native symbols to alls DCs.

## Reason for change

When resolving symbols for crash reports in a DC different from us1, the deobfuscation API is unable to get the symbol for a specific frame.

## Implementation details

- Create a github environment and create secrets with all the API keys we need.
- update the action to take as parameter all the API keys
- Use the github environment on the different jobs we need to upload the symbol files

## Test coverage
https://github.com/DataDog/dd-trace-dotnet/actions/runs/19435230104/job/55604220909
-> Upload to preprod is failing (same with the current state of the job)

### DIFF
--- a/.github/actions/publish-debug-symbols/action.yml
+++ b/.github/actions/publish-debug-symbols/action.yml
@@ -5,14 +5,27 @@ inputs:
   artifacts_path:
     description: "Path to the build artifacts"
     required: true
-  preprod_key:
-    description: "Preprod API key"
-    default: ""
-    required: false
-  public_symbols_key:
-    description: "Public symbols API key"
-    default: ""
-    required: false
+  preprod_api_key:
+    description: "Pre-production API key"
+    required: true
+  us1_api_key:
+    description: "US1 API key"
+    required: true
+  us3_api_key:
+    description: "US3 API key"
+    required: true
+  us5_api_key:
+    description: "US5 API key"
+    required: true
+  eu1_api_key:
+    description: "EU1 API key"
+    required: true
+  ap1_api_key:
+    description: "AP1 API key"
+    required: true
+  ap2_api_key:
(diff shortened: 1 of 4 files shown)


.github/actions/publish-debug-symbols/action.yml
```diff
@@ -84,17 +97,14 @@ runs:
 
     - name: Push debug symbols
       shell: bash
+      env:
+        BOLD_GREEN: "\e[1;32m"
+        RESET: "\e[0m"
```
andrewlock, line 102: Very important addition 😂

.github/workflows/create_draft_release.yml
```diff
@@ -13,6 +13,7 @@ on:
 jobs:
   create_draft_release:
     runs-on: ubuntu-latest
+    environment: publish-debug-symbols-env
```
andrewlock, line 16: Should the environment be called something more generic? 🤔 We can only have a single environment "active" at a time, right?

Also, secrets that _aren't_ in an environment... are they still available?

Also also, how does `environment` actually "improve" anything here? any ideas?

gleocadie: > Should the environment be called something more generic? 🤔 We can only have a single environment "active" at a time, right?

we can, I was doing it for `symbols publishing` first and see after.
From what I read, you can have only one environment for a job. A job can use secrets that aren't in an environment in addition to the ones defined in an environment. The ones in an environment will override if there secrets with the same name not in a environment.

> Also also, how does environment actually "improve" anything here? any ideas?

The only benefit I see, is not being pinged by DD security bot.... joke aside, I do not know. I haven't used them a lot (first time though)

PR #7833, opened 2025-11-17
[Dynamic Instrumentation] DEBUG-4089 Support file-based debugger probes

dudikeleti:
## Summary of changes
Added support to load log/metric/span/span‑decoration probes from a JSON file (DebuggerSettings.ProbeFile)

## Reason for change
Dynamic Instrumentation only worked with RCM‑delivered probes. There was no support for loading probes from a local file, and the internal configuration state wasn’t designed for combining file and RCM probes or handling file‑only 

## Implementation details

- Added support to load log/metric/span/span‑decoration probes from a JSON file (DebuggerSettings.ProbeFile) during DynamicInstrumentation initialization.
- Ensured Dynamic Instrumentation can operate in three modes: RCM‑only, file‑only, and combined file+RCM.
- Implemented source‑level merge (MergeProbes) where RCM wins on ID conflicts with file probes, and configuration‑level union (MergeConfigurations).
- Updated initialization to start background processing and status/snapshot upload when either RCM is available or file probes exist.

## Test coverage
Updated DynamicInstrumentationTests to cover file probe loading (multiple probe types, invalid/missing/empty files, no file configured, partially valid file entries, and duplicate IDs within a file) and merge semantics between file and RCM probes.

### DIFF
--- a/tracer/src/Datadog.Trace/Debugger/Configurations/ConfigurationUpdater.cs
+++ b/tracer/src/Datadog.Trace/Debugger/Configurations/ConfigurationUpdater.cs
@@ -20,7 +20,7 @@ namespace Datadog.Trace.Debugger.Configurations
         private readonly string? _env;
         private readonly string? _version;
         private readonly int _maxProbesPerType;
-        private readonly HashSet<string> _removedRcmProbeIds = new(StringComparer.Ordinal);
+        private readonly HashSet<string> _removedRcmProbeIds = new();
         private Func<IReadOnlyList<ProbeDefinition>, List<UpdateResult>>? _handleAddedProbesChanges;
         private Action<string[]>? _handleRemovedProbesChanges;
 
--- a/tracer/src/Datadog.Trace/Debugger/Configurations/ConfigurationUpdater.cs
+++ b/tracer/src/Datadog.Trace/Debugger/Configurations/ConfigurationUpdater.cs
@@ -40,7 +40,13 @@ namespace Datadog.Trace.Debugger.Configurations
         {
             var result = new List<UpdateResult>();
             var filteredConfiguration = ApplyConfigurationFilters(configuration);
-            var comparer = new ProbeConfigurationComparer(_currentConfiguration, filteredConfiguration);
+
+            // Merge the new filtered configuration with the existing one so that
+            // _currentConfiguration always represents the union of all accepted
+            // configurations from both file and RCM
+            var mergedConfiguration = MergeConfigurations(_currentConfiguration, filteredConfiguration);
+
+            var comparer = new ProbeConfigurationComparer(_currentConfiguration, mergedConfiguration);
 
             if (comparer.HasProbeRelatedChanges)
             {
@@ -52,7 +58,7 @@ namespace Datadog.Trace.Debugger.Configurations
                 HandleRateLimitChanged(comparer);
             }
 
-            _currentConfiguration = configuration;
+            _currentConfiguration = mergedConfiguration;
 
             return result;
         }
@@ -67,6 +73,30 @@ namespace Datadog.Trace.Debugger.Configurations
             {
                 Log.Error(ex, "Failed to remove configurations");
             }
+        }
+
+        private static ProbeConfiguration MergeConfigurations(ProbeConfiguration current, ProbeConfiguration incoming)
+        {
+            return new ProbeConfiguration
+            {
+                LogProbes = current.LogProbes
+                                   .Concat(incoming.LogProbes)
+                                   .Distinct()
+                                   .ToArray(),
+                MetricProbes = current.MetricProbes
+                                      .Concat(incoming.MetricProbes)
+                                      .Distinct()
+                                      .ToArray(),
+                SpanProbes = current.SpanProbes
+                                    .Concat(incoming.SpanProbes)
+                                    .Distinct()
+                                    .ToArray(),
+                SpanDecorationProbes = current.SpanDecorationProbes
+                                              .Concat(incoming.SpanDecorationProbes)
+                                              .Distinct()
+                                              .ToArray(),
+                ServiceConfiguration = incoming.ServiceConfiguration ?? current.ServiceConfiguration
+            };
         }
 
         private ProbeConfiguration ApplyConfigurationFilters(ProbeConfiguration configuration)
--- a/tracer/src/Datadog.Trace/Debugger/Configurations/ConfigurationUpdater.cs
+++ b/tracer/src/Datadog.Trace/Debugger/Configurations/ConfigurationUpdater.cs
@@ -20,8 +20,11 @@ namespace Datadog.Trace.Debugger.Configurations
         private readonly string? _env;
         private readonly string? _version;
         private readonly int _maxProbesPerType;
+        private readonly HashSet<string> _removedRcmProbeIds = new(StringComparer.Ordinal);
 
         private ProbeConfiguration _currentConfiguration;
+        private ProbeConfiguration? _fileConfiguration;
+        private ProbeConfiguration _rcmConfiguration;
 
         private ConfigurationUpdater(string? env, string? version, int maxProbesPerType)
         {
@@ -29,6 +32,7 @@ namespace Datadog.Trace.Debugger.Configurations
             _version = version;
             _maxProbesPerType = maxProbesPerType;
             _currentConfiguration = new ProbeConfiguration();
+            _rcmConfiguration = new ProbeConfiguration();
         }
 
         public static ConfigurationUpdater Create(string? environment, string? serviceVersion, int maxProbesPerType)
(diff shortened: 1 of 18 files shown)


tracer/src/Datadog.Trace/Debugger/Configurations/ConfigurationUpdater.cs
```diff
@@ -20,26 +20,85 @@ internal sealed class ConfigurationUpdater
         private readonly string? _env;
         private readonly string? _version;
         private readonly int _maxProbesPerType;
+        private readonly HashSet<string> _removedRcmProbeIds = new(StringComparer.Ordinal);
```
andrewlock: nit: ordinal is the default, and it's better to not specify it in these cases, as there are often internal fast-paths in the types 🙂 

```suggestion
        private readonly HashSet<string> _removedRcmProbeIds = new();
```

tracer/src/Datadog.Trace/Debugger/Configurations/ProbeConfigurationUtils.cs
```diff
@@ -0,0 +1,130 @@
...
+        {
+            return higherPriority;
+        }
+
+        if (higherPriority.Length == 0)
+        {
+            return lowerPriority;
+        }
+
+        var mergedProbes = new Dictionary<string, T>(StringComparer.Ordinal);
```
andrewlock: nit: Use default comparer where possible
```suggestion
        var mergedProbes = new Dictionary<string, T>();
```

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
```diff
@@ -556,21 +590,21 @@ private ApplyDetails[] AcceptAddedConfiguration(List<RemoteConfiguration>? confi
...
             };
 
             try
             {
-                var updateResults = _configurationUpdater.AcceptAdded(probeConfiguration);
+                var updateResults = _configurationUpdater.AcceptAdded(rcmUpdate);
                 foreach (var updateResult in updateResults)
                 {
-                    var config = configs.FirstOrDefault(c => c.Path.Id == updateResult.Id);
+                    var config = configs.FirstOrDefault(c => ProbeConfigurationUtils.GetProbeIdFromPath(c.Path) == updateResult.Id);
```
andrewlock: `GetProbeIdFromPath` is very allocatey, as it creates a substring. Given you're calling this for every config, in every result, that's a lot of extra strings when you just are looking for the config.

I'd suggest rewriting `GetProbeIdFromPath()` to either return a `ReadOnlySpan<char>` which you can use here for the comparison without allocating, or instead create a `IsProbeId(path, id)` method that does something similar

tracer/src/Datadog.Trace/Debugger/Configurations/ProbeConfigurationFileLoader.cs
```diff
@@ -0,0 +1,191 @@
...
+    private static T[] DeduplicateProbes<T>(List<T> probes)
+        where T : ProbeDefinition
+    {
+        if (probes.Count == 0)
+        {
+            return [];
+        }
+
+        var uniqueProbes = new List<T>(probes.Count);
+        var seenIds = new HashSet<string>(StringComparer.Ordinal);
```
andrewlock:
```suggestion
        var seenIds = new HashSet<string>();
```

tracer/src/Datadog.Trace/Debugger/Configurations/ProbeConfigurationFileLoader.cs
```diff
@@ -0,0 +1,191 @@
...
+                fileContent = await reader.ReadToEndAsync().ConfigureAwait(false);
+            }
+
+            if (string.IsNullOrWhiteSpace(fileContent))
+            {
+                Log.Debug("Probe file is empty: {ProbeFile}", probeFile);
+                return null;
+            }
+
+            var jArray = JsonHelper.ParseJArray(fileContent);
```
andrewlock: This reads everything into a string, and then parses that string: it's better to parse directly rom the file stream, as it avoids the intermediate `string` allocation (which is just thrown away)

tracer/src/Datadog.Trace/Debugger/Configurations/ProbeConfigurationFileLoader.cs
```diff
@@ -0,0 +1,191 @@
...
+            var logs = new List<LogProbe>();
+            var metrics = new List<MetricProbe>();
+            var spans = new List<SpanProbe>();
+            var spanDecorations = new List<SpanDecorationProbe>();
```
andrewlock: nit: may be better to initialize these to null, and then do logs ??= new(); before the `Add()` calls, to reduce some allocation

tracer/src/Datadog.Trace/Debugger/Configurations/ProbeConfigurationFileLoader.cs
```diff
@@ -0,0 +1,191 @@
...
+                    continue;
+                }
+
+                var type = typeToken.ToString();
+                try
+                {
+                    switch (type)
+                    {
+                        case "LOG_PROBE":
+                            var logProbe = jObject.ToObject<LogProbe>();
```
andrewlock: Is there a `JsonHelper` version of this we can use instead, so that we use the array pool? 🤔

tracer/src/Datadog.Trace/Debugger/Configurations/ProbeConfigurationFileLoader.cs
```diff
@@ -0,0 +1,191 @@
...
+            }
+
+            var totalProbes = logs.Count + metrics.Count + spans.Count + spanDecorations.Count;
+            if (totalProbes == 0)
+            {
+                Log.Warning("No valid probes found in file: {ProbeFile}", probeFile);
+                return null;
+            }
+
+            var uniqueLogs = DeduplicateProbes(logs);
```
andrewlock: Why not de-duplicate when you're adding, instead of creating a whole new list? 🤔 I think the behaviour would be the same.

Also, if you don't actually expect that many items in the file, then it may not be worth the overhead of creating a hashset, and using a simple `Contains()` on the `List<>` may be preferable

tracer/src/Datadog.Trace/Debugger/Configurations/ProbeConfigurationUtils.cs
```diff
@@ -0,0 +1,130 @@
...
+    {
+        if (removedProbeIds.Count == 0 && !removeServiceConfiguration)
+        {
+            return configuration;
+        }
+
+        return new ProbeConfiguration
+        {
+            ServiceConfiguration = removeServiceConfiguration ? null : configuration.ServiceConfiguration,
+            LogProbes = configuration.LogProbes.Where(probe => !removedProbeIds.Contains(probe.Id)).ToArray(),
```
andrewlock: This feels very allocate-y, lots of Linq and `ToArray()` calls in many places - is that necessary? 🤔 Wondering if we can either refactor to just mutate lists instead of creating many instances of arrays etc

tracer/src/Datadog.Trace/Debugger/Configurations/ConfigurationUpdater.cs
andrewlock, line 1: I feel like this file should be heavily unit-tested, independent of the "integration" tests in DynamicInstrumentation, as most of the behaviour should be "given inputs, expect these outputs"

tracer/src/Datadog.Trace/Debugger/Configurations/ProbeConfigurationFileLoader.cs
andrewlock, line 1: I feel like this file should be heavily unit-tested, independent of the "integration" tests in DynamicInstrumentation, so that you can test all the edge cases

tracer/src/Datadog.Trace/Debugger/Configurations/ProbeConfigurationUtils.cs
andrewlock, line 1: The methods in this class are another example of things that should be independently unit tested, so that we cover all the edge cases

andrewlock (approved):
Approving for the tracer side - I glanced over some of the debugger stuff, mostly just to point out a few opportunities for perf improvements. One general comment is that I think most of the "helper" classes should all have their own dedicated unit tests for their behaviour, so we can cover all the edge cases, without going through the "full" dynamic instrumentation initialization code

tracer/src/Datadog.Trace/Debugger/Configurations/ConfigurationUpdater.cs
dudikeleti: dca07cf75d8aa42be99374d4028d5593240b3216

tracer/src/Datadog.Trace/Debugger/Configurations/ProbeConfigurationUtils.cs
dudikeleti: dca07cf75d8aa42be99374d4028d5593240b3216

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
dudikeleti: dca07cf75d8aa42be99374d4028d5593240b3216

tracer/src/Datadog.Trace/Debugger/Configurations/ProbeConfigurationFileLoader.cs
dudikeleti: dca07cf75d8aa42be99374d4028d5593240b3216

tracer/src/Datadog.Trace/Debugger/Configurations/ProbeConfigurationFileLoader.cs
dudikeleti: dca07cf75d8aa42be99374d4028d5593240b3216

tracer/src/Datadog.Trace/Debugger/Configurations/ProbeConfigurationFileLoader.cs
dudikeleti: dca07cf75d8aa42be99374d4028d5593240b3216

tracer/src/Datadog.Trace/Debugger/Configurations/ProbeConfigurationFileLoader.cs
dudikeleti: dca07cf75d8aa42be99374d4028d5593240b3216

tracer/src/Datadog.Trace/Debugger/Configurations/ProbeConfigurationUtils.cs
dudikeleti: dca07cf75d8aa42be99374d4028d5593240b3216

tracer/src/Datadog.Trace/Debugger/Configurations/ProbeConfigurationFileLoader.cs
dudikeleti: dca07cf75d8aa42be99374d4028d5593240b3216

tracer/src/Datadog.Trace/Debugger/Configurations/ProbeConfigurationUtils.cs
dudikeleti: dca07cf75d8aa42be99374d4028d5593240b3216

tracer/src/Datadog.Trace/Debugger/Configurations/ConfigurationUpdater.cs
dudikeleti: dca07cf75d8aa42be99374d4028d5593240b3216

tylfin (approved):

PR #7834, opened 2025-11-17
[Debugger] Add memory pressure monitoring telemetry for Dynamic Instrumentation (observe-only)

dudikeleti:
## Summary of changes

- Adds an observe-only memory pressure monitor for Dynamic Instrumentation.
- Samples runtime/system memory pressure around DI activity (no background timer).
- Emits low-cardinality telemetry only when entering or exiting high memory pressure, plus a one-shot counter if the monitor disables itself.
- Does not block, throttle, skip, or otherwise change probe behavior.

## Reason for change

Dynamic Instrumentation can perform memory-sensitive work when capturing variables, creating snapshots, serializing payloads, and enqueueing/uploading data.

Before using memory pressure to affect DI behavior, we need production data showing how often high pressure occurs, what drives it, how severe each signal is at entry, how long episodes last, where the signal is even available, and whether it is stable enough for future throttling decisions. This PR adds exactly that observability and nothing else.

## Implementation details

The monitor detects high pressure from concrete runtime/system signals only:

- Memory load ratio (fraction of available memory in use, so the value is comparable across platforms):
  - On .NET Core 3.1+, uses `GC.GetGCMemoryInfo()` (`MemoryLoadBytes / TotalAvailableMemoryBytes`). This is container/cgroup-aware, and is system/container-wide load rather than process-private bytes - the right scope for "are we close to an OOM".
  - On .NET Framework/Windows, uses `GlobalMemoryStatusEx` (`dwMemoryLoad`, clamped to `[0,1]`). This is machine-wide and **not** container/job-object aware.
  - On runtimes/platforms where neither is available, the memory signal is reported as unsupported.
- Gen2 collections per second:
  - Calculated from `GC.CollectionCount(2)` deltas over elapsed time. Identical and process-specific on every platform.

Sampling is activity-driven rather than always-on. DI calls into the monitor around capture/snapshot/upload-relevant work, and the monitor refreshes only when the last sample is stale (default min interval `1s`). If DI is idle, the monitor does not continuously sample memory.

Time is read from `Environment.TickCount64` on .NET Core 3.0+ (a monotonic `Stopwatch`-based fallback is used on net461/netstandard2.0), so the hot `RefreshIfStale()` fast path is a couple of volatile reads with no clock abstraction.

If neither memory nor GC signals are available on the host, or a provider throws, the monitor **permanently disables itself** (it never samples again), writes a single log line (debug for "no signals", error for an exception), and records a one-shot `disabled` counter tagged by reason so a silent zero in the transition metrics is explainable rather than ambiguous.

Telemetry is emitted on transitions only (enter high pressure / exit high pressure):

- Transition count tagged by `state` (`enter`/`exit`) and `trigger` (`memory`/`gen2`/`both` on enter; `none` on exit) - i.e. which signal drove entry.
- Severity values on transition: memory usage percent and Gen2 collections/sec (tagged by `state`).
- Duration of the high-pressure period, recorded on exit.

No high-cardinality tags such as probe ID, service name, file path, exception type, or method name are added.

```mermaid
flowchart TD
    diStart["DynamicInstrumentation starts"] --> createMonitor["Create MemoryPressureMonitor"]
    createMonitor --> idleMonitor["No background sampling while idle"]

    probeActivity["DI capture/snapshot/upload activity"] --> refreshIfStale["RefreshIfStale (min interval)"]
    refreshIfStale --> signalsAvailable{"Memory or GC signal available?"}
    signalsAvailable -->|"No / provider error"| disable["Disable monitor (logged) + disabled{reason}"]
    signalsAvailable -->|"Yes"| sampleSignals["Sample memory load and Gen2/sec"]
    sampleSignals --> pressureDecision{"High pressure transition?"}

    pressureDecision -->|"ENTER"| enterTelemetry["transitions{state:enter, trigger:*} + severity"]
    pressureDecision -->|"EXIT"| exitTelemetry["transitions{state:exit, trigger:none} + severity + duration"]
    pressureDecision -->|"No transition"| noOp["No telemetry emission"]

    enterTelemetry --> observeOnly["No DI behavior change"]
    exitTelemetry --> observeOnly
    noOp --> observeOnly
    disable --> observeOnly

    diStart --> dispose["DynamicInstrumentation.Dispose"]
    dispose --> stopMonitor["Dispose monitor (lock-free)"]
```

Risk is kept low by avoiding per-second telemetry gauges, background sampling while idle, probe-level cardinality and allocation on hot paths. Refreshes are guarded by a non-blocking single-writer CAS so concurrent DI activity does not run overlapping samples.

## Metrics implementation

All metrics are instrumentation-telemetry metrics, defined as **common** metrics (`isCommon: true`) under the `live_debugger` namespace. They surface in the backend as `dd.instrumentation_telemetry_data.live_debugger.memory_pressure.*`. Tags are fixed, low-cardinality enums.

| Metric | Type | Tags | Emitted |
|---|---|---|---|
| `memory_pressure.transitions` | count | `state` = `enter`\|`exit`, `trigger` = `none`\|`memory`\|`gen2`\|`both` | Once per high-pressure state change. `trigger` is the entry cause on `enter`; `none` on `exit`. |
| `memory_pressure.disabled` | count | `reason` = `no_signals`\|`error` | Once, when the monitor permanently disables itself (no signals available, or a provider threw). |
| `memory_pressure.memory_usage_pct` | count | `state` = `enter`\|`exit`, `bucket` = `lt_70`\|`70_80`\|`80_85`\|`85_90`\|`gte_90` | On each transition - count bucketed by memory load percent at that moment. |
| `memory_pressure.gen2_per_sec` | count | `state` = `enter`\|`exit`, `bucket` = `lt_1`\|`1_2`\|`2_5`\|`gte_5` | On each transition - count bucketed by Gen2 collections/sec at that moment. |
| `memory_pressure.duration_ms` | count | `bucket` = `lt_1s`\|`1_5s`\|`5_30s`\|`gte_30s` | On `exit` - count bucketed by length of the high-pressure episode. |

Runtime/platform segmentation (runtime name, OS, architecture, tracer version) is **not** added as per-metric tags; it comes from the telemetry payload's `application`/`host` metadata and is applied at query time in the backend. This keeps series count bounded while still allowing per-runtime analysis.

The severity and duration metrics use normal telemetry `count` metrics with a fixed `bucket` tag instead of distributions. Distribution telemetry is not supported for this internal telemetry path, and bucketed counts preserve the decision-making signal with bounded cardinality.

The `trigger` tag lives only on the transition count, not on the severity bucket counts: if `trigger:memory`, the memory bucket already captures how severe the memory signal was, and splitting every bucket by trigger would mostly restate the transition breakdown with extra series. The transition count's `trigger` plus the `state`-tagged bucket counts answer "what drove it" and "how severe / how long".

## Dashboards and how we will use the data

The goal is a single dashboard that answers, per runtime/OS, whether DI should ever gate capture under memory pressure and, if so, on which signal and at what threshold.

| Widget | Source metric | Question | Decision it informs |
|---|---|---|---|
| Coverage timeseries / count by `reason` | `memory_pressure.disabled` | Where does the monitor not run at all? | Defines the observed population - distinguishes "no pressure" from "no data". |
| Enter rate timeseries, split by `trigger` | `memory_pressure.transitions{state:enter}` | How often does high pressure happen, and is it memory- or GC-driven? | Whether gating is worth building at all, and which signal to gate on. |
| Trigger breakdown (top-list/pie) | `memory_pressure.transitions{state:enter}` | `memory` vs `gen2` vs `both` mix | Which threshold actually matters. |
| Memory severity bucket breakdown | `memory_pressure.memory_usage_pct{state:enter}` | How deep is memory pressure at entry? | Where to set a memory gating threshold. |
| GC severity bucket breakdown | `memory_pressure.gen2_per_sec{state:enter}` | How hot is GC at entry? | Where to set a gen2 gating threshold. |
| Episode duration bucket breakdown | `memory_pressure.duration_ms` | How long do episodes last? | Whether gating needs hysteresis/cooldown or episodes are too short to act on. |

Every widget can be grouped by the dimensions the telemetry intake stamps onto each series from the payload's `application`/`host` metadata.

How the data is gathered:

1. These metrics flow through the existing tracer telemetry pipeline (generate-metrics payloads) to the Datadog backend - no new transport.
2. Because emission is transition-only and activity-driven, idle apps cost nothing and noisy apps cannot spam per-second points.
3. We collect across runtimes for a few weeks, then read the dashboard top-down: `memory_pressure.disabled` (coverage) → `memory_pressure.transitions{state:enter}` by `trigger` (frequency + cause) → severity distributions (thresholds) → duration (persistence).
4. The resulting numbers define concrete gating thresholds (or show that gating is unnecessary) for a follow-up PR.

### Scope: what this telemetry answers (and what it doesn't)

This is a fleet-aggregate, pre-decision dataset: it tells us whether high memory pressure occurs during DI work often, severely, and long enough to justify gating capture, and on which signal at what threshold.
It is deliberately **not** a tool for measuring the impact of a future gate.

## Test coverage

Added/updated tests cover:

- Memory threshold enter/exit behavior and hysteresis.
- Gen2/sec threshold and rate behavior.
- Consecutive high/low cycle (debounce) behavior.
- Unavailable memory/GC signals → self-disable with `reason:no_signals`, no further sampling.
- Provider exception handling → self-disable with `reason:error`, no crash.
- Refresh overlap handling (non-blocking single writer) and dispose during an in-flight refresh.
- Dispose/lifecycle behavior (no further sampling or emission after dispose).
- Activity-driven stale refresh (no sampling while idle; samples only when stale).
- Transition telemetry: enter emitted once, exit emitted once, severity recorded on transition, duration recorded on exit, no repeat while state is unchanged.
- `trigger` classification: `gen2` when only GC drives entry, `both` when memory and GC cross together, `memory`/`none` on the enter/exit pair.
- Extreme/edge configuration values are clamped to safe minimums.
- Windows `GlobalMemoryStatusEx` interop returns valid values.
- DI integration: the monitor is sampled only through the dedicated hook and is disposed by `DynamicInstrumentation`.
- Telemetry aggregation: transitions tagged by `state`+`trigger`, the `disabled` counter, and the three bucketed count metrics aggregate into the expected series/tags.

## Other details

Initial thresholds are observation thresholds, not throttling thresholds:

- Memory pressure threshold: `0.85`.
- Memory exit threshold with default hysteresis: `0.80`.
- Gen2 threshold: `2` collections/sec.
- Gen2 exit threshold with default hysteresis: `1` collection/sec.

These values are intentionally conservative starting points for data collection and can be tuned after production telemetry is available.

This PR intentionally does not include an explicit/manual pressure event signal. There is no concrete production caller or well-defined meaning for that signal yet. If a future DI path has a specific condition, such as snapshot capture exceeding a memory budget, it should add a named API and metric for that scenario.

Any future DI throttling or blocking based on memory pressure should be done in a separate PR after reviewing the collected telemetry.

## Note
Incidental DynamicInstrumentation lifecycle hardening (not memory-pressure specific): Wiring the new monitor in as a disposable owned by DynamicInstrumentation, exposed pre-existing races in the start/subscribe/dispose paths. DynamicInstrumentation is disposed at runtime via RCM, not just at shutdown. Since the monitor's lifecycle is tied to those paths, we hardened them here.


tracer/src/Datadog.Trace/Debugger/Expressions/ProbeProcessor.cs
```diff
@@ -135,8 +145,13 @@ public bool Process<TCapture>(ref CaptureInfo<TCapture> info, IDebuggerSnapshotC
             }
 
             var probeInfo = state.ProbeInfo;
+            var dynamicInstrumentation = DebuggerManager.Instance.DynamicInstrumentation;
+            if (dynamicInstrumentation is not null && ShouldRefreshMemoryPressureBeforeCapture(info.MethodState))
+            {
+                dynamicInstrumentation.RefreshMemoryPressureIfStale();
+            }
```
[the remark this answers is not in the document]
dudikeleti: DebuggerManager.DynamicInstrumentation only returns a non-null instance when DI is not disposed and IsInitialized == true; otherwise the getter returns null. Because the refresh uses that getter result, RefreshMemoryPressureIfStale() cannot run before DI initialization. The following IsInitialized == false check is pre-existing/redundant, but moving the refresh is not needed for correctness.

tracer/src/Datadog.Trace/Debugger/RateLimiting/SystemMemorySource.cs
```diff
@@ -0,0 +1,75 @@
...
+                {
+                    return true;
+                }
+
+                Log.Debug("GlobalMemoryStatusEx did not return memory information. Treating memory usage ratio as unsupported on this platform.");
+            }
+            catch (Exception ex)
+            {
+                Log.Error(ex, "Failed to get system memory status");
+            }
```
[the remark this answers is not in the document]
dudikeleti: 6b8f06cb75f679b0f13b9d6fcf55583906dc8f47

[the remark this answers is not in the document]
dudikeleti: 6b8f06cb75f679b0f13b9d6fcf55583906dc8f47

tracer/src/Datadog.Trace/Debugger/RateLimiting/MemoryPressureMonitor.cs
```diff
@@ -0,0 +1,436 @@
...
+                }
+                else
+                {
+                    durationMs = _hasHighPressureStart ? nowMs - _highPressureStartMs : 0;
+                    _highPressureStartMs = 0;
+                    _hasHighPressureStart = false;
+                }
+
+                // Telemetry runs outside the CAS region intentionally; it is the only observable side effect
+                // and may race a concurrent Dispose() with no ill effect (observational only).
```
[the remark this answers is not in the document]
dudikeleti: 6b8f06cb75f679b0f13b9d6fcf55583906dc8f47

tracer/test/Datadog.Trace.Tests/Debugger/RateLimiting/MemoryPressureMonitorTests.cs
```diff
@@ -0,0 +1,580 @@
...
+                {
+                    ratio = 0.10;
+                    return true;
+                }
+
+                _blocked.Set();
+                _release.Wait();
+                ratio = _blockedRatio;
+                return true;
+            }
```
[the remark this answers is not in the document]
dudikeleti: a652bda3de7435e73779d98457ac2c57a7c75096

tracer/src/Datadog.Trace/Debugger/RateLimiting/MemoryPressureMonitor.cs
```diff
@@ -0,0 +1,436 @@
...
+                        Log.Debug("MemoryPressureMonitor disabled: no memory or GC info available on this runtime/platform.");
+                    }
+
+                    return;
+                }
+
+                // Establish Gen2 baseline if needed.
+                if (!_hasGen2Baseline && gcAvailable)
+                {
+                    _lastRefreshMs = nowMs;
```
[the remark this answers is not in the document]
dudikeleti: a652bda3de7435e73779d98457ac2c57a7c75096

tracer/src/Datadog.Trace/Debugger/RateLimiting/MemoryPressureMonitor.cs
```diff
@@ -0,0 +1,440 @@
...
+using System.Runtime.CompilerServices;
+using System.Threading;
+using Datadog.Trace.Logging;
+using Datadog.Trace.Telemetry;
+using Datadog.Trace.Telemetry.Metrics;
+
+namespace Datadog.Trace.Debugger.RateLimiting
+{
+    /// <summary>Reads the current process memory-load ratio (0-1+). Returns false when unsupported.</summary>
+    internal delegate bool TryReadMemoryUsageRatio(out double ratio);
```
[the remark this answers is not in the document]
dudikeleti: f639b663b7eae31d93f3b042070ac2613df6d347

tracer/test/Datadog.Trace.Tests/Debugger/RateLimiting/MemoryPressureMonitorTests.cs
```diff
@@ -0,0 +1,582 @@
...
+        public void WindowsMemoryInfo_TryGetMemoryLoadRatio_ReturnsValidValue()
+        {
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                return;
+            }
+
+            WindowsMemoryInfo.TryGetMemoryLoadRatio(out var ratio).Should().BeTrue();
+            ratio.Should().BeGreaterThan(0);
+            ratio.Should().BeLessOrEqualTo(1);
```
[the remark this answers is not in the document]
dudikeleti: c00a1e568d608e9186e43da31ee8192fa4992958

tracer/src/Datadog.Trace/Telemetry/Metrics/Count.cs
```diff
@@ -206,6 +206,18 @@ internal enum Count
     [TelemetryMetric<MetricTags.ApiError>("direct_log_api.errors", isCommon: false)] DirectLogApiErrors,
 
 #endregion
+#region Live Debugger Namespace
+
+    /// <summary>
+    /// The number of Dynamic Instrumentation memory-pressure state transitions, tagged by state and the signal that triggered entry.
+    /// </summary>
+    [TelemetryMetric<MetricTags.DebuggerMemoryPressureState, MetricTags.DebuggerMemoryPressureTrigger>("memory_pressure.transitions", isCommon: true, NS.LiveDebugger)] DebuggerMemoryPressureTransitions,
```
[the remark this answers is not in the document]
dudikeleti: The PR description has been updated to match the implementation: these are common metrics under the live_debugger namespace, emitted as memory_pressure.*

tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
```diff
@@ -184,16 +188,55 @@ private async Task InitializeAsync()
...
+                // Start the runtime exactly once.
+                if (!IsInitialized)
+                {
+                    var assemblyLoadSubscribed = false;
+                    try
+                    {
+                        AppDomain.CurrentDomain.AssemblyLoad += CheckUnboundProbes;
+                        assemblyLoadSubscribed = true;
+                        StartBackgroundProcess();
+                        Volatile.Write(ref _initializationState, 2);
```
jpbempel, line 219: nit: make it a named constant

tracer/src/Datadog.Trace/Debugger/RateLimiting/MemoryPressureMonitor.cs
```diff
@@ -0,0 +1,440 @@
...
+        private readonly int _consecutiveLowToExit;
+        private readonly TryReadMemoryUsageRatio _tryReadMemoryRatio;
+        private readonly TryReadGen2CollectionCount _tryReadGen2;
+        private readonly MemoryPressureTransitionHandler _onTransition;
+        private readonly Action<MetricTags.DebuggerMemoryPressureDisabledReason> _onDisabled;
+
+        // Refresh writes these fields through a single CAS-guarded writer; every cross-thread access
+        // goes through Volatile.Read/Volatile.Write (and Interlocked where a read-modify-write is needed).
+        // We deliberately use the Volatile.* helpers rather than the `volatile` keyword so the idiom is
+        // uniform across the int/long fields too (`long` cannot be marked `volatile`).
```
jpbempel, line 69: TIL: cannot mark a long field volatile! Now I understand why you are using it so scarcely while in Java we are using it more often :/

jpbempel: that being said, why not using a sub object with all stats and using a volatile ref to update in one shot for the refresh?
That way you have only one place to manage the concurency.

example:
```
class stats 
{
        private long _lastGen2Count;
        private long _lastRefreshMs;
        private long _highPressureStartMs;
        private bool _hasHighPressureStart;
        private int _highStreak;
        private int _lowStreak;
        private bool _hasGen2Baseline;
}

void refresh() 
{
  Stats stats = new stats();
  stats._lastGen2Count = ...;
  // ...
  stats._hasGen2Baseline = ...;
  currentStats = stats; // currentStats is volatile field, assigned 
}
```
then the reader just need to Volatile read `currentStats` to snapshot the values then use them locally

jpbempel: But yes the downside of this is one alloc for stats at each call of refresh.

tracer/src/Datadog.Trace/Debugger/RateLimiting/MemoryPressureMonitor.cs
```diff
@@ -0,0 +1,440 @@
...
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private bool IsStale(long nowMs, long lastRefreshMs)
+        {
+            return (nowMs - lastRefreshMs) >= _staleThresholdMs;
+        }
+
+        // Only ever called from inside the CAS-guarded Refresh, so no synchronization is required.
+        private bool DisableCore(MetricTags.DebuggerMemoryPressureDisabledReason reason)
```
jpbempel: For my knowledge as probably I missed something: if it is only used in `Refresh` method why not put it like `ComputeNextHigh` or `ToScaledInt` as "inner method" of `Referesh`?

jpbempel (approved):

andrewlock (changes requested):
As discussed, I have concerns that these metrics won't _actually_ be usable, because we don't tag them with service identity. So you have no way to compare between enter/exit conditions? 🤔 However, I now realise that you're relying on distributions and going to try to do some statistical comparison between them? I'm not sure..

However, the count metrics are fine in general, but we literally don't support distribution metrics at the moment (intentionally, because they have perf problems), so we'll need to remove those.

tracer/src/Datadog.Trace/Debugger/RateLimiting/MemoryPressureMonitor.cs
andrewlock: You _can_ use `Interlocked.Exchange(ref _long)` though, which is generally preferable anyway IMO, but snapshotting is probably still the better design IMO.

In terms of allocations, you can use a "front and back buffer object with atomic exchange" - that way you create 2 objects for the life of the process 🙂

dudikeleti:
> As discussed, I have concerns that these metrics won't _actually_ be usable, because we don't tag them with service identity. So you have no way to compare between enter/exit conditions? 🤔 However, I now realise that you're relying on distributions and going to try to do some statistical comparison between them? I'm not sure..
> 
> However, the count metrics are fine in general, but we literally don't support distribution metrics at the moment (intentionally, because they have perf problems), so we'll need to remove those.

The missing service tag isn't a blocker. We dont have to pair a given app's enter with its own exit. We want to be able to say:
"Memory pressure during DI is common, driven mainly by memory (not GC), clusters at 90%+, and episodes last tens of seconds - so gating is worth building, gate on memory at ~0.90 with a short cooldown." Or the opposite: "episodes are sub-second and rare - don't bother.

tracer/src/Datadog.Trace/Debugger/RateLimiting/MemoryPressureMonitor.cs
dudikeleti: > You can use Interlocked.Exchange(ref _long) though

Yes but for this case volatile is enough imo

> but snapshotting is probably still the better design IMO.
In terms of allocations, you can use a "front and back buffer object with atomic exchange" - that way you create 2 objects for the life of the process 🙂

I like it. Currently, no one really uses it in production, but I would definitely do that in the future if these values become part of a decision-making process. Noted.

dudikeleti: On second thought, I need to get back to this PR anyway because of the distribution metrics limitation, so I might handle this while I'm at it.

tracer/src/Datadog.Trace/Debugger/RateLimiting/MemoryPressureMonitor.cs
dudikeleti: The reasoning behind that is that, unlike `ComputeNextHigh` and `ToScaledInt`, which are purely local helpers and should remain local functions, `DisableCore` has meaning beyond the refresh algorithm, and I haven't fully settled on its future shape yet (since this PR is observe-only).

But you didn't miss anything. For this PR, it is currently only called from `Refresh`, as you noted.

Done 5c99511f77465eead2a5e453e1e7730d90682864

tracer/src/Datadog.Trace/Debugger/RateLimiting/MemoryPressureMonitor.cs
dudikeleti: On third thought 😆

I tried the two-buffer snapshot approach locally, but I’m leaning toward keeping the current simpler publication model for now.

It does give a more coherent internal snapshot, but it also adds quite a bit of complexity/versioning, more synchronization.

I think it’s worth revisiting if/when this becomes decision-making state or we expose a single snapshot API, but for the current observe-only telemetry use, I’d prefer to keep the simpler volatile fields.

dudikeleti:
> However, the count metrics are fine in general, but we literally don't support distribution metrics at the moment (intentionally, because they have perf problems), so we'll need to remove those.

I changed the memory-pressure severity/duration metrics from distributions to bucketed count metrics.

tracer/src/Datadog.Trace/Telemetry/Metrics/DistributionShared.cs
```diff
@@ -6,6 +6,7 @@
 #nullable enable
 using System.Diagnostics.CodeAnalysis;
 using Datadog.Trace.SourceGenerators;
+using NS = Datadog.Trace.Telemetry.MetricNamespaceConstants;
```
andrewlock:
```suggestion
```

tracer/test/Datadog.Trace.Tests/Telemetry/Metrics/common_metrics.json
```diff
@@ -1,4 +1,59 @@
 ﻿{
+  "live_debugger": {
+    "memory_pressure.transitions": {
+      "tags": [
+        "state",
+        "trigger"
+      ],
+      "metric_type": "count",
+      "data_type": "transitions",
+      "description": "The number of Dynamic Instrumentation memory-pressure state transitions, tagged by state and the signal that triggered entry",
```
andrewlock: It's best to include the _actual_ state tags allowed here, e.g. 

```suggestion
      "description": "The number of Dynamic Instrumentation memory-pressure state transitions, tagged by state (`state:enter` or `state:exit`) and the signal that triggered entry (`trigger:none`, `trigger:memory`, `trigger:gen2`, `trigger:both`)",
```

The same applies to all the other definitions here.

Additionally, these metrics need to be added to [the telemetry intake](https://github.com/DataDog/dd-go/blob/prod/trace/apps/tracer-telemetry-intake/telemetry-metrics/static/common_metrics.json), and deployed, otherwise they will be blocked. These metrics are common across _all_ languages too, just FYI, so we should try to keep the tags applicable to a wide range of languages if possible (e.g. gen2 is very .NET specific, maybe we can use a more generic tag name?)

tracer/src/Datadog.Trace/Telemetry/Metrics/Count.cs
```diff
@@ -206,6 +206,33 @@ internal enum Count
...
+
+    /// <summary>
+    /// Count of Dynamic Instrumentation memory-pressure transitions bucketed by Gen2 collections per second at the transition.
+    /// </summary>
+    [TelemetryMetric<MetricTags.DebuggerMemoryPressureState, MetricTags.DebuggerMemoryPressureGen2Bucket>("memory_pressure.gen2_per_sec", isCommon: true, NS.LiveDebugger)] DebuggerMemoryPressureGen2PerSec,
+
+    /// <summary>
+    /// Count of Dynamic Instrumentation high-memory-pressure periods bucketed by duration, recorded on exit.
+    /// </summary>
+    [TelemetryMetric<MetricTags.DebuggerMemoryPressureDurationBucket>("memory_pressure.duration_ms", isCommon: true, NS.LiveDebugger)] DebuggerMemoryPressureDurationMs,
```
andrewlock: Is this _really_ what you want? The _total_ duration of high-memory pressure periods? Maybe it is, just feels kind of strange to me 😅

tracer/src/Datadog.Trace/Telemetry/Metrics/Count.cs
```diff
@@ -206,6 +206,33 @@ internal enum Count
...
+
+    /// <summary>
+    /// Count of Dynamic Instrumentation memory-pressure transitions bucketed by memory load percentage at the transition.
+    /// </summary>
+    [TelemetryMetric<MetricTags.DebuggerMemoryPressureState, MetricTags.DebuggerMemoryPressureMemoryBucket>("memory_pressure.memory_usage_pct", isCommon: true, NS.LiveDebugger)] DebuggerMemoryPressureMemoryUsagePct,
+
+    /// <summary>
+    /// Count of Dynamic Instrumentation memory-pressure transitions bucketed by Gen2 collections per second at the transition.
+    /// </summary>
+    [TelemetryMetric<MetricTags.DebuggerMemoryPressureState, MetricTags.DebuggerMemoryPressureGen2Bucket>("memory_pressure.gen2_per_sec", isCommon: true, NS.LiveDebugger)] DebuggerMemoryPressureGen2PerSec,
```
andrewlock: It seems a bit weird to have a count of a rate... I guess this is just trying to hack around current lack of distribution support? 🤔

tracer/src/Datadog.Trace/Telemetry/Metrics/MetricTags.cs
```diff
@@ -188,6 +188,51 @@ internal enum InstrumentationComponent
...
+    {
+        [Description("reason:no_signals")] NoSignals,
+        [Description("reason:error")] Error,
+    }
+
+    internal enum DebuggerMemoryPressureMemoryBucket
+    {
+        [Description("bucket:lt_70")] LessThan70,
+        [Description("bucket:70_80")] From70To80,
+        [Description("bucket:80_85")] From80To85,
```
andrewlock, line 215: What if it's 80%? Which bucket does it go in? Might be preferable to make that clear

andrewlock (approved):
I haven't reviewed the Debugger code, just the shared metrics code. The important thing is you'll need to merge your changes into the telemetry intake before these new metrics will be accepted.

tracer/src/Datadog.Trace/Telemetry/Metrics/MetricTags.cs
dudikeleti: 86f6bb3cd59a512bd61d8c4fe63ffa9e4d6f31ae

tracer/test/Datadog.Trace.Tests/Telemetry/Metrics/common_metrics.json
dudikeleti: 86f6bb3cd59a512bd61d8c4fe63ffa9e4d6f31ae

tracer/src/Datadog.Trace/Telemetry/Metrics/DistributionShared.cs
dudikeleti: 86f6bb3cd59a512bd61d8c4fe63ffa9e4d6f31ae

PR #7835, opened 2025-11-17
[AWS Lambda] Add request-id as header to Lambda start/end invocation

rithikanarayan:
## Summary of changes
Adds` lambda-runtime-aws-request-id` header in start invocation and end invocation requests from tracer to extension. Assigns header to the aws request id of the span.

## Reason for change
Allows the lambda extension to identify which start/end invocation requests belong to which invocation if multiple Lambda requests are being served concurrently.

## Implementation details
Uses the LambdaContext from the `IInvocationRequest proxyinstance` to get the aws-request-id in `SendStartInvocation` and adds it to request headers in `WriteRequestHeaders`. Saves the aws-request-id as part of the state passed from `OnDelegateBegin `to `EndInvocationAsync`. `EndInvocationAsync` passes the state down to `GetEndInvocationRequest`, which adds it to the request headers. 

Updates the LambdaCommonTests and LambdaRequestBuilderTests to match new function signatures with state passed down and to test for the assignment to the lambda-runtime-aws-request-id. 

## Test coverage
Added unit test in LambdaRequestBuilderTests to check for ` lambda-runtime-aws-request-id` header and value in end invocation request. Manually tested for headers in start and end invocation using custom dd-trace-dotnet layer `arn:aws:lambda:sa-east-1:425362996713:layer:dotnet-request-id-header-rithika:2` and custom lambda extension `arn:aws:lambda:sa-east-1:425362996713:layer:Debug-extension-rithika:1` which prints the start and end invocation headers. Sample trace with logs of the headers [here](https://support-admin.us1.prod.dog/admin/switch_handle_get/org_id/1543931?next_url=%2Fapm%2Ftrace%2F219302768028734365%3FgraphType%3Dservice_map%26shouldShowLegend%3Dtrue%26spanID%3D4621724130062486128%26timeHint%3D1763567215307.0002%26trace%3D2193027680287343654621724130062486128%26traceQuery%3D)

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Lambda/LambdaCommon.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Lambda/LambdaCommon.cs
@@ -48,7 +48,7 @@ internal abstract class LambdaCommon
         WriteRequestHeaders(request, context?.ClientContext?.Custom);
         if (context?.AwsRequestId != null)
         {
-            request.Headers.Add(LambdaRuntimeAwsRequestIdHeader, context?.AwsRequestId);
+            request.Headers.Add(LambdaRuntimeAwsRequestIdHeader, context.AwsRequestId);
         }
 
         using var response = (HttpWebResponse)request.GetResponse();


lucaspimentel:
@codex review

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Lambda/HandlerWrapperSetHandlerIntegration.cs
```diff
@@ -80,11 +81,12 @@ public object OnDelegateBegin<TArg1>(object sender, ref TArg1 arg)
             else
             {
                 var jsonString = ConvertPayloadStream(proxyInstance.InputStream);
-                scope = LambdaCommon.SendStartInvocation(new LambdaRequestBuilder(), jsonString, proxyInstance.LambdaContext?.ClientContext?.Custom);
+                scope = LambdaCommon.SendStartInvocation(new LambdaRequestBuilder(), jsonString, proxyInstance.LambdaContext);
+                state = proxyInstance.LambdaContext.AwsRequestId;
```
lucaspimentel: I see `LambdaCommon.SendStartInvocation()` checks if the `proxyInstance.LambdaContext` value we pass above is null. We should handle that here as well.

```suggestion
                state = proxyInstance.LambdaContext?.AwsRequestId;
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Lambda/HandlerWrapperSetHandlerIntegration.cs
```diff
@@ -108,18 +110,18 @@ public async Task<TInnerReturn> OnDelegateEndAsync<TInnerReturn>(object sender,
                 if (proxyInstance == null)
                 {
                     LambdaCommon.Log("DuckCast.IInvocationResponse got null proxyInstance", debug: false);
-                    await LambdaCommon.EndInvocationAsync(string.Empty, exception, ((CallTargetState)state!).Scope, RequestBuilder).ConfigureAwait(false);
+                    await LambdaCommon.EndInvocationAsync(string.Empty, exception, ((CallTargetState)state!).Scope, ((CallTargetState)state!).State, RequestBuilder).ConfigureAwait(false);
```
lucaspimentel: To minimize all the casting (`((CallTargetState)state!)`) in this method, you can do it just once. Maybe also rename the method parameter from `state` to `stateObject` so we can use `state` below:
```csharp
public async Task<TInnerReturn> OnDelegateEndAsync<TInnerReturn>(object sender, TInnerReturn returnValue, Exception exception, object stateObject)
{
    var state = (CallTargetState)state!;
    ...
```

Then you can simply use `state.State` here and `state.Scope` below.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/SDK/ILambdaContext.cs
```diff
@@ -17,7 +17,7 @@ internal interface ILambdaContext
     // /// This is the same ID returned to the client that called invoke().
     // /// This ID is reused for retries on the same request.
     // /// </summary>
```
lucaspimentel: You can uncomment these lines now (remove the first `//`). They must have been commented with the `AwsRequestId` below.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Lambda/HandlerWrapperSetHandlerIntegration.cs
```diff
@@ -71,6 +71,7 @@ public object OnDelegateBegin<TArg1>(object sender, ref TArg1 arg)
             LambdaCommon.Log("DelegateWrapper Running OnDelegateBegin");
 
             Scope scope;
+            object state = null;
```
duncanista: let's rename this state to `requestId`

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Lambda/LambdaCommon.cs
```diff
@@ -38,7 +41,7 @@ internal static Scope CreatePlaceholderScope(Tracer tracer, NameValueHeadersColl
         return tracer.TracerManager.ScopeManager.Activate(span, false);
     }
 
-    internal static Scope SendStartInvocation(ILambdaExtensionRequest requestBuilder, string data, IDictionary<string, string> context)
+    internal static Scope SendStartInvocation(ILambdaExtensionRequest requestBuilder, string data, ILambdaContext context)
```
duncanista, line 44: You should update line 48, as the context written is not the `ILambdaContext` but the `.LambdaContext?.ClientContext?.Custom`

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Lambda/LambdaCommon.cs
```diff
@@ -22,6 +24,7 @@ internal abstract class LambdaCommon
     private const string PlaceholderOperationName = "placeholder-operation";
     private const double ServerlessMaxWaitingFlushTime = 3;
     private const string LogLevelEnvName = "DD_LOG_LEVEL";
+    private const string LambdaRuntimeAwsRequestHeaderId = "lambda-runtime-aws-request-id";
```
duncanista:
```suggestion
    private const string LambdaRuntimeAwsRequestIdHeader = "lambda-runtime-aws-request-id";
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Lambda/LambdaCommon.cs
```diff
@@ -116,13 +121,22 @@ private static void WriteRequestPayload(WebRequest request, string data)
...
-                request.Headers.Add(kv.Key, kv.Value);
+                foreach (var kv in clientContext)
+                {
+                    request.Headers.Add(kv.Key, kv.Value);
+                }
+            }
+
+            if (context.AwsRequestId != null)
+            {
+                request.Headers.Add(LambdaRuntimeAwsRequestHeaderId, context.AwsRequestId);
```
duncanista: Not sure if we should update this here, seems unnecessary? we could just do the same as GetEndInvocationRequest where you add the requestId on the method?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Lambda/LambdaCommon.cs
```diff
@@ -38,11 +41,16 @@ internal static Scope CreatePlaceholderScope(Tracer tracer, NameValueHeadersColl
...
-    internal static Scope SendStartInvocation(ILambdaExtensionRequest requestBuilder, string data, IDictionary<string, string> context)
+    internal static Scope SendStartInvocation(ILambdaExtensionRequest requestBuilder, string data, ILambdaContext context)
     {
         var request = requestBuilder.GetStartInvocationRequest();
         WriteRequestPayload(request, data);
-        WriteRequestHeaders(request, context);
+        WriteRequestHeaders(request, context?.ClientContext?.Custom);
+        if (context?.AwsRequestId != null)
+        {
+            request.Headers.Add(LambdaRuntimeAwsRequestIdHeader, context?.AwsRequestId);
```
lucaspimentel: Not a big deal, but already know `context` is not null inside the `if`, so we can skip the extra check.
```suggestion
            request.Headers.Add(LambdaRuntimeAwsRequestIdHeader, context.AwsRequestId);
```

lucaspimentel (approved):
I'll add some non-blocking comments for things we can clean up later in a separate PR.

duncanista (approved):

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Lambda/LambdaRequestBuilder.cs
```diff
@@ -35,12 +35,17 @@ WebRequest ILambdaExtensionRequest.GetStartInvocationRequest()
...
+    WebRequest ILambdaExtensionRequest.GetEndInvocationRequest(Scope scope, object state, bool isError)
     {
         var request = WebRequest.Create(Uri + EndInvocationPath);
         request.Method = "POST";
         request.Headers.Set(HttpHeaderNames.TracingEnabled, "false");
 
+        if (state != null)
+        {
+            request.Headers.Set("lambda-runtime-aws-request-id", (string)state);
+        }
```
lucaspimentel, line 47: Although it should never happen, if `state` is ever a different type from `string`, this casting operation will throw an exception. It would be safer to check that `state` is a `string`, not simply not null

The more modern way of doing this in C# is by combining the type check with a `string` variable declaration and removing the `(string)` casting operator:
```csharp
if (state is string requestId)
{
    request.Headers.Set("lambda-runtime-aws-request-id", requestId);
}

File: tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Lambda/LambdaCommon.cs [hunk 10]
```diff
@@ -55,9 +63,9 @@ internal static Scope SendStartInvocation(ILambdaExtensionRequest requestBuilder
         return CreatePlaceholderScope(tracer, headers);
     }
 
-    internal static void SendEndInvocation(ILambdaExtensionRequest requestBuilder, Scope scope, bool isError, string data)
+    internal static void SendEndInvocation(ILambdaExtensionRequest requestBuilder, Scope scope, object state, bool isError, string data)
```
lucaspimentel, line 66: In general we should try to avoid using the `object` type when passing arguments around. C# developers like the strong type safety that the language provides and treating anything as `object` disable that language feature.

To avoid that, we should cast the `state` to `CallTargetState` as early as possible and pass it around as a `CallTargetState`, not as `object`.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Lambda/LambdaCommon.cs
```diff
@@ -38,11 +41,16 @@ internal static Scope CreatePlaceholderScope(Tracer tracer, NameValueHeadersColl
         return tracer.TracerManager.ScopeManager.Activate(span, false);
     }
 
-    internal static Scope SendStartInvocation(ILambdaExtensionRequest requestBuilder, string data, IDictionary<string, string> context)
+    internal static Scope SendStartInvocation(ILambdaExtensionRequest requestBuilder, string data, ILambdaContext context)
     {
         var request = requestBuilder.GetStartInvocationRequest();
         WriteRequestPayload(request, data);
-        WriteRequestHeaders(request, context);
+        WriteRequestHeaders(request, context?.ClientContext?.Custom);
```
lucaspimentel, line 48: If either `context` or `context.ClientContext` or `context.ClientContext.Custom` are null, we will effectively call `WriteRequestHeaders(request, null)`.

Does `WriteRequestHeaders()` handle null gracefully? Or should we check for null before calling?
```csharp
if (context?.ClientContext?.Custom is { } c)
{
    WriteRequestHeaders(request, c);
}
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Lambda/LambdaRequestBuilder.cs
```diff
@@ -35,12 +35,17 @@ WebRequest ILambdaExtensionRequest.GetStartInvocationRequest()
         return request;
     }
 
-    WebRequest ILambdaExtensionRequest.GetEndInvocationRequest(Scope scope, bool isError)
+    WebRequest ILambdaExtensionRequest.GetEndInvocationRequest(Scope scope, object state, bool isError)
```
lucaspimentel, line 38: Here's a good example of why not to not use `object` in method signature if we can avoid it: in some methods `state` is a `CallTargetState`, but here it's a `string`. It's easy to get them confused and try to cast to the wrong type, which will throw an exception at run time.

Or if someone mistakenly checks `if (state is CallTargetState)` before casting, it will always be false (because this one is a `string`) and we'll never hit the code path below that adds the request header.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Lambda/LambdaCommon.cs
rithikanarayan: WriteRequestHeaders starts by checking if context is null (if it is null it does nothing). This was how it was structured prior to my changes, so I am choosing to leave it that way.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Lambda/HandlerWrapperSetHandlerIntegration.cs
```diff
@@ -71,6 +71,7 @@ public object OnDelegateBegin<TArg1>(object sender, ref TArg1 arg)
             LambdaCommon.Log("DelegateWrapper Running OnDelegateBegin");
 
             Scope scope;
+            object requestid = null;
```
lucaspimentel, line 74: [naming conventions]
We should use `requestId` instead of `requestid` (this is the nit-pickiest comment in this PR, I promise)

[type safety]
`requestid` can be `string` since `ILambdaContext.AwsRequestId` is declared as string. We want to avoid using `object` as much as possible.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Lambda/LambdaCommon.cs
```diff
@@ -67,8 +75,10 @@ internal static void SendEndInvocation(ILambdaExtensionRequest requestBuilder, S
         }
     }
 
-    internal static async Task EndInvocationAsync(string returnValue, Exception exception, Scope scope, ILambdaExtensionRequest requestBuilder)
+    internal static async Task EndInvocationAsync(string returnValue, Exception exception, object stateObject, ILambdaExtensionRequest requestBuilder)
```
lucaspimentel, line 78: Same comment here about the using `object` as parameter type.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Lambda/LambdaCommon.cs
```diff
@@ -67,8 +75,10 @@ internal static void SendEndInvocation(ILambdaExtensionRequest requestBuilder, S
...
     {
+        var state = (CallTargetState)stateObject!;
```
lucaspimentel, line 80: This is a dangerous cast that will throw an exception if `stateObject` is null or otherwise not a `CallTargetState`. Since `stateObject` is declared as `object`, it could be _anything_, so we should always be defensive and check first:
```csharp
if (state is CallTargetState)
{
    ...
```

Even better would be to use `CallTargetState` as the parameter type instead of `object`. Then the type is guaranteed and we won't need to either check or cast here.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Lambda/LambdaCommon.cs
lucaspimentel: That works, thanks!

PR #7836, opened 2025-11-17
Update .gitlab/one-pipeline.locked.yaml

pawelchcki:
Automatic PR. Updates .gitlab/one-pipeline.locked.yaml


mhlidd:
> Just going to block this until we know what it does - these updates are dangerous without knowing the implications

@andrewlock Sorry, I should have added more info to the PR description (auto-generated PR). I added 2 CI Checks for Config Registry to support a new format for the [supported-configurations.json](https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/src/Datadog.Trace/Configuration/supported-configurations.json) file ([ref1](https://github.com/DataDog/libdatadog-build/pull/151), [ref2](https://github.com/DataDog/libdatadog-build/pull/152)). I manually triggered the job to update the hash in our dd-trace-* libraries to allow tracers to migrate to using the new CI checks and new format for the supported-configurations.json file.

anna-git:
> Thanks for the extra details @mhlidd 🙂 @anna-git you might want to look into this seeing as it's about your config work!

Yes it seems all good, as long as the version is after the one I had changed to, and I see the config json validation step is still here :)

PR #7837, opened 2025-11-18
[Tracer] Add Cosmos DB CRUD instrumentation

pablomartinezbernardo:
## Summary of changes

Trace CRUD operations in addition to already traced queries. With additional tags:
* Response status codes
* Connection mode
* User Agent

## Other details

`MinimumVersion = "3.12.0"` as that is the lowest we can test with the emulator


pablomartinezbernardo:
Note for @bouwkast : This will likely modify the snapshots made with the manual win emulator tests

lucaspimentel:
@codex review

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/CosmosDb/RequestInvokerHandlerSendAsyncIntegration.cs
```diff
@@ -0,0 +1,139 @@
...
+    {
+        var scope = state.Scope;
+
+        if (scope != null && returnValue.Instance is not null)
+        {
+            try
+            {
+                var tags = (CosmosDbTags)scope.Span.Tags;
+                var statusCode = (int)returnValue.StatusCode;
+                tags.ResponseStatusCode = statusCode.ToString();
```
lucaspimentel:
```suggestion
                tags.ResponseStatusCode = statusCode.ToString(CultureInfo.InvariantCulture);
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/CosmosDb/RequestInvokerHandlerSendAsyncIntegration.cs
```diff
@@ -0,0 +1,139 @@
...
+    TypeName = "Microsoft.Azure.Cosmos.Handlers.RequestInvokerHandler",
+    MethodName = "SendAsync",
+    ReturnTypeName = "System.Threading.Tasks.Task`1[Microsoft.Azure.Cosmos.ResponseMessage]",
+    ParameterTypeNames = [ClrNames.String, "Microsoft.Azure.Documents.ResourceType", "Microsoft.Azure.Documents.OperationType", "Microsoft.Azure.Cosmos.RequestOptions", "Microsoft.Azure.Cosmos.ContainerInternal", ClrNames.Ignore, ClrNames.Stream, "System.Action`1[Microsoft.Azure.Cosmos.RequestMessage]", ClrNames.Ignore, ClrNames.CancellationToken],
+    MinimumVersion = "3.12.0",
+    MaximumVersion = "3.*.*",
+    IntegrationName = nameof(IntegrationId.CosmosDb))]
+[Browsable(false)]
+[EditorBrowsable(EditorBrowsableState.Never)]
+public class RequestInvokerHandlerSendAsyncIntegration
```
lucaspimentel: See #7879
```suggestion
public sealed class RequestInvokerHandlerSendAsyncIntegration
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/CosmosDb/RequestInvokerHandlerSendAsyncIntegration.cs
```diff
@@ -0,0 +1,139 @@
...
+                return CallTargetState.GetDefault();
+            }
+
+            string? containerId = null;
+            string? databaseId = null;
+
+            if (cosmosContainerCore?.Instance != null)
+            {
+                containerId = cosmosContainerCore.Id;
+                databaseId = cosmosContainerCore.Database.Id;
```
lucaspimentel: We check if `cosmosContainerCore` is `null` above, but is `cosmosContainerCore.Database` guaranteed to not be `null`?

```suggestion
                databaseId = cosmosContainerCore.Database?.Id;
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/CosmosDb/IResponseMessage.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// Microsoft.Azure.Cosmos.ResponseMessage interface for duck typing
+    /// </summary>
+    /// <remarks>
+    /// https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/Handler/ResponseMessage.cs
+    /// </remarks>
+    internal interface IResponseMessage : IDuckType
+    {
+        HttpStatusCode StatusCode { get; }
+
+        IResponseHeaders Headers { get; }
```
lucaspimentel: We check for null `Headers` in `RequestInvokerHandlerSendAsyncIntegration:121`
```csharp
if (returnValue.Headers?.Instance != null)
```

If it's nullable, we should make that explicit:
```suggestion
        IResponseHeaders? Headers { get; }
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/CosmosDb/RequestInvokerHandlerSendAsyncIntegration.cs
```diff
@@ -0,0 +1,139 @@
...
+            tags.DatabaseId = databaseId;
+            tags.SetEndpoint(instance?.Client.Endpoint);
+
+            var clientContextObject = instance?.Client.ClientContext;
+            if (clientContextObject != null && clientContextObject.TryDuckCast<CosmosContextClientStruct>(out var clientContext))
+            {
+                tags.UserAgent = clientContext.UserAgent;
+
+                var connectionMode = clientContext.ClientOptions.ConnectionMode;
+                tags.ConnectionMode = connectionMode == 0 ? "gateway" : "direct";
```
lucaspimentel: Currently, `0  == gateway` and `1 == direct`. To be safe (and for future-proofing), should we use something like `"other"` if we get any other value?

```
tags.ConnectionMode = connectionMode switch
{
    0 => "gateway"
    1 => "direct"
    _ => "other"
};
```

ref: https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/ConnectionMode.cs

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/CosmosDb/IContainer.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// Microsoft.Azure.Cosmos.Container interface for duck typing
+    /// </summary>
+    /// <remarks>
+    /// https://github.com/Azure/azure-cosmos-dotnet-v3/blob/a25730a77ab43a8e460ddc292f1a6d8eb193395a/Microsoft.Azure.Cosmos/src/Resource/Container/Container.cs
+    /// </remarks>
+    internal interface IContainer : IDuckType
+    {
+        string Id { get; }
+
+        IDatabase Database { get; }
```
lucaspimentel: Check with @bouwkast, but I think recently we've opted to make most of these nullable to enforce null checks.
```suggestion
        string? Id { get; }

        IDatabase? Database { get; }
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/CosmosDb/IDatabase.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    /// <summary>
+    /// Microsoft.Azure.Cosmos.Database interface for duck typing
+    /// </summary>
+    /// <remarks>
+    /// https://github.com/Azure/azure-cosmos-dotnet-v3/blob/a25730a77ab43a8e460ddc292f1a6d8eb193395a/Microsoft.Azure.Cosmos/src/Resource/Database/DatabaseCore.cs
+    /// </remarks>
+    internal interface IDatabase : IDuckType
+    {
+        string Id { get; }
```
lucaspimentel: See comment in `IContainer.cs`
```suggestion
        string? Id { get; }
```

tracer/src/Datadog.Trace/Tags.cs
```diff
@@ -211,6 +211,11 @@ public static partial class Tags
         /// </summary>
         internal const string HttpUserAgent = "http.useragent";
 
+        /// <summary>
+        /// The original user agent string. Used for SDK clients such as Azure SDK.
+        /// </summary>
+        internal const string UserAgentOriginal = "user_agent.original";
```
bouwkast: Isn't this `http.useragent`? Or at least I thought that is our convention / mapping to this OpenTelemetry one

tracer/build/PackageVersionsLatestMajors.g.props
```diff
@@ -4992,12 +4992,6 @@ NOTE:  This code was generated by the GeneratePackageVersions tool. To safely
       <RequiresDockerDependency>All</RequiresDockerDependency>
       <SampleName>Samples.Kafka</SampleName>
     </PackageVersionSample>
-    <PackageVersionSample Include="test\test-applications\integrations\Samples.CosmosDb\Samples.CosmosDb.csproj">
```
bouwkast, line 5145: I think these have all been removed due to running the GeneratePackageVersions script on Mac?
Not a major deal though and probably something I will note down to try and improve.

Everything seems to work correctly though on .NET FX

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/CosmosDb/IContainer.cs
bouwkast: So the analyzer that I added for the nullability would actually be usages of this `IContainer` to assert that we correctly check that the `.Instance` of it isn't null like so here https://github.com/DataDog/dd-trace-dotnet/pull/7837/changes/BASE..9e76fdeb01fdc793552cf359a499630ac8517d31#diff-6d20fdf5e3a9fdde27d0c555d0d967ce44f5ae83910a05acb7b4ad7c0d511756R65

So not necessarily the properties _within_ the object

lucaspimentel: Our internal analyzer aside, is making these properties nullable something we recommend so that the C# compiler itself will warn you if you don't check for null?

bouwkast: Yeah I think marking them nullable is a good idea 👍

PR #7838, opened 2025-11-18
[IAST] Fix racing conditions in Dataflow

NachoEchevarria:
## Summary of changes

This PR fixes multi-threading issues in Dataflow that could lead to corruption of the internal std::map structures (_modules, _appDomains, _moduleAspects) and subsequent crashes inside MSVC STL (_Tree::_Find_lower_bound). These issues were observed [as flaky CI failures on Windows x86](https://dev.azure.com/datadoghq/a51c4863-3eb4-4c5d-878a-58b41a049e4e/_apis/build/builds/190706/logs/22028), manifesting as:

```
2025-11-10T12:05:43.1623592Z 12:05:43 [DBG]  Standard Output Messages:
2025-11-10T12:05:43.1624040Z 12:05:43 [DBG]  Platform: X86
2025-11-10T12:05:43.1624418Z 12:05:43 [DBG]  TargetPlatform: X86
2025-11-10T12:05:43.1624695Z 12:05:43 [DBG]  Configuration: Release
2025-11-10T12:05:43.1624968Z 12:05:43 [DBG]  TargetFramework: net7.0
2025-11-10T12:05:43.1625239Z 12:05:43 [DBG]  .NET Core: True
2025-11-10T12:05:43.1625595Z 12:05:43 [DBG]  Native Loader DLL: D:\a\_work\1\s\shared\bin\monitoring-home\win-x86\Datadog.Trace.ClrProfiler.Native.dll
2025-11-10T12:05:43.1625940Z 12:05:43 [DBG]  Starting aspnetcore sample, agentPort: 53110
2025-11-10T12:05:43.1626334Z 12:05:43 [DBG]  Starting Application: D:\a\_work\1\s\artifacts\bin\Samples.Security.AspNetCore5\release_net7.0\Samples.Security.AspNetCore5.dll
2025-11-10T12:05:43.1626821Z 12:05:43 [DBG]  ProcessId: 4216
2025-11-10T12:05:43.1627140Z 12:05:43 [DBG]  [webserver][stderr] Fatal error. Internal CLR error. (0x80131506)
2025-11-10T12:05:43.1627502Z 12:05:43 [DBG]  [webserver][stderr]    at System.Threading.PortableThreadPool+GateThread.GateThreadStart()
2025-11-10T12:05:43.1627849Z 12:05:43 [DBG]  [webserver][stderr]    at System.Threading.Thread.StartCallback()
2025-11-10T12:05:43.1628229Z 12:05:43 [DBG]  [webserver][stdout] [createdump] Writing full dump for process 4216 to file C:\Users\AzDevOps\AppData\Local\Temp\dump.4216.dmp
2025-11-10T12:05:43.1628594Z 12:05:43 [DBG]  [webserver][stdout] [createdump] Dump successfully written in 11266ms
2025-11-10T12:05:43.1628956Z 12:05:43 [DBG]  [webserver][stdout] info: Microsoft.AspNetCore.DataProtection.KeyManagement.XmlKeyManager[63]
2025-11-10T12:05:43.1629405Z 12:05:43 [DBG]  [webserver][stdout]       User profile is available. Using 'C:\Users\AzDevOps\AppData\Local\ASP.NET\DataProtection-Keys' as key repository and Windows DPAPI to encrypt keys at rest.
2025-11-10T12:05:43.1629853Z 12:05:43 [DBG]  [webserver][stdout] info: Microsoft.Hosting.Lifetime[14]
2025-11-10T12:05:43.1630499Z 12:05:43 [DBG]  [webserver][stdout]       Now listening on: http://127.0.0.1:53179
2025-11-10T12:05:43.1630959Z 12:05:43 [DBG]  [webserver][stdout] info: Microsoft.Hosting.Lifetime[0]
2025-11-10T12:05:43.1631295Z 12:05:43 [DBG]  [webserver][stdout]       Application started. Press Ctrl+C to shut down.
```

Inspecting the dump, we get the following stack:

```
02d7e2d8 6ac018b6     02d7d760 02d7d864 b928276d Datadog_Tracer_Native!std::_Tree<std::_Tset_traits<unsigned int,std::less<unsigned int>,std::allocator<unsigned int>,0> >::_Find_lower_bound<unsigned int>+0x33
(Inline) --------     -------- -------- -------- Datadog_Tracer_Native!std::_Tree<std::_Tmap_traits<unsigned int,iast::ModuleInfo *,std::less<unsigned int>,std::allocator<std::pair<unsigned int const ,iast::ModuleInfo *> >,0> >::_Find+0x1f
(Inline) --------     -------- -------- -------- Datadog_Tracer_Native!std::_Tree<std::_Tmap_traits<unsigned int,iast::ModuleInfo *,std::less<unsigned int>,std::allocator<std::pair<unsigned int const ,iast::ModuleInfo *> >,0> >::find+0x1f
02d7e2d8 6ac01f24     2be9035c 034fbf20 02d7e7d4 Datadog_Tracer_Native!iast::Dataflow::GetModuleInfo+0xc6
02d7e2f8 6abb96b4     2be9035c 0600001d 00000000 Datadog_Tracer_Native!iast::Dataflow::JITProcessMethod+0x24
(Inline) --------     -------- -------- -------- Datadog_Tracer_Native!iast::Dataflow::IsInlineEnabled+0x16
02d7e54c 6ade76d8     034fbf20 2be90824 02d7e7d4 Datadog_Tracer_Native!trace::CorProfiler::JITCachedFunctionSearchStarted+0x1e4
02d7e5b8 2b0644d8     000002ad 00000000 02a302f7 Datadog_Trace_ClrProfiler_Native+0x476d8
00000000 00000000     00000000 00000000 00000000 0x2b0644d8
```

This is where the exception was thrown, but not necessarily where the problem is. The map could get corrupted earlier and later the exception thrown in some internal checks.

---

## Root Causes

The Dataflow instance is accessed from multiple CLR profiler callbacks (`JITCachedFunctionSearchStarted`, `JITCompilationStarted`, module load/unload, ReJIT), which execute on various .NET threads. While most code paths synchronize access to shared maps via `_cs`, several critical issues existed:

### 1. Unsynchronized Destructor
`Dataflow::~Dataflow()` modified `_modules`, `_appDomains`, and `_moduleAspects` without holding `_cs`, allowing races where:
- One thread was destructing/clearing the maps
- Another thread entered methods like `GetModuleInfo()` under the lock
- Result: STL tree corruption and crashes

### 2. Time-of-Check-Time-of-Use (TOCTOU) Race Conditions
Multiple methods checked `_initialized` **outside** the critical section, then accessed shared maps **inside** the lock:
- `GetAppDomain()`, `GetModuleInfo()`, `GetAspectsModule()`
- `AppDomainShutdown()`, `ModuleUnloaded()`
- `GetAspects()`

This created a race window where:
1. Thread A checks `IsInitialized()` → returns TRUE (no lock held)
2. Thread B enters destructor, acquires lock, sets `_initialized = false`, destroys all maps
3. Thread A acquires lock, accesses destroyed maps → **CRASH**

### 3. Non-Atomic `_initialized` Flag
`_initialized` was a plain `bool` accessed concurrently without synchronization, causing:
- Data races (undefined behavior in C++)
- Lack of memory ordering guarantees between destructor and callbacks

### 4. Unsynchronized Map Access in LoadAspects()
`LoadAspects()` cleared `_moduleAspects` without locking.

---

### 1. Eliminate TOCTOU Races - Move Checks Inside Critical Sections
All methods that directly access shared maps now check `IsInitialized()` **inside** the `CSGUARD(_cs)` lock:
- `~Dataflow()` - Acquires lock first, then sets flag and destroys maps atomically
- `GetAppDomain()` - Lock → check → access
- `GetModuleInfo()` - Lock → check → access
- `AppDomainShutdown()` - Lock → check → access
- `ModuleUnloaded()` - Lock → check → access
- `GetAspects()` - Lock → check → access

This ensures that once the destructor sets `_initialized = false` inside the lock, no other thread can proceed past the check to access destroyed state.

**Before (vulnerable):**
```cpp
if (!_isInitialized)    // Thread A checks outside lock → TRUE
    return;
                         // Thread B: lock → set false → destroy maps → unlock
CSGUARD(_cs);            // Thread A: acquires lock
auto x = _modules.find() // Thread A: accesses destroyed map → CRASH
```

**After (safe):**
```cpp
CSGUARD(_cs);            // Thread A: acquires lock
if (!IsInitialized())    // Thread A: checks inside lock
    return;              // If Thread B already set false, exits safely
auto x = _modules.find() // Only proceeds if initialized
```

### 2. Atomic `_initialized` with Proper Memory Ordering
Changed `_initialized` from `bool` to `std::atomic<bool>` with:
- `memory_order_acquire` for loads (in `IsInitialized()`)
- `memory_order_release` for stores (in `SetInitialized()`)

This ensures:
- No data races (well-defined behavior)
- Proper memory visibility: when a thread sees `_initialized == false`, it's guaranteed to see all prior writes made before the release store
- Required because many methods still check `IsInitialized()` outside locks (e.g., `IsInlineEnabled()`, `JITProcessMethod()`) but safely delegate to methods that lock properly

### 3. Inline Helper Methods in Header
Added `IsInitialized()` and `SetInitialized()` to the header file as inline methods for:
- Better code generation (inlining)
- Consistent atomic access patterns
- Clear abstraction over atomic operations

### 4. Optimize LoadAspects() Locking
Updated `LoadAspects()` to:
- Hold `_cs` lock only for the `swap()` operation on `_moduleAspects`
- Destroy old `ModuleAspects` objects **outside** the lock
- Minimizes lock contention and prevents potential deadlocks

### 5. Remove Redundant Lock in GetAspectsModule()
Removed `CSGUARD(_cs)` from `GetAspectsModule()` since it delegates to `GetModuleInfo()`, which now properly locks. Avoids unnecessary nested locking (though safe due to recursive `CRITICAL_SECTION` on Windows).

---

## Reason for change

Fix critical race conditions causing flaky crashes in CI, particularly on Windows x86. The crashes manifested as STL internal errors during concurrent access to Dataflow's internal maps during JIT compilation and shutdown scenarios.

This error could also happen in customer installations, causing crashes.

---

## Implementation details

**Files changed:**
- `tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp` - TOCTOU fixes, lock reordering, destructor synchronization
- `tracer/src/Datadog.Tracer.Native/iast/dataflow.h` - Atomic flag declaration, inline helper methods with acquire/release semantics

**Key synchronization pattern:**
The fix leverages the recursive nature of Windows `CRITICAL_SECTION` (used by `CSGUARD`) while eliminating TOCTOU races by ensuring all shared state access happens atomically with the initialization check under the same lock.

**Methods updated:**
- `~Dataflow()` - Lock before setting flag and destroying maps
- `GetAppDomain()`, `GetModuleInfo()`, `GetAspectsModule()` - Check inside lock
- `AppDomainShutdown()`, `ModuleUnloaded()`, `GetAspects()` - Check inside lock
- Added `IsInitialized()`/`SetInitialized()` - Atomic with acquire/release memory ordering

---

## Test coverage

This fix addresses the root cause of the flaky Windows x86 CI failures. The existing CI pipeline validates the fix through:
- Multi-threaded scenarios with concurrent JIT compilation and module loading
- Windows x86 integration tests (where the issue was originally observed)
- AppSec security tests that exercise IAST Dataflow heavily

### DIFF
--- a/tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
+++ b/tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
@@ -700,6 +700,10 @@ HRESULT Dataflow::RewriteMethod(MethodInfo* method, trace::FunctionControlWrappe
     HRESULT hr = S_OK;
 
     CSGUARD(_cs);
+    if (!IsInitialized())
+    {
+        return S_FALSE;
+    }
 
     if (!pFunctionControl)
     {


tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
```diff
@@ -191,9 +193,9 @@ Dataflow::~Dataflow()
 
 void Dataflow::LoadAspects(WCHAR** aspects, int aspectsLength, UINT32 enabledCategories, UINT32 platform)
 {
-    if (!_initialized)
+    if (!IsInitialized())
     {
-        _initialized = true;
+        SetInitialized(true);
```
andrewlock, line 198: Silly question, but could this also be part of the problem? 🤔 

We set `initialized = true` here _before_ we've finished the initialization. That means that all the other places where we do `if (!IsInitialized())` and bail out will _stop_ bailing 🤔 Also, we don't have a CSGUARD in here, should we?

I feel like we probably need something more like this?

```cpp
CSGUARD(_cs);
if (!IsInitialized())
{
    // Do all the other work in the if block

    SetInitialized(true);
}
```

tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
```diff
@@ -369,12 +377,12 @@ void Dataflow::LoadSecurityControls()
 
 HRESULT Dataflow::AppDomainShutdown(AppDomainID appDomainId)
 {
-    if (!_initialized)
+    CSGUARD(_cs);
```
andrewlock, line 380: Do we want to use double checking here to avoid contention? I'm not sure it's a big problem for `AppDomainShutdown` really, and how often would we _not_ be initialized? 🤔 

```cpp
// Initial check outside the lock
if (!IsInitialized())
{
    return S_OK;
}

CSGUARD(_cs);
// Double check inside the lock
if (!IsInitialized())
{
    return S_OK;
}
```

tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
```diff
@@ -388,7 +396,7 @@ HRESULT Dataflow::AppDomainShutdown(AppDomainID appDomainId)
 
 HRESULT Dataflow::ModuleLoaded(ModuleID moduleId, ModuleInfo** pModuleInfo)
 {
-    if (!_initialized)
+    if (!IsInitialized())
```
andrewlock, line 399: This still has the CSGUARD _after_ the `IsInitialized()` check (because it takes the lock in `GetModuleInfo`... should we move this check inside `GetModuleInfo()` instead?

tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
```diff
@@ -595,7 +613,11 @@ ModuleInfo* Dataflow::GetModuleInfo(ModuleID id)
 
 ModuleInfo* Dataflow::GetAspectsModule(AppDomainID id)
 {
-    CSGUARD(_cs);
+    if (!IsInitialized())
```
andrewlock, line 616: where did the guard go?! 😅

tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
```diff
@@ -618,7 +640,7 @@ MethodInfo* Dataflow::GetMethodInfo(ModuleID moduleId, mdMethodDef methodId)
 
 bool Dataflow::IsInlineEnabled(ModuleID calleeModuleId, mdToken calleeMethodId)
 {
-    if (!_initialized)
+    if (!IsInitialized())
```
andrewlock, line 643: Seems like there's a bunch of places here without the guards... is that safe?

tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
NachoEchevarria: Actually, GetModuleInfo locks the sensitive part. We have an additional early check here to avoid calling unnecessarily to GetProfilerAssemblyModuleId, but:

ModuleInfo* Dataflow::GetModuleInfo(ModuleID id) 
{
    CSGUARD(_cs);
    if (!IsInitialized())
    {
        return nullptr;
    }

tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
tonyredondo: yes I do think you need to guard in this one as well.

tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
```diff
@@ -182,7 +182,9 @@ Dataflow::Dataflow(ICorProfilerInfo* profiler, std::shared_ptr<RejitHandler> rej
 
 Dataflow::~Dataflow()
 {
-    _initialized = false;
+    CSGUARD(_cs);
```
dromanol, line 185: Dataflow is destroyed on profiler unload and this is the only place where _initialized is set to false. This CS is unnecessary (although does not hurt).

tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
tonyredondo: If we add the guard for each IsInitialized() call we should consider to just use a normal bool instead of the atomic bool.

tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
dromanol: The main problem is here, in LoadAspects, and this function is the one that has to be well covered. _isInitialized should be set to true as the last thing done here, not in the begining of the function.

tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
NachoEchevarria: It is safe as long as we dont perform sensitive operations. In this case, IsInitialized() is already safe because we are using an atomic bool and we will recheck that later in a locked envirinment.
Methods like IsInlineEnabled(), JITProcessMethod(), and JITCompilationStarted() check IsInitialized() outside locks because the atomic read itself is safe, they perform no "sensitive operations" (directly accessing shared state _modules, _appDomains, _moduleAspects) and the real protection is at the leaf level. When these methods call GetModuleInfo() or GetAppDomain(), those methods acquire the lock first, check IsInitialized() inside the lock (TOCTOU-safe) and only then access shared maps.

tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
dromanol: We are not going to find the situation where a threar reads true while it has been set to false. The problem is the oposite, in initialization. IMO the whole thing could be solved with deferring the set to true in _isInitialized to the end of the LoadAspects function.

tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
NachoEchevarria: GetModuleInfo() already checks for that, this is just an early check.

PR #7839, opened 2025-11-18
Add process tags to dynamic instrumentation snapshots

vandonr:
## Summary of changes

add a process tags field next to where the service name is added in the snapshot json.

dotnet version of https://github.com/DataDog/dd-trace-java/pull/8779

## Reason for change

AIDM-195

## Implementation details

I wanted to inject the setting from https://github.com/DataDog/dd-trace-dotnet/blob/79b266b0c7a329eae91584356a1cfc7c5599c1ee/tracer/src/Datadog.Trace/Debugger/DebuggerFactory.cs#L30 but looking at the call chain to bring it to the snapshot creator, it was quite tedious and messy.
Another option was to put it in the debugger settings, but it felt out of place as those settings are specific to the live debugger, not a repeat of global tracer settings.
In the end I decided to inject _a little bit_ so that I could specify the value in tests.

## Test coverage

just made sure the test that checks json correctness includes this

### DIFF
--- a/.azure-pipelines/ultimate-pipeline.yml
+++ b/.azure-pipelines/ultimate-pipeline.yml
@@ -2249,7 +2249,7 @@ stages:
       parameters:
         build: true
         baseImage: $(baseImage)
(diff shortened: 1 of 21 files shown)


GreenMatan (approved):
LGTM

PR #7840, opened 2025-11-18
Add process tags to remote config

vandonr:
## Summary of changes

adding process tags next to the service name in remote config requests

equivalent Java PR: https://github.com/DataDog/dd-trace-java/pull/8705
related change on the agent side: https://github.com/DataDog/datadog-agent/pull/36049

## Reason for change

AIDM-195

## Test coverage

I didn't see any existing tests on the content of the requests, didn't really feel like adding some.

### DIFF
--- a/tracer/src/Datadog.Trace/RemoteConfigurationManagement/Protocol/RcmClientTracer.cs
+++ b/tracer/src/Datadog.Trace/RemoteConfigurationManagement/Protocol/RcmClientTracer.cs
@@ -12,7 +12,7 @@ namespace Datadog.Trace.RemoteConfigurationManagement.Protocol
 {
     internal class RcmClientTracer
     {
-        public RcmClientTracer(string runtimeId, string tracerVersion, string service, string env, string? appVersion, List<string> tags, List<string> processTags)
+        public RcmClientTracer(string runtimeId, string tracerVersion, string service, string env, string? appVersion, List<string> tags, List<string>? processTags)
         {
             RuntimeId = runtimeId;
             Language = TracerConstants.Language;
@@ -37,7 +37,7 @@ namespace Datadog.Trace.RemoteConfigurationManagement.Protocol
         public string Service { get; }
 
         [JsonProperty("process_tags")]
-        public List<string> ProcessTags { get; }
+        public List<string>? ProcessTags { get; }
 
         [JsonProperty("extra_services")]
         public string[]? ExtraServices { get; set; }
--- a/tracer/src/Datadog.Trace/RemoteConfigurationManagement/RemoteConfigurationManager.cs
+++ b/tracer/src/Datadog.Trace/RemoteConfigurationManagement/RemoteConfigurationManager.cs
@@ -69,7 +69,7 @@ namespace Datadog.Trace.RemoteConfigurationManagement
             return new RemoteConfigurationManager(
                     discoveryService,
                     remoteConfigurationApi,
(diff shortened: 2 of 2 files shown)


vandonr:
> Will this be covered by system-tests? Historically we've broken RCM and not known about it, so I'm hesitant to just ok this as-is.

yes it should be, though I have not enabled them yet because it's made as all-or-nothing for process tags. I'll double check.

>  add some snapshots for the response

We do have some tests with sample responses, in https://github.com/DataDog/dd-trace-dotnet/blob/20246d2febfbbc43aad6598e9ddebfbb36d89aa2/tracer/test/Datadog.Trace.Tests/RemoteConfigurationManagement/RcmResponses.cs, but here I modify the request, so I can create snapshots of requests, but it'd be kinda self-asserting

vandonr:
ran [the system test](https://github.com/DataDog/system-tests/blob/3e0095f43287fc88502cf8f7856a4f6e9cfb2382/tests/apm_tracing_e2e/test_process_tags.py#L35)
and realized the format was incorrect 😅, this is the only place where we don't want the serialized list in string form, but in list form in the json. I rearranged the code in ProcessTags to expose both, in a way that is cleaner in the end I think :)

tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -20,52 +20,56 @@ internal static class ProcessTags
...
-    /// From the full path of a directory, get the name of the leaf directory.
-    /// </summary>
-    private static string GetLastPathSegment(string directoryPath)
+    private static List<string> GetTagsList()
     {
-        // Path.GetFileName returns an empty string if the path ends with a '/'.
-        // We could use Path.TrimEndingDirectorySeparator instead of the trim here, but it's not available on .NET Framework
-        return Path.GetFileName(directoryPath.TrimEnd('\\').TrimEnd('/'));
+        // ⚠️ make sure entries are added in alphabetical order of keys
+        List<string> tags = [];
```
andrewlock: Ironically, as you know the max size of list you need, you may as well _not_ use collection initializers here, so that you can set the capacity:
```suggestion
        var tags = new List<string>(3); // Update if you add more entries below
```

tracer/src/Datadog.Trace/RemoteConfigurationManagement/Protocol/RcmClientTracer.cs
```diff
@@ -35,6 +36,9 @@ public RcmClientTracer(string runtimeId, string tracerVersion, string service, s
         [JsonProperty("service")]
         public string Service { get; }
 
+        [JsonProperty("process_tags")]
+        public List<string> ProcessTags { get; }
```
andrewlock: Is it correct that we send an empty list here when `tracerSettings.PropagateProcessTags == false`? Or should we omit the property entirely (e.g. by passing in null to the constructor)?

vandonr: hmm yes you're right we should omit it entirely, otherwise it'll still show up in the json

andrewlock: In which case passing null through instead of an empty list should solve that I think 👍

vandonr: done

PR #7841, opened 2025-11-19
Set benchmarking jobs as interruptible when outside master

igoragoli:
## Summary of changes

- Set benchmarking jobs as interruptible when outside master.

Also:
- Consolidate microbenchmarking artifacts into `platform/artifacts`, following https://github.com/DataDog/benchmarking-platform/commit/733f92a71a1e153b274ad401b4057b888a2fedc1.

## Reason for change

[APMSP-2369](https://datadoghq.atlassian.net/browse/APMSP-2369) and [#incident-45572](https://dd.enterprise.slack.com/archives/C09S0H9M3B9)

Why set jobs to interruptible? Because by doing so, we prevent redundant runs from overloading our APM runner capacity.

Why only outside of master? Because benchmarks on master are never redundant: we need every run on master to collect results and maintain a continuous performance history.

## Test coverage

A dummy commit, [a95a4f9](https://github.com/DataDog/dd-trace-dotnet/pull/7841/commits/a95a4f985588fabdcd965810a5e760ce1640ff21), cancelling previous microbenchmarks from [1cd3cd2](https://github.com/DataDog/dd-trace-dotnet/pull/7841/commits/1cd3cd2d5f41ba384f5cc012933b28394cf5d0cf):

<div align="center">
image
</div>

For macrobenchmarks:

<div align="center">
image
</div>

## Other details

[APMSP-2369]: https://datadoghq.atlassian.net/browse/APMSP-2369?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.gitlab/benchmarks/dsm-throughput.yml
+++ b/.gitlab/benchmarks/dsm-throughput.yml
@@ -25,7 +25,6 @@ check_azure_pipeline:
     expire_in: 3 months
   tags: ["arch:amd64"]
   timeout: 1h
-  when: on_success
   rules:
     - if: $CI_COMMIT_REF_NAME == "master"
       interruptible: false
@@ -50,7 +49,6 @@ dsm_throughput:
     paths:
       - artifacts/
     expire_in: 3 months
-  when: on_success
   rules:
     - if: $CI_COMMIT_REF_NAME == "master"
       interruptible: false


.gitlab/benchmarks/macrobenchmarks.yml
```diff
@@ -81,8 +88,11 @@ update-bp-infra:
       - scheduler_failure
       - api_failure
       - script_failure
+  when: on_success
   rules:
-    - when: on_success
+    - if: $CI_COMMIT_REF_NAME == "master"
+      interruptible: false
+    - interruptible: true
```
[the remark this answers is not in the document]
igoragoli: https://docs.gitlab.com/ci/yaml/#when

.gitlab/benchmarks/microbenchmarks.yml
```diff
@@ -78,14 +82,16 @@ run-benchmarks:
   id_tokens:
     DDOCTOSTS_ID_TOKEN:
       aud: dd-octo-sts
+  when: on_success
   rules:
-    - when: on_success
+    - if: $CI_COMMIT_REF_NAME == "master"
```
andrewlock: We should probably exclude hot fix branches too? 🤔 And I wonder if we need to exclude tags (which happen on master, and therefore could impact our green-CI metrics if they're cancelled?

.gitlab/benchmarks/macrobenchmarks.yml
```diff
@@ -12,6 +12,10 @@ stages:
   - benchmarks
   - benchmarks-win
 
+workflow:
+  auto_cancel:
+    on_new_commit: interruptible
```
andrewlock, line 17: I think we have DSM throughput tests too - maybe we should update those similarly  @robcarlan-datadog?

.gitlab/benchmarks/microbenchmarks.yml
igoragoli: Thanks!

The condition is now 
```
$CI_COMMIT_REF_NAME == "master" || 
$CI_COMMIT_REF_NAME =~ /^hotfix.*$/ || 
$CI_COMMIT_REF_NAME =~ /^release.*$/ || 
$CI_COMMIT_REF_NAME =~ /^v[0-9]+\.[0-9]+\.[0-9]+(-prerelease)?$/
```

.gitlab/benchmarks/macrobenchmarks.yml
igoragoli: Thanks! I wasn't aware those jobs also used APM runners. Added the same interrupt policy to them 👍

.gitlab/benchmarks/microbenchmarks.yml
igoragoli: As per our chat on Slack, the condition is now reverted to `$CI_COMMIT_REF_NAME == "master"` only and microbenchmarks don't run on tags anymore.

igoragoli:
Microbenchmarks were already not running on tags because they couldn't find an open PR: https://gitlab.ddbuild.io/DataDog/apm-reliability/dd-trace-dotnet/-/jobs/1238672257

But now we enforce it through code, and the job won't even start.

PR #5206, opened 2024-02-17
[Tracing] Update instrumentation point for DD_TRACE_DELAY_WCF_INSTRUMENTATION_ENABLED=true

zacharycmontoya:
## Summary of changes
Updates the instrumentation point for `DD_TRACE_DELAY_WCF_INSTRUMENTATION_ENABLED=true` so that a server span is created before IDispatchMessageInspector implementations are run, so application code can access the root span from inside a IDispatchMessageInspector.AfterReceiveRequest callback.

This PR also does some cleanup to remove unused Wcf files and it makes the entire Wcf instrumentation use nullable reference types.

## Reason for change
We have users who want to get the WCF server span using `Tracer.Instance.ActiveScope` from inside a IDispatchMessageInspector.AfterReceiveRequest callback and more information to it.

## Implementation details
~This PR is meant to be reviewed commit-by-commit:~

After adding several commits to respond to PR feedback this is no longer really suitable for reviewing commit-by-commit, but the high-level changes are as follows:

1. A IDispatchMessageInspector implementation is added to the `Samples.Wcf` application to show that when `DD_TRACE_DELAY_WCF_INSTRUMENTATION_ENABLED=false` we are already creating a span before IDispatchMessageInspector implementations are invoked. The result is that WCF spans in the `DD_TRACE_DELAY_WCF_INSTRUMENTATION_ENABLED=false` scenario now have a custom tag `custom-tag: DispatchMessageInspector` without the code changes in this PR.
2. The instrumentation for `DD_TRACE_DELAY_WCF_INSTRUMENTATION_ENABLED=true` is changed
    - Existing integrations: Since the existing `MethodInvoker` integrations are able to observe exceptions thrown by the WCF endpoint, they are updated to only record exceptions on the current span, but they do not start/stop spans.
    - New integrations: A new `AfterReceiveRequestIntegration` integration is added that starts a span before IDispatchMessageInspector.AfterReceiveRequest implementations are invoked, and a new `BeforeSendReplyIntegration` integration is added that stops the span after the IDispatchMessageInspector.BeforeSendReply implementations are invoked. This enables the new functionality without causing a regression from the original [issue](https://github.com/DataDog/dd-trace-dotnet/pull/1992) that prompted us to create the `DD_TRACE_DELAY_WCF_INSTRUMENTATION_ENABLED` flag in the first place.
[...]

PR #1992, opened 2021-11-05
Add modified WCF CallTarget instrumentation via opt-in environment variable

zacharycmontoya:
### Issue
Our current WCF instrumentation (both CallSite and CallTarget) targets `System.ServiceModel.Dispatcher.ChannelHandler.HandleRequest`, which is very early in the callstack before a lot of the other WCF infrastructure is called, so any exceptions that propagate outside of the instrumentation will crash the server. For our instrumentation to create meaningful spans that are also connected to distributed parents, we must access the request message to gather information and create the span.

Unfortunately, in one known scenario, the first access to the request message should throw an exception that the WCF infrastructure should catch and it should cease processing of that request. Since we currently swallow exceptions during scope creation, the message continues to be processed and the application behavior is changed. Even if we decided to throw an exception, because we instrument a method before the main message processing loop, we would actually crash the server instead of stopping the processing of that one message. As a result, the only fix for this scenario is to defer our instrumentation until later in the WCF infrastructure code.

Since this is a big change and our WCF server testing is not as robust as other integrations, this change is being implemented by a feature flag.

### Changes
- In previous CallSite and CallTarget WCF instrumentation, check whether the WCF integration is enabled before accessing the `MessageRequest` object
- Add a new CallTarget capability: We can now explicitly state which method parameters should be passed from the instrumented method into the `OnMethodBegin` callbacks. This allows us to now instrument methods that have `ref`/`in`/`out` parameters, as long as we bypass those specific parameters
  - This is required to instrument the desired methods for the types below
- Add new WCF CallTarget instrumentation to the 3 built-in `IOperationInvoker` implementations, which is the last bit of WCF infrastructure code before the actual WCF services are invoked
  - `SyncMethodInvoker`
  - `AsyncMethodInvoker`
  - `TaskMethodInvoker`
- Disable the previous CallTarget WCF instrumentation and enable the new CallTarget WCF instrumentation via opt-in environment variable `DD_TRACE_DELAY_WCF_INSTRUMENTATION_ENABLED`

### Testing
- Update `Samples.Wcf` test application to invoke all 3 built-in `IOperationInvoker` implementations: `SyncMethodInvoker`, `AsyncMethodInvoker`, and `TaskMethodInvoker`
  - Before, we were only testing `SyncMethodInvoker`
- Update `Samples.Wcf` test application with a new custom binding that throws an exception on the first `RequestMessage` access, exposing the issue with our existing CallSite and CallTarget WCF instrumentation
[...]

PR #7842, opened 2025-11-19
Fix orphaned spans in WCF traces

andrewlock:
## Summary of changes

Fixes orphan traces in WCF

## Reason for change

Many customers have reported that in v3, with "delayed WCF" instrumentation enabled, they get orphan spans. This can be easily reproduced, so this PR fixes the issue.

## Background

There are two "modes" WCF can operate in: delayed instrumentation, and "simple" instrumentation. Delayed instrumentation was introduced in https://github.com/DataDog/dd-trace-dotnet/pull/1992 to avoid a problematic error scenario. Instead of instrumenting early across the whole "channel", we delayed instrumentation. A subsequent PR https://github.com/DataDog/dd-trace-dotnet/pull/5206 updated the instrumentation point to ensure the wcf span was active while `IDispatchMessageInspector` were running. 

This was somewhat complex, due to the convoluted and legacy way wcf uses async and threading. Basically, we use a weak table to hold the current span, and us the `MessageRpc` object as a key. Later on, we retrieve the scope and dispose it.

## Implementation details

The missing part was that we were not "restoring" the wcf scope for when the actual Invoker implementations executed. The process was:

- Update the tests and snapshots to:
  - Record _all_ spans and traces (it makes it easier to [understand the snapshots using](https://andrewlock.github.io/DatadogSpanSnapshotViewer/))
  - Create a span inside the "service" calls.
    - These _should_ have been children of the wcf server spans, but they were not. They were orphaned
- Update the "Invoker" implementations to grab the wcf spans, make it active, and then deactivate it afterwards
- Regenerate the snapshots, and confirm the custom spans are no longer orphaned 🎉 

Additionally, do a small amount of refactoring:
- "Fix" nullability in some places
- Extract the "restore wcf span" functionality to `WcfCommon` as it's used in multiple places

## Test coverage

Covered by existing, after the tweaks above. The snapshots change like this:

Before the fix, after adding the additional custom spans, `WSHttpBinding` traces:

image

After the fix, `WSHttpBinding` traces:

image

Before the fix, after adding the additional custom spans,  WebHttp traces:

image

After the fix, Web Http traces

image

Unfortunately, this ends up being somewhat unstable in regards to sorting, so I did some more refactoring to separate the root span for each separate httpcall. This should make sure we have more stable ordering of the spans using the built-in sort order

The result is we get separate traces for each API call instead:

image

## Other details

I wonder if we really need a weak table here? 🤔 seeing as we have an instrumentation point that explicitly disposes the span, shouldn't we just remove the entry at the same time?

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Wcf/WcfCommon.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Wcf/WcfCommon.cs
@@ -227,7 +227,6 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Wcf
             // First, capture the active scope
             var tracer = Tracer.Instance;
             var activeScope = tracer.InternalActiveScope;
-            var spanContextRaw = DistributedTracer.Instance.GetSpanContextRaw() ?? activeScope?.Span?.Context;
 
             var operationContext = WcfCommon.GetCurrentOperationContext?.Invoke();
             if (operationContext != null && operationContext.TryDuckCast<IOperationContextStruct>(out var operationContextProxy))
@@ -238,9 +237,12 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Wcf
                 if (requestContext?.Instance is { } requestContextInstance
                  && Scopes.TryGetValue(requestContextInstance, out var scope)
                     && scope is not null
-                    && scope.Span.SpanId != activeScope?.Span?.SpanId)
+                    && scope.Span.SpanId != activeScope?.Span.SpanId)
                 {
+                    // capture the raw context for later activation
+                    var spanContextRaw = DistributedTracer.Instance.GetSpanContextRaw() ?? activeScope?.Span.Context;
                     Log.Debug("Activating scope from operation context {ActivatedSpan}", scope.Span);
+
                     tracer.ActivateSpan(scope.Span);
                     // Add the exception but do not dispose the span.
                     // BeforeSendReplyIntegration is responsible for closing the span.


tracer/test/test-applications/integrations/Samples.Wcf/Client/Startup.cs
```diff
@@ -57,6 +53,8 @@ private static async Task<int> Invoke_ServerSyncAdd_Endpoints(CalculatorClient c
 
             try
             {
+                using var scope = SampleHelpers.CreateScope("ServerEmptyActionAdd");
+                calculator.DatadogScope = scope;
```
zacharycmontoya, line 57: Since you have some integration tests in place, can we try to remove this `DatadogScope` property? If it doesn't affect the results, we should remove it so we have less custom code

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Wcf/WcfCommon.cs
```diff
@@ -216,6 +221,47 @@ internal class WcfCommon
 
             return null;
         }
+
+        public static CallTargetState ActivateScopeFromContext()
```
zacharycmontoya, line 225: super small nit: It seems like this method has slightly too many responsibilities in that it's calling both `Tracer.Instance.ActivateSpan(Span)` and returning a `CallTargetState`. The only refactoring that comes to mind is to make this a `TryGetScopeFromContext` and return the fields used to populate the CallTargetState. And then from the caller you can activate the span and create the `CallTargetState`, but then you're making each callsite have to repeat multiple lines of code. I'm not sure if that's better than what you've done here, so no need to take my suggestion

zacharycmontoya (approved):
LGTM. As discussed, we have a corner case of WCF message headers overriding HTTP message headers. I think this PR is good as-is so we can tackle that as a follow-up PR if you'd like

tracer/test/test-applications/integrations/Samples.Wcf/Client/Startup.cs
andrewlock: Unfortunately, if we remove this scope, it will make the snapshots flakey. The nature of how WCF works means that all the client spans are identical, which in turn means the spans end up with a non-deterministic order if we remove these. This is also why I split each individual operation into a separate span, instead of having the single over-arching span. 

FWIW, I'm not sure why we need to set the `DatadogScope` property - AFAICT it's not used, but as it was doing it before, I cargo-culted instead of making more changes 😄

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Wcf/WcfCommon.cs
andrewlock: Yeah, I agree with you, but then also it's easy to screw up and forget to capture one or more parts that you need.. plus we can delay capturing the distributed context unless we need it (though we _should_ always hit the "activate" path in normal behaviour, so maybe not a big deal)

As you say, given every callsite would need to copy the exact same code to use the API, I'm not sure it's worth splitting out. We _could_ still capture the previous and then return a different carrier object instead of `CallTargetState` from the method, and then _convert_ it into a `CallTargetState` at the callsite, but that seems like more hassle for little gain? 🤷‍♂️ meh 😅

zacharycmontoya: Agreed: meh 😆

tracer/test/test-applications/integrations/Samples.Wcf/Client/Startup.cs
zacharycmontoya: Ah sorry, yes what I meant was removing the line `calculator.DatadogScope = scope;`. I understand that the manual scope in general helps with snapshot stability 👍🏼

andrewlock: Cool, I'll do that in the follow up PR that addresses other minor issues 🙂

PR #7844, opened 2025-11-19
[Dynamic Instrumentation] DEBUG-4758 Fix type mismatch with @duration expression

dudikeleti:
## Summary of changes
Fixes a bug in the debugger expression parser when compiling expressions with T == object that return value types (e.g. @duration as TimeSpan), causing a runtime error (“Expression of type 'System.TimeSpan' cannot be used for assignment to type 'System.Object'”).

## Test coverage
ProbeExpressionParser_ObjectReturnType_AllowsValueTypeResults


GreenMatan (approved):
LGTM

PR #7847, opened 2025-11-19
Add standard output and error when a test fails and does not display it

NachoEchevarria:
## Summary of changes

We get [some CI errors ](https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=191248&view=logs&j=5635f724-ec42-5e82-7816-fb263b6cfcc0&t=e80b9b00-921f-539a-741c-f91926c040ef)in tests caused by a non 0 exit code from a sample but we get no output from the sample itself.

This PR improves error reporting when tests fail due to non-zero exit codes by ensuring that process stdout and stderr are logged to the test output before assertions fail. This makes debugging CI failures significantly easier, especially for signal-related crashes (SIGABRT, SIGSEGV, etc.).

### DIFF
--- a/tracer/test/Datadog.Trace.Debugger.IntegrationTests/Helpers/DebuggerSampleProcessHelper.cs
+++ b/tracer/test/Datadog.Trace.Debugger.IntegrationTests/Helpers/DebuggerSampleProcessHelper.cs
@@ -8,6 +8,7 @@ using System.Diagnostics;
 using System.Net;
 using System.Threading.Tasks;
 using Datadog.Trace.TestHelpers;
+using Xunit.Abstractions;
 
 namespace Datadog.Trace.Debugger.IntegrationTests.Helpers
 {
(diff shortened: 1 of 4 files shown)


tracer/test/Datadog.Trace.Debugger.IntegrationTests/Helpers/DebuggerSampleProcessHelper.cs
```diff
@@ -36,6 +39,13 @@ internal async Task StopSample()
                 throw new InvalidOperationException($"The process did not exit after {timeout}ms");
             }
 
+            if (Process.ExitCode != 0)
+            {
+                _output.WriteLine($"[DebuggerSampleProcessHelper] Process exited with code {Process.ExitCode} (0x{Process.ExitCode:X})");
```
andrewlock, line 44: I'm not sure if `_output` is _always_ non null? 🤔 If it is this is all good, otherwise we should maybe add a null check in the if block 🙂

NachoEchevarria: It should never be null. If it is, something pretty weird is going on.

PR #7848, opened 2025-11-19
[Dynamic Instrumentation] DEBUG-4689 Remove the 100-probe limit for Dynamic Instrumentation

dudikeleti:
## Summary of changes
Remove the hard-coded 100-probe limit per ptobe type for Dynamic Instrumentation and replace it with an internal, configurable per-type limit that defaults to “no limit”.

## Reason for change
The MCP server and agentic debugging use cases are becoming increasingly popular. This growth is expected to lead to hitting the 100 probe limit in services more frequently.

### DIFF
--- a/tracer/src/Datadog.Trace/Configuration/supported-configurations-docs.yaml
+++ b/tracer/src/Datadog.Trace/Configuration/supported-configurations-docs.yaml
@@ -514,6 +514,10 @@ DD_INSTRUMENTATION_TELEMETRY_URL: |
   telemetry is in use (otherwise telemetry is sent to the agent using
   <see cref="Datadog.Trace.Configuration.ExporterSettings.AgentUri"/> instead)
 
+DD_INTERNAL_DYNAMIC_INSTRUMENTATION_MAX_PROBES_PER_TYPE: |
+  Internal configuration key to optionally limit the maximum number of probes of each type.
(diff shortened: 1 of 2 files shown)


tracer/src/Datadog.Trace/Configuration/supported-configurations-docs.yaml
```diff
@@ -690,9 +690,13 @@ DD_SYMBOL_DATABASE_THIRD_PARTY_DETECTION_INCLUDES: |
   Default value is empty.
 
 DD_SYMBOL_DATABASE_UPLOAD_ENABLED: |
-  Configuration key for allowing upload of symbol data (such as method names, parameter names, etc) to Datadog.
+  Internal configuration key to optionally limit the maximum number of probes of each type.
```
andrewlock: It looks like the description for these two keys is a bit mangled?

dudikeleti: Weird, I was sure I fixed that. Thanks for the heads up.

dudikeleti: fixed 3f7bf63fb537960404a974a45dbb045e9ebb2c10

PR #7849, opened 2025-11-20
Add `#nullable enable` and fix `NullReferenceException` in DynamoDB

andrewlock:
## Summary of changes

- Add `#nullable enable` to a bunch of files
- Fix `NullReferenceException` in DynamoDB integrations

## Reason for change

While looking at logs, spotted null reference exception in DynamoDB integration. Adding `#nullable enable` (and being conservative about what _could_ be null) revealed multiple cases

## Implementation details

- Add `#nullable enable` to DynamoDB duck types and integrations
- Update source generator to respect updated nullability
- Fixed incorrect access on null objects
- Add missing dispose on enumerator

## Test coverage

Should be covered by existing. These are generally edge cases

## Other details

Made other cosmetic changes too, removing unnecessary comments and namespaces etc. Recommend commit-by-commit review for simplicity (bug-fixes are isolated to a single commit)


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/DynamoDb/DeleteItemIntegration.cs
NachoEchevarria, line 56: Thought this was a possible null reference exception scenario, but then, realized that: 
public static T? DuckCast<T>()

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/DynamoDb/AwsDynamoDbCommon.cs
```diff
@@ -73,16 +75,17 @@ public static void TagTableNameAndResourceName(string tableName, AwsDynamoDbTags
...
         {
-            if (request.RequestItems.Count != 1)
+            if (request.RequestItems?.Count != 1)
             {
                 return;
             }
 
             // TableName tagging only when batch is from one table.
             var iterator = request.RequestItems.GetEnumerator();
+            using var disposable = iterator as IDisposable;
```
lucaspimentel, line 88: cool trick

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/DynamoDb/IAmazonDynamoDbRequestWithKnownKeys.cs
```diff
@@ -9,64 +9,65 @@
 using System.IO;
 using Datadog.Trace.DuckTyping;
 
-namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.AWS.DynamoDb
+namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.AWS.DynamoDb;
+
+/// <summary>
+/// Interface for duck typing AmazonDynamoDbRequest implementations with the TableName and Key properties.
+/// https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Services/DynamoDBv2/Generated/Model/UpdateItemRequest.cs
+/// https://github.com/aws/aws-sdk-net/blob/main/sdk/src/Services/DynamoDBv2/Generated/Model/DeleteItemRequest.cs
```
lucaspimentel, line 17: thanks!

PR #7850, opened 2025-11-20
Implement Serverless Service Representation (SSR) in AWS Lambda

rithikanarayan:
## Summary of changes
Fixes the Serverless Service Representation (SSR) for Dotnet.
In serverless environments, sets peer.service tag for outbound AWS managed service spans to {service type}.{region}.amazonaws.com:

SQS: sqs.us-east-1.amazonaws.com
SNS: sns.us-east-1.amazonaws.com
Kinesis: kinesis.us-east-1.amazonaws.com
Eventbridge: events.us-east-1.amazonaws.com
s3: bucket-name.s3.us-east-1.amazonaws.com
Dynamodb: dynamodb.us-east-1.amazonaws.com

Sets peer.service.source to peer.service for these serverless cases. 

In non-serverless cases we assign the peer.service tag to the values of the following tags:

SQS: AwsSqsTags.QueueName
SNS: AwsSnsTags.TopicName
Kinesis: AwsKinesisTags.StreamName
Eventbridge: AwsEventBridgeTags.RuleName
S3: AwsS3Tags.BucketName
Dynamodb: AwsDynamoDbTags.TableName
StepFunctions: AwsStepFunctionsTags.StateMachineName

Sets peer.service source to the respective tag itself. (Ex. for an SQS queue with the name "MyQueue", peer.service = "MyQueue" and peer.service.source = "queuename").

Adds DerivePeerService function in Util/PeerServiceHelpers to do the above assignments. 

## Reason for change
Before SSR implementation:
Screenshot 2025-11-18 at 5 38 00 PM

After SSR implementation:
Screenshot 2025-11-24 at 1 38 17 PM

## Implementation details
See summary of changes. 

## Test coverage
Covered by AWS integration tests in non-serverless case. Added unit tests for the DerivePeerService function in the PeerServiceHelpersTests.cs file. Manually tested in serverless case as follows due to challenges in assigning the region tag in an integration test. Tested functionality in AWS Lambda by using custom dd-trace-dotnet layer with arn:`arn:aws:lambda:us-east-1:425362996713:layer:ssr-dotnet-rithika:15` on a sample app that had an API Gateway trigger a producer lambda, which sent data through Kinesis, SQS, SNS, and EventBridge to consumer lambdas, which wrote to S3 and DynamoDB. Visually inspected the service map (see images in Reason for change section) to check for correct labelling of the inferred spans for these services. Also checked tags in the trace overview to confirm the correct values of the peer.service and peer.service.source tags.

### DIFF
--- a/tracer/src/Datadog.Trace/Util/PeerServiceHelpers.cs
+++ b/tracer/src/Datadog.Trace/Util/PeerServiceHelpers.cs
@@ -29,37 +29,40 @@ namespace Datadog.Trace.Util
                 switch (service)
                 {
                     case "DynamoDB":
-                        tags.PeerService = "dynamodb." + region + ".amazonaws.com";
+                        tags.PeerService = $"dynamodb.{region}.amazonaws.com";
                         break;
                     case "EventBridge":
-                        tags.PeerService = "events." + region + ".amazonaws.com";
+                        tags.PeerService = $"events.{region}.amazonaws.com";
                         break;
                     case "Kinesis":
-                        tags.PeerService = "kinesis." + region + ".amazonaws.com";
+                        tags.PeerService = $"kinesis.{region}.amazonaws.com";
                         break;
                     case "S3":
                         if (tags is AwsS3Tags s3Tags)
                         {
                             if (s3Tags.BucketName != null)
                             {
-                                tags.PeerService = s3Tags.BucketName + ".s3." + region + ".amazonaws.com";
+                                tags.PeerService = $"{s3Tags.BucketName}.s3.{region}.amazonaws.com";
                             }
                             else
                             {
-                                tags.PeerService = "s3." + region + ".amazonaws.com";
+                                tags.PeerService = $"s3.{region}.amazonaws.com";
                             }
                         }
 
                         break;
                     case "SNS":
-                        tags.PeerService = "sns." + region + ".amazonaws.com";
+                        tags.PeerService = $"sns.{region}.amazonaws.com";
                         break;
                     case "SQS":
-                        tags.PeerService = "sqs." + region + ".amazonaws.com";
+                        tags.PeerService = $"sqs.{region}.amazonaws.com";
                         break;
                 }
 
-                tags.PeerServiceSource = "peer.service";
+                if (tags.PeerService != null)
+                {
+                    tags.PeerServiceSource = "peer.service";
+                }
             }
             else if (!isAwsLambda)
             {
--- a/tracer/test/Datadog.Trace.Tests/Util/PeerServiceHelpersTests.cs
+++ b/tracer/test/Datadog.Trace.Tests/Util/PeerServiceHelpersTests.cs
@@ -25,174 +25,152 @@ public class PeerServiceHelpersTests
     }
 
     [Theory]
-    [InlineData("DynamoDB", true, false)]
-    [InlineData("DynamoDB", false, false)]
-    [InlineData("EventBridge", true, false)]
-    [InlineData("EventBridge", false, false)]
-    [InlineData("Kinesis", true, false)]
-    [InlineData("Kinesis", false, false)]
-    [InlineData("S3", true, false)]
-    [InlineData("S3", false, false)]
-    [InlineData("S3", true, true)]
-    [InlineData("S3", false, true)]
-    [InlineData("SNS", true, false)]
-    [InlineData("SNS", false, false)]
-    [InlineData("SQS", true, false)]
-    [InlineData("SQS", false, false)]
-    [InlineData("StepFunctions", false, false)]
-    public void DerivePeerService(string service, bool isAwsLambda, bool hasBucket)
+    [PairwiseData]
+    public void DerivePeerServiceDynamoDB(bool isAwsLambda)
     {
-        switch (service)
+        var tags = new AwsDynamoDbTags();
+        tags.Service = "DynamoDB";
+        tags.Region = "us-east-1";
+        tags.TableName = "example-table";
+        PeerServiceHelpers.DerivePeerService(tags, isAwsLambda);
+        if (isAwsLambda)
         {
-            case "DynamoDB":
-            {
-                var tags = new AwsDynamoDbTags();
-                tags.Service = "DynamoDB";
-                tags.Region = "us-east-1";
-                tags.TableName = "example-table";
-                PeerServiceHelpers.DerivePeerService(tags, isAwsLambda);
-                if (isAwsLambda)
-                {
-                    tags.PeerService.Should().Be("dynamodb.us-east-1.amazonaws.com");
-                    tags.PeerServiceSource.Should().Be("peer.service");
-                }
-                else
-                {
-                    tags.PeerService.Should().Be("example-table");
-                    tags.PeerServiceSource.Should().Be("tablename");
-                }
-
-                break;
-            }
-
-            case "EventBridge":
-            {
-                var tags = new AwsEventBridgeTags();
-                tags.Service = "EventBridge";
-                tags.Region = "us-east-1";
-                tags.RuleName = "example-rule";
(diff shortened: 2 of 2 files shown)


tracer/src/Datadog.Trace/Util/PeerServiceHelpers.cs
```diff
@@ -0,0 +1,106 @@
...
+// </copyright>
+
+#nullable enable
+
+using Datadog.Trace.Tagging;
+
+namespace Datadog.Trace.Util
+{
+    /// <summary>
+    /// Helpers to access environment variables
```
lucaspimentel: I think this summary was copied from another class 😅

tracer/src/Datadog.Trace/Util/PeerServiceHelpers.cs
```diff
@@ -0,0 +1,106 @@
...
+    {
+        /// <summary>
+        /// Sets peer.service tag for AwsSdk spans based on the service, environment, and region
+        /// </summary>
+        /// <param name="tags">AwsSdkTags for the current span</param>
+        public static AwsSdkTags DerivePeerService(AwsSdkTags tags)
+        {
+            var service = tags.AwsService;
+            var region = tags.Region;
+            var isServerless = EnvironmentHelpers.IsAwsLambda();
```
lucaspimentel: We should rename `isServerless` to `isAwsLambda`. There are other serverless environments 😄

tracer/src/Datadog.Trace/Util/PeerServiceHelpers.cs
```diff
@@ -0,0 +1,106 @@
...
+            if (isServerless && tags.AwsRegion != null)
```
lucaspimentel: Are `tags.AwsRegion` and `tags.Region` different things?

This code is checking if `tags.AwsRegion` is null, but concatenating `region` (`tags.Region`) below without checking if `region` itself is null.

tracer/src/Datadog.Trace/Util/PeerServiceHelpers.cs
```diff
@@ -0,0 +1,106 @@
...
+                        tags.PeerService = "dynamodb." + region + ".amazonaws.com";
+                        break;
+                    case "EventBridge":
+                        tags.PeerService = "events." + region + ".amazonaws.com";
+                        break;
+                    case "Kinesis":
+                        tags.PeerService = "kinesis." + region + ".amazonaws.com";
+                        break;
+                    case "S3":
+                        var s3Tags = (AwsS3Tags)tags;
```
lucaspimentel: Even if we're certain that `tags` can only be of type `AwsS3Tags` here, we should be more defensive and confirm by using `if (tags is AwsS3Tags)` before casting.

Same for all the other casting below (`AwsDynamoDbTags`, `AwsEventBridgeTags`, etc).

tracer/src/Datadog.Trace/Util/PeerServiceHelpers.cs
rithikanarayan: I think AwsRegion gets remapped to Region but I made them consistent to be region

tracer/src/Datadog.Trace/Util/PeerServiceHelpers.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// <summary>
+        /// Sets peer.service tag for AwsSdk spans based on the service, environment, and region
+        /// </summary>
+        /// <param name="tags">AwsSdkTags for the current span</param>
+        public static AwsSdkTags DerivePeerService(AwsSdkTags tags)
+        {
+            var service = tags.AwsService;
+            var region = tags.Region;
+            var isAwsLambda = EnvironmentHelpers.IsAwsLambda();
+            if (isAwsLambda && tags.Region != null)
```
rithikanarayan: going to replace tags.Region with region (variable declared two lines above)

tracer/src/Datadog.Trace/Util/PeerServiceHelpers.cs
```diff
@@ -0,0 +1,134 @@
...
+                }
+
+                tags.PeerServiceSource = "peer.service";
+            }
+            else if (!isAwsLambda)
+            {
+                switch (service)
+                {
+                    case "DynamoDB":
+                        if (tags is AwsDynamoDbTags)
```
rithikanarayan: will replace this pattern with `if (tags is AwsDynamoDbTags dbTags)` to avoid the declaration of var dbTags in the next line

tracer/src/Datadog.Trace/Util/PeerServiceHelpers.cs
rithikanarayan: can remove returning the tags - it's a class

rithikanarayan: make isAwsLambda an argument, makes this more testable
add unit tests for DerivePeerService

lucaspimentel:
@codex review

tracer/test/Datadog.Trace.Tests/Util/PeerServiceHelpersTests.cs
```diff
@@ -0,0 +1,198 @@
...
+    [InlineData("S3", true, false)]
+    [InlineData("S3", false, false)]
+    [InlineData("S3", true, true)]
+    [InlineData("S3", false, true)]
+    [InlineData("SNS", true, false)]
+    [InlineData("SNS", false, false)]
+    [InlineData("SQS", true, false)]
+    [InlineData("SQS", false, false)]
+    [InlineData("StepFunctions", false, false)]
+    public void DerivePeerService(string service, bool isAwsLambda, bool hasBucket)
```
lucaspimentel: Since you had to use a `switch` for `service` and there's no shared code at all between the different services (and only S3 uses `hasBucket`), I'm wondering if it would be cleaner to have a separate test method for each service instead of passing it as a parameter.

Something like
```
DerivePeerServiceDynamoDB(bool isAwsLambda)
DerivePeerServiceEventBridge(bool isAwsLambda)
DerivePeerServiceS3(bool isAwsLambda, bool hasBucket)
...
```

Not a blocker to merge this PR, though.

tracer/src/Datadog.Trace/Util/PeerServiceHelpers.cs
```diff
@@ -0,0 +1,128 @@
...
+                        break;
+                    case "SNS":
+                        tags.PeerService = "sns." + region + ".amazonaws.com";
+                        break;
+                    case "SQS":
+                        tags.PeerService = "sqs." + region + ".amazonaws.com";
+                        break;
+                }
+
+                tags.PeerServiceSource = "peer.service";
```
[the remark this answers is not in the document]
rithikanarayan: Going to wrap in a check for peer.service != null

tracer/src/Datadog.Trace/Util/PeerServiceHelpers.cs
```diff
@@ -0,0 +1,128 @@
...
+        public static void DerivePeerService(AwsSdkTags tags, bool isAwsLambda)
+        {
+            var service = tags.AwsService;
+            var region = tags.Region;
+            if (isAwsLambda && region != null)
+            {
+                switch (service)
+                {
+                    case "DynamoDB":
+                        tags.PeerService = "dynamodb." + region + ".amazonaws.com";
```
lucaspimentel: [coding style]
In "modern" C# we prefer to use string interpolation instead of concatenation:
```csharp
tags.PeerService = $"dynamodb.{region}.amazonaws.com";
```

There's no performance difference, though. Just coding style preference. In both cases the compiler will convert it into the bytecode equivalent of
```csharp
tags.PeerService = string.Concat("dynamodb.", region, ".amazonaws.com");
```

PR #7853, opened 2025-11-23
Simplify dataflow synchronization

daniel-romano-DD:
## Summary of changes

This PR simplifies the lifecycle and usage of iast::Dataflow within the .NET tracer by removing unnecessary synchronization complexity and consolidating initialization logic. After validating that the profiler’s lifecycle guarantees make the race condition targeted in [#7838](https://github.com/DataDog/dd-trace-dotnet/pull/7838) impossible, the extra locking and state checks introduced there were identified as unnecessary. This PR therefore removes them and replaces them with a clearer, more maintainable initialization model.

The end result is the same runtime behavior, but with cleaner, more predictable code paths.

## Reason for change
During investigation of the suspected race in Dataflow, we discovered:

_dataflow is created before any IAST-related callbacks can run.

_dataflow is destroyed only during CorProfiler::Shutdown, after all runtime callbacks have stopped.

Therefore, there is no window where runtime threads can access a partially-initialized or null Dataflow instance.

Because the profiler’s lifecycle already prevents the problematic interleavings, the additional locking and defensive checks from #7838 were redundant. They increased complexity without improving correctness.

This PR cleans that up.

## Implementation details
What this PR does
1. Consolidates Dataflow initialization logic

Dataflow::LoadAspects becomes the single, authoritative entry point for initialization. No scattered flags, no duplicated checks.

2. Removes unnecessary locking around initialization

Since lifecycle guarantees prevent concurrent initialization, the critical-section guarding and initialization state added in #7838 are removed.

3. Simplifies callback paths

Profiler callbacks (ModuleLoaded, ModuleUnloaded, IAST instrumentation setup, etc.) now:

Assume _dataflow exists.

Only guard the internal data structures that are actually accessed concurrently.

Remove early-return branches and over-defensive locking patterns.

4. Cleans up dead code and unused members

Any state variables, flags, or locking constructs made obsolete by the simplified lifecycle are removed, shrinking the Dataflow class footprint.

Result

Same behavioral correctness.

Fewer code paths and edge cases.

Clearer lifecycle contract: Dataflow always exists during the period where callbacks may use it.

Easier to maintain and reason about for future IAST development.

## Test coverage

Re-ran IAST tests on Windows x86/x64 to confirm no regressions.

Validated profiler lifecycle behavior to ensure _dataflow is always valid during runtime callbacks.

Confirmed no reappearance of previous intermittent failures.

### DIFF
--- a/tracer/src/Datadog.Tracer.Native/cor_profiler.cpp
+++ b/tracer/src/Datadog.Tracer.Native/cor_profiler.cpp
@@ -2083,16 +2083,14 @@ int CorProfiler::RegisterIastAspects(WCHAR** aspects, int aspectsLength, UINT32 
     {
         Logger::Info("Registering Callsite Aspects.");
         dataflow->LoadAspects(aspects, aspectsLength, enabledCategories, platform);
-        if (_dataflow == nullptr)
-        {
-            _dataflow = dataflow;
-        }
+        _dataflow = dataflow;
         return aspectsLength;
     }
     else
     {
         Logger::Info("Callsite instrumentation is disabled.");
     }
+
     return 0;
 }
 


tracer/src/Datadog.Tracer.Native/cor_profiler.cpp
```diff
@@ -2072,18 +2070,23 @@ long CorProfiler::DisableCallTargetDefinitions(UINT32 disabledCategories)
 int CorProfiler::RegisterIastAspects(WCHAR** aspects, int aspectsLength, UINT32 enabledCategories, UINT32 platform)
 {
     auto _ = trace::Stats::Instance()->InitializeProfilerMeasure();
+    auto definitions = definitions_ids.Get(); // Synchronize Aspects loading
 
     auto dataflow = _dataflow;
     if (dataflow == nullptr && IsCallSiteManagedActivationEnabled())
     {
+        Logger::Info("Creating Dataflow.");
```
andrewlock: nit:
```suggestion
        Logger::Debug("Creating Dataflow.");
```

tracer/src/Datadog.Tracer.Native/cor_profiler.cpp
```diff
@@ -2072,18 +2070,23 @@ long CorProfiler::DisableCallTargetDefinitions(UINT32 disabledCategories)
...
     if (dataflow != nullptr)
     {
         Logger::Info("Registering Callsite Aspects.");
         dataflow->LoadAspects(aspects, aspectsLength, enabledCategories, platform);
-        _dataflow = dataflow;
+        if (_dataflow == nullptr)
+        {
+            _dataflow = dataflow;
+        }
         return aspectsLength;
```
andrewlock, line 2087: If `_dataflow != null` for some reason, then we wont "save" the `dataflow` values, and we might return the wrong number of aspectsLength here, right? 🤔 

It seems like we should probably move this check right to the top of the method and/or ensure _can't_ be non-null when this method is called?

tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
```diff
@@ -369,11 +365,6 @@ void Dataflow::LoadSecurityControls()
 
 HRESULT Dataflow::AppDomainShutdown(AppDomainID appDomainId)
 {
-    if (!_initialized)
```
andrewlock, line 372: Is there no scenario in which the app could shut down _before_ we call `LoadAspects`? 🤔 It seems like it could, and that keeping these checks here would be safest?

The "fix" is that because `_initialized` can't go from `true -> false`, only `false -> true`, then you're "safe" once you've made the check and hit the CSGUARD. But I think you must still make sure not to call these if `LoadAspects` hasn't been called?

dromanol: In the end it does not make any difference, as it will remove the app domain from the dictionary if it was there (and do nothing if not). IMO it's better to keep simplicity. Also, this is how it worked before the first stable config refactor

tracer/src/Datadog.Tracer.Native/cor_profiler.cpp
dromanol: Agree. On it

PR #7856, opened 2025-11-24
process tags : extra-careful normalization

vandonr:
## Summary of changes

 - replace potential `:` in values _before_ we call the normalize method, so that it can squash them with other special chars if present
 - truncate to 100 char to avoid transmitting unnecessary bytes

## Reason for change

image

## Test coverage

I need to find a way to inject arbitrary values 😬


tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -64,8 +64,16 @@ private static string GetSerializedTags()
 
     private static string NormalizeTagValue(string tagValue)
```
andrewlock: Can we add some basic unit tests for this to demonstrate/catch this issue and prevent regressions?

```suggestion
    [TestingAndPrivateOnly]
    internal static string NormalizeTagValue(string tagValue)
```

That way I'd be more comfortable optimizing this later (e.g there's no point normalizing characters that are going to be subsequently truncated)

vandonr: ooh I didn't know about this annotation

vandonr: about normalizing truncated characters, you are right, but the normalization may shorten a string (because clusters of éspecial characters" get replaced by one single `_`), which is why I apply the truncation after the normalization.
We'd need to embed the character limit into the normalization method. Which could be a good idea for other usages too though.

tracer/test/Datadog.Trace.Tests/ProcessTagsTests.cs
```diff
@@ -30,4 +30,61 @@ public void TagsPresentWhenEnabled()
...
+    [InlineData("123456", "")]
+    [InlineData("7.0", "")] // this is not ideal
+    [InlineData("ünicöde", "ünicöde")]
+    [InlineData("ünicöde:metäl", "ünicöde_metäl")]
+    [InlineData("Data🐨dog🐶 繋がっ⛰てて", "data_dog_繋がっ_てて")]
+    [InlineData(" spaces   ", "spaces")]
+    [InlineData(" #hashtag!@#spaces #__<>#  ", "hashtag_spaces")]
+    [InlineData(":testing", "testing")]
+    [InlineData("_foo", "foo")]
+    [InlineData(":::test", "test")]
```
andrewlock, line 50: ~Wouldn't we expect this to be `_test` (for example) 🤔?~

Ah, I see, we trim leading `_` too, fair enough.

You should probably add a couple of cases that explicitly test the code paths you added in the process tags method, e.g. 🙂 
```suggestion
    [InlineData(":::test", "test")]
    [InlineData("te::st", "te_st")]
    [InlineData("te:_st", "te_st")]
    [InlineData("te:🐨st", "te_st")]
```

tracer/test/Datadog.Trace.Tests/ProcessTagsTests.cs
```diff
@@ -30,4 +30,61 @@ public void TagsPresentWhenEnabled()
...
+    [InlineData("a��b", "a_b")]
+    public void TestNormalization(string tagValue, string expectedValue)
+    {
+        ProcessTags.NormalizeTagValue(tagValue).Should().Be(expectedValue);
+    }
+
+    [Fact]
+    public void TestNormalizationsTruncation()
+    {
+        // cannot write those as `Theory` because the parameters need to be constant values
```
andrewlock, line 84: Meh, that's fine 😄 
```suggestion
```

vandonr: I just figured it'd save someone some time, if they come across this code and are like "well, this test could be merged with the one above" ^^

tracer/test/Datadog.Trace.Tests/ProcessTagsTests.cs
vandonr: yes fair, I actually took the test cases from [the ruby PR](https://github.com/DataDog/dd-trace-rb/pull/5033/files#diff-f75d18d7d50afc15d91882b2f3ee73b4d92ca459e499ad4bf9c14863291f8251) and... adapted the "expected" values to what's actually happening 🫠

tracer/test/Datadog.Trace.Tests/ProcessTagsTests.cs
andrewlock: Well, it _could_ be merged with the one above 😉

vandonr: like... with a generator method ?

tracer/test/Datadog.Trace.Tests/ProcessTagsTests.cs
andrewlock: > adapted the "expected" values to what's actually happening

wait... what? surely that means that one of the implementations is incorrect?!

tracer/test/Datadog.Trace.Tests/ProcessTagsTests.cs
andrewlock: Like, type `a` 100 times 😝

tracer/test/Datadog.Trace.Tests/ProcessTagsTests.cs
vandonr: it's because I relied on the existing code that cleans tags, while in Ruby, some custom code was written.
It's actually not as bad as it seems, because the difference is mostly on corner cases, and I don't think any customer has expectations that strings of special characters would be treated the same in their ruby server name and in their IIS application.

andrewlock: Whatever is the _correct_ implementation is what we should implement

vandonr: I see the purity of your remark, but the correct implementation is the one defined in the spec, and the spec is extremely vague on this topic. Also, the "reference implementation" is supposed to be Java, and Java doesn't even respect the spec at the moment, so...

PR #7858, opened 2025-11-24
[AWS Lambda] Cleanup from PR#7835

rithikanarayan:
## Summary of changes
Cast object state to CallTargetState earlier in its usage to add type safety. Added type checks combined with variable declarations when using the state and scope attributes of CallTargetState.

## Reason for change
Merged [earlier PR adding request id headers](https://github.com/DataDog/dd-trace-dotnet/pull/7835) to start and end invocation calls to the lambda extension, needed to improve code quality post-merge.

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Lambda/HandlerWrapperSetHandlerIntegration.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Lambda/HandlerWrapperSetHandlerIntegration.cs
@@ -71,7 +71,7 @@ public class HandlerWrapperSetHandlerIntegration
             LambdaCommon.Log("DelegateWrapper Running OnDelegateBegin");
 
             Scope scope;
-            object requestid = null;
+            object requestId = null;
             var proxyInstance = arg.DuckCast<IInvocationRequest>();
             if (proxyInstance == null)
             {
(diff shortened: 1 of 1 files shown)


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Lambda/HandlerWrapperSetHandlerIntegration.cs
```diff
@@ -104,27 +104,31 @@ public TReturn OnDelegateEnd<TReturn>(object sender, TReturn returnValue, Except
         public async Task<TInnerReturn> OnDelegateEndAsync<TInnerReturn>(object sender, TInnerReturn returnValue, Exception exception, object state)
         {
             LambdaCommon.Log("DelegateWrapper Running OnDelegateEndAsync");
-            try
+            if (state is CallTargetState callTargetState)
```
andrewlock, line 107: We should never pass `CallTargetState` to an `object` parameter, because it will cause boxing and extra allocation 🙂 Unfortunately I'm not very familiar with this integration, so I can't suggest a great alternative, but if you need to pass two different types of `state` object, it's probably best to just create two different methods with the two possible parameter types

rithikanarayan: I don't think it's possible to pass in CallTargetState directly (or avoid passing in an object) because this function comes from DelegateInstrumentation.cs, which is not a lambda-specific file and specifies that an object is passed in to this function. The state is already boxed when it comes to us.

zacharycmontoya: It looks like we may need to follow up to make sure that the `CallTargetState` is always passed as a strongly typed `CallTargetState` rather than `object?`, for the sake of performance. I think that change is out of scope for this PR.

Like Rithika pointed out, in `DelegateInstrumentation.cs`, the various callbacks are defined as holding the state as `object state`, but in this file the various `OnDelegateBegin` methods do in fact return a `CallTargetState` object, even though the method signature says `object`

PR #7860, opened 2025-11-24
Bump the version of `Microsoft.CodeAnalysis.CSharp` and others used in our internal analyzers

andrewlock:
## Summary of changes

- Update `Microsoft.CodeAnalysis.CSharp`, `Microsoft.CodeAnalysis.CSharp.Workspaces`, and `Microsoft.CodeAnalysis.Analyzers` to latest
- Split the code fixes into a separate project

## Reason for change

We should update these dependencies as they're old. Unfortunately, doing so hits [this build error](https://github.com/dotnet/roslyn-analyzers/blob/main/docs/rules/RS1038.md):

> **RS1038: Compiler extensions should be implemented in assemblies with compiler-provided references**
> Types which implement compiler extension points should not be declared in assemblies that contain references to assemblies which are not provided by all compilation scenarios. Doing so may cause the feature to behave unpredictably.

tl;dr; codefixes need to be in a separate assembly to analyzers and source generators 🙄 

## Implementation details

- Bump the versions
- Extract diagnostic ID definitions to standalone types so they can be easily linked to
- Move the code fixes to a separate project
- Add links back to the analyzer project where appropriate
- Minor tweaks so everything still builds

## Test coverage

As long as the unit tests pass I think we're good

## Other details

### DIFF
--- a/tracer/test/Datadog.Trace.SourceGenerators.Tests/Datadog.Trace.SourceGenerators.Tests.csproj
+++ b/tracer/test/Datadog.Trace.SourceGenerators.Tests/Datadog.Trace.SourceGenerators.Tests.csproj
@@ -6,8 +6,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.8.0" PrivateAssets="all" />
-    <PackageReference Include="System.Text.Json" Version="10.0.0-rc.2.25502.107" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="5.0.0" PrivateAssets="all" />
+    <PackageReference Include="System.Text.Json" Version="10.0.0" />
   </ItemGroup>
 
   <ItemGroup>


tracer/src/Datadog.Trace.Tools.Analyzers.CodeFixes/DuckTypeAnalyzer/DuckTypeNullCheckCodeFixProvider.cs
andrewlock, line 1: These are all just moving from the analyzer project to the code fix project

tracer/src/Datadog.Trace.Tools.Analyzers.CodeFixes/Helpers/ObjectDisplay.cs
andrewlock, line 1: This was a public type provided in the analyzers package, so just basically vendoring the minimum amount needed here

tracer/src/Datadog.Trace.Tools.Analyzers/Datadog.Trace.Tools.Analyzers.csproj
```diff
@@ -8,9 +8,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.4" PrivateAssets="all" />
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.8.0" PrivateAssets="all" />
-	  <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.8.0" PrivateAssets="all" />
```
andrewlock, line 13: This is the important bit, removing the reference to `Microsoft.CodeAnalysis.CSharp.Workspaces` from this project

NachoEchevarria (approved):

PR #7865, opened 2025-11-26
Add more cleanup to benchmarks

andrewlock:
## Summary of changes

- Update the JetBrains SelfApi package to latest version
- Don't run background tasks like remote config, telemetry, discovery service in benchmarks
- Cleanup "global" tracers after benchmark

## Reason for change

Benchmarks should be isolated to keep them consistent, but unfortunately a lot of our code has `Tracer.Instance` and `Security.Instance` scattered around. That means we need to set up _global_ state for some of the benchmarks. Currently, we're never cleaning up that global state, which could introduce variance (flake) in the benchmarks).

> Note that this actually doesn't impact us in _most_ cases, only when using the inprocess tool chain, which is primarily used for when we're using dotmemory etc locally

Similarly, new `Tracer` instances start a bunch of background tasks, such as remote configuration, telemetry, discovery service. These all have the possibility of skewing results and introducing flake.
 
## Implementation details

- Cleanup any resources (e.g. `Tracer` instances, but anything disposable) in `[GlobalCleanup]`
- Ensure we _remove_ and replace the global `Tracer` instances in cleanup
- Disable services that start background jobs (discovery service, telemetry, remote config)
- Create a helper for creating a tracer with default benchmarking configuration

## Test coverage

This is the test - we may see some variation from the current results, but hopefully the results will be more stable subsequently

## Other details

I _wish_ we could not use `Tracer.Instance` (or `Security.Instance`, or `Iast.Instance`) at all. It would make all of this much easier to reason about and reduce the chance of accidental cross-talk. That's something we'll move towards slowly, for now this is probably the best we can do.

### DIFF
--- a/tracer/test/benchmarks/Benchmarks.Trace/TracerHelper.cs
+++ b/tracer/test/benchmarks/Benchmarks.Trace/TracerHelper.cs
@@ -34,7 +34,7 @@ public static class TracerHelper
         Tracer.UnsafeSetTracerInstance(CreateTracer(config));
     }
 
-    public static void CleanupGlobalTracer(Dictionary<string, object> config = null)
+    public static void CleanupGlobalTracer()
     {
         Tracer.Instance.TracerManager.ShutdownAsync().GetAwaiter().GetResult();
         Tracer.UnsafeSetTracerInstance(null);


tracer/test/benchmarks/Benchmarks.Trace/TracerHelper.cs
```diff
@@ -0,0 +1,43 @@
...
+
+    public static Tracer CreateTracer(TracerSettings settings)
+        => new(settings, new DummyAgentWriter(), null, null, null, NullTelemetryController.Instance, NullDiscoveryService.Instance);
+
+    public static void SetGlobalTracer(Dictionary<string, object> config = null)
+    {
+        Tracer.UnsafeSetTracerInstance(CreateTracer(config));
+    }
+
+    public static void CleanupGlobalTracer(Dictionary<string, object> config = null)
```
NachoEchevarria: Guess we can remove config here as it is not used.

andrewlock:
```suggestion
    public static void CleanupGlobalTracer()
```

tracer/test/benchmarks/Benchmarks.Trace/AgentWriterBenchmark.cs
```diff
@@ -39,13 +39,9 @@ public void GlobalSetup()
...
 
-            var overrides = new NameValueConfigurationSource(new()
-            {
-                { ConfigurationKeys.StartupDiagnosticLogEnabled, false.ToString() },
-                { ConfigurationKeys.TraceEnabled, false.ToString() },
-            });
-            var sources = new CompositeConfigurationSource(new[] { overrides, GlobalConfigurationSource.Instance });
-            var settings = new TracerSettings(sources);
+            var config = TracerHelper.DefaultConfig;
+            config.Add(ConfigurationKeys.TraceEnabled, false);
```
lucaspimentel, line 49: `StartupDiagnosticLogEnabled` is no longer disabled. Was that intentional?

lucaspimentel: Never mind, reading the diffs top to bottom 😅

PR #7868, opened 2025-11-26
Benchmark improvements

andrewlock:
## Summary of changes

- Remove the obsolete `[AgentFilter]`
- Allow running some benchmarks on master only (not PRs)

## Reason for change

The `[AgentFilter]` is left over from before we moved to benchmarking platform, so delete it. 

Additionally some benchmarks don't seem like they need to be run _every_ PR, so added separate categories to allow controlling this.

> [!IMPORTANT]
> This PR _doesn't_ change which benchmarks we run on PR vs master, it just _allows_ us to do that easily by adding/removing a category

## Implementation details

- Delete the old filter
- Add new categories - add them to every benchmark currently tagged with tracer/ASM
- Change how the filtering works in `RunBenchmarks` Nuke target to work with this
- Requires benchmarking platform update:
  - https://github.com/DataDog/benchmarking-platform/pull/215

## Test coverage

This has the same coverage as today

### DIFF
--- a/.gitlab/benchmarks/microbenchmarks.yml
+++ b/.gitlab/benchmarks/microbenchmarks.yml
@@ -97,7 +97,7 @@ run-benchmarks:
     # Branch containing 1. scripts to launch Windows benchmarks on ephemeral 
     # instances (to be used by GitLab CI runners) and 2. scripts to run Windows 
     # benchmarks (to be used by the ephemeral instances).
-    BP_INFRA_BENCHMARKING_PLATFORM_BRANCH: "andrew/benchmarks"
+    BP_INFRA_BENCHMARKING_PLATFORM_BRANCH: "dd-trace-dotnet/micro"
 
     # Where benchmarking results will be stored
     BP_INFRA_ARTIFACTS_BUCKET_NAME: "windows-benchmarking-results-us-east-1"


.gitlab/benchmarks/microbenchmarks.yml
```diff
@@ -97,7 +97,7 @@ run-benchmarks:
     # Branch containing 1. scripts to launch Windows benchmarks on ephemeral 
     # instances (to be used by GitLab CI runners) and 2. scripts to run Windows 
     # benchmarks (to be used by the ephemeral instances).
-    BP_INFRA_BENCHMARKING_PLATFORM_BRANCH: "dd-trace-dotnet/micro"
```
NachoEchevarria, line 100: Guess that this change should be reverted after merging https://github.com/DataDog/benchmarking-platform/pull/215

tracer/test/benchmarks/Benchmarks.Trace/Program.cs
```diff
@@ -62,21 +64,16 @@ private static int Main(string[] args)
...
-            {
-                var attributeName = $"{benchmarkAgent}Attribute";
-                Console.WriteLine($"Found agent name {agentName}; executing only benchmarks decorated with '{attributeName}'");
-                config = config.AddFilter(new AttributesFilter(new[] { attributeName }));
-            }
-            else
+            if (args?.Any(a => a == isPrTrue) == true)
             {
-                Console.WriteLine($"Unknown agent name {agentName}; executing all benchmarks");
+                Console.WriteLine("Setting Jetbrains trace collection... (could take time downloading collector binaries)");
```
NachoEchevarria: If it's a PR, we remove jetBrainsDotTrace from args and display "Setting Jetbrains trace collection..." Is that ok?

tracer/build/_build/Build.cs
```diff
@@ -602,13 +602,22 @@ void DeleteReparsePoints(string path)
...
+                    };
+
                     DotNetRun(s => s
                         .SetProjectFile(benchmarksProject)
                         .SetConfiguration(BuildConfiguration)
                         .SetFramework(framework)
                         .EnableNoRestore()
                         .EnableNoBuild()
-                        .SetApplicationArguments($"-r {runtimes} -m -f {Filter ?? "*"} --anyCategories {BenchmarkCategory ?? "tracer"} --iterationTime 200")
+                        .SetApplicationArguments($"-r {runtimes} -m -f {Filter ?? "*"} -is_pr:{isPr} --allCategories {categories} --iterationTime 200")
```
NachoEchevarria: We are just passing this new is_pr arg but not using it (for now) right?

.gitlab/benchmarks/microbenchmarks.yml
andrewlock: yep!

tracer/test/benchmarks/Benchmarks.Trace/Program.cs
andrewlock: It looks like I pushed some changes that were WIP I didn't _mean_ to, I'll fix this and the next one 😅

PR #7871, opened 2025-11-26
Add process tags to runtime metrics & dynamic instrumentation metric probes

vandonr:
## Summary of changes

add process tags as constant tags for both runtime metrics and the metrics sent by dynamic instrumentation probes

## Reason for change

AIDM-195

## Test coverage

Will append existing integration tests

### DIFF
--- a/tracer/src/Datadog.Trace/DogStatsd/StatsdFactory.cs
+++ b/tracer/src/Datadog.Trace/DogStatsd/StatsdFactory.cs
@@ -18,7 +18,12 @@ internal static class StatsdFactory
 {
     private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(StatsdFactory));
 
-    internal static IDogStatsd CreateDogStatsdClient(MutableSettings settings, ExporterSettings exporter, bool includeDefaultTags, IReadOnlyCollection<string> processTags, string? prefix = null)
+    internal static IDogStatsd CreateDogStatsdClient(
+        MutableSettings settings,
+        ExporterSettings exporter,
+        bool includeDefaultTags,
+        IReadOnlyCollection<string> processTags,
+        string? prefix = null)
     {
         var customTagCount = settings.GlobalTags.Count;
         var tagsCount = (includeDefaultTags ? 5 + customTagCount : 0) + processTags.Count;
@@ -53,7 +58,11 @@ internal static class StatsdFactory
         return CreateDogStatsdClient(settings, exporter, constantTags, prefix);
     }
 
-    private static IDogStatsd CreateDogStatsdClient(MutableSettings settings, ExporterSettings exporter, List<string>? constantTags, string? prefix = null)
+    private static IDogStatsd CreateDogStatsdClient(
+        MutableSettings settings,
+        ExporterSettings exporter,
+        List<string>? constantTags,
+        string? prefix = null)
     {
         try
         {
--- a/tracer/src/Datadog.Trace/DogStatsd/StatsdManager.cs
+++ b/tracer/src/Datadog.Trace/DogStatsd/StatsdManager.cs
@@ -34,9 +34,11 @@ internal sealed class StatsdManager : IStatsdManager
     }
 
     // Internal for testing
-    internal StatsdManager(TracerSettings tracerSettings, Func<MutableSettings, ExporterSettings, IReadOnlyCollection<string>, StatsdClientHolder> statsdFactory)
+    internal StatsdManager(
+        TracerSettings tracerSettings,
+        Func<MutableSettings, ExporterSettings, IReadOnlyCollection<string>, StatsdClientHolder> statsdFactory)
     {
-        // The initial factory, assuming there's no updates
+        // The initial factory, assuming there are no updates
         _factory = () => statsdFactory(
             tracerSettings.Manager.InitialMutableSettings,
             tracerSettings.Manager.InitialExporterSettings,
@@ -176,9 +178,13 @@ internal sealed class StatsdManager : IStatsdManager
         return hasChanges;
     }
 
-    private static StatsdClientHolder CreateClient(MutableSettings settings, ExporterSettings exporter, IReadOnlyCollection<string> processTags)
+    private static StatsdClientHolder CreateClient(
+        MutableSettings settings,
+        ExporterSettings exporter,
+        IReadOnlyCollection<string> processTags)
     {
-        return new StatsdClientHolder(StatsdFactory.CreateDogStatsdClient(settings, exporter, includeDefaultTags: true, processTags));
+        var client = StatsdFactory.CreateDogStatsdClient(settings, exporter, includeDefaultTags: true, processTags);
+        return new StatsdClientHolder(client);
     }
 
     private void EnsureClient(bool ensureCreated, bool forceRecreate)
--- a/tracer/src/Datadog.Trace/RemoteConfigurationManagement/Protocol/RcmClientTracer.cs
+++ b/tracer/src/Datadog.Trace/RemoteConfigurationManagement/Protocol/RcmClientTracer.cs
@@ -16,16 +16,23 @@ namespace Datadog.Trace.RemoteConfigurationManagement.Protocol
     {
         // Don't change this constructor - it's used by Newtonsoft.JSON for deserialization
         // and that can mean the provided properties are not _really_ nullable, even though we "require" them to be
-        public RcmClientTracer(string runtimeId, string tracerVersion, string service, string env, string? appVersion, List<string> tags, IReadOnlyCollection<string>? processTags)
+        public RcmClientTracer(
+            string runtimeId,
+            string tracerVersion,
+            string service,
+            string env,
+            string? appVersion,
+            List<string> tags,
+            IReadOnlyCollection<string>? processTags)
         {
             RuntimeId = runtimeId;
-            Language = TracerConstants.Language;
             TracerVersion = tracerVersion;
+            Language = TracerConstants.Language;
             Service = service;
-            ProcessTags = processTags;
             Env = env;
             AppVersion = appVersion;
             Tags = tags ?? [];
+            ProcessTags = processTags;
         }
 
         [JsonIgnore]
@@ -58,7 +65,14 @@ namespace Datadog.Trace.RemoteConfigurationManagement.Protocol
         [JsonProperty("tags")]
         public List<string> Tags { get; }
 
-        public static RcmClientTracer Create(string runtimeId, string tracerVersion, string service, string env, string? appVersion, ReadOnlyDictionary<string, string> globalTags, IReadOnlyCollection<string>? processTags)
+        public static RcmClientTracer Create(
+            string runtimeId,
+            string tracerVersion,
+            string service,
+            string env,
+            string? appVersion,
+            ReadOnlyDictionary<string, string> globalTags,
+            IReadOnlyCollection<string>? processTags)
             => new(runtimeId, tracerVersion, service, env, appVersion, GetTags(env, service, globalTags), processTags);
 
         private static List<string> GetTags(string? environment, string? serviceVersion, ReadOnlyDictionary<string, string>? globalTags)
--- a/tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
+++ b/tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
@@ -1392,7 +1392,11 @@ namespace Datadog.Trace.Configuration
         internal static TracerSettings Create(Dictionary<string, object?> settings)
             => Create(settings, LibDatadogAvailabilityHelper.IsLibDatadogAvailable);
 
-        internal static TracerSettings Create(Dictionary<string, object?> settings, LibDatadogAvailableResult isLibDatadogAvailable)
-            => new(new DictionaryConfigurationSource(settings.ToDictionary(x => x.Key, x => x.Value?.ToString()!)), new ConfigurationTelemetry(), new(), isLibDatadogAvailable);
+        internal static TracerSettings Create(Dictionary<string, object?> settings, LibDatadogAvailableResult isLibDatadogAvailable) =>
+            new(
+                new DictionaryConfigurationSource(settings.ToDictionary(x => x.Key, x => x.Value?.ToString()!)),
+                new ConfigurationTelemetry(),
+                new OverrideErrorLog(),
+                isLibDatadogAvailable);
     }
 }
--- a/tracer/test/Datadog.Trace.Tests/DogStatsd/StatsdManagerTests.cs
+++ b/tracer/test/Datadog.Trace.Tests/DogStatsd/StatsdManagerTests.cs
@@ -732,7 +732,7 @@ public class StatsdManagerTests
         using var manager = new StatsdManager(settings, (_, _, processTags) =>
         {
             capturedProcessTags = processTags;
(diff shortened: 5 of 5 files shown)


dudikeleti (approved):

tracer/src/Datadog.Trace/Debugger/DebuggerFactory.cs
```diff
@@ -72,6 +72,7 @@ private static IDogStatsd GetDogStatsd(TracerSettings tracerSettings, string ser
                 tracerSettings.Manager.InitialMutableSettings,
                 tracerSettings.Manager.InitialExporterSettings,
                 includeDefaultTags: false,
+                tracerSettings.PropagateProcessTags ? ProcessTags.TagsList : [],
```
lucaspimentel, line 75: Since the `processTags` parameter is an `IList<string>`, the compiler will to allocate a new `List<string>` here when using `[]`. I think you can use `Array.Empty<string>()` instead to use the cached empty array.

Also consider changing the `processTags` parameter type from `IList<T>` to `ICollection<T>`. Even though `IList<T>` works, `ICollection<T>` is safer since it exposes the `Count` property (which we use) but does not expose methods that change the collection's size like `Add()` and `Remove()`, which would fail on an array.

tracer/src/Datadog.Trace/DogStatsd/StatsdFactory.cs
```diff
@@ -18,19 +18,19 @@ internal static class StatsdFactory
 {
     private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(StatsdFactory));
 
-    internal static IDogStatsd CreateDogStatsdClient(MutableSettings settings, ExporterSettings exporter, bool includeDefaultTags, string? prefix = null)
+    internal static IDogStatsd CreateDogStatsdClient(MutableSettings settings, ExporterSettings exporter, bool includeDefaultTags, IList<string> processTags, string? prefix = null)
```
lucaspimentel: We're only accessing `processTags.Count`, so `ICollection<string>` is enough and makes it safer to pass an `string[]` if needed, like in my other comment about using `[]`.

(`string[]` implements `IList<string>`, but some methods like `Add()` or `Remove()` will throw an exception)

```suggestion
    internal static IDogStatsd CreateDogStatsdClient(MutableSettings settings, ExporterSettings exporter, bool includeDefaultTags, ICollection<string> processTags, string? prefix = null)
```

tracer/src/Datadog.Trace/Debugger/DebuggerFactory.cs
lucaspimentel: Not a big deal, though, if it means changing a ton of code downstream, since this empty list would only be created once and then thrown away. But every little allocation counts at this point 😅  Your call.

lucaspimentel (approved):

tracer/src/Datadog.Trace/Debugger/DebuggerFactory.cs
vandonr: thanks for the pro tips, I figured dotnet would be smart enough to handle `[]` properly without creating a new object and then throwing it away, but I guess I was wrong ah ah

lucaspimentel: It tries to be smart when possible, but with `IList<T>` is has no other option than to create a new object every time. The list can be mutated so you don't want to reuse the same one. An empty array can't be mutated so it can be cached and reused.

tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -21,7 +21,7 @@ internal static class ProcessTags
     public const string EntrypointWorkdir = "entrypoint.workdir";
 
     // two views on the same data
-    public static readonly List<string> TagsList = GetTagsList();
+    public static readonly ReadOnlyCollection<string> TagsList = GetTagsList().AsReadOnly();
```
lucaspimentel: (I was going to leave comments in all the other places you used `[]` to change them to `Array.Empty<string>()` like in `DebuggerFactory`, but instead...)

I tried a few things to make it easier to avoid the heap allocation with the `[]` initializer. Turns out if you change this type to `IReadOnlyCollection<string>` then `[]` will use `Array.Empty<string>()` (since it's guaranteed that the collection won't be mutated).

If you do that, you also don't need `AsReadOnly()` here because `List<T>` implements `IReadOnlyCollection<T>`, avoiding another allocation for the `ReadOnlyCollection<T>`. And you can change that `Array.Empty<string>()` in `DebuggerFactory` back to `[]` the way you had it originally and keep all the other `[]` 😅 

You'll also have to change `IList<string>` to `IReadOnlyCollection<T>` in a few other places to fix compiler errors. I'll leave suggestions for you.

```suggestion
    public static readonly IReadOnlyCollection<string> TagsList = GetTagsList();
```

tracer/src/Datadog.Trace/Debugger/DebuggerFactory.cs
```diff
@@ -72,6 +72,7 @@ private static IDogStatsd GetDogStatsd(TracerSettings tracerSettings, string ser
                 tracerSettings.Manager.InitialMutableSettings,
                 tracerSettings.Manager.InitialExporterSettings,
                 includeDefaultTags: false,
+                tracerSettings.PropagateProcessTags ? ProcessTags.TagsList : Array.Empty<string>(),
```
lucaspimentel: See the comment in `tracer/src/Datadog.Trace/ProcessTags.cs`.
```suggestion
                tracerSettings.PropagateProcessTags ? ProcessTags.TagsList : [],
```

tracer/src/Datadog.Trace/DogStatsd/StatsdFactory.cs
```diff
@@ -18,19 +18,19 @@ internal static class StatsdFactory
 {
     private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(StatsdFactory));
 
-    internal static IDogStatsd CreateDogStatsdClient(MutableSettings settings, ExporterSettings exporter, bool includeDefaultTags, string? prefix = null)
+    internal static IDogStatsd CreateDogStatsdClient(MutableSettings settings, ExporterSettings exporter, bool includeDefaultTags, ICollection<string> processTags, string? prefix = null)
```
lucaspimentel: See the comment in `tracer/src/Datadog.Trace/ProcessTags.cs`.
```suggestion
    internal static IDogStatsd CreateDogStatsdClient(MutableSettings settings, ExporterSettings exporter, bool includeDefaultTags, IReadOnlyCollection<string> processTags, string? prefix = null)
```

tracer/src/Datadog.Trace/DogStatsd/StatsdManager.cs
```diff
@@ -33,12 +34,13 @@ public StatsdManager(TracerSettings tracerSettings)
     }
 
     // Internal for testing
-    internal StatsdManager(TracerSettings tracerSettings, Func<MutableSettings, ExporterSettings, StatsdClientHolder> statsdFactory)
+    internal StatsdManager(TracerSettings tracerSettings, Func<MutableSettings, ExporterSettings, IList<string>, StatsdClientHolder> statsdFactory)
```
lucaspimentel: See the comment in `tracer/src/Datadog.Trace/ProcessTags.cs`.
```suggestion
    internal StatsdManager(TracerSettings tracerSettings, Func<MutableSettings, ExporterSettings, IReadOnlyCollection<string>, StatsdClientHolder> statsdFactory)
```

tracer/src/Datadog.Trace/DogStatsd/StatsdManager.cs
```diff
@@ -173,8 +176,10 @@ internal static bool HasImpactingChanges(TracerSettings.SettingsManager.SettingC
         return hasChanges;
     }
 
-    private static StatsdClientHolder CreateClient(MutableSettings settings, ExporterSettings exporter)
-        => new(StatsdFactory.CreateDogStatsdClient(settings, exporter, includeDefaultTags: true));
+    private static StatsdClientHolder CreateClient(MutableSettings settings, ExporterSettings exporter, IList<string> processTags)
```
lucaspimentel: See the comment in `tracer/src/Datadog.Trace/ProcessTags.cs`.
```suggestion
    private static StatsdClientHolder CreateClient(MutableSettings settings, ExporterSettings exporter, IReadOnlyCollection<string> processTags)
```

tracer/src/Datadog.Trace/RemoteConfigurationManagement/RemoteConfigurationManager.cs
```diff
@@ -42,7 +42,7 @@ private RemoteConfigurationManager(
             TimeSpan pollInterval,
             IGitMetadataTagsProvider gitMetadataTagsProvider,
             IRcmSubscriptionManager subscriptionManager,
-            List<string>? processTags)
+            IList<string>? processTags)
```
lucaspimentel: See the comment in `tracer/src/Datadog.Trace/ProcessTags.cs`.
```suggestion
            IReadOnlyCollection<string>? processTags)
```

tracer/src/Datadog.Trace/RemoteConfigurationManagement/Protocol/RcmClientTracer.cs
```diff
@@ -60,7 +58,7 @@ public RcmClientTracer(string runtimeId, string tracerVersion, string service, s
         [JsonProperty("tags")]
         public List<string> Tags { get; }
 
-        public static RcmClientTracer Create(string runtimeId, string tracerVersion, string service, string env, string? appVersion, ReadOnlyDictionary<string, string> globalTags, List<string>? processTags)
+        public static RcmClientTracer Create(string runtimeId, string tracerVersion, string service, string env, string? appVersion, ReadOnlyDictionary<string, string> globalTags, IList<string>? processTags)
```
lucaspimentel: See the comment in `tracer/src/Datadog.Trace/ProcessTags.cs`.
```suggestion
        public static RcmClientTracer Create(string runtimeId, string tracerVersion, string service, string env, string? appVersion, ReadOnlyDictionary<string, string> globalTags, IReadOnlyCollection<string>? processTags)
```

tracer/src/Datadog.Trace/RemoteConfigurationManagement/Protocol/RcmClientTracer.cs
```diff
@@ -10,15 +10,13 @@
...
-#nullable enable
-
 namespace Datadog.Trace.RemoteConfigurationManagement.Protocol
 {
     internal sealed class RcmClientTracer
     {
         // Don't change this constructor - it's used by Newtonsoft.JSON for deserialization
         // and that can mean the provided properties are not _really_ nullable, even though we "require" them to be
-        public RcmClientTracer(string runtimeId, string tracerVersion, string service, string env, string? appVersion, List<string> tags, List<string>? processTags)
+        public RcmClientTracer(string runtimeId, string tracerVersion, string service, string env, string? appVersion, List<string> tags, IList<string>? processTags)
```
lucaspimentel: See the comment in `tracer/src/Datadog.Trace/ProcessTags.cs`.
```suggestion
        public RcmClientTracer(string runtimeId, string tracerVersion, string service, string env, string? appVersion, List<string> tags, IReadOnlyCollection<string>? processTags)
```

tracer/src/Datadog.Trace/RemoteConfigurationManagement/Protocol/RcmClientTracer.cs
```diff
@@ -46,7 +44,7 @@ public RcmClientTracer(string runtimeId, string tracerVersion, string service, s
         public string? Service { get; }
 
         [JsonProperty("process_tags")]
-        public List<string>? ProcessTags { get; }
+        public IList<string>? ProcessTags { get; }
```
lucaspimentel: See the comment in `tracer/src/Datadog.Trace/ProcessTags.cs`.
```suggestion
        public IReadOnlyCollection<string>? ProcessTags { get; }
```

tracer/src/Datadog.Trace/Debugger/DebuggerFactory.cs
lucaspimentel: I changed this back to `[]` in 51ae1df1fbdec84859e60a2fbee00dd6c5ca8ebc. See the comment in `tracer/src/Datadog.Trace/ProcessTags.cs` for details.

tracer/src/Datadog.Trace/ProcessTags.cs
lucaspimentel: Done in 51ae1df1fbdec84859e60a2fbee00dd6c5ca8ebc.

tracer/src/Datadog.Trace/Debugger/DebuggerFactory.cs
lucaspimentel: Done in 51ae1df1fbdec84859e60a2fbee00dd6c5ca8ebc.

tracer/src/Datadog.Trace/DogStatsd/StatsdFactory.cs
lucaspimentel: Done in 51ae1df1fbdec84859e60a2fbee00dd6c5ca8ebc.

tracer/src/Datadog.Trace/DogStatsd/StatsdManager.cs
lucaspimentel: Done in 51ae1df1fbdec84859e60a2fbee00dd6c5ca8ebc.

tracer/src/Datadog.Trace/RemoteConfigurationManagement/RemoteConfigurationManager.cs
lucaspimentel: Done in 51ae1df1fbdec84859e60a2fbee00dd6c5ca8ebc.

tracer/src/Datadog.Trace/RemoteConfigurationManagement/Protocol/RcmClientTracer.cs
lucaspimentel: Done in 51ae1df1fbdec84859e60a2fbee00dd6c5ca8ebc.

tracer/src/Datadog.Trace/RemoteConfigurationManagement/Protocol/RcmClientTracer.cs
lucaspimentel: Done in 51ae1df1fbdec84859e60a2fbee00dd6c5ca8ebc.

tracer/src/Datadog.Trace/RemoteConfigurationManagement/Protocol/RcmClientTracer.cs
lucaspimentel: Done in 51ae1df1fbdec84859e60a2fbee00dd6c5ca8ebc.

tracer/src/Datadog.Trace/DogStatsd/StatsdManager.cs
lucaspimentel: Done in 51ae1df1fbdec84859e60a2fbee00dd6c5ca8ebc.

PR #7874, opened 2025-11-27
Reduce allocations in `TraceContext`

andrewlock:
## Summary of changes

Replace `ArraySegment<Span>` with `SpanCollection`

## Reason for change

`ArraySegment<Span>` is a wrapper around an array, and by design, always requires allocating an array. 

However, if we look at the [distribution of traces](https://app.datadoghq.com/notebook/13270656/partial-flush-worth-it-or-not?computational_onboarding=true&fullscreen_end_ts=1760111663000&fullscreen_paused=true&fullscreen_refresh_mode=paused&fullscreen_start_ts=1759829808000&fullscreen_widget=3c8idtpl) received for processing, we see that 50% of traces only have a single span.

Consequently, `SpanCollection` takes a similar approach [to `StringValues`](https://andrewlock.net/a-brief-look-at-stringvalues/), in which it is an abstraction around _either_ a single `Span` _or_ a `Span[]`. This means we can avoid allocating the array entirely until we need to.

For small traces (with a single span), this saves 56 Bytes per scope/trace or ~8% of the basic trace size. For larger traces, `SpanCollection` and `ArraySegment<Span>` are essentially identical, so the allocation is the same.

Given `SpanCollection` (and `ArraySegment` actually) are `readonly struct`, also added `in` to the signatures (given that both structs are the same size and > pointer size).

> The only practical way I could see to actually make `SpanCollection` pointer-sized is to remove the `Count` parameter. But that means we _either_ need to allocate an "ArraySegment" wrapper around the Array, to hold the count, _or_ we always allocate an array of the "correct" length. I explored the latter in a separate PR, using an array pool during the "builder" step, and then allocating an array of the correct size subsequently, but the allocation gains were marginal, and it didn't

## Implementation details

The changes are essentially:
- `SpanCollection` holds _either_ a `Span` _or_ a `Span[]` (_or_ `null`)
- We store this in the same field (much like `StringValues` does), as it reduces the size of the struct which brings small perf benefits
- Pass the span around via `in` to reduce chance of defensive copies
- Fix/replace uses of Moq which requires different usage for `in`/`ref` fields, and can't provide the same functionality as a stub

## Test coverage

Added some unit tests for the implementation, but the important thing is the benchmarks. We see an 8-8.5% reduction in the allocations for create span/create scope:

|  Benchmark         | Base Allocated | Diff Allocated | Change   | Change % |
|:----------|-----------:|-----------:|--------:|--------:|
| Benchmarks.Trace.SpanBenchmark.StartFinishScope&#8209;net6.0 | 696 B | 640 B | -56 B | -8.05%
| Benchmarks.Trace.SpanBenchmark.StartFinishScope&#8209;netcoreapp3.1 | 696 B | 640 B | -56 B | -8.05%
| Benchmarks.Trace.SpanBenchmark.StartFinishScope&#8209;net472 | 658 B | 602 B | -56 B | -8.51%
| Benchmarks.Trace.SpanBenchmark.StartFinishSpan&#8209;net472 | 578 B | 522 B | -56 B | -9.69%
| Benchmarks.Trace.SpanBenchmark.StartFinishSpan&#8209;net6.0 | 576 B | 520 B | -56 B | -9.72%
| Benchmarks.Trace.SpanBenchmark.StartFinishSpan&#8209;netcoreapp3.1 | 576 B | 520 B | -56 B | -9.72%

Other benchmarks which create a single span see similar improvements.

Note that the _two_ scope benchmark (added in #7869) shows essentially no change (as expected).

## Other details

I tried a variety of variations on this approach:
- Keep separate fields for `Span` and `Span[]`
  - Nicer as it's type safe again, but increases allocations (by a pointer) so prob not worth it as this is hot path
- Don't pass via `in`
  - Slows things down slightly
- Remove the `Count` field (and make `SpanCollection` pointer-sized)
  - Requires allocating exact-sized arrays, so not practical to handle array growth 
- Same as above, but use an array pool for the build stage, and then only allocate the fixed size on close
  - Does show improvements in allocation, but more complexity. May be worth considering (I'll create a separate PR for it)
- Don't have a builder, just use an array pool + a Count field, and rely on reliably cleaning up the pooling, 

https://datadoghq.atlassian.net/browse/LANGPLAT-841

### DIFF
--- a/tracer/src/Datadog.Trace/Agent/MessagePack/TraceChunkModel.cs
+++ b/tracer/src/Datadog.Trace/Agent/MessagePack/TraceChunkModel.cs
@@ -260,7 +260,11 @@ internal readonly struct TraceChunkModel
             return _spans[0].SpanId == spanId ? 0 : -1;
         }
 
-        var array = _spans.ToArray();
+        if (_spans.TryGetArray() is not { } array)
+        {
+            // Shouldn't be possible, because we handle 0 and 1 spans above
+            return -1;
+        }
 
         // iterate over the span array starting at the specified index + 1
         for (var i = startIndex; i < count; i++)
--- a/tracer/src/Datadog.Trace/Agent/SpanCollection.cs
+++ b/tracer/src/Datadog.Trace/Agent/SpanCollection.cs
@@ -151,15 +151,13 @@ internal readonly struct SpanCollection : IEnumerable<Span>
     }
 
     /// <summary>
-    /// Creates a Span array from the current <see cref="SpanCollection"/> object.
-    /// Note that this allocates if the <see cref="SpanCollection"/> does not have a count greater than 1
+    /// Try to get the underlying Span array from the current <see cref="SpanCollection"/> object as an `ArraySegment`.
+    /// If the <see cref="SpanCollection"/> does _not_ contain an array (because it contains 0 or 1 spans) then returns
+    /// <c>null</c>.
     /// </summary>
-    /// <returns>A Span array represented by this instance.</returns>
-    /// <remarks>
-    /// <para>If the <see cref="SpanCollection"/> contains a single Span internally, it is copied to a new array.</para>
-    /// <para>If the <see cref="SpanCollection"/> contains an array internally it returns that array instance.</para>
-    /// </remarks>
-    public ArraySegment<Span> ToArray()
+    /// <returns>A wrapper around the Span array represented by this instance if this instance represents more than
+    /// one span, otherwise null.</returns>
+    public ArraySegment<Span>? TryGetArray()
     {
         // Take local copy of _values so type checks remain valid even if the SpanCollection is overwritten in memory
         object? value = _values;
@@ -167,15 +165,9 @@ internal readonly struct SpanCollection : IEnumerable<Span>
         {
             return new ArraySegment<Span>(values, 0, Count);
         }
-        else if (value != null)
-        {
-            // value not array, can only be Span
-            return new ArraySegment<Span>([Unsafe.As<Span>(value)], 0, 1);
-        }
-        else
-        {
-            return new ArraySegment<Span>(Array.Empty<Span>(), 0, 0);
-        }
+
+        // zero or one spans, so return null
+        return null;
     }
 
     private static Span[] GrowIfNeeded(Span[] array, int currentCount)
--- a/tracer/src/Datadog.Trace/Ci/Processors/TestSuiteVisibilityProcessor.cs
+++ b/tracer/src/Datadog.Trace/Ci/Processors/TestSuiteVisibilityProcessor.cs
@@ -1,4 +1,4 @@
-// <copyright file="TestSuiteVisibilityProcessor.cs" company="Datadog">
+﻿// <copyright file="TestSuiteVisibilityProcessor.cs" company="Datadog">
 // Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2 License.
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
 // </copyright>
@@ -46,7 +46,12 @@ internal class TestSuiteVisibilityProcessor : ITraceProcessor
         }
 
         // we know we have multiple spans, so get the underlying array
-        var segment = trace.ToArray();
+        if (trace.TryGetArray() is not { } segment)
+        {
+            // Shouldn't be possible, because we handle 0 and 1 spans above
+            return trace;
+        }
+
         Span[]? spans = null;
         var copiedCount = 0;
         var haveDrops = false;
--- a/tracer/test/Datadog.Trace.Tests/Agent/SpanCollectionTests.cs
+++ b/tracer/test/Datadog.Trace.Tests/Agent/SpanCollectionTests.cs
@@ -20,7 +20,7 @@ public class SpanCollectionTests
 
         collection.Count.Should().Be(0);
         collection.FirstSpan.Should().BeNull();
-        collection.ToArray().Array.Should().BeEmpty();
+        collection.TryGetArray().HasValue.Should().BeFalse();
         foreach (var span in collection)
         {
             Assert.Fail("We shouldn't have a span to enumerate: " + span);
@@ -37,7 +37,7 @@ public class SpanCollectionTests
 
         collection.Count.Should().Be(1);
         collection.FirstSpan.Should().BeSameAs(span);
-        collection.ToArray().Array.Should().ContainSingle().Which.Should().BeSameAs(span);
+        collection.TryGetArray().HasValue.Should().BeFalse();
         var spans = new List<Span>();
         foreach (var x in collection)
         {
@@ -57,10 +57,11 @@ public class SpanCollectionTests
 
         collection.Count.Should().Be(0);
         collection.FirstSpan.Should().BeNull();
-        var array = collection.ToArray();
-        array.Count.Should().Be(0);
-        array.Offset.Should().Be(0);
-        array.Array!.Length.Should().Be(length);
+        var array = collection.TryGetArray();
+        array.HasValue.Should().BeTrue();
+        array!.Value.Count.Should().Be(0);
+        array.Value.Offset.Should().Be(0);
+        array.Value.Array!.Length.Should().Be(length);
         foreach (var span in collection)
         {
             Assert.Fail("We shouldn't have a span to enumerate: " + span);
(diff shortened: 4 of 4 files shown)


tracer/src/Datadog.Trace/Agent/MessagePack/TraceChunkModel.cs
lucaspimentel:
```diff
- iteration
+ iterating
```

tracer/src/Datadog.Trace/Agent/MessagePack/TraceChunkModel.cs
```diff
@@ -243,15 +244,28 @@ private bool Contains(ulong spanId, int startIndex)
...
+        if (count == 1)
+        {
+            return _spans[0].SpanId == spanId ? 0 : -1;
+        }
+
+        var array = _spans.ToArray();
+
         // iterate over the span array starting at the specified index + 1
-        for (var i = startIndex; i < _spans.Count; i++)
+        for (var i = startIndex; i < count; i++)
```
lucaspimentel, line 270: Can we `foreach` over `_spans` (since it has a `struct` iterator) to avoid creating the array copy?

tracer/src/Datadog.Trace/Span.cs
```diff
@@ -47,10 +47,8 @@ internal Span(SpanContext context, DateTimeOffset? start, ITags tags, IEnumerabl
 
             if (links is not null)
             {
-                foreach (var link in links)
-                {
-                    AddLink(link);
-                }
+                // We know we're not finished and this allows optimizations vs using AddLink()
```
lucaspimentel: > "We know we're not finished"

I might be more clear to say something like, "we're in the Span ctor, so we know SpanLinks is empty here."

tracer/src/Datadog.Trace/Agent/MessagePack/TraceChunkModel.cs
andrewlock: 1. I don't think so, because we're potentially starting half-way through the array here?
2. Maybe we could special case startIndex = 0?
3. There's no array copy 🙂  We're already checked count > 2, which means we have an array, and we just return it directly.

tracer/src/Datadog.Trace/Agent/MessagePack/TraceChunkModel.cs
andrewlock:
```suggestion
    /// Searches for the specified spanId by iterating the array of spans.
```

tracer/src/Datadog.Trace/Agent/SpanCollection.cs
```diff
@@ -0,0 +1,298 @@
...
+    [TestingOnly]
+    internal SpanCollection(Span[] values)
+        : this(values, values.Length)
+    {
+    }
+
+    /// <summary>
+    /// Gets the first span in the <see cref="SpanCollection" />, or returns null if the collection is empty
+    /// </summary>
+    public Span? RootSpan
```
lucaspimentel: Consider calling this `FirstSpan` or `FirstOrDefault` or something else. It's not guaranteed to be the root span.

tracer/src/Datadog.Trace/Span.cs
andrewlock:
```suggestion
                // We're in the constructor, so doing a direct replace allows optimizations vs using AddLink()
```

tracer/src/Datadog.Trace/Agent/MessagePack/TraceChunkModel.cs
lucaspimentel: > we're potentially starting half-way

Ah, right, the `startIndex` parameter.

> we could special case startIndex = 0

Nah, probably not worth it.

> There's no array copy

My mistake. I saw the name `ToArray()` and expected it to return a _new_ array (.NET BCL convention?), so it usually works for getting a snapshot of a collection. (one exception is returning a cached empty array).

But also you'd expect `Span[]`, not `ArraySegment<Span>`, so 🤷🏽‍♂️

tracer/src/Datadog.Trace/Agent/MessagePack/TraceChunkModel.cs
lucaspimentel: I tried "Suggested change" but the diff looked weird so I didn't trust it 😅 

image

tracer/src/Datadog.Trace/Agent/MessagePack/TraceChunkModel.cs
lucaspimentel: Consider renaming `ToArray()` to avoid confusion?

andrewlock: weeeeellll the thing is, it _does_ allocate array if you don't use it "correctly" 🤔 So I'm not sure... Ideally, this wouldn't be necessary or exposed at all tbh, but this is one of the few places where we need it because we're _not_ just doing a `foreach` 🤔

Other options would include `TryGetArray()` which returns a nullable if the underlying stored value isn't an array? WDYT? 🤔

lucaspimentel:
```csharp
foreach (var span in _spans.Skip(startIndex))
{
    ...
}
```
:trollface:

andrewlock: image

tracer/src/Datadog.Trace/Agent/SpanCollection.cs
```diff
@@ -0,0 +1,298 @@
...
+#if NETFRAMEWORK
+using Datadog.Trace.VendoredMicrosoftCode.System.Runtime.CompilerServices.Unsafe;
+#endif
+
+namespace Datadog.Trace.Agent;
+
+/// <summary>
+/// Represents zero/null, one, or many Spans in an efficient way.
+/// </summary>
+internal readonly struct SpanCollection : IEnumerable<Span>
```
lucaspimentel, line 22: Don't stop here! Write a generic `ZeroOneOrMany<T>` (`ZeroOrMore<T>`?) while you're at it 😄

andrewlock: I'm not averse to it, it would be a simple change in practice, but do we actually have arrays of other things to manage? If not, then I'm inclined to keep it narrowly focused, mostly because there's potentially other optimizations that we can bake in here, e.g. pooling the arrays etc, which we wouldn't necessarily want in the _general_ case.

So yeah, if you have other places you're thinking of, I'd say let's consider it. If not then I'd say let's leave it as-is for now?

tracer/src/Datadog.Trace/Agent/SpanCollection.cs
andrewlock: Renamed to `FirstSpan` in https://github.com/DataDog/dd-trace-dotnet/pull/7874/commits/32e2660aea1ac8a2d9a4903d8a4bfff70d798db2

tracer/src/Datadog.Trace/Agent/MessagePack/TraceChunkModel.cs
andrewlock: So I renamed this method to be `TryGetArray()`, and also updated it to return a `null` if there _isn't_ an underlying array in https://github.com/DataDog/dd-trace-dotnet/pull/7874/commits/c02cbdbe994d1a7a4f0e104e1c2d4c2e3236e150, to try to make it a bit clearer, plus also avoid _accidental_ allocation (caused by calling `ToArray()` when there's only one span).

The downside (or upside, depending on how you look at it) is that you have to check the result of this call to see if you got null, even though you _should_ have already handled the 0/1 cases distinctly anyway. That makes the call sites a little uglier in general.

That said, there _are_ only 2 callsites, and we don't really want more, so maybe it's ok that it's a bit ugly? 🤷‍♂️ WDYT @lucaspimentel, do you prefer this design?

anna-git (approved):
Nice!

tonyredondo (approved):

tracer/src/Datadog.Trace/Agent/SpanCollection.cs
lucaspimentel: I was joking (mostly). If we do find other uses for this type of collection, we can do that later.

tracer/src/Datadog.Trace/Agent/MessagePack/TraceChunkModel.cs
lucaspimentel: That works, thanks! (although I would've been happy with just renaming `ToArray()` because that name implied the wrong behavior sometimes).

PR #7879, opened 2025-11-28
Mark all classes as `sealed` or `static`

andrewlock:
## Summary of changes

- Upgrades the `SealedAnalyzer` added in #7878 from `info` to `warning` for specific projects
- Fix all the errors it flagged

## Reason for change

As described in `#7878`, making types `sealed` can have performance benefits. It's basically a free small perf bump, and there's not _really_ any downsides.

## Implementation details

After writing the analyzer, I got claude code to fix all the violations. Annoyingly, `dotnet format` doesn't seem to work with this analyzer because it's a "compilation end" analyzer.

Then I went through all the changes and confirmed they're valid. I also changed a bunch to `static` where they were stateless and effectively _should_ have been static.

> I initially made all the integrations and aspects `static` instead of `sealed`, but then realised the integration are used as generic arguments, and I suspect that's not allowed at runtime

I have written some comments on the PR for places where we had to explicitly ignore, but GitHub can't handle it, so here's some others 😅 

- `Tracer` - had to not marked this sealed, because we derive from it in tests 🙁 This one is very annoying, I may try to fix in a subsequent PR.
- `DirectSubmissionLogSink` - we were deriving from this in tests so that we could call a protected method, so just added a way to call that instead.
- `PerformanceCountersListener` - we derive from this in tests.

## Test coverage

As long as everything passes, we should be good 🤞 

## Other details

Stacked on
- https://github.com/DataDog/dd-trace-dotnet/pull/7878

https://datadoghq.atlassian.net/browse/LANGPLAT-813

### DIFF
--- a/tracer/test/Datadog.Trace.Tests/Snapshots/PublicApiTests.Datadog.Trace.Manual.PublicApiHasNotChanged.verified.txt
+++ b/tracer/test/Datadog.Trace.Tests/Snapshots/PublicApiTests.Datadog.Trace.Manual.PublicApiHasNotChanged.verified.txt
@@ -339,11 +339,11 @@ namespace Datadog.Trace.Ci
 }
 namespace Datadog.Trace.Ci.Coverage.Attributes
 {
-    public class AvoidCoverageAttribute : System.Attribute
+    public sealed class AvoidCoverageAttribute : System.Attribute
     {
         public AvoidCoverageAttribute() { }
     }
-    public class CoveredAssemblyAttribute : System.Attribute
+    public sealed class CoveredAssemblyAttribute : System.Attribute
     {
         public CoveredAssemblyAttribute() { }
     }


tracer/src/Datadog.Trace/Agent/MessagePack/SpanFormatterResolver.cs
```diff
@@ -10,11 +10,11 @@
 
 namespace Datadog.Trace.Agent.MessagePack
 {
-    internal class SpanFormatterResolver : IFormatterResolver
+    internal sealed class SpanFormatterResolver : IFormatterResolver
     {
         public static readonly IFormatterResolver Instance = new SpanFormatterResolver();
 
-        private readonly SpanMessagePackFormatter _formatter;
+        private readonly object _formatter;
```
andrewlock, line 17: This was necessary because line 28 no longer compiles otherwise:

```csharp
return (IMessagePackFormatter<T>)_formatter;
```

tracer/src/Datadog.Trace/Agent/Transports/HttpClientRequestFactory.cs
```diff
@@ -12,7 +12,11 @@
 
 namespace Datadog.Trace.Agent.Transports
 {
+#if NET5_0_OR_GREATER // in .NET 6 we derive a SocketHandlerRequestFactory
```
andrewlock, line 15: In .NET 6 we derive from this to create `SocketHandlerRequestFactory` so we _can't_ be sealed

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Logging/NLog/LogsInjection/MappedDiagnosticsContextSetterProxy.cs
```diff
@@ -8,7 +8,9 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Logging.NLog.LogsInjecti
     /// <summary>
     /// Duck type for MappedDiagnosticsContext in NLog 1.0+
     /// </summary>
+#pragma warning  disable DDSEAL001 // Types should be sealed where possible. Types used for duck-typing cannot be sealed.
```
andrewlock, line 11: This is actually a (rare) `class` based duck type. We don't recommend these now (and can probably update to use `interface` instead, so just ignoring it for now

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Logging/NLog/LogsInjection/MappedDiagnosticsLogicalContextSetterProxy.cs
```diff
@@ -11,7 +11,9 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Logging.NLog.LogsInjecti
     /// <summary>
     /// Duck type for MappedDiagnosticsLogicalContext in NLog 4.6+
     /// </summary>
+#pragma warning  disable DDSEAL001 // Types should be sealed where possible. Types used for duck-typing cannot be sealed.
     internal class MappedDiagnosticsLogicalContextSetterProxy
+#pragma warning restore DDSEAL001
```
andrewlock, line 16: This is actually a (rare) `class` based duck type. We don't recommend these now (and can probably update to use `interface` instead, so just ignoring it for now

tracer/src/Datadog.Trace/DiagnosticListeners/EndpointFeatureProxy.cs
```diff
@@ -12,7 +12,9 @@ namespace Datadog.Trace.DiagnosticListeners
     /// explicitly, e.g. by https://github.com/dotnet/aspnetcore/blob/v3.0.3/src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
     /// Also see AspNetCoreDiagnosticObserver.EndpointFeatureStruct
     /// </summary>
+#pragma warning  disable DDSEAL001 // Types should be sealed where possible. Types used for duck-typing cannot be sealed.
     internal class EndpointFeatureProxy
+#pragma warning restore DDSEAL001
```
andrewlock, line 17: This is actually a (rare) `class` based duck type. We don't recommend these now (and can probably update to use `interface` instead, so just ignoring it for now

NachoEchevarria (approved):

PR #7887, opened 2025-12-01
Small perf improvement for some `TagList` implementations

andrewlock:
## Summary of changes

Tags that are readonly should be implemented as such, and return a value directly

## Reason for change

We have some tags that currently always return the same value, but are implemented as writeable fields. This requires a backing field, and so means the object is _sliiiightly_ more allocatey than it needs to be.

## Implementation details

Switch to using ` => ` readonly properties instead of `{ get; set; }`

## Test coverage

If the tests pass, it's fine

### DIFF
--- a/tracer/src/Datadog.Trace/Tagging/AzureServiceBusTags.cs
+++ b/tracer/src/Datadog.Trace/Tagging/AzureServiceBusTags.cs
@@ -18,7 +18,7 @@ namespace Datadog.Trace.Tagging
         private string _spanKind;
 
         [Tag(Trace.Tags.InstrumentationName)]
-        public string InstrumentationName => "AzureServiceBus";
+        public string InstrumentationName => nameof(Configuration.IntegrationId.AzureServiceBus);
 
         [Metric(Trace.Tags.Analytics)]
         public double? AnalyticsSampleRate { get; set; }


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/AzureServiceBusCommon.cs
```diff
@@ -126,7 +126,6 @@ private static CallTargetState CreateSenderSpanInternal(
             tags.MessagingDestinationName = entityPath;
             tags.MessagingOperation = operationName;
             tags.MessagingSystem = "servicebus";
-            tags.InstrumentationName = "AzureServiceBus";
```
bouwkast, line 129: I thought there was a reason why this writable but seems fine to me

tracer/src/Datadog.Trace/Tagging/AzureServiceBusTags.cs
```diff
@@ -18,7 +18,7 @@ internal partial class AzureServiceBusTags : OpenTelemetryTags
         private string _spanKind;
 
         [Tag(Trace.Tags.InstrumentationName)]
-        public string InstrumentationName { get; set; }
+        public string InstrumentationName => "AzureServiceBus";
```
lucaspimentel: We could use `nameof(Configuration.IntegrationId.AzureServiceBus)` for consistency with the change in `RabbitMQTags.cs`.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Azure/ServiceBus/AzureServiceBusCommon.cs
```diff
@@ -126,7 +126,6 @@ private static CallTargetState CreateSenderSpanInternal(
             tags.MessagingDestinationName = entityPath;
             tags.MessagingOperation = operationName;
             tags.MessagingSystem = "servicebus";
```
lucaspimentel, line 128: Probably more than you wanted to do this PR, but I was looking into why we can't do the same with `MessagingSystem`, and noticed that `AzureServiceBusTags` inherits that property from `OpenTelemetryTags`, but I don't think `AzureServiceBusTags` needs to inherit from `OpenTelemetryTags` anymore now that the Service Bus integration uses call-target instrumentation instead of the OpenTelemetry observer.

Maybe something @pablomartinezbernardo can look into later.

andrewlock: Yeah, I was going to take a look at this, but given our lack of extensive ServiceBus testing, I was a little hesitant, so will move to a separate PR

andrewlock: Actually, as far as I can tell, [we still have an OpenTelemetry observer](https://github.com/DataDog/dd-trace-dotnet/blob/fc81dab19992225516587bf0d28ba4bfcf7f37d8/tracer/src/Datadog.Trace/Activity/Handlers/AzureServiceBusActivityHandler.cs#L22)? 🤔

lucaspimentel: I don't know if we still use the observer. Question for Pablo 😅

PR #7893, opened 2025-12-03
Read container tags hash from agent

vandonr:
## Summary of changes

same as https://github.com/DataDog/dd-trace-java/pull/9156

 - send container ID to the agent with feature discovery requests
 - get back the container tags hash and store it

This hash is meant to be matched in the backend with the actual container tags, with the advantage that :
 - it's less cardinality than container ID
 - it's smaller than propagating th tag values themselves

related RFC: https://docs.google.com/document/d/15GtNOKGBCt6Dc-HsDNnMmCdZwhewFQx8yUlI9in5n3M

### DIFF
--- a/tracer/test/Datadog.Trace.Tests/Agent/DiscoveryServiceTests.cs
+++ b/tracer/test/Datadog.Trace.Tests/Agent/DiscoveryServiceTests.cs
@@ -402,7 +402,7 @@ public class DiscoveryServiceTests
             responseContent: GetConfig(),
             responseHeaders: new Dictionary<string, string> { { AgentHttpHeaderNames.ContainerTagsHash, expectedTagsHash } }));
 
-        var containerMetadata = NullContainerMetadata;
+        var containerMetadata = new ContainerMetadata(containerId: null, entityId: null);
 
         var ds = new DiscoveryService(factory, containerMetadata, InitialRetryDelayMs, MaxRetryDelayMs, RecheckIntervalMs);
         ds.SubscribeToChanges(x => mutex.Set());


tracer/src/Datadog.Trace/Agent/DiscoveryService/DiscoveryService.cs
```diff
@@ -219,6 +220,13 @@ private async Task FetchConfigurationLoopAsync()
 
                     var api = requestFactory.Create(uri);
 
+                    // Add container ID header if available
+                    var containerId = ContainerMetadata.GetContainerId();
+                    if (containerId != null)
+                    {
+                        api.AddHeader(AgentHttpHeaderNames.ContainerId, containerId);
+                    }
```
andrewlock: These are fixed for the lifetime of the app, right? In that case, we should likely include it as a default header instead, as part of the `CreateApiRequestFactory` call

tracer/src/Datadog.Trace/Agent/DiscoveryService/DiscoveryService.cs
```diff
@@ -254,6 +262,13 @@ int GetNextSleepDuration(int? previousDuration) =>
 
         private async Task ProcessDiscoveryResponse(IApiResponse response)
         {
+            // Extract and store container tags hash from response headers
+            var containerTagsHash = response.GetHeader(AgentHttpHeaderNames.ContainerTagsHash);
+            if (containerTagsHash != null)
+            {
+                ContainerMetadata.ContainerTagsHash = containerTagsHash;
```
andrewlock: I swear you're trolling me with all this static stuff 😅 Having static readonly state is one thing, but having static _mutable_ state is a whole new ball game 😛  I think the `ContainerMetadata` object likely needs refactoring to be an instance type and passed in to where it's used, so that we don't destroy our tests completely. I'd suggest doing that in a separate PR first, then rebasing this on top

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
```diff
@@ -46,6 +46,11 @@ internal static class ContainerMetadata
 
         private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(ContainerMetadata));
 
+        /// <summary>
+        /// Gets or sets the container tags hash received from the agent, used by DBM/DSM
+        /// </summary>
+        public static string ContainerTagsHash { get; set; }
```
andrewlock: AFAICT, this isn't actually used anywhere? 🤔

tracer/src/Datadog.Trace/Agent/DiscoveryService/DiscoveryService.cs
vandonr: ok fair 😅

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
vandonr: _yet_ 😏 
I could use it in the same PR, but I wanted to do baby steps. It's gonna be used in DSM and DBM injection

tracer/src/Datadog.Trace/Agent/DiscoveryService/DiscoveryService.cs
```diff
@@ -373,13 +389,33 @@ public Task DisposeAsync()
...
+
+        private static IApiRequestFactory CreateApiRequestFactory(ExporterSettings exporterSettings, string? containerId, TimeSpan tcpTimeout)
+        {
+            return AgentTransportStrategy.Get(
                 exporterSettings,
                 productName: "discovery",
                 tcpTimeout: tcpTimeout,
-                AgentHttpHeaderNames.MinimalHeaders,
+                BuildHeaders(containerId),
                 () => new MinimalAgentHeaderHelper(),
```
andrewlock: If you change one of these, you need to change _both_ of them. Different ones are use for different transports. Which means you likely need a custom `MinimalAgentHeaderHelper()` here, something like this:

```csharp
containerId is null
  ? () => new MinimalAgentHeaderHelper()
  : () => new MinimalWithContainerIdHelper()
```

I know, it sucks 😅

tracer/src/Datadog.Trace/Agent/DiscoveryService/DiscoveryService.cs
```diff
@@ -373,13 +389,33 @@ public Task DisposeAsync()
...
+        /// </summary>
+        internal static KeyValuePair<string, string>[] BuildHeaders(string? containerId)
+        {
+            if (containerId != null)
+            {
+                // if container ID is available, add it to headers
+                var headers = new KeyValuePair<string, string>[AgentHttpHeaderNames.MinimalHeaders.Length + 1];
+                Array.Copy(AgentHttpHeaderNames.MinimalHeaders, headers, AgentHttpHeaderNames.MinimalHeaders.Length);
+                headers[AgentHttpHeaderNames.MinimalHeaders.Length] = new KeyValuePair<string, string>(AgentHttpHeaderNames.ContainerId, containerId);
+                return headers;
```
andrewlock: We _maybe_ should cache this generation, given that this method will be called whenever the exporter settings change, but we _probably_ don't want or need to do that for now seeing as that shouldn't happen very often.

Also, collection expressions are awesome 😉 
```suggestion
                // if container ID is available, add it to headers
                return
                [
                    ..AgentHttpHeaderNames.MinimalHeaders,
                    new(AgentHttpHeaderNames.ContainerId, containerId),
                ];
```

tracer/src/Datadog.Trace/Agent/DiscoveryService/DiscoveryService.cs
```diff
@@ -373,13 +389,33 @@ public Task DisposeAsync()
...
 
-        private static IApiRequestFactory CreateApiRequestFactory(ExporterSettings exporterSettings, TimeSpan tcpTimeout)
-            => AgentTransportStrategy.Get(
+        /// <summary>
+        /// Builds the headers array for the discovery service, including the container ID if available.
+        /// Internal for testing purposes.
+        /// </summary>
+        internal static KeyValuePair<string, string>[] BuildHeaders(string? containerId)
+        {
+            if (containerId != null)
```
andrewlock, line 465: We never "discover" this later, right? i.e. if `containerId` is non null the first time this is called, it will be non-null for ever more (and never change) right?

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
```diff
@@ -59,6 +59,12 @@ public ContainerMetadata(string containerId, string entityId)
             _entityId = new Lazy<string>(() => entityId);
         }
 
+        /// <summary>
+        /// Gets or sets the container tags hash received from the agent, used by DBM/DSM
+        /// This is set when we receive a value for it in an http response from the agent
+        /// </summary>
+        public string ContainerTagsHash { get; set; }
```
andrewlock: Given this will be written to and from different threads, you may need to make sure you handle volatility. You could also use `Interlocked` depending on how this _will_ be used

```csharp
    public string ContainerTagsHash
    {
        set => Volatile.Write(ref field, value);
    }
```

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.NetFramework.cs
```diff
@@ -26,6 +26,12 @@ public ContainerMetadata(string containerId, string entityId)
         // nothing to do, just to match the other version
     }
 
+    /// <summary>
+    /// Gets or sets the container tags hash received from the agent, used by DBM/DSM
+    /// This is set when we receive a value for it in an http response from the agent
+    /// </summary>
+    public string? ContainerTagsHash { get; set; }
```
andrewlock: As above
```suggestion
    public string? ContainerTagsHash
    {
        get => Volatile.Read(ref field);
        set => Volatile.Write(ref field, value);
    }
```

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
vandonr: see https://github.com/DataDog/dd-trace-dotnet/pull/8061

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
vandonr: TIL about the `field` keyword

tracer/src/Datadog.Trace/Agent/DiscoveryService/DiscoveryService.cs
vandonr: yes correct. Are you asking because if yes, then you'd have this code written differently, or just out of curiosity ?

tracer/src/Datadog.Trace/Agent/DiscoveryService/DiscoveryService.cs
vandonr: 🤯

tracer/src/Datadog.Trace/Agent/DiscoveryService/DiscoveryService.cs
vandonr: hmm, yeah, not great, I have to do a different type because we use the instance to access a static field :/
btw, do you know why we don't use a normal `Lazy` in that helper ?

vandonr: I really didn't like the idea of duplicating the whole class, so i went another route, and made MinimalAgentHeaderHelper return a different value depending on whether it was given a container ID in the ctor.

tracer/src/Datadog.Trace/Agent/DiscoveryService/DiscoveryService.cs
andrewlock: Yeah, if the answer had been "yes" we would have to write this differently, because we're building this collection once and never re-evaluating. Which is fine if this always gives the same result, but not otherwise 🙂

tracer/src/Datadog.Trace/Agent/DiscoveryService/DiscoveryService.cs
```diff
@@ -297,6 +306,13 @@ internal bool RequireRefresh(string? currentHash, DateTimeOffset utcNow)
 
         private async Task ProcessDiscoveryResponse(IApiResponse response)
         {
+            // Extract and store container tags hash from response headers
+            var containerTagsHash = response.GetHeader(AgentHttpHeaderNames.ContainerTagsHash);
```
andrewlock, line 310: I was wondering about this code, as we're going to do the work to extract and set this hash with every response, but I think that's fine, given that we _shouldn't_ call the discovery service so much any more since #7979 (and also given the amount of work we do later in this method)

tracer/src/Datadog.Trace/HttpOverStreams/MinimalAgentHeaderHelper.cs
```diff
@@ -22,6 +29,18 @@ protected override string MetadataHeaders
                 _metadataHeaders = string.Concat(headers);
             }
 
+            if (_containerId != null)
```
andrewlock, line 32: I think we should probably refactor the `MinimalAgentHeaderHelper` code in general, there's some perf opportunities, and it feels like this is overall more complex than it needs to be, but I'll do that in a subsequent PR after you merge this 🙂

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
```diff
@@ -3,6 +3,7 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
 // </copyright>
 
+#nullable enable
```
andrewlock, line 6: 🙏

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
```diff
@@ -43,28 +44,38 @@ internal sealed class ContainerMetadata
...
         {
-            _containerId = new Lazy<string>(GetContainerIdInternal, LazyThreadSafetyMode.ExecutionAndPublication);
-            _entityId = new Lazy<string>(() => GetEntityIdInternal(_containerId), LazyThreadSafetyMode.ExecutionAndPublication);
+            _containerId = new Lazy<string?>(GetContainerIdInternal, LazyThreadSafetyMode.ExecutionAndPublication);
+            _entityId = new Lazy<string?>(() => GetEntityIdInternal(_containerId), LazyThreadSafetyMode.ExecutionAndPublication);
         }
 
         // For use in tests only
-        public ContainerMetadata(string containerId, string entityId)
+        public ContainerMetadata(string? containerId, string? entityId)
```
andrewlock, line 58:
```suggestion
        [TestingOnly]
        public ContainerMetadata(string? containerId, string? entityId)
```

tracer/test/Datadog.Trace.TestHelpers/TransportHelpers/TestApiResponse.cs
```diff
@@ -23,6 +23,7 @@ public TestApiResponse(int statusCode, string body, string contentType, Dictiona
         _body = body;
         _headers = headers;
         ContentTypeHeader = contentType;
+        _headers = headers ?? new Dictionary<string, string>();
```
andrewlock:
```suggestion
        _headers = headers ?? new();
        ContentTypeHeader = contentType;
```

tracer/test/Datadog.Trace.Tests/Agent/DiscoveryServiceTests.cs
```diff
@@ -25,6 +27,8 @@ public class DiscoveryServiceTests
     private const int MaxRetryDelayMs = 50;
     private const int RecheckIntervalMs = 300_000;
 
+    private static readonly ContainerMetadata NullContainerMetadata = new(containerId: null, entityId: null);
```
andrewlock, line 30: nit: should prob do this in the TraceExporterTests too

tracer/test/Datadog.Trace.Tests/Agent/DiscoveryServiceTests.cs
```diff
@@ -387,6 +391,51 @@ public async Task HandlesFailuresInApiWithBackoff()
...
+    {
+        const string expectedTagsHash = "test-container-tags-hash-123";
+        using var mutex = new ManualResetEventSlim();
+
+        var factory = new TestRequestFactory(x => new TestApiRequest(
+            x,
+            responseContent: GetConfig(),
+            responseHeaders: new Dictionary<string, string> { { AgentHttpHeaderNames.ContainerTagsHash, expectedTagsHash } }));
+
+        var containerMetadata = NullContainerMetadata;
```
andrewlock: If you're going to be modifying the object (you are) don't use the "shared" helper, just create a new one for the test 🙂 

```suggestion
        var containerMetadata = new ContainerMetadata(containerId: null, entityId: null);
```

tracer/test/Datadog.Trace.Tests/Agent/DiscoveryServiceTests.cs
```diff
@@ -387,6 +391,51 @@ public async Task HandlesFailuresInApiWithBackoff()
...
+
+        var ds = new DiscoveryService(factory, containerMetadata, InitialRetryDelayMs, MaxRetryDelayMs, RecheckIntervalMs);
```
andrewlock, line 407: ~`RecheckIntervalMs = 300s` - are you sure that's correct 🤔 If you "mis" the first check (which happens only 50ms (`InitialRetryDelayMs`) after calling `new()`), then this test will flake I think. It's rare, but looks like a definite possibility. If you set `RecheckIntervalMs` such that it is much _smaller_ than the timeout you use in the `mutex`, then that should reduce the chance of flake I think?~

EDIT: actually, this should be fine 😄 There's no real race condition here: the hash can be updated before subcribetochanges is called, but the mutex is guaranteed to be called in that case. Ignore me 😄

andrewlock (approved):
LGTM thanks 👍

tracer/src/Datadog.Trace/Agent/DiscoveryService/DiscoveryService.cs
vandonr: yeah, since it's not _supposed_ to change, we could check if we already have it, and only query it then. But if you say it's fine :)

tracer/src/Datadog.Trace/HttpOverStreams/MinimalAgentHeaderHelper.cs
vandonr: ok thanks. Yeah that code felt a bit... dated

tracer/test/Datadog.Trace.Tests/Agent/DiscoveryServiceTests.cs
vandonr: I did it here because it's used in many tests, but it's only used once in TraceExporterTests, so it feels less relevant to extract a property ?

tracer/test/Datadog.Trace.Tests/Agent/DiscoveryServiceTests.cs
vandonr: ah yes good catch, I used the refactor option to replace all usages and forgot about this

PR #7898, opened 2025-12-03
Refacto: Make ContainerMetadata an instance class

vandonr:
## Summary of changes

Everything in that class is public static, and it makes more sense to have it as an instance for testability.

## Implementation details

Kept the backing fields as static lazy values, to avoid building them twice. In the context of tests, one would create an override/mock and set the values that way.

I didn't really inject it anywhere though, it'd only make sense in the context of tests.
I _could_ take the opportunity to add tests, but also, I'm doing this to add an extra field to the container metadata, so I can do the work then ?

## Test coverage

existing tests

### DIFF
--- a/tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.NetFramework.cs
+++ b/tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.NetFramework.cs
@@ -16,6 +16,16 @@ internal sealed class ContainerMetadata
 {
     public static readonly ContainerMetadata Instance = new();
 
+    private ContainerMetadata()
+    {
+    }
+
+    // For use in tests only
+    public ContainerMetadata(string containerId, string entityId)
+    {
+        // nothing to do, just to match the other version
+    }
+
     /// <summary>
     /// Gets the id of the container executing the code.
     /// Return <c>null</c> if code is not executing inside a supported container.
--- a/tracer/test/Datadog.Trace.Tests/Agent/ApiTests.cs
+++ b/tracer/test/Datadog.Trace.Tests/Agent/ApiTests.cs
@@ -47,6 +47,7 @@ namespace Datadog.Trace.Tests.Agent
             requestMock.Verify(x => x.PostAsync(It.IsAny<ArraySegment<byte>>(), MimeTypes.MsgPack), Times.Once());
         }
 
+#if !NETFRAMEWORK
         [Theory]
         [CombinatorialData]
         public async Task SendWithContainerMetadata(
@@ -93,6 +94,7 @@ namespace Datadog.Trace.Tests.Agent
                 savedHeaders.Should().NotContainKey(AgentHttpHeaderNames.EntityId);
             }
         }
+#endif
 
         [Theory]
         [InlineData(429)]
--- a/tracer/test/Datadog.Trace.Tests/Telemetry/Transports/JsonTelemetryTransportTests.cs
+++ b/tracer/test/Datadog.Trace.Tests/Telemetry/Transports/JsonTelemetryTransportTests.cs
@@ -22,6 +22,7 @@ namespace Datadog.Trace.Tests.Telemetry.Transports
 {
     public class JsonTelemetryTransportTests
     {
+#if !NETFRAMEWORK
         [Theory]
         [CombinatorialData]
         public async Task ShouldContainRequiredHeaders(bool debugEnabled, [CombinatorialValues("", "gzip")] string compression, bool agentless)
@@ -83,6 +84,8 @@ namespace Datadog.Trace.Tests.Telemetry.Transports
             var expectedEncoding = compression == "gzip" ? "gzip" : null;
             requestMock.Verify(x => x.PostAsync(It.IsAny<ArraySegment<byte>>(), "application/json", expectedEncoding), Times.Once);
         }
+
+#endif
 
         [Fact]
         public void SerializedAppStartedShouldProduceJsonWithExpectedFormat()


tracer/test/Datadog.Trace.IntegrationTests/ContainerTaggingTests.cs
```diff
@@ -51,7 +51,7 @@ public async Task Http_Headers_Contain_ContainerId()
 
                 var headers = agent.TraceRequestHeaders.Should().ContainSingle().Subject;
 
-                var expectedContainedId = ContainerMetadata.GetContainerId();
+                var expectedContainedId = ContainerMetadata.Instance.ContainerId;
```
andrewlock: The idea in general is to _not_ use `ContainerMetadata.Instance` anywhere, other than at the "root" in the real application, which injects it in 🙂 Particularly in tests, you 100% don't want to be using the static instance. For example:

```suggestion
                var containerMetadata = new ContainerMetadata();
                var expectedContainedId = containerMetadata.ContainerId;
```

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
```diff
@@ -41,19 +39,21 @@ internal static class ContainerMetadata
         // if we're running in host namespace or not (does not work when running in DinD)
         private const long HostCgroupNamespaceInode = 0xEFFFFFFB;
 
-        private static readonly Lazy<string> ContainerId = new Lazy<string>(GetContainerIdInternal, LazyThreadSafetyMode.ExecutionAndPublication);
-        private static readonly Lazy<string> CgroupInode = new Lazy<string>(GetCgroupInodeInternal, LazyThreadSafetyMode.ExecutionAndPublication);
+        private static readonly Lazy<string> LazyContainerId = new(GetContainerIdInternal, LazyThreadSafetyMode.ExecutionAndPublication);
+        private static readonly Lazy<string> LazyEntityId = new(GetEntityIdInternal, LazyThreadSafetyMode.ExecutionAndPublication);
```
andrewlock: Having these as static defeats the purpose of the refactoring 🙂 
```suggestion
        private readonly Lazy<string> LazyContainerId = new(GetContainerIdInternal, LazyThreadSafetyMode.ExecutionAndPublication);
        private readonly Lazy<string> LazyEntityId = new(GetEntityIdInternal, LazyThreadSafetyMode.ExecutionAndPublication);
```

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
```diff
@@ -23,7 +21,7 @@ namespace Datadog.Trace.PlatformHelpers
     /// <summary>
     /// Utility class with methods to interact with container hosts.
     /// </summary>
-    internal static class ContainerMetadata
+    internal class ContainerMetadata
```
andrewlock:
```suggestion
    internal sealed class ContainerMetadata
```

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.NetFramework.cs
```diff
@@ -12,14 +12,16 @@ namespace Datadog.Trace.PlatformHelpers;
 /// <summary>
 /// Utility class with methods to interact with container hosts.
 /// </summary>
-internal static class ContainerMetadata
+internal class ContainerMetadata
```
andrewlock:
```suggestion
internal sealed class ContainerMetadata
```

tracer/src/Datadog.Trace/Telemetry/TelemetryConstants.cs
```diff
@@ -20,8 +20,6 @@ internal class TelemetryConstants
         public const string RequestTypeHeader = "DD-Telemetry-Request-Type";
         public const string ClientLibraryLanguageHeader = "DD-Client-Library-Language";
         public const string ClientLibraryVersionHeader = "DD-Client-Library-Version";
-        public const string ContainerIdHeader = Datadog.Trace.AgentHttpHeaderNames.ContainerId;
-        public const string EntityIdHeader = Datadog.Trace.AgentHttpHeaderNames.EntityId;
```
andrewlock, line 24: Were these just unused?

tracer/test/Datadog.Trace.IntegrationTests/TelemetryTransportTests.cs
```diff
@@ -86,14 +86,14 @@ public async Task SetsRequiredHeaders(bool agentless, bool useCloudAgentless)
                 { "DD-Client-Library-Version", TracerConstants.AssemblyVersion },
             };
 
-            if (ContainerMetadata.GetContainerId() is { } containerId)
+            if (ContainerMetadata.Instance.ContainerId is { } containerId)
```
andrewlock, line 89: don't use global instance here

tracer/src/Datadog.Trace/Telemetry/TelemetryConstants.cs
vandonr: they were used, but since they point to the same value as the constants in AgentHttpHeaderNames, and I factorized the code using those from 3 different places, they end up not being used anymore

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
vandonr: Yeah but then how do you mock the value of the properties in tests ? My idea was that you especially wouldn't want this sealed, so that tests can create mock of this.
Or I can create an interface `ContainerMetadata` that this implements.

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
vandonr: hmm, yes, I wanted to protect against creating those several times, but you are right. I wonder if we need to introduce some countermeasures as well, like a constructor that's `[TestingAndPrivateOnly]`

tracer/test/Datadog.Trace.IntegrationTests/ContainerTaggingTests.cs
vandonr: you don't need to inject all the way from the root though, only far enough that tests can pass it as a parameter when they need to, right ?
For sure it makes dependencies clearer, but we'd also end up with huge lists of parameters if we did that for everything.

I figured it'd make sense to create an instance specifically for a test if you can inject it, but in integration tests like this, the setup is very far from the tested code, so maybe I'm wrong, but I feel like passing the container metadata all the way down the call chain would be annoying

tracer/test/Datadog.Trace.IntegrationTests/TelemetryTransportTests.cs
vandonr: so, after discussing, I think we can use the global here because it's an integration test. I could remove it in favor of the one in `JsonTelemetryTransportTests`, but this one tests more things, so I wasn't comfortable in doing so.

Another way to avoid using the global in the test, would be to change the assert to check that the key is set with a non-empty value, but don't assert on the value itself.

vandonr:
@codex review

tracer/src/Datadog.Trace/PlatformHelpers/IContainerMetadata.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// <item>"in-&lt;inode&gt;" if the cgroup node controller's inode is available.
+    ///        We use the memory controller on cgroupv1 and the root cgroup on cgroupv2.</item>
+    /// <item><c>null</c> if neither are available.</item>
+    /// </list>
+    /// </summary>
+    /// <value>The entity id or <c>null</c>.</value>
+    public string? EntityId { get; }
+}
+
+internal static class ContainerMetadataHelper
```
andrewlock: nit: prob should just move this to its own class

tracer/src/Datadog.Trace/PlatformHelpers/IContainerMetadata.cs
```diff
@@ -0,0 +1,50 @@
...
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
+// </copyright>
+
+#nullable enable
+
+using Datadog.Trace.Agent;
+
+namespace Datadog.Trace.PlatformHelpers;
+
+internal interface IContainerMetadata
```
andrewlock: Do we _need_ an interface? 🤔 If it's just for testing purposes, I think being able to provide `ContainerId`/`EntityId` in the constructor is as good as anything, seeing as it's ultimately pretty stateless? 

The main reason _not_ to use an interface here is that it's faster not to, because there's no devirtulaisation or anything required :)

tracer/test/Datadog.Trace.IntegrationTests/TelemetryTransportTests.cs
andrewlock: If we're just going to use the global here, may as well use it in the `ContainerTaggingTests` too I guess, and not bother deleting that? That said, you have a unit test, which is just as good so meh 🤷‍♂️

tracer/test/Datadog.Trace.Tests/Agent/ApiTests.cs
```diff
@@ -40,13 +40,62 @@ public async Task SendTraceAsync_200OK_AllGood()
...
 
+        [Theory]
+        [CombinatorialData]
+        public async Task SendWithContainerMetadata(
+            [CombinatorialValues(null, "my-container-id")]
+            string containerId,
+            [CombinatorialValues(null, "my-entity-id")]
+            string entityId)
+        {
+            var responseMock = new Mock<IApiResponse>();
```
andrewlock, line 59: Just FYI, we have `TestApiRequest` and `TestApiResponse` which are much easier to work with than Moq IMO. The Moq in here predates those. No need to have to change, just an FYI

andrewlock (approved):
Thanks!

tracer/src/Datadog.Trace/PlatformHelpers/IContainerMetadata.cs
vandonr: hmm yeah right. I also created the interface to make compilation errors more obvious if the netfwk and netcore versions of the class differ, but it's true that there are perf implications too.

tracer/test/Datadog.Trace.Tests/Agent/ApiTests.cs
vandonr: ooooh ok, I just took inspiration from the other tests in the same class ^^

tracer/test/Datadog.Trace.IntegrationTests/TelemetryTransportTests.cs
vandonr: I think it's still nice to be able to get rid of an itest in favor of a utest, independently of this all

PR #7903, opened 2025-12-03
Add support for RedisServiceStack v10

andrewlock:
## Summary of changes

Adds support for RedisServiceStack v10

## Reason for change

We want to support the latest releases

## Implementation details

Bumped the numbers and 🤞 

## Test coverage

If the tests pass, we're good. Note that 10.0.0 dropped support for <.NET 6


tracer/build/PackageVersionsLatestMajors.g.props
```diff
@@ -3144,6 +3144,12 @@ NOTE:  This code was generated by the GeneratePackageVersions tool. To safely
       <RequiresDockerDependency>All</RequiresDockerDependency>
       <SampleName>Samples.ServiceStack.Redis</SampleName>
     </PackageVersionSample>
+    <PackageVersionSample Include="test\test-applications\integrations\Samples.ServiceStack.Redis\Samples.ServiceStack.Redis.csproj">
+      <Properties>ApiVersion=10.0.0;RestoreRecursive=false;BuildProjectReferences=false</Properties>
+      <TargetFramework>net48</TargetFramework>
+      <RequiresDockerDependency>All</RequiresDockerDependency>
```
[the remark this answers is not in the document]
andrewlock: .NET Framework is still supported

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Redis/ServiceStack/RedisNativeClientSendReceiveIntegration.cs
```diff
@@ -22,7 +22,7 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.Redis.ServiceStack
         ReturnTypeName = "!!0",
         ParameterTypeNames = new[] { "System.Byte[][]", "System.Func`1[!!0]", "System.Action`1[System.Func`1[!!0]]", ClrNames.Bool },
         MinimumVersion = "4.0.0",
-        MaximumVersion = "6.*.*",
```
bouwkast, line 25: Just checking if this is right as it wasn't an `8` before

andrewlock: oooh, yeah, you're right, this is just the early versions, thanks!

andrewlock: Reverted this change in https://github.com/DataDog/dd-trace-dotnet/pull/7903/commits/0d03dcae69ef88fe0d57d7a7f0204dcbdbe99b95

PR #7912, opened 2025-12-05
Use a null diagnostic observer to check impact

andrewlock:


andrewlock:
Part of R&D for #7964

PR #7917, opened 2025-12-08
[profiler] Refactor linux unwinder and remove unused one

gleocadie:
## Summary of changes

Extract the unwinder to its class + remove the old manual unwinding (never used)

## Reason for change

- Remove dead code
- Refactor the unwinder into its own class to allow other unwinder implementations (ex: ARM64)

## Implementation details

- Create IUnwinder + Backtrace2Unwinder classes
- Inject

## Test coverage

Current tests are enough.

### DIFF
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
@@ -564,12 +564,7 @@ void CorProfilerCallback::InitializeServices()
 #ifdef LINUX
     if (_pConfiguration->IsCpuProfilingEnabled() && _pConfiguration->GetCpuProfilerType() == CpuProfilerType::TimerCreate)
     {
-        _pUnwinder =
-#ifdef AMD64
-            std::make_unique<Backtrace2Unwinder>();
-#else
-            std::make_unique<HybridUnwinder>();
-#endif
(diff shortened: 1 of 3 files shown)


profiler/src/ProfilerEngine/Datadog.Profiler.Native.Linux/IUnwinder.h
```diff
@@ -0,0 +1,12 @@
...
+
+#include <cstdint>
+
+class IUnwinder
+{
+public:
+    virtual ~IUnwinder() = default;
+
+    // Returns the number of frames unwound
+    virtual std::int32_t Unwind(void* ctx, std::uintptr_t* buffer, std::size_t bufferSize) const = 0;
```
gleocadie, line 11: unwinder must be stateless (we will use one for all threads in timer_create-based CPU profiler)

profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
```diff
@@ -1323,7 +1331,7 @@ void CorProfilerCallback::PrintEnvironmentVariables()
     // --> should we dump the important ones to ensure that we get them during support investigations?
 
     Log::Info("Environment variables:");
-    PRINT_ENV_VAR_IF_SET(EnvironmentVariables::UseBacktrace2);
+    // TODO
```
chrisnas: maybe remove the "title" log if we log nothing

PR #7920, opened 2025-12-09
[Tracing] Instrument static methods defined in non-generic value types

zacharycmontoya:
## Summary of changes
Adds the capability to instrument static methods defined in non-generic value types

## Reason for change
We will need to instrument OpenTelemetry Baggage, which utilizes static methods in a Baggage struct

## Implementation details
- When instrumenting a method for CallTarget instrumentation, optionally update the LocalVarSig with a local variable of the owning type if the following conditions are met:
  1. The method is static
  2. The owning type of the method is a value type
- Update `TracerMethodRewriter::Rewrite` such that when it is loading the object instance to put on the stack for a static method, for non-generic value types it does the following:
  - `ldloca.s [localIndex]` (Loads the variable addresses)
  - `initobj [valueType]` (Initializes, or re-initializes, an empty struct)
  - `ldloc.s [localIndex]` (Loads the struct onto the stack)

## Test coverage
The `CallTargetNativeTest` is updated with the following cases to demonstrate the new functionality, especially as it compares to reference types:
- New `ArgumentsGenericStatic` types demonstrate the successful instrumentation of static methods in _generic_ reference types
- New `ArgumentsStaticStruct` types demonstrate the successful instrumentation of static methods in value types
  - New `ArgumentsGenericStaticStruct` types demonstrate the unsuccessful instrumentation of static methods in _generic_ value types

## Other details
Note: This does not implement the functionality for the live debugger IL rewriting, though it does update the method call so that everything compiles.

### DIFF
--- a/tracer/test/test-applications/instrumentation/CallTargetNativeTest/With0Arguments.cs
+++ b/tracer/test/test-applications/instrumentation/CallTargetNativeTest/With0Arguments.cs
@@ -123,6 +123,7 @@ partial class Program
         RunMethod(() => With0ArgumentsStaticStruct.ReturnGenericMethod<int>());
         Console.WriteLine();
         //
+        // Note: We skip instrumentation today for generic value types.
         Console.WriteLine($"{typeof(With0ArgumentsGenericStaticStruct<string>).FullName}.VoidMethod");
         RunMethod(() => With0ArgumentsGenericStaticStruct<string>.VoidMethod(), checkInstrumented: false);
         Console.WriteLine($"{typeof(With0ArgumentsGenericStaticStruct<string>).FullName}.ReturnValueMethod");
@@ -133,6 +134,7 @@ partial class Program
         RunMethod(() => With0ArgumentsGenericStaticStruct<string>.ReturnGenericMethod(),checkInstrumented: false);
         Console.WriteLine();
         //
+        // Note: We skip instrumentation today for generic value types.
         Console.WriteLine($"{typeof(With0ArgumentsGenericStaticStruct<int>).FullName}.VoidMethod");
         RunMethod(() => With0ArgumentsGenericStaticStruct<int>.VoidMethod(),checkInstrumented: false);
         Console.WriteLine($"{typeof(With0ArgumentsGenericStaticStruct<int>).FullName}.ReturnValueMethod");
@@ -317,6 +319,7 @@ partial class Program
         RunMethod(() => ArgumentsParentType.With0ArgumentsStaticStruct.ReturnGenericMethod<int>());
         Console.WriteLine();
         //
+        // Note: We skip instrumentation today for generic value types.
         Console.WriteLine($"{typeof(ArgumentsParentType.With0ArgumentsGenericStaticStruct<string>).FullName}.VoidMethod");
         RunMethod(() => ArgumentsParentType.With0ArgumentsGenericStaticStruct<string>.VoidMethod(), checkInstrumented: false);
         Console.WriteLine($"{typeof(ArgumentsParentType.With0ArgumentsGenericStaticStruct<string>).FullName}.ReturnValueMethod");
@@ -327,6 +330,7 @@ partial class Program
         RunMethod(() => ArgumentsParentType.With0ArgumentsGenericStaticStruct<string>.ReturnGenericMethod(), checkInstrumented: false);
         Console.WriteLine();
         //
+        // Note: We skip instrumentation today for generic value types.
         Console.WriteLine($"{typeof(ArgumentsParentType.With0ArgumentsGenericStaticStruct<int>).FullName}.VoidMethod");
         RunMethod(() => ArgumentsParentType.With0ArgumentsGenericStaticStruct<int>.VoidMethod(), checkInstrumented: false);
         Console.WriteLine($"{typeof(ArgumentsParentType.With0ArgumentsGenericStaticStruct<int>).FullName}.ReturnValueMethod");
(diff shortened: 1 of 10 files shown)


tracer/src/Datadog.Tracer.Native/tracer_method_rewriter.cpp
```diff
@@ -209,17 +209,35 @@ HRESULT TracerMethodRewriter::Rewrite(RejitHandlerModule* moduleHandler, RejitHa
...
-            // value. After the signature modification we need to emit the following IL to initialize and load into the
-            // stack.
-            //    ldloca.s [localIndex]
-            //    initobj [valueType]
-            //    ldloc.s [localIndex]
-            Logger::Warn("*** CallTarget_RewriterCallback(): Static methods in a ValueType cannot be instrumented. ");
-            return S_FALSE;
+            reWriterWrapper.LoadLocalAddress(staticValueTypeIndex);
+            if (caller->type.type_spec != mdTypeSpecNil)
+            {
```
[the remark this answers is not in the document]
zacharycmontoya: This was useful as I didn't cover this edge case. Although we don't instrument this case today, in our `CallTargetNativeTest` program we can enable this scenario. With the latest set of commits, we correctly handle this and load a `null` value as the CallTarget "instance" so we can in fact instrument this scenario.

tracer/src/Datadog.Tracer.Native/calltarget_tokens.cpp
```diff
@@ -399,7 +399,17 @@ HRESULT CallTargetTokens::ModifyLocalSig(ILRewriter* reWriter, TypeSignature* me
     }
     constexpr int variableNumber = 3;
     const auto additionalLocalsCount = GetAdditionalLocalsCount(*methodTypeArguments);
-    ULONG newLocalsCount = variableNumber + additionalLocalsCount;
+
+    // For instrumenting static methods in value types, see if we can get the TypeRef for the type.
+    // If so, include add a local variable with the value type in the signature.
+    bool isValueType = caller->type.valueType;
+    bool isStaticValueType = !(caller->method_signature.CallingConvention() & IMAGE_CEE_CS_CALLCONV_HASTHIS) && caller->type.valueType && GetCurrentTypeRef(&caller->type, isValueType) != mdTokenNil;
```
andrewlock: nit: 
```suggestion
    bool isStaticValueType = isValueType && !(caller->method_signature.CallingConvention() & IMAGE_CEE_CS_CALLCONV_HASTHIS) && GetCurrentTypeRef(&caller->type, isValueType) != mdTokenNil;
```

tracer/src/Datadog.Tracer.Native/tracer_method_rewriter.cpp
```diff
@@ -207,19 +207,39 @@ HRESULT TracerMethodRewriter::Rewrite(RejitHandlerModule* moduleHandler, RejitHa
     // *** Load instance into the stack (if not static)
     if (isStatic)
     {
-        if (caller->type.valueType)
+        bool callerTypeIsValueType = caller->type.valueType;
+        mdToken callerTypeToken = tracerTokens->GetCurrentTypeRef(&caller->type, callerTypeIsValueType);
+        if (caller->type.valueType && callerTypeToken != mdTokenNil)
```
andrewlock: nit:
```suggestion
        if (callerTypeIsValueType && callerTypeToken != mdTokenNil)
```

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/CallTargetNativeTests.cs
```diff
@@ -20,6 +20,9 @@ public CallTargetNativeTests(ITestOutputHelper output)
             : base(new EnvironmentHelper("CallTargetNativeTest", typeof(TestHelper), output, samplesDirectory: Path.Combine("test", "test-applications", "instrumentation"), prependSamplesToAppName: false), output)
         {
             SetServiceVersion("1.0.0");
+#pragma warning disable CS0618 // Type or member is obsolete
+            EnableDebugMode();
+#pragma warning restore CS0618 // Type or member is obsolete
```
andrewlock: nit: we shouldn't use debug mode permenantly, as it means we never test release mode 🙂  We have a "debug mode" run that runs nightly instead
```suggestion
```

tracer/test/test-applications/instrumentation/CallTargetNativeTest/With0Arguments.cs
```diff
@@ -91,6 +111,38 @@ private static void Argument0()
...
+        Console.WriteLine($"{typeof(With0ArgumentsStaticStruct).FullName}.ReturnValueMethod");
+        RunMethod(() => With0ArgumentsStaticStruct.ReturnValueMethod());
+        Console.WriteLine($"{typeof(With0ArgumentsStaticStruct).FullName}.ReturnReferenceMethod");
+        RunMethod(() => With0ArgumentsStaticStruct.ReturnReferenceMethod());
+        Console.WriteLine($"{typeof(With0ArgumentsStaticStruct).FullName}.ReturnGenericMethod<string>");
+        RunMethod(() => With0ArgumentsStaticStruct.ReturnGenericMethod<string>());
+        Console.WriteLine($"{typeof(With0ArgumentsStaticStruct).FullName}.ReturnGenericMethod<int>");
+        RunMethod(() => With0ArgumentsStaticStruct.ReturnGenericMethod<int>());
+        Console.WriteLine();
+        //
```
andrewlock, line 125: nit: might be nice to mention somewhere (in these comments perhaps, that these _don't_ work today). I know we have `checkInstrumented: false` for these, but might be hard to spot 🤷‍♂️

andrewlock (approved):
Nice, thanks! Just tiny nits/suggestions!

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/CallTargetNativeTests.cs
zacharycmontoya: Oops, my bad. Thanks for spotting this!

tracer/test/test-applications/instrumentation/CallTargetNativeTest/With0Arguments.cs
zacharycmontoya: Good idea, added some comments in b58ce81a349400035d99ec110b6763b7311e4960

PR #7921, opened 2025-12-09
[Tracing] Add interoperability between the Datadog Baggage API and the OpenTelemetry Baggage API

zacharycmontoya:
## Summary of changes
Enable updates to Baggage from either the Datadog Baggage API and the OpenTelemetry Baggage API to be visible to the other.

Relies on #7920 for properly instrumenting the static methods in the `OpenTelemetry.Baggage` struct.

## Reason for change
This allows users of either Baggage API to make changes to the automatically-propagated Baggage header. Before, only changes made through the Datadog Baggage API would be reflected.

### APIs supported
Instruments the following methods in the OpenTelemetry Baggage API:
- `static OpenTelemetry.Baggage.ClearBaggage(Baggage)`
- `static OpenTelemetry.Baggage.Current { get; set; }`
- `static OpenTelemetry.Baggage.RemoveBaggage(string, Baggage)`
- `static OpenTelemetry.Baggage.SetBaggage(string, string, Baggage)`
- `static OpenTelemetry.Baggage.SetBaggage(IEnumerable<KeyValuePair<string, string?>>, Baggage)`

The following public API's are also supported indirectly by our instrumentation of the `OpenTelemetry.Baggage.Current` getter:
- `static OpenTelemetry.Baggage.GetBaggage(Baggage)`
- `static OpenTelemetry.Baggage.GetBaggage(string, Baggage)`
- `static OpenTelemetry.Baggage.GetEnumerator(Baggage)`

### OpenTelemetry Baggage Model
In the OpenTelemetry model, baggage is immutable. This means every time the user wants to update `OpenTelemetry.Baggage.Current`, they must either call the setter directly with an new instance of `OpenTelemetry.Baggage`, or they must call one of the static APIs.

The static APIs are largely intuitive, but it's worth calling two interesting behaviors of the APIs that modify baggage:
1. If a `OpenTelemetry.Baggage` argument is provided (optional), that object will be the source from which the modification will be enacted on. If not, then `OpenTelemetry.Baggage.Current` will be the source.
2. Before returning from the API call, the resulting (immutable) `OpenTelemetry.Baggage` is set as the new `OpenTelemetry.Baggage.Current` value.

### Interoperability Approach
For this approach, `Datadog.Trace.Baggage.Current` will be the source of truth for the "current" baggage. For example, our HTTP server instrumentations will first populate `Datadog.Trace.Baggage.Current` with the contents of the incoming `baggage` HTTP header, and users of the `Datadog.Trace.Baggage` API can update the mutable `Datadog.Trace.Baggage.Current`. Then, if the user wants get or set `OpenTelemetry.Baggage.Current`, that is where our instrumentation must keep it in sync with `Datadog.Trace.Baggage.Current`.

Each time `OpenTelemetry.Baggage.set_Current` is invoked, we must replace the contents of `Datadog.Trace.Baggage.Current` entirely by clearing out the existing values and add all of the new KeyValuePairs.

Each time `OpenTelemetry.Baggage.get_Current` is invoked, we must create a new (immutable) `OpenTelemetry.Baggage` using the contents of `Datadog.Trace.Baggage.Current` and assign it to `OpenTelemetry.Baggage.Current` before the property is accessed and returned to the API caller.

By design, no work is done if only the `Datadog.Trace.Baggage` APIs are used. Also, users are aware of the immutable nature of the `OpenTelemetry.Baggage` APIs, so they will be used to doing at least one allocation per baggage modification, but we can certainly improve the overhead of this interop with approaches noted in the 'Other details' section below.

## Test coverage
Added a new `OtelBaggageController` to the `Samples.AspNetCoreMinimalApis` and `Samples.AspNetCoreMvc*` applications, and added integration test cases (with snapshots) to `tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNetCore/AspNetCoreMinimalApisTests.cs`.

## Other details
We can follow up this PR with changes to optimize the sharing of Baggage state between the two API's. In order to do this, we would need to do the following:
- Change the internal storage mechanism of items in `Datadog.Trace.Baggage` to be `Dictionary<string,string>`
- Change `Datadog.Trace.Baggage` to disallow `null` values in baggage key-value pairs, to align with the OpenTelemetry storage model
- When returning from the static `OpenTelemetry.Baggage` methods that modify state, take the reference to the internal `Dictionary<string,string>` object and assign it to the `Datadog.Trace.Baggage.Current` items dictionary

More immediately, we could implement tracking of writes to `Datadog.Trace.Baggage.Current` so we don't have to create a fresh copy of it every time `OpenTelemetry.Baggage.get_Current` is called, but I'd rather separate that out from this PR.


tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNetCore/AspNetCoreMinimalApisTests.cs
```diff
@@ -108,6 +108,53 @@ await Verifier.Verify(spans, settings)
...
+        [InlineData("/otel-baggage/set-baggage-items/foo_case_sensitive_key/overwrite_value", 200)]
+        [InlineData("/otel-baggage/set-baggage-items/new_key/new_value", 200)]
+        [InlineData("/otel-baggage/set-current/foo_case_sensitive_key/overwrite_value", 200)]
+        [InlineData("/otel-baggage/set-current/new_key/new_value", 200)]
+        public async Task OtelBaggageApiIntegration(string path, int statusCode)
+        {
+            SetInstrumentationVerification();
+
+            await Fixture.TryStartApp(this);
+            string[] baggageItems = [
```
zacharycmontoya, line 133: Note for reviewers: All of the test cases start with this baggage as the original contents. However, the OpenTelemetry Baggage API currently treats keys in a case-insensitive manner, so the final baggage output will be missing one of the baggage items below whose keys only differ by case. This is not aligned with our Datadog implementation and more importantly with the OpenTelemetry specification, so I'll submit a PR upstream to fix that. Once that change is shipped, we can then update the `OpenTelemetry.Api` package and update our snapshots.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/IApiBaggage.cs
```diff
@@ -0,0 +1,35 @@
...
+
+#nullable enable
+
+using System.Collections.Generic;
+using Datadog.Trace.DuckTyping;
+
+namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.OpenTelemetry
+{
+    /// <summary>
+    /// OpenTelemetry.Baggage interface for duck-typing
```
andrewlock, line 14: nit: it can be useful to have stable links to the APIs being instrumented in the comments if possible
```suggestion
    /// OpenTelemetry.Baggage interface for duck-typing
    /// https://github.com/open-telemetry/opentelemetry-dotnet/blob/db429bf642c1a2c2f71b49f88d63e0a661018298/src/OpenTelemetry.Api/Baggage.cs#L16
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/IApiBaggage.cs
```diff
@@ -0,0 +1,35 @@
...
+        IReadOnlyDictionary<string, string?> GetBaggage();
+
+        string? GetBaggage(string name);
+
+        IBaggageHolder EnsureBaggageHolder();
+    }
+
+    internal interface IBaggageHolder
+    {
+        [DuckField(Name = "Baggage")]
```
andrewlock, line 37: OMG the casing inconsistency in this code base 😅

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_ClearBaggageIntegration.cs
```diff
@@ -0,0 +1,59 @@
...
+    /// OpenTelemetry.Api Baggage.ClearBaggage calltarget instrumentation
+    /// </summary>
+    [InstrumentMethod(
+        AssemblyName = "OpenTelemetry.Api",
+        TypeName = "OpenTelemetry.Baggage",
+        MethodName = "ClearBaggage",
+        ReturnTypeName = "OpenTelemetry.Baggage",
+        ParameterTypeNames = new[] { "OpenTelemetry.Baggage" },
+        MinimumVersion = "1.0.0",
+        MaximumVersion = "1.0.0",
```
andrewlock, line 27: Is this really correct, we're strictly instrumenting version 1.0.0 _only_? That would be unusual...

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_ClearBaggageIntegration.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// <summary>
+        /// OnMethodEnd callback
+        /// </summary>
+        /// <typeparam name="TTarget">Type of the target</typeparam>
+        /// <typeparam name="TReturn">Type of the return value</typeparam>
+        /// <param name="instance">Instance value, aka `this` of the instrumented method.</param>
+        /// <param name="returnValue">Return value</param>
+        /// <param name="exception">Exception instance in case the original code threw an exception.</param>
+        /// <param name="state">Calltarget state value</param>
+        /// <returns>A response value, in an async scenario will be T of Task of T</returns>
```
andrewlock: We generally don't include these comments as they're basically just noise - did you use the auto instrumentation generator? (I'd recommend it if not!) (applies to all the integrations)

```suggestion
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_ClearBaggageIntegration.cs
```diff
@@ -0,0 +1,59 @@
...
+        internal static CallTargetReturn<TReturn> OnMethodEnd<TTarget, TReturn>(TTarget instance, TReturn returnValue, Exception exception, in CallTargetState state)
+        {
+            if (Tracer.Instance.CurrentTraceSettings.Settings.IsIntegrationEnabled(IntegrationId))
+            {
+                // Important: Before returning to the caller, the static OpenTelemetry.Baggage APIs set the to-be-returned baggage object as OpenTelemetry.Baggage.Current.
+                // However, this is not done through the public setter (which we instrument), but through the backing field, so we must manually update Datadog.Trace.Baggage.Current
+                // so it remains in-sync.
+                Baggage.Current.Clear();
```
andrewlock, line 43: If an exception was thrown in the method call (i.e. in `OpenTelemetry.Baggage.ClearBaggage()`) _should_ we still clear our internal baggage?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_GetCurrentIntegration.cs
```diff
@@ -0,0 +1,66 @@
...
+        ReturnTypeName = "OpenTelemetry.Baggage",
+        ParameterTypeNames = new string[0],
+        MinimumVersion = "1.0.0",
+        MaximumVersion = "1.0.0",
+        IntegrationName = IntegrationName)]
+    [Browsable(false)]
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public sealed class OTelBaggage_GetCurrentIntegration
+    {
+        internal const string IntegrationName = nameof(Configuration.IntegrationId.OpenTelemetry);
```
andrewlock: nit: can probably just inline all these in the attribute

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_GetCurrentIntegration.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// </summary>
+        /// <typeparam name="TInstance">Type of the instance (null for this static method)</typeparam>
+        /// <param name="instance">Instance value, aka `this` of the instrumented method.</param>
+        /// <returns>Calltarget state value</returns>
+        internal static CallTargetState OnMethodBegin<TInstance>(TInstance instance)
+        {
+            if (Tracer.Instance.CurrentTraceSettings.Settings.IsIntegrationEnabled(IntegrationId)
+                && OTelBaggageType is not null)
+            {
+                DuckType.CreateTypeResult proxyResult = DuckType.GetOrCreateProxyType(typeof(IApiBaggage), OTelBaggageType);
```
andrewlock: I'm not sure if we need `OTelBaggageType` - I think we _should_ be able to use `typeof(TInstance)`? 🤔 `instance` will be null, but I believe `TInstance` is the correct type?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_GetCurrentIntegration.cs
```diff
@@ -0,0 +1,66 @@
...
+                if (proxyResult.Success)
+                {
+                    var apiBaggage = proxyResult.CreateInstance<IApiBaggage>(Activator.CreateInstance(OTelBaggageType));
+
+                    // Since Datadog.Trace.Baggage.Current may have been updated since the last time OpenTelemetry.Baggage.Current was accessed,
+                    // we must update the underlying OpenTelemetry.Baggage.Current store with the latest Datadog.Trace.Baggage.Current items.
+                    // Note: When the user sets OpenTelemetry.Baggage.Current, those changes will override the contents of Datadog.Trace.Baggage.Current,
+                    // so we can always consider Datadog.Trace.Baggage.Current as being up-to-date.
+                    var baggageHolder = apiBaggage.EnsureBaggageHolder();
+                    baggageHolder.Baggage = apiBaggage.Create(Baggage.Current.ToDictionary(kvp => kvp.Key, kvp => kvp.Value));
```
andrewlock: I was wondering about this, and got stuck in the weeds 😅 

First, the `Dictionary<>` header has fast-paths for when you pass in a `Dictionary<>`, `Array` or `List<KeyValuePair<>>`, so I was wondering if we could/should just do something like `new Dictionary(Baggage.Current.GetList())`, as our baggage implementation is backed by a list.

_However_, that wouldn't be thread safe, whereas it looks like our `Baggage` implementation _is_ nominally thread safe... except... it isn't, because we implement IEnumerable, which exposes the List enumerator (although incorrectly, because we _force_ it to be boxed) 😅

So in conclusion:

- Our `Baggage` implementation is not thread safe when it's enumerated with `foreach` - I wonder if we should _stop_ implementing `IDictionary<string, string?>`... do we _need_ it? Because currently it's dangerous
- If we added an `AsDictionary()` method (implemented below), it should be faster than the `ToDictionary<>` case?
- WDYT?

```csharp
internal sealed class Baggage
{
    public static Dictionary<string, string?> AsDictionary()
    {
        if (_items is { } items)
        {
            lock (items)
            {
                 return new(items);
            }
        }
    
        return [];
    }
}
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_RemoveBaggageIntegration.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// <summary>
+        /// OnMethodBegin callback
+        /// </summary>
+        /// <typeparam name="TInstance">Type of the instance (null for this static method)</typeparam>
+        /// <typeparam name="TBaggage">Type of the baggage</typeparam>
+        /// <param name="instance">Instance value, aka `this` of the instrumented method.</param>
+        /// <param name="key">Key to be removed from baggage.</param>
+        /// <param name="baggage">Optional baggage, or default if not specified.</param>
+        /// <returns>Calltarget state value</returns>
+        internal static CallTargetState OnMethodBegin<TInstance, TBaggage>(TInstance instance, string key, TBaggage baggage)
```
andrewlock: Can't we use our standard constraints on TBaggage here to avoid the `TryDuckCast`? are there cases we expect that not to work? 🤔 
```suggestion
        internal static CallTargetState OnMethodBegin<TInstance, TBaggage>(TInstance instance, string key, TBaggage baggage)
            where TBaggage : IApiBaggage
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_RemoveBaggageIntegration.cs
```diff
@@ -0,0 +1,97 @@
...
+                && returnValue.TryDuckCast<IApiBaggage>(out var apiBaggage))
+            {
+                // Important: Before returning to the caller, the static OpenTelemetry.Baggage APIs set the to-be-returned baggage object as OpenTelemetry.Baggage.Current.
+                // However, this is not done through the public setter (which we instrument), but through the backing field, so we must manually update Datadog.Trace.Baggage.Current
+                // so it remains in-sync.
+                //
+                // Additional notes:
+                // - Since the Datadog Baggage model is mutable (allowing the user to get the Datadog.Trace.Baggage.Current once and continue to mutate that reference),
+                //   we must clear then add the new baggage items.
+                // - The API can be invoked with an arbitrary OpenTelemetry.Baggage object passed via the parameter, so we must replace all baggage items every time
```
andrewlock, line 67: To clarify, if the user calls `RemoveBaggage("mykey", someBaggage)`, and passes in a baggage object, _that_ (updated) baggage is set to be the new `Baggage.Current` value, correct? i.e. all of the `SetBaggage` and `RemoveBaggage()` APIs always manipulate the `Current` baggage, passing in an explicit baggage object just manipulates the "starting" point?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_RemoveBaggageIntegration.cs
```diff
@@ -0,0 +1,97 @@
...
+            if (Tracer.Instance.CurrentTraceSettings.Settings.IsIntegrationEnabled(IntegrationId)
+                && baggage.TryDuckCast<IApiBaggage>(out var apiBaggage)
+                && apiBaggage.Baggage is null)
+            {
+                // Since Datadog.Trace.Baggage.Current may have been updated since the last time OpenTelemetry.Baggage.Current was accessed,
+                // we must update the underlying OpenTelemetry.Baggage.Current store with the latest Datadog.Trace.Baggage.Current items.
+                // Note: When the user sets OpenTelemetry.Baggage.Current, those changes will override the contents of Datadog.Trace.Baggage.Current,
+                // so we can always consider Datadog.Trace.Baggage.Current as being up-to-date.
+                var baggageHolder = apiBaggage.EnsureBaggageHolder();
+                baggageHolder.Baggage = apiBaggage.Create(Baggage.Current.ToDictionary(kvp => kvp.Key, kvp => kvp.Value));
```
andrewlock: Given the OTel APIs treat baggage in a case-insensitive manner, what happens if we set the same key here? Are the duplicate keys overwritten, or does it throw?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_ClearBaggageIntegration.cs
zacharycmontoya: Yes, the `OpenTelemetry.Api` 1.x packages all share the `1.0.0` assembly version 😄

andrewlock: Ah I see, this is the old "keep your assembly version static" nonsense. In which case that's fine, but still makes me wonder if we should stick to the same patterns as elsewhere, just in case they do start incrementing them. Not bothered either way tbh in that case 🤷‍♂️

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_RemoveBaggageIntegration.cs
zacharycmontoya: No exception, the entry just gets overwritten. But I'm resolving this with https://github.com/open-telemetry/opentelemetry-dotnet/pull/6931

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_RemoveBaggageIntegration.cs
zacharycmontoya: Yes. Haha.

For the static methods, the baggage update will take place on the baggage object passed in or `Baggage.Current` if nothing is passed in. Then the resulting baggage is set as the new value of `Baggage.Current`.

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNetCore/AspNetCoreMinimalApisTests.cs
zacharycmontoya: I've posted a PR upstream to get the OTel API to use case-sensitive baggage names: https://github.com/open-telemetry/opentelemetry-dotnet/pull/6931

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_ClearBaggageIntegration.cs
zacharycmontoya: Hmm good question. The assignment to OpenTelemetry's `Baggage.Current` is the final thing that happens in the OpenTelemetry API method, so if an exception is thrown then that assignment would not be run. In that case, there's no update, so we also should not update our own Baggage. I'll make a follow-up change to address this

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_GetCurrentIntegration.cs
zacharycmontoya: I'll give it a try and report back!

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_RemoveBaggageIntegration.cs
zacharycmontoya: I think originally I was trying to test a case where `baggage == null` and then I moved away from it. I should be able to get this to work with the type constraint, so I'll try that out

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_GetCurrentIntegration.cs
zacharycmontoya: You're right, in this case because of the updated instrumentation, the type `TInstance` is set correctly and we pass a default value of that value type into the method. This is now handled in 5d067d8a7fdd5d0447c0e72acc1a29d1d9b5a508, which is much better now. Thanks for the feedback!

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_GetCurrentIntegration.cs
zacharycmontoya: Done in acfc98ddfaa4d529dc68908d40a7922893faa376

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/IApiBaggage.cs
zacharycmontoya: Done in e9ef73336b37cf4e9fe175355da7c4876c059fc6

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_ClearBaggageIntegration.cs
zacharycmontoya: Done in 7993bf7b68cb214e580ccd5f955942fbb107cc9d

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_RemoveBaggageIntegration.cs
zacharycmontoya: Done in 7a603e10739440e684f11965a7a0fcec7f9a1adb

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_ClearBaggageIntegration.cs
zacharycmontoya: I've added a new commit ae367aabf2e8231b59633e9222fee837ddc6857f so we only update our `Datadog.Trace.Baggage.Current` state if there is no exception, which is required for OpenTelemetry to successfully update its current state

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_GetCurrentIntegration.cs
zacharycmontoya: AFAIK we currently have our `Baggage` class implement `IDictionary<string, string?>` so that we can easily pass it back and forth in our `Datadog.Trace.Manual` instrumentation, since it's a BCL type, right @lucaspimentel?

It would be a breaking change to change the public type, but we could do it. Or we could tell users "actually, we're deprecating our own Baggage API in favor of the OpenTelemetry Baggage API, so you need to install the `OpenTelemetry.Api` package to use baggage". I actually don't mind the second one because then it limits how many times we copy state back and forth 🤔 

What are your thoughts on this?

andrewlock: > AFAIK we currently have our Baggage class implement IDictionary<string, string?> so that we can easily pass it back and forth in our Datadog.Trace.Manual instrumentation

Ooh, yeah, you're right, I missed that! We should leave it as-is for now IMO, and we can consider deprecating in the future...

I think the other statements still hold though:

- We should add an `AsDictionary()` method to `Baggage` so we can both safely enumerate and fast-path the creation of the `Dictionary` (by passing in the `List<>`)
- We should publicly expose the struct-based `Baggage.GetEnumerator()` so the compiler can use it in `foreach`!

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_GetCurrentIntegration.cs
```diff
@@ -0,0 +1,52 @@
...
+            where TInstance : IApiBaggage
+        {
+            if (Tracer.Instance.CurrentTraceSettings.Settings.IsIntegrationEnabled(IntegrationId))
+            {
+                // Since Datadog.Trace.Baggage.Current may have been updated since the last time OpenTelemetry.Baggage.Current was accessed,
+                // we must update the underlying OpenTelemetry.Baggage.Current store with the latest Datadog.Trace.Baggage.Current items.
+                // Note: When the user sets OpenTelemetry.Baggage.Current, those changes will override the contents of Datadog.Trace.Baggage.Current,
+                // so we can always consider Datadog.Trace.Baggage.Current as being up-to-date.
+                var baggageHolder = apiBaggage.EnsureBaggageHolder();
+                baggageHolder.Baggage = apiBaggage.Create(Baggage.Current.ToDictionary(kvp => kvp.Key, kvp => kvp.Value));
```
andrewlock: As discussed in a separate comment, I think an `AsDictionary()` method would be beneficial (for safety + perf)
```suggestion
                baggageHolder.Baggage = apiBaggage.Create(Baggage.Current.AsDictionary());
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_SetBaggageIntegration.cs
```diff
@@ -0,0 +1,98 @@
...
+        /// <param name="instance">Instance value, aka `this` of the instrumented method.</param>
+        /// <param name="returnValue">Return value</param>
+        /// <param name="exception">Exception instance in case the original code threw an exception.</param>
+        /// <param name="state">Calltarget state value</param>
+        /// <returns>A response value, in an async scenario will be T of Task of T</returns>
+        internal static CallTargetReturn<TReturn> OnMethodEnd<TTarget, TReturn>(TTarget instance, TReturn returnValue, Exception exception, in CallTargetState state)
+        {
+            if (Tracer.Instance.CurrentTraceSettings.Settings.IsIntegrationEnabled(IntegrationId)
+                && returnValue.TryDuckCast<IApiBaggage>(out var apiBaggage))
+            {
```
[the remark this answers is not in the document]
andrewlock: We're doing this correctly now I believe 🤔

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_SetCurrentIntegration.cs
```diff
@@ -0,0 +1,57 @@
...
+    [Browsable(false)]
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public sealed class OTelBaggage_SetCurrentIntegration
+    {
+        internal const IntegrationId IntegrationId = Configuration.IntegrationId.OpenTelemetry;
+
+        internal static CallTargetState OnMethodBegin<TInstance, TBaggage>(TInstance instance, TBaggage value)
+            where TBaggage : IApiBaggage
+        {
+            if (Tracer.Instance.CurrentTraceSettings.Settings.IsIntegrationEnabled(IntegrationId))
```
andrewlock, line 38: I guess we're assuming `set_Current()` can never throw - should we do this in the return value just in case (and check for `Exception`)?

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNetCore/AspNetCoreMvcTestBase.cs
```diff
@@ -61,6 +61,9 @@ protected AspNetCoreMvcTestBase(string sampleName, AspNetCoreTestFixture fixture
             SetEnvironmentVariable(ConfigurationKeys.FeatureFlags.RouteTemplateResourceNamesEnabled, (flags == AspNetCoreFeatureFlags.RouteTemplateResourceNames).ToString());
             SetEnvironmentVariable(ConfigurationKeys.FeatureFlags.SingleSpanAspNetCoreEnabled, (flags == AspNetCoreFeatureFlags.SingleSpan).ToString());
 
+            // Enable OpenTelemetry interop to test that the OTel Baggage API integration works correctly.
+            SetEnvironmentVariable("DD_TRACE_OTEL_ENABLED", "true");
```
andrewlock, line 65: We should probably set this in _all_ of our aspnetcore fixtures, regardless, so that we can catch any cases of weird regressions? 🤔 WDYT? We could do it in a separate PR obviously

tracer/test/snapshots/AspNetCoreMinimalApisTests.Single._OTelBaggageApi_path=_otel-baggage_clear-baggage_statusCode=200.verified.txt
```diff
@@ -0,0 +1,49 @@
...
+    TraceId: Id_1,
+    SpanId: Id_2,
+    Name: ClearBaggage,
+    Resource: ClearBaggage,
+    Service: Samples.AspNetCoreMinimalApis,
+    ParentId: Id_3,
+    Tags: {
+      env: integration_tests,
+      language: dotnet,
+      otel-baggage: ,
```
andrewlock: Should we still be adding the `Otel-Baggage` tag when it's empty?

tracer/test/snapshots/AspNetCoreMinimalApisTests.NoFF._OTelBaggageApi_path=_otel-baggage_set-baggage_foo_case_sensitive_key_overwrite_value_statusCode=200.verified.txt
```diff
@@ -0,0 +1,46 @@
...
+    TraceId: Id_1,
+    SpanId: Id_2,
+    Name: SetBaggage,
+    Resource: SetBaggage,
+    Service: Samples.AspNetCoreMinimalApis,
+    ParentId: Id_3,
+    Tags: {
+      env: integration_tests,
+      language: dotnet,
+      otel-baggage: foo_case_sensitive_key=overwrite_value,unused_key=unused_value,remove_me_key=remove_me_value,
```
andrewlock, line 12: is `remove_me_key=remove_me_Value` _meant to be here? I lost track 😅

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_SetBaggageIntegration.cs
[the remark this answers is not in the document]
zacharycmontoya: Thanks Codex, yes this should be resolved now. I'll go ahead and resolve this thread

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_GetCurrentIntegration.cs
zacharycmontoya: Sounds good, I'll add that 👍🏼

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/AspNetCore/AspNetCoreMvcTestBase.cs
zacharycmontoya: Yep sounds good to me!

tracer/test/snapshots/AspNetCoreMinimalApisTests.Single._OTelBaggageApi_path=_otel-baggage_clear-baggage_statusCode=200.verified.txt
zacharycmontoya: This is hand-written by custom instrumentation in the test application, but I think it makes sense to remove it when empty. I'll go ahead and do that

tracer/test/snapshots/AspNetCoreMinimalApisTests.NoFF._OTelBaggageApi_path=_otel-baggage_set-baggage_foo_case_sensitive_key_overwrite_value_statusCode=200.verified.txt
zacharycmontoya: Heh sorry, yes it's supposed to be there. We set the baggage item `remove_me_key=remove_me_value` in all of the test cases, and then when we specifically hit the endpoint `/otel-baggage/remove-baggage/remove_me_key` we call `OpenTelemetry.Baggage.RemoveBaggage` with the given key.

TLDR; this baggage item should exist in the `otel-baggage` tag on all snapshots EXCEPT the one for endpoint `/otel-baggage/remove-baggage/remove_me_key`

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_SetCurrentIntegration.cs
zacharycmontoya: Yeah good idea, will do. I originally did this in OnMethodBegin because the new baggage is passed via the `Baggage` argument, but I can simply pass this into our `CallTargetState` so it makes it to OnMethodEnd

zacharycmontoya: Fixed in 651312414b3636278decc8c2ebf094dc04562ead

tracer/test/snapshots/AspNetCoreMinimalApisTests.Single._OTelBaggageApi_path=_otel-baggage_clear-baggage_statusCode=200.verified.txt
zacharycmontoya: Fixed in 4cf6920049bce6b4016d68bd5cd4bf65a560d7b8

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_GetCurrentIntegration.cs
zacharycmontoya: Done in af3062616c1b214458f7ee82bb044833b7837e25

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/OpenTelemetry/OTelBaggage_GetCurrentIntegration.cs
zacharycmontoya: > I think the other statements still hold though:
> - We should add an AsDictionary() method to Baggage so we can both safely enumerate and fast-path the creation of the Dictionary (by passing in the List<>)
> - We should publicly expose the struct-based Baggage.GetEnumerator() so the compiler can use it in foreach!

Done in af3062616c1b214458f7ee82bb044833b7837e25

PR #7926, opened 2025-12-09
Add helper for taking memory dump in CI on mutex timeout

andrewlock:
## Summary of changes

Adds a helper, `WaitOrDump` which waits for a mutex, and takes a memory dump if it fails

## Reason for change

We've been seeing unexpected timeouts of mutexes in CI, and we want to get to the bottom of them

## Implementation details

Use the existing `MemoryDumpHelper.CaptureMemoryDump()` to take a memory dump of the test process. I kept the signature essentially the same as `Wait` for ease of replacing calls. We can pass in `Output` in the future for more details, just doing the minimum right now

## Test coverage

This is the test - if we don't see any failures, I'll retry until we do!

### DIFF
--- a/tracer/test/Datadog.Trace.Tests/Util/MutexHelper.cs
+++ b/tracer/test/Datadog.Trace.Tests/Util/MutexHelper.cs
@@ -42,18 +42,19 @@ internal static class MutexHelper
         if (!result)
         {
             output?.WriteLine($"Mutex timed out in {caller} after {timeout.TotalSeconds:F1}s. Capturing memory dump");
-          try
-          {
-              var process = Process.GetCurrentProcess();
-              var log = MemoryDumpHelper.CaptureMemoryDump(process)
-                            ? "Successfully captured memory dump"
-                            : "Failed to capture memory dump";
-              output?.WriteLine(log);
-          }
-          catch (Exception ex)
-          {
-              output?.WriteLine($"Exception while capturing memory dump: {ex}");
-          }
+
+            try
+            {
+                var process = Process.GetCurrentProcess();
+                var log = MemoryDumpHelper.CaptureMemoryDump(process)
+                                ? "Successfully captured memory dump"
+                                : "Failed to capture memory dump";
+                output?.WriteLine(log);
+            }
+            catch (Exception ex)
+            {
+                output?.WriteLine($"Exception while capturing memory dump: {ex}");
+            }
         }
 
         return result;


tracer/test/Datadog.Trace.Tests/Agent/StatsAggregatorTests.cs
```diff
@@ -58,7 +58,7 @@ public async Task CallFlushAutomatically()
                     // Flush is not called if no spans are processed
                     aggregator.Add(new Span(new SpanContext(1, 1), DateTime.UtcNow));
 
-                    if (mutex.Wait(TimeSpan.FromMilliseconds(100)))
+                    if (mutex.WaitOrDump(TimeSpan.FromMilliseconds(100)))
```
tonyredondo, line 61: I'm worried about this one, 100ms is not that much considering that this is running in CI (all those VMs). Maybe we can bump this one a little bit more?

tracer/test/Datadog.Trace.Tests/Logging/DirectSubmission/Sink/DatadogSinkTests.cs
```diff
@@ -202,13 +203,13 @@ bool LogsSentCallback(int x)
...
                 sink.EnqueueLog(new TestLogEvent(DirectSubmissionLogLevel.Debug, "A message"));
-                mutex.Wait(30_000).Should().BeTrue();
+                mutex.WaitOrDump(30_000).Should().BeTrue();
                 mutex.Reset();
             }
 
             // circuit should be broken
             sink.EnqueueLog(new TestLogEvent(DirectSubmissionLogLevel.Debug, "A message"));
-            mutex.Wait(3_000).Should().BeFalse(); // don't expect it to be set
+            mutex.WaitOrDump(3_000).Should().BeFalse(); // don't expect it to be set
```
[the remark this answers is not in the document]
tonyredondo: oh he's right, how this compiles? do we have a cast operator for int -> TimeSpan?

tonyredondo (approved):

tracer/test/Datadog.Trace.Tests/Logging/DirectSubmission/Sink/DatadogSinkTests.cs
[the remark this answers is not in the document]
andrewlock: > how this compiles

haha, you're assuming I built this locally before I pushed? 😅

tracer/test/Datadog.Trace.Tests/Agent/StatsAggregatorTests.cs
andrewlock: I agree with you, though this one _isn't_ one we've seen fail IIRC, so I'm somewhat inclined to leave it for this PR (and bump it later)?

tracer/test/Datadog.Trace.Tests/Logging/DirectSubmission/Sink/DatadogSinkTests.cs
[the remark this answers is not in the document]
andrewlock: Fixed, added an overload

tracer/test/Datadog.Trace.Tests/Logging/DirectSubmission/Sink/DatadogSinkTests.cs
```diff
@@ -44,7 +45,7 @@ public void SinkSendsMessagesToLogsApi()
             sink.EnqueueLog(new TestLogEvent(DirectSubmissionLogLevel.Debug, "First message"));
```
NachoEchevarria, line 46: I was taking a look at these tests and wondered if we should add this:
await sink.FlushAsync();

This would make the test more deterministic and would avoid it from relying on timing, WDYT?

NachoEchevarria:
Just noticed this in the CI. We should review those. 

##[error]Target "RunManagedUnitTests" has thrown an exception
12:38:08 [ERR] Target RunManagedUnitTests has thrown an exception
System.IO.IOException: There is not enough space on the disk. : 'D:\a\_work\1\s\artifacts\build_data\dumps\testhost.exe_251209_123600.dmp'.
   at System.IO.FileSystem.CopyFile(String sourceFullPath, String destFullPath, Boolean overwrite)
   at Nuke.Common.IO.FileSystemTasks.CopyFile(String source, String target, FileExistsPolicy policy, Boolean

tracer/test/Datadog.Trace.Tests/Util/MutexHelper.cs
```diff
@@ -0,0 +1,54 @@
...
+    {
+        var result = mutex.Wait(timeout);
+        if (!result)
+        {
+            output?.WriteLine($"Mutex timed out in {caller}. Capturing memory dump");
+            var process = Process.GetCurrentProcess();
+            var log = MemoryDumpHelper.CaptureMemoryDump(process)
+                          ? "Successfully captured memory dump"
+                          : "Failed to capture memory dump";
+            output?.WriteLine(log);
```
NachoEchevarria:
```suggestion
          try
          {
              var process = Process.GetCurrentProcess();
              var log = MemoryDumpHelper.CaptureMemoryDump(process)
                            ? "Successfully captured memory dump"
                            : "Failed to capture memory dump";
              output?.WriteLine(log);
          }
          catch (Exception ex)
          {
              output?.WriteLine($"Exception while capturing memory dump: {ex}");
          }
```
We should probably add a try/catch to avoid exceptions from masking test errors.

andrewlock:
> Just noticed this in the CI. We should review those.

Yeah, to be honest, we may not need to merge this PR at all. We're looking at those dumps and tl;dr; there's thread exhaustion as best as I can tell. The real question is _why_ 😬

tracer/test/Datadog.Trace.Tests/Logging/DirectSubmission/Sink/DatadogSinkTests.cs
andrewlock: So the thing is, we kinda want to test that you _don't_ have to call `FlushAsync` in these APIs I think 😕

tracer/test/Datadog.Trace.Tests/Util/MutexHelper.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// <param name="caller">The caller, populated automatically</param>
+    public static bool WaitOrDump(this ManualResetEventSlim mutex, TimeSpan timeout, ITestOutputHelper output = null, [CallerMemberName] string caller = null)
+        => DoWaitOrDump(mutex, timeout, output, caller);
+
+    private static bool DoWaitOrDump(ManualResetEventSlim mutex, TimeSpan timeout, ITestOutputHelper output, string caller)
+    {
+        var result = mutex.Wait(timeout);
+        if (!result)
+        {
+            output?.WriteLine($"Mutex timed out in {caller}. Capturing memory dump");
```
NachoEchevarria:
```suggestion
            output?.WriteLine($"Mutex timed out in {caller} after {timeout.TotalSeconds:F1}s. Capturing memory dump");
```
NIT

PR #7930, opened 2025-12-09
Fix HotChocolate duck typing error

NachoEchevarria:
## Summary of changes

Fixes a `DuckTypePropertyOrFieldNotFoundException` that occurs when HotChocolate GraphQL errors contain array index path segments instead of field name path segments.

## Reason for change

We have seen [the following error](https://app.datadoghq.com/error-tracking?query=service%3Ainstrumentation-telemetry-data%20%40lib_language%3Adotnet%20version%3A3.3%2A%20-Bson%20-unhandled%20%2Achocolate%2A&et-side=data&et-viz=events&fromUser=false&order=total_count&refresh_mode=sliding&source=all&sp=%5B%7B%22p%22%3A%7B%22issueId%22%3A%22f0cf1cd4-56b0-11f0-b7cb-da7ad0900002%22%7D%2C%22i%22%3A%22error-tracking-issue%22%7D%5D&from_ts=1764667683470&to_ts=1765272483470&live=true):

```
Error : Error creating GraphQL error SpanEvent.
Datadog.Trace.DuckTyping.DuckTypePropertyOrFieldNotFoundException
at Datadog.Trace.DuckTyping.DuckTypePropertyOrFieldNotFoundException.Throw(String name, String duckAttributeName, Type type)
at Datadog.Trace.DuckTyping.DuckType.CreatePropertiesFromStruct(TypeBuilder proxyTypeBuilder, Type proxyDefinitionType, Type targetType, FieldInfo instanceField)
at Datadog.Trace.DuckTyping.DuckType.CreateProxyType(Type proxyDefinitionType, Type targetType, Boolean dryRun)
at HotChocolate_Abstractions__.HotChocolate_Error.Datadog_Trace_ClrProfiler_AutoInstrumentation_GraphQL_HotChocolate_IError_44.get_Path()
at Datadog.Trace.ClrProfiler.AutoInstrumentation.GraphQL.HotChocolate.HotChocolateCommon.ConstructErrorEvents(List`1 executionErrors)
```

This error was detected in customer installations.

## Implementation details

 **Root cause:** HotChocolate's `Path` class is abstract with multiple implementations:
   - `NamePathSegment` - for field names (has a `Name` property)
   - `IndexerPathSegment` - for array indices (does NOT have a `Name` property)

   The previous implementation used duck typing with `PathStruct` that expected a `Name` property, which failed when encountering `IndexerPathSegment` instances.

   **Solution:**
   - Removed `PathStruct` duck type that assumed all path segments have a `Name` property
   - Created new `IPath` interface that duck types the `ToList()` method available on the abstract `Path` class
   - Changed `IError.Path` from `PathStruct` to `IPath`
   - Updated `HotChocolateCommon.ConstructErrorEvents()` to:
     1. Duck cast the path object to `IPath`
     2. Call `ToList()` to get all path segments (works for both `NamePathSegment` and `IndexerPathSegment`)
     3. Convert each segment to string for the span event attributes

   This approach is more robust as it relies on the public API (`ToList()`) defined on the abstract class rather than assuming all
   implementations share the same properties.

## Test coverage

Added new test case `ThrowExceptionIndex` to the existing HotChocolate integration tests:
   - New GraphQL query that triggers an error with array index path segments
   - `ErrorMiddleware` modifies the error path to include `["books", 1]` to create an `IndexerPathSegment`
   - Updated test snapshots across all schema versions to verify proper path serialization
   - Added regex scrubber for the new test method's stack traces

   The test ensures errors with `IndexerPathSegment` paths are properly captured and serialized in span events without throwing duck typing exceptions.

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/GraphQL/HotChocolate/HotChocolateCommon.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/GraphQL/HotChocolate/HotChocolateCommon.cs
@@ -198,10 +198,10 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.GraphQL.HotChocolate
                         eventAttributes.Add(new KeyValuePair<string, object>("locations", joinedLocations.ToArray()));
                     }
 
-                    var pathObject = executionError.Path;
-                    if (pathObject != null && pathObject.TryDuckCast<IPath>(out var pathProxy))
+                    var path = executionError.Path;
+                    if (path != null)
                     {
-                        var pathList = pathProxy.ToList();
+                        var pathList = path.ToList();
                         if (pathList is { Count: > 0 })
                         {
                             var pathArray = new string[pathList.Count];
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/GraphQL/HotChocolate/IError.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/GraphQL/HotChocolate/IError.cs
@@ -33,7 +33,7 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.GraphQL.HotChocolate
         /// Gets the path in the document where the error applies.
         /// Returns the HotChocolate.Path object which has a ToList() method.
         /// </summary>
-        object Path { get; }
+        IPath Path { get; }
 
         /// <summary>
         /// Gets the StackTrace of the error.


tracer/test/snapshots/HotChocolateTests.SubmitsTraces.SchemaV0.Pre_13_1_0.verified.txt
```diff
@@ -320,6 +350,49 @@ errors: [
       graphql.source:
 query ErrorQuery {
   throwException
+},
```
NachoEchevarria, line 353: This is the error thrown for indexes that previously would cause the exception.

bouwkast (approved):

NachoEchevarria:
Thanks for your reviews!

PR #7931, opened 2025-12-09
[Config Registry] 5/7 Replace manual ConfigurationKeys by generated ones in the whole solution

anna-git:
(This PR is actually https://github.com/DataDog/dd-trace-dotnet/pull/7697 that accidentally got closed)

## Context

Part of **Configuration Inversion (Step 5)** - Stack progress:
1. [#7548](https://github.com/DataDog/dd-trace-dotnet/pull/7548) - Add GitLab step and JSON configuration file
2. [#7688](https://github.com/DataDog/dd-trace-dotnet/pull/7688) - Cleanup configuration / platform keys + source generator
3. [#7698](https://github.com/DataDog/dd-trace-dotnet/pull/7698) - Aliases handling via source generator
4. [#7689](https://github.com/DataDog/dd-trace-dotnet/pull/7689) - Analyzers for platform and ConfigurationBuilder
5. **→ [#7931](https://github.com/DataDog/dd-trace-dotnet/pull/7697) - Replace manual ConfigurationKeys by generated ones in the whole solution (this PR)**
6. [#7932](https://github.com/DataDog/dd-trace-dotnet/pull/7932) - Forbid use of System.Environment methods and adapt everywhere 
7. [#7937](https://github.com/DataDog/dd-trace-dotnet/pull/7937) - Integration names to generated keys

## Summary of changes

Fixed the [ConfigurationKeysGenerator](cci:2://file:///Users/anna.yafi/go/src/github.com/DataDog/dd-trace-dotnet3/tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigurationKeysGenerator.cs:21:0-958:1) to properly read and apply the [configuration_keys_mapping.json](cci:7://file:///Users/anna.yafi/go/src/github.com/DataDog/dd-trace-dotnet3/tracer/src/Datadog.Trace/Configuration/configuration_keys_mapping.json:0:0-0:0) file, and extracted common file header comments to a reusable constant.

**Key changes:**
- Fixed JSON array extraction in [ParseMappingFile](cci:1://file:///Users/anna.yafi/go/src/github.com/DataDog/dd-trace-dotnet3/tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigurationKeysGenerator.cs:350:4-524:5) method - the generator was incorrectly trying to extract the `"mappings"` field as an object instead of an array
- Extracted configuration generator comments to `Constants.ConfigurationGeneratorComment` for reuse across multiple generators
[...]

PR #7932, opened 2025-12-09
[Config Registry] 6/8 Forbid use of System.Environment methods and adapt everywhere

anna-git:
## Context

Part of **Configuration Inversion (Step 6)** - Stack progress:
1.  [#7548](https://github.com/DataDog/dd-trace-dotnet/pull/7548) - Add GitLab step and JSON configuration file
2. [#7688](https://github.com/DataDog/dd-trace-dotnet/pull/7688) - Cleanup configuration / platform keys + source generator
3. [#7698](https://github.com/DataDog/dd-trace-dotnet/pull/7698) - Aliases handling via source generator
4. [#7689](https://github.com/DataDog/dd-trace-dotnet/pull/7689) - Analyzers for platform and ConfigurationBuilder
6. **→ [#7932](https://github.com/DataDog/dd-trace-dotnet/pull/7932) - Forbid use of System.Environment methods and adapt everywhere (this PR)**
7. [#7937](https://github.com/DataDog/dd-trace-dotnet/pull/7937) - Integration names to generated keys

I'll update the PR summary to mention the YAML documentation file:

## Summary of changes

Banned direct `System.Environment.GetEnvironmentVariable()` usage and migrated all environment variable access to use `EnvironmentHelpers` with strongly-typed [ConfigurationKeys](cci:2://file:///Users/anna.yafi/go/src/github.com/DataDog/dd-trace-dotnet3/tracer/src/Datadog.Trace/Generated/net461/Datadog.Trace.SourceGenerators/ConfigurationKeysGenerator/ConfigurationKeys.g.cs:15:0-737:1) and `PlatformKeys` constants.

### Key changes:
- **Banned `System.Environment.GetEnvironmentVariable()`** via `BannedApiAnalyzers`
- **Added `EnvironmentGetEnvironmentVariableAnalyzer` (DD0009)** to enforce [ConfigurationKeys](cci:2://file:///Users/anna.yafi/go/src/github.com/DataDog/dd-trace-dotnet3/tracer/src/Datadog.Trace/Generated/net461/Datadog.Trace.SourceGenerators/ConfigurationKeysGenerator/ConfigurationKeys.g.cs:15:0-737:1)/`PlatformKeys` usage only
- **Nested `PlatformKeys` by category** (Ci, Aws, AzureAppService, ServiceFabric, DotNet)
- **Migrated ~50+ files** across CI Visibility, AWS Lambda, AppSec, Telemetry, Profiler, and Agent components
- **Added missing keys** to [supported-configurations.json](cci:7://file:///Users/anna.yafi/go/src/github.com/DataDog/dd-trace-dotnet3/tracer/src/Datadog.Trace/Configuration/supported-configurations.json:0:0-0:0) and their documentation to [supported-configurations-docs.yaml](cci:7://file:///Users/anna.yafi/go/src/github.com/DataDog/dd-trace-dotnet3/tracer/src/Datadog.Trace/Configuration/supported-configurations-docs.yaml:0:0-0:0)

## Reason for change

Direct `System.Environment.GetEnvironmentVariable()` calls with string literals are error-prone. Centralizing through `EnvironmentHelpers` with strongly-typed constants provides compile-time validation, discoverability, and refactoring safety.

## Implementation details

1. **Banned API enforcement** - Added `BannedSymbols.txt` and configured `.editorconfig` to treat RS0030 as error (vendored code excluded)
2. **Custom analyzer** - DD0009 validates all `EnvironmentHelpers` calls accept only [ConfigurationKeys](cci:2://file:///Users/anna.yafi/go/src/github.com/DataDog/dd-trace-dotnet3/tracer/src/Datadog.Trace/Generated/net461/Datadog.Trace.SourceGenerators/ConfigurationKeysGenerator/ConfigurationKeys.g.cs:15:0-737:1)/`PlatformKeys` constants, rejecting hardcoded strings
3. **PlatformKeys organization** - Nested by category for better discoverability
4. **EnvironmentHelpers refactoring** - Added overloads for nested class constants, maintained backward compatibility
5. **Configuration documentation** - Added missing configuration keys to [supported-configurations.json](cci:7://file:///Users/anna.yafi/go/src/github.com/DataDog/dd-trace-dotnet3/tracer/src/Datadog.Trace/Configuration/supported-configurations.json:0:0-0:0) and comprehensive XML documentation to [supported-configurations-docs.yaml](cci:7://file:///Users/anna.yafi/go/src/github.com/DataDog/dd-trace-dotnet3/tracer/src/Datadog.Trace/Configuration/supported-configurations-docs.yaml:0:0-0:0), which the source generator uses to generate XML doc comments in [ConfigurationKeys](cci:2://file:///Users/anna.yafi/go/src/github.com/DataDog/dd-trace-dotnet3/tracer/src/Datadog.Trace/Generated/net461/Datadog.Trace.SourceGenerators/ConfigurationKeysGenerator/ConfigurationKeys.g.cs:15:0-737:1) classes

## Test coverage

- ✅ Comprehensive analyzer tests covering valid/invalid scenarios
- ✅ All existing unit tests pass
- ✅ Banned API analyzer correctly flags direct `System.Environment` usage

## Other details

+2,600/-940 lines. No breaking changes, negligible performance impact.


tracer/src/Datadog.Trace.Tools.Analyzers/ConfigurationAnalyzers/BannedSymbols.txt
```diff
@@ -0,0 +1,9 @@
+# Banned APIs for Datadog.Trace
+# This file defines APIs that should not be used in the Datadog.Trace project.
+# Format: Full type name with method signature; Optional message
+
+# Ban System.Environment.GetEnvironmentVariable - use EnvironmentHelpers instead
+M:System.Environment.GetEnvironmentVariable(System.String);Use EnvironmentHelpers.GetEnvironmentVariable instead to ensure proper configuration handling
+M:System.Environment.GetEnvironmentVariable(System.String,System.EnvironmentVariableTarget);Use EnvironmentHelpers.GetEnvironmentVariable instead to ensure proper configuration handling
+M:System.Environment.GetEnvironmentVariables();Avoid using this alltogether and get precise env vars through 
+ConfigurationBuilder
```
andrewlock:
```suggestion
M:System.Environment.GetEnvironmentVariables();Avoid using this alltogether and get precise env vars through ConfigurationBuilder
```

tracer/src/Datadog.Trace.Tools.Analyzers/ConfigurationAnalyzers/EnvironmentGetEnvironmentVariableAnalyzer.cs
```diff
@@ -0,0 +1,255 @@
...
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+            context.EnableConcurrentExecution();
+            context.RegisterSyntaxNodeAction(AnalyzeInvocationExpression, SyntaxKind.InvocationExpression);
+        }
+
+        private static void AnalyzeInvocationExpression(SyntaxNodeAnalysisContext context)
+        {
+            var invocation = (InvocationExpressionSyntax)context.Node;
+
+            // Get the method being called
```
andrewlock: nit: getting the semantic model is _comparatively_ expensive, and this is going to get called for every single node in the compilation. We should do everything we can to bail out asap, as cheaply as possible.

In particular, I would hardcode some initial name checks which will act as a good filter.  You _could_ also check based on argument lengths etc. Basically, anything you can reliably add to avoid using the SemanticModel is likely going to be a net win:


```suggestion
        if (invocation.Expression is not MemberAccessExpressionSyntax memberAccess
            || (memberAccess.Name.Identifier.Text is not "GetEnvironmentVariable" or "EnvironmentVariableExists" or "TryCheckEnvVar" or "GetValue")
        {
            return;
        }

            // Get the method being called
```

tracer/src/Datadog.Trace.Tools.Analyzers/ConfigurationAnalyzers/EnvironmentGetEnvironmentVariableAnalyzer.cs
```diff
@@ -0,0 +1,255 @@
...
+        }
+
+        /// <summary>
+        /// Checks if a method is one of our target methods to analyze.
+        /// </summary>
+        private static bool IsTargetMethod(IMethodSymbol method)
+        {
+            var methodName = method.Name;
+
+            foreach (var (targetType, targetMethod) in TargetMethods)
```
andrewlock: I would unroll this and hard-code this loop as much as you can, as it's going to be relatively expensive

tracer/src/Datadog.Trace.Tools.Analyzers/ConfigurationAnalyzers/EnvironmentGetEnvironmentVariableAnalyzer.cs
```diff
@@ -0,0 +1,255 @@
...
+            {
+                if (methodName != targetMethod)
+                {
+                    continue;
+                }
+
+                // Check if the containing type matches directly
+                var containingType = method.ContainingType?.ToDisplayString();
```
andrewlock: Calling `ToDisplayString()` allocates, and is relatively expensive. Ideally we should "find" the symbols and pass them through ourselves. e.g. you can see and example of that here: https://github.com/andrewlock/NetEscapades.EnumGenerators/blob/5d92a2b4a62a4b13f90b29ca8fa52ef8ac000dc6/src/NetEscapades.EnumGenerators/Diagnostics/UsageAnalyzers/GetNamesAnalyzer.cs#L32

tracer/src/Datadog.Trace.Tools.Analyzers/Datadog.Trace.Tools.Analyzers.csproj
```diff
@@ -10,6 +10,8 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="3.11.0" PrivateAssets="all" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="5.0.0" PrivateAssets="all" />
+    <!-- BannedApiAnalyzers needs to flow to consuming projects, so don't mark as PrivateAssets -->
+    <PackageReference Include="Microsoft.CodeAnalysis.BannedApiAnalyzers" Version="3.3.4" />
```
andrewlock: _Should_ this be in the analyzers project? 🤔 I think we should reference it in the Directory.Build.props file instead?

tracer/src/Datadog.Trace.Tools.Analyzers/ConfigurationAnalyzers/EnvironmentGetEnvironmentVariableAnalyzer.cs
```diff
@@ -0,0 +1,255 @@
...
+                if (!field.IsConst)
+                {
+                    return false;
+                }
+
+                // Check if it's from ConfigurationKeys, PlatformKeys, or any nested class within them
+                var containingType = field.ContainingType;
+                while (containingType != null)
+                {
+                    var fullTypeName = containingType.ToDisplayString();
```
andrewlock: Again, this is kind of expensive, if we're using symbols, we should probably find the symbols ahead of time and use default symbol comparison instead

tracer/src/Datadog.Trace.Tools.Runner/RunCiCommand.cs
```diff
@@ -54,8 +54,8 @@ private async Task ExecuteAsync(InvocationContext context)
             // Propagate original test command and working directory
             if (initResults.ProfilerEnvironmentVariables is { } profilerEnvironmentVariables)
             {
-                profilerEnvironmentVariables[TestSuiteVisibilityTags.TestSessionCommandEnvironmentVariable] = Environment.CommandLine;
-                profilerEnvironmentVariables[TestSuiteVisibilityTags.TestSessionWorkingDirectoryEnvironmentVariable] = Environment.CurrentDirectory;
+                profilerEnvironmentVariables[ConfigurationKeys.CIVisibility.TestSessionCommand] = Environment.CommandLine;
+                profilerEnvironmentVariables[ConfigurationKeys.CIVisibility.TestSessionWorkingDirectory] = Environment.CurrentDirectory;
```
andrewlock, line 58: nit: You might struggle less with merge conflicts and getting reviews if you extract all the basic renaming out into a separate PR 🙂

tracer/src/Datadog.Trace/AppSec/SecuritySettings.cs
```diff
@@ -153,31 +154,38 @@ public SecuritySettings(IConfigurationSource? source, IConfigurationTelemetry te
...
             ScaEnabled = config
-                             .WithKeys(ConfigurationKeys.AppSec.ScaEnabled)
-                             .AsBool();
+                        .WithKeys(ConfigurationKeys.AppSec.ScaEnabled)
+                        .AsBool();
 
             NoCustomLocalRules = Rules == null;
+
+            DdDotnetTracerHome = config.WithKeys(ConfigurationKeys.DotnetTracerHome).AsString();
+            InternalTraceNativeEnginePath = config.WithKeys(ConfigurationKeys.InternalTraceNativeEnginePath).AsString();
```
andrewlock: These are reading from the full config stack, but they shouldn't, they should specifically be read only from env vars I think? 🤔 

```suggestion
            var envConfig = new ConfigurationBuilder(new EnvironmentConfigurationSource(), telemetry);
            DdDotnetTracerHome = envConfig .WithKeys(ConfigurationKeys.DotnetTracerHome).AsString();
            InternalTraceNativeEnginePath = envConfig .WithKeys(ConfigurationKeys.InternalTraceNativeEnginePath).AsString();
```

There are likely a couple of other "non-settings" that should similarly be read with this pattern, but that's a separate issue 😄

tracer/src/Datadog.Trace/Util/EnvironmentHelpersNoLogging.cs
```diff
@@ -23,30 +22,32 @@ internal static bool IsServerlessEnvironment(out Exception? exceptionInReading)
...
+                        || (TryCheckEnvVar(PlatformKeys.GcpFunction.FunctionNameKey, ref firstException)
+                         && TryCheckEnvVar(PlatformKeys.GcpFunction.FunctionTargetKey, ref firstException))
+                        || (TryCheckEnvVar(PlatformKeys.GcpFunction.DeprecatedFunctionNameKey, ref firstException)
+                         && TryCheckEnvVar(PlatformKeys.GcpFunction.DeprecatedProjectKey, ref firstException));
         exceptionInReading = firstException;
         return isServerless;
+    }
 
-        static bool CheckEnvVar(string key, ref Exception? storedException)
+    private static bool TryCheckEnvVar(string key, ref Exception? storedException)
```
andrewlock: nit: we essentially have the same method in `EnvironmentHelpers` called `EnvironmentVariableExists` which IMO is a better name for it - can we use that here too?

tracer/src/Datadog.Trace/Util/EnvironmentHelpersNoLogging.cs
```diff
@@ -61,4 +62,12 @@ public static bool IsClrProfilerAttachedSafe()
             return false;
         }
     }
+
+    public static string? InjectionEnabled() => GetEnvironmentVariable(ConfigurationKeys.SsiDeployed);
```
andrewlock: I think we should change the name of this method. "injection enabled" could mean a _lot_ of things 😄 (and yes, `SsiDeployed` checks the variable `DD_INJECTION_ENABLED`, but the env var is terribly named, so `SsiDeployed` is a better name here too IMO). Or maybe `SsiInjectionEnabledVariable` ? Or something like that 😅

tracer/src/Datadog.Trace/Util/EnvironmentHelpersNoLogging.cs
```diff
@@ -61,4 +62,12 @@ public static bool IsClrProfilerAttachedSafe()
...
+
+    public static string? ProgramData() => GetEnvironmentVariable(PlatformKeys.ProgramData);
```
andrewlock, line 68: Do we use this in more than one place? If not, I would probably just inline it directly

tracer/src/Datadog.Trace/Util/EnvironmentHelpersNoLogging.cs
```diff
@@ -61,4 +62,12 @@ public static bool IsClrProfilerAttachedSafe()
...
+
+#pragma warning disable RS0030
+    private static string? GetEnvironmentVariable(string key) => Environment.GetEnvironmentVariable(key);
```
andrewlock, line 74: It probably _will_ be inlined, but seeing as we (presumably) only extracted it to simplify the `#pragma` I would probably do this:
```suggestion
    [MethodImpl(MethodImplOptions.AggressiveInlining)]
    private static string? GetEnvironmentVariable(string key) => Environment.GetEnvironmentVariable(key);
```

Or alternatively, just wrap the pragma around the whole file, and inline the method completely

.editorconfig
```diff
@@ -82,3 +82,17 @@ csharp_new_line_before_members_in_anonymous_types = true
...
+# Error when static analyzers aren't respected
+dotnet_diagnostic.RS0030.severity = error
+
+## Exceptions for banned apis:
+
+# Disables static analyzer errors for a specific folder
+[/tracer/src/Datadog.Trace/Vendors/**.*]
+dotnet_diagnostic.RS0030.severity = none
+
+# Allow Environment.GetEnvironmentVariable in EnvironmentConfigurationSource (it's the *one* place that needs it)
```
andrewlock: well... we have some other pragma disablements right? 😄 Should we document them here too? Or is it preferable to do that in source (I think the latter probably, just wondering!)

tracer/src/Datadog.Trace.Tools.Analyzers/ConfigurationAnalyzers/EnvironmentGetEnvironmentVariableAnalyzer.cs
anna-git: ah ok interesting thanks. I was wondering if creating an attribute, like a "marker" for this methods and just checking its presence would be preferable in the end? but maybe there's other issues with this

tracer/src/Datadog.Trace/Util/EnvironmentHelpersNoLogging.cs
anna-git: the issue is it's used in `DatadogLoggingFactory` as well 🤔

tracer/src/Datadog.Trace/Util/EnvironmentHelpersNoLogging.cs
anna-git: ah right, changed in 0bf577043dd9cc64429bb8b1d374f2c24a0b20b6

tracer/src/Datadog.Trace/Util/EnvironmentHelpersNoLogging.cs
anna-git: changed in 756c6b316619db20d993beb1109af17e9e4074b1

tracer/src/Datadog.Trace/Util/EnvironmentHelpersNoLogging.cs
anna-git: I added the inlining attribute in 756c6b316619db20d993beb1109af17e9e4074b1
Didn't want to add the pragma on the whole file as some other not allowed methods could potentially be introduced in it..

tracer/src/Datadog.Trace.Tools.Analyzers/ConfigurationAnalyzers/EnvironmentGetEnvironmentVariableAnalyzer.cs
anna-git: changed in 3796c4adb82bcb2c1881eefbc6e096d63183fe60

tracer/src/Datadog.Trace.Tools.Analyzers/ConfigurationAnalyzers/EnvironmentGetEnvironmentVariableAnalyzer.cs
anna-git: changed here https://github.com/DataDog/dd-trace-dotnet/pull/7932/changes/3796c4adb82bcb2c1881eefbc6e096d63183fe60#diff-5fcad5994e9d4e7836159e31d144880e90753a1c96ed2caa417cd4ca6d572c52R141

tracer/src/Datadog.Trace.Tools.Analyzers/ConfigurationAnalyzers/EnvironmentGetEnvironmentVariableAnalyzer.cs
anna-git: thanks very much for this comment, I changed it here 79b79c223e79200448999fc1701358afbadb60e4

tracer/src/Datadog.Trace.Tools.Analyzers/Datadog.Trace.Tools.Analyzers.csproj
anna-git: Ok I referenced it in Directory.Build.props  e541949032b4c237d83877f2d266fd25ffed8a12 but for now, only Datadog.Trace references it as additional file, as otherwise, warning pops for the loader as well and I didn't want to address them just yet..

tracer/src/Datadog.Trace.Tools.Analyzers/ConfigurationAnalyzers/EnvironmentGetEnvironmentVariableAnalyzer.cs
anna-git: thanks fixed in 3796c4adb82bcb2c1881eefbc6e096d63183fe60

tracer/src/Datadog.Trace.Tools.Runner/RunCiCommand.cs
anna-git: yeah 🫣... I tried having separate commits per team.. hopefully I can manage to get reviews! but probably next time I'll split this finer in smaller PR..

tracer/src/Datadog.Trace.Tools.Analyzers/ConfigurationAnalyzers/BannedSymbols.txt
anna-git: thanks fixed in ba5f8a79dfb7fd28fa5e6081977023745ad610a9

.editorconfig
anna-git: I changed it here ba5f8a79dfb7fd28fa5e6081977023745ad610a9
I replaced that by a pragma in source directly so that we can control which line we want to allow..

tonyredondo (approved):

tracer/src/Datadog.Trace.Tools.Runner/RunCiCommand.cs
tonyredondo: These weren't added as a configuration keys because they are not configuration knobs, these are part of a context propagation mechanism

tonyredondo: Test optimization uses the environment variables to do context propagation from the runner to the actual testhost process

tracer/src/Datadog.Trace/Ci/CiEnvironment/PlatformKeys.Ci.cs
```diff
@@ -0,0 +1,342 @@
...
+
+#nullable enable
+namespace Datadog.Trace.Configuration;
+
+internal static partial class PlatformKeys
+{
+    /// <summary>
+    /// Platform keys for CI/CD environment variables from external CI providers
+    /// </summary>
+    internal static class Ci
```
tonyredondo, line 14: I really don't know if I like this to be honest. I cannot see the wins moving the constants to this place. I really liked the total separation of the `CI` namespace, now I have CI over the place. If this has to be done, why not `Datadog.Trace.Ci.Configuration` instead?

tracer/src/Datadog.Trace.Tools.Runner/RunCiCommand.cs
anna-git: I see..  the issue is that everything starting with `DD_` and created by us is now considered a "configuration" key and has to be controlled [by config registry.](https://feature-parity.us1.prod.dog/configurations?viewType=configurations&configSearch=DD_TESTSESSION_WORKINGDIRECTORY).  and the mechanism here to control them is to have them in the class `ConfigurationKeys` automatically generated from config registry files 😅 
I guess configuration here is more to be understood in the broader sense of the term..

tracer/src/Datadog.Trace/Ci/CiEnvironment/PlatformKeys.Ci.cs
anna-git: As these are used to collect env vars, we want to control them as well and the [analyzers](https://github.com/DataDog/dd-trace-dotnet/blob/2a397eecd7d874a3df615067797884832a07d55f/tracer/src/Datadog.Trace.Tools.Analyzers/ConfigurationAnalyzers/EnvironmentGetEnvironmentVariableAnalyzer.cs#L35) make sure it's either coming from the generated `ConfigurationKeys` or the controlled `PlatformKeys` (not starting with DD_..) 
If I add them in a different class, then I'd have to change the 2 analyzers to add this control as well (allow an extra class with `ConfigurationBuilder` and `EnvironmentHelpers` and control this class doesnt have consts starting with the forbidden prefixes.. which complexifies a lot, when it obeys the same idea of either you query a DD created env var and it's generated in `ConfigurationKeys`, either you query an existing env var related to a platform or framework and we want to control  it indeed doesnt  start with OTEL", "DD_", "_DD_", "DATADOG_"..
On the plus side it's within the Ci static class, and maybe we could move it to the `Datadog.Trace.Ci` folder without changing the namespace? but at least it would remain all in the Ci folder?

tracer/src/Datadog.Trace.BenchmarkDotNet/DatadogDiagnoser.cs
```diff
@@ -129,7 +129,7 @@ public void DisplayResults(ILogger logger)
     private static IEnumerable<string?> GetHomeFolderPaths(DiagnoserActionParameters parameters)
     {
         // try to locate it from the environment variable
-        yield return EnvironmentHelpers.GetEnvironmentVariable("DD_DOTNET_TRACER_HOME");
+        yield return EnvironmentHelpers.GetEnvironmentVariable(ConfigurationKeys.DotnetTracerHome);
```
lucaspimentel: It looks like we use `Dotnet` (with lower case `n`) in some places and `DotNet` in others. Not super important, but it would be nice to be consistent.

tracer/src/Datadog.Trace/Configuration/PlatformKeys.AzureAppService.cs
```diff
@@ -51,5 +51,20 @@ internal static class AzureAppService
         /// or will have a value of "dynamic".
         /// </summary>
         public const string WebsiteSku = "WEBSITE_SKU";
+
+        /// <summary>
+        /// The name of the Azure Container App instance.
+        /// </summary>
+        internal const string ContainerAppName = "CONTAINER_APP_NAME";
```
lucaspimentel, line 53: This file is for Azure App Services, but Container Apps are separate (not AAS). I think it's fine to merge like this and we can clean it up later.

tracer/src/Datadog.Trace/Configuration/PlatformKeys.Aws.cs
```diff
@@ -10,5 +11,6 @@ internal static partial class PlatformKeys
     internal static class Aws
     {
         public const string FunctionName = "AWS_LAMBDA_FUNCTION_NAME";
+        public const string Handler = "_HANDLER";
```
lucaspimentel: It would be nice if we made it more clear that `FunctionName` and `Handler` are specifically for AWS Lambda (the filename and namespace only say `Aws`, which includes many other products aside from Lambda). We could either rename them to `LambdaFunctionName` and `LambdaHandler`, or add a `...PlatformKeys.Aws.Lambda` namespace. But I think it's fine to merge like this and we can clean it up later.

tracer/src/Datadog.Trace/PlatformHelpers/HostMetadata.cs
```diff
@@ -140,7 +141,7 @@ private static void TryGetKernelInformation(out string? kernel, out string? kern
                     return host;
                 }
 
-                return EnvironmentHelpers.GetEnvironmentVariable("COMPUTERNAME");
+                return EnvironmentHelpers.GetEnvironmentVariable(PlatformKeys.AzureAppService.InstanceNameKey);
```
lucaspimentel: `COMPUTERNAME` is a standard Windows env var, not specific to Azure App Service.

tracer/src/Datadog.Trace/Configuration/PlatformKeys.Aws.cs
anna-git: good point,  changed in [be89b22](https://github.com/DataDog/dd-trace-dotnet/pull/7932/commits/be89b2250e5601d3d5ca7e8a68295aad13942ada)

tracer/src/Datadog.Trace/PlatformHelpers/HostMetadata.cs
anna-git: good point,  changed in [be89b22](https://github.com/DataDog/dd-trace-dotnet/pull/7932/commits/be89b2250e5601d3d5ca7e8a68295aad13942ada)

tracer/src/Datadog.Trace/Configuration/PlatformKeys.AzureAppService.cs
anna-git: good point,  changed in [be89b22](https://github.com/DataDog/dd-trace-dotnet/pull/7932/commits/be89b2250e5601d3d5ca7e8a68295aad13942ada)

tracer/src/Datadog.Trace.BenchmarkDotNet/DatadogDiagnoser.cs
anna-git: good point,  changed in 84613129ea08b3fb8d9f740b90f96190935f04e0

tracer/src/Datadog.Trace/AppSec/SecuritySettings.cs
anna-git: I forgot this comment 🙈 good catch, changed in [be89b22](https://github.com/DataDog/dd-trace-dotnet/pull/7932/commits/be89b2250e5601d3d5ca7e8a68295aad13942ada)
I was wondering for a sec if it should be in all configs ... as stable config wants to now control all configs..  but I guess it's read too early in the app to even have these settings..

tracer/src/Datadog.Trace/ClrProfiler/ServerlessInstrumentation/LambdaMetadata.cs
```diff
@@ -39,11 +37,11 @@ private LambdaMetadata(bool isRunningInLambda, string functionName, string handl
 
     public static LambdaMetadata Create(string extensionPath = ExtensionFullPath)
     {
-        var functionName = EnvironmentHelpers.GetEnvironmentVariable(FunctionNameEnvVar);
+        var functionName = EnvironmentHelpers.GetEnvironmentVariable(PlatformKeys.Aws.LambdaFunctionName);
 
         var isRunningInLambda = !string.IsNullOrEmpty(functionName)
                              && File.Exists(
-                                    EnvironmentHelpers.GetEnvironmentVariable(ExtensionPathEnvVar)
+                                    EnvironmentHelpers.GetEnvironmentVariable(ConfigurationKeys.Aws.ExtensionPath)
```
lucaspimentel: Sorry, I missed it earlier, but `ExtensionPath` is also specific to Lambda 😅
Can we also rename it to `LambdaExtensionPath` like the others?

tracer/src/Datadog.Trace/Configuration/PlatformKeys.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Datadog.Trace.Configuration;
+
+internal static partial class PlatformKeys
+{
+    /// <summary>
+    /// Built-in Windows environment variable that holds the system's unique name
+    /// Also, the instance name in Azure where the traced application is running.
+    /// </summary>
+    internal const string InstanceNameKey = "COMPUTERNAME";
```
lucaspimentel: We should probably call the field `ComputerNameKey`. The "instance name" terminology is specific to Azure App Services, but `COMPUTERNAME` exists in every Windows host.

tracer/src/Datadog.Trace/Configuration/PlatformKeys.cs
```diff
@@ -0,0 +1,17 @@
...
+
+#nullable enable
+
+namespace Datadog.Trace.Configuration;
+
+internal static partial class PlatformKeys
+{
+    /// <summary>
+    /// Built-in Windows environment variable that holds the system's unique name
+    /// Also, the instance name in Azure where the traced application is running.
```
lucaspimentel:
```suggestion
    /// Built-in Windows environment variable that holds the system's unique name (hostname).
    /// Also, the instance name in Azure App Services where the instrumented application is running.
```

anna-git: Fixed in de535249db936260d4c8fcb7377b9f13ab32d3b6

tracer/src/Datadog.Trace/Configuration/PlatformKeys.cs
anna-git: Fixed in de535249db936260d4c8fcb7377b9f13ab32d3b6

tracer/src/Datadog.Trace/ClrProfiler/ServerlessInstrumentation/LambdaMetadata.cs
anna-git: thanks, fixed in de535249db936260d4c8fcb7377b9f13ab32d3b6

tracer/src/Datadog.Trace.Tools.Analyzers/ConfigurationAnalyzers/ConfigurationBuilderWithKeysAnalyzer.cs
```diff
@@ -56,70 +57,55 @@ public override void Initialize(AnalysisContext context)
...
+                    wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.PlatformKeys));
+
+                compilationContext.RegisterSyntaxNodeAction(
+                    c => AnalyzeInvocationExpression(c, targetTypes),
+                    SyntaxKind.InvocationExpression);
+            });
         }
 
-        private static void AnalyzeInvocationExpression(SyntaxNodeAnalysisContext context)
+        private static void AnalyzeInvocationExpression(SyntaxNodeAnalysisContext context, TargetTypeSymbols targetTypes)
```
andrewlock: Just FYI, _in general_, if we're passing around "big" `readonly struct`, we typically put `in` on both the caller and the target method, as it can avoid copying the struct so boof perf a bit. Don't know that it's necessary here, just an FYI 🙂

tracer/src/Datadog.Trace.Tools.Analyzers/ConfigurationAnalyzers/ConfigurationBuilderWithKeysAnalyzer.cs
```diff
@@ -56,70 +57,55 @@ public override void Initialize(AnalysisContext context)
         {
             context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
             context.EnableConcurrentExecution();
-            context.RegisterSyntaxNodeAction(AnalyzeInvocationExpression, SyntaxKind.InvocationExpression);
+            context.RegisterCompilationStartAction(compilationContext =>
+            {
+                var wellKnownTypeProvider = WellKnownTypeProvider.GetOrCreate(compilationContext.Compilation);
+                var targetTypes = new TargetTypeSymbols(
+                    wellKnownTypeProvider.GetOrCreateTypeByMetadataName(WellKnownTypeNames.ConfigurationBuilder),
```
andrewlock: Be aware that these can return null (if it can't be found). Maybe we should enable nullability in the analyzers? What should we do if the type _can't_ be found for some reason. should we bail _befre_ calling `compilationContext.RegisterSyntaxNodeAction` (that's the usual pattern 🙂 )

tracer/src/Datadog.Trace/Util/EnvironmentHelpersNoLogging.cs
andrewlock: Do we use it anywhere other than there I mean? 🙂

tracer/src/Datadog.Trace/Datadog.Trace.csproj
```diff
@@ -27,7 +27,7 @@
     from a source generator as opposed to something that is coming from some other tool. -->
     <Compile Remove="$(GeneratedFolder)/*/**/*.cs" />
     <None Include="$(GeneratedFolder)/*/**/*.cs" />
-    <Compile Update="Configuration\ConfigurationKeys.*.cs" DependentUpon="ConfigurationKeys.cs" />
```
andrewlock, line 30: Did you mean to remove this? You could do something similar for PlatformKeys FWIW

anna-git: Yes cause now they're generated so they just appear in the Generated folder already under `ConfigurationKeysGenerator` but the `PlatformKeys` aren't generated so I thought nesting them is still nicer

tracer/src/Datadog.Trace/Util/EnvironmentHelpersNoLogging.cs
anna-git: oh yes it's used only there, but then we would have to change method `EnvironmentHelpersNoLogging.GetEnvironmentVariable` from private to internal which we might not want to encourage?

tracer/src/Datadog.Trace.Tools.Analyzers/ConfigurationAnalyzers/ConfigurationBuilderWithKeysAnalyzer.cs
anna-git: Good point, actually I'm checking for Datadog.Trace that these types exist and otherwise the analyzers error, that way I can be sure nobody's renamed them and avoid analyzers silently failing, in 5dd057239670d8211130efdbb1e4be9607d099b0

tracer/src/Datadog.Trace.Tools.Analyzers/ConfigurationAnalyzers/ConfigurationBuilderWithKeysAnalyzer.cs
anna-git: yes I should take this good habit! fixed in 46a9848e1e4c2ab9bbfaf42df3297a5606e48375

PR #7935, opened 2025-12-09
Fix release GitHub Actions

andrewlock:
## Summary of changes

Add the missing secrets to the draft release process

## Reason for change

We recently added more secrets for uploading public symbols, and forgot to add them to the new workflows

Tried to use the new `create_hotfix_release` action earlier, and it didn't work. Luckily we kept the fallback around for just such an occasion 😄 

## Implementation details

Mark the secrets as required in the reusable workflow, and pass them in the workflows that use it

## Test coverage

YOLO


bouwkast (approved):
Thanks!

PR #7937, opened 2025-12-10
[Config Registry] 7/8 Integration names to keys 

anna-git:
## Context

Part of **Configuration Inversion (Step 6)** - Stack progress:
1.  [#7548](https://github.com/DataDog/dd-trace-dotnet/pull/7548) - Add GitLab step and JSON configuration file
2. [#7688](https://github.com/DataDog/dd-trace-dotnet/pull/7688) - Cleanup configuration / platform keys + source generator
3. [#7698](https://github.com/DataDog/dd-trace-dotnet/pull/7698) - Aliases handling via source generator
4. [#7689](https://github.com/DataDog/dd-trace-dotnet/pull/7689) - Analyzers for platform and ConfigurationBuilder
6. [#7932](https://github.com/DataDog/dd-trace-dotnet/pull/7932) - Forbid use of System.Environment methods and adapt everywhere
7. **→ [#7937](https://github.com/DataDog/dd-trace-dotnet/pull/7937) - Integration names to generated keys (this PR)**
## Summary of changes

Extends the `EnumExtensionsGenerator` to generate [IntegrationNameToKeys](cci:2://file:///Users/anna.yafi/go/src/github.com/DataDog/dd-trace-dotnet3/tracer/src/Datadog.Trace/Generated/net6.0/Datadog.Trace.SourceGenerators/EnumExtensionsGenerator/IntegrationNameToKeys.g.cs:14:4-392:5) class for the `IntegrationId` enum, providing centralized configuration key mapping for integrations including enabled, analytics enabled, and analytics sample rate keys. Also fixes AdoNet instrumentation compilation error.

## Reason for change

- Eliminates code duplication in configuration key generation across the codebase
- Provides compile-time safety for integration configuration key access
- Centralizes key pattern definitions with proper documentation
- Ensures consistency across all integration key formats (canonical, mixed case, short aliases)

## Implementation details

- **Source Generator**: Extended `EnumExtensionsGenerator` with [GenerateIntegrationNameToKeys](cci:1://file:///Users/anna.yafi/go/src/github.com/DataDog/dd-trace-dotnet3/tracer/src/Datadog.Trace.SourceGenerators/EnumExtensions/Sources.cs:82:4-178:5) method that generates:
  - `AllIntegrationEnabledKeys` array containing all enabled configuration keys
  - [GetIntegrationEnabledKeys(string)](cci:1://file:///Users/anna.yafi/go/src/github.com/DataDog/dd-trace-dotnet3/tracer/src/Datadog.Trace/Generated/net6.0/Datadog.Trace.SourceGenerators/EnumExtensionsGenerator/IntegrationNameToKeys.g.cs:102:8-197:9) - returns enabled keys for an integration
  - [GetIntegrationAnalyticsEnabledKeys(string)](cci:1://file:///Users/anna.yafi/go/src/github.com/DataDog/dd-trace-dotnet3/tracer/src/Datadog.Trace/Generated/net6.0/Datadog.Trace.SourceGenerators/EnumExtensionsGenerator/IntegrationNameToKeys.g.cs:198:8-294:9) - returns analytics enabled keys (marked obsolete)
  - [GetIntegrationAnalyticsSampleRateKeys(string)](cci:1://file:///Users/anna.yafi/go/src/github.com/DataDog/dd-trace-dotnet3/tracer/src/Datadog.Trace/Generated/net6.0/Datadog.Trace.SourceGenerators/EnumExtensionsGenerator/IntegrationNameToKeys.g.cs:295:8-391:9) - returns analytics sample rate keys (marked obsolete)
  - Key pattern constants (`IntegrationEnabledKeyPattern`, `AnalyticsEnabledKeyPattern`, `AnalyticsSampleRateKeyPattern`)
- **Configuration**: Updated [ConfigurationBuilder.WithIntegrationAnalyticsKey](cci:1://file:///Users/anna.yafi/go/src/github.com/DataDog/dd-trace-dotnet3/tracer/src/Datadog.Trace/Configuration/ConfigurationSources/Telemetry/ConfigurationBuilder.cs:32:4-42:5) and [WithIntegrationAnalyticsSampleRateKey](cci:1://file:///Users/anna.yafi/go/src/github.com/DataDog/dd-trace-dotnet3/tracer/src/Datadog.Trace/Configuration/ConfigurationSources/Telemetry/ConfigurationBuilder.cs:44:4-54:5) to use generated methods

## Test coverage

- Added [CanGenerateIntegrationNameToKeysForIntegrationId](cci:1://file:///Users/anna.yafi/go/src/github.com/DataDog/dd-trace-dotnet3/tracer/test/Datadog.Trace.SourceGenerators.Tests/EnumExtensionsGeneratorTests.cs:411:4-527:5) - verifies complete generated output with all three methods
- Added [DoesNotGenerateIntegrationNameToKeysForNonIntegrationIdEnum](cci:1://file:///Users/anna.yafi/go/src/github.com/DataDog/dd-trace-dotnet3/tracer/test/Datadog.Trace.SourceGenerators.Tests/EnumExtensionsGeneratorTests.cs:529:4-552:5) - ensures generation only for `IntegrationId` enum
- Added [IntegrationNameToKeysGeneratesCorrectKeyFormats](cci:1://file:///Users/anna.yafi/go/src/github.com/DataDog/dd-trace-dotnet3/tracer/test/Datadog.Trace.SourceGenerators.Tests/EnumExtensionsGeneratorTests.cs:554:4-670:5) - validates key formats for various integration names
- All tests verify obsolete attributes, key patterns, and switch expression logic

### DIFF
--- a/tracer/src/Datadog.Trace/Generated/net461/Datadog.Trace.SourceGenerators/EnumExtensionsGenerator/IntegrationNameToKeys.g.cs
+++ b/tracer/src/Datadog.Trace/Generated/net461/Datadog.Trace.SourceGenerators/EnumExtensionsGenerator/IntegrationNameToKeys.g.cs
@@ -256,6 +256,9 @@ namespace Datadog.Trace.Configuration
             "DD_TRACE_HANGFIRE_ENABLED", "DD_TRACE_Hangfire_ENABLED", "DD_Hangfire_ENABLED",
             "DD_TRACE_HANGFIRE_ANALYTICS_ENABLED", "DD_TRACE_Hangfire_ANALYTICS_ENABLED", "DD_Hangfire_ANALYTICS_ENABLED",
             "DD_TRACE_HANGFIRE_ANALYTICS_SAMPLE_RATE", "DD_TRACE_Hangfire_ANALYTICS_SAMPLE_RATE", "DD_Hangfire_ANALYTICS_SAMPLE_RATE", 
+            "DD_TRACE_OPENFEATURE_ENABLED", "DD_TRACE_OpenFeature_ENABLED", "DD_OpenFeature_ENABLED",
+            "DD_TRACE_OPENFEATURE_ANALYTICS_ENABLED", "DD_TRACE_OpenFeature_ANALYTICS_ENABLED", "DD_OpenFeature_ANALYTICS_ENABLED",
+            "DD_TRACE_OPENFEATURE_ANALYTICS_SAMPLE_RATE", "DD_TRACE_OpenFeature_ANALYTICS_SAMPLE_RATE", "DD_OpenFeature_ANALYTICS_SAMPLE_RATE", 
         ];
         /// <summary>
         /// Gets the configuration keys for the specified integration name.
@@ -344,6 +347,7 @@ namespace Datadog.Trace.Configuration
                 "AzureEventHubs" => new("DD_TRACE_AZUREEVENTHUBS_ENABLED", ["DD_TRACE_AzureEventHubs_ENABLED", "DD_AzureEventHubs_ENABLED"]),
                 "DatadogTraceVersionConflict" => new("DD_TRACE_DATADOGTRACEVERSIONCONFLICT_ENABLED", ["DD_TRACE_DatadogTraceVersionConflict_ENABLED", "DD_DatadogTraceVersionConflict_ENABLED"]),
                 "Hangfire" => new("DD_TRACE_HANGFIRE_ENABLED", ["DD_TRACE_Hangfire_ENABLED", "DD_Hangfire_ENABLED"]),
+                "OpenFeature" => new("DD_TRACE_OPENFEATURE_ENABLED", ["DD_TRACE_OpenFeature_ENABLED", "DD_OpenFeature_ENABLED"]),
                 _ => GetIntegrationEnabledKeysFallback(integrationName) // we should never get here
             };
         /// <summary>
@@ -434,6 +438,7 @@ namespace Datadog.Trace.Configuration
                 "AzureEventHubs" => new("DD_TRACE_AZUREEVENTHUBS_ANALYTICS_ENABLED", ["DD_TRACE_AzureEventHubs_ANALYTICS_ENABLED", "DD_AzureEventHubs_ANALYTICS_ENABLED"]),
                 "DatadogTraceVersionConflict" => new("DD_TRACE_DATADOGTRACEVERSIONCONFLICT_ANALYTICS_ENABLED", ["DD_TRACE_DatadogTraceVersionConflict_ANALYTICS_ENABLED", "DD_DatadogTraceVersionConflict_ANALYTICS_ENABLED"]),
                 "Hangfire" => new("DD_TRACE_HANGFIRE_ANALYTICS_ENABLED", ["DD_TRACE_Hangfire_ANALYTICS_ENABLED", "DD_Hangfire_ANALYTICS_ENABLED"]),
+                "OpenFeature" => new("DD_TRACE_OPENFEATURE_ANALYTICS_ENABLED", ["DD_TRACE_OpenFeature_ANALYTICS_ENABLED", "DD_OpenFeature_ANALYTICS_ENABLED"]),
                 _ => GetIntegrationAnalyticsEnabledKeysFallback(integrationName) // we should never get here
             };
         /// <summary>
@@ -524,6 +529,7 @@ namespace Datadog.Trace.Configuration
                 "AzureEventHubs" => new("DD_TRACE_AZUREEVENTHUBS_ANALYTICS_SAMPLE_RATE", ["DD_TRACE_AzureEventHubs_ANALYTICS_SAMPLE_RATE", "DD_AzureEventHubs_ANALYTICS_SAMPLE_RATE"]),
                 "DatadogTraceVersionConflict" => new("DD_TRACE_DATADOGTRACEVERSIONCONFLICT_ANALYTICS_SAMPLE_RATE", ["DD_TRACE_DatadogTraceVersionConflict_ANALYTICS_SAMPLE_RATE", "DD_DatadogTraceVersionConflict_ANALYTICS_SAMPLE_RATE"]),
                 "Hangfire" => new("DD_TRACE_HANGFIRE_ANALYTICS_SAMPLE_RATE", ["DD_TRACE_Hangfire_ANALYTICS_SAMPLE_RATE", "DD_Hangfire_ANALYTICS_SAMPLE_RATE"]),
+                "OpenFeature" => new("DD_TRACE_OPENFEATURE_ANALYTICS_SAMPLE_RATE", ["DD_TRACE_OpenFeature_ANALYTICS_SAMPLE_RATE", "DD_OpenFeature_ANALYTICS_SAMPLE_RATE"]),
                 _ => GetIntegrationAnalyticsSampleRateKeysFallback(integrationName) // we should never get here
             };
 
--- a/tracer/src/Datadog.Trace/Generated/net6.0/Datadog.Trace.SourceGenerators/EnumExtensionsGenerator/IntegrationNameToKeys.g.cs
+++ b/tracer/src/Datadog.Trace/Generated/net6.0/Datadog.Trace.SourceGenerators/EnumExtensionsGenerator/IntegrationNameToKeys.g.cs
@@ -256,6 +256,9 @@ namespace Datadog.Trace.Configuration
             "DD_TRACE_HANGFIRE_ENABLED", "DD_TRACE_Hangfire_ENABLED", "DD_Hangfire_ENABLED",
             "DD_TRACE_HANGFIRE_ANALYTICS_ENABLED", "DD_TRACE_Hangfire_ANALYTICS_ENABLED", "DD_Hangfire_ANALYTICS_ENABLED",
             "DD_TRACE_HANGFIRE_ANALYTICS_SAMPLE_RATE", "DD_TRACE_Hangfire_ANALYTICS_SAMPLE_RATE", "DD_Hangfire_ANALYTICS_SAMPLE_RATE", 
+            "DD_TRACE_OPENFEATURE_ENABLED", "DD_TRACE_OpenFeature_ENABLED", "DD_OpenFeature_ENABLED",
+            "DD_TRACE_OPENFEATURE_ANALYTICS_ENABLED", "DD_TRACE_OpenFeature_ANALYTICS_ENABLED", "DD_OpenFeature_ANALYTICS_ENABLED",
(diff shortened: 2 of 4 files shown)


tracer/src/Datadog.Trace.SourceGenerators/EnumExtensions/EnumExtensionsGenerator.cs
```diff
@@ -57,6 +57,13 @@ private static void Execute(in EnumToGenerate enumToGenerate, SourceProductionCo
         StringBuilder sb = new StringBuilder();
         var result = Sources.GenerateExtensionClass(sb, in enumToGenerate);
         context.AddSource(enumToGenerate.ExtensionsName + "_EnumExtensions.g.cs", SourceText.From(result, Encoding.UTF8));
+
+        // If this is the IntegrationId enum, also generate IntegrationNameToKeys
+        if (enumToGenerate.FullyQualifiedName == "Datadog.Trace.Configuration.IntegrationId")
```
andrewlock, line 62: This feels so hacky, but I love it 😂

tracer/src/Datadog.Trace.SourceGenerators/EnumExtensions/Sources.cs
```diff
@@ -80,6 +80,124 @@ public static string[] GetNames()
...
+            var sampleRateShortKey = $"DD_{name}_ANALYTICS_SAMPLE_RATE";
+            switchCasesAnalyticsSampleRate.AppendLine($"                \"{name}\" => new[] {{ \"{sampleRateUpperKey}\", \"{sampleRateMixedKey}\", \"{sampleRateShortKey}\" }},");
+            arrayKeys.AppendLine($"            \"{sampleRateUpperKey}\", \"{sampleRateMixedKey}\", \"{sampleRateShortKey}\", ");
+        }
+
+        sb.Clear();
+        sb.AppendLine(Constants.FileHeader);
+        sb.AppendLine("namespace Datadog.Trace.Configuration");
+        sb.AppendLine("{");
+        sb.AppendLine("    /// <summary>");
```
andrewlock: It looks neat, but doing these as individual `sb.AppendLine` is very expensive. You want to do as big a block as you can in general, so raw strings are the best option generally. 

e.g. 

```csharp
sb.AppendLine(
Constants.FileHeader + // as long as this is a constant, the concat happens at compile time
"""

namespace Datadog.Trace.Configuration
{
    /// Generated mapping of integration names to their configuration keys.");
    /// </summary>");
    internal static partial class IntegrationNameToKeys");
    {
        private const string ObsoleteMessage = DeprecationMessages.AppAnalytics;");

        /// <summary>");
        /// All integration enabled keys (canonical + aliases).");
        /// </summary>");
        public static readonly string[] AllIntegrationEnabledKeys = new[]");
        {
""");
```
etc


It also means you don't need to escape anything, just copy-paste the code you expect in there, and "break out" where you need something specific (like the actual values) 🙂

tracer/src/Datadog.Trace.SourceGenerators/EnumExtensions/Sources.cs
```diff
@@ -80,6 +80,124 @@ public static string[] GetNames()
...
 
+    public static string GenerateIntegrationNameToKeys(StringBuilder sb, in EnumToGenerate enumToGenerate)
+    {
+        var arrayKeys = new StringBuilder();
+        var switchCasesEnabled = new StringBuilder();
+        var switchCasesAnalyticsEnabled = new StringBuilder();
+        var switchCasesAnalyticsSampleRate = new StringBuilder();
+
+        // Single loop to build array keys and all switch cases
+        foreach (var member in enumToGenerate.Names)
```
andrewlock: nit, I would have just iterated it 3 times, and append to the same stringbuilder instead of having multiple builders. The iteration is low overhead compared to the stringbuilder (which will each have to grow their internal array multiple times)

tracer/src/Datadog.Trace.SourceGenerators/EnumExtensions/Sources.cs
```diff
@@ -80,6 +80,124 @@ public static string[] GetNames()
...
+        sb.AppendLine("        {");
+        sb.AppendLine("            return integrationName switch");
+        sb.AppendLine("            {");
+        sb.Append(switchCasesEnabled);
+        sb.AppendLine("                _ => new[]");
+        sb.AppendLine("                {");
+        sb.AppendLine("                    string.Format(\"DD_TRACE_{0}_ENABLED\", integrationName.ToUpperInvariant()),");
+        sb.AppendLine("                    string.Format(\"DD_TRACE_{0}_ENABLED\", integrationName),");
+        sb.AppendLine("                    $\"DD_{integrationName}_ENABLED\"");
+        sb.AppendLine("                }");
```
andrewlock: This is an interesting one tbh, do we even _need_ this branch 🤔 Given we handle this at compile time, we know it will never be called. That said, throwing probably isn't great. 

We _might_ want to generate something more like this:

```csharp
_ => GetEnabledFallbackFor(integrationName),
```

because it should generate smaller code, which has more chance of optimization, and the fallback will never be invoked,

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/Telemetry/ConfigurationBuilder.cs
```diff
@@ -19,36 +20,39 @@ internal readonly struct ConfigurationBuilder(IConfigurationSource source, IConf
...
-        _source,
-        _telemetry,
+    public HasKeys WithIntegrationKey(string integrationName)
+    {
+        var integrationEnabledKeys = IntegrationNameToKeys.GetIntegrationEnabledKeys(integrationName);
+        return new(
+            _source,
+            _telemetry,
+            integrationEnabledKeys[0],
+            integrationEnabledKeys.Skip(1).ToArray());
```
andrewlock: oh no, `Skip(1).ToArray()` 🙁 We definitely don't want to do that, If we're _going_ to need to do that, we should change the `GetIntegrationEnabledKeys()` API to return an array of the correct size (e.g. a "main" key plus an array of fallbacks). 

If _possible_ we should probably try to get away from using arrays at all... we _may_ be able to use `ReadOnlySpan<string>` in some places, or we should use `InlineArray` (if we can), or worse case we should create a specialized type to avoid creating all these small arrays if possible.

tracer/src/Datadog.Trace/Generated/net461/Datadog.Trace.SourceGenerators/EnumExtensionsGenerator/IntegrationNameToKeys.g.cs
```diff
@@ -0,0 +1,550 @@
...
+        /// Gets the configuration keys for the specified integration name.
+        /// Returns keys in order: [canonical key, alias1, alias2]
+        /// </summary>
+        /// <param name="integrationName">The integration name</param>
+        /// <returns>Array of configuration keys to check in order</returns>
+        public static string[] GetIntegrationEnabledKeys(string integrationName)
+        {
+            return integrationName switch
+            {
+                "HttpMessageHandler" => new[] { "DD_TRACE_HTTPMESSAGEHANDLER_ENABLED", "DD_TRACE_HttpMessageHandler_ENABLED", "DD_HttpMessageHandler_ENABLED" },
```
andrewlock: We should use collection expressions wherever possible
```suggestion
                "HttpMessageHandler" => ["DD_TRACE_HTTPMESSAGEHANDLER_ENABLED", "DD_TRACE_HttpMessageHandler_ENABLED", "DD_HttpMessageHandler_ENABLED"],
```

tracer/src/Datadog.Trace/Generated/net461/Datadog.Trace.SourceGenerators/EnumExtensionsGenerator/IntegrationNameToKeys.g.cs
```diff
@@ -0,0 +1,550 @@
...
+    /// Generated mapping of integration names to their configuration keys.
+    /// </summary>
+    internal static partial class IntegrationNameToKeys
+    {
+        private const string ObsoleteMessage = DeprecationMessages.AppAnalytics;
+
+        /// <summary>
+        /// All integration enabled keys (canonical + aliases).
+        /// </summary>
+        public static readonly string[] AllIntegrationEnabledKeys = new[]
```
andrewlock: AFAICT, this is only called in a test. But the array will be created as soon as we call `IntegrationNameToKeys.Something`. We can solve that by making it a method instead (and also marking it as only to be called in tests)

```suggestion
        [TestingOnly]
        public static readonly string[] GetAllIntegrationEnabledKeys() => new[]
```

tracer/src/Datadog.Trace.SourceGenerators/EnumExtensions/Sources.cs
anna-git: Ok! chhanged in e33db01789c68f55b884e0b1296f5b3073bc38f8

tracer/src/Datadog.Trace.SourceGenerators/EnumExtensions/Sources.cs
anna-git: yep thank you changed in e33db01789c68f55b884e0b1296f5b3073bc38f8

tracer/src/Datadog.Trace.SourceGenerators/EnumExtensions/Sources.cs
anna-git: thanks changed in e33db01789c68f55b884e0b1296f5b3073bc38f8

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/Telemetry/ConfigurationBuilder.cs
anna-git: very good point fixed in e33db01789c68f55b884e0b1296f5b3073bc38f8

tracer/src/Datadog.Trace/Generated/net461/Datadog.Trace.SourceGenerators/EnumExtensionsGenerator/IntegrationNameToKeys.g.cs
anna-git: fixed in e33db01789c68f55b884e0b1296f5b3073bc38f8

tracer/src/Datadog.Trace/Generated/net461/Datadog.Trace.SourceGenerators/EnumExtensionsGenerator/IntegrationNameToKeys.g.cs
anna-git: added in e33db01789c68f55b884e0b1296f5b3073bc38f8

duncanpharvey:
Removing @DataDog/serverless-azure-and-gcp as a reviewer since I don't see any Azure or GCP related changes. Maybe there were some changes before a force push but I don't see any now.

PR #7944, opened 2025-12-10
[Config Registry] Fix source generators: downgrade System.text.json version

anna-git:
## Summary of changes

Due to this [Rider](https://youtrack.jetbrains.com/projects/RIDER/issues/RIDER-132634/Source-Generated-files-missing-when-using-System.Text.Json) issue , the IDE doesn't index source generated files using the last version of the package properly which considerably degrades the dev experience. And following the merge [of this PR](https://github.com/DataDog/dd-trace-dotnet/pull/7931), we now have the issue:
image

It still builds but Rider doesn't see the generated sources and generates errors and doesn't autocompletes or index anything.

## Reason for change

Downgrade the package version to fix the dev experience.


anna-git:
> Nice! Did you try with the 10.0 GA version of system.text.json btw?

yes I tried with 10+ versions and it started working only from under 9...

PR #7946, opened 2025-12-11
Update code normalizing tag values

vandonr:
## Summary of changes

Since this code was initially written, some changes have been introduced in the "reference code" in the agent, mostly:
 - allowing leading digits (which also had the side effect of allowing leading non-alpha characters of all kinds, **currently working on clarifying if it was intended**): https://github.com/DataDog/datadog-agent/pull/17673
 - another small change fixing an edge case around max length: https://github.com/DataDog/datadog-agent/pull/13235

I'm porting both changes to the dotnet code. Also changed the tests, since the behavior changed.

## Reason for change

keeping things aligned

## Implementation details

also ported another small optimization where we do a first pass on the value to see if it's already well formatted. Benchmarking this would require running it against real world data, but I guess if it's a win in the agent, it's probably a win here as well ?

## Test coverage

just changed the expected value in existing tests (+ made it a `Theory` so that I can see all failures at once instead of it stopping on the first failure)


tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/RuntimeMetricsTests.cs
vandonr, line 87: not sure if I should be fixing the expected value where I did, or the service name (that may have been intended as a test on normalization)

vandonr: There is actually a _different_ [method](https://github.com/DataDog/datadog-agent/blob/055b7e3e80aa5c028504fe6c1c7ae00006cb5622/pkg/trace/traceutil/normalize/normalize.go#L80) in the agent if we normalize a service name 🫠 
Sooo... idk

tracer/src/Datadog.Trace/Processors/TraceUtil.cs
```diff
@@ -54,17 +54,12 @@ public static bool IsValidStatusCode(string statusCode)
             return false;
         }
 
-        // https://github.com/DataDog/datadog-agent/blob/eac2327c5574da7f225f9ef0f89eaeb05ed10382/pkg/trace/traceutil/normalize.go#L98-L209
+        // https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/traceutil/normalize/normalize.go#L152
```
lucaspimentel: I think it would be nice to have a commit hash here instead of `main` so we know exactly which version of the Go code this was based on. Copying the hash you used below:
```suggestion
        // https://github.com/DataDog/datadog-agent/blob/5e576f16449f2cc003d231ad50d54c920fdee08f/pkg/trace/traceutil/normalize/normalize.go#L152
```

tracer/src/Datadog.Trace/Processors/TraceUtil.cs
```diff
@@ -210,6 +201,52 @@ public static string NormalizeTag(string value)
...
 
+        // https://github.com/DataDog/datadog-agent/blob/5e576f16449f2cc003d231ad50d54c920fdee08f/pkg/trace/traceutil/normalize/normalize.go#L366
+        private static bool IsNormalizedAsciiTag(string tagValue)
+        {
+            if (string.IsNullOrEmpty(tagValue))
+            {
+                return true;
+            }
+
+            if (tagValue.Length > MaxTagLength)
```
lucaspimentel, line 212: I don't think we should do this check in `IsNormalizedAsciiTag()`. If a string is all "normalized ascii" (as per the method name) but is longer than the max, then we only needs to get a substring instead of going through the entire normalization code. We could optimize better for that case.

tracer/src/Datadog.Trace/Processors/TraceUtil.cs
```diff
@@ -54,17 +54,12 @@ public static bool IsValidStatusCode(string statusCode)
             return false;
         }
 
-        // https://github.com/DataDog/datadog-agent/blob/eac2327c5574da7f225f9ef0f89eaeb05ed10382/pkg/trace/traceutil/normalize.go#L98-L209
+        // https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/traceutil/normalize/normalize.go#L152
         public static string NormalizeTag(string value)
```
lucaspimentel, line 58: Is `NormalizeTag()` normalizing a tag name or a tag value? They have different normalization rules.

tracer/src/Datadog.Trace/Processors/TraceUtil.cs
```diff
@@ -210,6 +201,52 @@ public static string NormalizeTag(string value)
...
+            }
+
+            return true;
+        }
+
+        // https://github.com/DataDog/datadog-agent/blob/5e576f16449f2cc003d231ad50d54c920fdee08f/pkg/trace/traceutil/normalize/normalize.go#L403
+        private static bool IsValidAsciiTagChar(char c)
+        {
+            // where we use this method, the agent's code actually uses a lookup table for faster computation
+            return ('a' <= c && c <= 'z') || ('0' <= c && c <= '9') || c == ':' || c == '.' || c == '/' || c == '-';
```
lucaspimentel, line 247: What about capital letters? If they are also valid, you can replace this with `IsAlphaNumeric()` below plus the additional symbols:

```suggestion
            return IsAlphaNumeric(c) || c == ':' || c == '.' || c == '/' || c == '-';
```

tracer/test/Datadog.Trace.Tests/TraceProcessors/NormalizerTraceProcessorTests.cs
```diff
@@ -12,69 +12,65 @@ namespace Datadog.Trace.Tests.TraceProcessors
...
-                yield return new object[] { "a\xFFFD\xFFFD", "a" };
-                yield return new object[] { "a\xFFFD\xFFFDb", "a_b" };
-                yield return new object[] { null, null };
-            }
+        // https://github.com/DataDog/datadog-agent/blob/eac2327c5574da7f225f9ef0f89eaeb05ed10382/pkg/trace/traceutil/normalize_test.go#L17
+        [Theory]
+        [MemberData(nameof(GetNormalizeTagValues))]
+        public void NormalizeTagTests(string inValue, string expectedValue)
+        {
+            var actualValue = Processors.TraceUtil.NormalizeTag(inValue);
```
lucaspimentel, line 72: Since this is using `GetNormalizeTagValues()`, I assume it's testing normalization of tag values, not names? Can we rename the test itself, too?

tracer/src/Datadog.Trace/Processors/TraceUtil.cs
vandonr: I did that because I'm just porting the code from the go implem, that has this check too : https://github.com/DataDog/datadog-agent/blob/5e576f16449f2cc003d231ad50d54c920fdee08f/pkg/trace/traceutil/normalize/normalize.go#L370
(but yes you are right, this can probably be optimized, but in that case it might make sense to port the optimization to the agent too)

tracer/src/Datadog.Trace/Processors/TraceUtil.cs
vandonr: Capital letters are not valid, they need to be `toLower`ed.

tracer/test/Datadog.Trace.Tests/TraceProcessors/NormalizerTraceProcessorTests.cs
vandonr: I'll double check, but I _think_ we're normalizing entire key-value pairs already in the form `key:value`, so we're not really normalizing either key nor values. 
We're actually doing https://github.com/DataDog/datadog-agent/blob/5e576f16449f2cc003d231ad50d54c920fdee08f/pkg/trace/traceutil/normalize/normalize.go#L144, so I need to change the line the permalink is pointing at

vandonr: I really need to spend more time looking at where this is used, because I think we use this code both to normalize tags and service names, and the rules are not the same, + changing it would be a potential breaking change

vandonr:
too dangerous to merge and not important enough (no one asked for anything about it)

PR #7947, opened 2025-12-11
[Iast] Fix GetModuleInfo freezes

daniel-romano-DD:
## Summary of changes
Move all GetModuleInfo calls to the `ModuleLoadeFinished`  profiler event.

## Reason for change
Some freezes could happen in Linux

## Implementation details
Calling GetAssemblyInfo outside of the Module and Assembly events may corrupt the process

### DIFF
--- a/tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
+++ b/tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
@@ -384,10 +384,10 @@ HRESULT Dataflow::AppDomainShutdown(AppDomainID appDomainId)
 
 HRESULT Dataflow::ModuleLoaded(ModuleID moduleId, ModuleInfo** pModuleInfo)
 {
+    CSGUARD(_cs);
     // Retrieve all already modules at once to mimic initialization from creation behavior
     if (_preLoadedModuleIds.size() > 0)
     {
-        CSGUARD(_cs);
         for (auto const& id : _preLoadedModuleIds)
         {
             GetModuleInfo(id);


tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
```diff
@@ -379,6 +384,17 @@ HRESULT Dataflow::AppDomainShutdown(AppDomainID appDomainId)
...
 {
+    // Retrieve all already modules at once to mimic initialization from creation behavior
+    if (_preLoadedModuleIds.size() > 0)
+    {
+        CSGUARD(_cs);
+        for (auto const& id : _preLoadedModuleIds)
+        {
+            GetModuleInfo(id);
+        }
+        _preLoadedModuleIds.clear();
```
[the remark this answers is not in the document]
dromanol: The problem is the only safe places to call GetAssemblyInfo is in module related events

tracer/src/Datadog.Tracer.Native/iast/dataflow.cpp
```diff
@@ -526,20 +542,21 @@ ModuleInfo* Dataflow::GetModuleInfo(ModuleID id)
     }
 
     // Retrieve module information if not found
+    const int pathLen = 2048;
```
tonyredondo, line 545: why the change on the length? you had 300 before.

dromanol: Just to erase the shadow of a buffer overrun

andrewlock (approved):
Thanks!

e-n-0 (approved):

PR #7949, opened 2025-12-12
[DSM] Initial transaction tracking implementation

kr-igor:
# What Does This Do
Initial transaction tracking implementation.

* Support for extractors configuration
* Serialization for extractors 
* Extractors for http / kafka
* Updated / added unit tests

# Motivation
Multiple customers mentioned the need to track individual message across multiple service and environments and log messages which haven't reached the end of the pre-defined pipeline.
See [this](https://docs.google.com/document/d/1jvV63gUD0FZfr5hOewMsO5ibd99HkkjIpu0m2tXKTCs/edit?tab=t.0#heading=h.gv7mzukw9gk0) document for details.
Payload format description is [here](https://docs.google.com/document/d/1jvV63gUD0FZfr5hOewMsO5ibd99HkkjIpu0m2tXKTCs/edit?tab=t.0#heading=h.lqgt42jozm9j).

# Additional Notes
Transaction tracking is currently in closed beta, so the configuration flags will be documented after public availability.

# Technical details
Adds DSM transaction tracking by introducing a DataStreamsTransactionContainer that accumulates serialized transaction records (checkpoint ID, nanosecond timestamp, and UTF-8 transaction ID bytes) within the existing
  DataStreamsAggregator flush cycle, triggering an early flush when the buffer exceeds 512 KB. 
Transaction extraction is driven by a configurable DataStreamsExtractorRegistry (loaded from DD_DATA_STREAMS_TRANSACTION_EXTRACTORS 
  JSON) that maps named extractors to header keys for four integration types: 
* HTTP_IN_HEADERS (ASP.NET Core incoming requests)
* HTTP_OUT_HEADERS (HttpClient outgoing requests)
* KAFKA_CONSUME_HEADERS, and KAFKA_PRODUCE_HEADERS.

### DIFF
--- a/tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
+++ b/tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
@@ -263,7 +263,10 @@ internal sealed class DataStreamsWriter : IDataStreamsWriter
 
                 while (_transactionBuffer.TryDequeue(out var transactionPoint))
                 {
-                    _aggregator.AddTransaction(transactionPoint);
+                    if (!_aggregator.AddTransaction(transactionPoint))
+                    {
+                        Interlocked.Increment(ref _pointsDropped);
+                    }
                 }
             }
             catch (Exception ex)
@@ -370,7 +373,10 @@ internal sealed class DataStreamsWriter : IDataStreamsWriter
 
                 while (_transactionBuffer.TryDequeue(out var transactionPoint))
                 {
-                    _aggregator.AddTransaction(transactionPoint);
+                    if (!_aggregator.AddTransaction(transactionPoint))
+                    {
+                        Interlocked.Increment(ref _pointsDropped);
+                    }
                 }
 
                 if (_aggregator.ShouldFlushTransactions)
--- a/tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsTransactionContainer.cs
+++ b/tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsTransactionContainer.cs
@@ -16,6 +16,7 @@ namespace Datadog.Trace.DataStreamsMonitoring.TransactionTracking;
 internal sealed class DataStreamsTransactionContainer
 {
     private const int MaxSizeBytes = 512 * 1024;
+    private const int MaxDropSizeBytes = 2 * 1024 * 1024;
 
     private readonly int _initialByteSize;
 
@@ -30,8 +31,13 @@ internal sealed class DataStreamsTransactionContainer
 
     internal bool ShouldFlush => _size >= MaxSizeBytes;
 
-    public void Add(DataStreamsTransactionInfo transactionInfo)
+    public bool Add(DataStreamsTransactionInfo transactionInfo)
     {
+        if (_size >= MaxDropSizeBytes)
+        {
+            return false;
+        }
+
         var byteCount = transactionInfo.GetByteCount();
 
         if (_data.Length - _size < byteCount)
@@ -43,6 +49,7 @@ internal sealed class DataStreamsTransactionContainer
 
         transactionInfo.WriteTo(_data, _size);
         _size += byteCount;
+        return true;
     }
 
     public int Size() => _size;
--- a/tracer/src/Datadog.Trace/DataStreamsMonitoring/Aggregation/DataStreamsAggregator.cs
+++ b/tracer/src/Datadog.Trace/DataStreamsMonitoring/Aggregation/DataStreamsAggregator.cs
@@ -69,9 +69,9 @@ internal sealed class DataStreamsAggregator(DataStreamsMessagePackFormatter form
         }
     }
 
-    public void AddTransaction(in DataStreamsTransactionInfo transaction)
+    public bool AddTransaction(in DataStreamsTransactionInfo transaction)
     {
-        _dataStreamsTransactionContainer.Add(transaction);
+        return _dataStreamsTransactionContainer.Add(transaction);
     }
 
     /// <summary>


robcarlan-datadog:
@codex review

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
```diff
@@ -48,8 +50,10 @@ public DataStreamsManager(
         _isLegacyDsmHeadersEnabled = tracerSettings.IsDataStreamsLegacyHeadersEnabled;
         _writer = writer;
         _discoveryService = discoveryService;
-        _isInDefaultState = tracerSettings.IsDataStreamsMonitoringInDefaultState;
+        IsInDefaultState = tracerSettings.IsDataStreamsMonitoringInDefaultState;
+        _registry = new DataStreamsExtractorRegistry(tracerSettings.DataStreamsTransactionExtractors);
 
+        Log.Debug(@"Data Streams extractors loaded: {AsJson}", _registry.AsJson());
```
robcarlan-datadog: Is the @ needed?

tracer/test/Datadog.Trace.Tests/DataStreamsMonitoring/DataStreamsExtractorRegistryTest.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace Datadog.Trace.Tests.DataStreamsMonitoring;
+
+public class DataStreamsExtractorRegistryTest
+{
+    [Fact]
+    public void DeserializeCorrectly()
+    {
+        var registry = new DataStreamsExtractorRegistry("[{\"name\": \"transaction-origin\", \"type\": \"HTTP_OUT_HEADERS\", \"value\": \"transaction-id\"}]");
+        registry.AsJson().Should().Be("{\"HttpOutHeaders\":[{\"name\":\"transaction-origin\",\"type\":\"HTTP_OUT_HEADERS\",\"value\":\"transaction-id\",\"ExtractorType\":1}]}");
+    }
```
robcarlan-datadog, line 20: [nit] Could add a test for `GetExtractorsByType` that it can correctly store multiple values for the same type

tracer/test/Datadog.Trace.Tests/DataStreamsMonitoring/DataStreamsManagerTests.cs
```diff
@@ -249,6 +259,37 @@ public void WhenEnabled_SetCheckpoint_SetsSpanTags()
...
+    }
+
+    [Fact]
+    public void WhenDisabled_TrackTransaction_DoesNothing()
+    {
+        var dsm = GetDataStreamManager(false, out _);
+        var span = new Span(new SpanContext(traceId: 123, spanId: 456), DateTimeOffset.UtcNow);
+
+        var act = () => span.TrackTransaction(dsm, "tx-abc", "some-checkpoint");
+        act.Should().NotThrow();
```
robcarlan-datadog: Probably worth checking that writer.DataStreamsTransactions.Size() remains 0 here?

lucaspimentel:
@codex review

kr-igor:
All comments are now addressed.

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsTransactionInfo.cs
```diff
@@ -0,0 +1,123 @@
...
+#nullable enable
+
+using System;
+using System.Collections.Concurrent;
+using System.Text;
+using System.Threading;
+
+namespace Datadog.Trace.DataStreamsMonitoring.TransactionTracking;
+
+internal readonly struct DataStreamsTransactionInfo
```
bouwkast, line 37: Do we have a document/RFC that outlines this binary format?

kr-igor: Updated the doc, added comment to the code.

kr-igor:
> I think we need integration tests for this
> 
> `DataStreamsMonitoringKafkaTests.cs`
> 
> And it looks like since we want this added to HttpClient I think we'd get some tests there, I don't see any DSM stuff in it, unsure if we'd need a new sample or not

Added tests for http and kafka

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
bouwkast:
```suggestion
     if (Log.IsEnabled(LogEventLevel.Debug))
     {
         Log.Debug(@"Data Streams extractors loaded: {AsJson}", _registry.AsJson());
     }   
```

I think as written `.AsJson()` is called every single time this would be hit, unsure how often, but best to wrap I would think

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
```diff
@@ -175,6 +186,34 @@ public void InjectPathwayContext<TCarrier>(PathwayContext? context, TCarrier hea
...
+        var writer = Volatile.Read(ref _writer);
+        writer?.AddTransaction(new DataStreamsTransactionInfo(
+                                   transactionId,
+                                   DateTimeOffset.UtcNow.ToUnixTimeNanoseconds(),
+                                   checkpointName));
+    }
+
+    public void TrackTransaction(byte[] transactionIdBytes, string checkpointName)
+    {
+        if (!IsEnabled)
```
bouwkast:
```suggestion
        if (!IsTransactionTrackingEnabled)
```

Is this supposed to be this?

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
```diff
@@ -175,6 +186,34 @@ public void InjectPathwayContext<TCarrier>(PathwayContext? context, TCarrier hea
         DataStreamsContextPropagator.Instance.Inject(context.Value, headers, _isLegacyDsmHeadersEnabled);
     }
 
+    public void TrackTransaction(string transactionId, string checkpointName)
+    {
+        if (!IsEnabled)
```
bouwkast:
```suggestion
        if (!IsTransactionTrackingEnabled)
```

Is this supposed to be this?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Http/HttpClient/HttpMessageHandlerCommon.cs
```diff
@@ -42,6 +43,25 @@ public static CallTargetState OnMethodBegin<TTarget, TRequest>(TTarget instance,
...
+                    var extractors = dataStreamsManager.GetExtractorsByType(DataStreamsTransactionExtractor.Type.HttpOutHeaders);
+                    if (extractors != null)
+                    {
+                        foreach (var extractor in extractors)
+                        {
+                            if (headers.TryGetValues(extractor.Value, out var headerValues))
+                            {
+                                foreach (var headerValue in headerValues)
+                                {
+                                    dataStreamsManager.TrackTransaction(headerValue, extractor.Name);
```
bouwkast: Just wanted to confirm: Should we be tracking a transaction if `scope` is `null`? 

There is a check below on line 65 ` if (scope is not null)`

I think the transaction stuff isn't dependent on the scope, but just wanted to make sure that is accurate

tracer/src/Datadog.Trace/DataStreamsMonitoring/SpanExtensions.cs
```diff
@@ -9,6 +9,24 @@ namespace Datadog.Trace.DataStreamsMonitoring;
...
 internal static class SpanExtensions
 {
+    /// <summary>
+    /// Tracks a transaction at the given checkpoint and tags the span with the transaction ID.
+    /// </summary>
+    /// <param name="span">The span instance</param>
+    /// <param name="manager">The <see cref="DataStreamsManager"/> to use</param>
+    /// <param name="transactionId">The transaction identifier</param>
+    /// <param name="checkpointName">The checkpoint name at which the transaction is being tracked</param>
+    internal static void TrackTransaction(this Span span, DataStreamsManager? manager, string transactionId, string checkpointName)
```
bouwkast, line 19: I don't see any usages of this, is this still needed?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Http/HttpClient/HttpMessageHandlerCommon.cs
bouwkast: (If so it may be helpful to have a comment somewhere here stating why)

I also haven't checked when the `scope` could / would be `null` here

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsTransactionContainer.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        transactionInfo.WriteTo(_data, _size);
+        _size += byteCount;
+    }
+
+    public int Size() => _size;
+
+    public byte[] GetDataAndReset()
+    {
+        // trim zeros
```
bouwkast, line 64: Could we check if `_size` is `0` right away and if so just return an empty `[]`? Otherwise we always do `_data = new byte[_initialByteSize];` even when we don't need to (I think)

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsTransactionInfo.cs
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    internal static byte[] GetCacheBytes()
+    {
+        var result = new byte[512];
+        var index = 0;
+
+        foreach (var pair in Cache)
+        {
+            var keyBytes = Encoding.UTF8.GetBytes(pair.Key);
```
bouwkast:
```suggestion
            var keyBytes = Truncate(Encoding.UTF8.GetBytes(pair.Key));
```

Do we need to truncate this as well? I see we truncate the `_idBytes = Truncate(idBytes);` above

tracer/test/Datadog.Trace.Tests/DataStreamsMonitoring/DataStreamsTransactionExtractorTests.cs
```diff
@@ -0,0 +1,39 @@
...
+        registry.GetExtractorsByType(expected).Should().ContainSingle()
+                .Which.ExtractorType.Should().Be(expected);
+    }
+
+    [Fact]
+    public void ExtractorType_ReturnsSameValue_OnMultipleCalls()
+    {
+        var registry = new DataStreamsExtractorRegistry("[{\"name\": \"n\", \"type\": \"HTTP_OUT_HEADERS\", \"value\": \"v\"}]");
+        var extractor = registry.GetExtractorsByType(DataStreamsTransactionExtractor.Type.HttpOutHeaders)!.Single();
+        extractor.ExtractorType.Should().Be(extractor.ExtractorType);
```
bouwkast: Isn't this just validating that it is the same type as what it is?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Http/HttpClient/HttpMessageHandlerCommon.cs
```diff
@@ -42,6 +43,25 @@ public static CallTargetState OnMethodBegin<TTarget, TRequest>(TTarget instance,
                     integrationId,
                     out var tags);
 
+                var dataStreamsManager = tracer.TracerManager.DataStreamsManager;
+                if (dataStreamsManager.IsTransactionTrackingEnabled)
```
andrewlock, line 48: In the kafka helper you check for `dataStreamsManager.IsInDefaultState`. Should you be checking this here too? Shuoldn't the checks be the same?

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
```diff
@@ -48,8 +50,10 @@ public DataStreamsManager(
         _isLegacyDsmHeadersEnabled = tracerSettings.IsDataStreamsLegacyHeadersEnabled;
         _writer = writer;
         _discoveryService = discoveryService;
-        _isInDefaultState = tracerSettings.IsDataStreamsMonitoringInDefaultState;
+        IsInDefaultState = tracerSettings.IsDataStreamsMonitoringInDefaultState;
```
andrewlock: Exposing `IsInDefaultState` outside of this type feels like a smell. IMO we should expose the state of _features_, which _internally_ here can depend on `_isInDefaultState`. i.e. `IsTransactionTrackingEnabled` looks like it's doing the "correct" pattern, but I don't think we should expose this

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
andrewlock: 💯 definitely should do this (also remove the `@` 😄)

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
```diff
@@ -64,9 +68,11 @@ public bool IsEnabled
...
     }
 
-    public bool IsInDefaultState
+    public bool IsInDefaultState { get; }
+
+    public bool IsTransactionTrackingEnabled
     {
-        get => Volatile.Read(ref _isInDefaultState);
+        get => !IsInDefaultState && IsEnabled;
     }
```
andrewlock: nit
```suggestion
    public bool IsTransactionTrackingEnabled => !IsInDefaultState && IsEnabled;
```

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
```diff
@@ -64,9 +68,11 @@ public bool IsEnabled
         get => Volatile.Read(ref _isEnabled);
     }
 
-    public bool IsInDefaultState
+    public bool IsInDefaultState { get; }
+
+    public bool IsTransactionTrackingEnabled
```
andrewlock: Should we have a dedicated setting for this too 🤔 Because if there are problems with it, customers have no way to disable it without disabling the whole of DSM?

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
```diff
@@ -130,6 +132,24 @@ public void Add(in StatsPoint point)
...
+    {
+        if (!Volatile.Read(ref _isInitialized))
+        {
+            Initialize();
+        }
+
+        if (Volatile.Read(ref _isSupported) != SupportState.Unsupported)
+        {
+            if (_transactionBuffer.TryEnqueue(transaction))
+            {
```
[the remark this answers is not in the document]
andrewlock: I think this needs to be addressed

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsTransactionInfo.cs
```diff
@@ -0,0 +1,144 @@
...
+    private static int _counter;
+
+    private readonly byte[] _idBytes;
+    private readonly long _timestamp;
+    private readonly int _checkpointId;
+
+    internal DataStreamsTransactionInfo(string id, long timestamp, string checkpoint)
+    {
+        var encoded = Encoding.UTF8.GetBytes(id);
+        _idBytes = Truncate(encoded);
```
andrewlock: This is relatively expensive and a bit wasteful 🤔 There are more performant APIs you can use to (for example) `stackalloc` the byte[] if you're just going to truncate it anyway.

Additionally, consider _not_ making this a `readonly struct`, and using `ArrayPool<byte>.Shared` for renting and returning arrays, as long as you have strict control over the lifetime of these objects.

(Using `ArrayPool` probably isn't a good idea for structs, because it's _really_ hard to avoid copying and accidentally double-freeing an array 😅)

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsTransactionInfo.cs
```diff
@@ -0,0 +1,144 @@
...
+        _timestamp = timestamp;
+        _checkpointId = Cache.GetOrAdd(checkpoint, _ => Interlocked.Increment(ref _counter));
+    }
+
+    internal DataStreamsTransactionInfo(byte[] idBytes, long timestamp, string checkpoint)
+    {
+        _idBytes = Truncate(idBytes);
```
andrewlock, line 61: Can we not make sure we don't _generate_ the `idBytes` with too large a size?

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsTransactionInfo.cs
```diff
@@ -0,0 +1,144 @@
...
+        _timestamp = timestamp;
+        _checkpointId = Cache.GetOrAdd(checkpoint, _ => Interlocked.Increment(ref _counter));
+    }
+
+    internal long TimestampNs { get => _timestamp; }
```
andrewlock:
```suggestion
    internal long TimestampNs => _timestamp;
```

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsTransactionInfo.cs
```diff
@@ -0,0 +1,144 @@
...
+
+    internal string TransactionId { get => Encoding.UTF8.GetString(_idBytes); }
```
andrewlock: Do you _need_ this? It's expensive and allocatey, doesn't actually match the Transaction Id provided, and doesn't appear to be used 

```suggestion
```

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsTransactionInfo.cs
```diff
@@ -0,0 +1,144 @@
...
+            index += keyBytes.Length;
+        }
+
+        var trimmed = new byte[index];
+        Array.Copy(result, trimmed, index);
+        return trimmed;
+    }
+
+    // ClearCache is for using in tests only — resets both the map and the counter so IDs are deterministic
+    internal static void ClearCache()
```
andrewlock: If it's for testing only, we can add the attribute instead
```suggestion
    [InternalForTesting]
    internal static void ClearCache()
```

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsTransactionInfo.cs
```diff
@@ -0,0 +1,144 @@
...
+/// The checkpoint-name cache (<see cref="GetCacheBytes"/>) is serialized separately
+/// and sent alongside the transaction data so the receiver can resolve checkpoint IDs
+/// back to their string names. Each entry: <c>[1 byte ID] [1 byte name length] [N bytes name]</c>.
+/// Checkpoint IDs are assigned monotonically starting at 1 on first use.
+/// </summary>
+internal readonly struct DataStreamsTransactionInfo
+{
+    private const int MaxIdBytes = 255;
+
+    private static readonly ConcurrentDictionary<string, int> Cache = new();
```
andrewlock, line 45: This will grow unbounded as far as I can tell. Given there could be a huge number of transactions, should we cap the size, and make it FIFO or something? Unconstrained growth seems a bit problematic

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsTransactionInfo.cs
```diff
@@ -0,0 +1,144 @@
...
+        foreach (var pair in Cache)
+        {
+            var keyBytes = Encoding.UTF8.GetBytes(pair.Key);
+            // resize the buffer if needed
+            if (result.Length - index <= keyBytes.Length + 2)
+            {
+                var resized = new byte[result.Length * 2];
+                Array.Copy(result, 0, resized, 0, result.Length);
+                result = resized;
+            }
```
andrewlock: Yuk, there's _so_ many allocations here 😅 it feels like we should either pre-size this array and just do two loops (one to calculate sizes, one to assign), or we should be using an array pool. My preference is probably the former as it seems less error prone, but there's a race condition here in that case, so you would have to make sure to iterate the _same_ items (i.e. remember and skip the max keyValue I guess).

In fact... seeing as you never clear the cache, this is going to grow throughout the lifetime of the app which seems.... problematic 😟

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsExtractorRegistry.cs
```diff
@@ -0,0 +1,60 @@
...
+using Datadog.Trace.Logging;
+using Datadog.Trace.Util.Json;
+
+namespace Datadog.Trace.DataStreamsMonitoring.TransactionTracking;
+
+internal sealed class DataStreamsExtractorRegistry
+{
+    private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor<DataStreamsExtractorRegistry>();
+
+    private readonly Dictionary<DataStreamsTransactionExtractor.Type, List<DataStreamsTransactionExtractor>> _extractors = new();
```
andrewlock: This allocates unnecessarily if there are no extractors, I suggest moving it to the constructor.

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsExtractorRegistry.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        if (deserialized == null)
+        {
+            return;
+        }
+
+        foreach (var extractor in deserialized)
+        {
+            var list = _extractors.GetValueOrDefault(extractor.ExtractorType) ?? new();
```
andrewlock: You need to check `extractor` for null, because `[null, null, null]` is valid JSON here 😅

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsExtractorRegistry.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            deserialized = JsonHelper.DeserializeObject<List<DataStreamsTransactionExtractor>>(extractorsJson);
+        }
+        catch (Exception ex)
+        {
+            Log.Warning(ex, "Failed to parse DD_DATA_STREAMS_TRANSACTION_EXTRACTORS value. Transaction tracking extractors will be disabled.");
+            return;
+        }
+
+        if (deserialized == null)
```
andrewlock:
```suggestion
        if (deserialized == null || deserialized.Length == 0)
```

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -585,6 +585,10 @@ not null when string.Equals(value, "otlp", StringComparison.OrdinalIgnoreCase) =
                                                     .WithKeys(ConfigurationKeys.DataStreamsMonitoring.Enabled)
                                                     .AsBool() == null;
 
+            DataStreamsTransactionExtractors = config
+                                                  .WithKeys(ConfigurationKeys.DataStreamsMonitoring.TransactionExtractors)
```
andrewlock, line 590: We should do the JSON parsing here, as part of config extraction, so that errors are reported to telemetry correctly.

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsExtractorRegistry.cs
```diff
@@ -0,0 +1,60 @@
...
+    {
+        if (string.IsNullOrWhiteSpace(extractorsJson))
+        {
+            return;
+        }
+
+        List<DataStreamsTransactionExtractor>? deserialized;
+        try
+        {
+            deserialized = JsonHelper.DeserializeObject<List<DataStreamsTransactionExtractor>>(extractorsJson);
```
andrewlock: This seems like a very complicated and error prone format we're getting customers to need to use, have we considered a simpler public configuration API for this 😟

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsTransactionExtractor.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        ["HTTP_OUT_HEADERS"] = Type.HttpOutHeaders,
+        ["HTTP_IN_HEADERS"] = Type.HttpInHeaders,
+        ["KAFKA_CONSUME_HEADERS"] = Type.KafkaConsumeHeaders,
+        ["KAFKA_PRODUCE_HEADERS"] = Type.KafkaProduceHeaders,
+    };
+
+    private Type? _cachedType;
+
+    public enum Type
```
andrewlock: Please don't call it `Type`, it's too common an API and will clash with `System` namespace
```suggestion
    public enum ExtractorType
```

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsTransactionExtractor.cs
```diff
@@ -0,0 +1,58 @@
...
+    [JsonProperty(PropertyName = "value")]
+    public string Value { get; private set; } = string.Empty;
+
+    public Type ExtractorType
+    {
+        get
+        {
+            if (_cachedType is null)
+            {
+                _cachedType = TypeMap.TryGetValue(StringType, out var t) ? t : Type.Unknown;
```
andrewlock: nit: just use a switch expression and ditch the `TypeMap`. I'm not really sure you need the cachedType tbh:

```suggestion
                _cachedType = StringType switch
                {
                    "HTTP_OUT_HEADERS" => Type.HttpOutHeaders,
                    "HTTP_IN_HEADERS" => Type.HttpInHeaders,
                    "KAFKA_CONSUME_HEADERS" => Type.KafkaConsumeHeaders,
                    "KAFKA_PRODUCE_HEADERS" => Type.KafkaProduceHeaders,
                    _ => Type.Unknown,
                };
```

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsExtractorRegistry.cs
kr-igor: This format is flexible, since we don't know what other extractors we may need later. The intention was to use Remote Config in the future. For now we help customers onboard.

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
kr-igor: Good catch, thanks!

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsTransactionInfo.cs
kr-igor: I can't find `[InternalForTesting]`, so i renamed the method to `ClearCacheForTesting`

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsTransactionInfo.cs
kr-igor: It's OK for the checkpoint name to be longer than 255 characters. This limitation is in place only for transaction ids.

tracer/test/Datadog.Trace.Tests/DataStreamsMonitoring/DataStreamsTransactionExtractorTests.cs
kr-igor: Good catch, I used it for debugging, but forgot to cleanup

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Http/HttpClient/HttpMessageHandlerCommon.cs
kr-igor: `IsTransactionTrackingEnabled` now checks for default state.

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
kr-igor: Removing the extractors disables Transaction Tracking, adding one more option can be confusing.

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
kr-igor: The public `IsInDefaultState` was introduced in previous PRs and I don't want to do the refactoring in this PR. Happy to work on a follow up.

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsTransactionInfo.cs
kr-igor: This method is used for kafka, where header values are bytes. We use truncate to cur off extra bytes if the array is longer than the limit. In practice (from the data we've seen from Java customers) this never happens, as transaction ids are typically uuids.

tracer/src/Datadog.Trace/DataStreamsMonitoring/SpanExtensions.cs
kr-igor: Good catch, it should be used when there's an active span - we want to add transaction ids as span attributes.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Http/HttpClient/HttpMessageHandlerCommon.cs
kr-igor: It's not dependent, but we need to account for active spans for linking purpose.

tracer/src/Datadog.Trace/DataStreamsMonitoring/Aggregation/DataStreamsMessagePackFormatter.cs
```diff
@@ -146,11 +155,24 @@ public int Serialize(Stream stream, long bucketDurationNs, List<SerializableStat
             bytesWritten += MessagePackBinary.WriteStringBytes(stream, _tracerVersionValueBytes);
 
             bytesWritten += MessagePackBinary.WriteStringBytes(stream, _statsBytes);
-            bytesWritten += MessagePackBinary.WriteArrayHeader(stream, statsBuckets.Count + backlogsBuckets.Count);
+            bytesWritten += MessagePackBinary.WriteArrayHeader(stream, statsBuckets.Count + backlogsBuckets.Count + (hasTransactions ? 1 : 0));
+
+            if (hasTransactions)
+            {
+                var currentTs = DateTimeOffset.UtcNow.ToUnixTimeNanoseconds();
+                var bucketStartTime = currentTs - (currentTs % bucketDurationNs);
```
andrewlock, line 163: I don't know if it matters, but this seems like it has the potential to be essentially "wrong" if there are delays processing? Should this be passed into the `Serialize` method and tracked externally, when the data is actually recorded/snapshotted?

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsTransactionExtractor.cs
```diff
@@ -0,0 +1,72 @@
...
+                    "KAFKA_PRODUCE_HEADERS" => ExtractorType.KafkaProduceHeaders,
+                    _ => ExtractorType.Unknown,
+                };
+            }
+
+            return _cachedType.Value;
+        }
+    }
+
+    public static IReadOnlyList<DataStreamsTransactionExtractor> ParseList(string json)
```
andrewlock: nit: return the concrete type when you know it(it's better for perf)
```suggestion
    public static List<DataStreamsTransactionExtractor> ParseList(string json)
```

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsTransactionExtractor.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        Unknown,
+
+        HttpOutHeaders,
+
+        HttpInHeaders,
+
+        KafkaConsumeHeaders,
+
+        KafkaProduceHeaders,
```
andrewlock, line 23: nit spacing:
```suggestion
        Unknown,
        HttpOutHeaders,
        HttpInHeaders,
        KafkaConsumeHeaders,
        KafkaProduceHeaders,
```

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -585,6 +586,13 @@ not null when string.Equals(value, "otlp", StringComparison.OrdinalIgnoreCase) =
...
+                                                  .GetAs(
+                                                       defaultValue: new DefaultResult<IReadOnlyList<DataStreamsTransactionExtractor>>([], "[]"),
+                                                       converter: json => ParsingResult<IReadOnlyList<DataStreamsTransactionExtractor>>.Success(DataStreamsTransactionExtractor.ParseList(json)),
```
andrewlock: nit: use the concrete type when we know it (better for perf)
```suggestion
                                                       defaultValue: new DefaultResult<List<DataStreamsTransactionExtractor>>([], "[]"),
                                                       converter: json => ParsingResult<List<DataStreamsTransactionExtractor>>.Success(DataStreamsTransactionExtractor.ParseList(json)),
```

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -1208,6 +1216,11 @@ not null when string.Equals(value, "otlp", StringComparison.OrdinalIgnoreCase) =
         /// </summary>
         internal bool IsDataStreamsMonitoringInDefaultState { get; }
 
+        /// <summary>
+        /// Gets a raw value for DSM extractors
+        /// </summary>
+        internal IReadOnlyList<DataStreamsTransactionExtractor> DataStreamsTransactionExtractors { get; }
```
```suggestion
        internal List<DataStreamsTransactionExtractor> DataStreamsTransactionExtractors { get; }
```

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsExtractorRegistry.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Collections.Generic;
+using Datadog.Trace.Util.Json;
+
+namespace Datadog.Trace.DataStreamsMonitoring.TransactionTracking;
+
+internal sealed class DataStreamsExtractorRegistry
+{
+    private readonly Dictionary<DataStreamsTransactionExtractor.ExtractorType, List<DataStreamsTransactionExtractor>> _extractors;
+
+    internal DataStreamsExtractorRegistry(IReadOnlyList<DataStreamsTransactionExtractor> extractors)
```
andrewlock:
```suggestion
    internal DataStreamsExtractorRegistry(List<DataStreamsTransactionExtractor> extractors)
```

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsExtractorRegistry.cs
```diff
@@ -0,0 +1,36 @@
...
+    {
+        _extractors = new(extractors.Count);
+        foreach (var extractor in extractors)
+        {
+            var list = _extractors.GetValueOrDefault(extractor.ParsedType) ?? new();
+            list.Add(extractor);
+            _extractors[extractor.ParsedType] = list;
```
andrewlock: nit: tiny perf improvement (avoid setting if it already exists)
```suggestion
            if (!_extractors.TryGetValue(key, out var list)
            {
                list = new();
                _extractors[extractor.ParsedType] = list;
            }

            list.Add(extractor);
```

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsTransactionContainer.cs
```diff
@@ -0,0 +1,65 @@
...
+    internal bool ShouldFlush => _size >= MaxSizeBytes;
+
+    public void Add(DataStreamsTransactionInfo transactionInfo)
+    {
+        var byteCount = transactionInfo.GetByteCount();
+
+        if (_data.Length - _size < byteCount)
+        {
+            var resized = new byte[Math.Max(_data.Length * 2, _size + byteCount)];
+            Array.Copy(_data, 0, resized, 0, _size);
```
andrewlock, line 46: Theoretically this could keep growing indefinitely... Is that something we _want_ to happen? Should we consider capping this? How should we handle it in that case?

Currently it seems like we are always trying to keep all transactions, but should we account for cases where that's just too expensive in the design some how? 🤔

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsTransactionContainer.cs
```diff
@@ -0,0 +1,65 @@
...
+
+    public byte[] GetDataAndReset()
+    {
+        if (_size == 0)
+        {
+            return [];
+        }
+
+        // trim zeros
+        var result = new byte[_size];
```
andrewlock, line 65: Just to clarify:

> This class is NOT thread-safe

So we _won't_ ever be calling `GetDataAndReset()` at the same time as `Add()` or anything else, correct?

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsTransactionInfo.cs
```diff
@@ -0,0 +1,108 @@
...
+        buffer[offset + 2] = (byte)(_timestamp >> 48);
+        buffer[offset + 3] = (byte)(_timestamp >> 40);
+        buffer[offset + 4] = (byte)(_timestamp >> 32);
+        buffer[offset + 5] = (byte)(_timestamp >> 24);
+        buffer[offset + 6] = (byte)(_timestamp >> 16);
+        buffer[offset + 7] = (byte)(_timestamp >> 8);
+        buffer[offset + 8] = (byte)_timestamp;
+
+        // id size, up to 256 bytes
+        buffer[offset + 9] = (byte)_idBytes.Length;
```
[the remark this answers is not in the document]
andrewlock: I _think_ this is still applicable, i.e. as a question of correctness? 🤔

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
andrewlock: Oh, right, I misread, never mind 😄

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
```diff
@@ -64,9 +73,11 @@ public bool IsEnabled
...
     }
 
-    public bool IsInDefaultState
+    public bool IsInDefaultState => _isInDefaultState;
+
+    public bool IsTransactionTrackingEnabled
     {
-        get => Volatile.Read(ref _isInDefaultState);
+        get => !_isInDefaultState && IsEnabled;
     }
```
andrewlock: nit:
```suggestion
    public bool IsTransactionTrackingEnabled => !_isInDefaultState && IsEnabled;
```

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
andrewlock: Yeah, this looks like a bug to me?

tracer/test/Datadog.Trace.Tests/Telemetry/config_norm_rules.json
```diff
@@ -379,6 +379,7 @@
   "DD_TRACE_X_DATADOG_TAGS_MAX_LENGTH": "trace_x_datadog_tags_max_length",
   "DD_DATA_STREAMS_ENABLED": "data_streams_enabled",
   "DD_DATA_STREAMS_LEGACY_HEADERS": "data_streams_legacy_headers",
+  "DD_DATA_STREAMS_TRANSACTION_EXTRACTORS": "data_streams_transaction_extractors",
```
andrewlock, line 382: Don't forget to [add this here](https://github.com/DataDog/dd-go/blob/prod/trace/apps/tracer-telemetry-intake/telemetry-payload/static/config_norm_rules.json) too 🙂

tracer/src/Datadog.Trace/DataStreamsMonitoring/Aggregation/DataStreamsMessagePackFormatter.cs
kr-igor: We carry the timestamp with each transaction, so the actual bucket is not important. It is used only as a way of splitting the data into chunks.

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsTransactionContainer.cs
kr-igor: We will force the flush if the container size grows above 512 * 1024 bytes ([here](https://github.com/DataDog/dd-trace-dotnet/blob/eb1c7923aa7dbd08d1bd62e151a8d7d5ccd0ec63/tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs#L378)). So in practice it should not go much higher than that, but I will add a hard cap here too.

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsTransactionContainer.cs
kr-igor: Yes, that's correct. For now `Add` and `GetDataAndReset` are guarded by `_flushSemaphore` in `DataStreamsWriter`

tracer/src/Datadog.Trace/DataStreamsMonitoring/TransactionTracking/DataStreamsTransactionInfo.cs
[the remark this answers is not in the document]
kr-igor: True, fixed now.

tracer/test/Datadog.Trace.Tests/Telemetry/config_norm_rules.json
kr-igor: Didn't know that we need to keep these in sync. Will create a separate PR, thank you!

kr-igor: It's already added [here](https://github.com/DataDog/dd-go/blob/prod/trace/apps/tracer-telemetry-intake/telemetry-payload/static/registry_config_rules.json#L126).

PR #7953, opened 2025-12-15
Fix broken release scripts

bouwkast:
## Summary of changes

These were removed from the reusable workflow in https://github.com/DataDog/dd-trace-dotnet/pull/7828/, but didn't get removed from these. 

## Reason for change

They were broken and GitHub said:

```
The workflow is not valid. .github/workflows/create_normal_draft_release.yml (Line: 40, Col: 33): Invalid secret, DD_PREPROD_API_KEY is not defined in the referenced workflow. .github/workflows/create_normal_draft_release.yml (Line: 41, Col: 39): Invalid secret, DD_PUBLIC_SYMBOL_API_KEY is not defined in the referenced workflow.
```

## Implementation details

Deleted them

## Test coverage

Can try them out next time 😄 

## Other details

We still have our old release script which doesn't have the issue


andrewlock (approved):
Oops, thanks 👍

PR #7956, opened 2025-12-16
[Profiler] Add managed code cache

gleocadie:
## Summary of changes

Introduce Managed Code cache component.

## Reason for change

A managed code cache is a cache that holds managed (code) functions code ranges. The goal is two-fold:
- Prevent crashing: we have crash reports where crashes in the profiler happen when calling `GetFunctionFromIP`. This can occur when querying a function Id for an instruction pointer and we race against a deallocation.
- Contributing (in another PR) to ARM64 unwinding.

## Implementation details

- The managed code cache listens to jitted functions and R2R modules loading and caches functions code ranges.
- In case of R2R modules, we can cache the module code ranges by reading the PE file. When querying for a function id, if the instruction pointer belongs to a R2R module, we call into `GetFunctionFromIP` to enrich our cache.
- For now this managed code cache is not used by default but can be enabled by setting the env var `DD_INTERNAL_PROFILING_USE_MANAGED_CODE_CACHE=1`. We will ask customer to set this env var and if crash reports related to `GetFunctionFromIP` are less frequent, we will make it default. 

## Test coverage

- Add a test to make sure that an exception callstack is still good.

### DIFF
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
@@ -1336,6 +1336,14 @@ HRESULT STDMETHODCALLTYPE CorProfilerCallback::Initialize(IUnknown* corProfilerI
 {
     Log::Info("CorProfilerCallback is initializing.");
 
+    #if ARM64
+    if (!_pConfiguration->UseManagedCodeCache())
+    {
+        Log::Warn("Managed code cache is required to run the profiler on ARM64. Since it is not enabled, the profiler will not run.");
+        return E_FAIL;
+    }
+    #endif
+
     ConfigureDebugLog();
 
     _pMetadataProvider = std::make_unique<MetadataProvider>();
--- a/profiler/src/ProfilerEngine/Datadog.Profiler.Native/Configuration.cpp
+++ b/profiler/src/ProfilerEngine/Datadog.Profiler.Native/Configuration.cpp
@@ -124,7 +124,13 @@ Configuration::Configuration()
     _heapSnapshotCheckInterval = ExtractHeapSnapshotCheckInterval();
     _heapSnapshotMemoryPressureThreshold = GetEnvironmentValue(EnvironmentVariables::HeapSnapshotMemoryPressureThreshold, 85);
     _heapHandleLimit = ExtractHeapHandleLimit();
-    _useManagedCodeCache = GetEnvironmentValue(EnvironmentVariables::UseManagedCodeCache, false);
+    bool defaultUseManagedCodeCache =
+    #if ARM64
+        true;
+    #else
+        false;
+    #endif
+    _useManagedCodeCache = GetEnvironmentValue(EnvironmentVariables::UseManagedCodeCache, defaultUseManagedCodeCache);
 }
 
 fs::path Configuration::ExtractLogDirectory()


profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
```diff
@@ -162,6 +163,9 @@ void CorProfilerCallback::InitializeServices()
 
     _pDebugInfoStore = std::make_unique<DebugInfoStore>(_pCorProfilerInfo, _pConfiguration.get());
 
+    _managedCodeCache = std::make_unique<ManagedCodeCache>(_pCorProfilerInfo, _pConfiguration.get());
```
chrisnas: If it is a service, why not registering it as a service; Start/Sop will be called automatically 
BTW, Stop() is never called in CorProfilerCallback

profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
```diff
@@ -1858,6 +1870,7 @@ HRESULT STDMETHODCALLTYPE CorProfilerCallback::ModuleUnloadStarted(ModuleID modu
 
 HRESULT STDMETHODCALLTYPE CorProfilerCallback::ModuleUnloadFinished(ModuleID moduleId, HRESULT hrStatus)
 {
+    _managedCodeCache->RemoveModule(moduleId);
```
chrisnas: would it be possible that we end up to the same race condition as GetFuntionFromIP? i.e. since we symbolize frames asynchronously, a module could have been removed from the cache when looking for an address in its range.

profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
```diff
@@ -1898,6 +1911,12 @@ HRESULT STDMETHODCALLTYPE CorProfilerCallback::JITCompilationStarted(FunctionID
 
 HRESULT STDMETHODCALLTYPE CorProfilerCallback::JITCompilationFinished(FunctionID functionId, HRESULT hrStatus, BOOL fIsSafeToBlock)
 {
+    if (FAILED(hrStatus))
+    {
+        Log::Debug("JITCompilationFinished failed for functionId=0x", std::hex, functionId, std::dec, ". hrStatus=", std::hex, hrStatus, std::dec);
```
chrisnas: is it too verbose even in debug?

profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
```diff
@@ -2415,6 +2434,13 @@ HRESULT STDMETHODCALLTYPE CorProfilerCallback::GetReJITParameters(ModuleID modul
 
 HRESULT STDMETHODCALLTYPE CorProfilerCallback::ReJITCompilationFinished(FunctionID functionId, ReJITID rejitId, HRESULT hrStatus, BOOL fIsSafeToBlock)
 {
+    if (FAILED(hrStatus))
+    {
+        Log::Debug("ReJITCompilationFinished failed for functionId=0x", std::hex, functionId, std::dec, " ReJit ID=0x", std::hex, rejitId, std::dec, ". hrStatus=", std::hex, hrStatus, std::dec);
```
chrisnas: is it too verbose even in debug?

profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
```diff
@@ -2415,6 +2434,13 @@ HRESULT STDMETHODCALLTYPE CorProfilerCallback::GetReJITParameters(ModuleID modul
...
+        return S_OK;
+    }
+    Log::Debug("ReJITCompilationStarted for functionId=0x", std::hex, functionId, std::dec, " ReJit ID=0x", std::hex, rejitId, std::dec);
```
chrisnas: is it too verbose even in debug?

profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
```diff
@@ -2455,6 +2481,12 @@ HRESULT STDMETHODCALLTYPE CorProfilerCallback::DynamicMethodJITCompilationStarte
 
 HRESULT STDMETHODCALLTYPE CorProfilerCallback::DynamicMethodJITCompilationFinished(FunctionID functionId, HRESULT hrStatus, BOOL fIsSafeToBlock)
 {
+    if (FAILED(hrStatus))
+    {
+        Log::Debug("DynamicMethodJITCompilationFinished failed for functionId=0x", std::hex, functionId, std::dec, ". hrStatus=", std::hex, hrStatus, std::dec);
```
chrisnas: is it too verbose even in debug?

profiler/src/ProfilerEngine/Datadog.Profiler.Native/EnvironmentVariables.h
```diff
@@ -84,4 +84,5 @@ class EnvironmentVariables final
     inline static const shared::WSTRING CIVisibilityEnabled         = WStr("DD_CIVISIBILITY_ENABLED");
     inline static const shared::WSTRING InternalCIVisibilitySpanId  = WStr("DD_INTERNAL_CIVISIBILITY_SPANID");
     inline static const shared::WSTRING WaitHandleProfilingEnabled  = WStr("DD_INTERNAL_PROFILING_WAITHANDLE_ENABLED");
+    inline static const shared::WSTRING UseCustomGetFunctionFromIP = WStr("DD_EXPERIMENTAL_USE_CUSTOM_GET_FUNCTION_FROM_IP");
```
chrisnas: maybe use INTERNAL instead of EXPERIMENTAL in the env var

profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
gleocadie: Actually the module code range is not removed from the cache, but marked as removed.

profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
gleocadie: you are right, could be. But it would be only when the compilation failed.

profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
gleocadie: same

profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
gleocadie: same

profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
gleocadie: same.
The thing is if one of them fails, how would I be able to debug it ?

profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
gleocadie: In case of SSI, I would like the cache to be filled with the code ranges

gleocadie: Since I want the instance of ManagedCodeCache to live for the whole process lifetime, I removed it as a Service and replace Start by Initialize.
The destructor will take care of the clean up

profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
gleocadie: 🤔 more about it, I wonder if we actually need this:
if it fails, we won't register the code because there is no code range. The log message would not help at all.
we can remove it I would say, what do you think ?

profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
```diff
@@ -162,6 +163,9 @@ void CorProfilerCallback::InitializeServices()
 
     _pDebugInfoStore = std::make_unique<DebugInfoStore>(_pCorProfilerInfo, _pConfiguration.get());
 
+    _managedCodeCache = std::make_unique<ManagedCodeCache>(_pCorProfilerInfo, _pConfiguration.get());
+    _managedCodeCache->Initialize();
```
chrisnas: If `Initialize` does not return a bool, move the initialization into the constructor.
If there are situation where the initialization could fail, make it return true/false and log the error : it will help in investigations  ;^)

profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
chrisnas: A log here would not help...

profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
chrisnas: if this fails, well... we would be in big trouble anyway

profiler/src/ProfilerEngine/Datadog.Profiler.Native/FrameStore.cpp
```diff
@@ -91,71 +63,17 @@ std::pair<bool, FrameInfoView> FrameStore::GetFrame(uintptr_t instructionPointer
...
 
-    std::optional<std::pair<HRESULT, FunctionID>> result = GetFunctionFromIP(instructionPointer);
+    std::optional<FunctionID> functionId = _pManagedCodeCache->GetFunctionId(instructionPointer);
 
-    if (!result.has_value())
+    if (!functionId.has_value())
     {
-        // we still want a frame to display a good'ish callstack shape
-        return {true, {UnloadedModuleName, NotResolvedFrame, "", 0}};
+        return {false, {NotResolvedModuleName, NotResolvedFrame, "", 0}};
```
chrisnas: Is it expected to now return false instead of true before?

profiler/src/ProfilerEngine/Datadog.Profiler.Native/ManagedCodeCache.h
```diff
@@ -0,0 +1,185 @@
...
+    void WorkerThread(std::promise<void> startPromise);
+    
+    // Helper: Ensure a page exists in the map
+    void EnsurePageExists(uint64_t page);
+    
+    // Map from page number -> page entry (with its own lock)
+
+
+    std::unordered_map<uint64_t, PageEntry> _pagesMap;
+    std::unordered_set<ModuleID> _modules;
```
chrisnas: I'm not sure this is used

profiler/src/ProfilerEngine/Datadog.Profiler.Native/ManagedCodeCache.cpp
```diff
@@ -0,0 +1,613 @@
...
+
+    // Level 2: Check if the IP is within a module code range
+    
+    if (IsCodeInR2RModule(ip))
+    {
+        auto functionId = GetFunctionFromIP_Original(ip);
+        if (functionId.has_value()) {
+            // in that case no need to defer this action to another thread
+            // if we let another thread do this, we can remove the constraint on managed threads
+            AddFunctionImpl(*functionId, false);
```
chrisnas: prefer .value instead of leveraging the * overload

profiler/src/ProfilerEngine/Datadog.Profiler.Native/ManagedCodeCache.cpp
```diff
@@ -0,0 +1,613 @@
...
+        return info;
+    }
+
+    // Level 2: Check if the IP is within a module code range
+    
+    if (IsCodeInR2RModule(ip))
+    {
+        auto functionId = GetFunctionFromIP_Original(ip);
+        if (functionId.has_value()) {
+            // in that case no need to defer this action to another thread
```
chrisnas: provide more explanation because not obvious

profiler/src/ProfilerEngine/Datadog.Profiler.Native/ManagedCodeCache.cpp
```diff
@@ -0,0 +1,613 @@
...
+    return future.wait_for(2s) == std::future_status::ready;
+}
+
+bool ManagedCodeCache::IsCodeInR2RModule(std::uintptr_t ip) const noexcept
+{
+    std::shared_lock<std::shared_mutex> moduleLock(_modulesMutex);
+    
+    // Use lower_bound to find the first range where startAddress > ip
+    // The range we're looking for (if it exists) must be the PREVIOUS range
+    auto moduleIt = std::lower_bound(
```
chrisnas: would it be possible to have a wrapper with a more explicit name because it is not obvious that all this is a binary search?

profiler/src/ProfilerEngine/Datadog.Profiler.Native/ManagedCodeCache.cpp
```diff
@@ -0,0 +1,613 @@
...
+    if (moduleIt == _modulesCodeRanges.begin())
+    {
+        return false;
+    }
+    
+    --moduleIt;
+    
+    if (moduleIt->isRemoved)
+    {
+        LogOnce(Debug, "ManagedCodeCache::IsCodeInR2RModule: Module code range is removed for ip: 0x", std::hex, ip);
```
chrisnas: update the text because it could be confusing: we are not removing a a code range here

profiler/src/ProfilerEngine/Datadog.Profiler.Native/ManagedCodeCache.h
```diff
@@ -0,0 +1,185 @@
...
+    }
+    
+    // Helper: Find range in sorted vector using binary search (signal-safe)
+    static const CodeRange* FindRangeInVector(
+        const std::vector<CodeRange>& ranges,
+        UINT_PTR ip) noexcept;
+    
+    // Query the runtime for code ranges for a specific version
+    // This is called when a new tier is compiled
+    std::vector<CodeRange> QueryCodeRanges(FunctionID functionId);
```
chrisnas: For consistency sake, prefer `Get` to `Query`

profiler/src/ProfilerEngine/Datadog.Profiler.Native/ManagedCodeCache.cpp
```diff
@@ -0,0 +1,613 @@
...
+}
+
+// This function must be called by the worker thread only
+void ManagedCodeCache::AddFunctionRangesToCache(std::vector<CodeRange> newRanges)
+{    
+    // Step 1: Compute affected pages
+    std::set<uint64_t> affectedPages;
+    for (const auto& range : newRanges)
+    {
+        uint64_t startPage = GetPageNumber(range.startAddress);
```
chrisnas, line 344: add a comment to explain that a method could span 2 pages so it is needed to add the code range in both PageEntry

profiler/src/ProfilerEngine/Datadog.Profiler.Native/ManagedCodeCache.cpp
```diff
@@ -0,0 +1,613 @@
...
+        uint64_t endPage = GetPageNumber(range.endAddress - 1);
```
chrisnas: check if -1 is needed

```
PE1     PE2
|------ ---|
     __ _

```
in that specific case, I would like to add the range to Page Entry 2
must be in synch with GetPageNumber (i.e. GetPageNumber(64k + 1) should return PE2

profiler/src/ProfilerEngine/Datadog.Profiler.Native/ManagedCodeCache.cpp
```diff
@@ -0,0 +1,613 @@
...
+        for (uint64_t p = startPage; p <= endPage; ++p) {
+            affectedPages.insert(p);
```
chrisnas: It should be possible to  optimize the number of locks taken for the _fast path_ (i.e. most of the time, the pages will be here).
Today, with the set:
- if the page does not exist: reader + writer + reader
- if the page exists: reader + reader

The code in EnsurePageExists takes the reader and if the page exists, return. The same reader will be taken again in line 376. The writer is taken to add a page. It means that the code starting at line 386 could be moved in an helper function taking the page to update with the given ranges. This helper function would be called (1) if the page exists (under the reader) and (2) after a new page is added (i.e.) under the writer lock.

In term of performance, (2) might keep the writer longer than today but it will only happen a few time when the cache is filled up. Instead, most of the time (after the cache is filled), the reader will be taken only once.

profiler/src/ProfilerEngine/Datadog.Profiler.Native/ManagedCodeCache.cpp
```diff
@@ -0,0 +1,613 @@
...
+        : work(std::move(w)), next(nullptr) {}
+};
+
+std::unique_ptr<ManagedCodeCache::QueueNode> ManagedCodeCache::DequeueWorkItem()
+{
+    auto* currentHead = _workerQueueHead.load(std::memory_order_acquire);
+    while (currentHead != nullptr)
+    {
+        auto* next = currentHead->next;
+        if (_workerQueueHead.compare_exchange_weak(
```
chrisnas: maybe get the whole queue with 1 c`ompare_exchange(currentHeap, nullptr,...`) and then iterate without any lock on each workitem in the (former) queue (that is now empty): this avoid 1 `compare_exchange` per workitem.

This would change the code in `ManagedCodeCache::WorkerThread`: use a lock to (1) get the head and (2) replace the head with nullptr before iterating on the former queue

profiler/src/ProfilerEngine/Datadog.Profiler.Native.Windows/AutoResetEvent.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+        case WAIT_TIMEOUT:
+            // Timeout occurred
+            return false;
+
+        case WAIT_FAILED:
+            Log::Debug("AutoResetEvent::Wait: WaitForSingleObject failed. Error: ", GetLastError());
+            return false;
+
+        case WAIT_ABANDONED:
```
chrisnas: could be removed because, as mentioned in the comment, for mutex only

profiler/src/ProfilerEngine/Datadog.Profiler.Native.Windows/AutoResetEvent.cpp
```diff
@@ -0,0 +1,110 @@
...
+            // Should not happen with events (only with mutexes)
+            Log::Debug("AutoResetEvent::Wait: Unexpected WAIT_ABANDONED");
+            return false;
+
+        default:
+            Log::Debug("AutoResetEvent::Wait: Unexpected result: ", result);
```
chrisnas: logonce

profiler/src/ProfilerEngine/Datadog.Profiler.Native.Windows/AutoResetEvent.cpp
```diff
@@ -0,0 +1,110 @@
...
+    {
+        case WAIT_OBJECT_0:
+            return true;
+
+        case WAIT_TIMEOUT:
+            // Timeout occurred
+            return false;
+
+        case WAIT_FAILED:
+            Log::Debug("AutoResetEvent::Wait: WaitForSingleObject failed. Error: ", GetLastError());
```
chrisnas: logonce

profiler/src/ProfilerEngine/Datadog.Profiler.Native/FrameStore.cpp
gleocadie: changed

profiler/src/ProfilerEngine/Datadog.Profiler.Native/ManagedCodeCache.cpp
gleocadie: done

profiler/src/ProfilerEngine/Datadog.Profiler.Native/ManagedCodeCache.h
```diff
@@ -0,0 +1,181 @@
...
+    void InsertCodeRangeIntoPage(PagesMap::iterator pageIt, const CodeRange& range);
+
+    void WorkerThread(std::promise<void> startPromise);
+    
+    // Helper: Ensure a page exists in the map
+    void EnsurePageExists(uint64_t page);
+    
+    // Map from page number -> page entry (with its own lock)
+
+
```
chrisnas, line 150: remove these 2 empty lines

PR #7957, opened 2025-12-16
Revert "Restore null scope check to stop EventBridge injection"

bouwkast:
## Summary

Reverting #7919 as we _should_ inject independently of whether there is a span/scope active.
Will need to follow up on the issue noted.

## Reason for change

If the instrumentation is disabled we should still be injecting context / baggage as they are independent of eachother.


lucaspimentel (approved):

lucaspimentel (commented):
I changed my mind. I don't think we need to revert this, although a little cleanup would be nice (see Andrew's comment).

bouwkast:
Closing this, after some discussion the earlier PR is essentially a temporary workaround for the time being, but it doesn't resolve the underlying issue where the headers appear to be placed (for some certain message types) in the wrong location. We will need to figure out how to address that, but any changes will need to be synced amongst other languages / tracers as if we move them we need to be consistent between tracers.

PR #7962, opened 2025-12-17
Add a `ValueStringBuilder` for .NET 6

andrewlock:
## Summary of changes

Adds a `ValueStringBuilder` implementation, based on the one used internally in the .NET runtime

## Reason for change

- We can stackalloc the `Span<char>`
- It's a bit faster than our existing `StringBuilderCache` implementation, and in some places that matters.

## Implementation details

- Made in .NET 6 only for simplicity. We _could_ expose it earlier, but I wanted this for the updated aspnetcore integration, so .NET6+ only for now is good enough
- It's not without risks in its usage, so we have to be careful about things like passing it around (i.e. avoid doing that completely, for safety)
- Uses an array pool backed implementation (again, built into .NET 6)

## Test coverage

Imported the unit tests from the runtime too

## Other details

Part of a stack

- https://github.com/DataDog/dd-trace-dotnet/pull/7962 👈


lucaspimentel:
I thought we had this already, but it was me in a personal project :sweat_smile:
https://github.com/lucaspimentel/pwsh-prompt/blob/main/src/pwsh-prompt/ValueStringBuilder.cs

I also had a TODO for myself from a long time ago to do this in `dd-trace-dotnet`, so thanks!

PR #7963, opened 2025-12-17
Small perf tweaks for `AspNetCoreDiagnosticObserver`

andrewlock:
## Summary of changes

Small perf tweaks for existing `AspNetCoreDiagnosticObserver`

## Reason for change

While experimenting with single span observer, noticed some (safe) tweaks we could make to the existing implementation. There are some other improvements we could make as well, but those are a little riskier, so may be best to delay them.

## Implementation details

- Use `HttpContext.Items[string]` instead of `HttpContext.Features.Get<T>` as it's a bit faster
- Make `Tracer`/`Security`/`Iast` inputs readonly, instead of using a null-coalesce with every access
- Move check outside of `AddHeaderTagsToSpan` given common case is to avoid the function call

## Test coverage

Covered by existing tests

## Other details

Part of a stack

- https://github.com/DataDog/dd-trace-dotnet/pull/7963 👈

### DIFF
--- a/tracer/src/Datadog.Trace/PlatformHelpers/AspNetCoreHttpRequestHandler.cs
+++ b/tracer/src/Datadog.Trace/PlatformHelpers/AspNetCoreHttpRequestHandler.cs
@@ -123,7 +123,7 @@ namespace Datadog.Trace.PlatformHelpers
             scope.Span.DecorateWebServerSpan(resourceName, httpMethod, host, url, userAgent, tags);
 
             var headerTagsInternal = tracer.CurrentTraceSettings.Settings.HeaderTags;
-            if (!headerTagsInternal.IsNullOrEmpty())
+            if (headerTagsInternal.Count != 0)
             {
                 AddHeaderTagsToSpan(scope.Span, request, tracer, headerTagsInternal);
             }


tracer/src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
```diff
@@ -421,10 +412,8 @@ private static Span StartMvcCoreSpan(
 
         private void OnHostingHttpRequestInStart(object arg)
         {
-            var tracer = CurrentTracer;
-            var security = CurrentSecurity;
-            var shouldTrace = tracer.CurrentTraceSettings.Settings.IsIntegrationEnabled(IntegrationId);
-            var shouldSecure = security.AppsecEnabled;
+            var shouldTrace = _tracer.CurrentTraceSettings.Settings.IsIntegrationEnabled(IntegrationId);
+            var shouldSecure = _security.AppsecEnabled;
```
lucaspimentel: While you're here, can we rename these "should" variables?
Since you asked, I'd prefer `integrationEnabled` and `appsecEnabled` 😁

tracer/src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
```diff
@@ -602,11 +589,9 @@ private void OnRoutingEndpointMatched(object arg)
...
         private void OnMvcBeforeAction(object arg)
         {
-            var tracer = CurrentTracer;
-            var security = CurrentSecurity;
-            var shouldTrace = tracer.CurrentTraceSettings.Settings.IsIntegrationEnabled(IntegrationId);
-            var shouldSecure = security.AppsecEnabled;
-            var shouldUseIast = CurrentIast.Settings.Enabled;
+            var shouldTrace = _tracer.CurrentTraceSettings.Settings.IsIntegrationEnabled(IntegrationId);
+            var shouldSecure = _security.AppsecEnabled;
+            var shouldUseIast = _iast.Settings.Enabled;
```
lucaspimentel: ... and `iastEnabled` (see my other comment)

tracer/src/Datadog.Trace/PlatformHelpers/AspNetCoreHttpRequestHandler.cs
```diff
@@ -124,7 +121,13 @@ public Scope StartAspNetCorePipelineScope(Tracer tracer, Security security, Iast
 
             var scope = tracer.StartActiveInternal(_requestInOperationName, extractedContext.SpanContext, tags: tags, links: extractedContext.Links);
             scope.Span.DecorateWebServerSpan(resourceName, httpMethod, host, url, userAgent, tags);
-            AddHeaderTagsToSpan(scope.Span, request, tracer);
+
+            var headerTagsInternal = tracer.CurrentTraceSettings.Settings.HeaderTags;
+            if (!headerTagsInternal.IsNullOrEmpty())
```
lucaspimentel: eww, calling an extension method to check for null 😝

lucaspimentel (approved):

tracer/src/Datadog.Trace/PlatformHelpers/AspNetCoreHttpRequestHandler.cs
andrewlock: Nuffin' todo wiv me, guv. I just inlined the method 😝 

But yeah, fair point, fixed, thanks! 😄

tracer/src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
andrewlock: Agreed, fixed! And renamed them in the duplicate observer in #7964

tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
```diff
@@ -486,7 +486,11 @@ private static void StartDiagnosticManager()
             }
             else
             {
-                observers.Add(new AspNetCoreDiagnosticObserver());
+                // Tracer, Security, should both have been initialized by now.
+                // Iast hasn't yet, but doing it now is fine
+                // span origins is _not_ initialized yet, and we can't guarantee it will be
+                // so just be lazy instead
+                observers.Add(new AspNetCoreDiagnosticObserver(Tracer.Instance, Security.Instance, Iast.Iast.Instance, spanCodeOrigin: null));
```
NachoEchevarria, line 493: This forces the creation of IAST and security instances even if not enabled. On the other hand, we are calling Security.Instance and Iast.Instance in many other places, so, in practical terms, we are going to create the instances anyway. Actually, we need to call Instance to read the IAST and ASM settings. Just wondering if at some point we could separate the security settings and access them without actually creating a (not needed mostly) instance of Security or IAST. We are creating instances just to read the settings. Still, a sensitive change (we need to be careful with RC, for instance)

Anyway, definitely, totally out of the scope of this PR. The performance optimization here is sound given the current architecture.

andrewlock: > This forces the creation of IAST and security instances even if not enabled

So these are actually _already_ created automatically 😅 e.g. we explicitly do `_ = Security.Instance;` just above. 

> Actually, we need to call Instance to read the IAST and ASM settings. Just wondering if at some point we could separate the security settings and access them without actually creating a (not needed mostly) instance of Security or IAST. 

💯 This is something I'd definitely like to do, and I think is worthwhile, but for the exact reason you state "we need to be careful with RC, for instance", I suspect we will _generally_ need to do a certain amount of initialization no matter what, so I don't think it's a massive issue. And like you say, outside of the scope for this PR, but definitely something for us to think about! 🙂

Issue #4093, opened 2023-04-27
Duplicated spans using ASP.NET controllers


matteopey:
**Describe the bug**
I'm not sure if this is a bug or is by design, so I'm opening this issue to have some guidance on how to solve this.
I have an application that is using .NET Core 6. The controllers in this applications are implemented using the `ControllerBase` class from ASP.NET. 
The result on Datadog is that inside the trace two spans are shown, even though the controller path and the request made is only one. 
The service entry is of type `aspnet_core.request` the inner one is of type `aspnet_core_mvc_request`.
Instead, using the new routing like this `app.MapGet("/api/mapget", () => "Hello!");` only one trace is shown.

**To Reproduce**
I setup Datadog with an agent running on the host machine on Docker with APM enabled and the application is running on the same host, integrating the .NET tracer as explained in the `/apm/intro` guide on Datadog.

**Expected behavior**
One span is shown given the single request and the single controller.

**Screenshots**
**Controllers**
![immagine](https://user-images.githubusercontent.com/28830727/234980879-44a8e7d0-b90f-4185-ace2-447f45f637e0.png)

**MapGet**
![immagine](https://user-images.githubusercontent.com/28830727/234981243-44b33d8b-eb7b-40d8-8056-9b8a64a2567a.png)

**Runtime environment (please complete the following information):**
- Automatic instrumentation
- Tracer version: 2.29.0
- OS: Windows 10 on WSL2. Happening in production on K8S cluster.
- CLR: >= .NET Core 3.1

**Additional context**
Noticed this issue in our production systems, but is easily reproducible with a simple application locally.

andrewlock:
Hi @matteopey, the behaviour you're seeing is expected when you use minimal APIs. There is no controller/Razor Page with minimal APIs so the MVC request aspect doesn't apply here, and there's no additional span to create. The root span contains details about the selected endpoint.

Can I take it from this issue that you would _like_ there to be an additional span for the endpoint execution? If so, what information is it that you'd like to obtain? Is it the _duration_ of the endpoint relative to the whole request? Or is there some additional information you would hope to see? 

We've had some feedback that having two spans per request is problematic for some people, so it would be good to know if you _prefer_ the two-span setup we have. Thanks!

matteopey:
Hi @andrewlock I probably explained myself badly.
The behavior of the minimal API is the one that I think is **correct**. 
[...]

Issue #1147, opened 2021-01-19
Feature request: Set tags & errors on a root span from a child span


lukeg-unleashed:
**Is your feature request related to a problem? Please describe.**
I saw that the dd-trace-java APM client has the ability to access the root span for the current trace - this could be very helpful when tagging the trace with important request-specific metadata (e.g. user/account ids). 

**Describe the solution you'd like**
Ideally we would have the ability to do something similar to that described here: https://docs.datadoghq.com/tracing/setup_overview/custom_instrumentation/java#set-tags--errors-on-a-root-span-from-a-child-span

**Describe alternatives you've considered**
I've considered using reflection to access the internal Context?.TraceContext?.RootSpan property, but would rather avoid a messy workaround (for hopefully obvious reasons)!

**Additional context**
Please feel free to mention if there's any more detail I can provide.

fjmorel:
This would be super useful in ASP.NET Core requests in order for code running in subspans to add tags to the root span. For example, I add every feature flag key/value pair as a tag on the request span currently by storing a reference to it at the beginning of the request.

rparsonsbb:
+1 on this, I'm very surprised that this functionality isn't available out of the box.  Are there tips on how to bubble up tags to the root span to be able to search on it?

davenewza:
Yep, I'll be looking out for this feature 👀 👍

aforloney:
As a pulse check, is there any update on this?

mlstubblefield:
+1

lpodolak:
+1. 
I realised that by having these 2 spans, we are not able to filter traces by values that are present in both spans. 
For example, the aspnet_core.request span contains the "status_code", while the aspnet_core_mvc.request constains our custom metadata - e.g. "api.channel". As a result, if we want to filter by both of these values (no matter what we select as the scope: all spans, root spans or service entry spans), we get no results. Which isn't true! 

![image](https://github.com/DataDog/dd-trace-dotnet/assets/82114/1fee026c-3e84-4561-a3cc-d7426b9b90df)

Our problem is that we annotate every request but with values that are present only after data-binding, which means it's already after middleware is executed, hence we land in the "aspnet_core_mvc.request" scope. 

I also referenced the problem [here](https://github.com/DataDog/dd-trace-dotnet/issues/4093#issuecomment-1550217451)

lpodolak:
[...]

PR #7964, opened 2025-12-17
Add a new single-span diagnostic observer for ASP.NET Core

andrewlock:
## Summary of changes

Adds a new `DD_TRACE_SINGLE_SPAN_ASPNETCORE_ENABLED` in which we omit the MVC span

## Reason for change

- Addresses #4093 and partially #1147
- Omitting the MVC span has perf and some usability benefits in some cases (See issues above). It _does_ mean a certain amount of less information, particularly in pipeline re-execution scenarios

## Implementation details

- Add a new, disabled by default, `DD_TRACE_SINGLE_SPAN_ASPNETCORE_ENABLED` for opting in to the new behaviour.
- .NET 6+ only for various reasons
- Cloned the original observer, deleted everything related to the MVC span, and then optimized further based on .NET 6+ and other improvements we can make

## Test coverage

- Added in-process "integration" tests for the observer (more like unit tests), which mirror the exisitng.
- Added "full" sample-based integration tests for basically all scenarios we currently test (IIS in process/out of process / MVC / minimal APIs / proxy spans etc)
- Added snapshots by copying existing, and then updating, so if you look at the final commit, you can see the diff, which is basically just "delete the child span"

## Other details

Part of a stack

- https://github.com/DataDog/dd-trace-dotnet/pull/7964 👈

I experimented with some other perf improvements, but they didn't provide a significant improvement, so I reverted:
- Caching the resource name by saving it into the `IEndpointMetadata` collection. The work to do that was too much, and you _can't_ cache in some cases anyway so not worth it
- Splitting `WebTags` out to remove the IP headers unless we need them. They're disabled by default, so it just increase the size of the tags object, but wasn't a big enough improvement to make the complexity worth it.
- Tried pooling the `RequestTrackingFeature` objects. This one _may_ actually be worth exploring again later, but I didn't see any improvement in my testing, as they're just not very big and only last 1 request.

### DIFF
--- a/tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
+++ b/tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
@@ -387,8 +387,8 @@ namespace Datadog.Trace.Configuration
             {
                 SingleSpanAspNetCoreEnabled = false;
                 Log.Warning(
-                    $"{ConfigurationKeys.FeatureFlags.SingleSpanAspNetCoreEnabled} is set, but is only supported in .NET 6+");
-                _telemetry.Record(ConfigurationKeys.FeatureFlags.SingleSpanAspNetCoreEnabled, false, ConfigurationOrigins.Calculated);
+                    $"{ConfigurationKeys.FeatureFlags.SingleSpanAspNetCoreEnabled} is set to true, but is only supported in .NET 6+. Using false instead.");
+                telemetry.Record(ConfigurationKeys.FeatureFlags.SingleSpanAspNetCoreEnabled, false, ConfigurationOrigins.Calculated);
             }
 #endif
 


tracer/src/Datadog.Trace/Configuration/supported-configurations-docs.yaml
```diff
@@ -1037,6 +1037,13 @@ DD_TRACE_SERVICE_MAPPING: |
   Configuration key for a map of services to rename.
   <seealso cref="Datadog.Trace.Configuration.MutableSettings.ServiceNameMappings"/>
 
+DD_TRACE_SINGLE_SPAN_ASPNETCORE_ENABLED: |
+  Feature Flag: enables single-span ASP.NET Core traces, by only creating `aspnet_core.request` spans, without
```
lucaspimentel: "single-span ASP.NET Core traces" sounds like the entire trace is collapsed into a single span, not just the ASP.NET Core spans.

Consider these options:
- Feature Flag: enables generating a single span in the ASP.NET Core integration by only creating...
- Feature Flag: collapses (merges?) all ASP.NET Core spans into a single aspnet_core.request span, without...

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -380,9 +380,23 @@ not null when string.Equals(value, "otlp", StringComparison.OrdinalIgnoreCase) =
...
+            SingleSpanAspNetCoreEnabled = config
+                                         .WithKeys(ConfigurationKeys.FeatureFlags.SingleSpanAspnetcoreEnabled)
+                                         .AsBool(defaultValue: false);
+#if !NET6_0_OR_GREATER
+            // single span aspnetcore is only supported in .NET 6+, so override for telemetry purposes
+            if (SingleSpanAspNetCoreEnabled)
+            {
+                SingleSpanAspNetCoreEnabled = false;
+                Log.Warning(
+                    $"{ConfigurationKeys.FeatureFlags.SingleSpanAspnetcoreEnabled} is set, but is only supported in .NET 6+");
```
lucaspimentel: Make it more clear we're disabling the feature?
```suggestion
                    $"{ConfigurationKeys.FeatureFlags.SingleSpanAspnetcoreEnabled} is set to true, but is only supported in .NET 6+. Using false instead.");
```

tracer/src/Datadog.Trace/Configuration/configuration_keys_mapping.json
```diff
@@ -604,6 +604,10 @@
       "env_var": "DD_SITE",
       "const_name": "Site"
     },
+    {
+      "env_var": "DD_TRACE_SINGLE_SPAN_ASPNETCORE_ENABLED",
+      "const_name": "SingleSpanAspnetcoreEnabled"
```
lucaspimentel: I thought this was used for codegen and would've caused a compiler error?
```suggestion
      "const_name": "SingleSpanAspNetCoreEnabled"
```

lucaspimentel (approved):

tracer/src/Datadog.Trace/Configuration/configuration_keys_mapping.json
andrewlock: ~Yeah, me too 😕~ Ah, it is used for keygen, and it didn't cause an error, because I used the generated value before 😄 Fixed now

tracer/src/Datadog.Trace/Configuration/supported-configurations-docs.yaml
andrewlock: Makes sense, I'm not entirely happy with "Single span" as a moniker at all, other than the fact it's somewhat descriptive. I also considered "simplified", but was concerned in the future we might end up with "more simplified", and "even simpler again" settings too 😂 


Went with the first suggestion, thanks 👍

tracer/src/Datadog.Trace/DiagnosticListeners/SingleSpanAspNetCoreDiagnosticObserver.cs
```diff
@@ -0,0 +1,427 @@
...
+                    else if (routeEndpoint.Value.RequestDelegate?.TryDuckCast<Target>(out var target) == true && target is { Handler: { } handler })
+                    {
+                        Log.Debug("RouteEndpoint?.RequestDelegate?.Method is null. Extracting code origin from RouteEndpoint.RequestDelegate.Target.Handler {Handler}", handler);
+                        codeOrigin.SetCodeOriginForEntrySpan(rootSpan, handler.Target?.GetType(), handler.Method);
+                    }
+                    else
+                    {
+                        Log.Debug("RouteEndpoint?.RequestDelegate?.Method is null and could not extract handler from RouteEndpoint.RequestDelegate.Target");
+                    }
+                }
```
GreenMatan, line 261: Can we reduce code duplication between `AspNetCoreDiagnosticObserver` and this class? I'm concerned we'll fix some bugs related to this code and forget to update one of the two implementations

GreenMatan (approved):
Left a comment regarding code duplication

tracer/src/Datadog.Trace/DiagnosticListeners/SingleSpanAspNetCoreDiagnosticObserver.cs
andrewlock: I intentionally avoided removing duplication so far, _primarily_ because I consider this implementation to be somewhat experimental initially, and I'd rather not risk breaking on of our most used integrations (aspnetcore) by trying to reduce it. This implementation was also trying to be as fast as it can be, so I was actually inlining methods in various places too 😄 

That said, I would very much like to consolidate further soon 🙂 And that likely includes extracting this CodeOrigin handling stuff, especially as code origin is disabled by default IIRC!

GreenMatan: Got it. We'll have to be extra careful then until we have the full implementation in place

tonyredondo (approved):

PR #7966, opened 2025-12-17
Create a more optimized `AspNetCoreResourceNameHelper`

andrewlock:
## Summary of changes

Creates an optimized version of `AspNetCoreResourceNameHelper` for use by the single-span Diagnostic observer.

## Reason for change

We have the opportunity for a variety of optimizations thanks to not creating the MVC span, and the fact that we're .NET 6+. We could potentially include some or all of these optimizations in the existing observer path, but for safety/simplicity I'd rather keep it isolated for now, and we can look at expanding it more broadly later

## Implementation details

- Use the `ValueStringBuilder` to build a resource name
- Don't pre-grab MVC route values, wait to see if we have route params from them first
- Write directly into the builder as `ToLowerInvariant`, instead of doing an extra ToLower() at the end

## Test coverage

Added unit tests that use the existing data for the other implementations, to confirm it's identical. It's also covered indirectly by the integration tests

## Other details

Part of a stack

- https://github.com/DataDog/dd-trace-dotnet/pull/7966 👈


tracer/src/Datadog.Trace/DiagnosticListeners/AspNetCoreResourceNameHelper.cs
```diff
@@ -19,6 +19,118 @@ namespace Datadog.Trace.DiagnosticListeners;
...
+                        }
+
+                        if (routeValueDictionary.TryGetValue(parameterName, out var value)
+                            && value is string name)
+                        {
+                            sb.AppendAsLowerInvariant(name);
+                        }
+                        else
+                        {
+                            sb.Append(parameterName);
```
NachoEchevarria: NIT: Previously we were not appending the parameterName to the route name but write an empty value if no value was found. This change seems reasonable, but still worth to note the different which, I assume, is intended.

tracer/src/Datadog.Trace/DiagnosticListeners/AspNetCoreResourceNameHelper.cs
```diff
@@ -19,6 +19,118 @@ namespace Datadog.Trace.DiagnosticListeners;
...
 internal static class AspNetCoreResourceNameHelper
 {
+#if NET6_0_OR_GREATER
+    internal static string SimplifyRoutePattern(
+        RoutePattern routePattern,
+        RouteValueDictionary routeValueDictionary,
+        bool expandRouteParameters)
+    {
+        var sb = routePattern.RawText?.Length < 512
+                     ? new ValueStringBuilder(stackalloc char[512])
```
lucaspimentel, line 29: 😍

tracer/src/Datadog.Trace/DiagnosticListeners/AspNetCoreResourceNameHelper.cs
andrewlock: Good catch, this is essentially a bug in the _existing_ implementations, but _also_ in the new one! It's a case that shouldn't generally be hit, in that I'm not sure it can ever be hit by a _valid_ route template + route value combination 🤔 but we should probably handle it in any case. For example, if you have a (very contrived) route like this:

```
"{area?}/{controller?}/{action?}"
```

I'm not sure it would ever be matched. But if we assume it is, and we have a value for `controller` (`Home`) and `action` (`Index`) but _not_ `area`, then the calculated route we use should be:

```
/{area}/home/index
```

However, the existing implementations all omit it `{area}`, and generate:

```
//home/index
```

which is obviously wrong. I attempted to fix that, but the current fix accidentally is doing:

```
/area/home/iindex
```

which is _also_ wrong 😅 

So I will
- Add tests and fix the _existing_ handlers
- Fix the new handler

Thanks!

PR #7967, opened 2025-12-17
Add PR-level performance quality gates

igoragoli:
## Summary of changes

- Adds PR-level gates to dd-trace-dotnet with:
    - A regression threshold of 20%: https://github.com/DataDog/benchmarking-platform/pull/224/files#diff-e2c6df715b4e55dca3365e30ac3801dd5d9ca5c212e18ed45ead5e2cff811077R6
    - A `check-big-regressions-no-ignored-benchmarks` CI job that won't block merges (since `allow-failure` is set to true).
- Update `.gitlab-ci.yml`'s `microbenchmarks`:
    - Set `strategy: depends` to have `microbenchmarks` fail if `check-big-regressions` (the PR gate) fails.
    - Set `allow_failure: false` to have `microbenchmarks` failures prevent PRs from being merged.

## Reason for change

https://datadoghq.atlassian.net/browse/APMSP-2292

## Other details

`run-benchmarks` has a known issue @NachoEchevarria recently mentioned of jobs failing when AWS allocates a machine with a CPU different than the one we use for benchmarking. 

This happens rarely, but it might lead to some frustration when attempting merges, because it would require manual retries.

An automatic retry strategy would require changes to `bp-infra` I'm planning to do. But I think that we can get this PR merged as is, especially considering that there are other jobs on dd-trace-dotnet's CI that require manual retries sometimes.

If failures prove to be too disruptive, we can always set `allow_failure` to `true` and disable PR gates momentarily.

### DIFF
--- a/.gitlab/benchmarks/microbenchmarks.yml
+++ b/.gitlab/benchmarks/microbenchmarks.yml
@@ -157,6 +157,7 @@ run-benchmarks:
       when: always
     - interruptible: true
       when: always
+  allow_failure: true  # Allowed to fail since we don't want it to block PRs yet
   artifacts:
     name: "artifacts"
     when: always
@@ -193,6 +194,5 @@ check-big-regressions:
 # unstable has been made
 check-big-regressions-no-ignored-benchmarks:
   extends: .check-big-regressions
-  allow_failure: true  # Allowed to fail since we do not want it to block PRs 
   variables:
     IGNORED_BENCHMARKS_REGEX: ""


.gitlab-ci.yml
```diff
@@ -223,7 +223,8 @@ microbenchmarks:
   needs: []
   trigger:
     include: .gitlab/benchmarks/microbenchmarks.yml
-  allow_failure: true
+    strategy: depend
+  allow_failure: false
```
NachoEchevarria: According to slack conversations, we will set this to false before merging, right? I agree that's a good approach for now.

.gitlab/benchmarks/microbenchmarks.yml
```diff
@@ -138,3 +139,60 @@ run-benchmarks:
...
+    name: "artifacts"
+    when: always
+    paths:
+      - reports/
+    expire_in: "30 days"
+  variables:
+    BP_INFRA_BENCHMARKING_PLATFORM_BRANCH: "dd-trace-dotnet/micro"
+
+    # Regex to ignore benchmarks identified as unstable
+    IGNORED_BENCHMARKS_REGEX: "Trace.Iast.StringAspectsBenchmark|Trace.CharSliceBenchmark|Trace.Asm.AppSecWafBenchmark|Trace.Asm.AppSecBodyBenchmark|Trace.CIVisibilityProtocolWriterBenchmark"
```
NachoEchevarria, line 171: Checking hte jobs that are more flaky, we seem that we are mostly covered. Still some of the netcoreapp3.1 ones are not included in this list. I would say that, for now, we should exclude any netcoreapp3.1 test. Is that possible?

.gitlab-ci.yml
igoragoli: To `true`, actually! So that the `microbenchmarks` child pipeline failing does not impact the main pipeline.

igoragoli: Addressed in 0aeab66783251ce25ce24e88c2a676e16a8b3052.

.gitlab/benchmarks/microbenchmarks.yml
igoragoli: Targeting netcoreapp3.1 benchmarks is a bit more tricky, since we'd have to look inside benchmark result files. Absolutely doable! But maybe too complex for now :)

To explain a bit more, the regex matches benchmark files. A benchmark file can have many scenarios. The netcoreapp3.1 identifier appears in some scenarios inside the benchmark file.

PR #7968, opened 2025-12-17
fix(DSM): Fix race condition on DataStreamsWriter disposal

robcarlan-datadog:
## Summary of changes
Fixes a race condition where we attempt to dispose of a semaphore when it is being held by another task.

## Reason for change
I think this is the root cause behind some unit test flake, and some errors when disposing DataStreamsWriter.

## Implementation details
There are two tasks that run: `flushTask` and `processTask`. `DataStreamsWriter::DisposeAsync` calls `_flushSemaphore.Dispose();` after `FlushAndCloseAsync()`.

`FlushAndCloseAsync` sets an exit flag for the task, but only waits for `processTask` to complete. If `flushTask` is running and has already acquired the semaphore, the `DisposeAsync` will still dispose of the semaphore even though flushTask is still using it.

Fixed by waiting for both tasks to complete, as flushTask will finish once processTask finishes. And we still have the 1 second fallback.

This should also fix the test flake. The unit test immediately calls dispose after adding the data points:
```
writer.Add(CreateStatsPoint(timestamp));
writer.AddBacklog(CreateBacklogPoint(timestamp));

await writer.DisposeAsync();
```

So it seems likely that we might dispose the semaphore before flush has a chance to run.

### DIFF
--- a/tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
+++ b/tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
@@ -171,17 +171,18 @@ internal sealed class DataStreamsWriter : IDataStreamsWriter
         }
 
         // nothing else to do, since the writer was not fully initialized
-        if (!Volatile.Read(ref _isInitialized) || _processTask == null)
+        if (!Volatile.Read(ref _isInitialized) || _processTask == null || _flushTask == null)
         {
             return;
         }
 
+        var allTasks = Task.WhenAll(_processTask, _flushTask);
         var completedTask = await Task.WhenAny(
-                                           _processTask,
+                                           allTasks,
                                            Task.Delay(TimeSpan.FromSeconds(1)))
                                       .ConfigureAwait(false);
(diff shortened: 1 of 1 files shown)


tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsWriter.cs
```diff
@@ -171,17 +171,18 @@ private async Task FlushAndCloseAsync()
         }
 
         // nothing else to do, since the writer was not fully initialized
-        if (!Volatile.Read(ref _isInitialized) || _processTask == null)
+        if (!Volatile.Read(ref _isInitialized) || _processTask == null || _flushTask == null)
         {
             return;
         }
 
+        var allTasks = Task.WhenAll(_processTask, _flushTask);
```
robcarlan-datadog, line 179: Flush task will return after processTask returns (which returns because we set processExit above). So there will be nothing using the semaphore after this.

Hence, we are safe to dispose the semaphore on L163 after this method returns because no task will be using it.

PR #7969, opened 2025-12-17
LifetimeManager .NET10 fix

tonyredondo:
## Summary of changes

Fix the LifetimeManager on .NET 10

## Reason for change

https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/10.0/sigterm-signal-handler

## Test coverage

Before this PR:

image

After this PR:

image


tracer/src/Datadog.Trace/LifetimeManager.cs
```diff
@@ -168,5 +229,99 @@ static void SetSynchronizationContext(SynchronizationContext? context)
                 }
             }
         }
+
+#if NET6_0_OR_GREATER
+        private static void TryLogWarning(Exception ex, string message)
+        {
+            try
+            {
+                // The message template must be a constant to avoid allocations and to satisfy logging analyzers.
```
lucaspimentel: We don't need this comment here (or, we could add it _everywhere_ we use log message templates, you know, for consistency 😅 )

tonyredondo: yeah I don't like it either, I've just removed the helper

tracer/src/Datadog.Trace/LifetimeManager.cs
```diff
@@ -168,5 +229,86 @@ static void SetSynchronizationContext(SynchronizationContext? context)
...
+                if (Environment.Version.Major < 10)
+                {
+                    // On .NET <= 9, the runtime provided default termination handlers that result in graceful exit,
+                    // so we do not install our own to avoid changing long-standing behavior.
+                    return;
+                }
+
+                // Register termination handlers to restore the pre-.NET 10 behavior (graceful managed shutdown).
+                //
+                // Microsoft guidance suggests registering SIGTERM on Unix and SIGTERM+SIGHUP on Windows:
```
andrewlock, line 248: > Microsoft guidance suggests ... SIGTERM+SIGHUP on Windows

_Does_ it? 🤔 I read the guidance as SIGTERM on Unix and SIGHUP on Linux.. But it's definitely not clear 😅

tonyredondo: that's what I understood on this phrase:

>On Unix systems, the .NET runtime no longer provides a default SIGTERM signal handler. On Windows, the .NET runtime no longer provides default handlers for the [CTRL_SHUTDOWN_EVENT and CTRL_CLOSE_EVENT signals](https://learn.microsoft.com/en-us/windows/console/handlerroutine), which are equivalents of Unix SIGTERM and SIGHUP signals.

Unix SIGTERM
Windows both SIGTERM AND SIGHUP

andrewlock: Yeah, ok, that does sound like it 👍

tracer/src/Datadog.Trace/LifetimeManager.cs
```diff
@@ -168,5 +229,86 @@ static void SetSynchronizationContext(SynchronizationContext? context)
                 }
             }
         }
+
+#if NET6_0_OR_GREATER
+        private void TryRegisterTerminationSignalHandlers()
```
lucaspimentel, line 234: Have you tested this manually? Can we add automated tests?

tonyredondo: I've added tests in [cf14750](https://github.com/DataDog/dd-trace-dotnet/pull/7969/commits/cf14750d6c7e2104d184c21641e52aa36ad1b322)

tonyredondo:
> Nice, thanks! I wonder if we should add a smoke check to confirm our shutdown tasks are actually running (to confirm they weren't running previously). I'm kind of surprised we never saw this show up in our tests before 🤔

I've added a test

PR #7975, opened 2025-12-18
Fix `System.InvalidCastException` in Grpc during Header Removal

bouwkast:
## Summary of changes

This resolves a `System.InvalidCastException` that was happening in the GRPC client instrumentation (I think just GrpcLegacy) when we attempted to remove headers. 

## Reason for change

Errors in Error Tracking

## Implementation details

> before

```
public void Remove(string name)
{
    var entry = _headers.Get(name);
    if (entry is not null)
    {
        _headers.Remove(name);
    }
}
```

> after
```
public void Remove(string name)
{
    var entry = _headers.Get(name);
    if (entry is not null)
    {
        _headers.Remove(entry);
    }
}
```

## Test coverage

Updated the sample application to detect this and if it is detected it adds a tag which causes the tests to fail (and they did fail before I applied the fix).

## Other details

Used https://github.com/DataDog/dd-trace-dotnet/pull/7961 `/analyze-error` command based on the reported error in Error Tracking. Seemed to work okay, it identified the issue, follow up was able to get a reproduction quickly, but it really struggled to get the tests updated.

Helpfully, the command output a document that linked to and copied the actual `Grpc.Core.Api` code that we hit 👍 
Saves some time searching on GitHub ourselves I guess.

This is a bit of an edge case here so I don't think very wide spread.

Reproduced Error:

```
2025-12-18 10:03:42.678 -05:00 [ERR] Error creating inactive client span for GRPC call System.InvalidCastException: Unable to cast object of type 'System.String' to type 'Entry'.
   at Grpc_Core_Api__D754F35622E28BAD.Grpc_Core_Metadata.Datadog_Trace_ClrProfiler_AutoInstrumentation_Grpc_IMetadata_3.Remove(Object )
   at Datadog.Trace.ClrProfiler.AutoInstrumentation.Grpc.MetadataHeadersCollection.Remove(String name)
   at Datadog.Trace.ClrProfiler.AutoInstrumentation.Grpc.MetadataHeadersCollection.Set(String name, String value)
   at Datadog.Trace.Headers.HeadersCollectionAccesor`1.Set(TCarrier carrier, String key, String value)
   at Datadog.Trace.Propagators.W3CTraceContextPropagator.Inject[TCarrier,TCarrierSetter](PropagationContext context, TCarrier carrier, TCarrierSetter carrierSetter)
   at Datadog.Trace.Propagators.SpanContextPropagator.Inject[TCarrier,TCarrierSetter](PropagationContext context, TCarrier carrier, TCarrierSetter carrierSetter)
   at Datadog.Trace.Propagators.SpanContextPropagator.Inject[TCarrier](PropagationContext context, TCarrier headers)
   at Datadog.Trace.ClrProfiler.AutoInstrumentation.Grpc.GrpcLegacy.Client.GrpcLegacyClientCommon.InjectHeaders[TMethod,TCallOptions](Tracer tracer, TMethod method, TCallOptions& callOptionsInstance)
 { MachineName: ".", Process: "[84684 Samples.GrpcLegacy]", AppDomain: "[1 Samples.GrpcLegacy.exe]", TracerVersion: "3.34.0.0" }

```

### DIFF
--- a/tracer/test/test-applications/integrations/Samples.GrpcLegacy/ClientWorker.cs
+++ b/tracer/test/test-applications/integrations/Samples.GrpcLegacy/ClientWorker.cs
@@ -57,10 +57,7 @@ public class ClientWorker
                     meta.Add("client-value1", "some-client-value");
                     meta.Add("client-value2", "other-client-value");
 
-                    // Pre-populate W3C trace headers to trigger the bug
-                    // When dd-trace tries to inject its own headers, it will call Set()
-                    // which calls Remove() with a string instead of the entry object
-                    _logger.LogInformation("Adding pre-existing W3C trace headers to trigger Remove() bug");
+                    _logger.LogInformation("Adding pre-existing W3C trace headers to ensure we can replace them correctly");
                     meta.Add("traceparent", "00-0af7651916cd43dd8448eb211c80319c-b7ad6b7169203331-01");
                     meta.Add("tracestate", "dd=s:1");
 


tracer/test/test-applications/integrations/Samples.GrpcLegacy/ClientWorker.cs
```diff
@@ -56,6 +56,14 @@ public async Task ExecuteAsync(CancellationToken stoppingToken)
                 {
                     meta.Add("client-value1", "some-client-value");
                     meta.Add("client-value2", "other-client-value");
+
+                    // Pre-populate W3C trace headers to trigger the bug
+                    // When dd-trace tries to inject its own headers, it will call Set()
+                    // which calls Remove() with a string instead of the entry object
+                    _logger.LogInformation("Adding pre-existing W3C trace headers to trigger Remove() bug");
```
andrewlock: nit: given this is fixed now, probably makes sense to reword the comment in-code - we'll have the details of the bug in this PR 🙂  

```suggestion
                    _logger.LogInformation("Adding pre-existing W3C trace headers to ensure we can replace them correctly");
```

andrewlock (approved):
Nice, thanks!

PR #7979, opened 2025-12-19
Don't poll discovery service if we've received an update from the agent recently

andrewlock:
## Summary of changes

Don't poll discovery service if we're received an update from the agent recently

## Reason for change

The agent sends a sha256 hash of the `/info` endpoint data in every response. We can use that to track whether the `/info` value has changed, and if it hasn't, we can skip polling the endpoint needlessly.

This obviously removes the overhead of polling the endpoint, but it's not entirely clear the impact this will have on performance, as we're doing "extra" work by materializing the header and calling the discovery service, everytime we send a trace.

## Implementation details

- When polling the `/info` data, take a SHA256 of the data, and store it as a hex string
- After sending a trace, grab the state-config-hash header and pass it to the `DiscoveryService`
- When the `DiscoveryService` is next due to poll (every 30s) check if the existing hash matches the most recent value from the header (and that it was received within the last 30s). If so, skip the current poll.

## Test coverage

Added unit tests for all of the components

## Other details

https://datadoghq.atlassian.net/browse/LANGPLAT-815

Couple of points of interest:
- Using `CryptoStream` makes calculating the hash low allocation, but it assumes we're receiving the data as UTF-8 etc (which we are, but you know, we don't currently verify that)
- It's kind of tough to measure the perf impact of this. We're _assuming_ that the overhead of materialising the string (if it's not already, it will be in many if not all TFMs) and invoking the action on every send to the agent will be _lower_ than polling the agent once every 30s. But proving that will be tricky

### DIFF
--- a/tracer/test/Datadog.Trace.Tests/Agent/ApiTests.cs
+++ b/tracer/test/Datadog.Trace.Tests/Agent/ApiTests.cs
@@ -71,7 +71,7 @@ namespace Datadog.Trace.Tests.Agent
 
             var containerMetadata = new ContainerMetadata(containerId, entityId);
 
-            var api = new Api(factoryMock.Object, TestStatsdManager.NoOp, containerMetadata, updateSampleRates: null, partialFlushEnabled: false, healthMetricsEnabled: false);
+            var api = new Api(factoryMock.Object, TestStatsdManager.NoOp, containerMetadata, updateSampleRates: null, updateConfigState: null, partialFlushEnabled: false, healthMetricsEnabled: false);
 
             await api.SendTracesAsync(new ArraySegment<byte>(new byte[64]), numberOfTraces: 1, statsComputationEnabled: false, numberOfDroppedP0Traces: 0, numberOfDroppedP0Spans: 0, apmTracingEnabled: false);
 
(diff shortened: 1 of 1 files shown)


tracer/src/Datadog.Trace/Agent/DiscoveryService/DiscoveryService.cs
```diff
@@ -252,9 +281,44 @@ int GetNextSleepDuration(int? previousDuration) =>
                 previousDuration is null ? _initialRetryDelayMs : Math.Min(previousDuration.Value * 2, _maxRetryDelayMs);
         }
 
+        [TestingAndPrivateOnly]
+        internal bool RequireRefresh(string? currentHash, DateTimeOffset utcNow)
```
NachoEchevarria, line 285: Could there be a race condition here?
Thread A reads  _agentConfigStateHash 
Thread B calls SetCurrentConfigStateHash() and updates hash and time
Thread A reads _agentConfigStateHashUnixTime and gets a fresh timestamp, return false
Not critical though, it would be rare and the consecuences are not severe. WDYT?

andrewlock: > Could there be a race condition here?

Yup, absolutely, there could. When I first wrote this, I worked around that by creating a `record State(string Hash, DateTimeOffset utcNow)` and then atomically switching that. The problem is that `SetCurrentConfigStateHash()` is going to be called a lot, so it would result in _either_ a more complex "atomically read, compare, conditionally recreate  , atomically write" comparison, _or_ if we just always write (like we're doing currently) then we end up allocating every time. 

My conclusion was pretty much the same as yours, it should be rare, and not severe (we will pick it up next time), and given this is primarily meant to be a perf improvement, I think it's worth it 🙂

PR #7981, opened 2025-12-19
Fix OnSupportedFrameworkInSsi_CallsForwarderWithExpectedTelemetry flakiness

NachoEchevarria:
## Summary of changes

Fixes flaky test `InstrumentationTests.OnSupportedFrameworkInSsi_CallsForwarderWithExpectedTelemetry` by adding a retry mechanism when checking for the telemetry echo log file.

## Reason for change

The test was intermittently failing in CI with:
```
Expected File.Exists(echoLogFileName) to be True, but found False.
```

The test is already marked as flaky because of [this issue](https://github.com/NuGet/Home/issues/14343). This seems to be a different problem though. 

**Root cause:** Race condition between test assertion and asynchronous telemetry forwarder execution.

The telemetry echo log file is written by a **separate process** (`telemetry_echo.exe`), not by the sample application itself. The native loader spawns this forwarder process on a detached thread (`single_step_guard_rails.cpp:346`) to avoid blocking application startup. Since the forwarder runs in a separate process, the sample application can complete and exit before the forwarder finishes writing the log file. On CI machines with high CPU contention, this timing variance becomes observable, causing the test to check for the file before it exists.

**CI Failure:** https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=192494&view=logs&j=ca775ed2-8590-5a92-528d-855583aa2d34&t=9804592b-55de-5ba0-6562-0dbf7f7ff3f6

## Implementation details

Added a polling loop with 10-second timeout in `AssertHasExpectedTelemetry` (line 799-807):
- Polls every 100ms for file existence
- Allows up to 10 seconds for the detached telemetry thread to complete
- Provides clear failure message if timeout is exceeded
- Similar pattern already used elsewhere in the codebase (e.g., `Samples.Console/Program.cs:71-74`)

The fix acknowledges the production design (non-blocking telemetry on detached thread) while making the test tolerant to timing variance.

## Test coverage

This change fixes an existing flaky test. The test itself validates telemetry reporting in Single-Step Instrumentation scenarios and continues to provide the same coverage with improved reliability.

### DIFF
--- a/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/InstrumentationTests.cs
+++ b/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/InstrumentationTests.cs
@@ -433,7 +433,7 @@ namespace Foo
                                "name": "library_entrypoint.abort.runtime"
                              }]
                              """;
-            AssertHasExpectedTelemetry(logFileName, processResult, pointsJson,  "abort", ".NET Core 3.0 or lower", "incompatible_runtime");
+            await AssertHasExpectedTelemetry(logFileName, processResult, pointsJson,  "abort", ".NET Core 3.0 or lower", "incompatible_runtime");
         }
 
         [SkippableFact]
@@ -464,7 +464,7 @@ namespace Foo
                                "tags": ["injection_forced:true"]
                              }]
                              """;
-            AssertHasExpectedTelemetry(logFileName, processResult, pointsJson, "success", "Force instrumentation enabled, incompatible runtime, .NET Core 3.0 or lower", "success_forced");
+            await AssertHasExpectedTelemetry(logFileName, processResult, pointsJson, "success", "Force instrumentation enabled, incompatible runtime, .NET Core 3.0 or lower", "success_forced");
         }
 
 #endif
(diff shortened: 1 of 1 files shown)


tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/InstrumentationTests.cs
```diff
@@ -795,7 +799,18 @@ private void AssertNativeLoaderLogContainsString(string logDir, string requiredL
...
             using var s = new AssertionScope();
-            File.Exists(echoLogFileName).Should().BeTrue();
+
+            // Wait for the telemetry echo file to be written (with timeout)
+            // The telemetry forwarder may write the file asynchronously after process exit
+            var fileWaitTimeout = TimeSpan.FromSeconds(10);
+            var fileWaitStart = DateTime.UtcNow;
+            while (!File.Exists(echoLogFileName) && (DateTime.UtcNow - fileWaitStart) < fileWaitTimeout)
+            {
+                Thread.Sleep(100);
```
andrewlock: Given this is going to sleep for up to 10s, should we make this an async `Task.Delay` instead? 🤔

andrewlock (approved):
Thanks makes sense!

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/InstrumentationTests.cs
NachoEchevarria: Sure, that would mitigate the delay. Done!

PR #7986, opened 2025-12-20
[Dynamic Instrumentation] DEBUG-4871 Added a managed guard in LogLocal and LogArg to detect null byrefs

dudikeleti:
## Reason for change
Some methods can legitimately produce null managed byrefs (e.g., MemoryMarshal.GetReference(emptySpan)), which makes ref T point to address 0. Attempting to capture such locals/args would dereference an invalid reference and could crash in LogLocal/LogArg (NullReferenceException). This change makes DI runs stable and keeps probes installed while avoiding unsafe dereferences

## Implementation details
Before reading a ref T capture input, we check Unsafe.IsNullRef(ref value) and return early to avoid dereferencing an invalid reference.

Current behavior is to skip capturing that member entirely (no misleading null value for non-nullable types like byte).
Follow-up option: if we decide we want the snapshot to explicitly show this case, we can reuse the existing snapshot notCapturedReason mechanism (no new reason added in this PR).

## Test coverage
https://github.com/DataDog/dd-trace-dotnet/pull/7987

## Other details
https://github.com/DataDog/dd-trace-dotnet/pull/7988
https://github.com/DataDog/dd-trace-dotnet/pull/7989

### DIFF
--- a/tracer/src/Datadog.Trace/Debugger/Instrumentation/AsyncLineDebuggerInvoker.cs
+++ b/tracer/src/Datadog.Trace/Debugger/Instrumentation/AsyncLineDebuggerInvoker.cs
@@ -9,7 +9,6 @@ using System.Runtime.CompilerServices;
 using Datadog.Trace.Debugger.Expressions;
 using Datadog.Trace.Debugger.Helpers;
 using Datadog.Trace.Debugger.Instrumentation.Collections;
-using Datadog.Trace.Debugger.RateLimiting;
 using Datadog.Trace.Logging;
 
 namespace Datadog.Trace.Debugger.Instrumentation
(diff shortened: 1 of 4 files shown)


tracer/src/Datadog.Trace/Debugger/Instrumentation/AsyncLineDebuggerInvoker.cs
```diff
@@ -47,6 +46,20 @@ public static void LogLocal<TLocal>(ref TLocal local, int index, ref AsyncLineDe
                     return;
                 }
 
+                if (Datadog.Trace.VendoredMicrosoftCode.System.Runtime.CompilerServices.Unsafe.Unsafe.IsNullRef(ref local))
```
GreenMatan, line 49: Do you do that to avoid NRE? did it impact the structure of the snapshot itself when we encountered it?

dudikeleti: Yes for both. For the second part this PR should be merged together with #7988 ,
 #7989

PR #7991, opened 2025-12-20
[EL] DEBUG-4878 Fix null equality in expression language

dudikeleti:
## Summary of changes
Harden binary equality handling (`==` / `!=`) in the probe expression parser to avoid runtime exceptions and produce predictable results when comparing:
- non-nullable value types to `null`
- reference types to `null` when the runtime type differs from the compile-time type

## Implementation details
- Route `==` and `!=` through a dedicated `EqualExpression()` helper.
- For `non-nullable value type` vs `null`:
  - return a constant expression (`false` for `==`, `true` for `!=`) since the comparison outcome is deterministic.
- For `reference type` vs `null`:
  - use `Expression.ReferenceEqual` on `object`-converted operands to ensure object-reference semantics and avoid type-cast issues.

## Test coverage
Snapshot Exploration Test
ProbeExpressionParser_NonNullableValueTypeComparedToNull_DoesNotThrow

## Other details

### DIFF
--- a/tracer/test/Datadog.Trace.Tests/Debugger/DebuggerExpressionLanguageTests.cs
+++ b/tracer/test/Datadog.Trace.Tests/Debugger/DebuggerExpressionLanguageTests.cs
@@ -137,6 +137,52 @@ namespace Datadog.Trace.Tests.Debugger
             Assert.True(compiled.Errors == null || compiled.Errors.Length == 0);
         }
 
+        [Fact]
+        public void ProbeExpressionParser_NonNullableValueTypeComparedToNull_DoesNotThrow()
+        {
+            var scopeMembers = CreateScopeMembers();
+
+            const string equalsJson = """
+                                      {
+                                        "eq": [
+                                          { "ref": "IntLocal" },
+                                          null
+                                        ]
+                                      }
+                                      """;
+
+            const string notEqualsJson = """
+                                         {
+                                           "ne": [
+                                             { "ref": "IntLocal" },
+                                             null
+                                           ]
+                                         }
+                                         """;
+
+            var equalsCompiled = ProbeExpressionParser<bool>.ParseExpression(equalsJson, scopeMembers);
+            var equalsResult = equalsCompiled.Delegate(
+                scopeMembers.InvocationTarget,
+                scopeMembers.Return,
+                scopeMembers.Duration,
+                scopeMembers.Exception,
+                scopeMembers.Members);
+
+            Assert.False(equalsResult);
+            Assert.True(equalsCompiled.Errors == null || equalsCompiled.Errors.Length == 0);
+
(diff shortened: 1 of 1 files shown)


tracer/src/Datadog.Trace/Debugger/Expressions/ProbeExpressionParser.Binary.cs
```diff
@@ -75,8 +75,7 @@ private Expression BinaryOperation(JsonTextReader reader, List<ParameterExpressi
                 ">=" => Expression.GreaterThanOrEqual(left, right),
                 "<" => Expression.LessThan(left, right),
                 "<=" => Expression.LessThanOrEqual(left, right),
-                "==" => Expression.Equal(left, right),
-                "!=" => Expression.NotEqual(left, right),
+                "==" or "!=" => EqualExpression(),
```
tylfin, line 78: Should we add a failing unittest for the previous behaviour?

GreenMatan: +1, and also for the new behavior

tracer/src/Datadog.Trace/Debugger/Expressions/ProbeExpressionParser.Binary.cs
```diff
@@ -85,6 +84,38 @@ private Expression BinaryOperation(JsonTextReader reader, List<ParameterExpressi
             AddError($"{left?.ToString() ?? "N/A"} {operand} {right?.ToString() ?? "N/A"}", e.Message);
             return ReturnDefaultValueExpression();
         }
+
+        Expression EqualExpression()
+        {
+            // Check if comparing non-nullable value type to null constant
+            if ((right is ConstantExpression { Value: null } && left.Type.IsValueType && Nullable.GetUnderlyingType(left.Type) == null) ||
+                (left is ConstantExpression { Value: null } && right.Type.IsValueType && Nullable.GetUnderlyingType(right.Type) == null))
```
GreenMatan, line 92: Hmm, shouldn't `EqualExpression` accept `right`, `left` and `operand` args?

tracer/src/Datadog.Trace/Debugger/Expressions/ProbeExpressionParser.Binary.cs
dudikeleti: e33f17c8abada24ad75c2d85ed0f9e6af9428624

tracer/src/Datadog.Trace/Debugger/Expressions/ProbeExpressionParser.Binary.cs
dudikeleti: local func

PR #7992, opened 2025-12-20
[EL] DEBUG-4880 Align probe expression compilation with runtime types

dudikeleti:
## Summary of changes
Align probe expression compilation with runtime types for `this` and args/locals to prevent type-mismatch runtime failures and keep expression-cache behavior consistent.

## Reason for change
Expression compilation previously relied on declared types (`ScopeMember.Type`) for `this` and for args/locals. In practice, the expression cache key is based on runtime types, and values can differ from their declared types (e.g., declared `Int32` but actual value `Int64`, or declared base type but runtime derived type). This could lead to cached expressions being compiled with one set of types but executed with another, causing runtime `InvalidCastException` during lambda execution.

## Implementation details
- For args/locals:
  - Create variables using `Value.GetType()` (when available) and fall back to the declared type otherwise.
  - Use the same runtime type in the `Expression.Convert` when reading `ScopeMember.Value`.
- For `this`:
  - Resolve `thisType` once at the entry point (`ParseExpression`) preferring runtime type (`InvocationTarget.Value?.GetType()`), then pass it down via `thisTypeOverride`.
  - This ensures the same `this` type is used consistently throughout parsing/compilation.

## Test coverage
Snapshot Exploration Test

## Other details

### DIFF
--- a/tracer/test/Datadog.Trace.Tests/Debugger/ProbeExpressionsResources/Approvals/DebuggerExpressionLanguageTests.AccessNullObject.verified.txt
+++ b/tracer/test/Datadog.Trace.Tests/Debugger/ProbeExpressionsResources/Approvals/DebuggerExpressionLanguageTests.AccessNullObject.verified.txt
@@ -35,7 +35,7 @@ Expressions:
     var BooleanValue = (bool)scopeMemberArray[11].Value;
     var Char = (char)scopeMemberArray[12].Value;
     var AnotherChar = (char)scopeMemberArray[13].Value;
-    var NullableNotNullValueLocal = (Guid?)scopeMemberArray[14].Value;
+    var NullableNotNullValueLocal = (Guid)scopeMemberArray[14].Value;
     var NullableNullValueLocal = (Guid?)scopeMemberArray[15].Value;
     var IntArg = (int)scopeMemberArray[16].Value;
     var DoubleArg = (double)scopeMemberArray[17].Value;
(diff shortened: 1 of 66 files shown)


tracer/src/Datadog.Trace/Debugger/Expressions/ProbeExpressionParser.cs
```diff
@@ -470,7 +470,13 @@ private void AddLocalAndArgs(ScopeMember[] argsOrLocals, List<ParameterExpressio
...
 
             var argOrLocal = argsOrLocals[index];
-            var variable = Expression.Variable(argOrLocal.Type, argOrLocal.Name);
+
+            // CRITICAL: Use runtime type (Value.GetType()) instead of declared type (Type).
+            // The compilation must use the same types as runtime.
+            // If declared type is Int32 but actual value is Int64, using declared type
+            // would cause InvalidCastException when the lambda is executed.
+            var runtimeType = argOrLocal.Value?.GetType() ?? argOrLocal.Type;
+            var variable = Expression.Variable(runtimeType, argOrLocal.Name);
```
GreenMatan, line 487: We compile the expression only once and reuse it, so the type we end up getting in this line might differ.
This is a general issue we have with our expressions - we compile them using the args/locals that happened to be in there in the first invocation but that might (and usually is) different from subsequent invocations.
I'm not suggesting to do a revamp right now, but as a note to keep in mind.

GreenMatan (approved):
Left a comment

tracer/src/Datadog.Trace/Debugger/Expressions/ProbeExpressionParser.cs
GreenMatan: I see there's an effort to overcome this shortcoming in #7994. Great.

tracer/src/Datadog.Trace/Debugger/Expressions/ProbeExpressionParser.cs
```diff
@@ -470,7 +470,13 @@ private void AddLocalAndArgs(ScopeMember[] argsOrLocals, List<ParameterExpressio
...
             scopeMembers.Add(variable);
 
             expressions.Add(
```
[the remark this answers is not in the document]
dudikeleti: You’re right, but we’re handling this in #7994

PR #7993, opened 2025-12-20
[EL] DEBUG-4884 Fix null expression when type is struct

dudikeleti:
## Summary of changes
Prevent null-unboxing failures when binding `ScopeMember.Value` into value-type locals during probe expression compilation.

## Reason for change
When the target variable type is a non-nullable value type (struct), `Expression.Convert(valueField, type)` attempts to unbox `null`, which throws at execution time. This caused probe evaluation to fail even when a safe default value would be acceptable.

## Implementation details
- Build the assignment expression differently for value types vs reference types:
  - **Value types:** emit a conditional expression:
    - if `ScopeMember.Value == null` → assign `default(TValue)`
    - else → unbox via `Convert(object, TValue)`
  - **Reference types:** keep the existing `Convert`, which already handles null correctly.
- This change keeps behavior unchanged for valid values and avoids runtime `NullReferenceException` for null value-type bindings.

## Test coverage
Snapshot Exploration Test
ProbeTests.RecursionWithInnerRefStructTest
ProbeExpressionParser_ValueTypeNull_UsesDefaultValue

## Other details

### DIFF
--- a/tracer/test/Datadog.Trace.Tests/Debugger/DebuggerExpressionLanguageTests.cs
+++ b/tracer/test/Datadog.Trace.Tests/Debugger/DebuggerExpressionLanguageTests.cs
@@ -183,6 +183,35 @@ namespace Datadog.Trace.Tests.Debugger
             Assert.True(notEqualsCompiled.Errors == null || notEqualsCompiled.Errors.Length == 0);
         }
 
+        [Fact]
+        public void ProbeExpressionParser_ValueTypeNull_UsesDefaultValue()
+        {
+            // Arrange
+            var scopeMembers = CreateScopeMembers();
+            scopeMembers.Duration = new ScopeMember("@duration", typeof(TimeSpan), null, ScopeMemberKind.Duration);
+
+            const string json = """
+                                {
+                                  "ref": "@duration"
+                                }
+                                """;
+
+            // Act
+            var compiled = ProbeExpressionParser<object>.ParseExpression(json, scopeMembers);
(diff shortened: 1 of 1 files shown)


tracer/src/Datadog.Trace/Debugger/Expressions/ProbeExpressionParser.cs
```diff
@@ -578,7 +578,28 @@ private ParameterExpression AddParameterAndVariable(ScopeMember scopeMember, Typ
...
+                Expression.Default(type),
+                Expression.Convert(valueField, type));
+        }
+        else
+        {
+            // Reference types: Convert handles null correctly, throws on type mismatch
+            assignmentValue = Expression.Convert(valueField, type);
+        }
+
+        expressions.Add(Expression.Assign(variable, assignmentValue));
```
GreenMatan, line 602: Can you add a test that validates it works?

GreenMatan (approved):
left a comment

tracer/src/Datadog.Trace/Debugger/Expressions/ProbeExpressionParser.cs
tylfin: +1

dudikeleti: 5d12f5e941e76b2c093afc8373a91b48070d4578

PR #7994, opened 2025-12-20
[EL] DEBUG-4879 Support multiple expression compilations

dudikeleti:
## Summary of changes
- Reworked `ProbeExpressionEvaluator` to cache compiled expressions **per runtime type combination** (invocation target, return value, and scope members), instead of a single `Lazy<>` instance shared across all invocations.
- Introduced `ExpressionCacheKey` to uniquely identify a compilation context using **runtime types** (`Value.GetType()` when available) and member runtime types to support polymorphic calls.
- Replaced single cached fields with `ConcurrentDictionary<ExpressionCacheKey, …>` caches for:
  - Templates (`CompiledExpression<string>[]?`)
  - Condition (`CompiledExpression<bool>?`)
  - Metric (`CompiledExpression<double>?`)
  - Span decorations (compiled decorations array)
- Preserved evaluation semantics and error aggregation, while ensuring compiled delegates match the actual runtime types.

## Reason for change
The previous implementation compiled expressions once (on first invocation) and reused them for all subsequent calls. This fails in polymorphic scenarios where:
- The declared type differs from the runtime type (e.g., generics/base class/interface method invoked on a derived instance).
- Parameters/locals are declared as a base type (or `object`) but carry different concrete runtime types across invocations.

In these cases, delegates compiled against the first observed type can fail later (typically invalid casts / expression binding mismatches), causing evaluation errors and incorrect probe results.

## Implementation details
- Compute a **runtime-aware cache key** on each `Evaluate()` call:
  - `thisType`: `InvocationTarget.Value?.GetType()` (fallback to declared type, then `typeof(object)`)
  - `returnType`: `Return.Value?.GetType()` (fallback to declared type)
  - `members`: runtime types captured from `ScopeMember.Value?.GetType()` (fallback to declared `ScopeMember.Type`)
- Store compiled artifacts in thread-safe caches:
  - `ConcurrentDictionary<ExpressionCacheKey, ...>`
  - On cache miss: compile, then `TryAdd`
- Added `ExpressionCacheKey` struct:
  - Precomputes hash over `thisType`, `returnType`, and member runtime types.
  - Implements equality by comparing `ThisType`, `ReturnType`, and the captured member type sequence (hash used as a fast-path).
- Added debug logging for cache misses (type + hash + cache size) to help diagnose compilation churn and verify caching behavior.

## Test coverage
- Existing `DebuggerExpressionLanguageTests` remain supported via `CompiledTemplates` / `CompiledCondition` accessors (now returning the first cached entry).
- Snapshot Exploration Test

## Dependencies
- Depends on: #7991, #7992, #7993
- Merge order: #7991 → #7992 → #7993 → (this PR)

## Other details
- Caches may grow with many unique type combinations; however this is required for correctness under polymorphic workloads. If needed later, we can add eviction/size caps.
- This PR is part of an effort to make the Snapshot Exploration Test run successfully end-to-end.

### DIFF
--- a/tracer/src/Datadog.Trace.Trimming/build/Datadog.Trace.Trimming.xml
+++ b/tracer/src/Datadog.Trace.Trimming/build/Datadog.Trace.Trimming.xml
@@ -1065,7 +1065,6 @@
       <type fullname="System.Threading.ReaderWriterLockSlim" />
       <type fullname="System.Threading.SemaphoreSlim" />
       <type fullname="System.Threading.SpinLock" />
-      <type fullname="System.Threading.SpinWait" />
       <type fullname="System.Threading.SynchronizationContext" />
       <type fullname="System.Threading.ThreadLocal`1" />
       <type fullname="System.Threading.Volatile" />


tracer/src/Datadog.Trace/Debugger/Expressions/ProbeExpressionEvaluator.cs
```diff
@@ -90,31 +109,63 @@ internal CompiledExpression<double>? CompiledMetric
...
+        // The declared type can differ from runtime type when:
+        // - Method is on a base class/interface but called on derived type
+        // - Parameter is declared as base type but actual value is derived
+        // Without this, expressions compiled for declared type would fail to cast actual values.
+        var invocationTarget = scopeMembers.InvocationTarget;
+
+        // CRITICAL: Always use Value.GetType() for runtime type. If Value is null,
+        // we use the declared Type, but this can cause cache collisions when different
+        // runtime types share the same declared type (e.g., all inherit from Object).
+        var thisRuntimeType = invocationTarget.Value?.GetType();
```
GreenMatan: `invocationTarget` can be null?

tracer/src/Datadog.Trace/Debugger/Expressions/ExpressionCacheKey.cs
```diff
@@ -0,0 +1,126 @@
...
+        // Capture RUNTIME types of members (Value.GetType()), not declared types
+        // This is critical for polymorphic scenarios where declared type is Object/base
+        // but actual values are different concrete types
+        var hash = new HashCode();
+        hash.Add(thisType);
+        hash.Add(returnType);
+
+        if (members != null && members.Length > 0)
+        {
+            var types = new Type?[members.Length];
```
GreenMatan: can we minimize this allocation by reusing from a type pool? I'm worried we incur additional short-live allocation that might result in GC pressure if the method executes lots of time

tracer/src/Datadog.Trace/Debugger/Expressions/ProbeExpressionEvaluator.cs
```diff
@@ -20,15 +20,14 @@ internal sealed class ProbeExpressionEvaluator
...
-
-    private int _expressionsCompiled;
+    // Per-type caching using ConcurrentDictionary.
+    // Cache key includes thisType AND member types to handle polymorphic args/locals.
+    // When same method is called with different arg types (e.g., Stream declared but sometimes LimitedInputStream, sometimes MemoryStream),
+    // each combination gets its own compiled expression.
+    private readonly ConcurrentDictionary<ExpressionCacheKey, CompiledExpression<string>[]?> _compiledTemplatesByType = new();
+    private readonly ConcurrentDictionary<ExpressionCacheKey, CompiledExpression<bool>?> _compiledConditionByType = new();
+    private readonly ConcurrentDictionary<ExpressionCacheKey, CompiledExpression<double>?> _compiledMetricByType = new();
+    private readonly ConcurrentDictionary<ExpressionCacheKey, KeyValuePair<CompiledExpression<bool>, KeyValuePair<string?, CompiledExpression<string>[]>[]>[]?> _compiledDecorationsByType = new();
```
GreenMatan: Can we improve this slow path of inquiring a ConcurrentDictionary, perhaps by using a Reader/Writer (slim) lock or other lock free approach?
We touch them in hot paths

tracer/src/Datadog.Trace/Debugger/Expressions/ProbeExpressionEvaluator.cs
dudikeleti: static methods

tracer/src/Datadog.Trace/Debugger/Expressions/ExpressionCacheKey.cs
dudikeleti: The problem that we can't pool them if we want to use them as a long lived cache key, I made a few changes to the implementation to reduce allocations in most cases

tracer/src/Datadog.Trace/Debugger/Expressions/ProbeExpressionEvaluator.cs
dudikeleti: I made a few changes to the implementation to reduce dictionary access, but overall I don’t think it’s worth it.

tracer/src/Datadog.Trace/Debugger/Expressions/ExpressionCacheKey.cs
GreenMatan: The creation of `ExpressionCacheKey` (and the Type array in it) is done every time we try to evaluate an expression - in other words, whenever the probe executes. Keeping it in memory for long is true only for the `ExpressionCacheKey` that ends up being saved in the dictionary. We could think about an approach that takes the array from the pool for lookup, and materialize it as heap allocated array _only_ when the key doesn't exist (first time we encounter the specific type combination) while all subsequent executions will take & return the array from the pool. My concern is that we may induce GC pressure for short lived arrays

tracer/src/Datadog.Trace/Debugger/Expressions/ProbeExpressionEvaluator.cs
GreenMatan: Let's take it offline

tracer/src/Datadog.Trace/Debugger/Expressions/ProbeExpressionEvaluator.cs
GreenMatan: I mean, could we end up with NRE if we dereference `invocationTarget.X` throughout this method?

tracer/src/Datadog.Trace/Debugger/Expressions/ExpressionCacheKey.cs
dudikeleti: It turns out that I didn’t push my last commit. `ExpressionCacheKey` no longer exists.

PR #7996, opened 2025-12-22
Revert changes to libdatadog

andrewlock:
## Summary of changes

Reverts recent profiler and libdatadog changes that were causing a crash on macOS

## Reason for change

The changes to the libdatadog integration are causing a crash in macOS instrumentation (confirmed for arm64)

## Implementation details

Revert the recent changes (and dependent changes)

## Test coverage

@tonyredondo will test locally - we clearly have a gap in our coverage here

## Other details

I reverted the profiler changes too, as I suspected they were dependent, but I'm not sure


tonyredondo:
the fix works 👍🏻 

image

Before: 

image

andrewlock:
This was a red herring

PR #8002, opened 2025-12-23
Retry on profiler failed integration jobs

NachoEchevarria:
## Summary of changes

The following jobs we detected as flaky (pipeline failures / last month):
```
profiler_integration_tests_linux | Test alpine | 34
profiler_integration_tests_windows | Win x86 | 7
profiler_integration_tests_windows | Win x64 | 6
```

The complete flakiness report can be found here:
[https://docs.google.com/spreadsheets/d/1Gftmhb-66Dag4qFEXw9tyXp7U0fOQsdCE2gI-1voWyA/edit?gid=1708590243#gid=1708590243](https://docs.google.com/spreadsheets/d/1Gftmhb-66Dag4qFEXw9tyXp7U0fOQsdCE2gI-1voWyA/edit?gid=1708590243#gid=1708590243)

The affected jobs will be retried on error. This PR can be reverted after handling flaky errors.

### DIFF
--- a/.azure-pipelines/ultimate-pipeline.yml
+++ b/.azure-pipelines/ultimate-pipeline.yml
@@ -2723,6 +2723,7 @@ stages:
 
     - script: tracer\build.cmd BuildProfilerSamples BuildAndRunProfilerIntegrationTests --TargetPlatform $(targetPlatform)
       displayName: Run Profiler Integration tests
+      retryCountOnTaskFailure: 2
       env:
         DD_LOGGER_DD_API_KEY: $(ddApiKey)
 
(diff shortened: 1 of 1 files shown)


.azure-pipelines/ultimate-pipeline.yml
```diff
@@ -2723,6 +2723,7 @@ stages:
 
     - script: tracer\build.cmd BuildProfilerSamples BuildAndRunProfilerIntegrationTests --TargetPlatform $(targetPlatform)
       displayName: Run Profiler Integration tests
+      retryCountOnTaskFailure: 2
```
andrewlock, line 2726: I don't think this will work in isolation 🤔 These currently take a long time to run, and doubling the length with likely just cause timeouts...

I think we should probably
- Split out the "build" commands here from the "run" commands
- Add a bunch of retries to those tasks
- Split by target framework as well as target platform (primarily for perf)
- See if we can add features to their custom runner to add the retries
  - If we can't, retry the whole target-framework task: these should be short enough that we _can_ retry the whole lot without hitting timeouts

PR #8003, opened 2025-12-23
Skip some errors from being sent to telemetry

bouwkast:
## Summary of changes

Skips a few errors that are `Ignored` in Error Tracking and shouldn't really be sent to telemetry as we can't act on them.

## Reason for change

Nothing we can really do about these.

## Implementation details

Marked as `ErrorSkipTelemetry`

## Test coverage

N/A

## Other details

I guess one thing we do lose here is whether we _want_ to see how often these happen?

### DIFF
--- a/tracer/src/Datadog.Trace/Ci/Coverage/Util/CoverageUtils.cs
+++ b/tracer/src/Datadog.Trace/Ci/Coverage/Util/CoverageUtils.cs
@@ -73,7 +73,7 @@ internal static class CoverageUtils
             var jsonFiles = Directory.GetFiles(inputFolder, "*.json", SearchOption.TopDirectoryOnly);
             if (jsonFiles.Length == 0)
             {
-                Log.Error("'{InputFolder}' doesn't contain any json file.", inputFolder);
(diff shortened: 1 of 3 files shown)


tracer/src/Datadog.Trace/Ci/Coverage/Util/CoverageUtils.cs
```diff
@@ -71,7 +71,7 @@ private static bool TryCombineAndGetTotalCoverage(string? inputFolder, out Globa
             var jsonFiles = Directory.GetFiles(inputFolder, "*.json", SearchOption.TopDirectoryOnly);
             if (jsonFiles.Length == 0)
             {
-                Log.Error("'{InputFolder}' doesn't contain any json file.", inputFolder);
+                Log.ErrorSkipTelemetry("'{InputFolder}' doesn't contain any json file.", inputFolder);
```
andrewlock, line 76: @tonyredondo is this actually an error, or should it just be a warning? (`ErrorSkipTelemetry` is obvs fine for now, unless we hear otherwise)

andrewlock (approved):
Thanks!

PR #8011, opened 2025-12-24
Automatic retry ASM integration tests under failure

NachoEchevarria:
## Summary of changes

The following jobs have been found flaky, mostly due to a already reported crash condition:

```
integration_tests_windows.Win x86_net9.0_ASM  7 failures / last month
integration_tests_linux.Test alpine_net7.0_ASM  5 failures / last month
integration_tests_linux.Test alpine_net9.0_ASM  5 failures / last month
integration_tests_linux.Test debian_net9.0_ASM  3 failures / last month
```

The complete flakiness report can be found here:
https://docs.google.com/spreadsheets/d/1Gftmhb-66Dag4qFEXw9tyXp7U0fOQsdCE2gI-1voWyA/edit?gid=1708590243#gid=1708590243

The affected jobs will automatically be retried once to avoid CI flakiness.

For other teams, the flaky tests have been marked as flaky, which causes automatic retry. In the case of ASM, the failure does not occur on especific tests, so marking some of them would not solve the issue. This change can be reverted once the jobs are more stable.

This PR is part of an initiative of marking the most flaky tests or jobs of all the teams.


.azure-pipelines/ultimate-pipeline.yml
```diff
@@ -1710,12 +1710,25 @@ stages:
...
+        DD_LOGGER_DD_API_KEY: $(ddApiKey)
+        enable_crash_dumps: true
+        Filter: $(IntegrationTestFilter)
+        SampleName: $(IntegrationTestSampleName)
+        Area: $(area)
+
+    - script: tracer\build.cmd RunIntegrationTests RunWindowsRegressionTests -Framework $(framework) --code-coverage-enabled $(CodeCoverageEnabled)
+      displayName: Run integration tests (ASM)
+      condition: eq(variables['area'], 'ASM')
+      retryCountOnTaskFailure: 2
```
andrewlock, line 1725: Gah, it's so ugly that we have to split these jobs, but I think that's the best we can do 😅

.azure-pipelines/ultimate-pipeline.yml
```diff
@@ -1710,12 +1710,25 @@ stages:
...
     - script: tracer\build.cmd RunIntegrationTests RunWindowsRegressionTests -Framework $(framework) --code-coverage-enabled $(CodeCoverageEnabled)
-      displayName: Run integration tests
+      displayName: Run integration tests (Tracer)
+      condition: ne(variables['area'], 'ASM')
+      env:
+        DD_LOGGER_DD_API_KEY: $(ddApiKey)
+        enable_crash_dumps: true
+        Filter: $(IntegrationTestFilter)
+        SampleName: $(IntegrationTestSampleName)
+        Area: $(area)
```
andrewlock, line 1720: nit: I don't think we need this variable?
```suggestion
```

.azure-pipelines/ultimate-pipeline.yml
```diff
@@ -1710,12 +1710,25 @@ stages:
...
+    - script: tracer\build.cmd RunIntegrationTests RunWindowsRegressionTests -Framework $(framework) --code-coverage-enabled $(CodeCoverageEnabled)
+      displayName: Run integration tests (ASM)
+      condition: eq(variables['area'], 'ASM')
+      retryCountOnTaskFailure: 2
       env:
         DD_LOGGER_DD_API_KEY: $(ddApiKey)
         enable_crash_dumps: true
         Filter: $(IntegrationTestFilter)
         SampleName: $(IntegrationTestSampleName)
+        Area: $(area)
```
andrewlock, line 1731: nit: I don't think we need this variable?
```suggestion
```

.azure-pipelines/ultimate-pipeline.yml
```diff
@@ -2272,7 +2285,26 @@ stages:
...
+    - script: |
+        docker-compose -f docker-compose.yml -p $(DockerComposeProjectName) \
+          run --no-deps --rm \
+          -e baseImage=$(baseImage) \
+          -e framework=$(publishTargetFramework) \
+          -e CodeCoverageEnabled=$(CodeCoverageEnabled) \
+          -e IncludeTestsRequiringDocker=false  \
+          -e Filter=$(IntegrationTestFilter) \
+          -e SampleName=$(IntegrationTestSampleName) \
+          -e Area=$(area) \
```
andrewlock, line 2303: nit: I don't think we need this variable?
```suggestion
```

.azure-pipelines/ultimate-pipeline.yml
NachoEchevarria: I would say it's still needed to exclude non tracer tests in this run. We are still using  --filter (Area=Tracer) when building the final command line.

PR #8022, opened 2025-12-30
Fix invalid telemetry when using manual config in code with dictionaries

andrewlock:
## Summary of changes

Fixes a bug in telemetry that was recording invalid values for dictionary values coming from config-in-code

## Reason for change

When looking through some memory dumps, noticed a bunch of strings that looked like `"DD_TAGS:System.Collections.Generic.Dictionary``2[System.String,System.String]"` and tracked them down to the manual config, and ultimately to a bug in the `DictionaryObjectConfigurationSource.AsDictionary()` methods.

## Implementation details

- Write some unit tests to track down the problem. Show them failing with above string.
- Minor refactor of `CompositeConfigurationSource.AsDictionary()` to reduce duplication
- Fix bug in `DictionaryObjectConfigurationSource`
- Simplify (unused except in error) fallback path in `CompositeConfigurationSource.AsDictionary()`

## Test coverage

Added unit tests to demonstrate bug

### DIFF
--- a/tracer/src/Datadog.Trace/Configuration/ConfigurationSources/DictionaryObjectConfigurationSource.cs
+++ b/tracer/src/Datadog.Trace/Configuration/ConfigurationSources/DictionaryObjectConfigurationSource.cs
@@ -6,10 +6,10 @@
 #nullable enable
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using Datadog.Trace.Configuration.ConfigurationSources.Telemetry;
 using Datadog.Trace.Configuration.Telemetry;
 using Datadog.Trace.Telemetry;
+using Datadog.Trace.Util;
 
 namespace Datadog.Trace.Configuration;
 
@@ -138,11 +138,23 @@ internal class DictionaryObjectConfigurationSource : IConfigurationSource
                 return ConfigurationResult<IDictionary<string, string>>.ParseFailure();
             }
 
-#if NETCOREAPP
-            var dictAsString = string.Join(separator, value.Select(x => $"{x.Key}:{x.Value}"));
-#else
-            var dictAsString = string.Join($"{separator}", value.Select(x => $"{x.Key}:{x.Value}"));
-#endif
+            var dictAsString = string.Empty;
+            if (value.Count > 0)
+            {
+                var sb = StringBuilderCache.Acquire();
+                foreach (var kvp in value)
+                {
+                    sb.Append(kvp.Key)
+                      .Append(':')
+                      .Append(kvp.Value)
+                      .Append(separator);
+                }
+
+                // remove the final separator (we know there was at least one so this is safe)
+                sb.Remove(sb.Length - 1, 1);
+                dictAsString = StringBuilderCache.GetStringAndRelease(sb);
+            }
+
             if (validator is null || validator(value))
             {
                 telemetry.Record(key, dictAsString, recordValue: true, Origin);


tracer/src/Datadog.Trace/Configuration/ConfigurationSources/DictionaryObjectConfigurationSource.cs
```diff
@@ -138,7 +138,11 @@ public ConfigurationResult<IDictionary<string, string>> GetDictionary(string key
                 return ConfigurationResult<IDictionary<string, string>>.ParseFailure();
             }
 
-            var dictAsString = string.Join($"{separator}", value.Select(x => $"{key}:{value}"));
```
andrewlock, line 141: This is the fix: `$"{key}:{value}` => `$"{x.Key}:{x.Value}`

I just couldn't stand to see `$"{separator}"`, hence the `#if` 😅

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/CompositeConfigurationSource.cs
```diff
@@ -189,38 +189,17 @@ public ConfigurationResult<bool> GetBool(string key, IConfigurationTelemetry tel
 
         /// <inheritdoc />
         public ConfigurationResult<IDictionary<string, string>> GetDictionary(string key, IConfigurationTelemetry telemetry, Func<IDictionary<string, string>, bool>? validator)
```
andrewlock, line 191: These methods were essentially identical except for the `GetDictionary()` call (which all called different overloads)

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/CompositeConfigurationSource.cs
```diff
@@ -277,7 +237,8 @@ public ConfigurationResult<IDictionary<string, string>> GetDictionary(string key
 
             if (result.IsValid && !isLastFound)
             {
-                telemetry.Record(key, result.TelemetryOverride ?? result.Result?.ToString(), recordValue: true, origin);
+                // there should always be a telemetry override by convention, so just record a sentinel for now if there's not for some reason
+                telemetry.Record(key, result.TelemetryOverride ?? "<MISSING>", recordValue: true, origin);
```
andrewlock, line 241: We should never hit this path, but if we do, calling `ToString()` on a dictionary is pointless. We _could_ serialize it properly, but seeing as this shouldn't ever be hit (we should always have a fallback), it doesn't seem worth duplicating here to me

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/DictionaryObjectConfigurationSource.cs
lucaspimentel: Should we replace LINQ with a `foreach`+`StringBuilderCache` while we're at it?

lucaspimentel (approved):

tracer/src/Datadog.Trace/Configuration/ConfigurationSources/DictionaryObjectConfigurationSource.cs
andrewlock: Good point. This is only going to be hit when customers are doing manual updates, but ther'es not a good reason _not_ to do that, so will make the fix, thanks!

anna-git (approved):

PR #8025, opened 2026-01-01
[Dynamic Instrumentation] Disable AppDomain related log flooding on .NET Framework

GreenMatan:
## Summary of changes
In legacy .NET Framework, a single process can host multiple logically isolated applications ("AppDomains") managed by the runtime. The machine code produced from jitting a method is used among the app domains. Thus, instrumentation is also being shared. The runtime isolates those different domains and prevents data sharing (that is only possible through marshalling/proxying). Each probe has metadata associated with it that is either received through the network (Remote Configuration) or created locally as in Exception Replay. This metadata is kept in memory and is crucial for the instrumentation to function correctly. When a method with an instrumentation is executed in a different domain from the one that kept this metadata, it can't retrieve it, thus bails out with a log message.
If such method executes many times, we may log plenty of time, result in log files rotation and flooding the disk. We have a customer that complaint about it.

As a quick fix for now, I'm disabling those log messages on .NET Framework.

## Reason for change
Flooding log files with the log message `BeginMethod_StartMarker: Failed to receive the ProbeData associated with the executing probe <...>`

## Implementation details
Disabling the log message on .NET Framework using a preprocessor directive that prevents the compilation of this log message on .NET Framework builds.


andrewlock (approved):
Thanks!

Issue #5862, opened 2024-08-07
Resource leak with DD_RUNTIME_METRICS_ENABLED=true


npubl629:
**Describe the bug**
A resource leak with DD_RUNTIME_METRICS_ENABLED=true. This bug is most apparent when viewing the CPU usage of a service that has been running for several weeks without a reboot. It is unclear whether or not there is a memory leak associated with this bug.

**To Reproduce**
1. View [this repo](https://github.com/npubl629/dd-trace/v1)
2. Follow steps in the readme to start multiple docker services for which we can compare resource usage.
3. Observe the CPU usage over several weeks. Services with DD_RUNTIME_METRICS_ENABLED begin consuming relatively more CPU than the ones without DD_RUNTIME_METRICS_ENABLED.

**Expected behavior**
The CPU and memory usage of this library should be stable when the service is idle or serving basic status endpoints.

**Screenshots**
I have had the services from the example repo running for over 3 weeks and the issue is fairly clear. See [this public datadog dashboard](https://p.datadoghq.com/sb/0770c150-3108-11ef-af83-caa90115f9de-4945ba0c8efa9af723fac32b609963e6?fromUser=true&refresh_mode=sliding&from_ts=1721061462690&to_ts=1721666262690&live=true) where CPU is steadily rising for the containers trace-metrics-idle-1 and trace-metrics-healthcheck-1. Unfortunately the trace-nometrics-idle-1 container never started, but the difference in CPU usage for the three healthcheck containers makes the issue fairly clear.

![Chart of CPU Usage Difference](https://raw.githubusercontent.com/npubl629/dd-trace/master/v1/postprocessing/cpu-diff.png)
![Chart of CPU Usage](https://raw.githubusercontent.com/npubl629/dd-trace/master/v1/postprocessing/cpu-usage.png)
![Chart of Memory Usage](https://raw.githubusercontent.com/npubl629/dd-trace/master/v1/postprocessing/mem-diff.png)

This chart is the difference between container CPU usage of the three healthcheck containers. Note the blue and green lines are increasing, while the orange line is fairly stable. Also note the CPU usage of the idle service (serving no requests) has more than doubled in three weeks.

![image](https://github.com/user-attachments/assets/986594cb-ebad-48b0-b6ac-26fdf4f50594)

**Runtime environment:**
- Instrumentation mode: NuGet package
[...]

PR #8027, opened 2026-01-02
Create `System.Diagnostics.Metrics`-based runtime metrics listener

andrewlock:
## Summary of changes

Creates a .NET 6+ only implementation of `IRuntimeMetricsListener` that uses the `System.Diagnostics.Metrics` (and other) APIs

## Reason for change

.NET Core (probably all versions, but at least .NET 6+)  has a memory leak with the event pipes, which means if we enable runtime metrics, we likely have a slow memory leak 😬 [This was raised ~1 year ago with .NET team](https://github.com/dotnet/runtime/issues/111368), specifically citing dd-trace-dotnet. but doesn't have a fix yet. Also a PR has been open on the .NET repo with a tentative fix for ~2 months, so _at best_ this _might_ be fixed in .NET 11.

Separately, the `System.Diagnostics.Metrics` APIs were introduced in .NET 6, with support for aspnetcore-based metrics added in .NET 8, and support for "runtime" metrics in .NET 9.

This PR introduces a new (experimental for now) `IRuntimeMetricsListener` implementation that doesn't use `EventListener`, and instead uses the `System.Diagnostics.Metrics` APIs, aiming to provide essentially the same runtime metrics we currently do, just using a different source.

## Implementation details

- Created a new `IRuntimeMetricsListener` implementation, `DiagnosticsMetricsRuntimeMetricsListener`
- Added a config to enable it in .NET 6+ only, `DD_RUNTIME_METRICS_DIAGNOSTICS_METRICS_API_ENABLED`
  - Open to suggestions here. Other options include having an "enum" type for listener instead of just this one. That's harder to consume for customers, but more extensible theoretically.
- Added tests

To give as wide compatibility as possible, and to avoid any additional overhead, whenever the built-in runtime metrics use existing APIs (e.g. via `GC` calls), we use those instead of the metrics. 

In summary:

Thread metrics:
- `runtime.dotnet.threads.workers_count`: via `ThreadPool.ThreadCount` (same as `RuntimeEventListener`)
- `runtime.dotnet.threads.contention_count: via `Monitor.LockContentionCount`

GC metrics:
- `runtime.dotnet.gc.size.gen#` from info in `GC.GetGCMemoryInfo()`, which mirrors [the built-in approach](https://github.com/dotnet/runtime/blob/v10.0.1/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Metrics/RuntimeMetrics.cs#L185).
- `runtime.dotnet.gc.memory_load` was a tricky one as the built-in uses a new API, but I think the info we get in `GC.GetGCMemoryInfo()` is broadly good enough
- `runtime.dotnet.gc.count.gen#` uses `GC.CollectionCount()`, same as [built-in approach](https://github.com/dotnet/runtime/blob/v10.0.1/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Metrics/RuntimeMetrics.cs#L159)
- `runtime.dotnet.gc.pause_time` this was also a tricky one, more on it below...

`runtime.dotnet.gc.pause_time` is a runtime metric that's available in .NET 9, so when we're running in .NET 9 we just use that value. There's actually also a public API introduce in .NET 6, `GetTotalPauseDuration()`, but [it's _only_ available from 6.0.21](https://github.com/dotnet/runtime/pull/87143), so we can't directly reference it. Resorted to using a simple `CreateDelegate` call to invoke it in these cases. We could use duck typing, but didn't seem worth it. If we're running < 6.0.21, there's no feasible way to get the value, so we just don't emit it.

ASP.NET Core metrics:
- `runtime.dotnet.aspnetcore.requests.current`
- `runtime.dotnet.aspnetcore.requests.failed`
- `runtime.dotnet.aspnetcore.requests.total`
- `runtime.dotnet.aspnetcore.requests.queue_length`
- `runtime.dotnet.aspnetcore.connections.current`
- `runtime.dotnet.aspnetcore.connections.queue_length`
- `runtime.dotnet.aspnetcore.connections.total`

Note that the `.total` and `.failed` requests are recorded as _gauges_ (which monotonically increase), which doesn't feel right to me (they should be counters, surely), but that's what `RuntimeEventListener` is using, so we have to stick to the same thing (metric types are global by metric, so we can't change it). It means there's a risk of overflow there, but that's already the case for `RuntimeEventListener` so I guess we just ignore it 🤷‍♂️ 

I couldn't find a way to get the following metrics at all without using `EventListener`: 
- `runtime.dotnet.threads.contention_time`

## Test coverage

Added unit and integration tests for the listener behavior.

I also manually ran an aspnetcore app in a loop with both the `RuntimeEventListener` and the new listener producing metrics (hacked in, we wont ever do this in "normal" execution), and did a manual comparison of the metrics. Overall, the values were in broad agreement (slightly off, due to skew in sampling time) and helped identify some cases where I'd made incorrect assumptions (e.g. aspnetcore `.total` metrics are never "reset" to 0.

## Other details

Relates to:
- https://github.com/DataDog/dd-trace-dotnet/issues/5862#issuecomment-3578608955
- https://github.com/dotnet/runtime/issues/111368
- https://github.com/dotnet/runtime/pull/118415
- https://datadoghq.atlassian.net/browse/LANGPLAT-916

### DIFF
--- a/tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
+++ b/tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
@@ -182,7 +182,7 @@ namespace Datadog.Trace.Configuration
             RuntimeMetricsDiagnosticsMetricsApiEnabled = config.WithKeys(ConfigurationKeys.RuntimeMetricsDiagnosticsMetricsApiEnabled).AsBool(false);
 
 #if !NET6_0_OR_GREATER
-            if (RuntimeMetricsDiagnosticsMetricsApiEnabled)
+            if (RuntimeMetricsEnabled && RuntimeMetricsDiagnosticsMetricsApiEnabled)
             {
                 Log.Warning(
                     $"{ConfigurationKeys.RuntimeMetricsDiagnosticsMetricsApiEnabled} was enabled, but System.Diagnostics.Metrics is only available on .NET 6+. Using standard runtime metrics collector.");


tracer/test/benchmarks/Benchmarks.Trace/TelemetryHttpClientBenchmark.cs
```diff
@@ -0,0 +1,307 @@
...
+        var request = _apiWebRequestFactory.Create(_apiEndpointUri);
+        using var response = await request.PostAsJsonAsync(request, compression: MultipartCompression.None);
+        return response.StatusCode;
+    }
+
+    [BenchmarkCategory("ApiWebRequest", "Gzip"), Benchmark]
+    public async Task<int> ApiWebRequest_After_Gzip()
+    {
+        var request = _apiWebRequestFactory.Create(_apiEndpointUri);
+        using var response = await request.PostAsJsonAsync(request, compression: MultipartCompression.None);
```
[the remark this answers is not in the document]
andrewlock: oops, didn't mean to commit that. Removed it now

tracer/src/Datadog.Trace/RuntimeMetrics/DiagnosticsMetricsRuntimeMetricsListener.cs
```diff
@@ -0,0 +1,335 @@
...
+                  "http.server.request.duration")
+          || (string.Equals(meterName, "Microsoft.AspNetCore.Server.Kestrel", StringComparison.Ordinal) && instrumentName is
+                  "kestrel.active_connections" or
+                  "kestrel.queued_connections" or
+                  "kestrel.connection.duration" or
+                  "kestrel.queued_requests"))
+        {
+            if (Log.IsEnabled(LogEventLevel.Debug))
+            {
+                Log.Debug("Enabled measurement events for instrument: {MeterName}/{InstrumentName} ", instrumentName, meterName);
```
bouwkast: just swapped around
```suggestion
                Log.Debug("Enabled measurement events for instrument: {MeterName}/{InstrumentName} ", meterName, instrumentName);
```

tracer/src/Datadog.Trace/RuntimeMetrics/DiagnosticsMetricsRuntimeMetricsListener.cs
```diff
@@ -0,0 +1,335 @@
...
+            }
+        }
+
+        // This isn't strictly true, due to the "previous counts" behavior, but it's good enough, and what we in other listeners
+        Log.Debug($"Sent the following metrics to the DD agent: {MetricsNames.Gen0HeapSize}, {MetricsNames.Gen1HeapSize}, {MetricsNames.Gen2HeapSize}, {MetricsNames.LohSize}, {MetricsNames.ContentionCount}, {MetricsNames.Gen0CollectionsCount}, {MetricsNames.Gen1CollectionsCount}, {MetricsNames.Gen2CollectionsCount}, {MetricsNames.GcPauseTime}, {MetricsNames.GcMemoryLoad}");
+
+        // aspnetcore metrics
+        if (_aspnetcoreMetricsAvailable)
+        {
+            statsd.Gauge(MetricsNames.AspNetCoreCurrentRequests, Interlocked.Read(ref _activeRequests));
```
bouwkast: Wondering if we should do all the `Interlocked.Read`s right after each other to try and ensure ensure that they are collected roughly closer together? Unsure if that is even a thing to care about though

tracer/src/Datadog.Trace/RuntimeMetrics/DiagnosticsMetricsRuntimeMetricsListener.cs
```diff
@@ -0,0 +1,335 @@
...
+            //   - Working Set Size / Limit
+            //   - GlobalMemoryStatusEx -> (ullTotalVirtual - ullAvailVirtual) * 100.0 / (float)ms.ullTotalVirtual
+            //
+            // We try to roughly emulate that using the info in gcInfo:
+            var availableBytes = gcInfo.TotalAvailableMemoryBytes;
+
+            if (availableBytes > 0)
+            {
+                // This can return a value > 1 (for values I don't _entirely_ understand), so clamp it to 1.0
+                statsd.Gauge(MetricsNames.GcMemoryLoad, (double)gcInfo.MemoryLoadBytes * 100.0 / availableBytes);
```
bouwkast: > so clamp it to 1.0

Is this needed? I don't think it is clamped

Should it be ->

```suggestion
                // This can return a value > 1 (for values I don't _entirely_ understand), so clamp it to 1.0
                var ratio = Math.Min(1.0, (double)gcInfo.MemoryLoadBytes / availableBytes);
                statsd.Gauge(MetricsNames.GcMemoryLoad, ratio * 100.0);
```

tracer/src/Datadog.Trace/RuntimeMetrics/DiagnosticsMetricsRuntimeMetricsListener.cs
```diff
@@ -0,0 +1,335 @@
...
+            // Same here, seems risky to have this as a gauge, but I think that ship has sailed
+            // Note also that as _totalClosedConnections doesn't include _current_ connections, we add that in
+            statsd.Gauge(MetricsNames.AspNetCoreTotalConnections, Interlocked.Read(ref _totalClosedConnections) + currentConnections);
+            Log.Debug($"Sent the following metrics to the DD agent: {MetricsNames.AspNetCoreCurrentRequests}, {MetricsNames.AspNetCoreFailedRequests}, {MetricsNames.AspNetCoreTotalRequests}, {MetricsNames.AspNetCoreRequestQueueLength}, {MetricsNames.AspNetCoreCurrentConnections}, {MetricsNames.AspNetCoreConnectionQueueLength}, {MetricsNames.AspNetCoreTotalConnections}");
+        }
+    }
+
+    private static void OnMeasurementRecordedDouble(Instrument instrument, double measurement, ReadOnlySpan<KeyValuePair<string, object?>> tags, object? state)
+    {
+        var handler = (DiagnosticsMetricsRuntimeMetricsListener)state!;
```
bouwkast, line 234: Is the `!` safe here?

tracer/src/Datadog.Trace/RuntimeMetrics/DiagnosticsMetricsRuntimeMetricsListener.cs
```diff
@@ -0,0 +1,335 @@
...
+                    }
+                }
+
+                break;
+        }
+    }
+
+    private static void OnMeasurementRecordedLong(Instrument instrument, long measurement, ReadOnlySpan<KeyValuePair<string, object?>> tags, object? state)
+    {
+        var handler = (DiagnosticsMetricsRuntimeMetricsListener)state!;
```
bouwkast, line 267: Is the `!` safe? I am thinking so since it is the callback and is consistent with the other method as well

tracer/test/Datadog.Trace.Tests/RuntimeMetrics/DiagnosticMetricsRuntimeMetricsListenerTests.cs
```diff
@@ -0,0 +1,229 @@
...
+                { "http.request.method", "GET" },
+                { "network.protocol.version", "1.1" },
+                { "http.response.status_code", (object)200 },
+                { "http.route", "/" }
+            });
+
+        listener.Refresh();
+
+        statsd.Verify(s => s.Gauge(MetricsNames.AspNetCoreFailedRequests, 0.0, 1, null), Times.Once);
+        statsd.Verify(s => s.Gauge(MetricsNames.AspNetCoreTotalRequests, 1.0, 1, null), Times.Once);
```
andrewlock: Oops, forgot to fix these after I changed the behaviour to never reset:


```suggestion
        statsd.Verify(s => s.Gauge(MetricsNames.AspNetCoreFailedRequests, 1.0, 1, null), Times.Once);
        statsd.Verify(s => s.Gauge(MetricsNames.AspNetCoreTotalRequests, 4.0, 1, null), Times.Once);
```

tracer/src/Datadog.Trace/RuntimeMetrics/DiagnosticsMetricsRuntimeMetricsListener.cs
```diff
@@ -0,0 +1,335 @@
...
+    private static void OnMeasurementRecordedDouble(Instrument instrument, double measurement, ReadOnlySpan<KeyValuePair<string, object?>> tags, object? state)
+    {
+        var handler = (DiagnosticsMetricsRuntimeMetricsListener)state!;
+        switch (instrument.Name)
+        {
+            case "dotnet.gc.pause.time":
+                Interlocked.Exchange(ref handler._gcPauseTimeSeconds, measurement);
+                break;
+            case "kestrel.connection.duration":
+                Interlocked.Increment(ref handler._totalClosedConnections);
```
bouwkast, line 241: Is `totalClosedConnections` the thing to increment for this?

tracer/src/Datadog.Trace/RuntimeMetrics/DiagnosticsMetricsRuntimeMetricsListener.cs
```diff
@@ -0,0 +1,335 @@
...
+
+internal sealed class DiagnosticsMetricsRuntimeMetricsListener : IRuntimeMetricsListener
+{
+    private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor<DiagnosticsMetricsRuntimeMetricsListener>();
+
+    // Note that we don't currently record Pinned Object Heap sizes
+    private static readonly string[] GcGenSizeMetricNames = [MetricsNames.Gen0HeapSize, MetricsNames.Gen1HeapSize, MetricsNames.Gen2HeapSize, MetricsNames.LohSize];
+    private static readonly string[] GcGenCountMetricNames = [MetricsNames.Gen0CollectionsCount, MetricsNames.Gen1CollectionsCount, MetricsNames.Gen2CollectionsCount];
+
+    private static readonly int MaxGcSizeGenerations = Math.Min(GC.GetGCMemoryInfo().GenerationInfo.Length, GcGenSizeMetricNames.Length);
```
bouwkast, line 28: Just wondering if it is okay to call `GetGCMemoryInfo()` here  based on this comment later down

```
// GC Heap Size based on "dotnet.gc.last_collection.heap.size" metric
// from https://github.com/dotnet/runtime/blob/v10.0.1/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Metrics/RuntimeMetrics.cs#L185
// If we call this API before _any_ GCs happen, we'll get 0s for the heap size, so check for that and bail skip emitting if so
```

tracer/src/Datadog.Trace/RuntimeMetrics/DiagnosticsMetricsRuntimeMetricsListener.cs
bouwkast: Ah nevermind looking it up this is recorded when the connection is closed

tracer/src/Datadog.Trace/RuntimeMetrics/DiagnosticsMetricsRuntimeMetricsListener.cs
andrewlock: Yeah, I considered it, I'm not sure if it's worth it or not tbh, but happy to make the change 🙂

tracer/src/Datadog.Trace/RuntimeMetrics/DiagnosticsMetricsRuntimeMetricsListener.cs
andrewlock: Oops, no, that's a left over comment from a previous implementation
```suggestion
                statsd.Gauge(MetricsNames.GcMemoryLoad, (double)gcInfo.MemoryLoadBytes * 100.0 / availableBytes);
```

tracer/src/Datadog.Trace/RuntimeMetrics/DiagnosticsMetricsRuntimeMetricsListener.cs
andrewlock: Yes, the `state` variable is whatever you pass in when configuring the `OnMeasurementRecordedDouble` handlers, and as that's always non-null, we know this will be non-null here (though the compiler can't tell that)

tracer/src/Datadog.Trace/RuntimeMetrics/DiagnosticsMetricsRuntimeMetricsListener.cs
andrewlock: Yes, the `state` variable is whatever you pass in when configuring the `OnMeasurementRecordedDouble` handlers, and as that's always non-null, we know this will be non-null here (though the compiler can't tell that)

tracer/src/Datadog.Trace/RuntimeMetrics/DiagnosticsMetricsRuntimeMetricsListener.cs
andrewlock: I _believe_ it's safe, because even though there may not have been any collections, the _number_ of generations is fixed and known, so that info should be correct.

Also, it's what the buit-in metrics do, so I have an even higher confidence 😄  (They actually don't handle the `Index=0` case at all, so will give "incorrect" heap sizes of 0, which we _do_ handle gracefully)

tracer/src/Datadog.Trace/RuntimeMetrics/DiagnosticsMetricsRuntimeMetricsListener.cs
```diff
@@ -0,0 +1,335 @@
...
+            // - On Windows, memory_load is calculated differently depending if there's a restriction (src/coreclr/gc/unix/gcenv.windows.cpp#L1000)
+            //   - Working Set Size / Limit
+            //   - GlobalMemoryStatusEx -> (ullTotalVirtual - ullAvailVirtual) * 100.0 / (float)ms.ullTotalVirtual
+            //
+            // We try to roughly emulate that using the info in gcInfo:
+            var availableBytes = gcInfo.TotalAvailableMemoryBytes;
+
+            if (availableBytes > 0)
+            {
+                // This can return a value > 1 (for values I don't _entirely_ understand), so clamp it to 1.0
```
andrewlock:
```suggestion
```

tracer/src/Datadog.Trace/RuntimeMetrics/DiagnosticsMetricsRuntimeMetricsListener.cs
andrewlock:
```suggestion
            var activeRequests = Interlocked.Read(ref _activeRequests);
            var failedRequests = Interlocked.Read(ref _failedRequests);
            var successRequests = Interlocked.Read(ref _successRequests);
            var queuedRequests = Interlocked.Read(ref _queuedRequests);
            var currentConnections = Interlocked.Read(ref _activeConnections);
            var queuedConnections = Interlocked.Read(ref _queuedConnections);
            var totalClosedConnections = Interlocked.Read(ref _totalClosedConnections);

            statsd.Gauge(MetricsNames.AspNetCoreCurrentRequests, activeRequests);
```

tracer/src/Datadog.Trace/RuntimeMetrics/DiagnosticsMetricsRuntimeMetricsListener.cs
```diff
@@ -0,0 +1,335 @@
...
+            statsd.Gauge(MetricsNames.AspNetCoreTotalRequests, failed + Interlocked.Read(ref _successRequests));
+            statsd.Gauge(MetricsNames.AspNetCoreRequestQueueLength, Interlocked.Read(ref _queuedRequests));
+
+            var currentConnections = Interlocked.Read(ref _activeConnections);
+            statsd.Gauge(MetricsNames.AspNetCoreCurrentConnections, currentConnections);
+            statsd.Gauge(MetricsNames.AspNetCoreConnectionQueueLength, Interlocked.Read(ref _queuedConnections));
+
+            // Same here, seems risky to have this as a gauge, but I think that ship has sailed
+            // Note also that as _totalClosedConnections doesn't include _current_ connections, we add that in
+            statsd.Gauge(MetricsNames.AspNetCoreTotalConnections, Interlocked.Read(ref _totalClosedConnections) + currentConnections);
```
andrewlock:
```suggestion
            statsd.Gauge(MetricsNames.AspNetCoreFailedRequests, failedRequests);
            statsd.Gauge(MetricsNames.AspNetCoreTotalRequests, failedRequests + successRequests);
            statsd.Gauge(MetricsNames.AspNetCoreRequestQueueLength, queuedRequests);

            statsd.Gauge(MetricsNames.AspNetCoreCurrentConnections, currentConnections);
            statsd.Gauge(MetricsNames.AspNetCoreConnectionQueueLength, queuedConnections);

            // Same here, seems risky to have this as a gauge, but I think that ship has sailed
            // Note also that as _totalClosedConnections doesn't include _current_ connections, we add that in
            statsd.Gauge(MetricsNames.AspNetCoreTotalConnections, totalClosedConnections + currentConnections);
```

tracer/src/Datadog.Trace/RuntimeMetrics/DiagnosticsMetricsRuntimeMetricsListener.cs
```diff
@@ -0,0 +1,341 @@
...
+            // - On Windows, memory_load is calculated differently depending if there's a restriction (src/coreclr/gc/unix/gcenv.windows.cpp#L1000)
+            //   - Working Set Size / Limit
+            //   - GlobalMemoryStatusEx -> (ullTotalVirtual - ullAvailVirtual) * 100.0 / (float)ms.ullTotalVirtual
+            //
+            // We try to roughly emulate that using the info in gcInfo:
+            var availableBytes = gcInfo.TotalAvailableMemoryBytes;
+
+            if (availableBytes > 0)
+            {
+                // This can return a value > 1 (for values I don't _entirely_ understand), so clamp it to 1.0
```
andrewlock:
```suggestion
```

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
```diff
@@ -179,6 +179,16 @@ not null when string.Equals(value, "otlp", StringComparison.OrdinalIgnoreCase) =
 
             RuntimeMetricsEnabled = runtimeMetricsEnabledResult.WithDefault(false);
 
+            RuntimeMetricsDiagnosticsMetricsApiEnabled = config.WithKeys(ConfigurationKeys.RuntimeMetricsDiagnosticsMetricsApiEnabled).AsBool(false);
+
+#if !NET6_0_OR_GREATER
+            if (RuntimeMetricsDiagnosticsMetricsApiEnabled)
```
lucaspimentel: Should we only log this if `RuntimeMetricsEnabled` is also enabled? Since the log message says "Using standard runtime metrics collector."

lucaspimentel (approved):

tracer/src/Datadog.Trace/Configuration/TracerSettings.cs
andrewlock: Makes sense 👍
```suggestion
            if (RuntimeMetricsEnabled && RuntimeMetricsDiagnosticsMetricsApiEnabled)
```

PR #8033, opened 2026-01-06
[SINT-4550] Use CI Identities in Windows CI Jobs

cedricvanrompay-datadog:
## Summary of changes

We use the CI Identities system for authentication in Windows CI jobs.

We also:
* update the Windows code signing tool
* anticipate the renewal of the Windows code signing certificate

## Reason for change

The CI Identities allows finer-grained authentication than the existing systems, and give Windows CI Jobs access to Vault (for retrieving secrets for instance). 

## Implementation details

- Rebuild the GitLab images to grab the signing and identity tools as part of the container build
  - These tools are available as docker images on the internal registry (so requires auth, but otherwise a simple rebuild)
- Update the .gitlab.yml to set the required env vars for auth
- Call `assume-role` as part of the entrypoint execution

## Test coverage

This is the test, as long as the build works, and the artifacts are signed, we're good 👍

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -40,7 +40,7 @@ build:
         -e NUGET_CERT_REVOCATION_MODE=offline `
         -e CI_IDENTITIES_GITLAB_ID_TOKEN `
         -e CI_PROJECT_NAME `
-        datadog/dd-trace-dotnet-docker-build:ci-identities `
+        registry.ddbuild.io/images/mirror/datadog/dd-trace-dotnet-docker-build:ci-identities `
         Info Clean BuildTracerHome BuildProfilerHome BuildNativeLoader BuildDdDotnet PublishFleetInstaller PackageTracerHome ZipSymbols SignDlls SignMsi DownloadWinSsiTelemetryForwarder
     - mkdir artifacts-out
     - xcopy /e/s build-out\${CI_JOB_ID}\*.* artifacts-out


tracer/build/_build/Build.Gitlab.cs
```diff
@@ -88,7 +88,15 @@ partial class Build
...
-        const string validSignature = "59063C826DAA5B628B5CE8A2B32015019F164BF0";
+        // See list of certificates
+        // in https://datadoghq.atlassian.net/wiki/spaces/SECENG/pages/3217261499/Certificates+for+Windows+Code+Signing
+        List<string> expectedCertificateThumbprints =
+        [
+            "A0FB7BEE153FE31431062731306903B3A5CB1824",
+            // TODO remove this one when the new certificate is deployed;
+            // see https://github.com/DataDog/windows-code-signing-cert/blob/main/current-certs.toml
+            "59063C826DAA5B628B5CE8A2B32015019F164BF0",
+        ];
```
andrewlock: We can't use collection expressions in the `Build` jobs unfortunately:

```suggestion
        var expectedCertificateThumbprints = new []
        {
            "A0FB7BEE153FE31431062731306903B3A5CB1824",
            // TODO remove this one when the new certificate is deployed;
            // see https://github.com/DataDog/windows-code-signing-cert/blob/main/current-certs.toml
            "59063C826DAA5B628B5CE8A2B32015019F164BF0",
        };
```

tracer/build/_build/Build.Gitlab.cs
```diff
@@ -108,13 +116,7 @@ void SignBinary(AbsolutePath binaryPath)
             var output = signProcess.Output.Select(o => o.Text);
             foreach (var line in output)
             {
-                Logger.Information("[dd-wcs] {Line}", line);
-
-                // dd-wcs will return 0 even if there are errors
-                if (line.StartsWith("ERROR:", StringComparison.OrdinalIgnoreCase))
```
andrewlock, line 114: I take it this is no longer the case, and the sign tool returns an error now if it fails for any reason? Just checking because this was a cause of releasing unsigned artifacts and causing an incident previously 😅

tracer/build/_build/docker/gitlab/gitlab.windows.dockerfile
```diff
@@ -40,27 +40,20 @@ ENV DOTNET_VERSION="10.0.100" \
...
-ENV WINSIGN_SHA256 "b2ba5127a5c5141e04d42444ca115af4c95cc053a743caaa9b33c68dd6b13f68"
-ENV PYTHON_VERSION "3.8.2"
-
-# Install Python
-COPY install_python3.ps1 .
-RUN powershell -Command .\install_python3.ps1 -Version $ENV:PYTHON_VERSION
-
-COPY requirements.txt constraints.txt install_python_packages.ps1 ./
-RUN powershell -Command .\install_python_packages.ps1
+ENV WINSIGN_VERSION "v0.5.0"
```
andrewlock: Looks like there's a limitation in Docker that you can't expand this variable in the `COPY --from` command. All my attempts end up with `invalid from flag value registry.ddbuild.io/windows-code-signer/go:${WINSIGN_VERSION}: invalid reference format`, so just bake-in the version in the command directly (like you did above)
```suggestion
```

tracer/build/_build/docker/gitlab/gitlab.windows.dockerfile
```diff
@@ -40,27 +40,20 @@ ENV DOTNET_VERSION="10.0.100" \
...
 
 # Install JAVA
 COPY helpers.ps1 install_java.ps1 ./
 RUN powershell -Command .\install_java.ps1
 
-# Install 
-COPY install_winsign.ps1 .
-RUN powershell -Command .\install_winsign.ps1
+# Install Windows Code Signer
+COPY --from=registry.ddbuild.io/windows-code-signer/go:$WINSIGN_VERSION c:/windows-code-signer/windows-code-signer.exe c:/devtools/windows-code-signer.exe
```
andrewlock:
```suggestion
COPY --from=registry.ddbuild.io/windows-code-signer/go:v0.5.0 c:/windows-code-signer/windows-code-signer.exe c:/devtools/windows-code-signer.exe
```

.gitlab-ci.yml
```diff
@@ -37,6 +38,8 @@ build:
         -e AWS_NETWORKING=true `
         -e SIGN_WINDOWS=true `
         -e NUGET_CERT_REVOCATION_MODE=offline `
+        -e CI_IDENTITY_GITLAB_JWT `
+        -e CI_PROJECT_NAME `
         registry.ddbuild.io/images/mirror/datadog/dd-trace-dotnet-docker-build:dotnet10 `
```
andrewlock: I will push the image to docker hub for initial testing. 🙂 once we've confirmed this works, we can then mirror it into ddbuild.io

```suggestion
        datadog/dd-trace-dotnet-docker-build:ci-identities `
```

tracer/build/_build/Build.Gitlab.cs
cedricvanrompay-datadog: The previous signing tool had many issues so we re-wrote it entirely, and I am pretty confident it should exit with a non-zero status if there is any problem: https://github.com/DataDog/windows-code-signer/blob/ee081a1e9d092654c24a037a3e5c524b6f5cc52e/cmd/windows-code-signer/windows-code-signer.go#L21

Was the signature verification step added _after_ the incident? I would expect it to catch any missing signature.

andrewlock: > Was the signature verification step added after the incident? I would expect it to catch any missing signature.

Yes, it was 🙂

.gitlab-ci.yml
```diff
@@ -37,7 +38,9 @@ build:
         -e AWS_NETWORKING=true `
         -e SIGN_WINDOWS=true `
         -e NUGET_CERT_REVOCATION_MODE=offline `
-        registry.ddbuild.io/images/mirror/datadog/dd-trace-dotnet-docker-build:dotnet10 `
+        -e CI_IDENTITIES_GITLAB_ID_TOKEN `
+        -e CI_PROJECT_NAME `
+        datadog/dd-trace-dotnet-docker-build:ci-identities `
```
andrewlock: We added the image in https://github.com/DataDog/images/pull/8437, so you should now be able to use it:

```suggestion
        registry.ddbuild.io/images/mirror/datadog/dd-trace-dotnet-docker-build:ci-identities `
```

tracer/build/_build/docker/gitlab/entrypoint.bat
```diff
@@ -36,6 +36,10 @@ if "%nuke_args%"=="" (
     exit /b 1
 )
 
+:: the CI Identities client will write the credentials to the path in the environment variable AWS_SHARED_CREDENTIALS_FILE,
+:: and if the variable is not set, it will write to %USERPROFILE%\.aws\credentials
+c:\devtools\ci-identities-gitlab-job-client.exe assume-role
```
andrewlock, line 41: Thinking about this more, this _could_ fail, right? What happens if we can't assume the role? Does signing fail? Are we _sure_ that would result in a failure of the build pipeline? Or should we be checking for a non-zero exit code here?

```suggestion
c:\devtools\ci-identities-gitlab-job-client.exe assume-role

if %ERRORLEVEL% NEQ 0 exit /B %ERRORLEVEL%
```

andrewlock (approved):
LGTM, my only concern is whether we should be checking for failure of `assume-role`?

tracer/build/_build/docker/gitlab/entrypoint.bat
cedricvanrompay-datadog: If `assume-role` fails then no credentials are written in the "AWS shared credentials" file, meaning that the AWS SDK will instead use the instance profile credentials.

Right now, the CI job would still be able to do its job using these credentials, this is actually what made me lose some time during this PR because `assume-role` was failing but I did not notice it at first.

Now the reason why we are doing all of this is to _remove_ permissions currently given to the instance profile, so one day the CI job will _not_ be able to do its job if `assume-role` fail.

I would suggest that for now we _don't_ fail the job if `assume-role` fails, protecting you from a failure of the CI Identities system that's still quite young, and later we add this "if error, exit" logic.

PR #8034, opened 2026-01-06
Support custom `LogEventPropertyValue` values in direct log submission

andrewlock:
## Summary of changes

Handle the case where a customer is using a custom `LogEventPropertyValue` implementation and direct log submission

## Reason for change

#8010 handled a case where we were generating invalid JSON by dropping the offending value. This PR adds support for serializing the value as a string instead.

## Implementation details

Ducktyped `LogEventPropertyValue` and called its `Render` method with a cached `StringBuilder` to get the value, then use the same mechanisms

## Test coverage

Should be covered by the tests added in #8010

## Other details

### DIFF
--- a/tracer/src/Datadog.Trace/Logging/DirectSubmission/Sink/DirectSubmissionLogSink.cs
+++ b/tracer/src/Datadog.Trace/Logging/DirectSubmission/Sink/DirectSubmissionLogSink.cs
@@ -100,7 +100,15 @@ namespace Datadog.Trace.Logging.DirectSubmission.Sink
                         _logStringBuilder.Capacity = InitialBuilderSizeBytes;
                     }
 
-                    log.Format(_logStringBuilder, _formatter);
+                    try
+                    {
+                        log.Format(_logStringBuilder, _formatter);
+                    }
+                    catch (Exception ex)
+                    {
+                        _logger.Error(ex, "Log dropped due to an error during serialization");
+                        continue;
+                    }
 
                     // would be nice to avoid this, but can't see a way without direct UTF8 serialization (System.Text.Json)
                     // Currently a rogue giant message still pays the serialization cost but is then thrown away


NachoEchevarria (approved):

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Logging/Serilog/DirectSubmission/Formatting/SerilogLogFormatter.cs
```diff
@@ -107,6 +109,18 @@ private static void FormatLogEventPropertyValue(JsonTextWriter writer, object va
                 return;
             }
 
+            if (value.TryDuckCast<ILogEventPropertyValue>(out var propertyValue))
+            {
+                // This is likely an edge case, so we just write the value as a string and then embed that in the JSON
+                var sb = StringBuilderCache.Acquire();
+                var stringWriter = new StringWriter(sb);
+                propertyValue.Render(stringWriter, null, null);
```
NachoEchevarria, line 117: If Render() launches an exception, we should probably catch it, right?

andrewlock: That's actually an interesting one - currently we don't catch exceptions anywhere in here they bubble up and will cause a whole batch to be dropped. It seems like it would make sense to catch exceptions at the _log_ level? 🤔 i.e. if we get an exception anywhere in the serialization, we should drop that log, but not the whole batch. Does that make sense to you?

andrewlock: I added a try-catch at a higher level (where we call the top-level `Format` method, inside the direct log submission sink. That way we will catch _any_ errors coming from formatter methods (there are multiple implementations, so I think it's better to be high-level here). That single log will then be dropped, but the overall batch isn't which seems preferable

PR #8037, opened 2026-01-08
Introduce `ActivityKey` to remove string concatenation

andrewlock:
## Summary of changes

Reduce allocation of `ActivityHandlerCommon` by removing `string` concatenation

## Reason for change

The `ActivityHandlerCommon.ActivityStarted` and `ActivityStopped` methods need to store and retrieve `Activity` instances from a `ConcurrentDictionary<>`. Today they're doing that be concatenating the `Activity`'s `TraceId` and `SpanID`, or by using it's `ID`. All that concatenation causes a bunch of allocation, so instead introduce a simple `struct` to use as the key instead

## Implementation details

Introduce `ActivityKey`, which is essentially `internal readonly record struct ActivityKey(string TraceId, string SpanId`, and use that for all of the dictionary lookups. Which avoids all the string concatenation allocations.

## Test coverage

Added some unit tests for `ActivityKey`, by mostly covered by existing integration tests for correctness. Benchmarks show a significant improvement over [the previous results](https://github.com/DataDog/dd-trace-dotnet/pull/8036), particularly for Hierachical IDs which clearly were buggy

| Method                          | Runtime              |     Mean |    StdDev |   Gen0 | Allocated | Compared to #8036 |
| ------------------------------- | -------------------- | -------: | --------: | -----: | --------: | ----------------: |
| StartStopWithChild_Hierarchical | .NET 6.0             | 4.217 us | 0.3227 us | 0.0153 |   4.09 KB |          -6.38 KB |
| StartStopWithChild_Hierarchical | .NET 8.0             | 3.413 us | 0.2505 us | 0.0076 |   4.09 KB |          -6.39 KB |
| StartStopWithChild_Hierarchical | .NET Core 3.1        | 5.676 us | 0.4636 us | 0.0153 |   4.32 KB |          -6.43 KB |
| StartStopWithChild_Hierarchical | .NET Framework 4.7.2 | 6.813 us | 0.4969 us | 0.7324 |   4.53 KB |          -7.24 KB |
|                                 |                      |          |           |        |           |                   |
| StartStopWithChild              | .NET 6.0             | 4.105 us | 0.2677 us | 0.0153 |    4.3 KB |          -0.57 KB |
| StartStopWithChild              | .NET 8.0             | 3.475 us | 0.1570 us | 0.0114 |    4.2 KB |          -0.57 KB |
| StartStopWithChild              | .NET Core 3.1        | 5.647 us | 0.3129 us | 0.0153 |   4.48 KB |          -0.57 KB |
[...]

PR #8036, opened 2026-01-08
Update `Activity` benchmarks for stability

andrewlock:
## Summary of changes

Update `Activity`-based benchmarks to reduce variability and make comparisons easier

## Reason for change

The various `Activity`-related benchmarks call the global `Tracer` instance, so we should make sure to configure it with our default benchmarking settings (basically disabling background jobs like telemetry/discovery/remote config) to reduce variation. Also added a "baseline" job for comparison and a version that uses hiearachical IDs instead of W3C IDs. Was a prerequisite for a bunch of other work.

## Implementation details

- Configure `Benchmarks.OpenTelemetry.InstrumentedApi` project and `Benchmarks.Trace/ActivityBenchmark` to setup the global tracer with Telemetry etc disabled
- Add two extra benchmarks to `Benchmarks.Trace/ActivityBenchmark`
  - `StartStopWithChild_Baseline`, which is the same as `StartStopWithChild` but without the DD integration
  - `StartStopWithChild_Hierarchical`, which is the same as `StartStopWithChild` but uses hierarchical ID format
  - Don't run either of these in CI for now (to avoid extra load), just for local comparisons
- Simplify the benchmark to just do explicit duck typing (which is closer to what we do normally anyway, and removes a bunch of code)

## Test coverage

Running these benchmarks locally shows the improvements we need to make, and highlights that we clearly have a bug with hierarchical IDs 😬 

| Method                          | Runtime              |        Mean |       Error |      StdDev |   Gen0 |   Gen1 | Allocated |
| ------------------------------- | -------------------- | ----------: | ----------: | ----------: | -----: | -----: | --------: |
| StartStopWithChild_Baseline     | .NET 6.0             |    672.9 ns |    23.23 ns |    66.66 ns | 0.0038 |      - |   1.09 KB |
| StartStopWithChild_Hierarchical | .NET 6.0             | 35,478.1 ns | 1,035.96 ns | 3,021.94 ns |      - |      - |  10.47 KB |
| StartStopWithChild              | .NET 6.0             |  3,681.7 ns |    50.23 ns |    44.53 ns | 0.0153 |      - |   4.87 KB |
| StartStopWithChild_Baseline     | .NET 8.0             |    570.8 ns |    16.96 ns |    49.22 ns | 0.0029 |      - |   1.09 KB |
| StartStopWithChild_Hierarchical | .NET 8.0             | 32,478.5 ns | 1,215.04 ns | 3,466.57 ns |      - |      - |  10.48 KB |
[...]

PR #8038, opened 2026-01-08
Fix nullable annotations on `Activity` duck types

andrewlock:
## Summary of changes

Fix incorrect nullable annotations on `Activity` duck types

## Reason for change

While working on other performance things, noticed that the nullable annotations often declared non-nullability when they actually could be null.

A particularly confusing part are the `TraceId` and `SpanId` values in `IW3CActivity`. These were marked non-nullable because when you call `Activity.Start()` then these will always be non-null, but _only_ if you're using W3C IDs. If you're using hierarchical IDs ([the default in <.NET 5](https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/5.0/default-activityidformat-changed)) then these values _will_ be null.

As a side note, I suspect this explains the "we saw errors about these being null in error tracking but don't understand why" scenarios 😄 

Also, I think we should rename `IW3CActivity` to `IActivity3` instead. W3C _implies_ that it's a W3C activity, but that's not necessarily the case, and is essentially the source of the above confusion I think. `IActivity3` would then be consistently named with `IActivity5` and `IActivity6` we also currently have.

## Implementation details

Add nullable annotations to values that _can_ be null, and fix the fallout (mostly in `ActivityKey`)

## Test coverage

Covered by existing tests sufficiently I think

## Other details

- https://github.com/DataDog/dd-trace-dotnet/pull/8037 👈

### DIFF
--- a/tracer/src/Datadog.Trace/Activity/DuckTypes/IActivity.cs
+++ b/tracer/src/Datadog.Trace/Activity/DuckTypes/IActivity.cs
@@ -13,10 +13,13 @@ namespace Datadog.Trace.Activity.DuckTypes
 {
     internal interface IActivity : IDuckType
     {
-        // The corresponding property on the Activity object is nullable,
-        // but we are guaranteed that the value will not be null once the Activity is started.
-        // Since we are only accessing the Activity after it has started, we can treat the value
-        // as never having a null value.
+        /// <summary>
+        /// Gets an ID for the activity. If the Activity is using hierarchical IDs, the Id has a hierarchical structure:
+        /// '|root-id.id1_id2.id3_' and is generated when Start() is called by appending suffix to Parent.Id
+        /// or ParentId; Activity has no Id until it started, but as we only see started IDs, this will always be non null.
+        /// </summary>
+        /// <remarks>NOTE: this property allocates when using W3C activities, and so should generally not be called
+        /// unless you know that you're not using W3C activities (because <see cref="IW3CActivity.SpanId"/> is null</remarks>
         string Id { get; }
 
         string? ParentId { get; }


tracer/src/Datadog.Trace/Activity/DuckTypes/IW3CActivity.cs
```diff
@@ -11,11 +11,14 @@ namespace Datadog.Trace.Activity.DuckTypes
...
         [DuckField(Name = "_traceId")]
-        string TraceId { get; set; }
+        string? TraceId { get; set; }
 
         [DuckField(Name = "_spanId")]
-        string SpanId { get; set; }
+        string? SpanId { get; set; }
 
         [DuckField(Name = "_parentSpanId")]
         string ParentSpanId { get; set; }
```
bouwkast, line 24: Should we do this one as well? Unsure if there will be other updates necessary.
https://github.com/dotnet/dotnet/blob/b0f34d51fccc69fd334253924abd8d6853fad7aa/src/runtime/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.cs#L102

```suggestion
        string? ParentSpanId { get; set; }
```

bouwkast (approved):
Thanks!

> As a side note, I suspect this explains the "we saw errors about these being null in error tracking but don't understand why" scenarios 😄

I think if I recall correctly this was actually the `Activity.Id` being `null` and not the `SpanId`/`TraceId`

But there were a lot of weirdness with handling the different versions and formats. I think that if `SpanId`/`TraceId` are `null` for Hierarchical the `Id` wouldn't, but looking at the current code that is marked nullable too so 🤷 

https://github.com/dotnet/dotnet/blob/b0f34d51fccc69fd334253924abd8d6853fad7aa/src/runtime/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.cs#L96C9-L96C29

tracer/src/Datadog.Trace/Activity/DuckTypes/IW3CActivity.cs
bouwkast: I see this is done in https://github.com/DataDog/dd-trace-dotnet/pull/8040 👍

andrewlock:
> I think if I recall correctly this was actually the Activity.Id being null and not the SpanId/TraceId

Ah, ok, I actually assumed it was related to the fix in [the previous PR](https://github.com/DataDog/dd-trace-dotnet/pull/8037) where we were assuming that `IW3CActivity` meant it had a `SpanId`/`TraceId`, which isn't the case.

> I think that if SpanId/TraceId are null for Hierarchical the Id wouldn't, but looking at the current code that is marked nullable too so 🤷

Yeah, I think as long as you call `Start()` on the activity, then whichever ID type it is `ID` _shouldn't_ be null. And given in practice we only see "started" activities, I think the existing nullable annotation should be ok.

That said, we should avoid calling it if possible when using `W3C`, because it allocates 🤔 I think we currently only use it as a fallback, but will add that as an extra note, just to flag it

PR #8039, opened 2026-01-08
Avoid allocating a closure for `ActivityMappingById.GetOrAdd`

andrewlock:
## Summary of changes

Avoids allocating a closure in .NET Core if we can avoid it

## Reason for change

.NET Core's `ConcurrentDictionary.GetOrAdd()` method allows providing a "state" object which we can pass to the `GetOrAdd` method. Using this method avoids allocating a closure every time the method is hit, and we can pass the state using a value tuple to avoid additional allocation there

## Implementation details

`#if`/`#else` to glory 

## Test coverage

Functionality is covered by existing tests, benchmarks show an incremental improvement over #8037 for .NET Core, as expected:

| Method                          | Runtime              |     Mean |    StdDev |   Gen0 | Allocated | Compared to #8037 |
| ------------------------------- | -------------------- | -------: | --------: | -----: | --------: | ----------------: |
| StartStopWithChild_Hierarchical | .NET 6.0             | 3.746 us | 0.1128 us | 0.0114 |   3.81 KB |          -0.28 KB |
| StartStopWithChild_Hierarchical | .NET 8.0             | 2.759 us | 0.0374 us | 0.0114 |   3.81 KB |          -0.28 KB |
| StartStopWithChild_Hierarchical | .NET Core 3.1        | 4.762 us | 0.0584 us | 0.0153 |   4.04 KB |          -0.28 KB |
| StartStopWithChild_Hierarchical | .NET Framework 4.7.2 | 5.651 us | 0.0717 us | 0.7324 |   4.54 KB |   0.01 KB (noise) |
|                                 |                      |          |           |        |           |                   |
| StartStopWithChild              | .NET 6.0             | 3.607 us | 0.0508 us | 0.0153 |   4.02 KB |          -0.28 KB |
| StartStopWithChild              | .NET 8.0             | 2.921 us | 0.0617 us | 0.0114 |   3.91 KB |          -0.29 KB |
| StartStopWithChild              | .NET Core 3.1        | 4.922 us | 0.0407 us | 0.0153 |    4.2 KB |          -0.28 KB |
| StartStopWithChild              | .NET Framework 4.7.2 | 6.008 us | 0.0979 us | 0.7629 |   4.69 KB |                 0 |

## Other details

https://datadoghq.atlassian.net/browse/LANGPLAT-915

Part of a stack working to improve OTel performance

- https://github.com/DataDog/dd-trace-dotnet/pull/8036
- https://github.com/DataDog/dd-trace-dotnet/pull/8037
[...]

PR #8040, opened 2026-01-08
Various optimizations for Activity handling

andrewlock:
## Summary of changes

Various optimizations in the `Activity` handling code to reduce allocations and execution time

## Reason for change

Some properties are triksy, as they do a bunch of allocation, so we should avoid them if we can. The changes in here _look_ bigger than they are diff-wise, I've added comments to aid review.

## Implementation details

- Don't call `ParentId` until we definitely need it.
  - This property does a bunch of allocation to generate a "valid" value, so we should avoid it if we can. This makes the conditions a bit harder to read, but delays calling `ParentId` until we're sure we don't have something better already
- Avoid calling `Tracer.Instance.ActiveScope?.Span` until we know we need it (very minor optimisations but why not 🤷‍♂️) 
- Extract `StopActivitySlow` to a separate method, as we _shouldn't_ hit this now, so should help the JIT out with things like code size etc of the calling method (conjecture, not tested, but I think it's better from code understanding PoV too)
- Simplify `ShouldIgnoreByOperationName` which also improves execution time 10x from ~30us to ~3us.

## Test coverage

Functionally covered by existing tests. Benchmarks compared to #8039 show improvments. Allocations listed below, but execution time is also improved: 

| Method                          | Runtime              |     Mean |    StdDev |   Gen0 | Allocated | Compared to #8039 |
| ------------------------------- | -------------------- | -------: | --------: | -----: | --------: | ----------------: |
| StartStopWithChild_Hierarchical | .NET 6.0             | 3.410 us | 0.0658 us | 0.0153 |   3.79 KB |          -0.02 KB |
| StartStopWithChild_Hierarchical | .NET 8.0             | 2.582 us | 0.0443 us | 0.0114 |   3.79 KB |          -0.02 KB |
| StartStopWithChild_Hierarchical | .NET Core 3.1        | 4.272 us | 0.0731 us | 0.0153 |   4.02 KB |          -0.02 KB |
| StartStopWithChild_Hierarchical | .NET Framework 4.7.2 | 5.165 us | 0.1245 us | 0.7324 |   4.51 KB |          -0.03 KB |
|                                 |                      |          |           |        |           |                   |
| StartStopWithChild              | .NET 6.0             | 3.312 us | 0.0266 us | 0.0153 |   3.78 KB |          -0.04 KB |
| StartStopWithChild              | .NET 8.0             | 2.648 us | 0.0306 us | 0.0114 |   3.78 KB |          -0.13 KB |
| StartStopWithChild              | .NET Core 3.1        | 4.344 us | 0.0555 us | 0.0076 |   3.97 KB |          -0.23 KB |
| StartStopWithChild              | .NET Framework 4.7.2 | 5.234 us | 0.1568 us | 0.7095 |   4.39 KB |          -0.30 KB |

## Other details

- https://github.com/DataDog/dd-trace-dotnet/pull/8040 👈


tracer/src/Datadog.Trace/Activity/Handlers/ActivityHandlerCommon.cs
```diff
@@ -59,69 +58,73 @@ public static void ActivityStarted<T>(string sourceName, T activity, OpenTelemet
...
+                            traceId: newActivityTraceId,
+                            spanId: newActivitySpanId,
+                            rawTraceId: activityTraceId,
+                            rawSpanId: parentSpanId);
+                    }
+                }
+                else if ((string.IsNullOrEmpty(activityTraceId) && w3cActivity is { ParentSpanId: not null })
+                      || w3cActivity is { Parent: not null })
+                {
+                    // We know we have a parent context, but also that the traceID is (weirdly) null, so we fallback
```
bouwkast: I don't really see how this is possible really looking at the `Activity` code
Pretty sure I wrote it all though 🙈

bouwkast (approved):

tracer/src/Datadog.Trace/Activity/Handlers/ActivityHandlerCommon.cs
zacharycmontoya: Agreed, I'm not sure how this can happen...can we log an error for this? It might seem weird to the end-user, but I would love for us to more proactively get a better sense of when this is happening

zacharycmontoya: > `string.IsNullOrEmpty(activityTraceId) && w3cActivity is { ParentSpanId: not null }`

This condition makes sense in that it's the testing the case where the first `if (!StringUtil.IsNullOrEmpty(activityTraceId) && w3cActivity is { ParentSpanId: { } parentSpanId })` fails the activityTraceId portion of the conditional but passes the parentSpanId part of the conditional. However, I don't see how this is possible. This is the part where I think error logging could identify if we ever hit this.

> `w3cActivity is { Parent: not null }`

It seems like this would happen for .NET 5 but the Activity was set to a Hierarchical format. That one I can easily see happening and shouldn't error log.

What do you think of separating out the two conditionals and for the `ParentSpanId` condition we _only_ do error logging? In fact the current `ParentId` lookup you proposed doesn't look right for that branch

andrewlock: > I don't really see how this is possible really looking at the Activity code
> Pretty sure I wrote it all though 🙈

Yeah, I purposefully avoided changing the checks here because of the complexity of hierarchical IDs vs W3C and how `Activity` has evolved over the years 😅 Definitely opted for an "I trust you" possibility 😂 

From my own reading of Activity, `activityTraceId` will only ever be `null` if we are using hierarchical IDs, so I think your assertion makes sense - we can update this to log a warning when `string.IsNullOrEmpty(activityTraceId) && w3cActivity is { ParentSpanId: not null }`, though what should we _do_ in that case, in terms of finding a parent? 🤔 Just, don't have a parent? If we do that, then we'll end up falling into the _subsequent_ block, where we try to modify the activity to match any ambient span, which also feels a bit... off... 

> In fact the current `ParentId` lookup you proposed doesn't look right for that branch

Hmmm, you're right... the previous condition:

```csharp
if (!StringUtil.IsNullOrEmpty(activityTraceId) && w3cActivity is { ParentSpanId: { } parentSpanId })
```

could _only_ ever be hit for w3c IDs, the hierarchical ones would never be hit 🤔 With my change, we now _do_ hit this for hierarchical IDs... which means that this is a behaviour change, but then again, it feels like the previous logic was off - I don't think we would ever have correctly grabbed the parent for hierarchical IDs? 😕 I'll debug it and see what was happening in those cases, because I'm super confused 😅

andrewlock: Yeah, I just stepped through this, for hierarchical IDs, we would previously _never_ hit _any_ of the branches for setting the parent 🤦‍♂️ 

My fix now _does_ (correctly I think) hit the second branch when you're using hierarchical IDs, and looks up the parent correctly in `ActivityMappingById`, so I believe we need to keep that branch in some form, but it's all a bit confusing, so I've reworked it a lot, as below (pushing it shortly, but CI is getting tired, so will do it in a one-er shortly)

Also noticed we were completely ignore hierarchical IDs in < .NET Core 3 (i.e. `IActivity`) which didn't seem right?

```csharp
if (activity is IW3CActivity activity3)
{
    var activityTraceId = activity3.TraceId;
    var activitySpanId = activity3.SpanId;

    if (!StringUtil.IsNullOrEmpty(activityTraceId))
    {
        // W3C ID
        if (activity3 is { ParentSpanId: { } parentSpanId })
        {
            
            if (ActivityMappingById.TryGetValue(new ActivityKey(activityTraceId, parentSpanId), out ActivityMapping mapping))
            {
                parent = mapping.Scope.Span.Context;
            }
            else
            {
                // create a new parent span context for the ActivityContext
                // ... code elided, same as before
            }
        }
    }
    else
    {
        // No traceID, so much be Hierarchical ID
        if (activity3.ParentSpanId is { } parentSpanId)
        {
            Log.Error("Activity with ID {ActivityId} had parent span ID {ParentSpanId} but TraceID was missing", activity.Id, parentSpanId);
        }
        else
        {
            var parentId = activity3.ParentId;
            if (!StringUtil.IsNullOrEmpty(parentId) && ActivityMappingById.TryGetValue(new ActivityKey(parentId), out ActivityMapping mapping))
            {
                parent = mapping.Scope.Span.Context;
            }
        }
    }

    // ... Same "fixup" code as before, only for W3C ID spans
}
else
{
    // NEW BRANCH 
    // non-IW3CActivity, i.e. we're in .NET Core 2.x territory. Only have hierarchical IDs to worry about here
    var parentId = activity.ParentId;
    if (!StringUtil.IsNullOrEmpty(parentId) && ActivityMappingById.TryGetValue(new ActivityKey(parentId), out ActivityMapping mapping))
    {
        parent = mapping.Scope.Span.Context;
    }
}

```

tracer/src/Datadog.Trace/Activity/Handlers/ActivityHandlerCommon.cs
```diff
@@ -48,25 +47,27 @@ public static void ActivityStarted<T>(string sourceName, T activity, OpenTelemet
...
 
-            // for non-IW3CActivity interfaces we'll use Activity.Id and string.Empty as the key as they don't have a guaranteed TraceId+SpanId
-            // for IW3CActivity interfaces we'll use the Activity.TraceId + Activity.SpanId as the key
-            // have to also validate that the TraceId and SpanId actually exist and aren't null - as they can be in some cases
+            // for non-W3C activities using Hierarchical IDs (both IW3CActivity and IActivity) we use Activity.Id and string.Empty
+            // for W3C ID activities (always IW3CActivity) we'll use the Activity.TraceId + Activity.SpanId as the key
             ActivityKey? activityKey = null;
 
-            if (activity is IW3CActivity w3cActivity)
+            if (activity is IW3CActivity activity3)
```
lucaspimentel, line 54: Level of care low, but I actually prefer `w3cActivity` here because `activity3` could be taken to mean it's an `IActivity3` reference (even though `IActivity3` doesn't exist, but `IActivity5` and `IActivity6` do).

tracer/src/Datadog.Trace/Activity/Handlers/ActivityHandlerCommon.cs
```diff
@@ -85,43 +86,53 @@ public static void ActivityStarted<T>(string sourceName, T activity, OpenTelemet
                                 rawSpanId: parentSpanId);
                         }
                     }
+                }
+                else
+                {
+                    // No traceID, so much be Hierarchical ID
+                    if (activity3.ParentSpanId is { } parentSpanId)
+                    {
+                        // This is a weird scenario - we're in a hierarchical ID, we don't have a trace ID, but we _do_ have a _parentSpanID?!
```
lucaspimentel, line 96: Do we care about activities with hierarchical IDs? AFAIK those are not compatible with OpenTelemetry. Could we just ignore them? (general question, not specific to this PR)

except special cases like the aspnetcore observer, if we still use that without `DD_TRACE_OTEL_ENABLED`.

tracer/src/Datadog.Trace/Activity/Handlers/ActivityHandlerCommon.cs
```diff
@@ -85,43 +86,53 @@ public static void ActivityStarted<T>(string sourceName, T activity, OpenTelemet
...
 
-                        // SpanId (always 16 chars long)
-                        w3cActivity.ParentSpanId = activeSpan.Context.RawSpanId;
+                    // SpanId (always 16 chars long)
+                    activity3.RawParentSpanId = activeSpan.Context.RawSpanId;
 
-                        // We clear internals Id and ParentId values to force recalculation.
-                        w3cActivity.RawId = null;
-                        w3cActivity.RawParentId = null;
+                    // We clear internals Id and ParentId values to force recalculation.
```
lucaspimentel:
```suggestion
                    // We clear internal ids values to force recalculation.
```

tracer/src/Datadog.Trace/Activity/Handlers/ActivityHandlerCommon.cs
lucaspimentel: ahh, I see the comment below

> non-IW3CActivity, i.e. we're in .NET Core 2.x territory. Only have hierarchical IDs to worry about here

tracer/src/Datadog.Trace/Activity/Handlers/IgnoreActivityHandler.cs
```diff
@@ -30,36 +31,43 @@ internal sealed class IgnoreActivityHandler : IActivityHandler
...
-            if (ShouldIgnoreByOperationName(activity))
-            {
-                IgnoreActivity(activity, span);
-                return true;
-            }
+                    // SpanId (always 16 chars long)
+                    w3cActivity.RawParentSpanId = span.Context.RawSpanId;
 
-            return false;
+                    // We clear internals Id and ParentId values to force recalculation.
```
lucaspimentel:
```suggestion
                    // We clear internal ids to force recalculation.
```

lucaspimentel: alternatively
```suggestion
                    // We clear internal Id and ParentId values to force recalculation.
```

tracer/src/Datadog.Trace/Activity/Handlers/ActivityHandlerCommon.cs
lucaspimentel: alternatively
```suggestion
                    // We clear internal Id and ParentId values to force recalculation.
```

tracer/src/Datadog.Trace/Activity/Handlers/ActivityHandlerCommon.cs
bouwkast: Wow thanks and nice catch 👍

tracer/src/Datadog.Trace/Activity/Handlers/ActivityHandlerCommon.cs
andrewlock: > Level of care low, but I actually prefer w3cActivity here because activity3 could be taken to mean it's an IActivity3 reference

I have a pretty strong opinion on _not_ using `w3cActivity`, because it comes with implicit assumptions (I know, because the code made them, as did I 😅)

> even though IActivity3 doesn't exist, but IActivity5 and IActivity6 do)

Yeah I'm going to rename `IW3CActitvity` to `IActivity3` later, when it's less conflicting 😉

tracer/src/Datadog.Trace/Activity/Handlers/ActivityHandlerCommon.cs
andrewlock: > Could we just ignore them? (general question, not specific to this PR)

Great question, _maybe_ we could 🤔 Big breaking change though...

bouwkast (approved):
👍 lot easier to follow the various logic with the refactorings/renaming/comments thanks!

zacharycmontoya (approved):
Thanks for the refactoring and fixes!!

PR #8041, opened 2026-01-08
Switch to allocation-free enumeration of `Activity` objects

andrewlock:
## Summary of changes

Update code that enumerates properties of `IActivity` duck types to use allocation-free enumeration

## Reason for change

Our duck types all use the public APIs of `Activity` to grab tags, events, and links. These types are generally implemented internally as custom linked list objects, but are exposed as `IEnumerable<>` types. The internal types implement a `struct`-based enumerator, but as the types aren't exposed directly, the compiler can't use those and ends up allocating. That's... annoying.

To work around that, this adds a "hack" that OTel were [actually using themselves](https://github.com/open-telemetry/opentelemetry-dotnet/blob/73bff75ef653f81fe6877299435b21131be36dc0/src/OpenTelemetry/Internal/EnumerationHelper.cs#L58) (and [which has been discussed elsewhere](https://www.macrosssoftware.com/2020/07/13/enumerator-performance-surprises/)), of building a dynamic method that finds the `GetEnumerator` type and uses that. We can't use it everywhere because we don't have the types available (e.g. links and events), but we _can_ use it for tags. There's essentially a trade-off of startup time vs runtime allocation, which is worth it in this case IMO. 

Additionally, the `Activity` implementation uses an empty array as the return value when there _are_ no tags/links/events, so we can avoid the enumerator allocation in those cases by specifically looking for an empty array and bailing if so. 

## Implementation details

I chose to add the `EnumerationHelper` as an explicitly `Activity`-related thing for now, but we can consider exposing this more widely if we get confidence.

I also avoided the concurrent dictionary approach that is the "default" for this, and instead added an additional helper to read a static field that we populate once. We have to wait until we have a "real" activity to do the population, so that we get the right type, which is a bit annoying, but it works...

## Test coverage

Added a couple of unit tests for the result and confirmed it works as expected. Benchmarks vs #8040 shw a nice allocation improvement:

| Method                          | Runtime              |       Mean |    StdDev |   Gen0 |   Gen1 | Allocated | Compared to #8040 |
| ------------------------------- | -------------------- | ---------: | --------: | -----: | -----: | --------: | ----------------: |
| StartStopWithChild_Hierarchical | .NET 6.0             | 3,248.6 ns |  28.71 ns | 0.0114 | 0.0038 |    3.6 KB |          -0.19 KB |
| StartStopWithChild_Hierarchical | .NET 8.0             | 2,651.1 ns |  80.80 ns | 0.0114 |      - |    3.6 KB |          -0.19 KB |
| StartStopWithChild_Hierarchical | .NET Core 3.1        | 4,482.2 ns | 263.51 ns | 0.0153 |      - |   3.83 KB |          -0.19 KB |
| StartStopWithChild_Hierarchical | .NET Framework 4.7.2 | 4,915.8 ns |  63.81 ns | 0.7019 |      - |   4.33 KB |          -0.18 KB |
| StartStopWithChild              | .NET 6.0             | 3,176.8 ns |  37.73 ns | 0.0114 |      - |   3.59 KB |          -0.19 KB |
| StartStopWithChild              | .NET 8.0             | 2,598.0 ns |  40.05 ns | 0.0114 |      - |   3.59 KB |          -0.19 KB |
| StartStopWithChild              | .NET Core 3.1        | 4,260.7 ns |  59.98 ns | 0.0076 |      - |   3.78 KB |          -0.19 KB |
| StartStopWithChild              | .NET Framework 4.7.2 | 5,097.7 ns |  46.98 ns | 0.6790 |      - |    4.2 KB |          -0.19 KB |

## Other details

- https://github.com/DataDog/dd-trace-dotnet/pull/8041 👈

### DIFF
--- a/tracer/src/Datadog.Trace/Activity/Helpers/AllocationFreeEnumerator.cs
+++ b/tracer/src/Datadog.Trace/Activity/Helpers/AllocationFreeEnumerator.cs
@@ -5,7 +5,7 @@
 
 #nullable enable
 
-// Thus is based on code from
+// This is based on code from
 // https://github.com/open-telemetry/opentelemetry-dotnet/blob/73bff75ef653f81fe6877299435b21131be36dc0/src/OpenTelemetry/Internal/EnumerationHelper.cs#L96
 // and
 // https://github.com/Macross-Software/core/blob/a4ed6770099826af3cbfd07d1442af68bc664526/ClassLibraries/Macross.Performance.Extensions/Code/StructEnumerator.cs


tracer/src/Datadog.Trace/Activity/Helpers/ActivityEnumerationHelper.cs
```diff
@@ -0,0 +1,113 @@
...
+using System.Threading;
+using Datadog.Trace.Activity.DuckTypes;
+
+namespace Datadog.Trace.Activity.Helpers;
+
+internal static class ActivityEnumerationHelper
+{
+    private static AllocationFreeEnumerator<IEnumerable<KeyValuePair<string, object?>>, KeyValuePair<string, object?>, OtelTagsEnumerationState>.AllocationFreeForEachDelegate? _tagObjectsEnumerator;
+    private static AllocationFreeEnumerator<IEnumerable<KeyValuePair<string, string?>>, KeyValuePair<string, string?>, OtelTagsEnumerationState>.AllocationFreeForEachDelegate? _tagsEnumerator;
+#if DEBUG
```
bouwkast: What exactly are the `#if DEBUG` for?

tracer/src/Datadog.Trace/Activity/Helpers/AllocationFreeEnumerator.cs
bouwkast, line 1: Does this file need to be named `AllocationFreeEnumerator` as it is in the DD copyright header?

tracer/src/Datadog.Trace/Activity/Helpers/OtelTagsEnumerationState.cs
```diff
@@ -0,0 +1,15 @@
...
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
+// </copyright>
+
+#nullable enable
+
+using Datadog.Trace.Tagging;
+
+namespace Datadog.Trace.Activity.Helpers;
+
+internal readonly struct OtelTagsEnumerationState(Span span)
```
bouwkast, line 15: This is to avoid the closure allocation right? It might be good to leave a comment stating that so it doesn't get refactored away accidentally

tracer/src/Datadog.Trace/Activity/Helpers/ActivityEnumerationHelper.cs
andrewlock: The "risk" with these helpers are that you must always call this method with the same runtime-type of object. This is _generally_ guaranteed with the current usage, but there's a risk, if you use it incorrectly, that we pass in a different type. The `DEBUG` were intended as some simple guard rails to try to catch invalid behaviour locally, without adding overhead at runtime.

There are a couple of (safer) alternatives
- Add the `GetType()` call in every case, and bail if it's wrong. Small bit of extra overhead, for runtime safety
- Reinstate the `ConcurrentDictionary` lookup based on `GetType()`, and build new enumerators as necessary. This is slower (due to the dictionary lookup), but it does guarantee we'll always be able to do the lookup even if they do more stuff like "return type A in some cases, and type B in other cases"...

I'm ok with those other two options, I've just been pushing hard for speed, so was erring on that side, but I've maybe gone too far now 😅

tracer/src/Datadog.Trace/Activity/Helpers/AllocationFreeEnumerator.cs
andrewlock: Yep, fixed now, thanks 🙂

tracer/src/Datadog.Trace/Activity/Helpers/OtelTagsEnumerationState.cs
andrewlock: Yeah, essentially, but mostly it's because `AllocationFreeEnumerator` _requires_ the state is a `struct` 😄 I'll add a comment to it 👍

tonyredondo (approved):
👏🏻 👏🏻 👏🏻

tracer/src/Datadog.Trace/Activity/Helpers/AllocationFreeEnumerator.cs
```diff
@@ -0,0 +1,167 @@
+// <copyright file="AllocationFreeEnumerator.cs" company="Datadog">
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2 License.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
+// </copyright>
+
+#nullable  enable
+
+// Thus is based on code from
```
lucaspimentel:
```suggestion
// This is based on code from
```

lucaspimentel (approved):

tracer/src/Datadog.Trace/Activity/Helpers/ActivityEnumerationHelper.cs
andrewlock: After some soul searching (thanks @tonyredondo) decided to refactor this to be always safe, instead of only in `DEBUG` mode 😄 Now, if the provided type doesn't match the provided type it's an error, but we fallback to doing an allocating enumeration instead

PR #8042, opened 2026-01-08
Improve `OpenTelemetryTags` population to reduce allocation and work

andrewlock:
## Summary of changes

Improve performance around the "population" of tags from an `Activity` into a `Span`

## Reason for change

Currently we do a _lot_ of allocation in the `OtlpHelpers.AgentConvertSpan` method. The `OpenTelemetryTags` object also has a lot of properties on it that we never set, and are only used for _reading_ tags in the `OperationMapper`, even though many may never even be read. This increases the size of the tags object.

This PR aims to do various optimizations around the "close activity" paths:
- Reduce the size of the `OpenTelemetryTags` object by removing properties only used by `OperationMapper`
- Add properties to `OpenTelemetryTags` for values that we explictly set in `OtlpHelpers.AgentConvertSpan`

## Implementation details

- Currently we're always creating the `OpenTelemetryTags`, even though we _may_ throw it away if the activity is ignored, so that's an easy win.
- I think we can assume that the tags object passed to `AgentConvertSpan` is only ever `OpenTelemetryTags` based on the call sites, so we can interact with the tags object directly where posisble.
- "Simplify" some of the code paths by avoiding `span.SetTag` and favouring `tags.SetTag` when we know it's not a "special" tag.
- Inline a few method calls in places where they will often not be called, at the expense of some clarity
- Refactor `OperationMapper` to use the `GetTag()` API, and cache the values at various points. The resulting code is uglier, but overall means we reduce ~100 bytes off every span, so I think it's worth it

## Test coverage

Functionality is covered by existing, this is just a refactoring, and there's good tests for `OperationMapper` currently. Benchmarks vs https://github.com/DataDog/dd-trace-dotnet/pull/8041 show a nice allocation improvement, at the expense of slower execution (this was local though, so not 100% on that, will see what the CI results show). If there is a significant slow down, I'll try to isolate it, but I think the allocation improvments are probably worth it either way

| Method                          | Runtime              |     Mean |    StdDev |   Gen0 | Allocated | Compared to #8041 |
| ------------------------------- | -------------------- | -------: | --------: | -----: | --------: | ----------------: |
| StartStopWithChild_Hierarchical | .NET 6.0             | 4.438 us | 0.7347 us | 0.0076 |    3.2 KB |          -0.40 KB |
| StartStopWithChild_Hierarchical | .NET 8.0             | 3.216 us | 0.4260 us | 0.0076 |    3.2 KB |          -0.40 KB |
| StartStopWithChild_Hierarchical | .NET Core 3.1        | 5.293 us | 0.8316 us | 0.0076 |   3.42 KB |          -0.41 KB |
| StartStopWithChild_Hierarchical | .NET Framework 4.7.2 | 5.735 us | 0.5592 us | 0.6332 |   3.92 KB |          -0.41 KB |
|                                 |                      |          |           |        |           |                   |
| StartStopWithChild              | .NET 6.0             | 3.848 us | 0.5325 us | 0.0076 |   3.19 KB |          -0.40 KB |
| StartStopWithChild              | .NET 8.0             | 3.080 us | 0.2698 us | 0.0076 |   3.19 KB |          -0.40 KB |
| StartStopWithChild              | .NET Core 3.1        | 5.094 us | 0.4982 us | 0.0076 |   3.38 KB |          -0.40 KB |
| StartStopWithChild              | .NET Framework 4.7.2 | 6.172 us | 0.5948 us | 0.6104 |   3.79 KB |          -0.41 KB |

## Other details

Added a couple of extra tests that were missing from the previous PR in the stack too

- https://github.com/DataDog/dd-trace-dotnet/pull/8042 👈

### DIFF
--- a/tracer/src/Datadog.Trace/Activity/OtlpHelpers.cs
+++ b/tracer/src/Datadog.Trace/Activity/OtlpHelpers.cs
@@ -70,18 +70,6 @@ namespace Datadog.Trace.Activity
                 tags.OtelTraceId = w3cActivity.TraceId;
             }
 
-            // Fixup "version" tag
-            // Fallback to static instance if no tracer associated with the trace
-            var traceContext = span.Context.TraceContext;
-            var tracer = traceContext?.Tracer ?? Tracer.Instance;
-            // TODO: This is checking that the tag is being set on the _span_ not on the _activity_, is that really correct?
-            // If it's correct, we should probably update the SetTag logic and remove this?
-            if (tracer.PerTraceSettings.Settings.ServiceVersion is null
-             && span.GetTag("service.version") is { Length: > 1 } otelServiceVersion)
-            {
-                traceContext?.ServiceVersion = otelServiceVersion;
-            }
-
             // Copy over tags from Activity to the Datadog Span
             // Starting with .NET 5, Activity can hold tags whose value have type object?
             // For runtimes older than .NET 5, Activity can only hold tags whose values have type string
@@ -106,6 +94,15 @@ namespace Datadog.Trace.Activity
                     OtlpHelpers.SetTagObject(s.Span, kvp.Key, kvp.Value);
                     return true;
                 });
+            }
+
+            // Fixup "version" tag
+            var traceContext = span.Context.TraceContext;
+            if (traceContext is not null
+             && string.IsNullOrEmpty(traceContext.ServiceVersion)
+             && span.GetTag("service.version") is { Length: > 0 } otelServiceVersion)
+            {
+                traceContext.ServiceVersion = otelServiceVersion;
             }
 
             // Additional Datadog policy: Set tag "span.kind"


tracer/src/Datadog.Trace/Activity/OperationNameMapper.cs
```diff
@@ -32,127 +32,99 @@ internal static void MapToOperationName(Span span)
...
-                operationName = "http.server.request";
-            }
-            else if (tags.IsHttpClient())
+            span.OperationName = GetOperationName(tags);
+        }
+
+        internal static string GetOperationName(OpenTelemetryTags tags)
+        {
+            var httpRequestMethod = tags.GetTag("http.request.method");
+            if (tags.SpanKind == SpanKinds.Server && !string.IsNullOrEmpty(httpRequestMethod))
```
andrewlock: For each of these, the method was basically inlined, then reused tags we'd already fetched where possible

tracer/src/Datadog.Trace/Activity/OtlpHelpers.cs
```diff
@@ -57,16 +67,19 @@ private static void AgentConvertSpan<TInner>(TInner activity, Span span)
...
+                tags.OtelTraceId = w3cActivity.TraceId;
             }
 
             // Fixup "version" tag
             // Fallback to static instance if no tracer associated with the trace
-            var tracer = span.Context.TraceContext?.Tracer ?? Tracer.Instance;
+            var traceContext = span.Context.TraceContext;
+            var tracer = traceContext?.Tracer ?? Tracer.Instance;
+            // TODO: This is checking that the tag is being set on the _span_ not on the _activity_, is that really correct?
+            // If it's correct, we should probably update the SetTag logic and remove this?
```
andrewlock: Just highlighting that we haven't yet copied activity tags to the span yet, so there _could_ be a bug here if we're assuming that we have?

tracer/src/Datadog.Trace/Activity/OtlpHelpers.cs
```diff
@@ -130,44 +147,51 @@ private static void AgentConvertSpan<TInner>(TInner activity, Span span)
...
                 }
                 else
                 {
-                    span.SetTag("otel.status_code", "STATUS_CODE_UNSET");
+                    tags.OtelStatusCode = "STATUS_CODE_UNSET";
                 }
             }
 
             // Map the OTEL status to error tags
-            AgentStatus2Error(activity, span);
```
andrewlock, line 168: Inlined this method because it makes it simpler in terms of the tags and a couple of other things

tracer/src/Datadog.Trace/Tagging/AzureServiceBusTags.cs
```diff
@@ -23,6 +23,12 @@ internal partial class AzureServiceBusTags : OpenTelemetryTags
         [Metric(Trace.Tags.Analytics)]
         public double? AnalyticsSampleRate { get; set; }
 
+        [Tag(Tags.MessagingSystem)]
+        public string MessagingSystem { get; set; }
```
andrewlock, line 27: We removed these from `OpenTelemetryTags` but the service bus integration does actually set them

tracer/src/Datadog.Trace/Activity/OtlpHelpers.cs
```diff
@@ -103,20 +116,24 @@ private static void AgentConvertSpan<TInner>(TInner activity, Span span)
...
+                span.OperationName = OperationNameMapper.GetOperationName(tags);
+            }
 
             // TODO: Add container tags from attributes if the tag isn't already in the span
 
             // Fixup "env" tag
-            if (span.Context.TraceContext?.Environment is null
-                && span.GetTag("deployment.environment") is { Length: > 0 } otelServiceEnv)
+            if (traceContext is not null
+             && traceContext.Environment is null
```
anna-git, line 128: isn't this a different behavior than `span.Context.TraceContext?.Environment is null`  , I mean before if traceContext was null it would be true, and now it wouldn't

tracer/src/Datadog.Trace/Activity/OtlpHelpers.cs
```diff
@@ -130,44 +147,51 @@ private static void AgentConvertSpan<TInner>(TInner activity, Span span)
...
-                    span.SetTag("otel.library.version", activity5.Source.Version);
+                    tags.OtelLibraryVersion = activity5.Source.Version;
                 }
             }
 
             // Set OTEL status code and OTEL status description
-            if (span.GetTag("otel.status_code") is null)
+            if (tags.OtelStatusCode is null)
             {
                 if (activity6 is not null)
```
anna-git, line 157: nit: activity6 could be declared closer to usage?

tracer/src/Datadog.Trace/Activity/OtlpHelpers.cs
andrewlock: Technically, that's true, but in the code for `span.SetTag` we then check if the `TraceContext` is `null`, and if it is, we don't change anything 😄 (technically we write an error log, but in this case I don't think that adds much value, and should never happen)

tracer/src/Datadog.Trace/Activity/OtlpHelpers.cs
andrewlock: All of the activities are created using `as ` because they're originally structs, and this causes boxing, so I think it's best if we continue to keep them all together at the top?

I have ideas about removing _all_ that boxing later, but I'll have to see how big of a deal that is later 😄

tracer/src/Datadog.Trace/Activity/OperationNameMapper.cs
```diff
@@ -32,127 +32,99 @@ internal static void MapToOperationName(Span span)
...
+                return "http.server.request";
             }
-            else if (tags.IsDatabase())
+
+            if (tags.SpanKind == SpanKinds.Client && !string.IsNullOrEmpty(httpRequestMethod))
             {
-                operationName = $"{tags.DbSystem}.query";
+                // IsHttpClient
+                return "http.client.request";
             }
```
lucaspimentel, line 54: Could we nest the `if` conditions like this instead? Feel a little cleaner to me (not important, though). I'm not sure if you intentionally wanted to check `tags.SpanKind` first before `httpRequestMethod`, or if this was just the result of inlining the previous methods.

```csharp
if (!StringUtil.IsNullOrEmpty(tags.GetTag("http.request.method"))
{
    if (tags.SpanKind == SpanKinds.Server)
    {
        // IsHttpServer
        return "http.server.request";
    }
    
    if (tags.SpanKind == SpanKinds.Client)
    {
        // IsHttpClient
        return "http.client.request";
    }
}
```

tracer/src/Datadog.Trace/Activity/OperationNameMapper.cs
```diff
@@ -32,127 +32,99 @@ internal static void MapToOperationName(Span span)
...
-            else if (tags.IsMessaging())
+
+            var dbSystem = tags.GetTag("db.system");
+            if (tags.SpanKind == SpanKinds.Client && !StringUtil.IsNullOrEmpty(dbSystem))
```
lucaspimentel: Can we skip getting the tag if span kind is not client?

```csharp
    if (tags.SpanKind == SpanKinds.Client &&
        tags.GetTag("db.system") is { Length: > 0 } dbSystem)
    {
        // IsDatabase
        return $"{dbSystem.ToLowerInvariant()}.query";
    }
```

tracer/src/Datadog.Trace/Activity/OperationNameMapper.cs
```diff
@@ -32,127 +32,99 @@ internal static void MapToOperationName(Span span)
...
             {
-                operationName = $"{tags.MessagingSystem}.{tags.MessagingOperation}";
+                // IsDatabase
+                return $"{dbSystem.ToLowerInvariant()}.query";
             }
-            else if (tags.IsAwsClient())
+
+            var messagingSystem = tags.GetTag(Tags.MessagingSystem);
+            var messagingOperation = tags.GetTag(Tags.MessagingOperation);
```
lucaspimentel: Similar comment as `dbSystem` above.

tracer/src/Datadog.Trace/Activity/OperationNameMapper.cs
```diff
@@ -32,127 +32,99 @@ internal static void MapToOperationName(Span span)
...
+                  && !StringUtil.IsNullOrEmpty(messagingSystem) &&
+                     !StringUtil.IsNullOrEmpty(messagingOperation))
             {
-                operationName = !string.IsNullOrEmpty(tags.RpcService) ? $"aws.{tags.RpcService}.request" : "aws.client.request";
+                // IsMessaging
+                return $"{messagingSystem}.{messagingOperation}".ToLowerInvariant();
             }
-            else if (tags.IsRpcClient())
+
+            var rpcSystem = tags.GetTag(Tags.RpcSystem);
```
lucaspimentel, line 70: It looks like the checks for `rpcSystem` could also use a little refactoring.

```csharp
if (tags.GetTag(Tags.RpcSystem) is { Length: > 0 } rpcSystem)
{
    switch (tags.SpanKind)
    {
        case SpanKinds.Client when string.Equals(rpcSystem, "aws-api", StringComparison.OrdinalIgnoreCase):
        {
            // IsAwsClient
            return tags.GetTag(Tags.RpcService) is { Length: > 0 } service ?
                       $"aws.{service.ToLowerInvariant()}.request" :
                       "aws.client.request";
        }

        case SpanKinds.Client:
            // IsRpcClient
            return $"{rpcSystem.ToLowerInvariant()}.client.request";

        case SpanKinds.Server:
            // IsRpcServer
            return $"{rpcSystem.ToLowerInvariant()}.server.request";
    }
}
```

tracer/src/Datadog.Trace/Activity/OperationNameMapper.cs
```diff
@@ -32,127 +32,99 @@ internal static void MapToOperationName(Span span)
...
+
+            if (tags.SpanKind == SpanKinds.Server && !StringUtil.IsNullOrEmpty(rpcSystem))
             {
-                operationName = $"{tags.FaasTrigger}.invoke";
+                // IsRpcServer
+                return $"{rpcSystem.ToLowerInvariant()}.server.request";
             }
-            else if (tags.IsFaasClient())
+
+            var faasTrigger = tags.GetTag("faas.trigger");
```
lucaspimentel: similar to above: only get `faas.trigger` tag if span kind is server?

tracer/src/Datadog.Trace/Activity/OperationNameMapper.cs
```diff
@@ -32,127 +32,99 @@ internal static void MapToOperationName(Span span)
...
+            if (tags.SpanKind == SpanKinds.Server && !StringUtil.IsNullOrEmpty(faasTrigger))
             {
-                operationName = $"{tags.FaasInvokedProvider}.{tags.FaasInvokedName}.invoke";
+                // IsFaasServer
+                return $"{faasTrigger.ToLowerInvariant()}.invoke";
             }
-            else if (tags.IsGraphQLServer())
+
+            var faasInvokedProvider = tags.GetTag("faas.invoked_provider");
+            var faasInvokedName = tags.GetTag("faas.invoked_name");
```
lucaspimentel: similar to above: can we skip getting these tags unless span kind is client?

tracer/src/Datadog.Trace/Activity/OtlpHelpers.cs
lucaspimentel: > isn't this a different behavior...?

Good catch, though! I'd say this was a small oversight in the previous code. There was no point in trying to set `env` without a `TraceContext`.

tracer/src/Datadog.Trace/Activity/OperationNameMapper.cs
andrewlock: > Could we nest the if conditions like this instead?

Yes, I think in this case, we _could_, technically, the behaviour would be identical

> Feel a little cleaner to me (not important, though).

Yeah, I'm not a big fan of the code I refactored _into_ in general, but performance says yes 😅 

> I'm not sure if you intentionally wanted to check tags.SpanKind first before httpRequestMethod, or if this was just the result of inlining the previous methods.

It's basically the latter, there's lots of very specific ordering in here in general, so I didn't want to mess with it too much, as much to make sure that I didn't break anything! 😅 

There's also an element of the fact that these checks as-written map one-to-one with the remapping in the agent, so if/when we need to tweak this, it's more clear.

That said... what are the chances we're going to do that anytime soon? So I'll apply all your cleanup tweaks 🙂

andrewlock: Done in https://github.com/DataDog/dd-trace-dotnet/pull/8042/commits/ab631ca60fb1a4b3a383c76de4b99f335f39513e

tracer/src/Datadog.Trace/Activity/OperationNameMapper.cs
andrewlock: Done in https://github.com/DataDog/dd-trace-dotnet/pull/8042/commits/ab631ca60fb1a4b3a383c76de4b99f335f39513e

tracer/src/Datadog.Trace/Activity/OperationNameMapper.cs
andrewlock: Done in https://github.com/DataDog/dd-trace-dotnet/pull/8042/commits/ab631ca60fb1a4b3a383c76de4b99f335f39513e

tracer/src/Datadog.Trace/Activity/OperationNameMapper.cs
andrewlock: Done in https://github.com/DataDog/dd-trace-dotnet/pull/8042/commits/ab631ca60fb1a4b3a383c76de4b99f335f39513e (though stuck with `if` instead of switch because meh

tracer/src/Datadog.Trace/Activity/OperationNameMapper.cs
andrewlock: done in https://github.com/DataDog/dd-trace-dotnet/pull/8042/commits/ab631ca60fb1a4b3a383c76de4b99f335f39513e

tracer/src/Datadog.Trace/Activity/OperationNameMapper.cs
andrewlock: Done in https://github.com/DataDog/dd-trace-dotnet/pull/8042/commits/ab631ca60fb1a4b3a383c76de4b99f335f39513e

tracer/src/Datadog.Trace/Activity/OtlpHelpers.cs
```diff
@@ -57,16 +67,19 @@ private static void AgentConvertSpan<TInner>(TInner activity, Span span)
...
-            var tracer = span.Context.TraceContext?.Tracer ?? Tracer.Instance;
+            var traceContext = span.Context.TraceContext;
+            var tracer = traceContext?.Tracer ?? Tracer.Instance;
+            // TODO: This is checking that the tag is being set on the _span_ not on the _activity_, is that really correct?
+            // If it's correct, we should probably update the SetTag logic and remove this?
             if (tracer.PerTraceSettings.Settings.ServiceVersion is null
              && span.GetTag("service.version") is { Length: > 1 } otelServiceVersion)
             {
-                span.SetTag(Tags.Version, otelServiceVersion);
+                traceContext?.ServiceVersion = otelServiceVersion;
```
bouwkast: Unsure about this change, maybe it is right?

https://github.com/DataDog/datadog-agent/blob/67c353cff1a6a275d7ce40059aad30fc6a3a0bc1/pkg/trace/api/otlp.go#L474-L478

Is what it was based on, which was similar but different to ours 😕 


Wondering if instead, after we copy everything over if we are to do something like so?

```csharp
  if (span.GetTag(Tags.Version) is null
   && span.GetTag("service.version") is { Length: > 0 } serviceVersion)
  {
      span.SetTag(Tags.Version, serviceVersion);
  }
```

andrewlock: Fixed as discussed 👍

PR #8043, opened 2026-01-08
[SVLS-7360] Fix Exception Replay in Lambda

shreyamalpani:
## Summary of changes
- Sends the placeholder tracer span to the Lambda extension as `dd-tracer-serverless-span`
- Sets the Lambda request ID on the span so that the extension can copy the tracer tags to the `aws.lambda` span that is created in the extension

## Reason for change
To support Exception Replay in Lambda

## Test coverage
Manually tested with Lambda
image

## Other details

[SVLS-7360](https://datadoghq.atlassian.net/browse/SVLS-7360)

[SVLS-7360]: https://datadoghq.atlassian.net/browse/SVLS-7360?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tracer/test/snapshots/AwsLambdaTests.verified.txt
```diff
@@ -415,10 +415,7 @@
       runtime-id: Guid_1
     },
     Metrics: {
-      process_id: 0,
-      _dd.top_level: 1.0,
-      _dd.tracer_kr: 1.0,
-      _sampling_priority_v1: 1.0
+      _dd.top_level: 1.0
```
lucaspimentel, line 418: Do we know why the snapshot changed like this? All these spans long these 3 attributes and there seem to be a lot of new spans. Without looking too deep into it, my guess is that the new spans are new root spans so the attributes are being added there instead.

lucaspimentel:
@codex review

tracer/test/snapshots/AwsLambdaTests.verified.txt
shreyamalpani: Yup, the new spans are the new root spans and these tags are present on those spans

lucaspimentel: oops, it's right in the PR description, my bad!

> Sends the placeholder tracer span to the Lambda extension as `dd-tracer-serverless-span`

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Lambda/LambdaCommon.cs
```diff
@@ -20,35 +20,45 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.AWS.Lambda;
...
 
         var span = tracer.StartSpan(
-            PlaceholderOperationName,
+            operationName: InvocationSpanName,
             tags: null,
             parent: context.SpanContext,
-            serviceName: PlaceholderServiceName,
-            addToTraceContext: false);
+            serviceName: InvocationSpanName,
+            addToTraceContext: true);
```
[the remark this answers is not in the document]
lucaspimentel: This is a good call out, and I believe this is fine as long as the agent (in this case the "Lambda Extension" aka Bottlecap) is handling it correctly.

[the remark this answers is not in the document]
lucaspimentel: We should also consider what happens when using a newer version of the Extension layer with an older version of the tracer layer, and vice-versa. Is this change both backwards- and forwards-compatible?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Lambda/LambdaCommon.cs
```diff
@@ -20,35 +20,45 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.AWS.Lambda;
...
+
+        // The Lambda extension uses the resource name to identify placeholder span
+        span.ResourceName = InvocationSpanName;
+
+        // Need to set request_id to copy tracer tags to the aws.lambda span
+        if (awsRequestId != null)
+        {
+            span.SetTag("request_id", awsRequestId);
```
lucaspimentel, line 46: `request_id` is quite ambiguous. Is this the attribute name we're already using elsewhere (in the Extension and other client libraries)? If so, it may be too late to change it.

But we could've used `aws.requestid` instead which we already use when instrumenting the AWS SDK.

lucaspimentel: fwiw, we already send the request id in the http headers we send to the Extension's `/lambda/start-invocation` and `/lambda/end-invocation` endpoints. See #7835.

If this has to be in a span attribute for some reason, it seems wasteful to send a whole additional span (`dd-tracer-serverless-span`) only to carry the request id attribute. Couldn't we add the attribute to the existing `lambda.invocation` span?

lucaspimentel (changes requested):
Code LGTM. This "request changed" is just to prevent merging the PR until some questions I have are answered.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Lambda/LambdaCommon.cs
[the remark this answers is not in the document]
shreyamalpani: We previously were not actually copying the .NET tracer tags, since the extension looks for the span resource `dd-tracer-serverless-span` [here](https://github.com/DataDog/datadog-lambda-extension/blob/2be85c6cb00e2cb83821f2df6af1206385990390/bottlecap/src/traces/trace_agent.rs#L586C16-L593C18) which was never being sent, so this fixes that issue.

The extension always filtered that span and I verified that traces look the same as before, so I think it should be good.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Lambda/LambdaCommon.cs
shreyamalpani: Yup, the Lambda extension looks for the `dd-tracer-serverless-span` span with the matching `request_id` field to enrich the new invocation span [here](https://github.com/DataDog/datadog-lambda-extension/blob/2be85c6cb00e2cb83821f2df6af1206385990390/bottlecap/src/lifecycle/invocation/processor.rs#L1299C5-L1307C6). We want to send this span so that we are able to copy the tracer tags, including the Exception replay snapshot identifiers, to the new invocation span.

PR #8044, opened 2026-01-09
Fix build by splitting `FleetInstaller` integration tests into their own test project

andrewlock:
## Summary of changes

- Move `FleetInstaller` integration tests to their own project
- Only run those tests on .NET FX x64 (all that we support)

## Reason for change

The integration tests on .NET Framrework, x86, are failing due to a missing binary, and have been for some time 😬 This fixes those tests so that they actually run. Separately we will do work to detect if/when that scenario happens again.

## Implementation details

- Create a new test project for fleet installer
- Move the existing tests to that project
- Don't try to build/run the project unless we're on .NET Framework x64

Before resorting to this approach, I tried a bunch of other things, such as building the project for `AnyCPU` instead of the `x64` that we currently have. However, then NuGet restore failed, which I remember is the reason why it is like this in the first place 🙄 

Its not pretty, but at least now it works

## Test coverage

This is the test. [They're actually running now](https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=193673&view=logs&j=cc578895-adfa-5051-1356-3b220e12921d&t=1c63e335-ed52-58c7-827e-80e5f4113871).

### DIFF
--- a/tracer/test/Datadog.FleetInstaller.IntegrationTests/Datadog.FleetInstaller.IntegrationTests.csproj
+++ b/tracer/test/Datadog.FleetInstaller.IntegrationTests/Datadog.FleetInstaller.IntegrationTests.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <!-- only run .NET Framework tests on Windows -->
-    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">net48;net10.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">net48</TargetFrameworks>
     <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">net10.0</TargetFrameworks>
     <OutputType>Library</OutputType>
   </PropertyGroup>


tracer/test/Datadog.FleetInstaller.IntegrationTests/Datadog.FleetInstaller.IntegrationTests.csproj
```diff
@@ -0,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <!-- only run .NET Framework tests on Windows -->
+    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">net48;net10.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">net10.0</TargetFrameworks>
```
andrewlock: We don't actually want to run this project on .NET 10, but otherwise we get into a weird state on non-Windows 😅

NachoEchevarria: And why to include .net 10 for Windows too?

andrewlock: Mostly for consistency with how we do this elsewhere, but it's a good point, I'll try removing it 👍 
```suggestion
    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">net48</TargetFrameworks>
    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">net10.0</TargetFrameworks>
```

PR #8048, opened 2026-01-12
[Dynamic Instrumentation] DEBUG-4142 Subscribe to settings change in the debugger manager

dudikeleti:
## Reason for change
Dynamic Instrumentation (and other debugger products) previously used the tracer’s initial/fallback service name when the user didn’t set DD_SERVICE via environment/config, even if the user later set the service name in code. This caused debugger payloads (snapshots/diagnostics) to be attributed to the wrong service.
This change ensures debugger products track the current tracer service name and behave consistently with the tracer configuration.

## Implementation details
DebuggerManager subscribes to TracerSettings.SettingsManager.SubscribeToChanges(...) and updates ServiceName whenever MutableSettings changes.
Debugger components that need the service name use a Func<string> provider so they read the latest value when emitting data.

## Test coverage
Added a debugger integration test validating service-name updates via code:
New sample test-run: `ChangeServiceNameInCodeTest` emits one snapshot before and one snapshot after calling Tracer.Configure(...) with a new ServiceName.
New integration test: ServiceNameChangedInCode_IsReflectedInSnapshots installs two probes, collects two snapshots, and asserts:
the “before” snapshot has a default service name (exe name)
the “after” snapshot has service from code

## Other details
Not supported in this PR
DogStatsD / metric probes: service name updates are not currently propagated to the DogStatsD client used by debugger metric probes.

Symbol uploader: the symbol uploader intentionally captures the service name on first use and keeps it fixed for the uploader’s lifetime to avoid mixing symbols across services. Supporting service name changes there would require stopping/recreating the uploader (expensive) and is deferred until there’s an explicit requirement.

### DIFF
--- a/tracer/test/Datadog.Trace.Tests/Debugger/ProbeStatusSinkTests.cs
+++ b/tracer/test/Datadog.Trace.Tests/Debugger/ProbeStatusSinkTests.cs
@@ -26,7 +26,7 @@ namespace Datadog.Trace.Tests.Debugger
             Clock.SetForCurrentThread(_timeLord);
 
             _settings = DebuggerSettings.FromDefaultSource();
-            _sink = DiagnosticsSink.Create(() => "test", _settings);
+            _sink = DiagnosticsSink.Create(() => "ProbeStatusSinkTests", _settings);
         }
 
         [Fact]


tracer/src/Datadog.Trace/Debugger/Upload/SymbolUploadApi.cs
```diff
@@ -51,11 +52,13 @@ internal static IBatchUploadApi Create(
             IApiRequestFactory apiRequestFactory,
             IDiscoveryService discoveryService,
             IGitMetadataTagsProvider gitMetadataTagsProvider,
-            string serviceName,
+            Func<string> serviceNameProvider,
             bool enableCompression)
         {
             ArraySegment<byte> GetEventMetadataAsArraySegment()
             {
+                // Capture service name on first use, and keep it fixed for the lifetime of this API instance
```
andrewlock, line 60: How long does the API instance last? Is it recreated when the settings change? If not, won't it send the wrong settings? And if so, we probably don't need to use a `Func<string>`?

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -191,6 +200,58 @@ private void OneTimeSetup(TracerSettings tracerSettings)
...
         }
 
+        private void EnsureTracerSettingsSubscription(TracerSettings tracerSettings)
+        {
+            if (_processExit.Task.IsCompleted)
+            {
+                return;
+            }
+
+            // If the global manager replaced, UpdateConfiguration can be called with a different instance.
```
andrewlock, line 210: there should _never_ be another `TracerSettings` instance 🤔 The single `TracerSettings` instance is created once on app startup, and we shouldn't have another

tracer/test/test-applications/debugger/dependency-libs/Samples.Probes.TestRuns/Samples.Probes.TestRuns.csproj
```diff
@@ -6,6 +6,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Samples.Probes.External\Samples.Probes.External.csproj" />
+    <ProjectReference Include="..\..\..\..\..\src\Datadog.Trace.Manual\Datadog.Trace.Manual.csproj" />
```
andrewlock, line 9: I'm not sure, this _might_ cause issues in CI, but hopefully not 🤔

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -39,12 +40,15 @@ internal sealed class DebuggerManager
...
         private readonly TaskCompletionSource<bool> _processExit;
+        private string _serviceName;
         private volatile bool _isDebuggerEndpointAvailable;
         private int _initialized;
         private int _symDbInitialized;
         private volatile TaskCompletionSource<bool>? _diDebounceGate;
         private volatile DynamicInstrumentation? _dynamicInstrumentation;
         private int _diState; // 0 = disabled, 1 = initializing, 2 = initialized
+        private TracerSettings.SettingsManager? _subscribedSettingsManager;
+        private IDisposable? _tracerSettingsSubscription;
```
andrewlock, line 51: `DebuggerManager` lives for the lifetime of the app, right? If so this is fine, but if not, you should make sure to dispose this when `DebuggerManager` is disposed (it only really matters for tests so that we don't leak perf issues 🙂 )

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
```diff
@@ -191,6 +200,58 @@ private void OneTimeSetup(TracerSettings tracerSettings)
...
+            {
+                return;
+            }
+
+            ServiceName = GetServiceName(updatedMutable);
+
+            // Note: `SymbolsUploader` captures the service name on first use (symbol extraction/upload) and then keeps it fixed for its lifetime,
+            // to avoid mixing symbols across services. If the service name changes after that point, the correct behavior would be to stop and
+            // recreate the uploader, but that is expensive (symbol extraction/upload) and is intentionally deferred until we have an explicit
+            // requirement.
```
GreenMatan, line 252: Let's raise that in the team

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
dudikeleti: What about `ReplaceGlobalManager`? it seems that we subscribe to changes again to the new tracer in `TracerMAnagerFactory.CreateTracerManager` -> `tracer.Settings.Manager.SubscribeToChanges`

tracer/src/Datadog.Trace/Debugger/Upload/SymbolUploadApi.cs
dudikeleti: > we probably don't need to use a Func<string>

That is to delay the service name evaluation as much as possible, so we have the highest chance of using the updated value.

tracer/src/Datadog.Trace/Debugger/DebuggerManager.cs
andrewlock: So _it depends_ 😄 In theory, yes you're right, but `ReplaceGlobalManager` is only ever called by ci visibility (and would love to change that in the future) and I'm not even sure about that behaviour now (honestly, I forgot about it 😅). It should never be called in another situation.

PR #8055, opened 2026-01-14
[tracing] add MassTransit integration

chojomok:
## Summary of changes

  Adds automatic instrumentation for MassTransit 7.x and 8.x, a popular distributed application framework for .NET. The integration captures distributed traces across message
  producers and consumers, propagates trace context across transport boundaries (RabbitMQ, Amazon SQS, In-Memory), and supports both DiagnosticSource (MT7) and
  OpenTelemetry Activity (MT8) instrumentation patterns.

  ## Reason for change

  MassTransit is a widely-used message bus framework in the .NET ecosystem. This integration provides customers with end-to-end visibility into distributed message-driven
  architectures, allowing them to track message flow across services, identify performance bottlenecks, and correlate traces across synchronous HTTP and asynchronous messaging
  operations.

  ## Implementation details

  **MassTransit 7.x:**
  - Instruments via `DiagnosticSource` listeners (`MassTransit.ReceivePipe`, `MassTransit.ConsumeContext`)
  - Hooks into `ConsumeContext`, `SendContext`, and `PublishContext` for context propagation
  - Creates `RECEIVE` and `PROCESS` spans for incoming messages, `SEND` spans for outgoing messages

  **MassTransit 8.x:**
  - Leverages MassTransit's built-in OpenTelemetry `ActivitySource` instrumentation
  - Enhances Activity tags with Datadog-specific metadata
  - Injects/extracts trace context via message headers (`x-datadog-*`, `_datadog`)

  **Key technical decisions:**
  - Uses duck typing over reflection for ~30% performance improvement in header access
  - Transport-agnostic design handles RabbitMQ, SQS, and In-Memory
  - For MT8, enriches existing Activities rather than creating new spans to avoid duplication
  - Optimized Activity.Tags iteration (single `foreach` with `switch` vs multiple LINQ calls)

  ## Test coverage

  - `MassTransit7Tests.cs` - DiagnosticSource instrumentation across multiple transports
  - `MassTransit8Tests.cs` - OpenTelemetry Activity enhancement across multiple transports
  - Package version testing: MT7 (7.0.0 - 7.3.1), MT8 (8.0.3 - 8.5.8)
  - Sample applications for both versions with multi-transport scenarios, saga state machines, and error handling

  ## Other details

  **Compatibility:**
  - MassTransit 7.x: .NET Core 2.1+ (.NET 5, 6, 7, 8, 9, 10)
  - MassTransit 8.x: .NET Framework 4.8 / .NET Core 2.1+ (.NET 5, 6, 7, 8, 9, 10)
  - Tested with RabbitMQ 3.x, LocalStack (SQS), 

NEED TO TEST WITH Azure Service Bus !!!


tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/MassTransit7Tests.cs
```diff
@@ -0,0 +1,68 @@
...
+        using (await RunSampleAndWaitForExit(agent))
+        {
+            // Wait for spans to arrive - expecting send, receive, and process spans
+            const int minExpectedSpanCount = 12;
+            var spans = await agent.WaitForSpansAsync(minExpectedSpanCount, timeoutInMilliseconds: 30000);
+
+            using var s = new AssertionScope();
+            spans.Count.Should().BeGreaterOrEqualTo(minExpectedSpanCount);
+
+            // Filter to MassTransit spans - component tag should be "MassTransit"
```
chojomok: this is terrible - I should have checked first that we never actually checked the snapshot. 

I'll fix this in a bit

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/MassTransit7Tests.cs
```diff
@@ -0,0 +1,207 @@
...
+    public async Task SubmitsTraces(string packageVersion)
+    {
+        // Set environment variables for RabbitMQ and LocalStack (for SQS/SNS)
+        var rabbitHost = Environment.GetEnvironmentVariable("RABBITMQ_HOST") ?? "localhost";
+        var localStackEndpoint = Environment.GetEnvironmentVariable("AWS_SDK_HOST") ?? "localhost:4566";
+
+        SetEnvironmentVariable("RABBITMQ_HOST", rabbitHost);
+        SetEnvironmentVariable("LOCALSTACK_ENDPOINT", $"http://{localStackEndpoint}");
+        SetEnvironmentVariable("DD_TRACE_OTEL_ENABLED", "false");
+
```
chojomok, line 53: Do a check on env var MASSTRANSIT_INMEMORY_ONLY to see if we should skip the non-in memory tests

tracer/test/snapshots/MassTransit8TestsWindows.verified.txt
```diff
@@ -0,0 +1,850 @@
...
+    Service: Samples.MassTransit8,
+    Type: queue,
+    ParentId: Id_15,
+    Error: 1,
+    Tags: {
+      component: masstransit,
+      env: integration_tests,
+      error.msg: Handler failure test: Handler will fail,
+      error.stack: Scrubbed
+            error.type: System.InvalidOperationException,
```
chojomok: let's add a bit more info into
get the error.stack and remove all the new lines. 
change all the new lines into a space?

tracer/test/test-applications/integrations/Samples.MassTransit7/Samples.MassTransit7.csproj
```diff
@@ -0,0 +1,22 @@
...
+
+  <ItemGroup>
+    <PackageReference Include="MassTransit" Version="$(ApiVersion)" />
+    <PackageReference Include="MassTransit.RabbitMQ" Version="$(ApiVersion)" />
+    <PackageReference Include="MassTransit.AmazonSQS" Version="$(ApiVersion)" />
+    <PackageReference Include="MassTransit.Extensions.DependencyInjection" Version="$(ApiVersion)" />
+    <PackageReference Include="Automatonymous" Version="5.1.3" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="6.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="6.0.0" />
+  </ItemGroup>
```
chojomok, line 20: can we use shared files for some the masstransit sample apps and reference in the csproj?

chojomok: use compile include

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Shared/ContextPropagation.cs
```diff
@@ -85,12 +87,92 @@ public static void InjectHeadersIntoMessage(Tracer tracer, IContainsMessageAttri
             }
         }
 
+        /// <summary>
```
chojomok, line 90: When MassTransit uses AWS as a transport layer,  we need a way to propagate context from MassTransit to AWS/RabbitMQv.

Not tested yet for AzureServiceBus

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/SNS/AwsSnsHandlerCommon.cs
```diff
@@ -25,7 +25,18 @@ public static CallTargetState BeforePublish<TPublishRequest>(TPublishRequest req
         var requestProxy = request.DuckCast<IAmazonSNSRequestWithTopicArn>();
 
         var tracer = Tracer.Instance;
-        var scope = AwsSnsCommon.CreateScope(tracer, sendType.OperationName, SpanKinds.Producer, out var tags);
+
```
chojomok, line 28: For RabbitMQ when I didn't put the current active scope as a priority before checking the headers, it caused some tests to fail. 
Should I be doing the same here?

tracer/test/snapshots/MassTransit8Tests.pre_8_0_5.verified.txt
```diff
@@ -0,0 +1,1117 @@
+﻿[
```
chojomok, line 1: 8.0.4 is the latest version that works before there are issues on how the service name is created from the Activity 8.0.3 releases. 
Ideally - I just want to test support for 8.2.1 + just because there are less snapshots to keep track off and MassTransit v9 is coming out. V8 is about to be OLD (but probably still used)

PR #8060, opened 2026-01-14
fix(tests): make sure we include the codeowners tag on all tests

tonyredondo:
## Summary of changes

  - Add CODEOWNERS fallback resolution and source path normalization to keep matching consistent when SourceRoot differs from workspace or paths are absolute/Windows-style.
  - Expand CODEOWNERS test coverage with fallback scenarios and Windows separator cases.
  - Bump the DatadogTestLogger nuget package to 0.0.52
  - Bump time-it to 0.4.7

## Reason for change

  - CODEOWNERS matching can fail when source files are absolute (Windows) or when SourceRoot and workspace paths diverge; the updates make matching deterministic and cross-platform.
  - Try to fix some scenarios where we don't have CODEOWNERS in tests

## Implementation details

  - Track CodeOwnersRoot and add a fallback search that starts from the source file, then workspace root, stopping at git boundaries and capping search-start cache growth.
  - Normalize SourceFile tags via MakeRelativePathFromSourceRootWithFallback and gate matching with TryGetCodeOwnersRelativePath so only paths within the CODEOWNERS root are used.
  - Normalize path separators before matching in CodeOwners.Match.

## Test coverage

  - Added CodeOwnersFallbackTests covering fallback root resolution, workspace mismatch, and outside-root guards.
  - Extended CodeOwnersTests with Windows path separator cases.

### DIFF
--- a/tracer/Directory.Build.props
+++ b/tracer/Directory.Build.props
@@ -31,7 +31,7 @@
   <PropertyGroup>
     <SolutionDir Condition="'$(SolutionDir)'==''">$(MSBuildThisFileDirectory)..\</SolutionDir>
   </PropertyGroup>
-  
+
   <!-- OS Detection Properties -->
   <PropertyGroup>
     <IsWindows Condition="'$(OS)' == 'Windows_NT'">true</IsWindows>
@@ -59,5 +59,5 @@
       <PreprocessorDefinitions>%(PreprocessorDefinitions);MICROSOFT_WINDOWS_WINBASE_H_DEFINE_INTERLOCKED_CPLUSPLUS_OVERLOADS=1</PreprocessorDefinitions>
     </ClCompile>
   </ItemDefinitionGroup>
-  
+
 </Project>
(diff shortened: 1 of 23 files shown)


tracer/src/Datadog.Trace/Ci/CiEnvironment/CIEnvironmentValues.cs
```diff
@@ -20,10 +21,17 @@ namespace Datadog.Trace.Ci.CiEnvironment;
 
 internal abstract class CIEnvironmentValues
 {
+    private const int CodeOwnersSearchCacheLimit = 256;
     internal const string RepositoryUrlPattern = @"((http|git|ssh|http(s)|file|\/?)|(git@[\w\.\-]+))(:(\/\/)?)([\w\.@\:/\-~]+)(\.git)?(\/)?";
     protected static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(CIEnvironmentValues));
     private static readonly Lazy<CIEnvironmentValues> LazyInstance = new(Create);
     private static readonly Regex BranchOrTagsRegex = new(@"^refs\/heads\/tags\/(.*)|refs\/heads\/(.*)|refs\/tags\/(.*)|refs\/(.*)|origin\/tags\/(.*)|origin\/(.*)$", RegexOptions.Compiled);
+    private static readonly StringComparer CodeOwnersSearchComparer = Path.DirectorySeparatorChar == '\\'
```
lucaspimentel: Interesting OS check 😅

tonyredondo: haha, true, I've changed to use `FrameworkDescription.Instance.IsWindows()`

PR #8061, opened 2026-01-14
[DBM] Add container tags hash to queries (if enabled)

vandonr:
## Summary of changes

Add the ability to write the container tags hash to DBM queries + to the related span.
The goal is that DBM would then query the spans bearing that hash, and then use the container tags on this (those) spans(s)  to enrich the queries with it.
This is controlled by a setting that is disabled by default, and would be enabled if propagation mode is "service" or greater

see RFC: https://docs.google.com/document/d/15GtNOKGBCt6Dc-HsDNnMmCdZwhewFQx8yUlI9in5n3M
related PR in python: https://github.com/DataDog/dd-trace-py/pull/15293

## Reason for change

DBM and DSM propagate service context in outbound communications (SQL comments, message headers), but neither product has awareness of the container environment (e.g., `kube_cluster`, `namespace`, `pod_name`). Propagating full container tags is not feasible due to cardinality constraints (query cache invalidation in OracleDB/SQLServer, exponential pathway growth in DSM) and size limitations (64–128 bytes for DBM non-comment methods).

This is needed for the **service renaming initiative** (defining services based on container names) and **APM primary tags** (container-based dimensions like Kubernetes cluster).

The solution: the agent computes a hash of low-cardinality container tags and back-propagates it to the tracer, which includes it in outbound DBM/DSM communications. DBM then resolves the hash by correlating with APM spans that carry the same hash as a span tag.

## Implementation details

- Add `BaseHash` static class that computes an FNV-64 hash of `ProcessTags.SerializedTags` combined with the container tags hash from the agent, encoded as base64
- Read the container tags hash from the Datadog Agent via `DiscoveryService`, stored in `ContainerMetadata.ContainerTagsHash`
- `ContainerMetadata` converted from static to instance class (singleton via `ContainerMetadata.Instance`) to improve testability
- `DatabaseMonitoringPropagator` injects the base hash into SQL comments (as `ddsh`) when `DD_DBM_INJECT_SQL_BASEHASH` is true
- Add `_dd.dbm_container_tags_hash` span tag on `SqlTags` so DBM can correlate the hash back to the span's container tags
- New config key `DD_DBM_INJECT_SQL_BASEHASH` (disabled by default), intended to be enabled when DBM propagation mode is `service` or higher
- Add container ID header to `MinimalAgentHeaderHelper` for agent communication

## Test coverage

Adding a test in DbScopeFactoryTests.cs forced me to inject the value from pretty high, which I find a bit "dirty", but at least we don't have to rely on global static instance in tests.

### DIFF
--- a/tracer/src/Datadog.Trace/ProcessTags.cs
+++ b/tracer/src/Datadog.Trace/ProcessTags.cs
@@ -38,7 +38,10 @@ internal sealed class ProcessTags
     // two views on the same data
     public List<string> TagsList => field ??= GetTagsList(_serviceNameUserDefined, _autoServiceName);
 
-    public string SerializedTags => field ??= string.Join(",", TagsList); // don't forget to refresh the base hash on write if this value becomes mutable
+    public string SerializedTags
+    {
+        get => field ??= string.Join(",", TagsList); // don't forget to refresh the hash in ServiceRemappingHash on write if this value becomes mutable
+    }
 
     private static List<string> GetTagsList(bool serviceNameUserDefined, string autoServiceName)
     {
--- a/tracer/src/Datadog.Trace/ServiceRemappingHash.cs
+++ b/tracer/src/Datadog.Trace/ServiceRemappingHash.cs
@@ -68,13 +68,14 @@ internal sealed class ServiceRemappingHash
             hash = FnvHash64.GenerateHash(containerTagsHash, FnvHash64.Version.V1, hash);
         }
 
+        // ulong/UInt64 is 8 bytes
+        // Encoding to Base64 expands the number of bytes by 4 * ceil(n / 3) = 12 bytes
 #if NETCOREAPP3_1_OR_GREATER
         Span<byte> buf = stackalloc byte[12];
 #else
         // can't stackalloc into the vendored Span<T>
         var buf = new byte[12];
 #endif
-
         BinaryPrimitives.WriteUInt64LittleEndian(buf, hash); // write 8 bytes into a 12-byte buffer
         Base64.EncodeToUtf8InPlace(buf, dataLength: 8, out var bytesWritten);
 


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AdoNet/DbScopeFactory.cs
```diff
@@ -102,9 +103,14 @@ internal static class DbScopeFactory
                     }
                     else
                     {
-                        // PropagateDataViaComment (service) - this injects varius trace information as a comment in the query
+                        if (tracer.Settings.InjectSqlBasehash && !string.IsNullOrEmpty(baseHash))
+                        {
+                            tags.BaseHash = baseHash;
```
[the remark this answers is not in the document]
vandonr: hmm, yes, that's an interesting point, but I'm not sure we care, we only need one span with the hash to get the values, so we don't really need to tag all spans. I think in practice it works well like this.

tracer/src/Datadog.Trace/DatabaseMonitoring/DatabaseMonitoringPropagator.cs
```diff
@@ -104,6 +107,11 @@ propagationLevel is not (DbmPropagationLevel.Service or DbmPropagationLevel.Full
                 propagatorStringBuilder.Append(',').Append(SqlCommentVersion).Append("='").Append(Uri.EscapeDataString(versionTag)).Append('\'');
             }
 
+            if (!string.IsNullOrEmpty(baseHash))
+            {
+                propagatorStringBuilder.Append(',').Append(SqlCommentBaseHash).Append("='").Append(Uri.EscapeDataString(baseHash)).Append('\'');
+            }
```
bouwkast, line 112: If I understand the Python implementation correctly they appear to be recomputing the hash:
https://github.com/DataDog/dd-trace-py/blob/b5008005cd637a07e5c249f8d60ed07ee8328dc5/ddtrace/internal/process_tags/__init__.py#L92-L94

I _think_ based on the linked document that the method here of just passing in the has as given from the agent is the correct approach, but seeing that Python has been merged I'm wondering if the approach has been changed outside of the RFC? If not, I think that the Python implementation must not work.

From the RFC:

> The agent would compute a hash on low cardinality container tags that can be used to identify a workload

> The agent will then propagate that hash to the tracing libraries that can then use it to complete outbound communications with container information

Could we validate this?

vandonr:
I just realized I need to put process tags in there too

tracer/src/Datadog.Trace/DatabaseMonitoring/DatabaseMonitoringPropagator.cs
vandonr: the hash is recomputed here because it's _only_ injected from the instrumentation site, where we get it fresh from the global instance, so if the global instance changes, we'd pick up the latest one.
However, here I just implemented container tags, but I realized as you were reviewing that I need to add proces tags in there too (like in python), so there will be _some_ computation to do.

vandonr: basically, the agent computes a hash that's ready to use based on container tags, but we also need to pack process tags in that hash, which is why there is a re-hashing step needed in the tracer (then)

tracer/src/Datadog.Trace/Tagging/SqlTags.cs
```diff
@@ -14,23 +13,26 @@ internal partial class SqlTags : InstrumentationTags
...
-        [Tag(Trace.Tags.OutHost)]
+        [Tag(Tags.OutHost)]
         public string OutHost { get; set; }
 
-        [Tag(Trace.Tags.DbmTraceInjected)]
+        [Tag(Tags.DbmTraceInjected)]
         public string DbmTraceInjected { get; set; }
+
+        [Tag(Tags.BaseHash)]
+        public string BaseHash { get; set; }
```
bouwkast, line 35: Just noticing another discrepancy between .NET / Python

.NET is adding the hash to the `meta` as a `string`
Python is adding the hash to the `metrics` as an `int` (or whatever python uses 🤷 )

https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/propagation/_database_monitoring.py#L87-L88

I don't think the RFC clarifies exactly what is expected here though:

> A tag (_dd.propagated_hash) should be set on the current span when injection happens in order to correlate the container hash set on the query with the spans, and retrieve the container tags from there. The value of the tag should be the back propagated container tags hash, see [PR](https://github.com/DataDog/dd-trace-py/pull/15293)


Wait that comment on the RFC points back to the Python PR as the source of truth, so is it really that the Python PR is the source of truth here and not the RFC?

tracer/src/Datadog.Trace/DatabaseMonitoring/DatabaseMonitoringPropagator.cs
bouwkast: Okay 👍 

I also noticed we were putting it as a string compared to python, but also noticed that the RFC actually seems to point to that Python PR as being the source of truth so I think that answers both this comment and the one I just left [here](https://github.com/DataDog/dd-trace-dotnet/pull/8061#discussion_r2694747550)

tracer/src/Datadog.Trace/Tagging/SqlTags.cs
vandonr: yes it is really that. I agree that this is not great, but that's the world we live in. Even worse, the "real spec" used to be the java implem, but then as the python implem was written, it turned out that there was some holes in it, so now the reference is the python code 🙃 
I'll double check about the type of the tag, because it's important that we get all the digits, and I'm not 100% sure metric type tags do that.

tracer/src/Datadog.Trace/DatabaseMonitoring/DatabaseMonitoringPropagator.cs
vandonr: after some investigation and discussion ([here](https://dd.slack.com/archives/CC7Q58YRG/p1768819701733689)), we agreed that storing the hash in b64 was the best compromise, and python will align on that.
I amended the RFC to specify that the value should be in meta (as a tag), and the python code will be updated.

tracer/src/Datadog.Trace/BaseHash.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        get => Volatile.Read(ref field);
+        private set => Volatile.Write(ref field, value);
+    }
+
+        = Recompute(ProcessTags.SerializedTags, ContainerMetadata.Instance.ContainerTagsHash);
+
+    public static string Recompute(string processTags, string? containerTagsHash)
+    {
+        var hash = FnvHash64.GenerateHash(processTags, FnvHash64.Version.V1);
```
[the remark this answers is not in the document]
vandonr: there is no container ID in .net framework

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
```diff
@@ -0,0 +1,43 @@
...
+
+    public static string Recompute(string processTags, string? containerTagsHash)
+    {
+        var hash = FnvHash64.GenerateHash(processTags, FnvHash64.Version.V1);
+        if (containerTagsHash != null)
+        {
+            hash = FnvHash64.GenerateHash(containerTagsHash, FnvHash64.Version.V1, hash);
+        }
+
+        return B64Value = Convert.ToBase64String(BitConverter.GetBytes(hash));
```
dubloom: My understanding is if process tags are not present, you don't want to compute anything.

Additionally, shouldn't we compute the hash for containerTags hash and process tags at the same time ? 
That is done in Python and I'd argue this is not the same as what we are doing here

tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -23,7 +23,7 @@ internal static class ProcessTags
     // two views on the same data
     public static IReadOnlyCollection<string> TagsList => field ??= GetTagsList();
 
-    public static string SerializedTags => field ??= string.Join(",", TagsList);
+    public static string SerializedTags => field ??= string.Join(",", TagsList); // don't forget to refresh the base hash on write if this value becomes mutable
```
dubloom: Is this comment has any effect ? Because I don't feel we are doing something. Is it because right now this is immutable ?

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
```diff
@@ -68,7 +68,11 @@ public ContainerMetadata(string? containerId, string? entityId)
         public string? ContainerTagsHash
         {
             get => Volatile.Read(ref field);
-            set => Volatile.Write(ref field, value);
+            set
+            {
+                Volatile.Write(ref field, value);
+                BaseHash.Recompute(ProcessTags.SerializedTags, value);
+            }
```
dubloom: It means whenever we modify containerTagsHash, we recompute the basehash right ?

tracer/src/Datadog.Trace/DatabaseMonitoring/DatabaseMonitoringPropagator.cs
```diff
@@ -103,6 +106,11 @@ propagationLevel is not (DbmPropagationLevel.Service or DbmPropagationLevel.Full
...
+
```
dubloom, line 113: We are not checking the config are we ?

dubloom: actually, we are doing it but by setting base hash to null if the config is not activated, I think this is not super clear and it can be missed. It would probably be clearer to check the config here if possible

tracer/src/Datadog.Trace/ProcessTags.cs
vandonr: yes it is immutable for the moment (it's computed once and then never changes)

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
vandonr: about your first point, I'm not sure it's what you meant, but I realized that I didn't check that process tags were enabled when computing this. 
I think I'll keep this as is, where process tags is never null, and just add the check on the setting on usage.

About your second point, computing the hash in two steps like this allows us to avoid creating an extra string. We basically compute the second step picking up where we left off on the previous one.

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
vandonr: yes correct

tracer/src/Datadog.Trace/DatabaseMonitoring/DatabaseMonitoringPropagator.cs
vandonr: hmm, I see you point, I did it like that because we don't have access to the config object here, the necessary settings are passed directly as parameters. I'll see if I can make it clearer, especially since it duplicates the check from https://github.com/DataDog/dd-trace-dotnet/blob/a75fa609f913d5231fccffed1709f7c98d76065f/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AdoNet/DbScopeFactory.cs#L106

vandonr: ok, I did a change that moves a tiny bit in the right direction, changing the nullability is I think helping the trained dotnet eye read this correctly

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
```diff
@@ -68,7 +68,11 @@ public ContainerMetadata(string? containerId, string? entityId)
         public string? ContainerTagsHash
         {
             get => Volatile.Read(ref field);
-            set => Volatile.Write(ref field, value);
+            set
+            {
+                Volatile.Write(ref field, value);
+                BaseHash.Recompute(ProcessTags.SerializedTags, value);
```
bouwkast: Should this be in `ContainerMetadata.Framework`?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AdoNet/CommandExecuteNonQueryAsyncIntegration.cs
```diff
@@ -8,6 +8,7 @@
 using System.Data;
 using System.Threading;
 using Datadog.Trace.ClrProfiler.CallTarget;
+using Datadog.Trace.PlatformHelpers;
```
bouwkast: nitpick: can follow up on this in a later PR but VS is saying these are unused

tracer/src/Datadog.Trace/BaseHash.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// <summary>
+    /// Gets the base64 representation of the hash
+    /// </summary>
+    public static string B64Value
+    {
+        get => Volatile.Read(ref field);
+        private set => Volatile.Write(ref field, value);
+    }
+
+        = Recompute(ProcessTags.SerializedTags, ContainerMetadata.Instance.ContainerTagsHash);
```
bouwkast: I don't think we really have any usages of the actual [field](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/field) keyword in the project and it makes this file look _strange_ 

I think having an actual backing private variable would make this look more consistent with the rest of the project

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
```diff
@@ -0,0 +1,43 @@
...
+/// </summary>
+internal static class BaseHash
+{
+    /// <summary>
+    /// Gets the base64 representation of the hash
+    /// </summary>
+    public static string B64Value
+    {
+        get => Volatile.Read(ref field);
+        private set => Volatile.Write(ref field, value);
```
bouwkast: nitpick

I'd almost rather we didn't have a setter at all and just had calls for Recompute and internally in there write the value as at the moment I would expect it to be a basic setter and not something that does some work in the background as well.

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+        = Recompute(ProcessTags.SerializedTags, ContainerMetadata.Instance.ContainerTagsHash);
+
+    public static string Recompute(string processTags, string? containerTagsHash)
+    {
+        var hash = FnvHash64.GenerateHash(processTags, FnvHash64.Version.V1);
```
bouwkast, line 65: For the future: I wonder if we need to benchmark FnvHash64 to see if we need to optimize it (unsure how frequently this path is going to be hit)

tracer/test/Datadog.Trace.ClrProfiler.Managed.Tests/DbScopeFactoryTests.cs
```diff
@@ -151,6 +151,35 @@ public async Task CreateDbCommandScope_IgnoresReplacementServiceNameWhenNotProvi
...
+        [Theory]
+        [MemberData(nameof(GetEnabledDbmData))]
+        public async Task CreateDbCommandScope_HasBaseHashWhenConfigured(Type commandType, string dbmMode, bool hashPropagationEnabled)
+        {
+            var command = (IDbCommand)Activator.CreateInstance(commandType)!;
+            command.CommandText = DbmCommandText;
+
+            var tracerSettings = TracerSettings.Create(new Dictionary<string, object>
+            {
+                { ConfigurationKeys.DbmInjectSqlBasehash, hashPropagationEnabled.ToString() },
```
bouwkast, line 165: Do we need to set `PropagateProcessTags` as well?

tracer/src/Datadog.Trace/BaseHash.cs
lucaspimentel: fwiw, I'm ok with us using the (newish, added in C# 14) `field` keyword going forwards. I used it for the first time in one of my PRs recently, I can't remember where. Gotta embrace all the new C# features!

tracer/test/Datadog.Trace.TestHelpers/TransportHelpers/TestApiResponse.cs
```diff
@@ -23,6 +23,7 @@ public TestApiResponse(int statusCode, string body, string contentType, Dictiona
         _body = body;
         _headers = headers ?? new();
         ContentTypeHeader = contentType;
+        _headers = headers ?? new Dictionary<string, string>();
```
lucaspimentel: `_headers` is already initialized in line 24 above.

tracer/src/Datadog.Trace/Tags.cs
```diff
@@ -791,6 +791,9 @@ public static partial class Tags
         /// </summary>
         internal const string DbmTraceInjected = "_dd.dbm_trace_injected";
 
+        /// <summary> contains a hash of container tags (for now), to be matched with the hash injected in queries, so that we can retrieve the corresponding values from the span</summary>
+        internal const string BaseHash = "_dd.propagated_hash";
```
lucaspimentel, line 800: `propagated_hash` is _very_ ambiguous. A hash of what? What does it represent and what is it used for? If this specific to DBM, should it include "dbm" in the name or something?

Also, "propagated" tags are already a concept in the tracers, so this name could be confused with that.

tracer/src/Datadog.Trace/BaseHash.cs
```diff
@@ -0,0 +1,43 @@
...
+using Datadog.Trace.Util;
+
+namespace Datadog.Trace;
+
+/// <summary>
+/// This class is a container for the "base hash", a hash of process tags and container tags.
+/// Used by DBM to retrieve all the tag values from the spans, from just a single parameter (this hash),
+/// Used by DSM in the pathway to identify different sources that could have been service-remapped
+/// </summary>
+internal static class BaseHash
```
lucaspimentel: If this is "a hash of process tags and container tags" and is only used by DBM and DSM, we should improve the type name to be more specific (what's a "base" hash and how is that different from other hashes) and move it into a more narrower namespace than `Datadog.Trace`.

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
lucaspimentel: Yeah, having a static property which is just a simple get/set is basically implementing a global variable, which is not idiomatic C#.

lucaspimentel: Static state is troublesome (common cause of flaky tests, for example) and it encourages tight coupling because you can access `BaseHash.B64Value` from _anywhere_ instead of getting it injected only where needed. It's a pattern we try to avoid is possible.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AdoNet/CommandExecuteNonQueryAsyncIntegration.cs
lucaspimentel: in every `CommandExecute*Integration.cs`

tracer/src/Datadog.Trace/BaseHash.cs
lucaspimentel: [formatting] This initializer looks very weird hanging out there by itself 😅 

but more importantly: do we really want to call `Recompute()` in the static constructor? Isn't that a bit too eager?

tracer/test/Datadog.Trace.ClrProfiler.Managed.Tests/DbScopeFactoryTests.cs
```diff
@@ -151,6 +151,35 @@ public async Task CreateDbCommandScope_IgnoresReplacementServiceNameWhenNotProvi
             Assert.NotEqual("my-custom-type", scope.Span.ServiceName);
         }
 
+        [Theory]
+        [MemberData(nameof(GetEnabledDbmData))]
+        public async Task CreateDbCommandScope_HasBaseHashWhenConfigured(Type commandType, string dbmMode, bool hashPropagationEnabled)
```
lucaspimentel, line 156: [nit] The third parameter is defined as `storedProcInject` in `GetEnabledDbmData` (line 65), but here it's used as `hashPropagationEnabled`. It technically works (you just want `true`/`false`), but can be confusing. Maybe rename `storedProcInject` to something more generic like `enabled`?

tracer/src/Datadog.Trace/Tags.cs
lucaspimentel, line 800: According to 🤖 , based on the `dd-trace-py` PR you linked in the PR description, it seems like Python is storing this as a `uint64` numeric attribute in the `metrics` map, but in .NET we're using a base64 string attribute in `meta`?

lucaspimentel: Similar issue in the `ddsh` SQL comment:
- Python: injects `base_hash` which is the raw FNV64 integer
  `ddsh='1234567890'`
- .NET: injects `BaseHash.B64Value` which is `Convert.ToBase64String(BitConverter.GetBytes(hash))`
  `ddsh='AAAAAAAAAAA='`

tracer/src/Datadog.Trace/BaseHash.cs
lucaspimentel: 🤖 Comparing to the `dd-trace-py` PR you linked in the PR description:

`BaseHash.B64Value` is never empty/null when process tags exist

- .NET: `BaseHash` eagerly computes on static init from `ProcessTags.SerializedTags` even before container tags arrive. So `B64Value` is always a non-empty string (hash of just process tags). The guard `!string.IsNullOrEmpty(baseHash)` will always pass.
- Python: `base_hash` starts as `None` and only gets set when `compute_base_hash(container_tags_hash)` is called. So the hash is only injected after the agent responds with the container tags hash.

This means .NET will inject a partial hash (process tags only, no container tags) on startup, then silently change it after the agent responds. Python waits until container tags are available.

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
vandonr: Hmm, I don't think so, because in NetFramework, we never get container tags (because it never runs in a container), so actually what we do in the set there shouldn't matter because it's dead code... For what it's worth, I could even replace the `Write` with a warning log that this code shouldn't be called. Or like, send us some error telemetry, idk if I can do that directly without throwing an exception

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AdoNet/CommandExecuteNonQueryAsyncIntegration.cs
vandonr: I'll do it here

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
vandonr: I think it's used a lot more by DSM, but it looks relatively optimized already, on account that :
 - it was written by Andrew
 - it uses `Span` (not something you do when you don't have performance in mind)
 - in has a `SkipLocalsInit` annotation, again showing there was some thoughts put into perf

tracer/test/Datadog.Trace.ClrProfiler.Managed.Tests/DbScopeFactoryTests.cs
vandonr: yes indeed

tracer/src/Datadog.Trace/Tags.cs
vandonr: I mean, I hear you, but that ship has sailed, it's how it's specified in [the RFC](https://docs.google.com/document/d/15GtNOKGBCt6Dc-HsDNnMmCdZwhewFQx8yUlI9in5n3M/edit?pli=1&tab=t.0#heading=h.bw16yqis4bx6), and it's already implemented like this in the backend and in several other tracers :/

vandonr: I think this was changed in a later PR in python, I'll double check

vandonr: yes, so about metric, it was changed to a tag later:
https://github.com/DataDog/dd-trace-py/blob/b892f2fbdce342b949f09915a0868bf1a0195b26/ddtrace/propagation/_database_monitoring.py#L87

It still writes an int (as a string), but we discussed that since we're writing a string anyway, writing it as b64 was more efficient. This doesn't need to be consistent across tracers, it only needs to be consistent between the span and the SQL comment, since it's used to match one with the other.

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
vandonr: but mom, all my friends are using global static

tracer/src/Datadog.Trace/BaseHash.cs
vandonr: I think my approach makes more sense, because in the case where we're not running inside a container, we'd still want to have the process tags in the hash, even though the container tag hash would never be set. I'll check.

vandonr: Also, yes the formating is weird, but that's how the IDE auto-formatted it, so I went with it 🤷 

Recompute in the init is actually just "compute", but I don't think it's too eager, given the fact that we want process tags from the get go as I said above. Do you have any performance concerns about a static init ? I could eventually replace it with a lazy if that bothers you, but it'd _always_ be computed, and pretty early in the app init

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
vandonr: more seriously, I don't think we can go around using a global static given the fact it's called from an instrumentation. You cannot really inject anything in `OnMethodBegin`, right ? I then propagate it from there, and it's true that there is nothing preventing using it in other places in a bad way, but I don't see a way to enforce it in the code structure.
And since the read is static, I have the write it in a static way too.

tracer/src/Datadog.Trace/BaseHash.cs
vandonr: DBM and DSM don't really have a place they share in the code, do you think `PlatformHelpers` would be a better place for this ?
Also, I took the name from the java tracer. WDYT of "ServiceRemappingEnablementHash" ? It's a bit verbose, but describes exactly the purpose of this hash (to enable service remapping where it's used)

tracer/src/Datadog.Trace/BaseHash.cs
vandonr: python change prompted by your comment: https://github.com/DataDog/dd-trace-py/pull/16946

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
lucaspimentel: Consider adding a comment to `ContainerMetadata.NetFramework.cs` explaining why we don't call `BaseHash.Recompute()` there.

tracer/src/Datadog.Trace/BaseHash.cs
andrewlock: Yeah, to clarify, I think `field` is ok, but having `= Recompute` 2 lines later than the property _isn't_ ok IMO 😉

tracer/src/Datadog.Trace/Tags.cs
lucaspimentel: > This doesn't need to be consistent across tracers, it only needs to be consistent between the span and the SQL comment, since it's used to match one with the other.

That is very helpful, thanks!

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
andrewlock: In general, you should not use global statics. Anything that needs to be "global" should hang off the `TracerManager` instanc

tracer/src/Datadog.Trace/BaseHash.cs
lucaspimentel: > DBM and DSM don't really have a place they share in the code

Good point. I guess it's fine here for now. We can always move it later.

> WDYT of "ServiceRemappingEnablementHash"?

`ServiceRemappingHash` might be enough without being too verbose.

tracer/src/Datadog.Trace/BaseHash.cs
andrewlock: > python change prompted by your comment: DataDog/dd-trace-py#16946

I thought [the python implementation was the source of truth](https://github.com/DataDog/dd-trace-dotnet/pull/8061/changes#r2704771342) :trollface:

lucaspimentel: > I could eventually replace it with a lazy if that bothers you, but it'd always be computed, and pretty early in the app init

It's not a good pattern, imo. It makes the code harder to reason about and harder to debug when things go wrong. For example, one of the gotchas is that the the static init happens the first time the `BaseHash` _type_ is referenced anywhere, not when any of its _members_ is accessed. And if `Recompute()` ever throws an exception, it will throw in `BaseHash`'s static ctor, which can be tricky to debug and hard to handle since we can't catch it there.

I think it's safer to initialize it lazily in the property getter. Ignoring thread-safety, it could be as simple as
```
get { return field ??= Recompute(...); }
```
or using a `Lazy<T>` if you're into that kinda thing :)

vandonr: ok, I can force the format, but we should consider editing the rules to produce something we like more, because it's annoying to maintain a format that is auto-changed by the IDE

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
vandonr: I actually ported the changes to ContainerMetadata.NetFramework.cs that I did in https://github.com/DataDog/dd-trace-dotnet/pull/8163, where I enforce the value to be null. I think it is pretty self explanatory then.

tracer/src/Datadog.Trace/BaseHash.cs
lucaspimentel: > it's annoying to maintain a format that is auto-changed by the IDE

💯 

Unfortunately many formatting rules are not supported in `.editorconfig` where they are easy to share with the team across different IDEs, so they tend to be personal IDE settings that don't get committed to git.

For this particular case, I don't think any IDE would format it this way by default, so maybe it got into your personal settings accidentally? We could argue whether the assignment should be in a separate line or not, and I wouldn't mind it either way. It's that blank line between the property and the assignment that is _really_ weird. 😅

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
vandonr: I have something for y'all

tracer/src/Datadog.Trace/Ci/TestOptimization.cs
```diff
@@ -250,6 +250,7 @@ public void Initialize()
             getDiscoveryServiceFunc: static s => DiscoveryService.CreateUnmanaged(
                 s.TracerSettings.Manager.InitialExporterSettings,
                 ContainerMetadata.Instance,
+                new ServiceRemappingHash(null),
```
vandonr, line 253: this one I'm not 100% sure, but since it's only used for DBM for now, I don't think it'd play any role in that code path, so it should be safe to hardcode a disabled instance

tracer/src/Datadog.Trace/DatabaseMonitoring/DatabaseMonitoringPropagator.cs
```diff
@@ -28,6 +29,7 @@ internal static class DatabaseMonitoringPropagator
         private const string SqlCommentOuthost = "ddh";
         private const string SqlCommentVersion = "ddpv";
         private const string SqlCommentEnv = "dde";
+        private const string SqlCommentBaseHash = "ddsh";
```
lucaspimentel, line 32: The PR description says
>  injects the base hash into SQL comments (as `ddch`)

I couldn't find either one in the RFC, but the `dd-trace-py` PR uses `ddsh`, like this one. Is that a typo in the PR description?

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
```diff
@@ -0,0 +1,78 @@
...
+
+    private static string Compute(string processTags, string? containerTagsHash)
+    {
+        var hash = FnvHash64.GenerateHash(processTags, FnvHash64.Version.V1);
+        if (containerTagsHash != null)
+        {
+            hash = FnvHash64.GenerateHash(containerTagsHash, FnvHash64.Version.V1, hash);
+        }
+
+        var b64 = Convert.ToBase64String(BitConverter.GetBytes(hash));
```
lucaspimentel: `BitConverter.GetBytes()` will return different values in little-endian vs big-endian architectures. If this is relevant (I can't find endianness in the RFC), consider using `BinaryPrimitives.WriteInt64LittleEndian()` or the `BigEndian` version.

If not, then a comment here saying so would help ("don't care about endianness because...").

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /// <summary>
+    /// Gets the base64 representation of the hash
+    /// </summary>
+    public string? B64Value
+    {
+        get;
+        private set;
+    }
```
lucaspimentel, line 51: These properties [used to have](https://github.com/DataDog/dd-trace-dotnet/blob/b70e6f22def717fa3dd9a60a71b302adce136719/tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs#L68-L72) `Volatile.Read()`/`Volatile.Write()` and we should probably keep that since they are written from a background thread in `DiscoveryService` and read in the hot path when creating spans.

Furthermore, `UpdateContainerTagsHash` updates both values non-atomically, so a reader could see a stale `B64Value` with a new `ContainerTagsHash`. If consistency between the two is important, consider using a lock to read/write both values, or using immutable copies.

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
lucaspimentel: This code is allocating:

- `byte[]` in `BitConverter.GetBytes()`
- `char[]` for the parameter in `TrimEnd(params char[])` in .NET Framework (Newer runtimes have a `TrimEnd(char)` overload)
- `string` in `TrimEnd()` if it modifies the string
- more `string` instance for each `Replace()` if they modify the string

Good news! We have "vendored" versions of `BinaryPrimitives` and `Base64`, so we can avoid `BitConverter.GetBytes()` and `Convert.ToBase64String()`, and then trimming and replacing 1:1 chars can be done in-place, so this code should work in all TFMs:

```csharp
#if NETCOREAPP3_1_OR_GREATER
        Span<byte> buf = stackalloc byte[12];
#else
        // can't stackalloc into the vendored Span<T>
        var buf = new byte[12];
#endif

        BinaryPrimitives.WriteUInt64LittleEndian(buf, hash); // write 8 bytes into a 12-byte buffer
        Base64.EncodeToUtf8InPlace(buf, 8, out int bytesWritten);

        while (bytesWritten > 0 && buf[bytesWritten - 1] == (byte)'=')
        {
            bytesWritten--;
        }

        for (int i = 0; i < bytesWritten; i++)
        {
            if (buf[i] == (byte)'+')
            {
                buf[i] = (byte)'-';
            }
            else if (buf[i] == (byte)'/')
            {
                buf[i] = (byte)'_';
            }
        }

#if NETCOREAPP3_1_OR_GREATER
        return Encoding.ASCII.GetString(buf[..bytesWritten]);
#else
        // can't use Range
        return Encoding.ASCII.GetString(buf, 0, bytesWritten);
#endif
```
This has zero heap allocations on `NETCOREAPP3_1_OR_GREATER`, or only the `byte[12]` otherwise (aside from the final `string` we need to return in both cases which is unavoidable).

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
```diff
@@ -0,0 +1,78 @@
+// <copyright file="ServiceRemappingHash.cs" company="Datadog">
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2 License.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
+// </copyright>
+
+#nullable enable
+
+using System;
+using System.Threading;
+using Datadog.Trace.PlatformHelpers;
```
lucaspimentel: Not used.
```suggestion
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AdoNet/DbScopeFactory.cs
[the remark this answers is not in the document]
lucaspimentel: Consider adding a comment explaining this.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AdoNet/DbScopeFactory.cs
```diff
@@ -259,6 +265,7 @@ static Cache()
             public static Scope? CreateDbCommandScope(Tracer tracer, IDbCommand command)
             {
                 var commandType = command.GetType();
+                var baseHash = tracer.TracerManager.ServiceRemappingHash?.B64Value;
```
lucaspimentel: Should we guard this behind the setting?

```csharp
var baseHash = tracer.Settings.DbmInjectSqlBasehash ?
                   tracer.TracerManager.ServiceRemappingHash?.B64Value :
                   null;
```

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
```diff
@@ -0,0 +1,78 @@
...
+    {
+        ContainerTagsHash = containerTagsHash;
+
+        if (_processTags != null)
+        {
+            B64Value = Compute(_processTags.SerializedTags, containerTagsHash);
+        }
+    }
+
+    private static string Compute(string processTags, string? containerTagsHash)
```
lucaspimentel, line 63: While working on the "less-allocatey" code below, I noticed there are no unit tests for this method.

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.NetFramework.cs
```diff
@@ -32,20 +29,6 @@ public ContainerMetadata(string containerId, string entityId)
...
-    // always null in this implementation
-    public string? ContainerTagsHash
-    {
-        get => null;
-        set
-        {
-            if (!_warnedOnSet)
-            {
-                _warnedOnSet = true;
-                Log.Error("The code is trying to set the value '{Value}' to {Prop}, but this has no effect in .NET Framework.", value, nameof(ContainerTagsHash));
```
lucaspimentel, line 44: This log is now gone in the new version. Intentional?

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
lucaspimentel: Actually, .NET doesn't run on any big-endian architectures anymore. But I remember having an issue with this when implementing `HexConverter`/`HexString`. Hmm 🤔 

edit: Ahh, yes. Hex strings are conventionally big-endian (most significant byte first), so I had to make sure we converted. That doesn't apply here. Resolving this thread!

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
```diff
@@ -0,0 +1,78 @@
...
+    public string? ContainerTagsHash
+    {
+        get;
+        private set;
+    }
+
+    /// <summary>
+    /// Gets the base64 representation of the hash
+    /// </summary>
+    public string? B64Value
```
lucaspimentel: [Naming nit] The .NET naming conventions would use `Base64Value`, here, or simply `Base64`. No need to abbreviate "Base" to "B".

lucaspimentel:
related: https://github.com/DataDog/dd-trace-dotnet/pull/8363

tracer/src/Datadog.Trace/DatabaseMonitoring/DatabaseMonitoringPropagator.cs
vandonr: yes, I'll fix it

vandonr: You'll never guess who wrote that typo

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
vandonr: Also, what's important here is that we write the same thing to the span and the SQL comment, so conforming to a specific RFC-defined format is not critical

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
vandonr: Ah yes. TBH I have not fully understood when Volatile is needed versus regular get/set 😬. I guessed it was linked to it being global static, but apparently not. I'll use Volatile. 
Having consistency between B64 and container hash in this object is not critical.

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
vandonr: thanks for the verbose but efficient implem :p

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.NetFramework.cs
vandonr: yes 🤷 
I mean, I could add an #if NETFRAMEWORK in the `set` body, but I'm not sure it's really worth the extra code...

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AdoNet/DbScopeFactory.cs
vandonr: it's checked further down the execution path, but checking it here could make the logic simpler indeed

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    public void UpdateContainerTagsHash(string containerTagsHash)
+    {
+        ContainerTagsHash = containerTagsHash;
+
+        if (_serializedProcessTags != null)
+        {
+            Base64Value = Compute(_serializedProcessTags, containerTagsHash);
+        }
```
andrewlock, line 60: So just to confirm, we only have a `Base64Value` _if_ we have both process tags _and_ a container hash. And we don't care about the race condition/consistency here? e.g. for outside observers it's possible for there to be a `ContainerTagsHash` but no `Base64Value`, and vice versa. As long as we're fine with those limitations, this is fine. Otherwise, we can create a "holder" type to hold both values, and you can swap that atomically.

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    private static string Compute(string processTags, string? containerTagsHash)
+    {
+        var hash = FnvHash64.GenerateHash(processTags, FnvHash64.Version.V1);
```
andrewlock, line 65: Do we definitely want to use `v1` here, not `V1A`? IIRC, most places use V1A? 🤔

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+#if NETCOREAPP3_1_OR_GREATER
+        Span<byte> buf = stackalloc byte[12];
+#else
+        // can't stackalloc into the vendored Span<T>
+        var buf = new byte[12];
+#endif
+
+        BinaryPrimitives.WriteUInt64LittleEndian(buf, hash); // write 8 bytes into a 12-byte buffer
```
andrewlock, line 79: Just a suggestion:
```suggestion
        // UInt64 writes 8 bytes
        // Encoding to Base64 expands the number of bytes by 4 * ceil(n / 3) = 12 bytes
        // The output bytes are ASCII, so encoding to UTF-8 keeps it at 8 bytes
        BinaryPrimitives.WriteUInt64LittleEndian(buf, hash); // write 8 bytes into a 12-byte buffer
```

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
```diff
@@ -0,0 +1,107 @@
...
+        Base64.EncodeToUtf8InPlace(buf, dataLength: 8, out var bytesWritten);
```
andrewlock, line 80: _Technically_ we should check the return value for correct encoding - I'm kinda surprised the new analyzers don't flag it. But also, this _is_ fine, and we can make other assumptions later, so big meh 🤷‍♂️

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
```diff
@@ -0,0 +1,107 @@
...
+
+        // no padding
+        while (bytesWritten > 0 && buf[bytesWritten - 1] == (byte)'=')
+        {
+            bytesWritten--;
+        }
```
andrewlock: Technically we know that there will _always_ be a single `=` character (i.e. we will have 11 bytes of "usable" data, so we could optimize this to just always chop a single value 🤷‍♂️ 
```suggestion
        // Base64 always pads input to multiples of 3, so we know that we will
        // always have 11 bytes of "real" data, and one of padding `=`.
        bytesWritten--;
```

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
```diff
@@ -0,0 +1,107 @@
...
+                buf[i] = (byte)'-';
+            }
+            else if (buf[i] == (byte)'/')
+            {
+                buf[i] = (byte)'_';
+            }
+        }
+
+#if NETCOREAPP3_1_OR_GREATER
+        return Encoding.ASCII.GetString(buf[..bytesWritten]);
```
andrewlock: nit: Generally we avoid range anyway (for tiny perf reasons)

```suggestion
        return Encoding.ASCII.GetString(buf.Slice(0, bytesWritten));
```

tracer/src/Datadog.Trace/Agent/DiscoveryService/DiscoveryService.cs
```diff
@@ -310,7 +314,7 @@ private async Task ProcessDiscoveryResponse(IApiResponse response)
             var containerTagsHash = response.GetHeader(AgentHttpHeaderNames.ContainerTagsHash);
             if (containerTagsHash != null)
             {
-                _containerMetadata.ContainerTagsHash = containerTagsHash;
+                _serviceRemappingHash.UpdateContainerTagsHash(containerTagsHash);
```
andrewlock, line 317: I'm a little confused about how this interacts with the `ContainerMetadata`? 🤔  

As I understand it `ContainerMetadata` and "ContainerTagsHash" are _essentially_ two completely different concepts at this point? `ContainerMetadata` is stuff we calculate, the container tags hash is something entirely agent-provided, right?

(This confusion is all on me, mostly due to you having to redo things repeatedly based on my gripes 🙈)

tracer/src/Datadog.Trace/Agent/DiscoveryService/DiscoveryService.cs
```diff
@@ -429,7 +433,7 @@ private async Task ProcessDiscoveryResponse(IApiResponse response)
                 eventPlatformProxyEndpoint: eventPlatformProxyEndpoint,
                 telemetryProxyEndpoint: telemetryProxyEndpoint,
                 tracerFlareEndpoint: tracerFlareEndpoint,
-                containerTagsHash: _containerMetadata.ContainerTagsHash, // either the value just received, or the one we stored before (prevents overriding with null)
+                containerTagsHash: _serviceRemappingHash.ContainerTagsHash, // either the value just received, or the one we stored before (prevents overriding with null)
```
andrewlock, line 436: Is that correct though? 🤔 If the agent starts sending us `null`, is that not because something changed about the instance? Should we be second guessing the values we're sent here? (I haven't checked if the RFC spells this out)

tracer/src/Datadog.Trace/Ci/TestOptimization.cs
andrewlock: I agree, this seems fine to me, I can't imagine we support DBM or DSM with CI Visibility today 🤔

tracer/src/Datadog.Trace/Configuration/supported-configurations.yaml
```diff
@@ -533,6 +533,15 @@ supportedConfigurations:
       Configuration key for setting DBM propagation mode
       Default value is disabled, expected values are either: disabled, service or full
       <seealso cref="Datadog.Trace.Configuration.TracerSettings.DbmPropagationMode"/>
+  DD_DBM_INJECT_SQL_BASEHASH:
```
andrewlock, line 536: _Sigh_ this configuration key should end with `_ENABLED` so it should be something like `DD_DBM_SQL_BASEHASH_INJECTION_ENABLED`. If this ship hasn't already sailed in other languages, can we please fix it 🙏 🥺

tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -38,7 +38,7 @@ public ProcessTags(bool serviceNameUserDefined, string autoServiceName)
     // two views on the same data
     public List<string> TagsList => field ??= GetTagsList(_serviceNameUserDefined, _autoServiceName);
 
-    public string SerializedTags => field ??= string.Join(",", TagsList);
+    public string SerializedTags => field ??= string.Join(",", TagsList); // don't forget to refresh the base hash on write if this value becomes mutable
```
andrewlock: Thanks for this comment!

tracer/test/Datadog.Trace.Tests/ServiceRemappingHashTests.cs
```diff
@@ -0,0 +1,68 @@
...
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
+// </copyright>
+
+using System.Reflection;
+using FluentAssertions;
+using Xunit;
+
+namespace Datadog.Trace.Tests;
+
+public class ServiceRemappingHashTests
```
andrewlock, line 12: Do we have any example values from the RFC that we can use to validate the _actual_ expected hash given a set of process tags and container tags hash? Would be useful for validating that the calculation is correct across languages

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
vandonr: yes there can be (for a brief period) a ContainerTagsHash but a stale Base64Value, it's fine, no precise timing needed here.

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
vandonr: [python](https://github.com/DataDog/dd-trace-py/blob/8ff3ccb99c8457b1c2e7ee71ace472d16930844b/ddtrace/internal/process_tags/__init__.py#L109) and [Java](https://github.com/DataDog/dd-trace-java/blob/1a55ea4827be8053936267cffeacb4d198448a1a/internal-api/src/main/java/datadog/trace/api/BaseHash.java#L44-L50) both use V1 and not V1A, but what really matters is consistency _within a tracer_ between the sql comment and the tag on the span, so either would be fine

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
vandonr: 👍  except I don't understand the last line `The output bytes are ASCII, so encoding to UTF-8 keeps it at 8 bytes` 🤔 I understand as if we'd expect the output to be 8 bytes ?

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
vandonr: hmm, yeah, but the return values are either Done or DestinationTooSmall. DestinationTooSmall should never happen because we computed the size pretty deterministically...
I can like, throw an exception if this happens so that it's catched in tests if this ever happens because... we put more data in there

tracer/src/Datadog.Trace/ProcessTags.cs
vandonr: I just realized the "base hash" changed name during this PR life, so I'll update it so it points to a more obvious location :)

tracer/test/Datadog.Trace.Tests/ServiceRemappingHashTests.cs
vandonr: no expected hash, as I said in an other reply above, we just need to be consistent with ourselves

tracer/src/Datadog.Trace/Agent/DiscoveryService/DiscoveryService.cs
vandonr: yes, correct. Container Metadata is something we can know from the tracer, and it's computed once and constant.
There are some container tags that we cannot know from inside the app, so the agent collects them and enriches the spans with them. What we do here is that we ask the agent for a key (hash) corresponding to our container tags (by providing our container ID), so that we can "reference" those container tags in places the agent cannot reach, like DSM checkpoints or DBM query comments.

tracer/src/Datadog.Trace/Agent/DiscoveryService/DiscoveryService.cs
vandonr: I don't really see a practical case where we'd like to reset our container tags hash... Like, either they change or they stay the same, but they cannot really disappear, it'd mean that the app exited the container while still running ^^

At the same time, I don't see a real reason why the agent would reply with null, BUT it could happen if some code path don't request the value in the headers, which could happen for good reasons (we could, for performance reasons, say that we don't need the hash on every single request if we already have one). I feel like it's safer to write it that way, especialy considering how far in the code the code "requesting" the header is

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.NetFramework.cs
lucaspimentel: From the conversation in #8363 (closed), it's ok to remove this.

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
lucaspimentel: Yeah, that's a bit confusing. `BinaryPrimitives.WriteUInt64LittleEndian()` writes the `UInt64` as 8 bytes into the 12-byte buffer. Then `Base64.EncodeToUtf8InPlace()` converts them into a 12-char base64 string encoded as UTF-8.

I think what Andrew means is that we know all that chars in base64 are within the ASCII range (so they are all 1-byte chars, no multi-byte chars), so the 12 chars fit into the 12-byte buffer. If there was any chance of having multi-byte UTF-8 chars, we would need a bigger buffer.

vandonr: I ended up putting the first part of the comment above, where we allocate 12 bytes, I think it's enough to explain the magic number, not sure we need to go into more details ?

lucaspimentel:
```diff
- keeps it at 8 bytes
+ keeps it at 12 bytes
```

tracer/test/Datadog.Trace.Tests/ServiceRemappingHashTests.cs
```diff
@@ -0,0 +1,68 @@
...
+    [InlineData("svc.auto:service", "container0", "ANVkaLuv_RQ")]
+    public void Compute_ReturnsExpectedUrlSafeBase64WithoutPadding(string processTags, string containerTagsHash, string expected)
+    {
+        var hash = new ServiceRemappingHash(processTags);
+        hash.UpdateContainerTagsHash(containerTagsHash);
+
+        hash.Base64Value.Should().Be(expected);
+        hash.Base64Value.Should().NotContain("+");
+        hash.Base64Value.Should().NotContain("/");
+        hash.Base64Value.Should().NotContain("=");
```
lucaspimentel, line 66: [nit] If we assert that `hash.Base64Value == expected`, then these 3 `NotContain()` are redundant (assuming `expected` is valid and never has these chars).

vandonr: yes, I saw that (reviewing AI generated tests :D ), but I left them as I thought it conveyed well the intent of the test.
You're gonna say then remove the assert equals, BUT that one helps make it obvious (in the test params) that the hashes we are testing _would_ contain the incriminating characters
So in light of that, I decided to leave all the asserts

lucaspimentel: > You're gonna say then remove the assert equals

No, that's the best test because it checks the output is actually what we expect and that we don't have a bug which results in every hash being `AAAAAAA` or something 😅

> I left them as I thought it conveyed well the intent of the test

That's fair, thanks.

tracer/src/Datadog.Trace/ServiceRemappingHash.cs
andrewlock: Yeah, that was a typo, should have been 12 bytes at the end 😅

PR #8066, opened 2026-01-15
[ASM] Fix RC configs apply order

daniel-romano-DD:
## Summary of changes
- Refactors ASM Remote Configuration application to process **removals and updates together** per product, instead of handling them in two separate passes.

## Reason for change
- Fixes incorrect or unstable **apply order** when Remote Configuration delivers a mix of removals and updates for ASM products, which could lead to stale configs being retained or ordering-dependent behavior.

### DIFF
--- a/tracer/src/Datadog.Trace/AppSec/Rcm/AsmDdProduct.cs
+++ b/tracer/src/Datadog.Trace/AppSec/Rcm/AsmDdProduct.cs
@@ -1,4 +1,4 @@
-﻿// <copyright file="AsmDdProduct.cs" company="Datadog">
+// <copyright file="AsmDdProduct.cs" company="Datadog">
 // Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2 License.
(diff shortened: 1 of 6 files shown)


tracer/src/Datadog.Trace/AppSec/Rcm/AsmDdProduct.cs
```diff
@@ -17,34 +17,41 @@ internal sealed class AsmDdProduct : IAsmConfigUpdater
...
         }
-    }
 
-    public void ProcessRemovals(ConfigurationState configurationStatus, List<RemoteConfigurationPath> removedConfigsForThisProduct)
-    {
-        foreach (var removedConfig in removedConfigsForThisProduct)
+        if (files is { Count: > 0 })
         {
-            configurationStatus.RulesetConfigs.Remove(removedConfig.Path);
+            // var file = files.First();
```
e-n-0, line 32: leftover I guess 😬

PR #8068, opened 2026-01-15
Fix `Environment.Version` causing huge allocation

andrewlock:
## Summary of changes

- Don't call `Environment.Version` except in `FrameworkDescription`
- Fix `ActivityEnumerationHelper` to use `FrameworkDescription` instead of `Environment.Version`

## Reason for change

`Environment.Version` allocates every time you access it. So my "improve performance by checking if we're in .NET 10" PR caused a [massive allocation regression](https://app.datadoghq.com/dashboard/9eu-mfe-5ad?fromUser=true&refresh_mode=sliding&tpl_var_branch%5B0%5D=%28master%20OR%20benchmarks%5C%2F%2A%29&tpl_var_runtime%5B0%5D=%2A&from_ts=1767877815370&to_ts=1768482615370&live=true) 💀 😂 

image

So I think it makes sense to cache `Environment.Version` in our `FrameworkDescription` type, and access it from there.

## Implementation details

- Read (or calculate, where it can't be trusted) the `Environment.Version` for the app in `FrameworkDescription`
- Expose it as `FrameworkDescription.RuntimeVersion`
- Use the exposed version where possible
- Add `#nullable enable` seeing as I need to think about all that properly anyway

## Test coverage

This fixes the benchmarks again now 🙄 Everything else should be covered by existing, but I'll likely follow up with some more tests in a follow up PR (just don't want to delay this getting in!)

## Other details

The benchmarks _did_ show the regression in the PR https://github.com/DataDog/dd-trace-dotnet/pull/8058... but I wasn't looking, because those numbers are so flaky 😬

### DIFF
--- a/tracer/src/Datadog.Trace/Activity/Helpers/ActivityEnumerationHelper.cs
+++ b/tracer/src/Datadog.Trace/Activity/Helpers/ActivityEnumerationHelper.cs
@@ -24,6 +24,14 @@ namespace Datadog.Trace.Activity.Helpers;
 internal static class ActivityEnumerationHelper
 {
     private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(ActivityEnumerationHelper));
+
+    private static readonly bool IsDotNet10
+#if NET10_0_OR_GREATER
+        = true;
+#else
+        = FrameworkDescription.Instance.RuntimeVersion.Major >= 10;
+#endif
+
     private static TagObjectsEnumerator.AllocationFreeForEachDelegate? _tagObjectsEnumerator;
     private static TagsEnumerator.AllocationFreeForEachDelegate? _tagsEnumerator;
     private static Type? _tagObjectsType;
@@ -35,7 +43,7 @@ internal static class ActivityEnumerationHelper
     public static void EnumerateTagObjects<T>(T activity5, ref OtelTagsEnumerationState state, TagObjectsEnumerator.ForEachDelegate funcToRun)
         where T : IActivity5
     {
-        if (Environment.Version.Major >= 10)
+        if (IsDotNet10)
         {
             // .NET 10 doesn't allocate _anyway_, so we should just do naive enumeration
             foreach (var value in activity5.TagObjects)
@@ -83,7 +91,7 @@ internal static class ActivityEnumerationHelper
     public static void EnumerateTags<T>(T activity, ref OtelTagsEnumerationState state, TagsEnumerator.ForEachDelegate funcToRun)
         where T : IActivity
     {
-        if (Environment.Version.Major >= 10)
+        if (IsDotNet10)
         {
             // .NET 10 doesn't allocate _anyway_, so we should just do naive enumeration
             foreach (var value in activity.Tags)


tracer/src/Datadog.Trace/Activity/Helpers/ActivityEnumerationHelper.cs
```diff
@@ -24,6 +24,14 @@ namespace Datadog.Trace.Activity.Helpers;
 internal static class ActivityEnumerationHelper
 {
     private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(ActivityEnumerationHelper));
+
+    private static readonly bool IsDotNet10
+#if NET10_0_OR_GREATER
```
andrewlock, line 29: We don't build for this now, but maybe we will in the future, so why not

tracer/src/Datadog.Trace/FrameworkDescription.cs
```diff
@@ -15,36 +18,22 @@ internal sealed partial class FrameworkDescription
 
         private static readonly Assembly RootAssembly = typeof(object).Assembly;
 
-        private static readonly Tuple<int, string>[] DotNetFrameworkVersionMapping =
```
andrewlock, line 18: Instead of allocating this array, then enumerating it using linq, just converted into a big switch statement below

tracer/src/Datadog.Trace/FrameworkDescription.NetCore.cs
```diff
@@ -13,21 +15,20 @@ namespace Datadog.Trace
...
     {
-        private static FrameworkDescription _instance = null;
+        private const string Unknown = "unknown";
+        private static readonly Lazy<FrameworkDescription> _instance = new(Create);
 
-        public static FrameworkDescription Instance
-        {
-            get { return _instance ?? (_instance = Create()); }
-        }
+        public static FrameworkDescription Instance => _instance.Value;
```
andrewlock, line 21: I couldn't really see a reason that we were doing this in a "fake lazy" way 🤔 I _think_ we can get rid of the `Lazy<T>` too, but it seemed safer to not, just in case

tracer/src/Datadog.Trace/FrameworkDescription.NetCore.cs
```diff
@@ -77,69 +78,70 @@ public static FrameworkDescription Create()
...
 
         public bool IsCoreClr()
         {
             return Name.ToLowerInvariant().Contains("core") || IsNet5();
         }
 
-        private static string GetNetCoreOrNetFrameworkVersion()
+        private static void GetNetCoreOrNetFrameworkVersion(out Version version, out string productVersion)
         {
-            string productVersion = null;
```
andrewlock, line 90: These changes basically just grab `Environment.Version` now _or_ if it can't be trusted, try to parse it from the calculated `productVersion`

tracer/src/Datadog.Trace/FrameworkDescription.cs
```diff
@@ -59,6 +48,8 @@ private FrameworkDescription(
 
         public string OSDescription { get; }
 
+        public Version RuntimeVersion { get; }
+
         public static bool IsNet5()
         {
             return Environment.Version.Major >= 5;
```
bouwkast, line 55: Is this `Environment.Version` an issue as well?

andrewlock: Yup, fixed in https://github.com/DataDog/dd-trace-dotnet/pull/8069, just didn't want to delay getting _this_ fix in 😄

tracer/src/Datadog.Trace/FrameworkDescription.cs
```diff
@@ -110,5 +101,64 @@ private static string GetVersionFromAssemblyAttributes()
 
             return productVersion;
         }
+
+        private static bool GetDotNetFrameworkProductMapping(
```
lucaspimentel, line 105: This could probably also move to `FrameworkDescription.NetFramework.cs` to exclude the code entirely for other TFMs.

andrewlock: So `FrameworkDescription.NetFramework` is _only_ in .NET Framework, but given the above is needed for .NET Standard 2.0 as well, it's probably correct?

PR #8069, opened 2026-01-15
Replace remaining usages of `Environment.Version`

andrewlock:
## Summary of changes

Replaced remaining usages of `Environment.Version` other than `FrameworkDescription`

## Reason for change

`Environment.Version` allocates, but it's used in a bunch of places unnecessarily. Switching to use `FrameworkDescription.Runtimeversion` removes those allocations 

## Implementation details

- Find everywhere using `Environment.Version` and fix it to not
- Inlined a couple of places where it didn't seem necessary
- Remove an allocating path in `IsCoreClr()`

## Test coverage

Should be covered by existing

## Other details

Stacked on 
- https://github.com/DataDog/dd-trace-dotnet/pull/8068

### DIFF
--- a/tracer/src/Datadog.Trace/Debugger/ExceptionAutoInstrumentation/ExceptionReplayProbe.cs
+++ b/tracer/src/Datadog.Trace/Debugger/ExceptionAutoInstrumentation/ExceptionReplayProbe.cs
@@ -73,7 +73,7 @@ namespace Datadog.Trace.Debugger.ExceptionAutoInstrumentation
         private bool CheckIfMethodMayBeOmittedFromCallStack()
         {
             return ExceptionTrackManager.IsEditAndContinueFeatureEnabled &&
-                   FrameworkDescription.Instance.IsCoreClr() && RuntimeHelper.IsNetOnward(6) && Method.Method.DeclaringType?.Assembly != null && RuntimeHelper.IsModuleDebugCompiled(Method.Method.DeclaringType.Assembly);
+                   FrameworkDescription.Instance.RuntimeVersion.Major >= 6 && Method.Method.DeclaringType?.Assembly != null && RuntimeHelper.IsModuleDebugCompiled(Method.Method.DeclaringType.Assembly);
         }
 
         private void ProcessCase(ExceptionCase @case)
--- a/tracer/src/Datadog.Trace/Debugger/Helpers/RuntimeHelper.cs
+++ b/tracer/src/Datadog.Trace/Debugger/Helpers/RuntimeHelper.cs
@@ -17,11 +17,6 @@ namespace Datadog.Trace.Debugger.Helpers
 {
     internal static class RuntimeHelper
     {
-        internal static bool IsNetOnward(int major)
-        {
-            return Environment.Version.Major >= major;
-        }
-
         internal static bool IsModuleDebugCompiled(Assembly assembly)
         {
             var debuggableAttribute = assembly.GetCustomAttributes(typeof(DebuggableAttribute), false).FirstOrDefault() as DebuggableAttribute;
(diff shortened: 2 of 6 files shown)


tracer/src/Datadog.Trace/Debugger/Helpers/RuntimeHelper.cs
```diff
@@ -17,11 +17,6 @@ namespace Datadog.Trace.Debugger.Helpers
 {
     internal static class RuntimeHelper
     {
-        internal static bool IsNetOnward(int major)
```
andrewlock, line 20: This was only used in the one place, above, so just inlined it for clarity

tracer/src/Datadog.Trace/Debugger/ExceptionAutoInstrumentation/ExceptionReplayProbe.cs
```diff
@@ -73,7 +73,7 @@ private bool ShouldInstrument()
         private bool CheckIfMethodMayBeOmittedFromCallStack()
         {
             return ExceptionTrackManager.IsEditAndContinueFeatureEnabled &&
-                   FrameworkDescription.Instance.IsCoreClr() && RuntimeHelper.IsNetOnward(6) && Method.Method.DeclaringType?.Assembly != null && RuntimeHelper.IsModuleDebugCompiled(Method.Method.DeclaringType.Assembly);
```
andrewlock, line 76: `RuntimeHelper.IsNetOnward(6)` implies `FrameworkDescription.Instance.IsCoreClr()` so the first check is redundant

tracer/src/Datadog.Trace/RuntimeMetrics/DiagnosticsMetricsRuntimeMetricsListener.cs
```diff
@@ -57,15 +57,16 @@ public DiagnosticsMetricsRuntimeMetricsListener(IStatsdManager statsd)
         };
 
         // ASP.NET Core metrics are only available on .NET 8+
-        _aspnetcoreMetricsAvailable = Environment.Version.Major >= 8;
+        var version = FrameworkDescription.Instance.RuntimeVersion;
```
andrewlock, line 60: Just extract the variable and use the existing version

tracer/src/Datadog.Trace/FrameworkDescription.NetCore.cs
```diff
@@ -84,7 +84,7 @@ public static FrameworkDescription Create()
 
         public bool IsCoreClr()
         {
-            return Name.ToLowerInvariant().Contains("core") || IsNet5();
+            return RuntimeVersion.Major >= 5 || Name.IndexOf("core", StringComparison.OrdinalIgnoreCase) >= 0;
```
andrewlock, line 87: Don't allocate when looking for `"core"`

dudikeleti (approved):

tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs
```diff
@@ -106,9 +106,9 @@ internal RuntimeMetricsWriter(IStatsdManager statsd, TimeSpan delay, bool inAzur
                 {
                     { } x when x.IsWindows() => true, // Works on Windows
                     { } x when !x.IsCoreClr() => true, // Works on .NET Framework
-                    _ when Environment.Version is { Major: >= 5 } => true, // Works on .NET 5 and above
-                    _ when Environment.Version is { Major: 3, Minor: > 0 } => true, // 3.1 works
-                    _ when Environment.Version is { Major: 3, Minor: 0 } => false, // 3.0 is broken on linux
+                    { RuntimeVersion: { Major: >= 5 } } => true, // Works on .NET 5 and above
+                    { RuntimeVersion: { Major: 3, Minor: > 0 } } => true, // 3.1 works
+                    { RuntimeVersion: { Major: 3, Minor: 0 } } => false, // 3.0 is broken on linux
```
lucaspimentel, line 111: Why not remove this line and let it fall into the default branch?

PR #8070, opened 2026-01-15
[Test Optimization] perf: logging/timing helpers

tonyredondo:
## Summary of changes
- Add FixedSizeArrayPool and logger overloads to reduce per-log allocations.
- Add RefStopwatch/StopwatchHelpers/CodeDuration and apply timing to TraceClock/TracerManagerFactory/Instrumentation and CI processors.
- Lower init logs to Debug for CI protocol writer/sender and processors.

## Reason for change
Reduce allocations and improve timing diagnostics in hot paths.

## Implementation details
- New utility helpers in `Datadog.Trace.Util` for allocation-free timing and pooling.
- Logging changes to avoid params array allocations in common log paths.
- Timing/log-level tweaks in CI protocol writer/sender and trace processors.

## Test coverage
CI passes then all changes are good.

## Other details
- Stacked PRs (current marked):
  - **CURRENT**: PR1 https://github.com/DataDog/dd-trace-dotnet/pull/8070
  - PR2 https://github.com/DataDog/dd-trace-dotnet/pull/8071
  - PR3 https://github.com/DataDog/dd-trace-dotnet/pull/8072
  - PR4 https://github.com/DataDog/dd-trace-dotnet/pull/8073


tracer/src/Datadog.Trace/Logging/Internal/DatadogLogging.cs
```diff
@@ -54,7 +54,11 @@ static DatadogLogging()
         public static IDatadogLogger GetLoggerFor(Type classType)
         {
             // Tells us which types are loaded, when, and how often.
-            SharedLogger.Debug("Logger retrieved for: {AssemblyQualifiedName}", classType.AssemblyQualifiedName);
+            if (SharedLogger.IsEnabled(LogEventLevel.Debug))
```
tonyredondo, line 57: RuntimeType.AssemblyQualifiedName always allocates a new string and do some internals QCalls, is fast but we can avoid it.

tracer/src/Datadog.Trace/Logging/Internal/DatadogSerilogLogger.cs
```diff
@@ -200,38 +210,77 @@ public void CloseAndFlush()
...
+            // Avoid boxing + array allocation if disabled
+            if (!_logger.IsEnabled(level))
             {
-                // Avoid boxing + array allocation if disabled
-                WriteIfNotRateLimited(level, exception, messageTemplate, [property], sourceLine, sourceFile, skipTelemetry);
+                return;
             }
+
+            using var array = FixedSizeArrayPool<object?>.OneItemPool.Get();
+            array.Array[0] = property;
```
tonyredondo, line 220: previously we allocate an array for every single log line, here the idea is to have a small pool for it and reuse it.

tracer/src/Datadog.Trace/TraceClock.cs
```diff
@@ -51,7 +51,7 @@ public static TraceClock Instance
     public DateTimeOffset UtcNow
     {
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        get => _utcStart.Add(Elapsed);
+        get => _utcStart.AddTicks(StopwatchHelpers.GetElapsedTicks(Stopwatch.GetTimestamp() - _timestamp));
```
tonyredondo, line 54: we really don't need to create a TimeSpan struct for this by calling Elapsed, we just count the ticks and add them.

tracer/src/Datadog.Trace/Util/CodeDuration.cs
```diff
@@ -0,0 +1,173 @@
...
+using System;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using Datadog.Trace.Logging;
+using Datadog.Trace.Vendors.Serilog.Events;
+
+namespace Datadog.Trace.Util;
+
+internal struct CodeDuration : IDisposable
```
tonyredondo, line 29: This is the version for use in an async/await method because we cannot use ref struct there. So creating a `CodeDuration` in that case will allocate in the heap (struct boxing).

Below there's the `ref struct` version to avoid any heap allocation in a normal sync method.

tracer/src/Datadog.Trace/Util/CodeDuration.cs
```diff
@@ -0,0 +1,173 @@
...
+                _prefix,
+                StopwatchHelpers.GetElapsedMilliseconds(endTime - started),
+                _memberName,
+                _sourceFilePath,
+                _sourceLineNumber);
+        }
+    }
+}
+
+internal ref struct CodeDurationRef
```
tonyredondo, line 102: `ref struct` version to avoid an allocation in the heap.

tracer/src/Datadog.Trace/Util/FixedSizeArrayPool.cs
```diff
@@ -0,0 +1,144 @@
...
+        }
+#else
+        if (value is null || value.Length != _arrayItems)
+        {
+            return;
+        }
+#endif
+
+#if NETCOREAPP3_0_OR_GREATER
+        if (RuntimeHelpers.IsReferenceOrContainsReferences<T>())
```
tonyredondo, line 77: we clear the array if we know that `T` is a reference type (so we don't hold any reference from the GC)

tracer/src/Datadog.Trace/Util/RefStopwatch.cs
```diff
@@ -0,0 +1,35 @@
...
+// </copyright>
+
+#nullable enable
+using System;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+
+namespace Datadog.Trace.Util;
+
+internal ref struct RefStopwatch
```
tonyredondo, line 17: normal Stopwatch is a class, and we use multiple in the Initialization and in some of the CI visibility hotpath.
This version doesn't allocate (ref struct) but it's only useful in sync method, for async method is better to use the existing Stopwatch.

tracer/src/Datadog.Trace/Util/FixedSizeArrayPool.cs
```diff
@@ -0,0 +1,144 @@
...
+    private readonly ConcurrentStack<T[]> _items;
+    private readonly int _arrayItems;
+    private T[]? _fastPath;
+    private int _stackCount;
+
+    public FixedSizeArrayPool(int arrayItems)
+    {
+        if (arrayItems <= 0)
+        {
+            throw new ArgumentOutOfRangeException(nameof(arrayItems), arrayItems, "Array size must be greater than 0.");
```
lucaspimentel: Use `ThrowHelper.ThrowArgumentOutOfRangeException()`?

tracer/src/Datadog.Trace/Util/RefStopwatch.cs
lucaspimentel: Should we include this as a code comment?

tracer/src/Datadog.Trace/Util/FixedSizeArrayPool.cs
```diff
@@ -0,0 +1,144 @@
...
+            return value;
+        }
+
+        value = new T[_arrayItems];
+        return value;
+    }
+
+    private void InternalReturnArray(T[] value)
+    {
+#if DEBUG
```
lucaspimentel: The `Debug.Fail()` calls below get removed in release builds, so it becomes:

```csharp
if (value is null)
{
    return;
}

if (value.Length != _arrayItems)
{
    return;
}
```

Is it worth having the `#if DEBUG`?

tracer/src/Datadog.Trace/Util/FixedSizeArrayPool.cs
```diff
@@ -0,0 +1,144 @@
...
+
+using System;
+using System.Collections.Concurrent;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+using System.Threading;
+
+namespace Datadog.Trace.Util;
+
+internal sealed class FixedSizeArrayPool<T>
```
lucaspimentel, line 15: Can we add test for this?

tracer/src/Datadog.Trace/TraceClock.cs
lucaspimentel: is this measurably faster?

tracer/src/Datadog.Trace/Util/FixedSizeArrayPool.cs
tonyredondo: done in [6aee9f7](https://github.com/DataDog/dd-trace-dotnet/pull/8070/commits/6aee9f78b051bbf3f91b01fc3770b61180d1d8a0)

tracer/src/Datadog.Trace/Util/FixedSizeArrayPool.cs
tonyredondo: done in [6aee9f7](https://github.com/DataDog/dd-trace-dotnet/pull/8070/commits/6aee9f78b051bbf3f91b01fc3770b61180d1d8a0)

tracer/src/Datadog.Trace/Util/FixedSizeArrayPool.cs
tonyredondo: done in [6aee9f7](https://github.com/DataDog/dd-trace-dotnet/pull/8070/commits/6aee9f78b051bbf3f91b01fc3770b61180d1d8a0)

tracer/src/Datadog.Trace/TraceClock.cs
tonyredondo: not in recent frameworks, in olders frameworks the codegen is just a little better.

tracer/src/Datadog.Trace/Util/RefStopwatch.cs
tonyredondo: done in [4a649b1](https://github.com/DataDog/dd-trace-dotnet/pull/8070/commits/4a649b190a78ea4ba2150fef13de5119927e900f)

tracer/src/Datadog.Trace/Util/CodeDuration.cs
dudikeleti: Would it make sense to add a short comment in code about when to use CodeDuration vs CodeDurationRef?
My concern is that if someone uses the struct version in a non async methos and in a non-local way (copied by value) he could end up calling Dispose() on multiple copies and get an incorrect CodeDurationBase.Count.

tonyredondo: done in [5cb68e6](https://github.com/DataDog/dd-trace-dotnet/pull/8070/commits/5cb68e6cfddb714dd18cb49fbacb6ea25fb56375)

tracer/src/Datadog.Trace/Util/CodeDuration.cs
anna-git: NIT: wondering if the code inside the methods / ctors could have been shared between the two via some static helpers , for maintenance

PR #8074, opened 2026-01-15
[FFE] Fix exposures url

daniel-romano-DD:


andrewlock (commented):
:oops: we should definitely add some integration tests to make it easy to confirm behaviours once we get the system tests locked down 😅

PR #8079, opened 2026-01-16
Refactor Azure Functions instrumentation before #7628

lucaspimentel:
## Summary of changes

Refactor Azure Functions instrumentation code as a precursor to #7628.

## Reason for change

The changes in #7628 were building up, so I decided to split them into two PRs. This PR is mostly cleanup and refactoring. No functional changes.

## Implementation details

**Environment detection helpers** (`EnvironmentHelpers`)
- Added `IsAzureFunctionsIsolated()` - detects isolated worker runtime
- Added `GetAzureFunctionsWorkerRuntime()` - retrieves worker runtime value
- Added `GetAzureFunctionsExtensionVersion()` - retrieves extension version
- Refactored `IsRunningInAzureFunctionsHost()` to use new helpers

**Diagnostic observer initialization** (`Instrumentation`)
- Extracted `GetAspNetCoreDiagnosticObserver()`
- Extracted `SkipAspNetCoreDiagnosticObserver()`

**Azure Functions instrumentation cleanup** (`AzureFunctionsCommon`)
- Renamed parameter `context` to `functionContext` for clarity and consistency
- Made `CreateIsolatedFunctionScope()` private
- Formatting: improved comment alignment and whitespace
- Changed `if` to `switch` for trigger type handling

**ASP.NET Core tracking key rename**
- Renamed `HttpContextItemsKey` to `HttpContextTrackingKey` across all usages (will add another key in next PR, needed to distinguish them since they are both "item keys")

## Test coverage

No new tests added - this is a refactoring/cleanup PR with no functional changes. Existing Azure Functions tests provide coverage.

## Other details

Precursor to #7628 - Fix span parenting for isolated Azure Functions using ASP.NET Core integration

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
@@ -367,7 +367,7 @@ namespace Datadog.Trace.ClrProfiler
                     // ignore
                 }
             }
-#endif
+#endif // #if !NETFRAMEWORK
 
             try
             {
@@ -496,7 +496,7 @@ namespace Datadog.Trace.ClrProfiler
             {
                 return new SingleSpanAspNetCoreDiagnosticObserver(Tracer.Instance, Security.Instance, Iast.Iast.Instance, spanCodeOrigin: null);
             }
-#endif
+#endif // #if NET6_0_OR_GREATER
 
             return new AspNetCoreDiagnosticObserver(Tracer.Instance, Security.Instance, Iast.Iast.Instance, spanCodeOrigin: null);
         }


tracer/src/Datadog.Trace/ClrProfiler/Instrumentation.cs
```diff
@@ -473,43 +474,41 @@ private static void StartDiagnosticManager()
         {
             var observers = new List<DiagnosticObserver>();
 
-            // get environment variables directly so we don't access Trace.Instance yet
-            var functionsExtensionVersion = EnvironmentHelpers.GetEnvironmentVariable(PlatformKeys.AzureFunctions.FunctionsExtensionVersion);
-            var functionsWorkerRuntime = EnvironmentHelpers.GetEnvironmentVariable(PlatformKeys.AzureFunctions.FunctionsWorkerRuntime);
-
-            if (!string.IsNullOrEmpty(functionsExtensionVersion) && !string.IsNullOrEmpty(functionsWorkerRuntime))
+            if (!SkipAspNetCoreDiagnosticObserver())
```
pablomartinezbernardo, line 477: Previously this was looking at `FUNCTIONS_EXTENSION_VERSION` and `FUNCTIONS_WORKER_RUNTIME`, and now it is doing `EnvironmentHelpers.IsAzureFunctions()` which adds `WEBSITE_SITE_NAME` to the check. Seems totally intended, just mentioning it because the PR says there are no functional changes and I'm not sure if this has other consequences

lucaspimentel: Good catch, thanks! I'll take a closer look.

lucaspimentel: Ahh, yes. In `EnvironmentHelpers` we also check `WEBSITE_SITE_NAME` which we didn't do here before. It shouldn't make any difference in practice (all 3 env vars should always be set). We only check for the additional env vars to be safe, I guess.

PR #8080, opened 2026-01-16
[IBM MQ] Fix queue naming by removing URI prefix

johannbotha:
## Summary
- Sanitize IBM MQ queue names by stripping `queue://` and `topic://` URI scheme prefixes
- Applied to DSM edge tags, span tags, and resource names
- Aligns with how other messaging integrations report queue/topic names

**Before:** `queue://my_ibmmq.queue.1`
**After:** `my_ibmmq.queue.1`

## Changes
- Added `SanitizeQueueName()` method to `IbmMqHelper.cs`
- Updated `PutIntegration.cs` and `GetIntegration.cs` to use sanitized names in DSM edge tags
- Updated `CreateProducerScope()` and `CreateConsumerScope()` to use sanitized names

## Test plan
- [x] Added unit tests for `SanitizeQueueName` method covering:
  - `queue://` prefix removal (case insensitive)
  - `topic://` prefix removal (case insensitive)
  - Preserving names without prefixes
  - Null and empty string handling
  - Other URI schemes preserved (http, https, amqp, etc.)
- [x] All 22 tests passing

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/tracer/test/Datadog.Trace.Tests/ClrProfiler/AutoInstrumentation/IbmMq/IbmMqHelperTests.cs
+++ b/tracer/test/Datadog.Trace.Tests/ClrProfiler/AutoInstrumentation/IbmMq/IbmMqHelperTests.cs
@@ -74,10 +74,12 @@ namespace Datadog.Trace.Tests.ClrProfiler.AutoInstrumentation.IbmMq
             result.Should().Be(expected);
         }
 
-        [Fact]
-        public void SanitizeQueueName_HandlesEmptyQueueNameAfterPrefix()
+        [Theory]
+        [InlineData("queue://")]
+        [InlineData("topic://")]
+        public void SanitizeQueueName_HandlesEmptyNameAfterPrefix(string input)
         {
-            var result = IbmMqHelper.SanitizeQueueName("queue://");
+            var result = IbmMqHelper.SanitizeQueueName(input);
 
             result.Should().Be(string.Empty);
         }


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/IbmMq/IbmMqHelper.cs
```diff
@@ -16,6 +16,33 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.IbmMq;
...
+    private const string QueueUriPrefix = "queue://";
+    private const string TopicUriPrefix = "topic://";
+
+    /// <summary>
+    /// Sanitizes an IBM MQ queue name by removing URI scheme prefixes.
+    /// Converts names like "queue://my_queue" to "my_queue".
+    /// </summary>
+    internal static string SanitizeQueueName(string? queueName)
+    {
+        if (queueName is null || queueName.Length == 0)
```
robcarlan-datadog: [nit] ```string.IsNullOrEmpty(queueName)```instead

johannbotha: Actually changed this on purpose because got a compiler error with what you're suggesting

`error CS8602: Dereference of a possibly null reference.`

lucaspimentel: We have a custom `StringUtil.IsNullOrEmpty(string)` which fixes the issues with `string.IsNullOrEmpty(string)` nullability warnings on older runtimes. [source](https://github.com/DataDog/dd-trace-dotnet/blob/acbb3f99f645b66e1d80d8179931401899cfabe7/tracer/src/Datadog.Trace/Util/StringUtil.cs#L28)

johannbotha: Thanks for the pointer! Updated to use `StringUtil.IsNullOrEmpty`.

tracer/test/Datadog.Trace.Tests/ClrProfiler/AutoInstrumentation/IbmMq/IbmMqHelperTests.cs
```diff
@@ -0,0 +1,92 @@
...
+        {
+            var result = IbmMqHelper.SanitizeQueueName(input);
+
+            result.Should().Be(expected);
+        }
+
+        [Fact]
+        public void SanitizeQueueName_HandlesQueueNameWithSpecialCharacters()
+        {
+            var result = IbmMqHelper.SanitizeQueueName("queue://MY.QUEUE_NAME-123");
```
lucaspimentel: This test name says "with special characters." What are the special characters here?
Can we remove `SanitizeQueueName_HandlesQueueNameWithSpecialCharacters()`, or include this test case in `SanitizeQueueName_RemovesQueueUriPrefix()`?

tracer/test/Datadog.Trace.Tests/ClrProfiler/AutoInstrumentation/IbmMq/IbmMqHelperTests.cs
```diff
@@ -0,0 +1,92 @@
...
+
+            result.Should().Be("MY.QUEUE_NAME-123");
+        }
+
+        [Fact]
+        public void SanitizeQueueName_HandlesEmptyQueueNameAfterPrefix()
```
lucaspimentel: Can we convert this from `[Fact]` to `[Theory]` and test for `"topic://"` as well?

lucaspimentel (approved):

tracer/test/Datadog.Trace.Tests/ClrProfiler/AutoInstrumentation/IbmMq/IbmMqHelperTests.cs
johannbotha: Done - moved to the existing Theory.

tracer/test/Datadog.Trace.Tests/ClrProfiler/AutoInstrumentation/IbmMq/IbmMqHelperTests.cs
johannbotha: Done - converted to Theory with both `queue://` and `topic://`.

PR #8081, opened 2026-01-20
Fix Batching sink tests flakiness.

NachoEchevarria:
## Summary of changes

Fixes race conditions in `BatchingSinkTests` that caused [intermittent test failures in CI.](https://dev.azure.com/datadoghq/a51c4863-3eb4-4c5d-878a-58b41a049e4e/_apis/build/builds/193774/logs/8036)

## Reason for change

Failing tests are WhenRunning_AndAnEventIsQueued_ItIsWrittenToABatch and WhenRunning_AndAnEventIsQueued_ItIsWrittenToABatchOnDispose

Both tests were failing intermittently with:
```
Expected sink.Batches.TryPeek(out var batch) to be 1, but found 0 (difference of -1).
```

The tests have been marked as flaky until solved.

### DIFF
--- a/tracer/test/Datadog.Trace.Tests/Logging/DirectSubmission/Sink/PeriodicBatching/BatchingSinkTests.cs
+++ b/tracer/test/Datadog.Trace.Tests/Logging/DirectSubmission/Sink/PeriodicBatching/BatchingSinkTests.cs
@@ -13,6 +13,7 @@ using System.Threading.Tasks;
 using Datadog.Trace.Logging.DirectSubmission.Formatting;
 using Datadog.Trace.Logging.DirectSubmission.Sink;
 using Datadog.Trace.Logging.DirectSubmission.Sink.PeriodicBatching;
+using Datadog.Trace.TestHelpers;
 using FluentAssertions;
 using Xunit;
 using Xunit.Abstractions;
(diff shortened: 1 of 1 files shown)


bouwkast:
@codex review

tracer/test/Datadog.Trace.Tests/Logging/DirectSubmission/Sink/PeriodicBatching/BatchingSinkTests.cs
```diff
@@ -67,9 +68,10 @@ public async Task WhenRunning_AndAnEventIsQueued_ItIsWrittenToABatch()
             var evt = new TestEvent("Some event");
 
             sink.EnqueueLog(evt);
-            var batches = await WaitForBatchesAsync(sink);
+            await sink.FlushAsync();
+            await sink.DisposeAsync();
```
bouwkast: Granted I think these tests are old and it was noted that they were loosely defined when they came it, but I think the initial ordering was expected.

As they are written now they appear to be the same and we've lost that initial expectation, but I'm unsure how accurate that expectation is/was looking at the `FlushAsync`/`DisposeAsync` flow

Maybe we could just remove this test now ?

NachoEchevarria:
Note: After a small conversation, @bouwkast and I agreed that marking the tests as flaky would be the best approach here.

PR #8086, opened 2026-01-21
[FFE] Added support for Value evaluation type in OpenFeature

daniel-romano-DD:
## Summary of changes
Add full support for OpenFeature Value type evaluation (JSON)

### DIFF
--- a/tracer/src/Datadog.FeatureFlags.OpenFeature/Datadog.FeatureFlags.OpenFeature.csproj
+++ b/tracer/src/Datadog.FeatureFlags.OpenFeature/Datadog.FeatureFlags.OpenFeature.csproj
@@ -44,7 +44,12 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="OpenFeature" Version="[2.0.0, 3.0.0)" />
+    <PackageReference Include="OpenFeature" Version="2.0.0" />
+    <PackageReference Include="Newtonsoft.Json" Version="12.0.1" />
   </ItemGroup>
+
+  <PropertyGroup>
+    <DefineConstants>$(DefineConstants);INTERNAL_FFE</DefineConstants>
+  </PropertyGroup>
 
 </Project>
--- a/tracer/test/test-applications/integrations/Samples.OpenFeature/Evaluator.cs
+++ b/tracer/test/test-applications/integrations/Samples.OpenFeature/Evaluator.cs
@@ -1,5 +1,5 @@
 using System;
-using System.Threading;
+using System.Diagnostics;
 using OpenFeature.Model;
 
 namespace Samples.FeatureFlags;
@@ -51,5 +51,21 @@ class Evaluator
         }
 
         return (evaluation.Value, evaluation.ErrorMessage);
+    }
+
+    public static void ExtraChecks()
+    {
+        var key = "simple-json";
+        var context = EvaluationContext.Builder().Set("targetingKey", key).Build();
+
+        var defaultValue = new Value("Not found");
+        var evaluation = client.GetObjectDetailsAsync(key, defaultValue, context).Result;
+
+        Debug.Assert(evaluation is not null);
+        Debug.Assert(evaluation.ErrorMessage is null);
+        Debug.Assert(evaluation.Value != defaultValue);
+        Debug.Assert(evaluation.Value.IsStructure);
+        Debug.Assert(evaluation.Value.AsStructure.ContainsKey("integer"));
+        Debug.Assert(evaluation.Value.AsStructure.GetValue("integer").AsInteger == 1);
     }
 }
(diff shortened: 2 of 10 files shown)


tracer/src/Datadog.FeatureFlags.OpenFeature/Datadog.FeatureFlags.OpenFeature.csproj
```diff
@@ -44,7 +44,12 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="OpenFeature" Version="[2.0.0, 3.0.0)" />
+    <PackageReference Include="OpenFeature" Version="2.0.0" />
+    <PackageReference Include="Newtonsoft.Json" Version="12.0.1" />
```
NachoEchevarria, line 48: Why was Newtonsoft.Json chosen over the vendored version or System.Text.Json?

tracer/test/Datadog.Trace.Tests/Datadog.Trace.Tests.csproj
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <RunAnalyzersDuringBuild>true</RunAnalyzersDuringBuild>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <LangVersion>preview</LangVersion>
```
NachoEchevarria: Is this change intended?

tracer/test/test-applications/integrations/Samples.OpenFeature/Evaluator.cs
```diff
@@ -1,4 +1,6 @@
 using System;
+using System.Diagnostics;
+using System.Runtime.InteropServices.ComTypes;
```
NachoEchevarria: Are you using this one?

tracer/test/Datadog.Trace.Tests/Datadog.Trace.Tests.csproj
dromanol: Yes, it is. Compiler began complaining after last VS update

tracer/src/Datadog.FeatureFlags.OpenFeature/Datadog.FeatureFlags.OpenFeature.csproj
dromanol: Because this package does not have (must not) a dep with any of our libs. This is the most used version of Newtonsoft, as a min requirement

tracer/test/test-applications/integrations/Samples.OpenFeature/Evaluator.cs
dromanol: VS added it. This does no harm, specially in a Sample app.

dromanol: Removed

tracer/src/Datadog.FeatureFlags.OpenFeature/Datadog.FeatureFlags.OpenFeature.csproj
lucaspimentel: > this package does not have (must not) a dep with any of our libs

That explains why we don't use the vendored version, but what about `System.Text.Json`? It's supposed to have better performance.

tracer/test/test-applications/integrations/Samples.OpenFeature/Evaluator.cs
```diff
@@ -52,4 +52,20 @@ public static (string? Value, string? Error)? Evaluate(string key)
...
+
+        var defaultValue = new Value("Not found");
+        var evaluation = client.GetObjectDetailsAsync(key, defaultValue, context).Result;
+
+        Debug.Assert(evaluation is not null);
+        Debug.Assert(evaluation.ErrorMessage is null);
+        Debug.Assert(evaluation.Value != defaultValue);
+        Debug.Assert(evaluation.Value.IsStructure);
+        Debug.Assert(evaluation.Value.AsStructure.ContainsKey("integer"));
+        Debug.Assert(evaluation.Value.AsStructure.GetValue("integer").AsInteger == 1);
```
lucaspimentel, line 69: I don't know if we use `release` builds when we run tests, but if we do, none of these assert will run.

dromanol: Uhm, good point. I'll change it

tracer/src/Datadog.FeatureFlags.OpenFeature/Datadog.FeatureFlags.OpenFeature.csproj
lucaspimentel: Never mind. It doesn't support .NET Framework 4.6.1 and brings in a lot of other dependencies.

dromanol: That was the main reason (as Gemini suggested) XD

PR #8094, opened 2026-01-22
[Feature Flags] Fix ISO 8601 date parsing to support variable precision

leoromanovsky:
The date parser was using TryParseExact with a fixed 3-digit millisecond format, causing PARSE_ERROR for dates with 6-digit microsecond precision (e.g., "2025-09-23T15:48:37.235982Z") sent by the backend.

Changed to use DateTime.TryParse with DateTimeStyles.RoundtripKind which correctly handles all valid ISO 8601 date formats including:
- No fractional seconds: 2020-01-01T00:00:00Z
- 1-7 digit fractional seconds (beyond 7 digits truncated by .NET)

Added comprehensive unit tests for various date precisions and invalid date handling. Also added microsecond-date-test flag and test cases to the bundled test data.

## Test coverage

Associating a new system test (also discovered this condition in java): https://github.com/DataDog/system-tests/pull/6087

### DIFF
--- a/tracer/src/Datadog.Trace/FeatureFlags/FeatureFlagsEvaluator.cs
+++ b/tracer/src/Datadog.Trace/FeatureFlags/FeatureFlagsEvaluator.cs
@@ -405,20 +405,11 @@ namespace Datadog.Trace.FeatureFlags
                 return null;
             }
 
-            // Using TryParse instead of TryParseExact to support RFC 3339 dates with
-            // variable fractional second precision (0-9 digits). TryParseExact would
+            // Using Parse instead of ParseExact to support RFC 3339 dates with
+            // variable fractional second precision (0-9 digits). ParseExact would
             // require ~10 format strings. Since dates come from our controlled backend,
             // accepting broader date formats is acceptable.
-            if (DateTime.TryParse(
-                    dateString,
-                    CultureInfo.InvariantCulture,
-                    DateTimeStyles.RoundtripKind,
-                    out var dt))
-            {
-                return dt;
-            }
-
-            throw new FormatException($"Wrong date format: {dateString}");
+            return DateTime.Parse(dateString, CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind);
         }
 
         private static object? ResolveAttribute(string? name, EvaluationContext? context)


tracer/src/Datadog.Trace/FeatureFlags/FeatureFlagsEvaluator.cs
```diff
@@ -406,17 +431,16 @@ private static string GetMd5Hash(string input)
                 return null;
             }
 
-            if (DateTime.TryParseExact(
+            if (DateTime.TryParse(
```
lucaspimentel: `TryParseExact()` will have better performance than `TryParse()` since there's less "guessing" and less work overall. If we know the expected format, it might be better to stick to that spec and use `TryParseExact()`. If we need to support multiple formats, you can pass a `string[]` for formats to `TryParseExact()`.

Also, I don't think `RoundtripKind` applies here. That is used to keep the `DateTime.Kind` value when converting a `DateTime` to string and back. We're only parsing here, not converting to string.

tracer/test/Datadog.Trace.Tests/FeatureFlags/FeatureFlagsEvaluatorTests.cs
```diff
@@ -359,6 +359,84 @@ public void EvaluateExposureFlagLogsExposureEvent()
...
+    // ISO 8601 Date Parsing tests
+    // ---------------------------------------------------------------------
+
+    [Theory]
+    [InlineData("2020-01-01T00:00:00.123Z", "2099-12-31T23:59:59.456Z")] // 3-digit milliseconds
+    [InlineData("2020-01-01T00:00:00.123456Z", "2099-12-31T23:59:59.654321Z")] // 6-digit microseconds
+    [InlineData("2020-01-01T00:00:00.123456789Z", "2099-12-31T23:59:59.987654321Z")] // 9-digit nanoseconds (last 2 digits truncated by .NET)
+    [InlineData("2020-01-01T00:00:00Z", "2099-12-31T23:59:59Z")] // no fractional seconds
+    [InlineData("2020-01-01T00:00:00.1Z", "2099-12-31T23:59:59.9Z")] // 1-digit
+    [InlineData("2020-01-01T00:00:00.12Z", "2099-12-31T23:59:59.99Z")] // 2-digit
```
lucaspimentel, line 373: Do need to support all these formats?

tracer/src/Datadog.Trace/FeatureFlags/FeatureFlagsEvaluator.cs
lucaspimentel: I see we now tests for several formats, so maybe we need to support them all. If so, feel free to ignore this comment 😅

tracer/src/Datadog.Trace/FeatureFlags/FeatureFlagsEvaluator.cs
```diff
@@ -406,17 +431,16 @@ private static string GetMd5Hash(string input)
...
-                    System.Globalization.DateTimeStyles.AdjustToUniversal,
+                    CultureInfo.InvariantCulture,
+                    DateTimeStyles.RoundtripKind,
                     out var dt))
             {
                 return dt;
             }
 
-            throw new FormatException("Wrong date format");
+            throw new FormatException($"Wrong date format: {dateString}");
```
lucaspimentel: If we're always going to throw an exception when the format is invalid, any reason not use to `Parse()` (which throws `FormatException`) instead of `TryParse()`?

tracer/test/Datadog.Trace.Tests/FeatureFlags/FeatureFlagsEvaluatorTests.cs
leoromanovsky: The backend produces dates in RFC 3339 format; yes these are all possible values. In the future we hope to migrate to milliseconds since epoch for simpler parsing.

tracer/src/Datadog.Trace/FeatureFlags/FeatureFlagsEvaluator.cs
leoromanovsky: Agreed with your comments in general about performance. It's a goal of mine too. Big picture this parsing runs once every 5 seconds async, whereas flag evaluations have a very high rate of usage in the hot path of user code. What balance would you strike between being able to support rfc3339 dates and good performance here?

tracer/test/Datadog.Trace.Tests/FeatureFlags/FeatureFlagsEvaluatorTests.cs
lucaspimentel: I see, thanks! (resolving)

tracer/src/Datadog.Trace/FeatureFlags/FeatureFlagsEvaluator.cs
lucaspimentel: I did some more research and I think your code may be the best way to do this 😅. The RFC allows zero or more fractional digits. If we wanted to stick to `TryParseExact()` we would have to use a `string[]` with at least 10 (?) different formats, one for each number digits we want to support (0-9).

On the other hand, `TryParse()` allows _anything_ that looks like a date/time, so something like `"1/1/1"` would be considered valid, which feels dodgy. 🤷🏽‍♂️

PR #8099, opened 2026-01-23
Fix flaky test metrics tag values

NachoEchevarria:
## Summary of changes

We currently send telemetry metrcis related to flaky tests. These metrics, though, had some issues.

In some tests, the test name already contained the class name. The resulting test name would include repeated class names. For instance: "Datadog.Trace.Tests.Logging.DirectSubmission.Sink.PeriodicBatching.BatchingSinkTests.Datadog.Trace.Tests.Logging.DirectSubmission.Sink.PeriodicBatching.BatchingSinkTests.WhenRunning_AndAnEventIsQueued_ItIsWrittenToABatch"

We are sanitizing tag values by calling TryNormalizeTagName() which is useful for value normalization but adds some unneccessary restrictions. For instance, framewrok values were not being sent because they started with a number, which is fine for tag values but not for tag names. Therefore, TryNormalizeTagName will not be used for framework values.

Sometimes, the test name is not sent because we validate tags before sending them and we don't send tags with a length > 200. Test name has been trucated to 200 to avoid having empty test name tags.


tracer/test/Datadog.Trace.TestHelpers/CustomTestFramework.cs
```diff
@@ -349,7 +357,7 @@ async Task SendMetric(IMessageSink outputHelper, string metricName, string testF
                     var tags = $$"""
                                      "os.platform:{{SanitizeTagValue(FrameworkDescription.Instance.OSPlatform)}}",
                                      "os.architecture:{{SanitizeTagValue(EnvironmentTools.GetPlatform())}}",
-                                     "target.framework:{{SanitizeTagValue(FrameworkDescription.Instance.ProductVersion)}}",
+                                     "target.framework:{FrameworkDescription.Instance.ProductVersion}",
```
bouwkast:
```suggestion
                                     "target.framework:{{FrameworkDescription.Instance.ProductVersion}}",
```

Needs the extra set of `{` and `}`

tracer/test/Datadog.Trace.TestHelpers/CustomTestFramework.cs
```diff
@@ -296,7 +296,15 @@ protected override async Task<RunSummary> RunTestCaseAsync(IXunitTestCase testCa
                         }
 
                         _diagnosticMessageSink.OnMessage(new DiagnosticMessage($"RETRYING: {test} ({attemptsRemaining} attempts remaining, {retryReason})"));
-                        var testFullName = $"{TestMethod.TestClass.Class.Name}.{testCase.DisplayName}";
+                        var testFullName = (testCase.DisplayName.StartsWith(TestMethod.TestClass.Class.Name) ?
+                            testCase.DisplayName :
+                            $"{TestMethod.TestClass.Class.Name}.{testCase.DisplayName}").Trim();
+
+                        if (testFullName.Length > 200)
```
bouwkast: For readability in the future can we just extract `200` to some variable to note that it is the maximum name length? `var maxTestNameLength = 200`

tracer/test/Datadog.Trace.TestHelpers/CustomTestFramework.cs
NachoEchevarria: Thanks!

tracer/test/Datadog.Trace.TestHelpers/CustomTestFramework.cs
NachoEchevarria: Good idea! Done!

NachoEchevarria:
> Thanks! I think we have very-similar code copy-pasted in a bunch of places, we should probably fix those cases too?
> 
> e.g.:
> 
> * https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/test/Datadog.Trace.TestHelpers.AutoInstrumentation/ErrorHelpers.cs#L82
> * https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/build/_build/MetricHelper.cs
> 
> Thanks!

Nice catch, thanks! 
Done!

PR #8103, opened 2026-01-23
Fix `ProcessTags.GetLastPathSegment` for some edge cases, add  tests

lucaspimentel:
## Summary of changes

Simplified `ProcessTags.GetLastPathSegment()` to use `DirectoryInfo.Name` and fixed the edge case where root paths (e.g., `/`, `C:\`) would return an empty string.

## Reason for change

- **Bug fix**: The previous implementation returned an empty string for root paths like `/` or `C:\` because `Path.GetFileName()` returns empty for paths ending with a separator, and root paths always logically end with a separator.

## Implementation details

**ProcessTags.cs:**
- Replaced manual `TrimEnd` + `Path.GetFileName` logic with `new DirectoryInfo(path).Name`, which correctly handles:
  - Trailing separators
  - Root paths (returns the root itself)
  - Both Windows and Unix path formats
- Changed `TagsList` and `SerializedTags` to use `field ??=` pattern (`Lazy<T>` semantics) to delay initialization until first access
- Inlined `GetSerializedTagsFromList()` since it was a one-liner

## Test coverage

Added unit tests for `GetLastPathSegment`:
- `GetLastPathSegment_ReturnsLastDirectory_Windows` - Tests Windows paths (rooted, unrooted, with/without trailing separators, root paths, empty, null)
- `GetLastPathSegment_ReturnsLastDirectory_NonWindows` - Tests Unix paths (same scenarios)

## Other details

n/a

### DIFF
--- a/tracer/src/Datadog.Trace/ProcessTags.cs
+++ b/tracer/src/Datadog.Trace/ProcessTags.cs
@@ -60,6 +60,12 @@ internal static class ProcessTags
     [TestingAndPrivateOnly]
     internal static string GetLastPathSegment(string directoryPath)
     {
+        // See https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats
+        // Use DirectoryInfo.Name because it correctly handles
+        // - "/" and "\" separators
+        // - paths with or without trailing separators
+        // - root paths like "/" or "C:\"
+        // - other edge cases on Windows like device paths ("\\?\.\" etc) or "\\server\share" UNC paths
         return StringUtil.IsNullOrEmpty(directoryPath) ?
                    string.Empty :
                    new DirectoryInfo(directoryPath).Name;


tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -61,11 +61,72 @@ private static string GetSerializedTagsFromList(IEnumerable<string> tags)
     /// <summary>
     /// From the full path of a directory, get the name of the leaf directory.
     /// </summary>
-    private static string GetLastPathSegment(string directoryPath)
+    [TestingAndPrivateOnly]
+    internal static string GetLastPathSegment(string directoryPath)
     {
-        // Path.GetFileName returns an empty string if the path ends with a '/'.
-        // We could use Path.TrimEndingDirectorySeparator instead of the trim here, but it's not available on .NET Framework
-        return Path.GetFileName(directoryPath.TrimEnd('\\').TrimEnd('/'));
```
lucaspimentel, line 68: Note to reviewers: In .NET Framework, `TrimEnd(params char[])` would allocate an `char[]` for each of the two calls (because `params`) and would also allocate a new string if trimming was required. With these changes we avoid the two `params` arrays, and the intermediate trimmed string in .NET Core 3.1 and above.

Also, a bugfix: if `directoryPath` is a root path like `/` or `C:\` (although unlikely), `Path.GetFileName()` would return an empty string.

tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -61,11 +61,77 @@ private static string GetSerializedTagsFromList(IEnumerable<string> tags)
...
+        return Path.GetFileName(trimmedPath).ToString();
+#else
+        return Path.GetFileName(trimmedPath.ToString());
+#endif
+    }
+
+    [TestingOnly]
+    internal static string TrimEndingDirectorySeparator(string path)
+    {
+        return TrimEndingDirectorySeparator(path.AsSpan()).ToString();
```
bouwkast: I think this may need a null check even though it is just a null check

I think `path.AsSpan()` is throwing for one of the test cases that passes in `null`

tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -61,11 +61,77 @@ private static string GetSerializedTagsFromList(IEnumerable<string> tags)
...
+        if (path.IsEmpty)
+        {
+            return ReadOnlySpan<char>.Empty;
+        }
+
+        // Path.TrimEndingDirectorySeparator() is not available in older .NET versions.
+        // Not using "#if NETCOREAPP3_1_OR_GREATER" here to keep consistent behavior across TFMs.
+        var last = path[path.Length - 1];
+
+        return last == Path.DirectorySeparatorChar || last == Path.AltDirectorySeparatorChar ?
```
bouwkast: I think the tests are failing because the test input is a mix of Windows/Unix style and on Unix the Windows paths aren't working as the `DirectorySeparatorChar` and `AltDirectorySeparatorChar` are both `/`

Would we want to use something like ` last == '/' || last == '\\'  instead of `last == Path.DirectorySeparatorChar || last == Path.AltDirectorySeparatorChar`

lucaspimentel: Thanks! I ran this test locally on Windows only where one is ` \` and the other is `/` 🤦🏽
TIL they are both `/` on Linux. Looks like I need to hard-code them like you suggested.

tracer/src/Datadog.Trace/ProcessTags.cs
lucaspimentel: `AsSpan()` checks for null and returns an empty `ReadOnlySpan<char>` and the `null` tests pass on my machine.

bouwkast: Ahh gotcha thanks!

tracer/src/Datadog.Trace/ProcessTags.cs
bouwkast: Are these the `TrimEnd` that we vendored in and have the bug in them?

tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -53,19 +54,15 @@ private static void AddNormalizedTag(this List<string> tags, string key, string?
...
-    private static string GetLastPathSegment(string directoryPath)
+    [TestingAndPrivateOnly]
+    internal static string GetLastPathSegment(string directoryPath)
     {
-        // Path.GetFileName returns an empty string if the path ends with a '/'.
-        // We could use Path.TrimEndingDirectorySeparator instead of the trim here, but it's not available on .NET Framework
-        return Path.GetFileName(directoryPath.TrimEnd('\\').TrimEnd('/'));
+        return StringUtil.IsNullOrEmpty(directoryPath) ?
+                   string.Empty :
+                   new DirectoryInfo(directoryPath).Name;
```
bouwkast, line 71: Is creating a new `DirectoryInfo` the workaround to the TrimEnd bug?

bouwkast (approved):
Thanks 👍

tracer/src/Datadog.Trace/ProcessTags.cs
lucaspimentel: Not really. We could just use the `TrimEnd()` that uses `string` instead of `ReadOnlySpan<char>` and accept the extra allocations.

But I decided I was worrying to much about just a few allocations that only happen once. Handling paths is quite complicated and `DirectoryInfo.Name` handles things correctly:
- `/` and `\` separators 
- paths with or without trailing separators
- root paths like `/` or `C:\`
- other edge cases like device paths (`\\?\.\` etc) and `\\server\share` UNC paths
- probably other edge cases I haven't even thought of 😅 

The drawback is that we have to allocate the `DirectoryInfo` object and its constructor can also allocate some strings internally. But since this code only runs once, I think it's not a terrible trade-off.

tracer/src/Datadog.Trace/ProcessTags.cs
lucaspimentel: No. These were using the old `string.TrimEnd(params char[])` overload which has been in .NET for a long time (no need to vendor). As written, each of the two calls allocate a `char[1]` (because of the `params`). I was only hitting the bug when I tried to refactor from
```csharp
string TrimEnd(params char[])
```
to the vendored
```csharp
ReadOnlySpan<char> TrimEnd(this ReadOnlySpan<char> span, ReadOnlySpan<char> trimChars)
```
to avoid a few heap allocations.

(edit: formatting)

lucaspimentel:
I wanted to give some context on this PR since I did 180 from the original goal 😅 

It all started when I saw the `char[]` heap allocations from `TrimEnd(params char[])` and the potential intermediate trimmed strings. I thought these could be easily avoided. As I was refactoring that code, I noticed there were no unit tests, so I started adding some tests. That's when I found about some edge cases we were not handling well, so the code got more and more complicated to handle those cases. Then the tests started hanging and I found the bug in our vendored/decompiled version of the `TrimEnd(ReadOnlySpan<char>)`.

After going down that rabbit whole for a while, I realized that handling paths is _complex_, and I was too focused on reducing just a few small allocations that only happen once. (Yes, every bit counts, but sometimes you have to make trade-offs.) I pivoted the PR to focus less on the allocations (the original goal) and focused instead on a more correct handling of paths by using `DirectoryInfo` :stuck_out_tongue:

tracer/src/Datadog.Trace/ProcessTags.cs
bouwkast: Ahhhh I see thanks!

tracer/src/Datadog.Trace/ProcessTags.cs
bouwkast: Nice sounds like a lot easier to handle with `DirectoryInfo` 👍

tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -21,8 +21,9 @@ internal static class ProcessTags
     public const string EntrypointWorkdir = "entrypoint.workdir";
 
     // two views on the same data
-    public static readonly IReadOnlyCollection<string> TagsList = GetTagsList();
-    public static readonly string SerializedTags = GetSerializedTagsFromList(TagsList);
+    public static IReadOnlyCollection<string> TagsList => field ??= GetTagsList();
```
andrewlock, line 24: Is there a reason not to make these readonly fields any more? 🤔

lucaspimentel: In the previous code, they could be `readonly` because they were initialized in the (implicit) static ctor. I changed them to be more lazy (initialize on first acces) because the eager initialization was causing issues with some of the unit tests I was adding.

We don't need this anymore after further refactoring and I can change it back if you want (I personally don't like doing this much work in a static ctor, but level of care is low).

PR #8106, opened 2026-01-27
[Process Tags] Explicit whether the service name is the default one

pierotibou:
## Summary of changes

Implements [Andrea's specification](https://docs.google.com/document/d/1c47iSTWxIOHMHfZTF2nT9xfyQaIBP9KJvI9sRn5SvpM/edit?tab=t.0#heading=h.uoivahvmuvbg). 
TLDR: We need to send in process tags whether the service used is the one set by the user or the default one.
It was deemed acceptable not to handle the case where the service is set at the scope level when creating spans, so we focus on service set by the user through config. 

## Reason for change
Service renaming shouldn't rename a service if it's the one set by the customer but the backend doesn't know that today

## Implementation details
- As the service can be changed in the settings, I moved the process tags in the `MutableSettings`.  As a side effect it simplifies part of the code as we don't have to pass service tags anymore as they're in Settings. 
- The ProcessTags object is null when the PropagateServiceTag config key is false. We rely in the rest of the code on whether the tags are null or not.
- Process tags are still lazily initialized. The `ProcessTags` class is merely a container of the information needed when you need to serialize them. 

## Test coverage
Utests mainly.
There are system tests to update once merged: https://github.com/DataDog/system-tests/pull/6204

## Other details
AI generated and polished by hand (and repolished again :p) 

APMLP-1000

### DIFF
--- a/tracer/src/Datadog.Trace/Agent/DiscoveryService/AgentConfiguration.cs
+++ b/tracer/src/Datadog.Trace/Agent/DiscoveryService/AgentConfiguration.cs
@@ -21,6 +21,7 @@ internal sealed record AgentConfiguration
         string? eventPlatformProxyEndpoint,
         string? telemetryProxyEndpoint,
         string? tracerFlareEndpoint,
+        string? containerTagsHash,
         bool clientDropP0,
         bool spanMetaStructs,
         bool? spanEvents)
@@ -36,6 +37,7 @@ internal sealed record AgentConfiguration
         EventPlatformProxyEndpoint = eventPlatformProxyEndpoint;
         TelemetryProxyEndpoint = telemetryProxyEndpoint;
         TracerFlareEndpoint = tracerFlareEndpoint;
+        ContainerTagsHash = containerTagsHash;
         ClientDropP0s = clientDropP0;
         SpanMetaStructs = spanMetaStructs;
         SpanEvents = spanEvents ?? false;
@@ -74,6 +76,8 @@ internal sealed record AgentConfiguration
     public string? TelemetryProxyEndpoint { get; }
 
     public string? TracerFlareEndpoint { get; }
+
+    public string? ContainerTagsHash { get; }
 
     public bool ClientDropP0s { get; }
 
--- a/tracer/src/Datadog.Trace/Agent/DiscoveryService/DiscoveryService.cs
+++ b/tracer/src/Datadog.Trace/Agent/DiscoveryService/DiscoveryService.cs
@@ -429,6 +429,7 @@ namespace Datadog.Trace.Agent.DiscoveryService
                 eventPlatformProxyEndpoint: eventPlatformProxyEndpoint,
                 telemetryProxyEndpoint: telemetryProxyEndpoint,
                 tracerFlareEndpoint: tracerFlareEndpoint,
+                containerTagsHash: _containerMetadata.ContainerTagsHash, // either the value just received, or the one we stored before (prevents overriding with null)
                 clientDropP0: clientDropP0,
                 spanMetaStructs: spanMetaStructs,
                 spanEvents: spanEvents);
--- a/tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
+++ b/tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
@@ -15,6 +15,7 @@ using Datadog.Trace.DataStreamsMonitoring.Aggregation;
 using Datadog.Trace.DataStreamsMonitoring.Hashes;
 using Datadog.Trace.Headers;
 using Datadog.Trace.Logging;
+using Datadog.Trace.PlatformHelpers;
 using Datadog.Trace.Vendors.Serilog.Events;
 
 namespace Datadog.Trace.DataStreamsMonitoring;
@@ -26,46 +27,86 @@ internal sealed class DataStreamsManager
 {
     private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor<DataStreamsManager>();
     private static readonly AsyncLocal<PathwayContext?> LastConsumePathway = new(); // saves the context on consume checkpointing only
+    private readonly object _nodeHashUpdateLock = new();
     private readonly ConcurrentDictionary<string, RateLimiter> _schemaRateLimiters = new();
     private readonly IDisposable _updateSubscription;
     private readonly bool _isLegacyDsmHeadersEnabled;
+    private readonly IDiscoveryService _discoveryService; // only saved to be able to unsubscribe
     private long _nodeHashBase; // note that this actually represents a `ulong` that we have done an unsafe cast for
+    private MutableSettings _previousMutableSettings;
+    private string? _previousContainerTagsHash;
     private bool _isEnabled;
     private bool _isInDefaultState;
     private IDataStreamsWriter? _writer;
 
     public DataStreamsManager(
         TracerSettings tracerSettings,
-        IDataStreamsWriter? writer)
+        IDataStreamsWriter? writer,
+        IDiscoveryService discoveryService)
     {
-        UpdateNodeHash(tracerSettings.Manager.InitialMutableSettings);
         _isEnabled = writer is not null;
         _isLegacyDsmHeadersEnabled = tracerSettings.IsDataStreamsLegacyHeadersEnabled;
         _writer = writer;
+        _discoveryService = discoveryService;
         _isInDefaultState = tracerSettings.IsDataStreamsMonitoringInDefaultState;
-        _updateSubscription = tracerSettings.Manager.SubscribeToChanges(updates =>
-        {
-            if (updates.UpdatedMutable is { } updated)
-            {
-                UpdateNodeHash(updated);
-            }
-        });
 
-        void UpdateNodeHash(MutableSettings settings)
+        _previousMutableSettings = tracerSettings.Manager.InitialMutableSettings;
+        // even though the value will probably get updated by a callback when subscriptions happen just after,
+        // we still need to initialize it to a value from initial settings in case no callback fire
+        UpdateNodeHash(_previousMutableSettings, containerTagsHash: null);
+        // subscribing to changes calls the callback immediately if a value is present
+        discoveryService.SubscribeToChanges(UpdateHashWithContainerTags);
+        _updateSubscription = tracerSettings.Manager.SubscribeToChanges(UpdateHashWithNewSettings);
+    }
+
+    public bool IsEnabled
+    {
+        get => Volatile.Read(ref _isEnabled);
+    }
+
+    public bool IsInDefaultState
+    {
+        get => Volatile.Read(ref _isInDefaultState);
+    }
+
+    /// <summary> Callback for AgentConfiguration updates </summary>
+    private void UpdateHashWithContainerTags(AgentConfiguration conf)
+    {
+        lock (_nodeHashUpdateLock)
         {
-            // We don't yet support primary tag in .NET yet
-            var value = HashHelper.CalculateNodeHashBase(settings.DefaultServiceName, settings.Environment, primaryTag: null, settings.ProcessTags?.SerializedTags);
-            // Working around the fact we can't do Interlocked.Exchange with the struct
-            // and also that we can't do Interlocked.Exchange with a ulong in < .NET 5
-            Interlocked.Exchange(
-                ref _nodeHashBase,
-                unchecked((long)value.Value)); // reinterpret as a long
+            if (conf.ContainerTagsHash == _previousContainerTagsHash)
+            {
+                return;
+            }
+
+            UpdateNodeHash(_previousMutableSettings, conf.ContainerTagsHash);
+            _previousContainerTagsHash = conf.ContainerTagsHash;
         }
     }
 
-    public bool IsEnabled => Volatile.Read(ref _isEnabled);
+    /// <summary> Callback for MutableSettings updates </summary>
+    private void UpdateHashWithNewSettings(TracerSettings.SettingsManager.SettingChanges updates)
+    {
+        if (updates.UpdatedMutable is { } updated)
+        {
+            lock (_nodeHashUpdateLock)
+            {
+                UpdateNodeHash(updated, _previousContainerTagsHash);
+                _previousMutableSettings = updated;
+            }
+        }
+    }
 
-    public bool IsInDefaultState => Volatile.Read(ref _isInDefaultState);
+    private void UpdateNodeHash(MutableSettings settings, string? containerTagsHash)
+    {
+        // We don't yet support primary tag in .NET yet
+        var value = HashHelper.CalculateNodeHashBase(settings.DefaultServiceName, settings.Environment, primaryTag: null, settings.ProcessTags?.SerializedTags, containerTagsHash);
+        // Working around the fact we can't do Interlocked.Exchange with the struct
+        // and also that we can't do Interlocked.Exchange with a ulong in < .NET 5
+        Interlocked.Exchange(
+            ref _nodeHashBase,
+            unchecked((long)value.Value)); // reinterpret as a long
+    }
 
     public static DataStreamsManager Create(
         TracerSettings settings,
@@ -76,12 +117,13 @@ internal sealed class DataStreamsManager
                          ? DataStreamsWriter.Create(settings, profilerSettings, discoveryService)
                          : null;
 
-        return new DataStreamsManager(settings, writer);
+        return new DataStreamsManager(settings, writer, discoveryService);
     }
 
     public async Task DisposeAsync()
     {
         _updateSubscription.Dispose();
+        _discoveryService.RemoveSubscription(UpdateHashWithContainerTags);
         Volatile.Write(ref _isEnabled, false);
         var writer = Interlocked.Exchange(ref _writer, null);
 
--- a/tracer/src/Datadog.Trace/DataStreamsMonitoring/Hashes/HashHelper.cs
+++ b/tracer/src/Datadog.Trace/DataStreamsMonitoring/Hashes/HashHelper.cs
@@ -19,7 +19,7 @@ internal static class HashHelper
     /// Calculates the base NodeHash for a service.
     /// This can be used to create a <see cref="NodeHash"/> by calling <see cref="CalculateNodeHash"/>
     /// </summary>
-    public static NodeHashBase CalculateNodeHashBase(string service, string? env, string? primaryTag, string? processTags)
+    public static NodeHashBase CalculateNodeHashBase(string service, string? env, string? primaryTag, string? processTags, string? containerTagsHash)
     {
         var hash = FnvHash64.GenerateHash(service, HashVersion);
         if (!StringUtil.IsNullOrEmpty(env))
@@ -35,6 +35,11 @@ internal static class HashHelper
         if (!StringUtil.IsNullOrEmpty(processTags))
         {
             hash = FnvHash64.GenerateHash(processTags, HashVersion, hash);
+            // container tags are only added if process tags are in use
+            if (!StringUtil.IsNullOrEmpty(containerTagsHash))
+            {
+                hash = FnvHash64.GenerateHash(containerTagsHash, HashVersion, hash);
+            }
         }
 
         return new NodeHashBase(hash);
--- a/tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.NetFramework.cs
+++ b/tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.NetFramework.cs
@@ -6,6 +6,7 @@
 #nullable enable
 
 using System.Threading;
+using Datadog.Trace.Logging;
 
 #if NETFRAMEWORK
 
@@ -16,6 +17,9 @@ namespace Datadog.Trace.PlatformHelpers;
 /// </summary>
 internal sealed class ContainerMetadata
 {
+    private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(ContainerMetadata));
+    private static bool _warnedOnSet; // to log only once
+
     public static readonly ContainerMetadata Instance = new();
 
     private ContainerMetadata()
@@ -28,14 +32,18 @@ internal sealed class ContainerMetadata
         // nothing to do, just to match the other version
     }
 
-    /// <summary>
-    /// Gets or sets the container tags hash received from the agent, used by DBM/DSM
-    /// This is set when we receive a value for it in an http response from the agent
-    /// </summary>
+    // always null in this implementation
     public string? ContainerTagsHash
     {
-        get => Volatile.Read(ref field);
-        set => Volatile.Write(ref field, value);
+        get => null;
+        set
+        {
+            if (!_warnedOnSet)
+            {
+                _warnedOnSet = true;
+                Log.Error("The code is trying to set the value '{Value}' to {Prop}, but this has no effect in .NET Framework.", value, nameof(ContainerTagsHash));
+            }
+        }
     }
 
     /// <summary>
--- a/tracer/test/Datadog.Trace.Tests/Agent/DiscoveryServiceMock.cs
+++ b/tracer/test/Datadog.Trace.Tests/Agent/DiscoveryServiceMock.cs
@@ -26,6 +26,7 @@ internal class DiscoveryServiceMock : IDiscoveryService
         string eventPlatformProxyEndpoint = "eventPlatformProxyEndpoint",
         string telemetryProxyEndpoint = "telemetryProxyEndpoint",
         string tracerFlareEndpoint = "tracerFlareEndpoint",
+        string containerTagsHash = "containerTagsHash",
         bool clientDropP0 = true,
         bool spanMetaStructs = true,
         bool spanEvents = true)
@@ -42,6 +43,7 @@ internal class DiscoveryServiceMock : IDiscoveryService
                 eventPlatformProxyEndpoint: eventPlatformProxyEndpoint,
                 telemetryProxyEndpoint: telemetryProxyEndpoint,
                 tracerFlareEndpoint: tracerFlareEndpoint,
+                containerTagsHash: containerTagsHash,
                 clientDropP0: clientDropP0,
                 spanMetaStructs: spanMetaStructs,
                 spanEvents: spanEvents));
--- a/tracer/test/Datadog.Trace.Tests/Agent/DiscoveryServiceTests.cs
+++ b/tracer/test/Datadog.Trace.Tests/Agent/DiscoveryServiceTests.cs
@@ -301,6 +301,7 @@ public class DiscoveryServiceTests
             eventPlatformProxyEndpoint: "eventPlatformProxyEndpoint",
             telemetryProxyEndpoint: "telemetryProxyEndpoint",
             tracerFlareEndpoint: "tracerFlareEndpoint",
+            containerTagsHash: "containerTagsHash",
             clientDropP0: false,
             spanMetaStructs: true,
             spanEvents: true);
@@ -318,6 +319,7 @@ public class DiscoveryServiceTests
             eventPlatformProxyEndpoint: "eventPlatformProxyEndpoint",
             telemetryProxyEndpoint: "telemetryProxyEndpoint",
             tracerFlareEndpoint: "tracerFlareEndpoint",
+            containerTagsHash: "containerTagsHash",
             clientDropP0: false,
             spanMetaStructs: true,
             spanEvents: true);
@@ -335,6 +337,7 @@ public class DiscoveryServiceTests
             eventPlatformProxyEndpoint: "eventPlatformProxyEndpoint",
             telemetryProxyEndpoint: "telemetryProxyEndpoint",
             tracerFlareEndpoint: "tracerFlareEndpoint",
+            containerTagsHash: "containerTagsHash",
             clientDropP0: false,
             spanMetaStructs: true,
             spanEvents: true);
@@ -391,6 +394,8 @@ public class DiscoveryServiceTests
         factory.RequestsSent.Count.Should().BeInRange(3, 6, "Should make between 3 and 6 retries in 13s");
     }
 
+#if !NETFRAMEWORK
+
     [Fact]
     public async Task ExtractsContainerTagsHashFromResponseHeader()
     {
@@ -414,6 +419,8 @@ public class DiscoveryServiceTests
 
         await ds.DisposeAsync();
     }
+
+#endif
 
     private string GetConfig(bool dropP0 = true, string version = null)
         => JsonConvert.SerializeObject(new MockTracerAgent.AgentConfiguration() { ClientDropP0s = dropP0, AgentVersion = version });
--- a/tracer/test/Datadog.Trace.Tests/Agent/StatsAggregatorTests.cs
+++ b/tracer/test/Datadog.Trace.Tests/Agent/StatsAggregatorTests.cs
@@ -457,6 +457,7 @@ namespace Datadog.Trace.Tests.Agent
                              eventPlatformProxyEndpoint: "eventPlatformProxyEndpoint",
                              telemetryProxyEndpoint: "telemetryProxyEndpoint",
                              tracerFlareEndpoint: "tracerFlareEndpoint",
+                             containerTagsHash: "containerTagsHash",
                              clientDropP0: true,
                              spanMetaStructs: true,
                              spanEvents: true));
--- a/tracer/test/Datadog.Trace.Tests/DataStreamsMonitoring/DataStreamsManagerTests.cs
+++ b/tracer/test/Datadog.Trace.Tests/DataStreamsMonitoring/DataStreamsManagerTests.cs
@@ -8,13 +8,16 @@ using System.Collections.Concurrent;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using Datadog.Trace.Agent.DiscoveryService;
 using Datadog.Trace.Configuration;
 using Datadog.Trace.DataStreamsMonitoring;
 using Datadog.Trace.DataStreamsMonitoring.Aggregation;
 using Datadog.Trace.DataStreamsMonitoring.Hashes;
 using Datadog.Trace.ExtensionMethods;
+using Datadog.Trace.PlatformHelpers;
 using Datadog.Trace.TestHelpers.FluentAssertionsExtensions;
 using FluentAssertions;
+using Moq;
 using Xunit;
 
 namespace Datadog.Trace.Tests.DataStreamsMonitoring;
@@ -164,7 +167,7 @@ public class DataStreamsManagerTests
         var context = dsm.SetCheckpoint(parentPathway: null, CheckpointKind.Consume, edgeTags, 100, 100);
         context.Should().NotBeNull();
 
-        var baseHash = HashHelper.CalculateNodeHashBase(service, env, primaryTag: null, processTags: null);
+        var baseHash = HashHelper.CalculateNodeHashBase(service, env, primaryTag: null, processTags: null, containerTagsHash: null);
         var nodeHash = HashHelper.CalculateNodeHash(baseHash, edgeTags);
         var hash = HashHelper.CalculatePathwayHash(nodeHash, parentHash: new PathwayHash(0));
 
@@ -183,7 +186,7 @@ public class DataStreamsManagerTests
         var context = dsm.SetCheckpoint(parent, CheckpointKind.Consume, edgeTags, 100, 100);
         context.Should().NotBeNull();
 
-        var baseHash = HashHelper.CalculateNodeHashBase(service, env, primaryTag: null, processTags: null);
+        var baseHash = HashHelper.CalculateNodeHashBase(service, env, primaryTag: null, processTags: null, containerTagsHash: null);
         var nodeHash = HashHelper.CalculateNodeHash(baseHash, edgeTags);
         var hash = HashHelper.CalculatePathwayHash(nodeHash, parentHash: parent.Hash);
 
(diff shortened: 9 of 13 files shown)


pierotibou:
@codex review

tracer/src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
```diff
@@ -612,7 +607,8 @@ private int WriteTags(ref byte[] bytes, int offset, in SpanModel model, ITagProc
             // Process tags will be sent only once per buffer/payload (one payload can contain many chunks from different traces)
             if (model.IsFirstSpanInChunk && model.TraceChunk.IsFirstChunkInPayload && model.TraceChunk.ShouldPropagateProcessTags)
             {
-                var processTagsRawBytes = _processTagsValueBytes.Value;
+                var processTagsRawBytes = MessagePackStringCache.GetProcessTagsBytes(model.TraceChunk.ProcessTags);
+
                 if (processTagsRawBytes.Length > 0)
                 {
                     count++;
```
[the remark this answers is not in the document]
pierotibou: Nice catch. Also shows, I must be missing a test.

tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -14,24 +14,53 @@
...
+        AddNormalizedTag(tags, EntrypointBasedir, GetLastPathSegment(AppContext.BaseDirectory));
+        AddNormalizedTag(tags, EntrypointName, GetEntryPointName());
         // workdir can be changed by the code, but we consider that capturing the value when this is called is good enough
-        tags.AddNormalizedTag(EntrypointWorkdir, GetLastPathSegment(Environment.CurrentDirectory));
+        AddNormalizedTag(tags, EntrypointWorkdir, GetLastPathSegment(Environment.CurrentDirectory));
+        // Either svc.user or svc.auto, never both
+        if (serviceNameUserDefined)
+        {
+            // svc.user is only added when the user explicitly set the service name
+            tags.Add($"{ServiceSetByUser}:1");
```
amarziali: @pierotibou can you replace with `true` instead of `1` ? (we discussed that recently and has been changed)

pierotibou: Sure, but please change the RFC then :). I changed from `true` to `1` because that's what is in the RFC .

amarziali: it's already changed normally

tracer/src/Datadog.Trace/Agent/MessagePack/MessagePackStringCache.cs
```diff
@@ -32,6 +32,9 @@ internal static class MessagePackStringCache
     [ThreadStatic]
     private static CachedBytes _service;
 
+    // Today, process tags may change only when the service name isn't the default one anymore
+    private static CachedBytes _processTags;
```
lucaspimentel, line 37: This needs `[ThreadStatic]` to avoid race conditions. The ones below don't need it before those values never changes. If the value never changes, they can be cached directly in `SpanMessagePackFormatter` and don't need to be in this file. These file was for values that _can_ change but _rarely_ do, so we cache only the lastest value.

I added that NOTE comment below, but it should really be a TODO for us to fix it someday 😅

pierotibou: I saw the TODO and wanted to do it alongside another change. But I can't promise an ETA :)

tracer/src/Datadog.Trace/Configuration/MutableSettings.cs
```diff
@@ -91,6 +92,7 @@ private MutableSettings(
         GitRepositoryUrl = gitRepositoryUrl;
         GitCommitSha = gitCommitSha;
         ErrorLog = errorLog;
+        ProcessTags = propagateProcessTags ? new ProcessTags(!string.IsNullOrWhiteSpace(serviceName), defaultServiceName) : null;
```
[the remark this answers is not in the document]
pierotibou: This doesn't initialize the tags anymore, they are lazy loaded in the ProcessTags class

tracer/src/Datadog.Trace/Agent/MessagePack/MessagePackStringCache.cs
```diff
@@ -32,6 +32,10 @@ internal static class MessagePackStringCache
     [ThreadStatic]
     private static CachedBytes _service;
 
+    // Today, process tags may change only when the service name isn't the default one anymore
```
vandonr: I don't think that's true ? 🤔 
Currently, process tags are constant in dotnet, but we may need to change this in the future, when we introduce new values like some stuff from IIS that we can only access after init.
Anyway, I don't see how process tags interact with default service name. They can be used to build a service in the backend, but the tracer wouldn't know about it, from here it's only a bunch of values that we write in payloads.
Maybe I misunderstood your logic when you wrote this comment ?

tracer/src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
```diff
@@ -68,12 +68,7 @@ internal sealed class SpanMessagePackFormatter : IMessagePackFormatter<TraceChun
         private readonly byte[] _runtimeIdNameBytes = StringEncoding.UTF8.GetBytes(Trace.Tags.RuntimeId);
         private readonly byte[] _runtimeIdValueBytes = StringEncoding.UTF8.GetBytes(Tracer.RuntimeId);
 
-        // using a Lazy here to make sure we don't compute the value of the process tags too early in the life of the app,
-        // some values may need a bit of time to be accessible.
-        // With this construct, it should be queried after the first span(s) get closed, which should be late enough.
-        private readonly Lazy<byte[]> _processTagsValueBytes = new(() => StringEncoding.UTF8.GetBytes(ProcessTags.SerializedTags));
```
vandonr, line 75: good thing to remove this, because I wrote this at the very beginning of working on it, and later on we needed to access the tags much earlier, making this useless

tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -14,25 +14,77 @@
...
+
+    private readonly bool _serviceNameUserDefined;
+    private readonly string _autoServiceName;
+    private List<string>? _tagsList;
+    private string? _serializedTags;
+
+    // ProcessTags captures EntryAssemblyLocator.GetEntryAssembly() and Environment.CurrentDirectory
+    // If initialization happens before the entry assembly is available (common in ASP.NET/IIS or some test hosts),
+    // entrypoint.name/entrypoint.workdir will be empty
+    // We intentionally do not compute tags at construction and init them on first access only
```
vandonr, line 31: I'm not sure this is very useful, given how some components will query process tags super early (I don't remember who's the worse culprit, but I started trying to delay access, only to realize later that it was _peine perdue_)

tracer/src/Datadog.Trace/Agent/ClientStatsPayload.cs
```diff
@@ -11,22 +12,29 @@ namespace Datadog.Trace.Agent
...
+        private string? _processTags = GetProcessTags(settings);
         private long _sequence;
 
-        public string HostName { get; init; }
+        public string? HostName { get; init; }
 
         public AppSettings Details => _settings;
 
-        public string ProcessTags { get; init; }
+        public string? ProcessTags => _processTags;
```
vandonr: do we really need a public property with a private backing field here ?
I feel like we should be able to do with a private setter or something ?

tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -14,25 +14,77 @@
...
         // workdir can be changed by the code, but we consider that capturing the value when this is called is good enough
-        tags.AddNormalizedTag(EntrypointWorkdir, GetLastPathSegment(Environment.CurrentDirectory));
+        AddNormalizedTag(tags, EntrypointWorkdir, GetLastPathSegment(Environment.CurrentDirectory));
+
+        // Either svc.user or svc.auto, never both
+        // svc.user is only added when the user explicitly set the service name
+        // svc.auto contains the automatically determined service name when user didn't set it
+        if (serviceNameUserDefined)
+        {
+            tags.Add($"{ServiceSetByUser}:true");
```
vandonr: since we only ever write this tag with a value of true, we might as well put the `key:value` in a constant, don't you think ?
Or maybe the compiler recognizes this, idk :p

tracer/src/Datadog.Trace/Agent/MessagePack/MessagePackStringCache.cs
pierotibou: I'll rewrite the comment to: process tags can change when the service name is changed in configuration which shouldn't happen often.

> I don't think that's true ? 🤔
it is true today right?  process tags won't change unless the service name changes with a new config.

tracer/src/Datadog.Trace/Agent/ClientStatsPayload.cs
pierotibou: It's needed for the `Interlocked.Exchange`. 
But I'll change 2 things now that I see this code:
- I'll reference `ProcessTags` instead of the string. That way, I'll defer the initialization of the string to the actual first serialization rather than at construction time
- I'll add the processTags in the AppSettings record. Not really an appSetting but it's used all in the same place so I believe it makes more sense and make the code more readable.

tracer/src/Datadog.Trace/ProcessTags.cs
pierotibou: Yes the compiler does it all, but you're right that it will be easier to read with the whole value

tracer/src/Datadog.Trace/ProcessTags.cs
pierotibou: :hmm: I thought you had actually initialized this only at serialization time everywhere. But indeed, I see some use cases where we do the serialization at construction time (statsd for instance). So it is mostly useless today.

tracer/src/Datadog.Trace/Agent/MessagePack/MessagePackStringCache.cs
vandonr: ah ok, yes that makes sense. In the future, they may change a little bit after the app has started too, so it's helpful to already have a mechanism where we can update them.

tracer/src/Datadog.Trace/ProcessTags.cs
vandonr: yes it was a relic from ancient times

tracer/src/Datadog.Trace/Agent/MessagePack/MessagePackStringCache.cs
pierotibou: Done in 5d9a3665a59b3d9163618f0d7315b896c9148d0e

tracer/src/Datadog.Trace/Agent/ClientStatsPayload.cs
pierotibou: done in 5d9a3665a59b3d9163618f0d7315b896c9148d0e

tracer/src/Datadog.Trace/ProcessTags.cs
pierotibou: Done in 5d9a3665a59b3d9163618f0d7315b896c9148d0e

tracer/src/Datadog.Trace/Agent/MessagePack/MessagePackStringCache.cs
andrewlock: What race conditions are we actually thinking about here? 🤔 Don't we have a single-threaded span messagepack writer? 

I'm just thinking about the fact that with `[ThreadStatic]` we're effectively creating `n` copies of the same array if it's called by multiple threads.

tracer/src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
```diff
@@ -610,14 +605,15 @@ private int WriteTags(ref byte[] bytes, int offset, in SpanModel model, ITagProc
             }
 
             // Process tags will be sent only once per buffer/payload (one payload can contain many chunks from different traces)
-            if (model.IsFirstSpanInChunk && model.TraceChunk.IsFirstChunkInPayload && model.TraceChunk.ShouldPropagateProcessTags)
+            if (model.IsFirstSpanInChunk && model.TraceChunk.IsFirstChunkInPayload && model.TraceChunk.ProcessTags is not null)
             {
-                var processTagsRawBytes = _processTagsValueBytes.Value;
-                if (processTagsRawBytes.Length > 0)
+                var processTagsRawBytes = MessagePackStringCache.GetProcessTagsBytes(model.TraceChunk.ProcessTags?.SerializedTags);
```
andrewlock: nit: dammit isn't necessary, as we know it's non-null at this point

```suggestion
                var processTagsRawBytes = MessagePackStringCache.GetProcessTagsBytes(model.TraceChunk.ProcessTags.SerializedTags);
```

tracer/src/Datadog.Trace/Agent/ClientStatsPayload.cs
```diff
@@ -13,20 +14,20 @@ internal sealed class ClientStatsPayload(MutableSettings settings)
...
 
-        public string ProcessTags { get; init; }
-
         public long GetSequenceNumber() => Interlocked.Increment(ref _sequence);
 
         public void UpdateDetails(MutableSettings settings)
-            => Interlocked.Exchange(ref _settings, CreateSettings(settings));
+        {
+            Interlocked.Exchange(ref _settings, CreateSettings(settings));
+        }
```
andrewlock: How very dare you 😛 Trying to bump your lines of code committed? 😉

tracer/src/Datadog.Trace/Agent/StatsBuffer.cs
```diff
@@ -70,17 +72,16 @@ public void Serialize(Stream stream, long bucketDuration)
...
 
-            var details = _header.Details;
             MessagePackBinary.WriteString(stream, "Env");
             MessagePackBinary.WriteString(stream, details.Environment ?? string.Empty);
 
             MessagePackBinary.WriteString(stream, "Version");
             MessagePackBinary.WriteString(stream, details.Version ?? string.Empty);
 
-            if (!string.IsNullOrEmpty(_header.ProcessTags))
+            if (!string.IsNullOrEmpty(details.ProcessTags?.SerializedTags))
```
andrewlock: nit: if you use `StringUtils` here, it stops the compiler complaining about the null in the line below
```suggestion
            if (!StringUtils.IsNullOrEmpty(details.ProcessTags?.SerializedTags))
```

tracer/src/Datadog.Trace/Agent/StatsBuffer.cs
```diff
@@ -70,17 +72,16 @@ public void Serialize(Stream stream, long bucketDuration)
...
             {
                 MessagePackBinary.WriteString(stream, "ProcessTags");
-                MessagePackBinary.WriteString(stream, _header.ProcessTags);
+                MessagePackBinary.WriteString(stream, details.ProcessTags?.SerializedTags);
```
andrewlock:
```suggestion
                MessagePackBinary.WriteString(stream, details.ProcessTags.SerializedTags);
```

tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -14,25 +14,77 @@
...
+            }
 
-    public static string SerializedTags => field ??= string.Join(",", TagsList);
+            return _tagsList!;
+        }
+    }
 
-    private static List<string> GetTagsList()
+    public string SerializedTags
+    {
```
andrewlock: AFAICT, this is an uglier way of doing the same thing? 🤔

tracer/src/Datadog.Trace/Configuration/MutableSettings.cs
```diff
@@ -733,7 +742,9 @@ public static MutableSettings CreateUpdatedMutableSettings(
             serviceNameMappings: serviceNameMappings,
             gitRepositoryUrl: gitRepositoryUrl,
             gitCommitSha: gitCommitSha,
-            errorLog: errorLog);
+            errorLog: errorLog,
+            propagateProcessTags: tracerSettings.PropagateProcessTags,
+            fallbackServiceName: tracerSettings.FallbackApplicationName);
```
andrewlock: using `tracerSettings.FallbackApplicationName` is expensive where it doesn't need to be called, and I don't believe it's necessary here

```suggestion
            propagateProcessTags: tracerSettings.PropagateProcessTags);
```

tracer/src/Datadog.Trace/Configuration/MutableSettings.cs
```diff
@@ -1064,7 +1075,9 @@ public static MutableSettings CreateInitialMutableSettings(
...
```
andrewlock:
```suggestion
            propagateProcessTags: tracerSettings.PropagateProcessTags);
```

tracer/src/Datadog.Trace/Configuration/MutableSettings.cs
```diff
@@ -61,7 +61,9 @@ private MutableSettings(
         ReadOnlyDictionary<string, string> serviceNameMappings,
         string? gitRepositoryUrl,
         string? gitCommitSha,
-        OverrideErrorLog errorLog)
+        OverrideErrorLog errorLog,
+        bool propagateProcessTags,
+        string fallbackServiceName)
```
andrewlock: using `tracerSettings.FallbackApplicationName` is expensive where it doesn't need to be called, and I don't believe it's necessary here
```suggestion
        bool propagateProcessTags)
```

tracer/src/Datadog.Trace/Configuration/MutableSettings.cs
```diff
@@ -91,6 +93,7 @@ private MutableSettings(
         GitRepositoryUrl = gitRepositoryUrl;
         GitCommitSha = gitCommitSha;
         ErrorLog = errorLog;
+        ProcessTags = propagateProcessTags ? new ProcessTags(!string.IsNullOrWhiteSpace(serviceName), fallbackServiceName) : null;
```
andrewlock:
```suggestion
        ProcessTags = propagateProcessTags
          ? new ProcessTags(serviceNameUserDefined: !string.IsNullOrEmpty(serviceName), defaultServiceName)
          : null;
```

tracer/src/Datadog.Trace/DataStreamsMonitoring/Aggregation/DataStreamsMessagePackFormatter.cs
```diff
@@ -83,6 +84,8 @@ void UpdateSettings(MutableSettings settings)
 
                 var service = StringUtil.IsNullOrEmpty(settings.DefaultServiceName) ? [] : StringEncoding.UTF8.GetBytes(settings.DefaultServiceName);
                 Interlocked.Exchange(ref _serviceValueBytes!, service);
+
+                Interlocked.Exchange(ref _processTags, settings.ProcessTags);
```
andrewlock, line 88: Is there a reason to _not_ do the `StringEncoding.UTF8.GetBytes()` and store that, like we are for the other values? I understand the "delay" reasoning, but we've debunked that already haven't we? 😅

tracer/src/Datadog.Trace/DataStreamsMonitoring/Aggregation/DataStreamsMessagePackFormatter.cs
```diff
@@ -219,8 +222,8 @@ public int Serialize(Stream stream, long bucketDurationNs, List<SerializableStat
             if (withProcessTags)
             {
                 bytesWritten += MessagePackBinary.WriteStringBytes(stream, _processTagsBytes);
-                bytesWritten += MessagePackBinary.WriteArrayHeader(stream, ProcessTags.TagsList.Count);
-                foreach (var tag in ProcessTags.TagsList)
+                bytesWritten += MessagePackBinary.WriteArrayHeader(stream, _processTags!.TagsList.Count);
```
andrewlock: tiny nit: I would extract this as a variable to avoid the multiple access (each access we have to check if it's null before returning it)

tracer/src/Datadog.Trace/Debugger/Snapshots/DebuggerSnapshotCreator.cs
```diff
@@ -838,6 +838,7 @@ internal string FinalizeMethodSnapshot<T>(string probeId, int probeVersion, ref
         internal void FinalizeSnapshot(string methodName, string typeFullName, string probeFilePath)
         {
             var activeScope = Tracer.Instance.InternalActiveScope;
+            var processTags = Tracer.Instance.Settings.Manager.InitialMutableSettings.ProcessTags?.SerializedTags;
```
andrewlock: This is _likely_ wrong, seeing as it's accessing `InitialMutableSettings`, but I'll defer to debugger folks as they have entirely different patterns for managing this currently

tracer/src/Datadog.Trace/Telemetry/Collectors/ApplicationTelemetryCollector.cs
andrewlock: This whole part should likely move into `RecordMutableSettings`? 🤔

tracer/test/Datadog.Trace.Tests/Agent/MessagePack/SpanMessagePackFormatterTests.cs
```diff
@@ -509,6 +509,64 @@ public async Task LastParentId_Tag()
...
+    public async Task ProcessTags_Serialization(bool propagateProcessTags)
+    {
+        var mockApi = new MockApi();
+        var settings = TracerSettings.Create(new()
+        {
+            { ConfigurationKeys.PropagateProcessTags, propagateProcessTags.ToString() },
+            { ConfigurationKeys.ServiceName, "test-service" }
+        });
+        var agentWriter = new AgentWriter(mockApi, statsAggregator: null, statsd: TestStatsdManager.NoOp, automaticFlush: false);
+        await using var tracer = TracerHelper.Create(settings, agentWriter, sampler: null, scopeManager: null, statsd: null, NullTelemetryController.Instance, NullDiscoveryService.Instance);
```
andrewlock, line 639: note to self: the `TracerHelper.Create` helpers should probably default to everything being noop instances, unless you explicitly need the "real" values 🤔

tracer/test/Datadog.Trace.Tests/Agent/StatsBufferTests.cs
```diff
@@ -25,22 +26,29 @@ public void KeyEquality()
...
+            IConfigurationSource source = new NameValueConfigurationSource(collection);
 
             var settings = MutableSettings.CreateForTesting(
-                new(),
+                new(source),
                 new()
                 {
                     { ConfigurationKeys.Environment, "Env" },
                     { ConfigurationKeys.ServiceVersion, "v99.99" },
                 });
```
andrewlock, line 49: not sure why all the complexity, what was there should be good enough? 😄 
```suggestion
            var settings = MutableSettings.CreateForTesting(
                new(),
                new()
                {
                    { ConfigurationKeys.Environment, "Env" },
                    { ConfigurationKeys.ServiceVersion, "v99.99" },
                    { ConfigurationKeys.PropagateProcessTags, propagateProcessTags.ToString() },
                });
```

tracer/test/Datadog.Trace.Tests/Agent/StatsBufferTests.cs
```diff
@@ -73,6 +80,15 @@ public void Serialization()
             result.Service.Should().BeNull();
             result.Stats.Should().HaveCount(1);
 
+            if (propagateProcessTags)
+            {
+                result.ProcessTags.Should().Contain("svc.auto");
```
andrewlock: Should we add a test for the alternative too, where an explicit service name is set?

Could add `setServiceName` as an additional parameter in this theory test

tracer/test/Datadog.Trace.Tests/Configuration/MutableSettingsTests.cs
```diff
@@ -641,6 +641,25 @@ public void LogsInjectionEnabled(string value, bool expected)
...
+            var mutable = GetMutableSettings(source, settings);
+
+            if (propagateTags)
+            {
+                mutable.ProcessTags?.Should().NotBeNull();
+            }
+            else
+            {
+                mutable.ProcessTags?.Should().NotBeNull();
+            }
```
andrewlock, line 660: That doesn't look right 😅

tracer/src/Datadog.Trace/DataStreamsMonitoring/Aggregation/DataStreamsMessagePackFormatter.cs
pierotibou: I can, but it felt better to be consistent. I was more considering that if we want to go completely over it and remove any lazy init, I'd do it then.

tracer/src/Datadog.Trace/Agent/ClientStatsPayload.cs
pierotibou: I'm old man. This is how we were supposed to write code 🧌 :p

tracer/src/Datadog.Trace/DataStreamsMonitoring/Aggregation/DataStreamsMessagePackFormatter.cs
andrewlock: Ah I see. I guess it depends on your definition of consistency. It's inconsistent with the other usages in `DataStreamsMessagePackFormatter.cs` 😄 Not passionate about it

tracer/src/Datadog.Trace/DogStatsd/StatsdFactory.cs
```diff
@@ -22,11 +22,11 @@ internal static IDogStatsd CreateDogStatsdClient(
         MutableSettings settings,
         ExporterSettings exporter,
         bool includeDefaultTags,
-        IReadOnlyCollection<string> processTags,
         string? prefix = null)
     {
         var customTagCount = settings.GlobalTags.Count;
-        var tagsCount = (includeDefaultTags ? 5 + customTagCount : 0) + processTags.Count;
+        var processTags = settings.ProcessTags?.TagsList;
+        var tagsCount = (includeDefaultTags ? 5 + customTagCount : 0) + processTags?.Count ?? 0;
```
NachoEchevarria:
```suggestion
        var tagsCount = (includeDefaultTags ? 5 + customTagCount : 0) + (processTags?.Count ?? 0);
```
?? has lower precedence than + making the whole expression 0 when processTags is null, regardless of previous values.

tracer/test/Datadog.Trace.Tests/Configuration/MutableSettingsTests.cs
```diff
@@ -641,6 +641,25 @@ public void LogsInjectionEnabled(string value, bool expected)
...
+            var settings = new TracerSettings(source);
+            var mutable = GetMutableSettings(source, settings);
+
+            if (propagateTags)
+            {
+                mutable.ProcessTags?.Should().NotBeNull();
+            }
+            else
+            {
+                mutable.ProcessTags?.Should().NotBeNull();
```
NachoEchevarria: Also worth to note that because of the ? if mutable.ProcessTags is null, we would skip the assert statement

tracer/src/Datadog.Trace/Configuration/MutableSettings.cs
pierotibou: This is what I had done initially. But I believe it's wrong because [defaultServiceName actually is serviceName when it's set](https://github.com/DataDog/dd-trace-dotnet/blob/48dd38723764d678f69cc6e34eacbbe144f14630/tracer/src/Datadog.Trace/Configuration/MutableSettings.cs#L731).  So given it's expensive and we'd rather avoid doing it I can avoid computing it when we don't need it, thus moving the if in ProcessTags elsewhere. 
meh

tracer/src/Datadog.Trace/Agent/MessagePack/MessagePackStringCache.cs
lucaspimentel: > Don't we have a single-threaded span messagepack writer?

errr... it that's true, you make an excellent point and all of this can be much simpler 😅

tracer/src/Datadog.Trace/Agent/ClientStatsPayload.cs
lucaspimentel: I don't care which syntax you use initially, but changing it from one to the other just adds noise to PRs, imo 😝 

And you didn't change `CreateSettings()` below. Where's the consistency?!

lucaspimentel (changes requested):
There are a few blocking comments, so I'm going to "request changes" to prevent this from getting merged accidentally. (sorry)

lucaspimentel:
@codex review

tracer/src/Datadog.Trace/Configuration/MutableSettings.cs
andrewlock: > But I believe it's wrong because [defaultServiceName actually is serviceName when it's set](https://github.com/DataDog/dd-trace-dotnet/blob/48dd38723764d678f69cc6e34eacbbe144f14630/tracer/src/Datadog.Trace/Configuration/MutableSettings.cs#L731)

Yes, but if `serviceName` is set, then you don't care what `defaultServiceName` is for your purposes, right? 😄 In other words: you only care about `defaultServiceName` when `serviceName` is `null`, and in that scenario, it always contains the "correct" value?

tracer/src/Datadog.Trace/Configuration/MutableSettings.cs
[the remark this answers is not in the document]
andrewlock: > CI Visibility can auto-populate it from repository metadata when DD_SERVICE is unset

_Ouch_ 😅

tracer/test/Datadog.Trace.Tests/Agent/StatsBufferTests.cs
pierotibou: So this was done on purpose. aspropagateprocessTags is in TracerSettings, so it has to be in a different source.

tracer/src/Datadog.Trace/Configuration/MutableSettings.cs
pierotibou: Good point. I believe that originally the RFC (or I understood from the RFC 🤷 ) was asking to send both data all the time. But I didn't think of reverting when I changed this in the implem indeed. Thanks again.
Fixed in  1ee932a804d866352e3cabd6efdb0a47c33d69b2

tracer/src/Datadog.Trace/Configuration/MutableSettings.cs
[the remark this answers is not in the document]
pierotibou: I think CI VIz isn't in scope. Checking

tracer/src/Datadog.Trace/DataStreamsMonitoring/Aggregation/DataStreamsMessagePackFormatter.cs
pierotibou: Done in 0f37f1da474b898ac460be13e966d9a4e438b788

tracer/src/Datadog.Trace/Debugger/Snapshots/DebuggerSnapshotCreator.cs
pierotibou: Fixed in 407f5caeae5cff48af5989dd40ca83b8bbde84d3

tracer/src/Datadog.Trace/DogStatsd/StatsdFactory.cs
pierotibou: Thanks. Done in 0f37f1da474b898ac460be13e966d9a4e438b788

tracer/src/Datadog.Trace/ProcessTags.cs
pierotibou: Indeed, I missed that when rebasing on top of Lucas change.
Fixed in 1ee932a804d866352e3cabd6efdb0a47c33d69b2

tracer/test/Datadog.Trace.Tests/Agent/StatsBufferTests.cs
pierotibou: Added in ed40741ed315321dd017e3907cf11735ff99fd0d

tracer/test/Datadog.Trace.Tests/Configuration/MutableSettingsTests.cs
pierotibou: :oops: Thanks. Fixed in cff605ccea5d9356b9542fa4a497fc644a613069

tracer/src/Datadog.Trace/Agent/ClientStatsPayload.cs
pierotibou: Don't worry Lucas, it was just a joke. I didn't even realize when I did the change.
Fixed in 1ee932a804d866352e3cabd6efdb0a47c33d69b2

tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -14,25 +14,55 @@
...
     public const string EntrypointName = "entrypoint.name";
     public const string EntrypointBasedir = "entrypoint.basedir";
     public const string EntrypointWorkdir = "entrypoint.workdir";
+    public const string ServiceSetByUser = "svc.user:true";
+    public const string ServiceAuto = "svc.auto";
+
+    private readonly bool _serviceNameUserDefined;
+    private readonly string _autoServiceName;
+    private List<string>? _tagsList;
+    private string? _serializedTags;
```
NachoEchevarria: Both _tagsList and _serializedTags are declared and set in the Init method, but never used. TagsList =>and  SerializedTags => don't actually use them so they should probably be removed

tracer/src/Datadog.Trace/ProcessTags.cs
```diff
@@ -75,4 +105,10 @@ internal static string GetLastPathSegment(string directoryPath)
     {
         return EntryAssemblyLocator.GetEntryAssembly()?.EntryPoint?.DeclaringType?.FullName;
     }
+
+    private void InitTags()
```
NachoEchevarria: This method could also be removed

tracer/src/Datadog.Trace/Agent/StatsBuffer.cs
```diff
@@ -60,7 +59,9 @@ public void Reset()
         public void Serialize(Stream stream, long bucketDuration)
         {
             var count = 8;
-            if (!string.IsNullOrEmpty(_header.ProcessTags))
+            var details = _header.Details;
+
+            if (!string.IsNullOrEmpty(details.ProcessTags?.SerializedTags))
```
NachoEchevarria: We are checking by using string.IsNullOrEmpty and later with StringUtil.IsNullOrEmpty(details.ProcessTags?.SerializedTags), we should probably use StringUtil in both. 

Maybe also worth using for a single access:
var processTags = details.ProcessTags?.SerializedTags;
var hasProcessTags = !StringUtil.IsNullOrEmpty(processTags);

vandonr: StringUtil is useful for its nullability compatibility (compiler will know that a value is not null after this has been checked), otherwise there is no specific reason to use it

tracer/src/Datadog.Trace/Telemetry/Collectors/ApplicationTelemetryCollector.cs
pierotibou: Done in 6c33ace27d65122575f981340801a4837749f0e8. Sorry I had missed this one

tracer/src/Datadog.Trace/Configuration/MutableSettings.cs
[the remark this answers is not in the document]
pierotibou: Ci Viz is indeed not in scope and not implemented today, so we're good

tracer/src/Datadog.Trace/Agent/StatsBuffer.cs
pierotibou: Fixed in 6c33ace27d65122575f981340801a4837749f0e8 anyway. For sake of consistency as I was cleaning other things

tracer/src/Datadog.Trace/ProcessTags.cs
pierotibou: Fixed in 6c33ace27d65122575f981340801a4837749f0e8

tracer/src/Datadog.Trace/ProcessTags.cs
pierotibou: Fixed in 6c33ace27d65122575f981340801a4837749f0e8

tracer/src/Datadog.Trace/Configuration/MutableSettings.cs
```diff
@@ -261,6 +263,12 @@ private MutableSettings(
 
     internal OverrideErrorLog ErrorLog { get; }
 
+    /// <summary>
+    /// Gets the process tags instance including the service_name.user_defined tag.
+    /// Will be null if propagateProcessTags is null
+    /// </summary>
+    internal ProcessTags? ProcessTags { get; }
```
lucaspimentel, line 270: Should we add this to `Equals()` and `GetHashCode()`?

tracer/src/Datadog.Trace/Agent/StatsBuffer.cs
```diff
@@ -70,17 +71,16 @@ public void Serialize(Stream stream, long bucketDuration)
...
 
-            var details = _header.Details;
             MessagePackBinary.WriteString(stream, "Env");
             MessagePackBinary.WriteString(stream, details.Environment ?? string.Empty);
 
             MessagePackBinary.WriteString(stream, "Version");
             MessagePackBinary.WriteString(stream, details.Version ?? string.Empty);
 
-            if (!string.IsNullOrEmpty(_header.ProcessTags))
+            if (!StringUtil.IsNullOrEmpty(details.ProcessTags?.SerializedTags))
```
lucaspimentel: Minor nit, but you already check this in line 64 above so you could put that results into a local var. You also access `details.ProcessTags.SerializedTags` three times (lines 64, 80, and 83).

tracer/src/Datadog.Trace/Telemetry/Collectors/ApplicationTelemetryCollector.cs
lucaspimentel: If `tracerSettings.PropagateProcessTags` is `false` in line 64, `processTags` will be `null`, and this falls back to the previous `_applicationData?.ProcessTags`. Don't we want to pass `null` to telemetry if the feature is disabled?

vandonr: yeah since the setting is not supposed to change during the app lifetime, the fallback would be to null too, making it uselss

PR #8107, opened 2026-01-27
Check for labels in PRs.

NachoEchevarria:
## Summary of changes

This PR adds automated validation to ensure all pull requests have at least one label before merging. This eliminates the manual step of checking for unlabeled PRs during the release process.

## Reason for change

During releases, we currently need to manually:
1. Check if there are any PRs without labels in the release milestone
2. Add appropriate labels to any unlabeled PRs
3. Ensure proper categorization for release notes generation

This PR implements automated label validation that runs on every pull request:

1. **Auto-labeling**: Existing workflow automatically adds labels based on PR title and changed files
2. **Validation**: New validation step immediately checks that at least one label was applied
3. **Enforcement**: PRs without labels fail the GitHub Actions check, blocking merge


.github/workflows/auto_label_prs.yml
```diff
@@ -22,6 +22,12 @@ jobs:
 
       - name: "Add labels"
         run: ./tracer/build.sh AssignLabelsToPullRequest
+        env:
+          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+          PullRequestNumber: "${{ github.event.pull_request.number }}"
+
+      - name: "Validate PR has labels"
```
bouwkast, line 31: I think it may be better if this lives in its own workflow (that is required) with a trigger such as

```
on:
    pull_request:
      types: [labeled]
```

That way whenever someone changes a label it will run this automatically and then update.

Separate workflow would get us around having to run AssignLabelsToPullRequest which seems a bit heavy

And `[labeled]` from what I see [here ](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request)would correctly run the workflow when labels are added/removed

But this would be a great help when doing the release!

NachoEchevarria: Initially, I set it in a separate workflow. The main problem is about racing conditions with auto_label_prs.yml. This one can automatically add some tags after we fail the validation. For instance, if this check runs before auto_label_prs.yml, it could fail even if later auto_label_prs.yml adds a label. If we make it secuential, we have no racing issues.

About using types: [labeled], the labeled trigger would only fire when a label is added to a PR, right? If we have no labels, would it run (and fail)?

bouwkast: That makes sense for the race condition

So if we keep it in this workflow we could do something like

```
on:
    pull_request:
      types: [opened, synchronize, reopened, labeled]
```

Then we can filter like so
```
 - name: "Add labels"
          if: github.event.action != 'labeled'
```

And I think that this would get us there so that they work and don't have a race condition

NachoEchevarria: Yes, that would make sense. Thanks!

NachoEchevarria:
Thanks for the feedback!

PR #8111, opened 2026-01-28
Fix compilation issues under VS 2022

NachoEchevarria:
## Summary of changes

Adds `<NoWarn>NU1510</NoWarn>` to Datadog.Trace.Manual.csproj to suppress package pruning warnings when building in Visual Studio.
Removed <PackageReference Include="System.ValueTuple" Version="4.5.0" /> from Datadog.InstrumentedAssemblyGeneration since it does not provide any value and causes compilation problems.

Add a #pragma warning disable SYSLIB0014 to avoid compilation problems.

## Reason for change

With the .NET 10 SDK, Visual Studio shows NU1510 warnings ("PackageReference X will not be pruned. Consider removing this package from your dependencies, as it is likely unnecessary") that block compilation in src projects due to `TreatWarningsAsErrors=true`.

Additionally, `Datadog.Trace.csproj` shows SYSLIB0014 errors about obsolete `WebRequest` APIs in AWS Lambda integration code:
```
error SYSLIB0014: 'WebRequest.Create(string)' is obsolete: 'WebRequest, HttpWebRequest, ServicePoint, and WebClient are obsolete. Use HttpClient instead.'
```

These warnings appear specifically in **Visual Studio 2022 builds** but not in:
- Command-line builds via `tracer\build BuildTracerHome`
- Test projects (already suppressed in `tracer/test/Directory.Build.props`)

### DIFF
--- a/tracer/src/Datadog.Trace/Datadog.Trace.csproj
+++ b/tracer/src/Datadog.Trace/Datadog.Trace.csproj
@@ -85,10 +85,10 @@
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.4.1" />
     <!--DuckTyping-->
     <PackageReference Include="System.Reflection.Emit" Version="4.3.0" />
     <PackageReference Include="System.Reflection.Emit.Lightweight" Version="4.3.0" />
-    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.4.1" />
   </ItemGroup>
 
   <ItemGroup>


tracer/src/Datadog.Trace/Datadog.Trace.csproj
```diff
@@ -90,6 +88,7 @@
     <!--DuckTyping-->
     <PackageReference Include="System.Reflection.Emit" Version="4.3.0" />
     <PackageReference Include="System.Reflection.Emit.Lightweight" Version="4.3.0" />
+    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.4.1" />
```
andrewlock: nit: I'd move this _above_ the `    <!--DuckTyping-->` comment, because it's _not_ about duck typing 😄

NachoEchevarria: True, thanks!!!

PR #8118, opened 2026-01-30
Increase timeout DiscoveryService Tests

NachoEchevarria:
## Summary of changes

 Increases mutex timeouts from 10s to 30s in DiscoveryServiceTests to reduce flakiness on slow CI environments.

## Reason for change

Three DiscoveryServiceTests tests were intermittently failing.

## Implementation details

- Increased mutex.Wait timeouts from 10,000ms to 30,000ms across affected tests

### DIFF
--- a/tracer/test/Datadog.Trace.Tests/Agent/DiscoveryServiceTests.cs
+++ b/tracer/test/Datadog.Trace.Tests/Agent/DiscoveryServiceTests.cs
@@ -177,9 +177,6 @@ public class DiscoveryServiceTests
             });
 
         var ds = new DiscoveryService(factory, NullContainerMetadata, InitialRetryDelayMs, MaxRetryDelayMs, recheckIntervalMs);
-
-        // Subscribe BEFORE Request 1 completes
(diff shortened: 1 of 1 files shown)


NachoEchevarria:
> Hey, nice idea, but I don't think this will work...
> 
> AFAICT, none of the changes in this PR will resolve any flakiness 🤔 We already had mutexes at the appropriate place, this PR just adds an _extra_ mutex in the middle. But given it's always doing `mutex2.Wait()` immediately before `mutex3.Wait()`, this won't improve anything I think.
> 
> It seems like the issue is more likely some sort of resource exhaustion, as we've been seeing recently generally, and if anything, this may actually make the problem _worse_ as it gives us extra mutexes to wait for 😅
> 
> IMO, the biggest thing we could do to reduce the flake is probably revert all this, and just bump up the 10s wait time to be 30s, unless there's something else we're missing here 🤷‍♂️

Thanks for the feedback! 
I have increased the timeouts, hope it will remove flakiness.

PR #8120, opened 2026-01-30
Filter more dynamic assemblies

bouwkast:
## Summary of changes

This filters out more dynamic assemblies that we don't want to send to telemetry due to cardinality issues. The main
  additions are:
  - 10 hex character assembly names (e.g., `a43d8b99ea`)
  - dynamicclasses<digits> pattern (e.g., `dynamicclasses254`)

Both patterns only apply to assemblies with version 0.0.0.0.

## Reason for change

Our telemetry is still sending a lot of dynamically generated assemblies. The backend is being updated to filter these out, but we should avoid sending them in the first place to reduce payload size and processing overhead.

From doing some manual testing these can slip by our `IsDynamic` check via:
- https://learn.microsoft.com/en-us/dotnet/api/system.reflection.emit.persistedassemblybuilder?view=net-10.0
- https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.load?view=net-10.0

## Implementation details

Extended `IsZeroVersionAssemblyPattern` in `DependencyTelemetryCollector` to recognize two additional patterns:
  - 10 hex chars: Reuses existing `IsHexString` helper (already used for 32+ hex char filtering)
  - dynamicclasses<digits>: New `IsDynamicClassesPattern` method that checks for the literal lowercase prefix followed by
  one or more digits

Both checks use the character-by-character comparison.

## Test coverage

Updated the tests to check for the new cases

## Other details

Followed the existing design where we do the length and character-by-character comparison, unsure how scalable / maintainable / overhead this is, I guess it isn't too bad, but am a little concerned just tacking more stuff on here.

### DIFF
--- a/tracer/src/Datadog.Trace/Telemetry/Collectors/DependencyTelemetryCollector.cs
+++ b/tracer/src/Datadog.Trace/Telemetry/Collectors/DependencyTelemetryCollector.cs
@@ -220,8 +220,7 @@ namespace Datadog.Trace.Telemetry
                  && IsBase32Char(assemblyName[5])
                  && IsBase32Char(assemblyName[6])
                  && IsBase32Char(assemblyName[7]))
-                || (assemblyName.Length == 10 && IsHexString(assemblyName, 0))
-                || (assemblyName.Length >= 32 && IsHexString(assemblyName, 0))
+                || (assemblyName.Length is 10 or >= 32 && IsHexString(assemblyName, 0))
                 || IsDynamicClassesPattern(assemblyName);
         }
 


tracer/src/Datadog.Trace/Telemetry/Collectors/DependencyTelemetryCollector.cs
```diff
@@ -218,7 +220,49 @@ private static bool IsZeroVersionAssemblyPattern(string assemblyName)
                  && IsBase32Char(assemblyName[5])
                  && IsBase32Char(assemblyName[6])
                  && IsBase32Char(assemblyName[7]))
-                || (assemblyName.Length >= 32 && IsHexString(assemblyName, 0));
+                || (assemblyName.Length == 10 && IsHexString(assemblyName, 0))
+                || (assemblyName.Length >= 32 && IsHexString(assemblyName, 0))
```
andrewlock: nit: 
```suggestion
                || (assemblyName.Length is 10 or >= 32 && IsHexString(assemblyName, 0))
```

tracer/src/Datadog.Trace/Telemetry/Collectors/DependencyTelemetryCollector.cs
```diff
@@ -218,7 +220,49 @@ private static bool IsZeroVersionAssemblyPattern(string assemblyName)
...
+        private static bool IsDynamicClassesPattern(string assemblyName)
+        {
+            // Pattern: "dynamicclasses" followed by digits, e.g., dynamicclasses254
+            // Length must be > 14 ("dynamicclasses".Length) to have at least one digit
+            if (assemblyName.Length <= 14)
+            {
+                return false;
+            }
+
+            // Check prefix character by character to avoid allocation (lowercase only)
```
andrewlock, line 236: You know, I'm wondering if we should be using `StartsWith()` and similar here and in other places in this method, to take advantage of SIMD in newer runtimes 🤔 I might have a little play with benchmarking to see what the result is

lucaspimentel: I would use `StartsWith()` regardless. What allocation are we trying to avoid here?

andrewlock: FWIW, I explored this, but the results weren't that impressive...

For high number of iterations, where tiered jit kicks in, SIMD _can_ be a lot faster (2ns vs 14ns) for the "worst" case scenario (where a match is found)


| Method                          | Categories |      Mean | Ratio |
| ------------------------------- | ---------- | --------: | ----: |
| TempPathTrue_BestCase_Simd      | BestCase   |  2.543 ns |  1.33 |
| TempPathTrue_BestCase_Original  | BestCase   |  1.911 ns |  1.00 |
|                                 |            |           |       |
| TempPathTrue_WorstCase_Simd     | WorstCase  |  2.557 ns |  0.18 |
| TempPathTrue_WorstCase_Original | WorstCase  | 14.230 ns |  1.00 |


However, making some minor modifications (remove branches, aggressively inline) to the current code makes the discrepancy _much_ less:

| Method                          | Categories |      Mean | Ratio |
| ------------------------------- | ---------- | --------: | ----: |
| TempPathTrue_BestCase_Simd      | BestCase   | 2.5783 ns |  2.61 |
| TempPathTrue_BestCase_Updated   | BestCase   | 0.9896 ns |  1.00 |
|                                 |            |           |       |
| TempPathTrue_WorstCase_Simd     | WorstCase  | 2.5189 ns |  0.78 |
| TempPathTrue_WorstCase_Updated  | WorstCase  | 3.2491 ns |  1.00 |


and if we force the benchmarks to be low repetition and to _not_ hit Tier 1 jitting, simd looks much _worse_ overall:

| Method                          | Categories |      Mean | Ratio |
| ------------------------------- | ---------- | --------: | ----: |
| TempPathTrue_BestCase_Simd      | BestCase   | 28.898 ns |  4.15 |
| TempPathTrue_BestCase_Updated   | BestCase   |  6.992 ns |  1.00 |
|                                 |            |           |       |
| TempPathTrue_WorstCase_Simd     | WorstCase  | 32.573 ns |  1.03 |
| TempPathTrue_WorstCase_Updated  | WorstCase  | 31.874 ns |  1.01 |


Given that this is prob only called 100-1000 times(?ish, i.e. once per assembly), we need to think about the "early" cases almost as much as the "later" cases. So I think not using SIMD is the better option, and minor tweaks to existing code is prob"good enough".

I'll look at making those tweaks in subsequent PRs

lucaspimentel: Nice, thanks!

PR #8123, opened 2026-02-02
Performance improvement in taglists

NachoEchevarria:
## Summary of changes

This PR improves performance in the critical hot path of tag operations by replacing O(n) linear scans with O(1) dictionary lookups and eliminating unnecessary allocations in debug paths.

## Reason for change

`TagsList` used `List<KeyValuePair<string, T>>` for storing tags, metrics, and metaStruct data, requiring O(n) linear scans for every get/set operation. This is a critical hot path as tags are accessed frequently on every span.

**Solution:**
Replaced list-based storage with `Dictionary<string, T>` for all three collections:
- `_tags`: `List<KeyValuePair<string, string>>` → `Dictionary<string, string>`
- `_metrics`: `List<KeyValuePair<string, double>>` → `Dictionary<string, double>`
- `_metaStruct`: `List<KeyValuePair<string, byte[]>>` → `Dictionary<string, byte[]>`

Eliminate String Interpolation in ToString()
- `ToString()` used string interpolation in loops, allocating format strings unnecessarily.

Replaced string interpolation with chained `Append()` calls:
```csharp
// Before:
sb.Append($"{pair.Key} (tag):{pair.Value},");

// After:
sb.Append(pair.Key).Append(" (tag):").Append(pair.Value).Append(',');
```

Applied to tags, metrics, and metaStruct enumeration in `ToString()`.

**Optimized Methods:**
- `GetTag()`: Linear scan → `TryGetValue()` (O(n) → O(1))
- `SetTag()`: Linear scan with conditional updates → Direct indexer/Remove (O(n) → O(1))
- `GetMetric()`: Linear scan → `TryGetValue()` (O(n) → O(1))
- `SetMetric()`: Linear scan with conditional updates → Direct indexer/Remove (O(n) → O(1))
- `SetMetaStruct()`: Linear scan with conditional updates → Direct indexer/Remove (O(n) → O(1))
- All `Enumerate*()` methods: Changed from indexed access to `foreach` over Dictionary

**Code Simplification:**
```csharp
// Before: O(n) linear scan
lock (tags)
{
    for (int i = 0; i < tags.Count; i++)
    {
        if (tags[i].Key == key)
        {
            return tags[i].Value;
        }
    }
}

// After: O(1) lookup
lock (tags)
{
    if (tags.TryGetValue(key, out var value))
    {
        return value;
    }
}
```

**Tag Operations (Every span):**
- Get/Set operations: **O(n) → O(1)** complexity
- Eliminates linear scans on every tag read/write
- Particularly beneficial for spans with many tags (sampling rules, propagation, custom tags)
- **No allocations changed**, but significantly faster lookups

**Frequency:**
- Affects every `span.SetTag()`, `span.GetTag()` call
- Sampling decision tag lookups (hot path)
- Context propagation tag access (every distributed trace)
- Custom user tags


andrewlock (commented):
FYI, using a dictionary here is likely _not_ better for performance, unless we have specific data to suggest otherwise. 

- These collections are pretty much always small. In those cases, a linear scan is likely going to be faster than a dictionary lookup, even though the complexity implies otherwise. The real numbers matter here.
- The string interpolation changes are fine, but not a big concern, as these _only_ apply in slow debug mode, which is already super slow an allocating 😄

NachoEchevarria:
> FYI, using a dictionary here is likely _not_ better for performance, unless we have specific data to suggest otherwise.
> 
> * These collections are pretty much always small. In those cases, a linear scan is likely going to be faster than a dictionary lookup, even though the complexity implies otherwise. The real numbers matter here.
> * The string interpolation changes are fine, but not a big concern, as these _only_ apply in slow debug mode, which is already super slow an allocating 😄

Thanks for the feedback! 
Yes, this PR was a test. The benchmark results are not very impressive, though.

PR #8127, opened 2026-02-02
[CI] Add CultureInvariant to a test that was failing locally

pierotibou:
## Summary of changes
Fixing a test that was failing on my env, because of a `,` instead of a `.`.

## Reason for change
The test was the issue not the code it tested.

## Implementation details
Just adding a good old `CultureInfo.InvariantCulture` as Lucas has been doing everywhere since ever :p

## Test coverage
N/A

### DIFF
--- a/tracer/test/Datadog.Trace.ClrProfiler.Managed.Tests/AutoInstrumentation/MongoDb/TruncatedTextWriterTests.cs
+++ b/tracer/test/Datadog.Trace.ClrProfiler.Managed.Tests/AutoInstrumentation/MongoDb/TruncatedTextWriterTests.cs
@@ -4,6 +4,7 @@
 // </copyright>
 
 using System;
+using System.Globalization;
 using System.Text;
 using Datadog.Trace.ClrProfiler.AutoInstrumentation.MongoDb.BsonSerialization;
 using FluentAssertions;
@@ -35,7 +36,7 @@ public class TruncatedTextWriterTests
         using var sw = new TruncatedTextWriter(sb);
         for (var i = 0; i < 2 * TruncatedTextWriter.MaxLength; i++)
         {
-            sw.Write("Format {0:E} ", 1_000_000);
+            sw.Write(string.Format(CultureInfo.InvariantCulture, "Format {0:E} ", 1_000_000));
         }
 
         var finalString = sw.ToString();


tracer/test/Datadog.Trace.ClrProfiler.Managed.Tests/AutoInstrumentation/MongoDb/TruncatedTextWriterTests.cs
```diff
@@ -35,7 +36,7 @@ public void Truncates_FormatStrings()
         using var sw = new TruncatedTextWriter(sb);
         for (var i = 0; i < 2 * TruncatedTextWriter.MaxLength; i++)
         {
-            sw.Write("Format {0:E} ", 1_000_000);
+            sw.Write(string.Format(CultureInfo.InvariantCulture, "Format {0:E} ", 1_000_000));
```
andrewlock, line 39: This is actually an interesting question, because maybe we should be forcing invariant culture in the `TruncatedTextWriter` 🤔 

e.g. we could update the `TruncatedTextWriter` to pass `CultureInfo.InvariantCulture` to the base `TextWriter` constructor.

```csharp
    public TruncatedTextWriter(StringBuilder sb)
        : base(CultureInfo.InvariantCulture)
```

pierotibou: I took for granted that we didn't want to do that as the code was running on the same server with same culture as used on the MongoDb CLI.
I don't mind doing the change if you think it's a better practice overall.

andrewlock: I'm actually not sure what we want to do 🤔 This is for serializing a mongo query to add it to a tag. _Should_ it be in the local culture, or in the invariant culture? What does the Datadog backend expect? 🤔

andrewlock: Yeah, you know, I think I'm probably over-thinking things
I don't think the actual writer ever writes [anything than a string](https://github.com/mongodb/mongo-csharp-driver/blob/main/src/MongoDB.Bson/IO/JsonWriter.cs/)

andrewlock (approved):

PR #8128, opened 2026-02-02
Fix broken parenting when Ocelot API Gateway is used

bouwkast:
> Mostly generated by claude with a lot of handholding

## Summary of changes

Add auto-instrumentation for [Ocelot](https://github.com/ThreeMammals/Ocelot) API gateway to fix broken distributed traces when Ocelot is used with `DD_TRACE_OTEL_ENABLED` and `OpenTelemetry.Instrumentation.Http` is used.

## Reason for change

When customers use Ocelot v23.0.0+ alongside the `OpenTelemetry.Instrumentation.Http` NuGet package, distributed traces break. Ocelot creates a `SocketsHttpHandler` with a default `ActivityHeadersPropagator` (this is normal behavior), which causes `DiagnosticsHandler` to be added to the handler chain. When the OpenTelemetry HTTP instrumentation is registered, it subscribes to the `DiagnosticSource` events from `DiagnosticsHandler` and injects its own trace context headers into downstream requests which overwrites the `x-datadog-*` / W3C `traceparent` headers already set by our `HttpClient` instrumentation.

Without the OpenTelemetry HTTP instrumentation NuGet installed, the `DiagnosticsHandler` Activity events have no subscriber to inject competing headers, so the issue does not occur.

For customers this results in broken parenting for traces/spans.

## Implementation details

- **`OcelotMessageInvokerPoolIntegration`**: Instruments `Ocelot.Requester.MessageInvokerPool.CreateHandler` to set `ActivityHeadersPropagator = null` on the returned `SocketsHttpHandler`, preventing `DiagnosticsHandler` from being added to the handler chain.
  - Two `[InstrumentMethod]` attributes handle the return type change between versions:
    - v23.0.0–24.0.*: returns `HttpMessageHandler`
    - v24.1.0–24.*.*: returns `SocketsHttpHandler`

Note: YARP needs to be updated as well from  `CreateNoOutputPropagator()` to `null` I'll do this in a follow up

## Test coverage

- Added a sample application `Samples.Ocelot.DistributedTracing` that has Ocelot configured as a reverse proxy and  `OpenTelemetry.Instrumentation.Http` installed and setup. Note that _not_ having `OpenTelemetry.Instrumentation.Http` causes it to not re-produce.
- Tests are added in `OcelotDistributedTracingTests` which follow the same pattern as the YARP tests
- Added snapshot, note that there are a couple of commits here where I removed the fix, saved the snapshot (broken) and then re-ran the snapshots. This is just to showcase the before/after

## Other details

Supported Ocelot versions: 23.0.0 to 24.*
Earlier versions do not have the same behavior as `HttpMessageHandler` was introduced first in 23 and then changed to `SocketsHttpHandler` in 24.


andrewlock:
> Earlier versions do not have the same behavior as HttpMessageHandler was introduced first in 23 and then changed to SocketsHttpHandler in 24

So just to confirm, the original issue _doesn't_ repro on earlier versions?

tracer/build/PackageVersionsGeneratorDefinitions.json
```diff
@@ -974,5 +974,29 @@
...
+      "23.0.*",
+      "23.4.*",
+      "24.0.*",
+      "24.1.*"
+    ],
+    "VersionConditions": [
+      {
+        "MinVersion": "23.0.0",
+        "MaxVersionExclusive": "24.0.0",
+        "ExcludeTargetFrameworks": ["net48", "netcoreapp2.1", "netcoreapp3.0", "netcoreapp3.1", "net5.0", "net6.0", "net7.0", "net9.0", "net10.0"]
```
andrewlock: This excludes _all_ the target frameworks except `net8.0`, which is definitely not right 😅 We should include everything that's supported, i.e. net6.0+
```suggestion
        "ExcludeTargetFrameworks": ["net48", "netcoreapp2.1", "netcoreapp3.0", "netcoreapp3.1", "net5.0"]
```

tracer/build/supported_versions.json
```diff
@@ -791,6 +791,23 @@
       }
     ]
   },
+  {
+    "integrationName": "HttpMessageHandler",
```
andrewlock: Seems a bit weird this has `integrationName` of `HttpMessageHandler` not `Ocelot`, but this is auto-generated, right?

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Http/HttpClient/SocketsHttpHandler/OcelotMessageInvokerPoolIntegration.cs
```diff
@@ -0,0 +1,81 @@
...
+    ReturnTypeName = "System.Net.Http.SocketsHttpHandler",
+    ParameterTypeNames = ["Ocelot.Configuration.DownstreamRoute"],
+    MinimumVersion = "24.1.0",
+    MaximumVersion = "24.*.*",
+    IntegrationName = IntegrationName)]
+[Browsable(false)]
+[EditorBrowsable(EditorBrowsableState.Never)]
+public sealed class OcelotMessageInvokerPoolIntegration
+{
+    private const string IntegrationName = nameof(Configuration.IntegrationId.HttpMessageHandler);
```
andrewlock, line 52: Should this be a new integration, `Ocelot`, instead of `HttpMessageHandler`? 🤔 I'm undecided tbh, I get the reasoning.

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/OcelotDistributedTracingTests.cs
```diff
@@ -0,0 +1,74 @@
+// <copyright file="OcelotDistributedTracingTests.cs" company="Datadog">
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2 License.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
+// </copyright>
+
+#if NET8_0_OR_GREATER
```
andrewlock: I think we should be testing on .NET 6+, to cover the 23.0.0 cases like we usually do.

```suggestion
#if NET6_0_OR_GREATER
```

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/OcelotDistributedTracingTests.cs
```diff
@@ -0,0 +1,74 @@
...
+            SetEnvironmentVariable("DD_TRACE_OTEL_ENABLED", "true");
+        }
+
+        public override Result ValidateIntegrationSpan(MockSpan span, string metadataSchemaVersion) => Result.DefaultSuccess;
+
+        [SkippableTheory]
+        [Trait("Category", "EndToEnd")]
+        [Trait("RunOnWindows", "True")]
+        [MemberData(nameof(PackageVersions.Ocelot), MemberType = typeof(PackageVersions))]
+        public async Task SubmitsTraces(string packageVersion)
```
andrewlock:
```suggestion
        [SkippableTheory]
        [Trait("Category", "EndToEnd")]
        [Trait("RunOnWindows", "True")]
        [CombinatorialOrPairwiseData]
        public async Task SubmitsTraces([PackageVersionData(nameof(PackageVersions.Ocelot)] string packageVersion)
```

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/OcelotDistributedTracingTests.cs
```diff
@@ -0,0 +1,74 @@
...
+            const int expectedSpans = 4;
+
+            using (var agent = EnvironmentHelper.GetMockAgent())
+            using (await RunSampleAndWaitForExit(agent, packageVersion: packageVersion, aspNetCorePort: 0))
+            {
+                var spans = await agent.WaitForSpansAsync(expectedSpans);
+                spans.Count.Should().Be(expectedSpans);
+
+                // All Datadog spans should be in the same trace
+                spans.Select(x => x.TraceId).Distinct().Count().Should().Be(1);
```
andrewlock: nit: when we have snapshots, adding extra assertions often makes it _harder_ to understand what failed. All of the "relationship" tests are implicitly asserted by the snapshot anyway - should we remove them?

tracer/test/test-applications/integrations/Samples.Ocelot.DistributedTracing/Samples.Ocelot.DistributedTracing.csproj
```diff
@@ -0,0 +1,23 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <ApiVersion Condition="'$(ApiVersion)' == ''">24.1.0</ApiVersion>
+
+    <!-- Ocelot 23.x supports net6.0-8.0, 24.x supports net8.0+. Tests use net8.0+ for simplicity. -->
+    <TargetFrameworks>net8.0;net9.0;net10.0</TargetFrameworks>
```
andrewlock: I _think_ we should test with the full range, no? I'm fairly sure we handle this in other places? 🤔 

```suggestion
    <!-- Ocelot 23.x supports net6.0-8.0, 24.x supports net8.0+. Tests use net8.0+ for simplicity. -->
    <TargetFrameworks>net6.0;net7.0;net8.0;net9.0;net10.0</TargetFrameworks>
```

tracer/test/test-applications/integrations/Samples.Ocelot.DistributedTracing/Program.cs
```diff
@@ -0,0 +1,20 @@
+using Microsoft.AspNetCore.Hosting;
```
andrewlock: nit: given this is .NET6+ only, it would probably make sense to rewrite this using the modern approach. In other words, delete Startup.cs, and change Program.cs to the following (not tested, written raw 😅):

```csharp
using Microsoft.AspNetCore.Builder;
using Microsoft.AspNetCore.Hosting;
using Microsoft.AspNetCore.Http;
using Microsoft.Extensions.Configuration;
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Hosting;
using Ocelot.DependencyInjection;
using Ocelot.Middleware;
using OpenTelemetry.Resources;
using OpenTelemetry.Trace;

var builder = new WebApplication.CreateBuilder(args);

// Simulate a customer environment where the OpenTelemetry SDK is configured.
// The OTel SDK replaces DistributedContextPropagator.Current with its own propagator,
// which causes SocketsHttpHandler's internal DiagnosticsHandler to overwrite Datadog
// trace context headers on forwarded requests.
buider.Services.AddOpenTelemetry()
    .ConfigureResource(r => r.AddService("ocelot-sample"))
    .WithTracing(opt =>
    {
        opt
            .AddAspNetCoreInstrumentation()
            .AddHttpClientInstrumentation()
            .AddOtlpExporter(o =>
            {
                // Use a non-routable endpoint so exports silently fail
                o.Endpoint = new System.Uri("http://192.0.2.1:4317");
            });
    });

// Create a minimal in-memory configuration for Ocelot startup
// The actual routes will be configured at runtime by the Worker
var ocelotConfig = new Dictionary<string, string>
{
    { "Routes", "[]" },
    { "GlobalConfiguration:BaseUrl", "http://localhost:5000" }
};

var ocelotConfig = new ConfigurationBuilder()
    .AddInMemoryCollection(ocelotConfig)
    .Build();

builder.Services.AddOcelot(ocelotConfig);
builder.Services.AddHostedService<Worker>();

var app = builder.Build();

app.MapGet("/", () => "Hello world!");
await app.UseOcelot();

app.Run();
```

andrewlock (commented):
LGTM, just a couple of minor things:

- We generally try to cover the full range of supported TFMs, so we should probably support .NET 6+ as Ocelot does
- Should this be the `HttpMessageHandler` or a new `Ocelot` integration?
- Couple of nits (using the more compact aspnetcore setup, using combinatorial data attr etc for future)

andrewlock (approved):

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Http/HttpClient/SocketsHttpHandler/OcelotMessageInvokerPoolIntegration.cs
bouwkast: Yeah, this initially was `Ocelot` but Yarp went with `HttpMessageHandler` so I just forced Claude to follow that

Honestly I'm not sure what is ideal

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/OcelotDistributedTracingTests.cs
bouwkast: Thanks!

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/OcelotDistributedTracingTests.cs
bouwkast: Changed

tracer/test/test-applications/integrations/Samples.Ocelot.DistributedTracing/Samples.Ocelot.DistributedTracing.csproj
bouwkast: 👍

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/OcelotDistributedTracingTests.cs
bouwkast: Makes sense to me, removed

tracer/build/PackageVersionsGeneratorDefinitions.json
bouwkast: Updated it to be the correct set

bouwkast:
> So just to confirm, the original issue doesn't repro on earlier versions?

Correct I couldn't get it to reproduce in earlier versions

tracer/build/supported_versions.json
bouwkast: Yeah since I have the instrumentation name set to HttpMessageHandler this is expected

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Http/HttpClient/SocketsHttpHandler/OcelotMessageInvokerPoolIntegration.cs
bouwkast: I could probably go both ways, but we aren't really "instrumenting" YARP/Ocelot here (YARP uses `HttpMessageHandler` as well).

I think based on that and that we are essentially attempting to fix re-parenting issues here not calling it `Ocelot` is the better representation, but I would say that if for some reason someone needs to disable this re-parenting fix I guess they would ideally want the option to do that without having to disable all of the `HttpMessageHandler` instrumentation. 

Maybe a second PR to update both YARP and this to be controlled by either `HttpMessageHandler` or their integration name itself?

lucaspimentel:
@codex review

NachoEchevarria (approved):

PR #8129, opened 2026-02-02
[FFE] Ruggerize integration tests Remote Config retrieval

daniel-romano-DD:
## Summary of changes
Increase timeout and change logic order to try to minimize test flakiness

## Reason for change
Flakiness is bad

### DIFF
--- a/tracer/test/test-applications/integrations/Samples.FeatureFlags/Program.cs
+++ b/tracer/test/test-applications/integrations/Samples.FeatureFlags/Program.cs
@@ -1,11 +1,12 @@
 using System;
 using System.Threading;
+using System.Threading.Tasks;
 
 namespace Samples.FeatureFlags;
 class Program
 {
 
-    private static void Main(string[] args)
+    private async static Task Main(string[] args)
     {
         int configUpdates = 0;
         Evaluator.RegisterOnNewConfigEventHandler(() => Interlocked.Increment(ref configUpdates));
@@ -35,7 +36,7 @@ class Program
                     return;
                 }
                 Console.WriteLine($"Waiting for RC...");
-                System.Threading.Thread.Sleep(1000);
+                await Task.Delay(1_000);
             }
         }
 


tracer/test/test-applications/integrations/Samples.FeatureFlags/Program.cs
```diff
@@ -21,20 +24,19 @@ private static void Main(string[] args)
...
             {
-                Console.WriteLine($"No RC received");
-                return;
+                if (attempts-- == 0)
+                {
+                    Console.WriteLine($"No RC received");
+                    return;
+                }
+                Console.WriteLine($"Waiting for RC...");
+                System.Threading.Thread.Sleep(1000);
```
andrewlock: nit, threads are precious in CI, we should prob use `await Task.Delay(1_000)` 😄

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/FeatureFlags/FeatureFlagsTests.cs
```diff
@@ -81,7 +81,7 @@ public async Task FfeEnabled()
             if (e.Value.PathAndQuery.EndsWith("api/v2/exposures"))
             {
                 Interlocked.Increment(ref eventsReceived);
-                e.Value.Headers["Content-Encoding"].Should().Be(MimeTypes.Json);
```
andrewlock, line 84: Did we change something here? Because it looks like this probably was failing before, but it wasn't causing the test to fail :sus: I'm assuming that exceptions here are just swallowed?

dromanol: Yep, the assert was failing but not halting the process, as it was happening in the RC handler. It was just showing an ugly message in the console output

PR #8133, opened 2026-02-02
Adjust Log levels in OltpExporter

bouwkast:
## Summary of changes

Changes the log levels in `OtlpExporter` to skip telemetry for some errors (while making them go from Debug to Error) and swapping some Debug 

## Reason for change

Noticed a lot of new errors but they just seem to be transient networking errors.
https://app.datadoghq.com/error-tracking/issue/443da854-fe18-11f0-ae87-da7ad0900002

Some of the current logs were either the wrong level or absent.

## Implementation details

- Bad Request (400) - not retrying - swapped to Error from Warning
- Add two `Log.ErrorSkipTelemetry("An error occurred while sending OTLP request to {AgentEndpoint}.` for when we get transient network issue
- `Log.Error(ex, "Error sending OTLP request (attempt {Attempt})", (attempt + 1).ToString());` -> `Log.Debug<int>(ex, "Error sending OTLP request (attempt {Attempt})", attempt + 1);` along with removing the string allocation

### DIFF
--- a/tracer/src/Datadog.Trace/OpenTelemetry/Metrics/OtlpExporter.cs
+++ b/tracer/src/Datadog.Trace/OpenTelemetry/Metrics/OtlpExporter.cs
@@ -326,7 +326,7 @@ namespace Datadog.Trace.OpenTelemetry.Metrics
 
                     if (statusCode == 400)
                     {
-                        Log.Warning("Bad Request (400) - not retrying: {StatusCode}", response.StatusCode);
+                        Log.Error("Bad Request (400) - not retrying: {StatusCode}", response.StatusCode);
                         return false;
                     }
 
@@ -349,6 +349,7 @@ namespace Datadog.Trace.OpenTelemetry.Metrics
                         }
                     }
 
+                    Log.ErrorSkipTelemetry("An error occurred while sending OTLP request to {AgentEndpoint}. If the error isn't transient, please check https://docs.datadoghq.com/tracing/troubleshooting/connection_errors/?code-lang=dotnet for guidance.", _endpoint);
                     return false;
                 }
                 catch (TaskCanceledException) when (attempt < maxRetries)
@@ -358,7 +359,7 @@ namespace Datadog.Trace.OpenTelemetry.Metrics
                 }
                 catch (Exception ex)
                 {
-                    Log.Error(ex, "Error sending OTLP request (attempt {Attempt})", (attempt + 1).ToString());
+                    Log.Debug<int>(ex, "Error sending OTLP request (attempt {Attempt})", attempt + 1);
                     if (attempt < maxRetries)
                     {
                         retryDelay = TimeSpan.FromMilliseconds((long)(retryDelay.TotalMilliseconds * 2));
@@ -366,6 +367,7 @@ namespace Datadog.Trace.OpenTelemetry.Metrics
                     }
                     else
                     {
+                        Log.ErrorSkipTelemetry("An error occurred after {Attempt} attempts while sending OTLP request to {AgentEndpoint}. If the error isn't transient, please check https://docs.datadoghq.com/tracing/troubleshooting/connection_errors/?code-lang=dotnet for guidance.", attempt + 1, _endpoint);
                         return false;
                     }
                 }


tracer/src/Datadog.Trace/OpenTelemetry/Metrics/OtlpExporter.cs
```diff
@@ -358,7 +361,8 @@ private async Task<bool> SendWithRetry(byte[] otlpPayload, HttpRequestFactory re
                 }
                 catch (Exception ex)
                 {
-                    Log.Error(ex, "Error sending OTLP request (attempt {Attempt})", (attempt + 1).ToString());
+                    // Seeing network connectivity errors so skipping telemetry
+                    Log.ErrorSkipTelemetry(ex, "Error sending OTLP request (attempt {Attempt})", (attempt + 1).ToString());
```
andrewlock: Arguably, this shouldn't be an error at all IMO. Transient errors are _expected_, so this almost certainly shouldn't be logging an error at all. In `Api` for example, these "intermediate" attempts are `Debug`. Also nit, AI did the ToString() I assume 🙄 
```suggestion
                    Log.Debug<int>(ex, "Error sending OTLP request (attempt {Attempt})", attempt + 1);```

File: tracer/src/Datadog.Trace/OpenTelemetry/Metrics/OtlpExporter.cs
Comment (andrewlock) [inline tracer/src/Datadog.Trace/OpenTelemetry/Metrics/OtlpExporter.cs]: This almost certainly _should_ be an `Error`, because it indicates a bug
```suggestion
                        Log.Error("Bad Request (400) - not retrying: {StatusCode}", response.StatusCode);
```

tracer/src/Datadog.Trace/OpenTelemetry/Metrics/OtlpExporter.cs
```diff
@@ -292,7 +294,8 @@ private async Task<bool> SendGrpcRequest(byte[] otlpPayload)
             }
             catch (Exception ex)
             {
-                Log.Error(ex, "Exception when sending metrics OTLP gRPC request.");
+                // Seeing network connectivity errors so skipping telemetry
+                Log.ErrorSkipTelemetry(ex, "Exception when sending metrics OTLP gRPC request.");
```
andrewlock: AFAICT, an exception here would not come from network connectivity, and _should_ be logged as an error
```suggestion
                Log.Error(ex, "Exception when sending metrics OTLP gRPC request.");
```

tracer/src/Datadog.Trace/OpenTelemetry/Metrics/OtlpExporter.cs
```diff
@@ -207,7 +208,8 @@ private async Task<bool> SendOtlpRequest(IReadOnlyList<MetricPoint> metrics)
             }
             catch (Exception ex)
             {
-                Log.Error(ex, "Error sending OTLP request.");
+                // Seeing network connectivity errors so skipping telemetry
+                Log.ErrorSkipTelemetry(ex, "Error sending OTLP request.");
```
andrewlock: AFAICT, an exception here would not come from network connectivity, and _should_ be logged as an error
```suggestion
                Log.Error(ex, "Error sending OTLP request.");
```

tracer/src/Datadog.Trace/OpenTelemetry/Metrics/OtlpExporter.cs
```diff
@@ -117,7 +117,8 @@ public override async Task<ExportResult> ExportAsync(IReadOnlyList<MetricPoint>
             }
             catch (Exception ex)
             {
-                Log.Error(ex, "Error exporting OTLP metrics.");
+                // Seeing network connectivity errors so skipping telemetry
+                Log.ErrorSkipTelemetry(ex, "Error exporting OTLP metrics.");
```
andrewlock: AFAICT, an exception here would not come from network connectivity, and _should_ be logged as an error
```suggestion
                Log.Error(ex, "Error exporting OTLP metrics.");
```

tracer/src/Datadog.Trace/OpenTelemetry/Metrics/OtlpExporter.cs
andrewlock, line 353: I think _this_ is where we need an error log:
```suggestion
                    Log.ErrorSkipTelemetry("An error occurred while sending OTLP request to {AgentEndpoint}. If the error isn't transient, please check https://docs.datadoghq.com/tracing/troubleshooting/connection_errors/?code-lang=dotnet for guidance.", _endpoint);
                    return false;
```

andrewlock, line 371: and due to the way this code is factored, we need one here too:


```suggestion
                        Log.ErrorSkipTelemetry("An error occurred after {Attempt} attempts while sending OTLP request to {AgentEndpoint}. If the error isn't transient, please check https://docs.datadoghq.com/tracing/troubleshooting/connection_errors/?code-lang=dotnet for guidance.", attempt + 1, _endpoint);
                        return false;
```

tracer/src/Datadog.Trace/OpenTelemetry/Metrics/OtlpExporter.cs
bouwkast: Ah yes, I ran my `/analyze-error` on it, forgot to add the label

bouwkast: I'll swap these over thanks!

tracer/src/Datadog.Trace/OpenTelemetry/Metrics/OtlpExporter.cs
bouwkast: Thanks had to manually apply this one as GitHub couldn't

tracer/src/Datadog.Trace/OpenTelemetry/Metrics/OtlpExporter.cs
bouwkast: Hmm yes thanks for catching, it seems my /analyze-error command had this instruction `**Change Log.Error to Log.ErrorSkipTelemetry** in the specific file` which was followed here oops

link04 (approved):
TY

PR #8134, opened 2026-02-03
[Profiler] Measure memory consumption

chrisnas:
## Summary of changes
Add memory footprint metrics and log memory consumption at shutdown

## Reason for change
Monitor memory consumption

## Implementation details
Add metrics and logs for most profiler components

## Test coverage
N/A


profiler/src/ProfilerEngine/Datadog.Profiler.Native/CorProfilerCallback.cpp
```diff
@@ -1793,6 +1835,74 @@ HRESULT STDMETHODCALLTYPE CorProfilerCallback::Shutdown()
...
+        totalMemory += _pHeapSnapshotManager->GetMemorySize();
+    }
+
+    if (_pNetworkProvider != nullptr)
+    {
+        _pNetworkProvider->LogMemoryBreakdown();
+        totalMemory += _pNetworkProvider->GetMemorySize();
+    }
+
+    Log::Info("Total measured profiler memory: ", totalMemory, " bytes (", (totalMemory / 1024.0 / 1024.0), " MB)");
```
gleocadie: I'm a bit reluctant to add this to log file. Not sure this is something we would like customer to see 😬

PR #8153, opened 2026-02-04
Refactoring and improvements to `HttpHeaderHelpers`

andrewlock:
## Summary of changes

Various usability and performance improvements to `HttpHeaderHelper`

## Reason for change

`HttpHeaderHelper` is used in conjunction with `AgentTransportStrategy`, for the cases where we're using our custom `DatadogHttpClient` instead of the built-in `HttpClient` etc. However, it's a bit clunky to use, and the way we're doing things at the moment causes a bunch of additional allocations on startup which we can avoid. 

The PR is a linear set of self-contained commits, so feel free to either review one-by-one, or just go for it and review the whole lot. 

## Implementation details

- If the header values are constants, we can pre-concatenate them all as a single constant at compile time, instead of building it up dynamically at runtime (with all the associated intermediate allocations)
- Split the "container ID minimal headers" into its own type, so we can optimize these separately and more clearly
- We were doing a "poor man's lazy" for the header calculation, but we can just use a normal `Lazy<T>` for this and accept it. We want to be lazy because we only ever call this property if we're using our customer client
- Use `StringBuilder` to build the lazy values instead of linq and interpolation
- Turn the helpers into singletons, and just pass those in instead of `Func<T>` calls, to reduce some allocations
- Don't pass in `uri => uri` lambdas as pointless allocation that we can handle elsewhere
- Instead of having to pass in an `HttpHeaderHelper` _and_ a collection of default headers, expose the  default headers directly on `HttpHeaderHelper`, as these must always stay in sync anyway.
- Fix bug in `ExposureApi` that wasn't matching the two sets of headers it sent (see above point!)
- Remove `ContentType` property from `HttpHeaderHelperBase` as it's never used
- Delete `MultipartAgentHeaderHelper`, as it's now identical to `MinimalAgentHeaderHelper`
- Rename `MetadataHeaders` to `HttpSerializedDefaultHeaders` (as more descriptive)

## Test coverage

Added some unit tests to ensure all the `HttpHeaderHelper` implementations return the same headers (and the expected values) whether using the `DefaultHeaders` property or the private `HttpSerializedDefaultHeaders` property

### DIFF
--- a/tracer/src/Datadog.Trace/DataStreamsMonitoring/Transport/DataStreamsHttpHeaderHelper.cs
+++ b/tracer/src/Datadog.Trace/DataStreamsMonitoring/Transport/DataStreamsHttpHeaderHelper.cs
@@ -14,11 +14,11 @@ namespace Datadog.Trace.DataStreamsMonitoring.Transport;
 internal sealed class DataStreamsHttpHeaderHelper : HttpHeaderHelperBase
 {
     public static readonly DataStreamsHttpHeaderHelper Instance = new();
-    private readonly Lazy<string> _metadataHeaders;
+    private readonly Lazy<string> _serializedHeaders;
 
     private DataStreamsHttpHeaderHelper()
     {
-        _metadataHeaders = new(static () =>
+        _serializedHeaders = new(static () =>
         {
             var sb = StringBuilderCache.Acquire();
             foreach (var kvp in DataStreamsHttpHeaderNames.GetDefaultAgentHeaders())
@@ -35,5 +35,5 @@ internal sealed class DataStreamsHttpHeaderHelper : HttpHeaderHelperBase
 
     public override KeyValuePair<string, string>[] DefaultHeaders => DataStreamsHttpHeaderNames.GetDefaultAgentHeaders();
 
-    protected override string MetadataHeaders => _metadataHeaders.Value;
+    protected override string HttpSerializedDefaultHeaders => _serializedHeaders.Value;
 }
--- a/tracer/src/Datadog.Trace/HttpOverStreams/TraceAgentHttpHeaderHelper.cs
+++ b/tracer/src/Datadog.Trace/HttpOverStreams/TraceAgentHttpHeaderHelper.cs
@@ -14,11 +14,11 @@ namespace Datadog.Trace.HttpOverStreams
     internal sealed class TraceAgentHttpHeaderHelper : HttpHeaderHelperBase
     {
         public static readonly TraceAgentHttpHeaderHelper Instance = new();
-        private readonly Lazy<string> _metadataHeaders;
+        private readonly Lazy<string> _serializedHeaders;
 
         private TraceAgentHttpHeaderHelper()
         {
-            _metadataHeaders = new(static () =>
+            _serializedHeaders = new(static () =>
             {
                 var sb = StringBuilderCache.Acquire();
                 foreach (var kvp in AgentHttpHeaderNames.DefaultHeaders)
@@ -35,6 +35,6 @@ namespace Datadog.Trace.HttpOverStreams
 
         public override KeyValuePair<string, string>[] DefaultHeaders => AgentHttpHeaderNames.DefaultHeaders;
 
-        protected override string MetadataHeaders => _metadataHeaders.Value;
+        protected override string HttpSerializedDefaultHeaders => _serializedHeaders.Value;
     }
 }
--- a/tracer/src/Datadog.Trace/HttpOverStreams/EventPlatformHeaderHelper.cs
+++ b/tracer/src/Datadog.Trace/HttpOverStreams/EventPlatformHeaderHelper.cs
@@ -22,10 +22,10 @@ internal sealed class EventPlatformHeaderHelper : HttpHeaderHelperBase
             ..AgentHttpHeaderNames.MinimalHeaders,
             new(evpHeaderKey, evpHeaderValue),
         ];
-        MetadataHeaders = $"{AgentHttpHeaderNames.HttpSerializedMinimalHeaders}{evpHeaderKey}: {evpHeaderValue}{DatadogHttpValues.CrLf}";
+        HttpSerializedDefaultHeaders = $"{AgentHttpHeaderNames.HttpSerializedMinimalHeaders}{evpHeaderKey}: {evpHeaderValue}{DatadogHttpValues.CrLf}";
     }
 
     public override KeyValuePair<string, string>[] DefaultHeaders { get; }
 
-    protected override string MetadataHeaders { get; }
+    protected override string HttpSerializedDefaultHeaders { get; }
 }
--- a/tracer/src/Datadog.Trace/HttpOverStreams/HttpHeaderHelperBase.cs
+++ b/tracer/src/Datadog.Trace/HttpOverStreams/HttpHeaderHelperBase.cs
@@ -13,7 +13,7 @@ namespace Datadog.Trace.HttpOverStreams
     {
         public abstract KeyValuePair<string, string>[] DefaultHeaders { get; }
 
-        protected abstract string MetadataHeaders { get; }
+        protected abstract string HttpSerializedDefaultHeaders { get; }
 
         public Task WriteLeadingHeaders(HttpRequest request, TextWriter writer)
         {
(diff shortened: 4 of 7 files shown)


tracer/src/Datadog.Trace/AgentHttpHeaderNames.cs
```diff
@@ -86,6 +86,14 @@ internal static class AgentHttpHeaderNames
         /// </summary>
         public const string DroppedP0Spans = "Datadog-Client-Dropped-P0-Spans";
 
+        /// <summary>
+        /// Returns <see cref="MinimalHeaders"/>, in the format <c>Key: Value\r\n</c>. For use in HTTP headers
+        /// </summary>
+        internal const string HttpSerializedMinimalHeaders =
+            $"{Language}: .NET" +
```
andrewlock: The HttpSerializedMinimalHeaders constant is missing the carriage return and line feed delimiters between headers. Comparing this to MinimalHeaders (lines 113-118), there should be 3 headers but they need proper CRLF separators. The constant is used in MinimalAgentHeaderHelper and should match the format Key: Value followed by CRLF for each header. 🤖 Claude Code

tracer/src/Datadog.Trace/DataStreamsMonitoring/Transport/DataStreamsHttpHeaderHelper.cs
```diff
@@ -4,29 +4,39 @@
...
+                sb.Append(kvp.Key);
+                sb.Append(": ");
+                sb.Append(kvp.Value);
+                sb.Append(DatadogHttpValues.CrLf);
             }
 
-            return _metadataHeaders;
-        }
+            // remove last char
+            sb.Remove(sb.Length - 1, 1);
```
andrewlock: This removes only 1 character but CRLF is 2 characters. This leaves a trailing carriage return. Should be sb.Remove(sb.Length - 2, 2) to properly remove the full CRLF sequence. The same issue exists in TraceAgentHttpHeaderHelper.cs line 33. 🤖 Claude Code

tracer/src/Datadog.Trace/Agent/AgentTransportStrategy.cs
```diff
@@ -27,19 +27,17 @@ internal static class AgentTransportStrategy
     /// <param name="productName">The product this is transport for e.g. 'traces', 'telemetry'.
     /// Used in logging only </param>
     /// <param name="tcpTimeout">The timeout to use in TCP/IP requests</param>
-    /// <param name="defaultAgentHeaders">The default headers to add to HttpClient requests</param>
-    /// <param name="getHttpHeaderHelper">A func that returns an <see cref="HttpHeaderHelperBase"/> for use
+    /// <param name="httpHeaderHelper">A func that returns an <see cref="HttpHeaderHelperBase"/> for use
```
andrewlock: The XML doc says the parameter is named httpHeaderHelper but it should say it's an HttpHeaderHelperBase instance (not a func that returns one, which is what the old parameter was). The description is still referring to the old func-based design. 🤖 Claude Code

tracer/src/Datadog.Trace/HttpOverStreams/TraceAgentHttpHeaderHelper.cs
```diff
@@ -5,28 +5,39 @@
...
+                    sb.Append(kvp.Key);
+                    sb.Append(": ");
+                    sb.Append(kvp.Value);
+                    sb.Append(DatadogHttpValues.CrLf);
                 }
 
-                return _metadataHeaders;
-            }
+                // remove last char
+                sb.Remove(sb.Length - 1, 1);
```
andrewlock: Same CRLF removal issue as DataStreamsHttpHeaderHelper. Removing only 1 character leaves a trailing carriage return. Should be sb.Remove(sb.Length - 2, 2) to properly remove both characters of the CRLF sequence. 🤖 Claude Code

tracer/src/Datadog.Trace/Debugger/DebuggerFactory.cs
```diff
@@ -123,8 +123,6 @@ private static IApiRequestFactory GetApiFactory(TracerSettings tracerSettings, b
             tracerSettings.Manager.InitialExporterSettings,
             productName: "debugger",
             tcpTimeout: TimeSpan.FromSeconds(15),
-            AgentHttpHeaderNames.MinimalHeaders,
-            isMultipart ? () => new MultipartAgentHeaderHelper() : () => new MinimalAgentHeaderHelper(),
-            uri => uri);
+            httpHeaderHelper: MinimalAgentHeaderHelper.Instance);
```
andrewlock, line 126: The DebuggerFactory no longer respects the isMultipart parameter. It always uses MinimalAgentHeaderHelper.Instance now, whereas before it would use MultipartAgentHeaderHelper when isMultipart was true. This changes the Content-Type header behavior. Was this intentional? If MultipartAgentHeaderHelper was truly identical to MinimalAgentHeaderHelper then this is fine, but the isMultipart parameter is now effectively unused. 🤖 Claude Code

tracer/test/Datadog.Trace.Tests/AgentHttpHeaderNamesTests.cs
```diff
@@ -0,0 +1,33 @@
...
+
+public class AgentHttpHeaderNamesTests
+{
+    [Fact]
+    public void AgentHttpHeaderNames_HttpSerializedMinimalHeaders_IsCorrect()
+    {
+        var headers = AgentHttpHeaderNames.MinimalHeaders.Select(kvp => $"{kvp.Key}: {kvp.Value}{DatadogHttpValues.CrLf}");
+        var expected = string.Concat(headers);
+
+        AgentHttpHeaderNames.HttpSerializedMinimalHeaders.Should().Be(expected);
```
andrewlock: Good addition of tests to verify the serialized header constants match their array counterparts. However, the test only validates that they produce the same output, not that the output is correct. Consider adding an assertion that checks the exact expected format, especially since there might be CRLF handling issues in the production code. 🤖 Claude Code

andrewlock:
This PR successfully refactors HttpHeaderHelper to reduce allocations and improve usability by introducing singleton instances and pre-computed header strings. The overall approach is sound and achieves the performance goals.

Key issues found:
- Critical: CRLF handling bugs in multiple files where only 1 character is removed instead of 2, leaving trailing carriage returns in header strings
- Missing CRLF delimiters in the HttpSerializedMinimalHeaders constant
- Outdated XML documentation referring to the old func-based API
- The isMultipart parameter in DebuggerFactory is now effectively unused

The changes follow good patterns (singleton instances, lazy initialization, StringBuilderCache usage) but the CRLF bugs should be addressed before merging as they could cause malformed HTTP headers.

Review by 🤖 Claude Code

tracer/src/Datadog.Trace/Telemetry/TelemetryHttpHeaderNames.cs
```diff
@@ -4,21 +4,30 @@
...
         /// Gets the default constant headers that should be added to any request to the agent
         /// </summary>
         internal static KeyValuePair<string, string>[] GetDefaultAgentHeaders()
-            => new[]
-            {
-                new KeyValuePair<string, string>(TelemetryConstants.ClientLibraryLanguageHeader, TracerConstants.Language),
-                new KeyValuePair<string, string>(TelemetryConstants.ClientLibraryVersionHeader, TracerConstants.AssemblyVersion),
-                new KeyValuePair<string, string>(HttpHeaderNames.TracingEnabled, "false"), // don't add automatic instrumentation to requests directed to the agent
-            };
+            =>
```
NachoEchevarria, line 25: In this case, we are creating new values for every call, right? Maybe this would be better?

  internal static KeyValuePair<string, string>[] DefaultAgentHeaders { get; } =
  [
      new(TelemetryConstants.ClientLibraryLanguageHeader, TracerConstants.Language),
      new(TelemetryConstants.ClientLibraryVersionHeader, TracerConstants.AssemblyVersion),
      new(HttpHeaderNames.TracingEnabled, "false"),
  ];

NachoEchevarria: BTW, not in this PR but DataStreamsHttpHeaderNames also creates new values for each GetDefaultAgentHeaders call

andrewlock: Thanks, I actually had the same thought, and did that originally, but changed my mind 😅 

I decided against the property + initializer because then it will _always_ be created, even if we never invoke this method, causing uneccessary allocation. And in practice, we only call this method once, so it shouldn't be an issue. That's especially enforced by making the types that use them singletons 😄

dudikeleti (approved):
Thanks!

PR #8154, opened 2026-02-04
Add configuration log if RuntimeMetricsDiagnosticsMetricsApiEnabled is enabled

anna-git:
## Summary of changes
Add configuration log if RuntimeMetricsDiagnosticsMetricsApiEnabled is enabled

## Reason for change
to see it in the logs as otherwise if it's enabled we don't know

### DIFF
--- a/tracer/src/Datadog.Trace/TracerManager.cs
+++ b/tracer/src/Datadog.Trace/TracerManager.cs
@@ -446,6 +446,12 @@ namespace Datadog.Trace
                     writer.WritePropertyName("runtime_metrics_enabled");
                     writer.WriteValue(instanceSettings.RuntimeMetricsEnabled);
 
+                    if (instanceSettings.RuntimeMetricsDiagnosticsMetricsApiEnabled)
+                    {
+                        writer.WritePropertyName("runtime_metrics_diagnostics_metrics_api_enabled");
+                        writer.WriteValue(instanceSettings.RuntimeMetricsDiagnosticsMetricsApiEnabled);
+                    }
+
                     writer.WritePropertyName("disabled_integrations");
                     writer.WriteStartArray();
 


andrewlock (commented):
Just wondering if we need this? Everything we add here adds a certain amount of noise, and this doesn't seem likely to be critical in general? + it's available in telemetry config anyway?

anna-git:
> Just wondering if we need this? Everything we add here adds a certain amount of noise, and this doesn't seem likely to be critical in general? + it's available in telemetry config anyway?

It was more for debugging purposes when testing runtime metrics, I wanted to make sure I was indeed using the right api and couldn't make sure of it in logs.. maybe adding a debug log instead  in https://github.com/DataDog/dd-trace-dotnet/blob/1863e47b65a3b779395589f71d230750a9dd3066/tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs#L328 to say which class is going to be instantiated?

andrewlock:
> > Just wondering if we need this? Everything we add here adds a certain amount of noise, and this doesn't seem likely to be critical in general? + it's available in telemetry config anyway?
> 
> It was more for debugging purposes when testing runtime metrics, I wanted to make sure I was indeed using the right api and couldn't make sure of it in logs.. maybe adding a debug log instead in
> 
> https://github.com/DataDog/dd-trace-dotnet/blob/1863e47b65a3b779395589f71d230750a9dd3066/tracer/src/Datadog.Trace/RuntimeMetrics/RuntimeMetricsWriter.cs#L328
> 
> to say which class is going to be instantiated?

If you have debug mode enabled the Log factory already writes a log with that information, as long as the types use `IDatadogLogger` (which they do) 😄

anna-git:
oh ok! I'll need to look closer to the logs next time , I'll close this one for now!

PR #8155, opened 2026-02-04
[ASM] Custom data classification

daniel-romano-DD:
## Summary of changes
Added remote configuration capabilities:
  CAPABILITY_ASM_PROCESSOR_OVERRIDES (16)
  CAPABILITY_ASM_CUSTOM_DATA_SCANNERS (17)
These capabilities inform remote config that this tracer version supports processor overrides and custom data scanners, allowing remote config to send these configurations.

## Reason for change
[RFC](https://docs.google.com/document/d/1wBrd-ShGoA9-aP96o0VIe46eBgw73GL1315R8QjuMoc/edit?tab=t.0#heading=h.4kapbym3f1x2)

## Test coverage
Added integration tests

### DIFF
--- a/tracer/test/Datadog.Trace.Security.IntegrationTests/AspNetBase.cs
+++ b/tracer/test/Datadog.Trace.Security.IntegrationTests/AspNetBase.cs
@@ -62,7 +62,7 @@ namespace Datadog.Trace.Security.IntegrationTests
 #pragma warning restore SA1202 // Elements should be ordered by access
 #pragma warning restore SA1401 // Fields should be private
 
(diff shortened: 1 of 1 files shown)


tracer/src/Datadog.Trace/AppSec/Waf/RCMCapabilitiesHelper.cs
```diff
@@ -6,6 +6,7 @@
 using System;
 using System.Collections.Generic;
 using System.Numerics;
+using System.Security.Cryptography;
```
andrewlock, line 9: 🤨

dromanol: I really don't know why VS keeps adding these kind of useless usings

PR #8160, opened 2026-02-05
[ASM] Remove unused telemetry tag from waf.request metric

daniel-romano-DD:
## Summary of changes
Remove `request_excluded` tag that was always sent to false from the `waf.requests` metric.

## Reason for change
The tag was optional and as there is no current way of determining if the request has been excluded, it has been deprecated

### DIFF
--- a/tracer/test/Datadog.Trace.Tests/Telemetry/Metrics/MetricTests.cs
+++ b/tracer/test/Datadog.Trace.Tests/Telemetry/Metrics/MetricTests.cs
@@ -67,6 +67,11 @@ public class MetricTests
         { "init_time", new() { "total", "component" } }, // we only send one of these
     };
 
+    // Deprecated tags which may still appear in the shared common_metrics.json file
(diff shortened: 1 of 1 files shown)


e-n-0 (approved):

tracer/test/Datadog.Trace.Tests/Telemetry/Metrics/common_metrics.json
```diff
@@ -740,7 +740,6 @@
         "rule_triggered",
         "request_blocked",
         "waf_timeout",
-        "request_excluded",
```
andrewlock, line 743: Hmmm 🤔 This one is still specified in the common_metrics.json, so we _shouldn't_ remove it here (otherwise it will reappear later when we sync the file)...

If having it in here causes some tests to fail, we probably need to take a look at those tests 💀

dromanol: Ok, on it

PR #8163, opened 2026-02-05
[DSM] Include container tags hash in node hash calculation

vandonr:
## Summary of changes

Include container tags hash in the DSM (Data Streams Monitoring) node hash calculation, similar to how process tags are already included. Since container tags may not be available at startup, the hash is dynamically updated via subscriptions to `AgentConfiguration` changes from the discovery service.

## Reason for change

Service remapping support — [see RFC](https://docs.google.com/document/d/15GtNOKGBCt6Dc-HsDNnMmCdZwhewFQx8yUlI9in5n3M).

## Implementation details

- Add `ContainerTagsHash` property to `AgentConfiguration`, populated from `ContainerMetadata` in the discovery service response flow
- `DataStreamsManager` now subscribes to both `AgentConfiguration` changes (for container tags) and `MutableSettings` changes (for service name/env), recalculating the node hash when either changes
- `HashHelper.CalculateNodeHashBase` accepts a new `containerTagsHash` parameter, which is only included in the hash when process tags are also enabled
- Container tags hash is propagated through the discovery service rather than via a direct subscription on `ContainerMetadata`

## Test coverage

- `ContainerTagsHashUsedInBaseHash` — verifies the hash changes when container tags are provided (with process tags enabled)
- `ContainerTagsHashNotUsedWithoutProcessTags` — verifies container tags are ignored when process tags are disabled
- Updated `HashHelperTests.NodeHashSanityCheck` theory with container tags hash parameter
- Updated existing DSM and discovery service tests to pass the new `containerTagsHash` parameter

### DIFF
--- a/tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
+++ b/tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
@@ -54,6 +54,9 @@ internal sealed class DataStreamsManager
         _isInDefaultState = tracerSettings.IsDataStreamsMonitoringInDefaultState;
 
         _previousMutableSettings = tracerSettings.Manager.InitialMutableSettings;
+        // even though the value will probably get updated by a callback when subscriptions happen just after,
+        // we still need to initialize it to a value from initial settings in case no callback fire
+        UpdateNodeHash(_previousMutableSettings, containerTagsHash: null);
         // subscribing to changes calls the callback immediately if a value is present
         discoveryService.SubscribeToChanges(UpdateHashWithContainerTags);
         _updateSubscription = tracerSettings.Manager.SubscribeToChanges(UpdateHashWithNewSettings);


tracer/src/Datadog.Trace/DataStreamsMonitoring/Hashes/HashHelper.cs
```diff
@@ -19,7 +19,7 @@ internal static class HashHelper
     /// Calculates the base NodeHash for a service.
     /// This can be used to create a <see cref="NodeHash"/> by calling <see cref="CalculateNodeHash"/>
     /// </summary>
-    public static NodeHashBase CalculateNodeHashBase(string service, string? env, string? primaryTag, string? processTags)
+    public static NodeHashBase CalculateNodeHashBase(string service, string? env, string? primaryTag, string? processTags, string? containerTagsHash)
```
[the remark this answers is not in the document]
andrewlock: Pfft, I think what you've done is correct, we just need to update the tests

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
```diff
@@ -68,7 +70,17 @@ public ContainerMetadata(string? containerId, string? entityId)
         public string? ContainerTagsHash
         {
             get => Volatile.Read(ref field);
-            set => Volatile.Write(ref field, value);
+            set
+            {
+                Volatile.Write(ref field, value);
+                lock (_subscribers)
```
andrewlock: Locking while invoking arbitrary `Action<string>` may or may not be an issue 🤔 It probably should be fine, especially seeing as (AFAICT) noone else uses it. But it makes me suspect that `ContainerTagsHash` probably shouldn't live on `ContainerMetadata` at all, and instead it should just be exposed in the discovery service response.... that way you don't have to re-invent the pub sub mechanism _and_ you don't have the whole "mutable global state" issue I've been railing against 😉

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
```diff
@@ -300,6 +312,19 @@ private static bool IsHostCgroupNamespaceInternal()
         {
             return File.Exists(ControlGroupsNamespacesFilePath) && TryGetInode(ControlGroupsNamespacesFilePath, out long output) && output == HostCgroupNamespaceInode;
         }
+
+        /// <summary>
+        /// Get notified when the container tags hash get updated.
+        /// This is suposed to happen only once per app lifetime
```
andrewlock: Except that's not true, it'll happen every time the discovery service is called...

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
```diff
@@ -37,9 +38,12 @@ internal sealed class DataStreamsManager
     public DataStreamsManager(
         TracerSettings tracerSettings,
         IDataStreamsWriter? writer,
-        string? processTags)
+        string? processTags,
+        ContainerMetadata containerMetadata)
     {
-        UpdateNodeHash(tracerSettings.Manager.InitialMutableSettings);
+        // make sure we subscribe before calling it manually so that there is no gap
+        containerMetadata.SubscribeToContainerTagsHashChanges(newHash => UpdateNodeHash(tracerSettings.Manager.InitialMutableSettings, newHash));
```
andrewlock: Yeah, the AI is correct here, this is incorrect as proposed. Consider the example:

- This code runs, container Id is null. Hash 1 calculated.
- The second line runs, container Id is still null, Hash 1 calculated (again)
- Customer updates their settings, changes the service name, container Id is still null, Hash 2 calculated
- Discovery service returns, container Id is non null, this code runs with the _original_ settings (i.e. the wrong service name), Hash 3 calculated (incorrect).

I don't have a good proposal for the alternative, but I would favor using the existing discovery service mechanisms if possible + you're probably going to need to just cache the latest service name/env locally for when the discovery service runs later...

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
vandonr: hmm, so I'd stuff them into the mutable settings, just like Pierre did for the process tags in his PR.
I thought about that at some point, but I figured "well, those are not settings" and stashed the idea.

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
vandonr: hmm, yes, true, actually what I had in mind was that we'd call the callback only if the value _changed_, but I didn't implement it 🤦 Also, doing so is not entirely straightforward.

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
andrewlock: > so I'd stuff them into the mutable settings

I was suggesting stuffing them [in `AgentConfiguration` instead](https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/src/Datadog.Trace/Agent/DiscoveryService/AgentConfiguration.cs) 😅 

Given this hash is currently only _set_ by the discovery service, and the discovery service already has a pub/sub implementation, that kind of seems like the "correct" place for them 🤔

vandonr:
@codex review

tracer/src/Datadog.Trace/Agent/DiscoveryService/DiscoveryService.cs
```diff
@@ -429,6 +429,7 @@ private async Task ProcessDiscoveryResponse(IApiResponse response)
                 eventPlatformProxyEndpoint: eventPlatformProxyEndpoint,
                 telemetryProxyEndpoint: telemetryProxyEndpoint,
                 tracerFlareEndpoint: tracerFlareEndpoint,
+                containerTagsHash: containerTagsHash,
```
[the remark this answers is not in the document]
vandonr: true, I'll change that

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
```diff
@@ -30,32 +31,47 @@ internal sealed class DataStreamsManager
...
         _isInDefaultState = tracerSettings.IsDataStreamsMonitoringInDefaultState;
+
+        _previousMutableSettings = tracerSettings.Manager.InitialMutableSettings;
+        _previousContainerTagsHash = containerMetadata.ContainerTagsHash;
+        UpdateNodeHash(tracerSettings.Manager.InitialMutableSettings, containerMetadata.ContainerTagsHash);
+        discoveryService.SubscribeToChanges(conf =>
+        {
+            _previousContainerTagsHash = conf.ContainerTagsHash;
+            UpdateNodeHash(_previousMutableSettings, conf.ContainerTagsHash);
+        });
```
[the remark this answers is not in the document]
vandonr: ok, I think this is not a great pattern to have to provide the same callback to unsubscribe and have no IDisposable provided, but let's fix it the way it currently works

vandonr:
@codex review

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
```diff
@@ -29,45 +30,73 @@ internal sealed class DataStreamsManager
...
+        }
+    }
+
+    /// <summary> Callback for MutableSettings updates </summary>
+    private void UpdateHashWithNewSettings(TracerSettings.SettingsManager.SettingChanges updates)
+    {
+        if (updates.UpdatedMutable is { } updated)
+        {
+            _previousMutableSettings = updated;
+            UpdateNodeHash(updated, _previousContainerTagsHash);
```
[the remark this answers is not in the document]
vandonr: I think this is true, but I also think those events are rare enough that the race condition is never going to happen

tracer/src/Datadog.Trace/DataStreamsMonitoring/Hashes/HashHelper.cs
```diff
@@ -35,6 +35,11 @@ public static NodeHashBase CalculateNodeHashBase(string service, string? env, st
         if (!StringUtil.IsNullOrEmpty(processTags))
         {
             hash = FnvHash64.GenerateHash(processTags, HashVersion, hash);
+            // container tags are only added if process tags are in use
+            if (!StringUtil.IsNullOrEmpty(containerTagsHash))
+            {
+                hash = FnvHash64.GenerateHash(containerTagsHash, HashVersion, hash);
+            }
```
wantsui, line 42: This seems functionally similar to what Ruby is doing: https://github.com/DataDog/dd-trace-rb/blob/c5d8dee9674e1c9e985225f1e47ce9a6313df3cc/lib/datadog/core/environment/agent_info.rb#L123-L137 and what Python does: https://github.com/DataDog/dd-trace-py/blob/4f7c157b2d6427f13764bd784de0a3cd7a013758/ddtrace/internal/process_tags/__init__.py#L85-L92

wantsui: (As in container tags are only referenced if process tags are enabled)

wantsui (approved):
LGTM.

I will a disclaimer that I'm not 100% familiar with the inner workings in .NET code and how the DiscoveryService call is accessing Agent information. For Ruby, I am relying on an "agent /info" call to fetch container tags, and it seems Python and .NET have their own unique ways to getting this data which result in slightly different hash logic (the components of what goes into the hash like process tags and container tags are the same). **_Functionally_**, .NET achieves the same goal as what I saw in the Python and Ruby tracer.

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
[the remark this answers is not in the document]
lucaspimentel: > those events are rare enough that the race condition is never going to happen

Famous last words 😅

tracer/src/Datadog.Trace/DataStreamsMonitoring/Hashes/HashHelper.cs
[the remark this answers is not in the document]
lucaspimentel: Why "pfft"? The comment is correct 😅

tracer/src/Datadog.Trace/Agent/DiscoveryService/DiscoveryService.cs
```diff
@@ -429,6 +429,7 @@ private async Task ProcessDiscoveryResponse(IApiResponse response)
                 eventPlatformProxyEndpoint: eventPlatformProxyEndpoint,
                 telemetryProxyEndpoint: telemetryProxyEndpoint,
                 tracerFlareEndpoint: tracerFlareEndpoint,
+                _containerMetadata.ContainerTagsHash, // either the value just received, or the one we stored before (prevents overriding with null)
```
lucaspimentel: for consistency with all the other arguments
```suggestion
                containerTagsHash: _containerMetadata.ContainerTagsHash, // either the value just received, or the one we stored before (prevents overriding with null)
```

tracer/test/Datadog.Trace.Tests/DataStreamsMonitoring/DataStreamsManagerTests.cs
```diff
@@ -196,12 +199,36 @@ public void ProcessTagsUsedInBaseHash()
...
+        var service = "bar";
+
+        var hashWithout = HashHelper.CalculateNodeHashBase(service, env, primaryTag: null, processTags: "hello:world", containerTagsHash: null);
+        var hashWith = HashHelper.CalculateNodeHashBase(service, env, primaryTag: null, processTags: "hello:world", "12345ABCDE");
+
+        hashWith.Value.Should().NotBe(hashWithout.Value);
+    }
+
+    [Fact]
+    public void ContainerTagsHashNotUsedUsedWithoutProcessTags()
```
lucaspimentel: double `Used` typo?
```suggestion
    public void ContainerTagsHashNotUsedWithoutProcessTags()
```

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.NetFramework.cs
```diff
@@ -57,5 +55,10 @@ public string? ContainerTagsHash
     /// </summary>
     /// <value>The entity id or <c>null</c>.</value>
     public string? EntityId => null;
+
+    public void SubscribeToContainerTagsHashChanges(Action<string?> action)
```
lucaspimentel: Do we need this? It doesn't seem to be called anywhere.

vandonr: no you're right it's a leftover from a previous version

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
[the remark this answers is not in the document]
lucaspimentel: Seriously, though, the rest of the PR LGTM but I don't feel comfortable approving with a known race condition that is relatively simple to avoid.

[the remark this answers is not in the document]
vandonr: ok, yeah I was thinking that it's not worth having a lock just for this, but also it's not called very often, so probably not such a big deal

vandonr:
@codex review

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.NetFramework.cs
```diff
@@ -28,10 +28,7 @@ public ContainerMetadata(string containerId, string entityId)
         // nothing to do, just to match the other version
     }
 
-    /// <summary>
-    /// Gets or sets the container tags hash received from the agent, used by DBM/DSM
-    /// This is set when we receive a value for it in an http response from the agent
-    /// </summary>
+    // always null in this implementation
```
lucaspimentel, line 35: > _// always null in this implementation_

There's nothing stopping someone from setting it to something else, making this comment false. Maybe consider making the getter return `null` and making the setter a no-op (assuming we need a setter to keep the api consistent with the other version in `ContainerMetadata.cs`)?

tracer/test/Datadog.Trace.Tests/Debugger/DynamicInstrumentationTests.cs
```diff
@@ -109,6 +109,7 @@ public void SubscribeToChanges(Action<AgentConfiguration> callback)
                     eventPlatformProxyEndpoint: "eventPlatformProxyEndpoint",
                     telemetryProxyEndpoint: "telemetryProxyEndpoint",
                     tracerFlareEndpoint: "tracerFlareEndpoint",
+                    "containerTagsHash",
```
lucaspimentel:
```suggestion
                    containerTagsHash: "containerTagsHash",
```

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.NetFramework.cs
vandonr: yes we need the set for the compiler

tracer/src/Datadog.Trace/DataStreamsMonitoring/Hashes/HashHelper.cs
andrewlock: Is that behavior specified in the RFC?

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
```diff
@@ -26,47 +27,82 @@ internal sealed class DataStreamsManager
...
     private bool _isInDefaultState;
     private IDataStreamsWriter? _writer;
 
     public DataStreamsManager(
         TracerSettings tracerSettings,
         IDataStreamsWriter? writer,
-        string? processTags)
+        IDiscoveryService discoveryService,
+        string? processTags,
+        ContainerMetadata containerMetadata)
```
andrewlock: If the `containerTagsHash` comes from the discoveryService, why do we need to inject it directly here? 🤔

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.cs
```diff
@@ -12,6 +12,7 @@
 using System.Linq;
 using System.Text.RegularExpressions;
 using System.Threading;
+using Datadog.Trace.Agent.DiscoveryService;
```
andrewlock:
```suggestion
```

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.NetFramework.cs
```diff
@@ -28,14 +32,18 @@ public ContainerMetadata(string containerId, string entityId)
...
     {
-        get => Volatile.Read(ref field);
-        set => Volatile.Write(ref field, value);
+        get => null;
+        set
+        {
+            if (!_warnedOnSet)
+            {
+                _warnedOnSet = true;
+                Log.Warning("The code is trying to set the value '{Value}' to {Prop}, but this has no effect in NetFramework.", value, nameof(ContainerTagsHash));
```
andrewlock: Should this actually be an error so we can see it in telemetry? This would be a programming error wouldn't it?
```suggestion
                Log.Error("The code is trying to set the value '{Value}' to {Prop}, but this has no effect in NetFramework.", value, nameof(ContainerTagsHash));
```

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
```diff
@@ -77,12 +113,13 @@ public static DataStreamsManager Create(
                          ? DataStreamsWriter.Create(settings, profilerSettings, discoveryService)
                          : null;
 
-        return new DataStreamsManager(settings, writer, settings.PropagateProcessTags ? ProcessTags.SerializedTags : null);
+        return new DataStreamsManager(settings, writer, discoveryService, settings.PropagateProcessTags ? ProcessTags.SerializedTags : null, ContainerMetadata.Instance);
```
andrewlock: AFAICT, we don't actually need to inject the `ContainerMetadata.Instance`?
```suggestion
        return new DataStreamsManager(settings, writer, discoveryService, settings.PropagateProcessTags ? ProcessTags.SerializedTags : null);
```

tracer/src/Datadog.Trace/DataStreamsMonitoring/Hashes/HashHelper.cs
vandonr: uuuh, in the RFC I'm not sure, but it's enforced by system tests, so I guess that's almost as good (if not better)

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
vandonr: hmm, they are not made accessible to the outside, but we could

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.NetFramework.cs
vandonr: I guess customer code could set this value ? That'd be very wicked though.
If it's set in netframework, it's probably something we should investigate yeah.

tracer/src/Datadog.Trace/DataStreamsMonitoring/DataStreamsManager.cs
vandonr: It comes from the discovery service on updates, but we need it here to compute the hash the first time. We cannot rely solely on updates, because If this happens after we received container metadata, we might wait a long time before we get an update

vandonr: ooooh one thing I had missed it that subscribing calls the callback immediately if there is data

tracer/src/Datadog.Trace/PlatformHelpers/ContainerMetadata.NetFramework.cs
lucaspimentel: Can you change `NetFramework` to `.NET Framework`, please? This is a user-facing message and branding is important :)

PR #8164, opened 2026-02-05
Add Multi-Azure Function Unique Named Pipe Configuration

Lewis-E:
## Summary of changes
* Allow use of unique windows named pipes for IPC, coordinated with the serverless compatibility layer's mini-agent.
* See: https://github.com/DataDog/datadog-serverless-compat-dotnet/pull/20 for compat-side changes.
* See: https://github.com/DataDog/serverless-compat-self-monitoring/pull/54 for integration testing in Azure.

## Reason for change
* Improve support for running multiple Azure Functions in the same hosting plan. [[jira](https://datadoghq.atlassian.net/browse/SVLS-8244)]

## Implementation details
For [Azure Functions](https://docs.datadoghq.com/serverless/azure_functions/?tab=dotnet#setup), the tracer loads first, then the compat layer, and then the compat layer launches a rust binary to act as the agent. Additionally, the client may or may not be using the DogstatsD client, and it is also possible to have the compat layer and DogstatsD in place without the tracer. We need to coordinate named pipes across all three/four of these codebases, noting that the tracer will use DogstatsD's pipe for runtime metrics if available.

Given this order of operations, and that we want this to work with immutable configuration, the tracer should set pipe names for the compatibility layer via instrumentation if the tracer is present. 

1. Tracer loads -- pipe name is generated
2. We check the presence of the compat layer + windows configs (using undocumented envvar `DD_SERVERLESS_COMPAT_BINARY_PATH` if necessary to adapt to Azure changes) to see if named pipes can be used or if we should rely on TCP
4. Compat layer loads
5. Tracer uses instrumentation to force the compat layer to generate the pipe names <pipe-a>, <pipe-b>
6. Compat layer passes the name to the mini-agent tracer <pipe-a> and dogstatsd <pipe-b>.

| Scenario | Pipe generated? | Transport | Result |                                                                      
  |---|---|---|---|                               
  | New tracer + no compat at all | No (binary not found) | TCP | Works |                                                  
  | New tracer + old compat (< 1.5) | No (version check fails) | TCP | Works |                                             
  | New tracer + new compat (≥ 1.5) | Yes | Named pipe | Works (integration coordinates) |                                 
  | New tracer + dev compat (0.0.0) | Yes | Named pipe | Works (integration coordinates) |
  | Old tracer + no compat | No | TCP | Works |
  | Old tracer + old compat (< 1.5) | No | TCP | Works (neither side knows about pipes) |
  | Old tracer + new compat (≥ 1.5) | No | TCP | **Broken** — tracer sends via TCP, but compat binary listens on pipe only. Requires [more serverless-components changes](https://github.com/DataDog/serverless-components/pull/94). |
  | Old tracer + dev compat (0.0.0) | No | TCP | Not supported |

+ dd-go PR for new configuration key: https://github.com/DataDog/dd-go/pull/228812

## Test coverage
* Self Monitoring
* New Unit Tests


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Serverless/CompatibilityLayer_CalculateDogStatsDPipeName_Integration.cs
```diff
@@ -0,0 +1,101 @@
...
+    /// silently skips instrumentation — no exception is thrown.
+    /// </summary>
+    [InstrumentMethod(
+        AssemblyName = "Datadog.Serverless.Compat",
+        TypeName = "Datadog.Serverless.CompatibilityLayer",
+        MethodName = "CalculateDogStatsDPipeName",
+        ReturnTypeName = ClrNames.String,
+        ParameterTypeNames = [],
+        MinimumVersion = "0.0.0",
+        MaximumVersion = "2.*.*",
```
duncanpharvey: What's the significance of maximum version 2.*.*? Would we need to update these if we ever release a datadog-serverless-compat-dotnet 3.*.*?

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -217,6 +259,107 @@ private static string GetMetricsHostNameFromAgentUri(Uri agentUri)
...
+                var version = assemblyName.Version;
+
+                if (version is null)
+                {
+                    Log.Warning("Could not read Serverless Compatibility Layer details at {Path}, using fallback agent communication methods. (No Named Pipes)", compatDllPath);
+                    return false;
+                }
+
+                // Allow 0.0.0 (dev builds) or >= 1.4.0 (first release with pipe support)
+                var minVersion = new Version(1, 4, 0);
```
duncanpharvey: Should we make sure to release this version of serverless compat before merging the tracer changes?

duncanpharvey (approved):
> We check the presence of the compat layer + windows configs (using undocumented envvar DD_SERVERLESS_COMPAT_BINARY_PATH if necessary to adapt to Azure changes) to see if named pipes can be used or if we should rely on TCP

What is meant by this? I don't see `DD_SERVERLESS_COMPAT_BINARY_PATH` in any of the PR changes.

duncanpharvey:
Should this dependency for `Datadog.Serverless.Compat` also be updated as part of this PR? https://github.com/DataDog/dd-trace-dotnet/blob/53edb3f261bf6687316dd44920ccea8cfe2a036d/tracer/src/Datadog.AzureFunctions/Datadog.AzureFunctions.csproj#L33

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Serverless/CompatibilityLayer_CalculateDogStatsDPipeName_Integration.cs
lucaspimentel: > What's the significance of maximum version

It's a gating mechanism we use in all our instrumentations. We don't want instrument the next major version of any library automatically without human intervention (testing). This is mostly for 3rd party libraries that we don't have control of.

> Would we need to update these

Yes. I suppose you _could_ use a larger major version here if you pinky-swear you won't make breaking changes to the APIs we're instrumenting in this PR.

lucaspimentel:
@codex review

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -26,6 +29,24 @@ namespace Datadog.Trace.Configuration
...
     {
+        private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(ExporterSettings));
+
+        /// <summary>
+        /// Backing field for AzureFunctionsGeneratedTracesPipeName.
+        /// Initialized statically by checking environment variables directly.
+        /// Only generates a name when running in Azure Functions without AAS Site Extension
+        /// and without an explicit DD_TRACE_PIPE_NAME configuration.
+        /// </summary>
+        private static readonly string? _azureFunctionsGeneratedTracesPipeName = InitAzureFunctionsTracesPipeName();
```
lucaspimentel: For various reasons, we generally try to avoid static state in the tracer and static initializers with non-trivial logic in them. (Happy to chat about this more if you're interested in the reasons.)

Instead of adding these static fields and their initializers, I think we can move this into the `ExporterSettings` instance constructor below, around lines 115-116. Something like:

```csharp
var tracesPipeName = GetAzureFunctionsTracesPipeName() ?? rawSettings.TracesPipeName;
var metricsPipeName = GetAzureFunctionsMetricsPipeName() ?? rawSettings.MetricsPipeName;
```

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -217,6 +259,107 @@ private static string GetMetricsHostNameFromAgentUri(Uri agentUri)
             return string.IsNullOrEmpty(traceHostname) ? DefaultDogstatsdHostname : traceHostname;
         }
 
+        private static string? InitAzureFunctionsTracesPipeName()
+        {
+            if (Util.EnvironmentHelpers.IsAzureFunctions()
+                && !Util.EnvironmentHelpers.IsUsingAzureAppServicesSiteExtension()
+                && IsCompatLayerAvailableWithPipeSupport())
+            {
+                var baseName = Util.EnvironmentHelpers.GetEnvironmentVariable(ConfigurationKeys.TracesPipeName);
```
lucaspimentel: This is only reading the "base" name from environment variables, but the tracer can be configured in other ways (e.g. config files), and those configs would be ignored there. Instead of reading from env vars directly, I think you can read the user setting from `rawSettings.TracesPipeName`, passed in via parameter.

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -217,6 +259,107 @@ private static string GetMetricsHostNameFromAgentUri(Uri agentUri)
...
+                && !Util.EnvironmentHelpers.IsUsingAzureAppServicesSiteExtension()
+                && IsCompatLayerAvailableWithPipeSupport())
+            {
+                var baseName = Util.EnvironmentHelpers.GetEnvironmentVariable(ConfigurationKeys.MetricsPipeName);
+                if (StringUtil.IsNullOrEmpty(baseName))
+                {
+                    baseName = "dd_dogstatsd";
+                }
+
+                var name = GenerateUniquePipeName(baseName!);
```
lucaspimentel: You're using `StringUtil.IsNullOrEmpty()` above, so you shouldn't need to suppress any nullability warnings.

```suggestion
                var name = GenerateUniquePipeName(baseName);
```

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -217,6 +259,107 @@ private static string GetMetricsHostNameFromAgentUri(Uri agentUri)
...
+                && !Util.EnvironmentHelpers.IsUsingAzureAppServicesSiteExtension()
+                && IsCompatLayerAvailableWithPipeSupport())
+            {
+                var baseName = Util.EnvironmentHelpers.GetEnvironmentVariable(ConfigurationKeys.TracesPipeName);
+                if (StringUtil.IsNullOrEmpty(baseName))
+                {
+                    baseName = "dd_trace";
+                }
+
+                var name = GenerateUniquePipeName(baseName!);
```
lucaspimentel: You're using `StringUtil.IsNullOrEmpty()` above, so you shouldn't need to suppress any nullability warnings.

```suggestion
                var name = GenerateUniquePipeName(baseName);
```

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -217,6 +259,107 @@ private static string GetMetricsHostNameFromAgentUri(Uri agentUri)
...
+                if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                {
+                    return false;
+                }
+#endif
+
+                // Check that the compat binary exists — it's what actually listens on the named pipe
+                // Check that the compat DLL exists and has a version that supports named pipes.
+                // Named pipe support was added in compat version 1.4.0 (dev builds use 0.0.0).
+                const string compatBinaryPath = @"C:\home\site\wwwroot\datadog\bin\windows-amd64\datadog-serverless-compat.exe";
```
lucaspimentel: We should take into account that this path can be overridden with `DD_SERVERLESS_COMPAT_PATH` ([see here](https://github.com/DataDog/datadog-serverless-compat-dotnet/blob/f06f78a9522fd9387375c95ac11ef584a6ff0e26/Datadog.Serverless/CompatibilityLayer.cs#L69)).

tracer/test/Datadog.Trace.Tests/Configuration/ExporterSettingsAzureFunctionsPipeTests.cs
```diff
@@ -0,0 +1,107 @@
...
+            {
+                var azureFuncName = ExporterSettings.AzureFunctionsGeneratedMetricsPipeName;
+                settings.MetricsPipeName.Should().Be(azureFuncName);
+            }
+        }
+
+        // Note: AzureFunctionsGeneratedPipeName_IsSet_WhenInAzureFunctionsWithoutExplicitConfig
+        // and AzureFunctionsGeneratedPipeName_IsNull_WhenExplicitPipeNameConfigured cannot be
+        // reliably tested as unit tests because the static fields are initialized at type-load
+        // time from real environment variables (WEBSITE_SITE_NAME, FUNCTIONS_WORKER_RUNTIME,
```
lucaspimentel: > cannot be reliably tested as unit tests because the static fields are initialized at type-load time

See my other comment about not using static fields and static initializers to avoid this issue.

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -217,6 +259,107 @@ private static string GetMetricsHostNameFromAgentUri(Uri agentUri)
             return string.IsNullOrEmpty(traceHostname) ? DefaultDogstatsdHostname : traceHostname;
         }
 
+        private static string? InitAzureFunctionsTracesPipeName()
+        {
+            if (Util.EnvironmentHelpers.IsAzureFunctions()
+                && !Util.EnvironmentHelpers.IsUsingAzureAppServicesSiteExtension()
+                && IsCompatLayerAvailableWithPipeSupport())
```
lucaspimentel: Consider moving these checks up the call stack so they don't have to be done twice (traces and metrics), especially `IsCompatLayerAvailableWithPipeSupport()` which does some I/O.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Serverless/CompatibilityLayer_CalculateDogStatsDPipeName_Integration.cs
andrewlock: [AFAICT we only ship 0.x.x and 1.x.x](https://www.nuget.org/packages/Datadog.Serverless.Compat#versions-body-tab) currently, so we should gate to those. 

```suggestion
        MaximumVersion = "1.*.*",
```

> Yes. I suppose you could use a larger major version here if you pinky-swear you won't make breaking changes to the APIs we're instrumenting in this PR.

IMO it's not worth the risk. It's worse to break customers by auto-rolling forward, and given the difficulty of testing serverless stuff, I think it would be sensible to err on the side of caution

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
andrewlock: 💯 no statics please 😄

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -87,9 +108,16 @@ internal ExporterSettings(Raw rawSettings, Func<string, bool> fileExists, IConfi
 
             ValidationWarnings = new List<string>();
 
+            // Use statically-generated pipe names for Azure Functions if available.
+            // The generated names already incorporate any explicit DD_TRACE_PIPE_NAME / DD_DOGSTATSD_PIPE_NAME
+            // as the base, with a unique GUID suffix appended. This ensures each function instance gets
+            // its own pipe even when sharing a hosting plan.
```
andrewlock: If/when customers set the pipe name in code, I think that still needs to take precedence, even though it could "break" their setup

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
andrewlock: Agreed, but we also need to make sure we record the new "calculated" value in telemetry as well

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
andrewlock: Most of these helpers feel like they shouldn't live in `ExporterSettings` at all 🤔 I'd suggest moving them out to a separate helper, and writing unit tests for them there.

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -217,6 +259,107 @@ private static string GetMetricsHostNameFromAgentUri(Uri agentUri)
...
+        /// that supports named pipe transport. This is called during static initialization
+        /// (before the compat assembly is loaded) so it checks files on disk rather than
+        /// loaded assemblies.
+        /// </summary>
+        private static bool IsCompatLayerAvailableWithPipeSupport()
+        {
+            try
+            {
+                // Named pipes are Windows-only
+#if !NETFRAMEWORK
```
andrewlock: Why the `#if`? 🤔 Named pipes are windows-only on _all_ TFMs, right?

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -217,6 +259,107 @@ private static string GetMetricsHostNameFromAgentUri(Uri agentUri)
...
+                {
+                    Log.Warning("Could not read Serverless Compatibility Layer details at {Path}, using fallback agent communication methods. (No Named Pipes)", compatDllPath);
+                    return false;
+                }
+
+                // Allow 0.0.0 (dev builds) or >= 1.4.0 (first release with pipe support)
+                var minVersion = new Version(1, 4, 0);
+                var devVersion = new Version(0, 0, 0);
+
+                if (version == devVersion || version >= minVersion)
```
andrewlock: It would be slightly preferable (perf wise) to avoid the `Version` allocations here I think, and just compare against Major/Minor etc given you're comparing against constant values anyway

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Serverless/CompatibilityLayer_CalculateDogStatsDPipeName_Integration.cs
```diff
@@ -0,0 +1,101 @@
...
+            if (exception is not null)
+            {
+                // If there was an exception, pass it through
+                return new CallTargetReturn<string>(returnValue);
+            }
+
+            try
+            {
+                // Get the tracer's pipe name (generated in ExporterSettings if on Windows with no explicit config)
+                // Use lazy caching to avoid repeated lookups
```
andrewlock: This is instrumenting, `CalculateDogStatsDPipeName()` - surely that's _not_ called repeatedly, right? I'm not sure that we need the extra complexity of the caching layer here. Plus there's a question of how this interacts if/when a customer manually sets the pipe name in code, but my guess is that we simply don't support that scenario 😅

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Serverless/CompatibilityLayer_CalculateDogStatsDPipeName_Integration.cs
```diff
@@ -0,0 +1,101 @@
...
+                {
+                    lock (_lock)
+                    {
+                        if (_cachedDogStatsDPipeName is null)
+                        {
+                            // Use the pipe name pre-generated by ExporterSettings if available (Azure Functions scenario)
+                            // If no pre-generated name exists (e.g. explicit config or compat check failed),
+                            // generate a unique name so both tracer and compat layer agree on the same pipe
+                            _cachedDogStatsDPipeName = ExporterSettings.AzureFunctionsGeneratedMetricsPipeName
+                                ?? ServerlessCompatPipeNameHelper.GenerateUniquePipeName("dd_dogstatsd", "DogStatsD");
```
andrewlock: Is there any point in generating a new unique name here instead of just using the `returnValue`? Seeing as "we" won't know about it in this scenario, it seems like unnecessary extra work for a scenario that's just not going to work anyway? 🤔

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Serverless/CompatibilityLayer_CalculateTracePipeName_Integration.cs
```diff
@@ -0,0 +1,101 @@
...
+    /// silently skips instrumentation — no exception is thrown.
+    /// </summary>
+    [InstrumentMethod(
+        AssemblyName = "Datadog.Serverless.Compat",
+        TypeName = "Datadog.Serverless.CompatibilityLayer",
+        MethodName = "CalculateTracePipeName",
+        ReturnTypeName = ClrNames.String,
+        ParameterTypeNames = [],
+        MinimumVersion = "0.0.0",
+        MaximumVersion = "2.*.*",
```
andrewlock:
```suggestion
        MaximumVersion = "1.*.*",
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Serverless/ServerlessCompatPipeNameHelper.cs
```diff
@@ -0,0 +1,47 @@
...
+            const int maxBaseLength = 214;
+
+            if (baseName.Length > maxBaseLength)
+            {
+                Log.Warning<string, int, int>("{PipeType} pipe base name exceeds {MaxLength} characters ({ActualLength}). Truncating to allow for GUID suffix.", pipeType, maxBaseLength, baseName.Length);
+                baseName = baseName.Substring(0, maxBaseLength);
+            }
+
+            var guid = Guid.NewGuid().ToString("N"); // "N" format removes hyphens (32 chars)
+            var uniqueName = $"{baseName}_{guid}";
```
andrewlock: We _could_ improve the allocations here on .NET Core, but prob not worth it. Inlining into the interpolated handler likely improves recent .NET at least (`ISpanFormattable`)

```suggestion
            // "N" format removes hyphens (32 chars)
            var uniqueName = $"{baseName}_{Guid.NewGuid():N}";
```

andrewlock (commented):
LGTM in general, but there's some specific things to address before we merge it (mostly already highlighted by Lucas)

- We should only target existing major versions, and bump support later, otherwise we need strict backward _and_ forward compatibility, which is very hard to maintain and very easy to screw up 😅 
- We shouldn't have static state if we can help it. If we want to cache things, we typically handle that by:
  - Move the calculating and storage of the value to a helper type as _instance_ methods.
  - Create a "singleton" instance of the helper
  - Pass the singleton in to the constructor (or, if this gives too much blast radius, allow passing in a null value for the helper, and grab the singleton in the body. This isn't the preferred approach, but it's sometimes the practical one)
- We need to ensure we record the new calculated values in telemetry
- The instrumentation seems a little overly complex considering the behaviours we need and the failure cases

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
Lewis-E: _azureFunctionsGeneratedTracesPipeName pulls from names set by customers -- it just also appends the guid, which is consistent across other compat implementations for this feature. Are you saying we should not append the guid?

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
Lewis-E: What's a good way to include `DD_SERVERLESS_COMPAT_PATH`? Claude and I both don't see any obvious options.

> Environment.GetEnvironmentVariable is banned (RS0030) and EnvironmentHelpers.GetEnvironmentVariable requires a constant from ConfigurationKeys or PlatformKeys (DD0011). Adding it to ConfigurationKeys via supported-configurations.yaml fails config registry validation since it's not a tracer config. And PlatformKeys rejects the DD_ prefix (DD0010)

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
lucaspimentel: Looks like this is addressed so I'll resolve the comment. Thanks!

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
lucaspimentel: Looks like my comment is addressed (no reading env vars directly), but I'll leave this open for Andrew's comment (configuration telemetry).

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
lucaspimentel: Fixed, thanks. Resolving comment.

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
lucaspimentel: Fixed, thanks. Resolving comment.

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
lucaspimentel: Try giving Claude this doc: `docs/development/Configuration/AddingConfigurationKeys.md`. I was hoping it would loaded automatically with [this note](https://github.com/DataDog/dd-trace-dotnet/blob/b70e6f22def717fa3dd9a60a71b302adce136719/AGENTS.md?plain=1#L328-L329), but maybe we need better wording in the `Load when...` to trigger that, or maybe we should move some of this things into claude skills now that those are a thing.

(AI agent tooling is still very WIP 😅)

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -325,6 +347,15 @@ private static string GetMetricsHostNameFromAgentUri(Uri agentUri)
             return string.IsNullOrEmpty(traceHostname) ? DefaultDogstatsdHostname : traceHostname;
         }
 
+        private string GenerateUniquePipeName(string? configuredBaseName, string defaultBaseName, string configKey)
+        {
+            var baseName = StringUtil.IsNullOrEmpty(configuredBaseName) ? defaultBaseName : configuredBaseName;
+            var name = ClrProfiler.AutoInstrumentation.Serverless.ServerlessCompatPipeNameHelper.GenerateUniquePipeName(baseName, "ExporterSettings");
```
lucaspimentel: More of a nit, but consider moving `ServerlessCompatPipeNameHelper` into maybe `Datadog.Trace/Util`, or into a new `Datadog.Trace/Serverless` (I'm already using the latter in #8231). It's a little jarring to see code outside of `ClrProfiler.AutoInstrumentation` referencing code inside `ClrProfiler.AutoInstrumentation` (something, something, tight coupling).

tracer/src/Datadog.Trace/Configuration/supported-configurations.yaml
```diff
@@ -1468,6 +1468,14 @@ supportedConfigurations:
...
           "[{"service": "cart*"}]"
       <seealso cref="Datadog.Trace.Configuration.TracerSettings.SpanSamplingRules"/>
+  DD_SERVERLESS_COMPAT_PATH:
+  - implementation: A
+    type: string
+    default: null
+    const_name: ServerlessCompatPath
+    documentation: |-
+      Overrides the default path to the serverless compat layer binary.
+      Default value is <c>C:\home\site\wwwroot\datadog\bin\windows-amd64\datadog-serverless-compat.exe</c>.
```
lucaspimentel: There are different defaults for Windows vs Linux. We should list them both.

tracer/test/Datadog.Trace.ClrProfiler.Managed.Tests/AutoInstrumentation/Serverless/ServerlessCompatIntegrationTests.cs
```diff
@@ -0,0 +1,203 @@
...
+    {
+        var traceIntegrationType = typeof(CompatibilityLayer_CalculateTracePipeName_Integration);
+        var dogstatsdIntegrationType = typeof(CompatibilityLayer_CalculateDogStatsDPipeName_Integration);
+
+        var traceCacheField = traceIntegrationType.GetField("_cachedTracePipeName", BindingFlags.NonPublic | BindingFlags.Static);
+        var dogstatsdCacheField = dogstatsdIntegrationType.GetField("_cachedDogStatsDPipeName", BindingFlags.NonPublic | BindingFlags.Static);
+
+        traceCacheField?.SetValue(null, tracePipeName);
+        dogstatsdCacheField?.SetValue(null, metricsPipeName);
+    }
```
lucaspimentel: The fact that you had to do this using reflection was not a good sign. 😅 

Do you still need them, though? It looks like you removed `_cachedTracePipeName` and `_cachedDogStatsDPipeName` from the integration classes.

tracer/test/Datadog.Trace.Tests/Configuration/ExporterSettingsAzureFunctionsPipeTests.cs
lucaspimentel: Fixed, thanks!

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
lucaspimentel: > Are you saying we should not append the guid?

Yes, I think that's what he meant. The idea is that we should always honor any explicit user configuration and not change it silently behind the scenes, even if it breaks the tracer.

It's a case of, "you're not doing what I'm telling you to do!"

tracer/test/Datadog.Trace.ClrProfiler.Managed.Tests/AutoInstrumentation/Serverless/ServerlessCompatIntegrationTests.cs
Lewis-E: Didn't still need them!

tracer/src/Datadog.Trace/Configuration/supported-configurations.yaml
Lewis-E: Good call-- added linux default

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
Lewis-E: The tracer should degrade gracefully and only enable pipe transport when it finds version 1.4.0+ or a 0.0.0 dev build. Older compat versions, or missing compat altogether, fall back. So we don't need to wait to merge this.

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
Lewis-E: Alright-- I've updated it so a customer-configured TracesPipeName / MetricsPipeName passes through. Also added comments on my original doc noting this.

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
Lewis-E: Moved to tracer/src/Datadog.Trace/Serverless/

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Serverless/CompatibilityLayer_CalculateDogStatsDPipeName_Integration.cs
Lewis-E: Should just return now.

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
Lewis-E: Yeah, this looks like a minor optimization Claude dropped in. With this #if in place, it avoids doing the actual !RuntimeInformation.IsOSPlatform check and skips that package. Not doing any harm to leave it I think.

Lewis-E:
@lucaspimentel @andrewlock Sorry for the slow back-and-forths here. I've recorded the new setting in telemetry, documented DD_SERVERLESS_COMPAT_PATH, and moved the helpers around. Does GenerateUniquePipeName need to be inlined further & are you happy with the tests? Right now we have one flaky integration test that I don't seem to have permissions to re-run, but other than that, it appears to be passing.

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Serverless/CompatibilityLayer_CalculateDogStatsDPipeName_Integration.cs
```diff
@@ -0,0 +1,87 @@
...
+                // If there was an exception, pass it through
+                return new CallTargetReturn<string>(returnValue);
+            }
+
+            try
+            {
+                // Use the tracer's configured pipe name, which in Azure Functions will be the
+                // unique name generated by ExporterSettings. Fall back to the compat layer's
+                // own calculated name if the tracer isn't initialized yet.
+                var pipeName = Tracer.Instance.Settings?.Manager?.InitialExporterSettings?.MetricsPipeName ?? returnValue;
```
andrewlock: nit: none of these intermediate values can be null:

```suggestion
                var pipeName = Tracer.Instance.Settings.Manager.InitialExporterSettings.MetricsPipeName ?? returnValue;
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Serverless/CompatibilityLayer_CalculateDogStatsDPipeName_Integration.cs
```diff
@@ -0,0 +1,87 @@
...
+            Exception exception,
+            in CallTargetState state)
+        {
+            if (exception is not null)
+            {
+                // If there was an exception, pass it through
+                return new CallTargetReturn<string>(returnValue);
+            }
+
+            try
```
andrewlock: nit, tbh, It might be better to remove can probably remove the `try-catch` here. We catch exceptions escaping from here (and disable the integration) and we report it in telemetry, so that signal might be preferable. try-catch also means less chance of inlining, but this is a once-called thing, so not a big deal

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Serverless/CompatibilityLayer_CalculateDogStatsDPipeName_Integration.cs
```diff
@@ -0,0 +1,87 @@
...
+                // Use the tracer's configured pipe name, which in Azure Functions will be the
+                // unique name generated by ExporterSettings. Fall back to the compat layer's
+                // own calculated name if the tracer isn't initialized yet.
+                var pipeName = Tracer.Instance.Settings?.Manager?.InitialExporterSettings?.MetricsPipeName ?? returnValue;
+
+                Log.Debug(
+                    "ServerlessCompat integration: Overriding compat layer DogStatsD pipe name. " +
+                    "Compat layer calculated: {CompatPipeName}, Tracer using: {TracerPipeName}",
+                    returnValue,
+                    pipeName);
```
andrewlock: Technically you're _not_ overriding the pipe name if `MetricsPipeName` is null here 😄

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Serverless/CompatibilityLayer_CalculateTracePipeName_Integration.cs
```diff
@@ -0,0 +1,87 @@
...
+                // If there was an exception, pass it through
+                return new CallTargetReturn<string>(returnValue);
+            }
+
+            try
+            {
+                // Use the tracer's configured pipe name, which in Azure Functions will be the
+                // unique name generated by ExporterSettings. Fall back to the compat layer's
+                // own calculated name if the tracer isn't initialized yet.
+                var pipeName = Tracer.Instance.Settings?.Manager?.InitialExporterSettings?.TracesPipeName ?? returnValue;
```
andrewlock: Same nits as the other instrumentation 🙂

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -88,12 +92,36 @@ internal ExporterSettings(Raw rawSettings, Func<string, bool> fileExists, IConfi
...
             ValidationWarnings = new List<string>();
 
+            // In Azure Functions (without AAS Site Extension), generate unique default pipe names so
+            // each function instance gets its own pipe even when sharing a hosting plan. A
+            // customer-configured pipe name always takes precedence — we only fabricate one when the
+            // customer has not set one themselves.
+            string? tracesPipeName;
+            string? metricsPipeName;
+
+            if (AzureInfo.Instance.IsAzureFunction
```
andrewlock: I wish we didn't access global state here 🙁 I'm a big proponent of _injecting_ dependencies in the constructor if we need them.

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -88,12 +92,36 @@ internal ExporterSettings(Raw rawSettings, Func<string, bool> fileExists, IConfi
...
+                && !AzureInfo.Instance.IsUsingSiteExtension
+                && ServerlessCompatPipeNameHelper.IsCompatLayerAvailableWithPipeSupport())
+            {
+                tracesPipeName = !StringUtil.IsNullOrEmpty(rawSettings.TracesPipeName)
+                    ? rawSettings.TracesPipeName
+                    : GenerateUniquePipeName("dd_trace", ConfigurationKeys.TracesPipeName);
```
andrewlock: Hmmmm, I see an issue here:

- If the customer configures settings in code, this constructor may be re-executed
- Even if they don't change the metric/trace pipe names, this code will be re-run
- Because we're using a static generator, we'll generate a _new_ unique pipe name
- This pipe name may no longer match the one we told the compat layer.

Somewhere along the line, you'll need to ensure you re-use a previously generated instance. The easiest way to do this (without storing global state that makes testing painful) is to pass an object around that is responsible for generating and _storing_ the pipe names. And then inject that in the constructor.

tracer/src/Datadog.Trace/Serverless/ServerlessCompatPipeNameHelper.cs
```diff
@@ -0,0 +1,120 @@
...
+            // Validate base pipe name length before appending GUID
+            // Windows pipe path format: \\.\pipe\{base}_{guid}
+            // Max total: 256 - 9 (\\.\pipe\) - 1 (underscore) - 32 (GUID) = 214
+            const int maxBaseLength = 214;
+
+            if (baseName.Length > maxBaseLength)
+            {
+                Log.Warning<string, int, int>("{PipeType} pipe base name exceeds {MaxLength} characters ({ActualLength}). Truncating to allow for GUID suffix.", pipeType, maxBaseLength, baseName.Length);
+                baseName = baseName.Substring(0, maxBaseLength);
+            }
```
andrewlock: nit: we explicitly pass in only `dd_trace` or `dd_dogstatsd` so it feels like we can simplify and inline this whole method tbh. But if we _do_ keep it, we should use `AsSpan().Slice(0, maxBaseLength)` to avoid allocations I think

tracer/src/Datadog.Trace/Serverless/ServerlessCompatPipeNameHelper.cs
```diff
@@ -0,0 +1,120 @@
...
+            return uniqueName;
+        }
+
+        /// <summary>
+        /// Checks whether the Datadog Serverless Compat layer is deployed and has a version
+        /// that supports named pipe transport. This is called during ExporterSettings construction
+        /// (before the compat assembly is loaded) so it checks files on disk rather than
+        /// loaded assemblies.
+        /// </summary>
+        internal static bool IsCompatLayerAvailableWithPipeSupport()
```
andrewlock: @lucaspimentel should these checks be here, or should it be on `AzureInfo` with the other Azure Function detection stuff? 🤔

tracer/src/Datadog.Trace/Serverless/ServerlessCompatPipeNameHelper.cs
```diff
@@ -0,0 +1,120 @@
...
+                {
+                    return false;
+                }
+#endif
+
+                // Check that the compat binary exists — it's what actually listens on the named pipe.
+                // DD_SERVERLESS_COMPAT_PATH overrides the default binary location
+                // (matches CompatibilityLayer.cs in datadog-serverless-compat-dotnet).
+                const string defaultCompatBinaryPath = @"C:\home\site\wwwroot\datadog\bin\windows-amd64\datadog-serverless-compat.exe";
+                var compatBinaryPath = Util.EnvironmentHelpers.GetEnvironmentVariable(Configuration.ConfigurationKeys.ServerlessCompatPath)
```
andrewlock: Given that this is a configuration value, should we be reading this using the normal telemetry-enabled configuration readers?

tracer/src/Datadog.Trace/Serverless/ServerlessCompatPipeNameHelper.cs
```diff
@@ -0,0 +1,120 @@
...
+                {
+                    Log.Debug("Did not find Serverless Compatibility Layer or related DLLs.");
+                    return false;
+                }
+
+                var version = getAssemblyVersion(compatDllPath);
+
+                if (version is null)
+                {
+                    Log.Warning("Could not read Serverless Compatibility Layer details at {Path}, using fallback agent communication methods. (No Named Pipes)", compatDllPath);
```
andrewlock: I don't know if this should be a warning? 🤔 

```suggestion
                    Log.Debug("Could not read Serverless Compatibility Layer details at {Path}, using fallback agent communication methods. (No Named Pipes)", compatDllPath);
```

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -88,12 +92,36 @@ internal ExporterSettings(Raw rawSettings, Func<string, bool> fileExists, IConfi
...
 
+            // In Azure Functions (without AAS Site Extension), generate unique default pipe names so
+            // each function instance gets its own pipe even when sharing a hosting plan. A
+            // customer-configured pipe name always takes precedence — we only fabricate one when the
+            // customer has not set one themselves.
+            string? tracesPipeName;
+            string? metricsPipeName;
+
+            if (AzureInfo.Instance.IsAzureFunction
+                && !AzureInfo.Instance.IsUsingSiteExtension
```
andrewlock: We only want this code on Windows, right? We don't want named pipes on linux etc? If so, we should add that check first, as it's cheap

tracer/src/Datadog.Trace/Serverless/ServerlessCompatPipeNameHelper.cs
```diff
@@ -0,0 +1,120 @@
...
+                    Log.Debug("Compat layer version {Version} supports named pipes.", version);
+                    return true;
+                }
+
+                Log.Debug("Compat layer version {Version} does not support named pipes (requires v1.4.0 or greater. Using fallback communication methods.)", version);
+                return false;
+            }
+            catch (Exception ex)
+            {
+                Log.Warning(ex, "Failed to determine Serverless Compatibility layer availability or Named Pipe Support.");
```
andrewlock: This one probably _should_ be an error, no? so that we get telemetry?
```suggestion
                Log.Error(ex, "Failed to determine Serverless Compatibility layer availability or Named Pipe Support.");
```

tracer/test/Datadog.Trace.ClrProfiler.Managed.Tests/AutoInstrumentation/Serverless/ServerlessCompatIntegrationTests.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        result.GetReturnValue().Should().Be(originalValue);
+    }
+
+    [Theory]
+    [InlineData("trace", "dd_trace_from_compat_layer")]
+    [InlineData("dogstatsd", "dd_dogstatsd_from_compat_layer")]
+    public void OnMethodEnd_WhenTracerHasNoPipeName_FallsBackToReturnValue(string pipeType, string compatLayerName)
+    {
+        // In a unit test environment, Tracer.Instance won't have pipe names configured,
```
andrewlock: In general, it's _Strongly_ recommended you don't run code that relies on Tracer.Instance in unit tests. If you want to test this behaviour (which I'd argue would be better tested in an integration test instead), then I'd suggest extracting the code to a separate method, and testing that, and avoiding the `Tracer.Instance` call (which you _really_ can't control here: insert standard thoughts on accessing global state 😄 )

tracer/test/Datadog.Trace.Tests/Configuration/ExporterSettingsAzureFunctionsPipeTests.cs
```diff
@@ -0,0 +1,80 @@
...
+            {
+                settings.MetricsPipeName.Should().Be(expectedPipeName);
+                settings.MetricsTransport.Should().Be(MetricsTransportType.NamedPipe);
+            }
+            else
+            {
+                // When no explicit pipe name is configured and not in Azure Functions,
+                // pipe transport is not used so MetricsPipeName should be null.
+                // In Azure Functions, the constructor generates a unique name which
+                // is covered by integration tests.
```
andrewlock, line 78: I'm not sure it _is_ covered by integration tests 😄 

I don't actually see any "real" integration tests, that does end-to-end with named pipes - do they live in a separate repo? Is there anything more we can add there?

andrewlock (commented):
Thanks! Mostly just a few nits but there's one scenario that I think is a potential bug, where the customer configures in code, we would re-generate a different pipe name, which might mean we inject the wrong value in the compat layer

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
Lewis-E: Ok, I've tried to move the state into a `ServerlessCompatPipeNames` object that can be injected. `AzureInfo.Instance` and `FrameworkDescription.Instance `are still accessed from `SettingsManager.CreatePipeNames` -- is that ok? I made the object immutable, since that seemed consistent with DogStatsdPort / MetricsUnixDomainSocketPath / etc.

tracer/test/Datadog.Trace.Tests/Configuration/ExporterSettingsAzureFunctionsPipeTests.cs
Lewis-E: https://github.com/DataDog/serverless-compat-self-monitoring/pull/54 but I think I agree the comment here can be improved also

tracer/test/Datadog.Trace.ClrProfiler.Managed.Tests/AutoInstrumentation/Serverless/ServerlessCompatIntegrationTests.cs
Lewis-E: Happy to rely on the integration tests in compat-self-monitoring instead, if you feel that's appropriate. (+also expanding them, adding e2e is in the plans)

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
Lewis-E: Added the Windows check as the first gate in `SettingsManager.CreatePipeNames` [here](https://github.com/DataDog/dd-trace-dotnet/pull/8164/changes#diff-e7a15ee8837af54f547fd2ae7a5054adfc08c2e91454d16746fb028b412bef06R144)

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
Lewis-E: See above comment about adding `ServerlessCompatPipeNames` ✔️

tracer/src/Datadog.Trace/Serverless/ServerlessCompatPipeNameHelper.cs
```diff
@@ -0,0 +1,88 @@
...
+using System;
+using System.IO;
+using System.Reflection;
+using Datadog.Trace.Logging;
+
+namespace Datadog.Trace.Serverless
+{
+    /// <summary>
+    /// Checks whether the Datadog Serverless Compat layer is available and supports named-pipe
+    /// transport. Pipe-name generation itself lives in <see cref="ServerlessCompatPipeNames"/>.
```
lucaspimentel: > Pipe-name generation itself lives in `<see cref="ServerlessCompatPipeNames"/>`

Is this comment accurate? `ServerlessCompatPipeNames` looks like a record type with no logic.

tracer/src/Datadog.Trace/Serverless/ServerlessCompatPipeNameHelper.cs
lucaspimentel: I don't feel strongly about it either way, as long as it's in `tracer/src/Datadog.Trace/Serverless/`. It's one called from one place (aside from tests) so it would be very easy to refactor later as needed.

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -76,24 +72,37 @@ internal ExporterSettings(Raw rawSettings, IConfigurationTelemetry telemetry)
...
+            // non-Windows / non-Azure-Function / missing-compat-layer runs).
+            var tracesPipeName = !StringUtil.IsNullOrEmpty(rawSettings.TracesPipeName)
+                ? rawSettings.TracesPipeName
+                : pipeNames.TracesPipeName;
+            var metricsPipeName = !StringUtil.IsNullOrEmpty(rawSettings.MetricsPipeName)
+                ? rawSettings.MetricsPipeName
+                : pipeNames.MetricsPipeName;
+
+            RecordCalculatedPipeName(ConfigurationKeys.TracesPipeName, tracesPipeName, rawSettings.TracesPipeName);
+            RecordCalculatedPipeName(ConfigurationKeys.MetricsPipeName, metricsPipeName, rawSettings.MetricsPipeName);
```
andrewlock: _super_ tiny (optional) nit, I would personally _probably_ combine the "recording" with the actual reading of the setting - extracting a small function which also embeds some logic as to what to record _probably_ isn't a net win IMO. 🤷‍♂️  e.g.


```suggestion
            // Customer-configured pipe names always take precedence; otherwise use whatever the
            // SettingsManager pre-computed for the Serverless Compat layer (may be null on
            // non-Windows / non-Azure-Function / missing-compat-layer runs).
            string? tracesPipeName = null;
            string? metricsPipeName = null;
            
            if (!StringUtil.IsNullOrEmpty(rawSettings.TracesPipeName))
            {
                tracesPipeName = rawSettings.TracesPipeName;
            }
            else if (pipeNames.TracesPipeName is { } pipeName)
            {
                tracesPipeName = pipeName;
                _telemetry.Record(ConfigurationKeys.TracesPipeName, pipeName, recordValue: true, ConfigurationOrigins.Calculated);
            }
            
            if (!StringUtil.IsNullOrEmpty(rawSettings.MetricsPipeName))
            {
                tracesPipeName = rawSettings.MetricsPipeName;
            }
            else if (pipeNames.TracesPipeName is { } metricsName)
            {
                tracesPipeName = pipeName;
                _telemetry.Record(ConfigurationKeys.MetricsPipeName, metricsName, recordValue: true, ConfigurationOrigins.Calculated);
            }
```

tracer/src/Datadog.Trace/Serverless/ServerlessCompatPipeNameHelper.cs
```diff
@@ -0,0 +1,88 @@
...
+            Func<string, bool> fileExists,
+            Func<string, Version?> getAssemblyVersion)
+        {
+            try
+            {
+                // Check that the compat binary exists — it's what actually listens on the named pipe.
+                // DD_SERVERLESS_COMPAT_PATH overrides the default binary location
+                // (matches CompatibilityLayer.cs in datadog-serverless-compat-dotnet).
+                const string defaultCompatBinaryPath = @"C:\home\site\wwwroot\datadog\bin\windows-amd64\datadog-serverless-compat.exe";
+                var compatBinaryPath = !string.IsNullOrEmpty(compatPathOverride) ? compatPathOverride! : defaultCompatBinaryPath;
```
andrewlock: nit: if you use `StringUtil` here, you don't need the `!` dammit
```suggestion
                var compatBinaryPath = !StringUtil.IsNullOrEmpty(compatPathOverride) ? compatPathOverride : defaultCompatBinaryPath;
```

andrewlock (approved):
LGTM, thanks! A couple of remaining nit suggestions, but definitely non-blocking, just feels 😄

tracer/src/Datadog.Trace/Serverless/ServerlessCompatPipeNameHelper.cs
Lewis-E: Fixed this, and a comment in the instrumentation that was also no longer accurate.

PR #8166, opened 2026-02-06
Add `#nullable enable` to UriHelpers

andrewlock:
## Summary of changes

Adds `#nullable enable` to `UriHelpers` and fix usages

## Reason for change

I was working with the file, so decided to cleanup

## Implementation details

Add `#nullable enable`, fix the complaints

## Test coverage

I'm assuming already has coverage

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/AwsApiGatewaySpanFactory.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/AwsApiGatewaySpanFactory.cs
@@ -24,7 +24,7 @@ internal sealed class AwsApiGatewaySpanFactory : IInferredSpanFactory
     {
         try
         {
-            var resourceUrl = UriHelpers.GetCleanUriPath(data.Path).ToLowerInvariant();
+            var resourceUrl = data.Path is null ? string.Empty : UriHelpers.GetCleanUriPath(data.Path).ToLowerInvariant();
 
             var tags = new InferredProxyTags
             {
--- a/tracer/src/Datadog.Trace/Util/UriHelpers.cs
+++ b/tracer/src/Datadog.Trace/Util/UriHelpers.cs
@@ -3,6 +3,8 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
 // </copyright>
 
+#nullable enable
+
 using System;
 
 namespace Datadog.Trace.Util
@@ -47,9 +49,9 @@ namespace Datadog.Trace.Util
         /// <param name="absolutePath">The path to clean</param>
         /// <param name="virtualPathToRemove">The optional virtual path to remove from the front of the path</param>
         /// <returns>The cleaned path</returns>
-        public static string GetCleanUriPath(string absolutePath, string virtualPathToRemove)
+        public static string GetCleanUriPath(string absolutePath, string? virtualPathToRemove)
         {
-            if (string.IsNullOrWhiteSpace(absolutePath) || (absolutePath.Length == 1 && absolutePath[0] == '/'))
+            if (StringUtil.IsNullOrWhiteSpace(absolutePath) || (absolutePath.Length == 1 && absolutePath[0] == '/'))
             {
                 return absolutePath;
             }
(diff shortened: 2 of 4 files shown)


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/AwsApiGatewaySpanFactory.cs
```diff
@@ -24,7 +24,7 @@ internal sealed class AwsApiGatewaySpanFactory : IInferredSpanFactory
     {
         try
         {
-            var resourceUrl = UriHelpers.GetCleanUriPath(data.Path).ToLowerInvariant();
+            var resourceUrl = data.Path is null ? string.Empty : UriHelpers.GetCleanUriPath(data.Path).ToLowerInvariant();
```
andrewlock, line 27: This is technically a change in behaviour, and `""` is a poor fallback, but previously it would just throw so open to alternative suggestions 🤷‍♂️

tracer/src/Datadog.Trace/Util/UriHelpers.cs
```diff
@@ -62,18 +64,20 @@ public static string GetCleanUriPath(string absolutePath, string virtualPathToRe
             // If the virtual path is "/" then we're hosted at the root, so nothing to remove
             // If not, it will be of the form "/myapp", so remove whole thing
             // Make sure we only remove _whole_ segment i.e. /myapp/controller, but not /myappcontroller
-            var hasPrefix = !string.IsNullOrEmpty(virtualPathToRemove)
-                         && virtualPathToRemove != "/"
-                         && virtualPathToRemove[0] == '/'
-                         && absolutePath.StartsWith(virtualPathToRemove, StringComparison.OrdinalIgnoreCase)
-                         && absolutePath.Length > virtualPathToRemove.Length
-                         && absolutePath[virtualPathToRemove.Length] == '/';
+            var prefixLength = !StringUtil.IsNullOrEmpty(virtualPathToRemove)
```
andrewlock, line 67: The switch from `hasPrefix` to `prefixLength` here is because otherwise the compiler couldn't subsequently reason about the fact that `hasPrefix==true` implies `virtualPathToRemove != null`. Doing it in one here resolves the issue without needing `!`

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Proxy/AwsApiGatewaySpanFactory.cs
bouwkast: Okay so looking at the other implementations we were the only ones who were throwing here and not creating a span. The other implementations all create a span, but then they mainly differ from there in the tag data generated for the inferred span they create.

It looks like both Java/JavaScript would concatenate the the name with the literal `null` string for the `http.url` so something like `"domainnull"`
Go uses an empty string
Python skips it if it is null and won't include the tag

So with this we'd match Go. I'd say that is good enough, the RFC doesn't specify what to do in this case.

bouwkast: Okay so looking at the other implementations we were the only ones who were throwing here and not creating a span. The other implementations all create a span, but then they mainly differ from there in the tag data generated for the inferred span they create.

It looks like both Java/JavaScript would concatenate the the name with the literal `null` string for the `http.url` so something like `"domainnull"`
Go uses an empty string
Python skips it if it is null and won't include the tag

So with this we'd match Go. I'd say that is good enough, the RFC doesn't specify what to do in this case.

bouwkast: Okay so looking at the other implementations we were the only ones who were throwing here and not creating a span. The other implementations all create a span, but then they mainly differ from there in the tag data generated for the inferred span they create.

It looks like both Java/JavaScript would concatenate the the name with the literal `null` string for the `http.url` so something like `"domainnull"`
Go uses an empty string
Python skips it if it is null and won't include the tag

So with this we'd match Go. I'd say that is good enough, the RFC doesn't specify what to do in this case.

bouwkast: Okay so looking at the other implementations we were the only ones who were throwing here and not creating a span. The other implementations all create a span, but then they mainly differ from there in the tag data generated for the inferred span they create.

It looks like both Java/JavaScript would concatenate the the name with the literal `null` string for the `http.url` so something like `"domainnull"`
Go uses an empty string
Python skips it if it is null and won't include the tag

So with this we'd match Go. I'd say that is good enough, the RFC doesn't specify what to do in this case FWIW, at least we'd be more consistent as we wouldn't be generating a span prior

lucaspimentel: >  concatenate the the name with the literal `null` string for the `http.url` so something like `"domainnull"`

that sounds like a bug 😅 

or maybe you're misreading the code and that's how they coalesce null values? like `path ?? "null"` in C#?

lucaspimentel: > the RFC doesn't specify what to do in this case

we should fix that?

PR #8167, opened 2026-02-06
Make `ValueStringBuilder` available in `.NET Core 3.1`

andrewlock:
## Summary of changes

Makes `ValueStringBuilder` usable in .NET Core 3.1+, instead of just .NET 6

## Reason for change

I want to use it in some aspnetcore improvements, but currently it's only exposed in .NET 6

## Implementation details

Update the `#if`. Note that I originally made this available in _all_ TFMs, using our vendored spans, but benchmarking showed that this was actively harmful compared to just using `StringBuilderCache`, so to avoid incorrect use, just restricting it for now.

## Test coverage

Updated the tests to cover .NET Core 3.1 too.

## Other details

Part of a stack

- https://github.com/DataDog/dd-trace-dotnet/pull/8167 👈
- https://github.com/DataDog/dd-trace-dotnet/pull/8170

### DIFF
--- a/tracer/src/Datadog.Trace/Util/ValueStringBuilder.cs
+++ b/tracer/src/Datadog.Trace/Util/ValueStringBuilder.cs
@@ -3,17 +3,17 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
 // </copyright>
 
-#if NET6_0_OR_GREATER
+#if NETCOREAPP
+
+#nullable enable
+
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
 using System.Buffers;
-using System.Diagnostics;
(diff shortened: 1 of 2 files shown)


NachoEchevarria:
## Summary

Clean, well-scoped PR. The approach of widening the outer `#if` to `NETCOREAPP` while adding an inner `#if NET6_0_OR_GREATER` guard around `AppendSpanFormattable` (which depends on `ISpanFormattable`) is correct. The removal of the unused `using System.Diagnostics` is a nice cleanup. The `#if !NET` blocks already present inside the file handle API differences for pre-.NET 5 targets, so the code is ready for the wider TFM support.

Only minor observation: the source file uses `#if NETCOREAPP` while the test file uses the more precise `#if NETCOREAPP3_1_OR_GREATER` — a small inconsistency, though functionally equivalent given the current TFMs (`net461;netstandard2.0;netcoreapp3.1;net6.0`). See inline comment for details.

Review by 🤖 Claude Code

PR #8168, opened 2026-02-06
Termination signals fix

### LINKED ISSUE #8165: [BUG]: LifetimeManager's Environment.Exit(0) on SIGTERM prevents host graceful shutdown on .NET 10
### Tracer Version(s)

v3.36.0

### Operating system and platform

Linux

### Instrumentation Mode

Manual

### TFM

net10.0

## Summary

On .NET 10, `LifetimeManager.TerminationSignalHandler` calls `Environment.Exit(0)` upon receiving `SIGTERM`, which immediately initiates process termination before the host's graceful shutdown (`IHost.RunAsync`) completes. This causes hosted services — specifically Microsoft Orleans silos — to not complete their shutdown lifecycle (e.g., membership deregistration, grain deactivation).

## Behavior

| Scenario | `CORECLR_ENABLE_PROFILING` | Shutdown behavior |
|---|---|---|
| Without profiler | `0` | Orleans silo deregisters from membership table cleanly. Silo status transitions to `Dead`. |
| With profiler | `1` | Orleans silo remains `Active` in membership table. Other silos must wait for missed I-Am-Alive checks to declare it dead. |

In both scenarios, the pod terminates in approximately 1 second. The issue is not a timeout — Orleans simply never gets to run its shutdown sequence.

## Likely Culprit

On .NET 10, `LifetimeManager` registers a `PosixSignalRegistration` for `SIGTERM`:

https://github.com/DataDog/dd-trace-dotnet/blob/57f90c182627fa71d289a94795cc5805ac79d042/tracer/src/Datadog.Trace/LifetimeManager.cs#L301-L304

The `Environment.Exit(0)` call bypasses the normal cooperative shutdown that `RunAsync` relies on. The host never gets the chance to stop its services gracefully.

## Workaround

Explicitly call `StopAsync` in a `ProcessExit` handler to block the process from exiting until the host completes shutdown:

```csharp
AppDomain.CurrentDomain.ProcessExit += (s, e) =>
{
    app.StopAsync().GetAwaiter().GetResult();
};
```

This works because `ProcessExit` handlers run sequentially, and blocking here gives Orleans time to complete its lifecycle before the process terminates.

### Reproduction Code

1. Create an ASP.NET Core application with Orleans on .NET 10
2. Deploy to Kubernetes with `CORECLR_ENABLE_PROFILING=1` and the Datadog tracer installed
3. Delete/restart the pod
4. Observe that the silo remains `Active` in the Orleans membership table

tonyredondo:
## Summary of changes

Fixes #8165  an issue of the termination signals.

## Reason for change

Posix handler registration and call to be not deterministic.

## Test coverage

Adapted the tests


tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/LifetimeManager/TerminationSignalTests.cs
```diff
@@ -90,7 +84,7 @@ private async Task RunSigtermTestAsync(int signalCount, bool usePublishWithRid)
             SendTerminationSignals(process, signalCount);
             WaitForExit(process, helper);
 
-            process.ExitCode.Should().Be(0);
+            process.ExitCode.Should().Be(143);
```
andrewlock, line 87: This also seems like a bug fix in and of itself 😄

andrewlock (approved):
Thanks 👍

PR #8171, opened 2026-02-06
[Profiler] Add tracer in the profiling benchmarks

chrisnas:
## Summary of changes
Benchmarks are now looking at tracer only and tracer+profiler scenarios

## Reason for change
Try to understand regression identified when benchmarks were switched from .NET 7.0 to .NET 10 

## Implementation details
Switch env vars 

## Test coverage
N/A

### DIFF
--- a/profiler/build/Allocations.linux.json
+++ b/profiler/build/Allocations.linux.json
@@ -14,7 +14,7 @@
       "name": "Profiler_allocation",
       "environmentVariables": {
         "DD_CLR_ENABLE_NGEN": "true",
-        "DD_PROFILING_ENABLED": "1",
+        "DD_PROFILING_ENABLED": "0",
         "DD_PROFILING_WALLTIME_ENABLED": "0",
         "DD_PROFILING_CPU_ENABLED": "0",
         "DD_PROFILING_EXCEPTION_ENABLED": "0",
@@ -24,7 +24,7 @@
         "DD_THREAD_LIFETIME_ENABLED": "0",
 
         "DD_PROFILING_ALLOCATION_ENABLED": "1",
-        "DD_TRACE_ENABLED" : "0",
+        "DD_TRACE_ENABLED" : "1",
         "DD_PROFILING_MANAGED_ACTIVATION_ENABLED" : "0"
       }
     },
@@ -42,7 +42,7 @@
         "DD_THREAD_LIFETIME_ENABLED": "0",
 
         "DD_PROFILING_ALLOCATION_ENABLED": "1",
-        "DD_TRACE_ENABLED" : "0",
+        "DD_TRACE_ENABLED" : "1",
         "DD_PROFILING_MANAGED_ACTIVATION_ENABLED" : "0"
       }
     }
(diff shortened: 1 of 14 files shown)


profiler/build/Allocations.linux.json
```diff
@@ -14,7 +14,7 @@
       "name": "Profiler_allocation",
       "environmentVariables": {
         "DD_CLR_ENABLE_NGEN": "true",
-        "DD_PROFILING_ENABLED": "1",
+        "DD_PROFILING_ENABLED": "0",
```
gleocadie, line 17: This benchmark is `Profiler_allocation`, but 
1. Profiler is disabled
2. Tracer enabled

Do you think about StableConfig/SSI enabling the profiler ?

That said, I would do another benchmark for that
`Profiler_tracer_allocation`

profiler/build/Allocations.windows.json
```diff
@@ -14,7 +14,7 @@
       "name": "Profiler_allocation",
       "environmentVariables": {
         "DD_CLR_ENABLE_NGEN": "true",
-        "DD_PROFILING_ENABLED": "1",
+        "DD_PROFILING_ENABLED": "0",
```
gleocadie, line 17: same

profiler/build/Contention.linux.json
```diff
@@ -14,7 +14,7 @@
       "name": "Profiler_contention",
       "environmentVariables": {
         "DD_CLR_ENABLE_NGEN": "true",
-        "DD_PROFILING_ENABLED": "1",
+        "DD_PROFILING_ENABLED": "0",
```
gleocadie, line 17: same

profiler/build/Contention.windows.json
```diff
@@ -14,7 +14,7 @@
       "name": "Profiler_contention",
       "environmentVariables": {
         "DD_CLR_ENABLE_NGEN": "true",
-        "DD_PROFILING_ENABLED": "1",
+        "DD_PROFILING_ENABLED": "0",
```
gleocadie, line 17: same

profiler/build/CpuWallTime.linux.json
```diff
@@ -14,7 +14,7 @@
       "name": "Profiler_walltime",
       "environmentVariables": {
         "DD_CLR_ENABLE_NGEN": "true",
-        "DD_PROFILING_ENABLED": "1",
+        "DD_PROFILING_ENABLED": "0",
```
gleocadie, line 17: same

profiler/build/CpuWallTime.windows.json
```diff
@@ -14,7 +14,7 @@
       "name": "Profiler_walltime",
       "environmentVariables": {
         "DD_CLR_ENABLE_NGEN": "true",
-        "DD_PROFILING_ENABLED": "1",
+        "DD_PROFILING_ENABLED": "0",
```
gleocadie, line 17: same

gleocadie (changes requested):
I did not go and comment each line, but the idea is the same
- Create a separate benchmark test for profiler + tracer and keep those profiler only
- in profiler+tracer tests, we should let the profiler enabled or document somewhere that the profiler will be enabled by the tracer. Otherwise it's not obvious

PR #8174, opened 2026-02-07
[Azure Functions] Improve local NuGet build script

lucaspimentel:
## Summary of changes

Improves the `Build-AzureFunctionsNuget.ps1` script with better verbosity control, clearer output, improved error handling, and enhanced NuGet cache cleanup.

## Reason for change

The script previously had:
- Inconsistent verbosity levels and minimal user feedback during execution
- Only removed `Datadog.AzureFunctions` from NuGet cache, missing `Datadog.Serverless.Compat`
- Missing `dotnet restore` step that was sometimes required

## Implementation details

Add missing NuGet restore
- Adds explicit `dotnet restore` step before building the package to ensure dependencies are resolved. This was step is sometimes required and had to be done manually.

Output verbosity
- Adds `-Verbose` parameter support to control verbosity of `dotnet` and `nuke` commands
- Maps PowerShell's `$VerbosePreference` to appropriate verbosity levels for dotnet (`detailed`/`quiet`) and nuke (`normal`/`quiet`)
- Improves output messages with color-coded success indicators using `Write-Host -ForegroundColor Green`
- Miscellaneous message formatting consistency

NuGet cache cleanup
- Refactors NuGet cache cleanup to use a loop over a list of packages (`Datadog.AzureFunctions`, `Datadog.Serverless.Compat`)
- Fixes NuGet cache cleanup to report success/failure clearly

Misc
- Improves error handling when copying the package to destination with proper file existence check

## Test coverage

Tested manually.

## Other details

🤖 Co-Authored-By: Claude Code

### DIFF
--- a/tracer/tools/Build-AzureFunctionsNuget.ps1
+++ b/tracer/tools/Build-AzureFunctionsNuget.ps1
@@ -70,9 +70,8 @@ $globalPackagesPath = & dotnet nuget locals global-packages --list | ForEach-Obj
 }
 Write-Verbose "NuGet global packages path: $globalPackagesPath"
 
-# Remove package Datadog.AzureFunctions from NuGet cache
-$packageId = 'Datadog.AzureFunctions'
-Write-Verbose "Removing $packageId from NuGet cache."
+# Remove packages from NuGet cache
+$packagesToRemove = @('Datadog.AzureFunctions', 'Datadog.Serverless.Compat')
 
 if (-not $globalPackagesPath)
 {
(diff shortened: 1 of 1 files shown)


tracer/tools/Build-AzureFunctionsNuget.ps1
```diff
@@ -77,8 +84,9 @@ else
 
     if (Test-Path $packagePath)
     {
-        Write-Host "Deleting `"$packagePath`"."
-        Remove-Item -Path $packagePath -Recurse -Force -ErrorAction SilentlyContinue
+        Write-Verbose "Deleting `"$packagePath`" from NuGet cache."
+        Remove-Item -Path $packagePath -Recurse -Force # -ErrorAction SilentlyContinue
+        Write-Host "Deleted `"$packagePath`" from NuGet cache." -ForegroundColor Green
```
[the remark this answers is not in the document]
lucaspimentel: You're not wrong, but this is the desired behavior. We want the script to fail if the item where not remove from the cache. We want to ensure the newly-built package is always used.

tracer/tools/Build-AzureFunctionsNuget.ps1
```diff
@@ -59,60 +64,82 @@ Write-Verbose "Tracer directory: $tracerDir"
...
-$packageId = 'Datadog.AzureFunctions'
+# Get NuGet cache path
 $globalPackagesPath = & dotnet nuget locals global-packages --list | ForEach-Object {
     if ($_ -match "global-packages:\s*(.+)$") { $matches[1] }
 }
-Write-Verbose "Global packages path: $globalPackagesPath"
+Write-Verbose "NuGet global packages path: $globalPackagesPath"
+
+# Remove packages from NuGet cache
+$packagesToRemove = @('Datadog.AzureFunctions', 'Datadog.Serverless.Compat')
```
andrewlock, line 74: Yuuuuuk. The "right" thing to do here is to never _add_ them to the global packages cache and to use a local packages cache instead 😬

lucaspimentel: (I tried posting this comment before our standup, but github was having issues)

> The "right" thing to do here is to never add them to the global packages cache

How do I do this? If I build an Azure Function app that references `Datadog.AzureFunctions`, the package will be saved into the nuget's global cache by default.

The issue I'm trying to get around is this (this is the dev edit/build/test loop)
- make changes to `Datadog.Trace`
- build `Datadog.AzureFunctions` package and copy nupkg into a local nuget feed
- build a sample Azure Functions app that references `Datadog.AzureFunctions`
  - `Datadog.AzureFunctions` goes into nuget's global cache. You're saying I can stop this from happening?
- test app, repeat steps

lucaspimentel: Discussed over zoom. You mean something like nuget's previous behavior where it would download packages to a path inside your solution solution directory instead of a global cache.

I'll merge this for now and will try out a few things and update in a separate PR.

CLI:
```
dotnet restore --packages <path>
```

Env var:
```
NUGET_PACKAGES=<path>
```

`nuget.config` file:
```
<configuration>
  <config>
    <add key="globalPackagesFolder" value="<path>" />
  </config>
</configuration>
```

PR #8175, opened 2026-02-07
Cache MessagePack bytes rather than just the bytes

pierotibou:
## Summary of changes
In `SpanMessagePackFormatter` cache Serialized MessagePack bytes (thus with the header) instead of just the Bytes. Take it a step further and generate the byte arrays at compile time with source generators. Also remove AAS tags from `MessagePackStringCache` as they never change.

## Reason for change
Mostly to make the code consistent with `MessagePackStringCache` where we were using that method. I chose that solution instead of impacting `MessagePackStringCache` as It's a micro optim.
Also the cache should contain only values that can change.

## Implementation details
- Moved AAS Tags from `MessagePackStringCache` to `SpanMessagePackFormatter`.
- Use `MessagePackSerializer.Serialize` and `WriteRaw` instead of `StringEncoding.UTF8.GetBytes` and `WriteStringBytes`
- Implement source generators. 
- There's one caveat that I duplicated the `dotnet` constant as TracerConstants is linked in dd-dotnet whereas sourcegenerators aren't. I didn't know if we want to have an almost empty shared project to share, or a extra dependency for dd_dotnet so I kept things simple for now.

## Test coverage
Asked an agent to add a unit test that is testing all values to make sure I didn't forget one case. I didn't plan to commit it as I'm not sure it's valuable in other cases, but I did as a separate commit so it's easy to drop if we don't want to keep it.

## Other details
Benchmark results (ran them 3 times as there was a good variance between runs, after was always better).

Before:
  | Method                                 | Mean     | Error    | StdDev   | Allocated |
  |--------------------------------------- |---------:|--------:|--------:|----------:|                                                                                           
  | 'Serialize simple span (minimal tags)' | 293.7 ns | 3.88 ns | 3.44 ns |         - |                                                                                           
  | 'Serialize span with some tags'        | 363.2 ns | 4.36 ns | 3.41 ns |         - |                                                                                           
  | 'Serialize span with many tags'        | 942.5 ns | 3.41 ns | 3.02 ns |         - |  

After:
  | Method                                 | Mean     | Error    | StdDev   | Allocated |
  |--------------------------------------- |---------:|---------:|---------:|----------:|                                                                                         
  | 'Serialize simple span (minimal tags)' | 238.4 ns |  3.69 ns |  3.50 ns |         - |                                                                                         
  | 'Serialize span with some tags'        | 316.0 ns |  6.56 ns |  6.43 ns |         - |                                                                                         
  | 'Serialize span with many tags'        | 887.7 ns | 17.71 ns | 20.06 ns |         - |


tracer/src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
```diff
@@ -22,79 +23,80 @@ internal sealed class SpanMessagePackFormatter : IMessagePackFormatter<TraceChun
...
-        private readonly byte[] _nameBytes = StringEncoding.UTF8.GetBytes("name");
-        private readonly byte[] _resourceBytes = StringEncoding.UTF8.GetBytes("resource");
-        private readonly byte[] _serviceBytes = StringEncoding.UTF8.GetBytes("service");
-        private readonly byte[] _typeBytes = StringEncoding.UTF8.GetBytes("type");
-        private readonly byte[] _startBytes = StringEncoding.UTF8.GetBytes("start");
-        private readonly byte[] _durationBytes = StringEncoding.UTF8.GetBytes("duration");
-        private readonly byte[] _parentIdBytes = StringEncoding.UTF8.GetBytes("parent_id");
-        private readonly byte[] _errorBytes = StringEncoding.UTF8.GetBytes("error");
-        private readonly byte[] _metaStructBytes = StringEncoding.UTF8.GetBytes("meta_struct");
+        private readonly byte[] _traceIdBytes = MessagePackSerializer.Serialize("trace_id");
```
andrewlock: There's no reason we can't do this at compile time, right, very similar to what we do for `TagList` bytes, by using a small source generator 🤔 Would remove a small amount of startup time theoretically... Also theoretically we could remove some of those allocations in .NET Core+ by using `ReadOnlySpan<byte>` too, but I'm not sure the messagepackserializer.WriteRaw would support that, so maybe not worth it...

tracer/src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
```diff
@@ -957,22 +964,35 @@ private byte[] GetAppSecRulesetVersion(string version)
...
+                _aasSubscriptionIdTagNameBytes = MessagePackSerializer.Serialize(Datadog.Trace.Tags.AzureAppServicesSubscriptionId);
+                _aasResourceIdTagNameBytes = MessagePackSerializer.Serialize(Datadog.Trace.Tags.AzureAppServicesResourceId);
+                _aasInstanceIdTagNameBytes = MessagePackSerializer.Serialize(Datadog.Trace.Tags.AzureAppServicesInstanceId);
+                _aasInstanceNameTagNameBytes = MessagePackSerializer.Serialize(Datadog.Trace.Tags.AzureAppServicesInstanceName);
+                _aasOperatingSystemTagNameBytes = MessagePackSerializer.Serialize(Datadog.Trace.Tags.AzureAppServicesOperatingSystem);
+                _aasRuntimeTagNameBytes = MessagePackSerializer.Serialize(Datadog.Trace.Tags.AzureAppServicesRuntime);
+                _aasExtensionVersionTagNameBytes = MessagePackSerializer.Serialize(Datadog.Trace.Tags.AzureAppServicesExtensionVersion);
+
+                // Cache value bytes (these are constant for the lifetime of the process)
+                _aasSiteNameValueBytes = SerializeIfNotNullOrWhiteSpace(azureAppServiceSettings.SiteName);
```
andrewlock: Arguably we don't need to serialize the "name" keys for each of these, if the "value" isn't going to get serialized. i.e. we _could_ pair each of these with the name above and skip serializing the name. 

However, if we _do_ decide to switch to the compile-time serialization for the tag names, then this becomes unnecessary, so meh.

tracer/test/Datadog.Trace.Tests/Agent/MessagePack/SpanMessagePackFormatterTests.cs
```diff
@@ -567,6 +568,120 @@ public async Task ProcessTags_Serialization(bool propagateProcessTags)
...
+        var settings = new TracerSettings(compositeSource);
+        var agentWriter = new AgentWriter(mockApi, statsAggregator: null, statsd: TestStatsdManager.NoOp, automaticFlush: false);
+        await using var tracer = TracerHelper.Create(settings, agentWriter, sampler: null, scopeManager: null, statsd: null, NullTelemetryController.Instance, NullDiscoveryService.Instance);
+
+        using (_ = tracer.StartActive("test-operation"))
+        {
+        }
+
+        await tracer.FlushAsync();
+        var traceChunks = mockApi.Wait(TimeSpan.FromSeconds(1));
```
andrewlock: Always play it safe in CI 😉 

```suggestion
        var traceChunks = mockApi.Wait(TimeSpan.FromSeconds(30));
```

tracer/test/Datadog.Trace.Tests/Agent/MessagePack/SpanMessagePackFormatterTests.cs
```diff
@@ -567,6 +568,120 @@ public async Task ProcessTags_Serialization(bool propagateProcessTags)
...
+        // ===== Verify metrics (numeric tags) that use cached names =====
+        span.Metrics.Should().ContainKey(Datadog.Trace.Metrics.ProcessId, "process_id metric should be present with cached name");
+        span.Metrics[Datadog.Trace.Metrics.ProcessId].Should().BeGreaterThan(0, "process_id should be valid");
+
+        // For root spans (no parent), _dd.top_level should be 1.0
+        span.Metrics.Should().ContainKey(Datadog.Trace.Metrics.TopLevelSpan, "_dd.top_level metric should be present for root span");
+        span.Metrics[Datadog.Trace.Metrics.TopLevelSpan].Should().Be(1.0, "root span should have _dd.top_level=1.0");
+
+        // ===== Verify span fields that use cached names =====
+        span.Service.Should().NotBeNullOrEmpty("service name should be present");
```
andrewlock, line 597: tiny nit: probably over-asserting a bit in this test IMO - this test is liable to break in the future if we change/refactor things, but it's not too bad.

andrewlock (approved):
Seems reasonable to me, thanks!

tracer/src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
pierotibou: I agree, I thought about it after reading a related comment on SLack. But I've shied away from it. Can give it a try through in a next PR though.

tracer/test/Datadog.Trace.Tests/Agent/MessagePack/SpanMessagePackFormatterTests.cs
pierotibou: Ok I'll simplify it a bit before merging then. Thanks.

pierotibou:
I was surprised to see the degradation on the benchmark. After looking, IIUC those bench run the Sample app each time so it is actually possible to have a degradation of startup time and overall time (even though given the improvement I'm still surprised of). So TLDR, I can ignore this particular feedback even though it advocates for pushing this work at compile time. Right @andrewlock ?

andrewlock:
> I was surprised to see the degradation on the benchmark. After looking, IIUC those bench run the Sample app each time so it is actually possible to have a degradation of startup time and overall time (even though given the improvement I'm still surprised of). So TLDR, I can ignore this particular feedback even though it advocates for pushing this work at compile time. Right @andrewlock ?

Yeah, those are explicitly trying to look at the impact on startup time, by running a relatively short-lived app repeatedly. There is naturally some variance though (which has recently got worse even in the baseline scenario) and so given that the other .NET Framework app and all the other TFMs didn't see a significant hit, I would not be too worried about it I think. 

To be safe, you could try re-running the execution time benchmarks to see if you get the same result. If so, maybe there's something else going on there we need to look into 🤷‍♂️

pierotibou:
> To be safe, you could try re-running the execution time benchmarks to see if you get the same result. If so, maybe there's something else going on there we need to look into 🤷‍♂️

It ran twice as I forced push. Not sure it was the same test, but IIRC it was indeed raising an issue on NET FX the first time as well. Anyway, this PR sits on top of the ProcessTags one, so maybe I'll have time to investigate compilation changes by the time I can merge this one

tracer/src/Datadog.Trace/Agent/MessagePack/SpanMessagePackFormatter.cs
pierotibou: Done in 34f7afe371752b915f9d66980dd708778ec87ef3

tracer/test/Datadog.Trace.Tests/Agent/MessagePack/SpanMessagePackFormatterTests.cs
pierotibou: done in c91512b87e1cfeed5fc40d0ed33b7588f53b1dc1 🙇

tracer/test/Datadog.Trace.Tests/Agent/MessagePack/SpanMessagePackFormatterTests.cs
pierotibou: done in c91512b87e1cfeed5fc40d0ed33b7588f53b1dc1 but I haven't simplified so much, I felt the rest was relevant.

vandonr:
@autotest review

andrewlock:
@codex review

vandonr:
@autotest review

vandonr:
@autotest review

PR #8183, opened 2026-02-10
[Config Registry] feat: migrate supported-configurations to V2 format

genesor:
## Summary of changes

Migrate the file used to power config registry to V2 format (default and type fields)

## Reason for change

The registry needs more data to validate the configurations

### DIFF
--- a/tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigKeyAliasesSwitcherGenerator.cs
+++ b/tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigKeyAliasesSwitcherGenerator.cs
@@ -44,7 +44,7 @@ public class ConfigKeyAliasesSwitcherGenerator : IIncrementalGenerator
             static (spc, result) => Execute(spc, result));
     }
 
-    private static void Execute(SourceProductionContext context, Result<ConfigurationAliases> result)
+    private static void Execute(SourceProductionContext context, Result<ConfigurationAliases?> result)
     {
         // Report any diagnostics first
         foreach (var diagnostic in result.Errors)
@@ -52,21 +52,22 @@ public class ConfigKeyAliasesSwitcherGenerator : IIncrementalGenerator
             context.ReportDiagnostic(Diagnostic.Create(diagnostic.Descriptor, diagnostic.Location?.ToLocation()));
         }
 
-        // Generate source code even if there are errors (use empty configuration as fallback)
+        // Generate source code even if there are errors (use empty configuration as fallback).
+        // `result.Value` can be null when parsing failed; treat that as "no aliases" and generate an empty switch.
         var configurationAliases = result.Value ?? new ConfigurationAliases(new Dictionary<string, string[]>());
         var generatedSource = GenerateConfigurationKeyMatcher(configurationAliases);
         context.AddSource($"{ClassName}.g.cs", SourceText.From(generatedSource, Encoding.UTF8));
     }
 
-    private static Result<ConfigurationAliases> ParseAliasesFromV2File(AdditionalText file, CancellationToken cancellationToken)
+    private static Result<ConfigurationAliases?> ParseAliasesFromV2File(AdditionalText file, CancellationToken cancellationToken)
     {
         try
         {
             var sourceText = file.GetText(cancellationToken);
             if (sourceText is null)
             {
-                return new Result<ConfigurationAliases>(
-                    null!,
+                return new Result<ConfigurationAliases?>(
+                    null,
                     new EquatableArray<DiagnosticInfo>(
                     [
                         CreateDiagnosticInfo("DDSG0003", "Configuration file not found", $"The file '{file.Path}' could not be read. Make sure the supported-configurations.json file exists and is included as an AdditionalFile.", DiagnosticSeverity.Error)
@@ -81,8 +82,8 @@ public class ConfigKeyAliasesSwitcherGenerator : IIncrementalGenerator
             if (!root.TryGetProperty("supportedConfigurations", out var supportedConfigurationsElement) ||
                 supportedConfigurationsElement.ValueKind != JsonValueKind.Object)
             {
-                return new Result<ConfigurationAliases>(
-                    null!,
+                return new Result<ConfigurationAliases?>(
+                    null,
                     new EquatableArray<DiagnosticInfo>(
                     [
                         CreateDiagnosticInfo("DDSG0002", "Aliases parsing error", "Missing or invalid 'supportedConfigurations' section", DiagnosticSeverity.Error)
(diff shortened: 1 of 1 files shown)


tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigKeyAliasesSwitcherGenerator.cs
```diff
@@ -87,93 +75,91 @@ private static Result<string> ExtractAliasesSection(AdditionalText file, Cancell
...
+                    if (string.IsNullOrEmpty(alias))
+                    {
+                        continue;
+                    }
 
-        foreach (var property in root.EnumerateObject())
-        {
-            var mainKey = property.Name;
-            var aliasArray = property.Value;
+                    aliasList ??= new List<string>();
```
anna-git:
```suggestion
                    aliasList ??= [];
```

tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigurationKeysGenerator.cs
```diff
@@ -268,18 +268,30 @@ private static Dictionary<string, ConfigEntry> ParseConfigurationEntries(JsonEle
...
+            // Extract the product field (first non-empty product in the implementations, if any)
             var product = string.Empty;
-            if (value.TryGetProperty("product", out var productElement) &&
-                productElement.ValueKind == JsonValueKind.String)
+            foreach (var implementation in value.EnumerateArray())
             {
-                product = productElement.GetString() ?? string.Empty;
+                if (implementation.ValueKind != JsonValueKind.Object)
+                {
+                    continue;
```
anna-git: we should maybe throw if this happens as that would mean someone didn't respect the structure adding directly a property for example

tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigurationKeysGenerator.cs
```diff
@@ -268,18 +268,30 @@ private static Dictionary<string, ConfigEntry> ParseConfigurationEntries(JsonEle
...
+
+                if (implementation.TryGetProperty("product", out var productElement) &&
+                    productElement.ValueKind == JsonValueKind.String)
+                {
+                    product = productElement.GetString() ?? string.Empty;
+                    if (!string.IsNullOrEmpty(product))
+                    {
+                        break;
+                    }
+                }
```
anna-git:
```suggestion
                 if (implementation.TryGetProperty("product", out var productElement))
                {
                    if (productElement.ValueKind != JsonValueKind.String)
                    {
                        throw new InvalidOperationException($"Configuration entry '{key}' has a 'product' field that is not a string");
                    }

                    product = productElement.GetString();
                    if (string.IsNullOrEmpty(product))
                    {
                        throw new InvalidOperationException($"Configuration entry '{key}' has an empty 'product' field");
                    }

                    break;
                }
```

shouldn't we throw as well if product is not of type string? if someone writes product: 1, we want to prevent that, as well as if the string is empty, if it's going to be empty, we shouldnt add product in the first place probably

tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigurationKeysGenerator.cs
```diff
@@ -268,18 +268,30 @@ private static Dictionary<string, ConfigEntry> ParseConfigurationEntries(JsonEle
...
-            if (value.ValueKind != JsonValueKind.Object)
+            // v2 schema: each entry is an array of implementation objects
+            if (value.ValueKind != JsonValueKind.Array)
             {
-                throw new InvalidOperationException($"Configuration entry '{key}' must be an object");
+                throw new InvalidOperationException($"Configuration entry '{key}' must be an array");
             }
 
-            // Extract the product field if it exists
+            // Extract the product field (first non-empty product in the implementations, if any)
```
anna-git, line 277: this is a bit funny, I'm thinking that product would never change among implementations... but I guess the schema, with arrays for each key, doesn't really allow "head" properties... (just food for thought)

tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigurationKeysGenerator.cs
genesor: Yep, it's better to throw than to have a weird/invalid value in the file.

tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigurationKeysGenerator.cs
genesor: Yup having an array directly under the key is not ideal for this kind of thing. I guess someday (hopefully not soon) we will have a v3 😅

tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigurationKeysGenerator.cs
genesor: I've added throws where needed, if not an array and if not an object.
Its working as expected after modifying the JSON with invalid data. Failing the build when an invalid type is detected.

.gitlab/one-pipeline.locked.yml
```diff
@@ -1,4 +1,4 @@
-# DO NOT EDIT THIS FILE MANUALLY 
+# DO NOT EDIT THIS FILE MANUALLY
```
andrewlock, line 1: 👀 😅

tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigKeyAliasesSwitcherGenerator.cs
```diff
@@ -70,15 +58,15 @@ private static void Execute(SourceProductionContext context, Result<Configuratio
...
     {
         try
         {
             var sourceText = file.GetText(cancellationToken);
             if (sourceText is null)
             {
-                return new Result<string>(
-                    string.Empty,
+                return new Result<ConfigurationAliases>(
+                    null!,
```
andrewlock: nit: having to use `!` is generally a sign that your nullability expectations/attributes are wrong 🙂 In this case, your method signature should be be returning `Result<ConfigurationAliases?>` if it could be null

EDIT: I just noticed the original code was doing that too, ah well 😅

.gitlab/one-pipeline.locked.yml
genesor: I don't really want to handle two PRs in parallel one being automatically generated and mine using the updated file 😅 

This file is updated automatically when clicking (manually) on a gitlab step on libdatadog build

tracer/src/Datadog.Trace.SourceGenerators/Configuration/ConfigKeyAliasesSwitcherGenerator.cs
genesor: I've changed this to handle properly `null`

PR #8185, opened 2026-02-10
Catch FileNotFoundException when accessing assemblies

NachoEchevarria:
## Summary of changes

Add per-assembly try-catch so one failing assembly doesn't prevent finding the OWIN entry assembly

## Reason for change

[Customers on .NET Framework 4.8](https://app.datadoghq.com/error-tracking?query=service%3Ainstrumentation-telemetry-data%20%40lib_language%3Adotnet%20version%3A3.3%2A&et-side=activity&fromUser=false&issue_states=open&order=total_count&refresh_mode=sliding&source=all&sp=%5B%7B%22p%22%3A%7B%22issueId%22%3A%22bb868232-3ae6-11f0-b1f9-da7ad0900002%22%7D%2C%22i%22%3A%22error-tracking-issue%22%7D%5D&from_ts=1770125633120&to_ts=1770730433120&live=true) see `Error: Cannot find entry assembly` with a `FileNotFoundException` from `ModuleHandle.ResolveMethod`. This happens because the OWIN fallback scans all loaded assemblies' `CustomAttributes`, which lazily resolves attribute constructor tokens — and any assembly with an attribute referencing an unloadable dependency throws and aborts the entire scan.

We catch `Exception` rather than just `FileNotFoundException` because reflection over arbitrary assemblies can also throw `FileLoadException`, `TypeLoadException`, `BadImageFormatException`, or `SecurityException`. Skipped assemblies are logged at `Debug` level.

### DIFF
--- a/tracer/src/Datadog.Trace/Ci/CiEnvironment/GithubActionsEnvironmentValues.cs
+++ b/tracer/src/Datadog.Trace/Ci/CiEnvironment/GithubActionsEnvironmentValues.cs
@@ -6,6 +6,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.IO;
 using System.Linq;
@@ -14,7 +15,6 @@ using System.Text.RegularExpressions;
 using Datadog.Trace.Configuration;
 using Datadog.Trace.SourceGenerators;
 using Datadog.Trace.Telemetry.Metrics;
-using Datadog.Trace.VendoredMicrosoftCode.System.Diagnostics.CodeAnalysis;
 using Datadog.Trace.Vendors.Newtonsoft.Json;
 using Datadog.Trace.Vendors.Newtonsoft.Json.Linq;
 


tracer/src/Datadog.Trace/Configuration/EntryAssemblyLocator.cs
```diff
@@ -54,10 +54,20 @@ internal static class EntryAssemblyLocator
             // Therefore, we'll try to find the entry assembly by scanning through all loaded assemblies.
             foreach (var assembly in AppDomain.CurrentDomain.GetAssemblies())
             {
-                if (IsMicrosoftAssembly(assembly!) == false &&
-                    assembly.CustomAttributes.Any(x => x.AttributeType.FullName == "Microsoft.Owin.OwinStartupAttribute"))
+                try
                 {
-                    return assembly;
+                    if (IsMicrosoftAssembly(assembly!) == false &&
```
andrewlock, line 59: We should _probably_ exclude ourselves from this too if possible, though I wonder if it will be dangerous🤔 e.g. if a customer has called their app `Datadog.MyAppthatSendsToDatadog` then we would run into issues 😅

NachoEchevarria: Looks like we already have these exclusions in other places:
  - ExceptionRedactor.cs — name.StartsWith("Datadog.", StringComparison.Ordinal)
  - InstrumentationDefinitions/Sources.cs — same pattern

Maybe better to exclude for " Datadog.Trace*" instead of " Datadog.*"?

NachoEchevarria: I have added a "Datadog.Trace*" check.

PR #8193, opened 2026-02-11
[SymDB] Increase SymDB upload batch default to 1MB

dudikeleti:
## Summary of changes
Update the default SymDB upload batch size from 100KB to 1MB (DD_SYMBOL_DATABASE_BATCH_SIZE_BYTES still overrides).
Update DebuggerSettingsTests to assert against DebuggerSettings.DefaultSymbolBatchSizeInBytes.

## Reason for change
We have to keep number of uploads as low as possible.

## Test coverage
`DebuggerSettingsTests` and existing `SymbolUploaderTest`

### DIFF
--- a/tracer/src/Datadog.Trace/Debugger/Symbols/SymbolsUploader.cs
+++ b/tracer/src/Datadog.Trace/Debugger/Symbols/SymbolsUploader.cs
@@ -349,6 +349,10 @@ namespace Datadog.Trace.Debugger.Symbols
 
         private async Task<bool> SendSymbol(string symbol, SymDbUploadMetadata metadata)
         {
+            // TODO: Consider streaming SymDB payload bytes directly to the stream
+            // to avoid the extra `string -> UTF8 bytes` encoding step and intermediate byte[] allocations/copies.
+            // Similar allocation reductions were done in https://github.com/DataDog/dd-trace-dotnet/pull/8017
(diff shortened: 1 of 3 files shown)


tracer/src/Datadog.Trace/Debugger/Symbols/SymbolsUploader.cs
```diff
@@ -344,6 +344,10 @@ private async Task Upload(StringBuilder builder, int prefixLength, string suffix
 
         private async Task<bool> SendSymbol(string symbol)
         {
+            // TODO: Consider streaming SymDB payload bytes directly to the stream
+            // to avoid the extra `string -> UTF8 bytes` encoding step and intermediate byte[] allocations/copies.
+            // Similar allocation reductions were done in https://github.com/DataDog/dd-trace-dotnet/pull/8017
+            // and https://github.com/DataDog/dd-trace-dotnet/pull/8019).
```
andrewlock, line 355: I would suggest that we should do this _before_ merging this PR as allocating multiple 1MB blocks is likely going to be a source of perf issues...

dudikeleti: https://github.com/DataDog/dd-trace-dotnet/pull/8691

jpbempel (approved):

PR #8196, opened 2026-02-12
Remove allocations for operation name/service name versioning code

andrewlock:
## Summary of changes

Remove allocations from paths that are called for every span creation by hard-coding the required values up-front 

## Reason for change

When doing some profiling work, I noticed a bunch of string allocations. Digging into it, we can pre-compute the required values, allocated them up front, and avoid allocating every time we create a span.

## Implementation details

In summary, the implementations change from doing string interpolation with every call on `v0`:

```csharp
public string GetOperationNameForProtocol(string protocol) =>
    _version switch
    {
        SchemaVersion.V0 => $"{protocol}.request",
        _ => V1Values.ProtocolOperationNames,
    };
```

to creating an enum of "allowed" values with an array of the "updated" values:

```csharp
public enum Protocol
{
    Http,
    Grpc
}

public static readonly string[] ProtocolOperationNames =
[
    "http.request",
    "grpc.request",
];
```

and then the runtime call becomes a simple index into the array:

```csharp
public string GetOperationNameForProtocol(Protocol protocol) => _protocols[(int)protocol];
```

This removes the allocation, _and_ the switch, so it's nice and quick. There are similar patterns and updates for all the `NamingSchema` paths. I focused on the ones that were doing string concatenation, but tried to keep consistent patterns where possible.

Note that in places where we need to choose between a "v1" and "v0" array, I put the array in nested types, to avoid the static initializers from running where the type isn't going to be accessed. I _haven't_ confirmed this works as I'm hoping, but I _think_ it does, and tbh, it's not a big deal if not 😅

All in all, these changes get us some nice benchmarking improvements (the following is testing the `ClientSchema` methods, but I expect similar improvements across the board)

| Method                       | Runtime            |       Mean | Allocated | Alloc Ratio |
| ---------------------------- | ------------------ | ---------: | --------: | ----------: |
| V0_GetOperationName_Original | .NET 10.0          |  9.3603 ns |      48 B |        0.86 |
| V0_GetOperationName_Updated  | .NET 10.0          |  0.3735 ns |         - |        0.00 |
| V0_GetOperationName_Original | .NET 6.0           | 10.8276 ns |      48 B |        0.86 |
| V0_GetOperationName_Updated  | .NET 6.0           |  0.1621 ns |         - |        0.00 |
| V0_GetOperationName_Original | .NET Core 2.1      | 14.7789 ns |      56 B |        1.00 |
| V0_GetOperationName_Updated  | .NET Core 2.1      |  0.0003 ns |         - |        0.00 |
| V0_GetOperationName_Original | .NET Core 3.1      | 14.9634 ns |      48 B |        0.86 |
| V0_GetOperationName_Updated  | .NET Core 3.1      |  0.0888 ns |         - |        0.00 |
| V0_GetOperationName_Original | .NET Framework 4.8 | 12.9418 ns |      56 B |        1.00 |
| V0_GetOperationName_Updated  | .NET Framework 4.8 |  0.0101 ns |         - |        0.00 |
|                              |                    |            |           |             |
| V1_GetOperationName_Original | .NET 10.0          | 11.1258 ns |      64 B |        1.00 |
| V1_GetOperationName_Updated  | .NET 10.0          |  0.2705 ns |         - |        0.00 |
| V1_GetOperationName_Original | .NET 6.0           | 10.7945 ns |      64 B |        1.00 |
| V1_GetOperationName_Updated  | .NET 6.0           |  0.0325 ns |         - |        0.00 |
| V1_GetOperationName_Original | .NET Core 2.1      | 16.0458 ns |      64 B |        1.00 |
| V1_GetOperationName_Updated  | .NET Core 2.1      |  0.0339 ns |         - |        0.00 |
| V1_GetOperationName_Original | .NET Core 3.1      | 16.9130 ns |      64 B |        1.00 |
| V1_GetOperationName_Updated  | .NET Core 3.1      |  0.0085 ns |         - |        0.00 |
| V1_GetOperationName_Original | .NET Framework 4.8 | 13.5547 ns |      64 B |        1.00 |
| V1_GetOperationName_Updated  | .NET Framework 4.8 |  0.0630 ns |         - |        0.00 |
|                              |                    |            |           |             |
| V0_GetServiceName_Original   | .NET 10.0          | 13.9854 ns |      64 B |        0.89 |
| V0_GetServiceName_Updated    | .NET 10.0          |  1.0276 ns |         - |        0.00 |
| V0_GetServiceName_Original   | .NET 6.0           | 16.9312 ns |      64 B |        0.89 |
| V0_GetServiceName_Updated    | .NET 6.0           |  0.0834 ns |         - |        0.00 |
| V0_GetServiceName_Original   | .NET Core 2.1      | 21.3370 ns |      72 B |        1.00 |
| V0_GetServiceName_Updated    | .NET Core 2.1      |  0.0128 ns |         - |        0.00 |
| V0_GetServiceName_Original   | .NET Core 3.1      | 22.9226 ns |      64 B |        0.89 |
| V0_GetServiceName_Updated    | .NET Core 3.1      |  0.0154 ns |         - |        0.00 |
| V0_GetServiceName_Original   | .NET Framework 4.8 | 19.1088 ns |      72 B |        1.00 |
| V0_GetServiceName_Updated    | .NET Framework 4.8 |  0.0018 ns |         - |        0.00 |
|                              |                    |            |           |             |
| V1_GetServiceName_Original   | .NET 10.0          |  0.3646 ns |         - |          NA |
| V1_GetServiceName_Updated    | .NET 10.0          |  0.2475 ns |         - |          NA |
| V1_GetServiceName_Original   | .NET 6.0           |  0.9440 ns |         - |          NA |
| V1_GetServiceName_Updated    | .NET 6.0           |  0.0601 ns |         - |          NA |
| V1_GetServiceName_Original   | .NET Core 2.1      |  1.5292 ns |         - |          NA |
| V1_GetServiceName_Updated    | .NET Core 2.1      |  0.0165 ns |         - |          NA |
| V1_GetServiceName_Original   | .NET Core 3.1      |  1.4162 ns |         - |          NA |
| V1_GetServiceName_Updated    | .NET Core 3.1      |  0.0176 ns |         - |          NA |
| V1_GetServiceName_Original   | .NET Framework 4.8 |  1.3256 ns |         - |          NA |
| V1_GetServiceName_Updated    | .NET Framework 4.8 |  0.0593 ns |         - |          NA |

## Test coverage

To avoid any potential regressions, I refactored the tests to specify the explicit expected output, instead of essentially duplicating the method inside the test. It's a bit more verbose, but gave me more confidence.

Also, the code as written requires that we keep the enums in sync with the arrays. There's a potential failure case where we update one and don't update the other. We _could_ use code-gen to avoid that, but it seems overkill. Instead, I just added some unit tests to ensure that every value of each enum returns _something_

## Other details

I did the benchmarking and the ClientSchema changes manually, used 🤖 Claude to do the same thing for the other schema types, and then changed the stuff I didn't like (half of it 😅)

Spotted while working on https://datadoghq.atlassian.net/browse/LANGPLAT-842

### DIFF
--- a/tracer/src/Datadog.Trace/Configuration/Schema/ClientSchema.cs
+++ b/tracer/src/Datadog.Trace/Configuration/Schema/ClientSchema.cs
@@ -39,7 +39,7 @@ namespace Datadog.Trace.Configuration.Schema
             _serviceNames =
             [
                 useSuffix ? $"{defaultServiceName}-{HttpClientComponent}" : defaultServiceName,
-                useSuffix ? $"{defaultServiceName}-grpc-client" : defaultServiceName,
+                useSuffix ? $"{defaultServiceName}-{GrpcClientComponent}" : defaultServiceName,
             ];
             if (serviceNameMappings is not null)
             {
--- a/tracer/test/Datadog.Trace.Tests/Configuration/Schema/ClientSchemaTests.cs
+++ b/tracer/test/Datadog.Trace.Tests/Configuration/Schema/ClientSchemaTests.cs
@@ -48,9 +48,7 @@ namespace Datadog.Trace.Tests.Configuration.Schema
         public static IEnumerable<(int SchemaVersion, string ExpectedSuffix)> GetOperationNameSuffixForRequestData()
         {
             yield return (0, string.Empty);           // SchemaVersion.V0
-            yield return (0,  string.Empty);
             yield return (1, ".request");   // SchemaVersion.V1
-            yield return (1, ".request");
         }
 
         [Theory]


tracer/src/Datadog.Trace/Configuration/Schema/ClientSchema.cs
```diff
@@ -13,82 +13,100 @@ namespace Datadog.Trace.Configuration.Schema
...
-                SchemaVersion.V0 when !_removeClientServiceNamesEnabled => $"{_defaultServiceName}-{component}",
-                _ => _defaultServiceName,
-            };
+        /// <summary>
+        /// WARNING: when adding new values, you _must_ update the corresponding arrays in <see cref="V0Values"/> and <see cref="V1Values"/>
+        /// </summary>
+        public enum Protocol
+        {
+            Http,
+            Grpc
```
lucaspimentel, line 64: If the backing int values are important (like here), I like to make them explicit:
```
Http = 0,
Grpc = 1
```

This avoids issue is someone inserts a value _not_ at the end. But in this case they also have to update the array so 🤷🏽‍♂️

tracer/src/Datadog.Trace/Configuration/Schema/DatabaseSchema.cs
```diff
@@ -15,86 +16,104 @@ namespace Datadog.Trace.Configuration.Schema
...
-            _version = version;
-            _peerServiceTagsEnabled = peerServiceTagsEnabled;
-            _removeClientServiceNamesEnabled = removeClientServiceNamesEnabled;
-            _defaultServiceName = defaultServiceName;
-            _serviceNameMappings = serviceNameMappings;
-        }
+            _useV0Tags = version == SchemaVersion.V0 && !peerServiceTagsEnabled;
 
-        public string GetOperationName(string databaseType) => $"{databaseType}.query";
+            // Calculate service names once, to avoid allocations with every call
```
lucaspimentel, line 34: Would it make sense to calculate each one on its first call to avoid calculating ones we never use?

tracer/src/Datadog.Trace/Configuration/Schema/DatabaseSchema.cs
```diff
@@ -15,86 +16,104 @@ namespace Datadog.Trace.Configuration.Schema
...
+                    serviceNames[(int)dbType] = mappedName;
+                }
+            }
+        }
+
+        /// <summary>
+        /// WARNING: when adding new values, you _must_ update the corresponding array in <see cref="OperationNames"/>
+        /// and update the service name initialization in the constructor.
+        /// </summary>
+        public enum ServiceType
```
lucaspimentel, line 70: Should we use `static class` with `const int` fields instead of all these `enum` to avoid all the `(int)` casting everywhere?

tracer/src/Datadog.Trace/Configuration/Schema/DatabaseSchema.cs
andrewlock: IMO, no, it's probably not worth it 🤔  Using an index into an array makes this super fast, in addition to removing the amortized allocations. If we want to avoid allocating the strings we don't need, then we need to add a branch for every access, and code to conditionally assign it etc, which makes things more complicated and slower at the expense of potentially 5 long-lived strings?

I could be wrong, but unless this has a measurable impact on startup time, I'd be more inclined to leave it as-it?

tracer/src/Datadog.Trace/Configuration/Schema/ClientSchema.cs
andrewlock: Yeah, and also this risks a classic copy pasta error 😅 

```csharp
Http = 0,
Grpc = 1
NewProtocol = 1
```

> If the backing int values are important (like here), I like to make them explicit

In some ways, the _Values_ aren't important (because they're always the "correct" values from the compiler), but the _order_ is important. My hope is that the unit tests around this are enough to handle and detect where you forget to update the arrays, or you add the item in the wrong place in the array 🤞

tracer/src/Datadog.Trace/Configuration/Schema/DatabaseSchema.cs
andrewlock: I'm not passionate about it either way tbh, but `enum` is nice because you don't have to care about the numbers, just the order, and it's all the same at runtime.

Happy to go with the majority consensus here

tracer/src/Datadog.Trace/Configuration/Schema/DatabaseSchema.cs
lucaspimentel: Make sense, thanks.

tracer/src/Datadog.Trace/Configuration/Schema/DatabaseSchema.cs
lucaspimentel: I see now that the enum types are used as parameters in some places like `GetOperationName(OperationType)` below. For this use it's nice to an enum (so people still rely on IntelliSense in this era of AI agents?). So, "meh."

PR #8197, opened 2026-02-12
[Code Origin] DEBUG-5199 Reduce per-span allocations and locks

dudikeleti:
## Summary of changes

- Add code origin tags for entry spans (frame 0 only) to WebTags to avoid allocating the underlying tag list for common server entry spans.
- Reduce per-span locking when adding code-origin tags for entry spans (`_dd.code_origin.*`) to `TagsList`.
- Cache PDB-derived code origin values (file URL + stringified line/column) during per-assembly cache population (stored per endpoint token) to avoid per-span `ToString()` allocations.

## Reason for change

- Code-origin tags are added on a very hot path and were introducing avoidable allocations and overhead (line/column `ToString()` and repeated tag container growth/locking).

## Implementation details

- **Entry span optimization**: code origin for entry spans is represented as WebTags properties for the base 4 tags (type/index/method/type). This is explicitly not a valid representation for exit spans, which can include multiple frames and dynamic keys.
- **Batch tag add fast-path**: `TagBatch` preserves `TagsList.SetTag()` semantics. Callers must keep the critical section small and avoid slow/allocating work while holding the lock.
- **PDB caching**: line/column strings are computed once during per-assembly cache population (per endpoint token).

## Test coverage
`SpanCodeOriginTests`
`EndpointDetectorTests`


tracer/src/Datadog.Trace/Tagging/TagsList.cs
```diff
@@ -299,5 +342,50 @@ protected virtual void WriteAdditionalTags(StringBuilder builder)
...
+            public void Add(string key, string value)
+            {
+                _tags.Add(new KeyValuePair<string, string>(key, value));
+            }
+
+            public void Dispose()
+            {
+                Monitor.Exit(_tags);
+            }
+        }
```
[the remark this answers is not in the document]
dudikeleti: 764d58f839d3a66292e1386409bfd807a3c9d1eb for `default(TagBatch)`

tracer/src/Datadog.Trace/Tagging/TagsList.cs
```diff
@@ -21,6 +21,49 @@ internal class TagsList : ITags
...
+        /// The returned <see cref="TagBatch"/> holds the underlying tag list lock and MUST be disposed (use <c>using</c>).
+        /// Callers should keep the critical section small: do not perform any slow/allocating work while holding the batch.
+        /// The batch is append-only: it does not perform key lookups and will not replace existing keys.
+        /// Callers are responsible for ensuring keys are not already present (or guarding under the same lock).
+        /// </summary>
+        internal TagBatch BeginTagBatch(int additionalTagCount)
+        {
+            if (additionalTagCount < 0)
+            {
+                additionalTagCount = 0;
```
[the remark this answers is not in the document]
dudikeleti: `additionalTagCount` is a best-effort capacity hint. The batch still functions correctly with 0; it just may reserve less capacity. Clamping is intentionally “fail-soft”. Worst case we lose a tiny allocation optimization, we don’t change tag behavior or correctness.

tracer/src/Datadog.Trace/Debugger/SpanCodeOrigin/SpanCodeOrigin.cs
```diff
@@ -106,20 +107,47 @@ private void AddEntrySpanTags(Span span, Type type, MethodInfo method)
                     return;
                 }
 
+                var sp = GetPdbInfo(assembly, method);
```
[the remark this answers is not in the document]
dudikeleti: 672be9f2b99483e42f816154ddafcec84f500572

tylfin:
I see the benchmarks above, but are we able to introduce a unittest that provably shows the reduction in allocations?

tracer/src/Datadog.Trace/Tagging/WebTags.cs
```diff
@@ -32,5 +32,21 @@ internal partial class WebTags : InstrumentationTags, IHasStatusCode
...
+        public string CodeOriginType { get; set; }
+
+        [Tag(Trace.Tags.CodeOriginFrameIndex)]
+        public string CodeOriginFrames0Index { get; set; }
+
+        [Tag(Trace.Tags.CodeOriginFrameMethod)]
+        public string CodeOriginFrames0Method { get; set; }
+
+        [Tag(Trace.Tags.CodeOriginFrameType)]
+        public string CodeOriginFrames0Type { get; set; }
```
andrewlock: I don't think we should add these to the base `WebTags` type, as that increases the allocations of _all_ derived types, whereas we will only ever actually add these to the root span tag type, which will be a limited subset of those 🤔 

Also, we shouldn't do this until _after_ (or ideally, at the same time) as we enable code origins by default 😄

tracer/src/Datadog.Trace/Tagging/TagsList.cs
```diff
@@ -299,5 +341,63 @@ protected virtual void WriteAdditionalTags(StringBuilder builder)
...
+        /// Note: this is a <see langword="struct"/> that contains a reference. Do not copy it, and dispose exactly once.
+        /// </summary>
+        internal readonly struct TagBatch : IDisposable
+        {
+            private readonly List<KeyValuePair<string, string>>? _tags;
+
+            internal TagBatch(List<KeyValuePair<string, string>> tags)
+            {
+                _tags = tags;
+            }
```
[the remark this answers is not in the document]
andrewlock: This does concern me... it's really hard to reason about exactly when these copies might or might not be added

Since we last spoke, I've realised that we _can_ have some simpler alternatives. IN particular, we can create multiple `SetTags` methods, which side-steps the allocation issue, at the expense of some minor inconvenience/duplication implementation wise 🙂 


```csharp
public void SetTags(KeyValuePair<string, string> tag1, KeyValuePair<string, string> tag2);
public void SetTags(KeyValuePair<string, string> tag1, KeyValuePair<string, string> tag2, KeyValuePair<string, string> tag3);
public void SetTags(KeyValuePair<string, string> tag1, KeyValuePair<string, string> tag2, KeyValuePair<string, string> tag3, KeyValuePair<string, string> tag4);

// I don't know if it's worth adding this too - it's much nicer to use and implement
// but can't work with .NET FX/.NET Standard, so may never be used in practice 🤷‍♂️ 
#if NETCOREAPP
public void SetTags(params ReadOnlySpan<KeyValuePair<string, string>> tags);
#endif
```

Internally, these methods would do the same thing as the existing `SetTag`, but would ensure the capacity is sufficient first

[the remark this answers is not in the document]
dudikeleti: fixed in 6414c1c78c8b8af1c035d2d5d353b3febd1cf50a & 12b0e4a8deaf3c457070db40f4fde632f9e286c3

tracer/src/Datadog.Trace/Tagging/WebTags.cs
dudikeleti: It seems that we do add code origin to all derived types. The only one that is still missing is WCF, but I plan to add it soon.
image

I'll take another look.

dudikeleti: > Also, we shouldn't do this until after (or ideally, at the same time) as we enable code origins by default 😄

Well… now is that time 😉

andrewlock: > It seems that we do add code origin to all derived types. The only one that is still missing is WCF, but I plan to add it soon

We don't have ASP.NET support yet do we? 🤔 Also, we don't need it in `AspNetCoreEndpointTags` or `AspNetCoreMvcTags` as those aren't service-entry spans.

Tbh, the Endpoint and MVC Tags _might_ not need to be `WebTags` at all 🤔 we should check whether they really need to be or not. I can look into that separately as part of an effort on aspnetcore perf I'm working on atm

> Well… now is that time 😉

Fair enough, I really meant in the same PR, but meh 😉

dudikeleti: > We don't have ASP.NET support yet do we? 🤔

#8083 

> Also, we don't need it in AspNetCoreEndpointTags or AspNetCoreMvcTags as those aren't service-entry spans.

Got it. Let's start with netcore only in this PR. I'll add the others later.

> Fair enough, I really meant in the same PR, but meh 😉

Ah got it.

dudikeleti: 58f2a67c138dc19a2dd73704f971e9504fbbcb88

tracer/src/Datadog.Trace/Tagging/TagsList.cs
```diff
@@ -21,61 +21,216 @@ internal class TagsList : ITags
...
+
+            var requiredCapacity = list.Count + additionalCountUpperBound;
+            if (list.Capacity < requiredCapacity)
+            {
+                list.Capacity = requiredCapacity;
+            }
+        }
+
+        private List<KeyValuePair<string, string>> GetOrCreateTagsList(int additionalCountUpperBound)
```
andrewlock: nit: I think we should rename `additionalCountUpperBound`, because it's not clear what it does, plus it could lead to incorrect usages. I think something like `defaultCapacity` might be clearer, and defaulting it to 0 if you don't need anything beyond the default?

```suggestion
        private List<KeyValuePair<string, string>> GetOrCreateTagsList(int defaultCapacity = 0)
```

tracer/src/Datadog.Trace/Tagging/TagsList.cs
```diff
@@ -21,61 +21,216 @@ internal class TagsList : ITags
...
-                    {
-                        if (tags[i].Key == key)
-                        {
-                            return tags[i].Value;
-                        }
-                    }
-                }
+                var newTags = additionalCountUpperBound > 0
+                                  ? new List<KeyValuePair<string, string>>(additionalCountUpperBound)
+                                  : new List<KeyValuePair<string, string>>();
```
andrewlock: Given that if you use `new List(capacity: 0)` it does the same thing as `new List()`, I think we can remove the ternary for simplicity🙂 (Also assumes you agree with the renaming above!

```suggestion
                var newTags = new List<KeyValuePair<string, string>>(defaultCapacity);
```

tracer/src/Datadog.Trace/Tagging/TagsList.cs
```diff
@@ -21,61 +21,216 @@ internal class TagsList : ITags
...
+            if (value is null)
+            {
+                if (tags is null)
+                {
+                    return;
+                }
+            }
+            else
+            {
+                tags = GetOrCreateTagsList(additionalCountUpperBound: 1);
```
andrewlock: This is the kind of pattern I'm concerned about. Passing `1` here will create a list with a backing array of length `1`, instead of the default `4`, which would end up causing 2 more array allocations and more work doing the doubling. 

So I think _either_ we add a constant "DefaultCapacity", and pass that in here, or we automatically _add_ `additionalCountUpperBound` to the `DefaultCapacity` inside `GetOrCreateTagsList` (and we would update this to pass in 0 instead).

IMO, changing this callsite to just use the default capacity is probably better, e.g.:


```suggestion
                tags = GetOrCreateTagsList(DefaultCapacity);
```

tracer/src/Datadog.Trace/Tagging/TagsList.cs
```diff
@@ -21,61 +21,216 @@ internal class TagsList : ITags
...
+        /// Sets multiple tags.
+        /// Uses the same semantics as <see cref="SetTag"/> for each tag (replace/remove existing keys).
+        /// </summary>
+        public virtual void SetTags(
+            string key1,
+            string? value1,
+            string key2,
+            string? value2,
+            string key3,
+            string? value3)
```
andrewlock: I'd prefer the API for this used `KeyValuePair<string, string?>`, as it's still pretty easy to use at the callsite (see previous comments) and it neatly pairs the keys and tags. e.g.


```suggestion
        public virtual void SetTags(
            KeyValuePair<string, string?> tag1,
            KeyValuePair<string, string?> tag2,
            KeyValuePair<string, string?> tag3)
```

what's more, we can literally set those values into the List directly. We're creating `KeyValuePair` _somewhere_, so with the multi-tag case, it seems like this would be preferable? 🤷‍♂️

tracer/src/Datadog.Trace/Tagging/TagsList.cs
```diff
@@ -21,61 +21,216 @@ internal class TagsList : ITags
...
+        {
+            var additionalCountUpperBound =
+                CountNotNull(value1) +
+                CountNotNull(value2) +
+                CountNotNull(value3);
+
+            var tags = GetOrCreateTagsList(additionalCountUpperBound);
```
andrewlock: We're only going to add a maximum of 3 tags here, so we should probably just use default capacity:

```suggestion
            var tags = GetOrCreateTagsList(DefaultCapacity);
```

tracer/src/Datadog.Trace/Tagging/TagsList.cs
```diff
@@ -21,61 +21,216 @@ internal class TagsList : ITags
...
+            string key4,
+            string? value4)
+        {
+            var additionalCountUpperBound =
+                CountNotNull(value1) +
+                CountNotNull(value2) +
+                CountNotNull(value3) +
+                CountNotNull(value4);
+
+            var tags = GetOrCreateTagsList(additionalCountUpperBound);
```
andrewlock:
```suggestion
            var tags = GetOrCreateTagsList(DefaultCapacity);
```

tracer/src/Datadog.Trace/Tagging/TagsList.cs
```diff
@@ -21,61 +21,216 @@ internal class TagsList : ITags
         private List<KeyValuePair<string, double>>? _metrics;
         private List<KeyValuePair<string, byte[]>>? _metaStruct;
 
-        public virtual string? GetTag(string key)
+        private static int CountNotNull(string? value) => value is null ? 0 : 1;
+
+        private static void EnsureAdditionalCapacity<T>(List<KeyValuePair<string, T>> list, int additionalCountUpperBound)
+        {
+            if (additionalCountUpperBound <= 0)
```
andrewlock: This would be 100% a programming error, and isn't exposed publicly, so I think we can just remove the check tbh

tracer/src/Datadog.Trace/Tagging/TagsList.cs
```diff
@@ -21,61 +21,216 @@ internal class TagsList : ITags
         private List<KeyValuePair<string, double>>? _metrics;
         private List<KeyValuePair<string, byte[]>>? _metaStruct;
 
-        public virtual string? GetTag(string key)
+        private static int CountNotNull(string? value) => value is null ? 0 : 1;
+
+        private static void EnsureAdditionalCapacity<T>(List<KeyValuePair<string, T>> list, int additionalCountUpperBound)
```
andrewlock: I was thinking about this again, and tl;dr; I'm not sure whether we need this method, or whether it could actually be harmful 😅 

My reasoning is this:
- `List<T>` automatically expands as more items are added
- It uses a default capacity of `4`, and doubles every time it needs more capacity, to ensure we don't allocate too many arrays, or spend too much time copying
- With the current design, after calling this method, if you need to expand the tags list, we guarantee the backing array is _exactly_ the size required to hold the current tags.
- That means if you add other tag afterwards, you _guarantee_ that you're going to need to reallocate a new array

I think _at a minimum_ this method should use the same expansion algorithm as `List<T>` does internally, i.e. do `2 * list.Length`. But in that case, duplicating the internal `List<T>` logic isn't _really_ saving anything compared to the internal expansion? So maybe we should just drop this entirely 😅

tracer/src/Datadog.Trace/Debugger/SpanCodeOrigin/SpanCodeOrigin.cs
```diff
@@ -106,20 +108,57 @@ private void AddEntrySpanTags(Span span, Type type, MethodInfo method)
...
+                            _tags.Method[0],
+                            methodName,
+                            _tags.TypeName[0],
+                            typeFullName,
+                            _tags.File[0],
+                            cached.Url,
+                            _tags.Line[0],
+                            cached.Line,
+                            _tags.Column[0],
+                            cached.Column);
```
andrewlock: I think it would make sense to make this API take KeyValuePair<> instead, as it would be easier to read with target typing 🙂 
```suggestion
                        tagsList.SetTags(
                            new (_tags.Type,"entry"),
                            new (_tags.Index[0],"0"),
                            new (_tags.Method[0], methodName),
                            new (_tags.TypeName[0], typeFullName),
                            new (_tags.File[0], cached.Url),
                            new (_tags.Line[0], cached.Line),
                            new (_tags.Column[0], cached.Column));
```

tracer/src/Datadog.Trace/Debugger/SpanCodeOrigin/SpanCodeOrigin.cs
```diff
@@ -106,20 +108,57 @@ private void AddEntrySpanTags(Span span, Type type, MethodInfo method)
...
+                    {
+                        tagsList.SetTags(
+                            _tags.Type,
+                            "entry",
+                            _tags.Index[0],
+                            "0",
+                            _tags.Method[0],
+                            methodName,
+                            _tags.TypeName[0],
+                            typeFullName);
```
andrewlock:
```suggestion
                        tagsList.SetTags(
                            _tags.Type,
                            "entry",
                            _tags.Index[0],
                            "0",
                            _tags.Method[0],
                            methodName,
                            _tags.TypeName[0],
                            typeFullName);
                        tagsList.SetTags(
                            new (_tags.Type,"entry"),
                            new (_tags.Index[0],"0"),
                            new (_tags.Method[0], methodName),
                            new (_tags.TypeName[0], typeFullName));
```

tracer/src/Datadog.Trace/Debugger/SpanCodeOrigin/SpanCodeOrigin.cs
```diff
@@ -106,20 +108,57 @@ private void AddEntrySpanTags(Span span, Type type, MethodInfo method)
...
+                    }
+                }
+                else
```
andrewlock, line 138: Hmmm, we should probably just add the new methods to `ITags` to be honest. Or remove `ITags` entirely, and just make `span.Tags` always be a `TagsList`! 😄 But the first one is probably easier 😉 Did you try that already?

tracer/test/Datadog.Trace.Tests/Debugger/SpanCodeOriginTests.cs
```diff
@@ -320,4 +345,122 @@ internal class TestController : Microsoft.AspNetCore.Mvc.ControllerBase
...
+                MethodValue,
+                CodeOriginFrames0TypeKey,
+                TypeValue,
+                CodeOriginFrames0FileKey,
+                FileValue,
+                CodeOriginFrames0LineKey,
+                LineValue,
+                CodeOriginFrames0ColumnKey,
+                ColumnValue);
+        }
```
andrewlock: Put another `SetTag()` at the end I think this test fails 😅

tracer/src/Datadog.Trace/Tagging/TagsList.cs
dudikeleti: fixed in b31b245ce0fdcd380fdbb36a6a4065db267f1e9b

tracer/src/Datadog.Trace/Tagging/TagsList.cs
dudikeleti: fixed in b31b245ce0fdcd380fdbb36a6a4065db267f1e9b

tracer/src/Datadog.Trace/Tagging/TagsList.cs
dudikeleti: fixed in b31b245ce0fdcd380fdbb36a6a4065db267f1e9b

tracer/src/Datadog.Trace/Tagging/TagsList.cs
dudikeleti: fixed in b31b245ce0fdcd380fdbb36a6a4065db267f1e9b

tracer/src/Datadog.Trace/Tagging/TagsList.cs
dudikeleti: fixed in b31b245ce0fdcd380fdbb36a6a4065db267f1e9b

tracer/src/Datadog.Trace/Tagging/TagsList.cs
dudikeleti: fixed in b31b245ce0fdcd380fdbb36a6a4065db267f1e9b

tracer/src/Datadog.Trace/Tagging/TagsList.cs
dudikeleti: fixed in b31b245ce0fdcd380fdbb36a6a4065db267f1e9b

tracer/src/Datadog.Trace/Tagging/TagsList.cs
dudikeleti: fixed in b31b245ce0fdcd380fdbb36a6a4065db267f1e9b

> But in that case, duplicating the internal List<T> logic isn't really saving anything compared to the internal expansion? So maybe we should just drop this entirely 😅

Yes, The difference is small, but still slightly better, when we know the size ahead of time, we can skip one growth/reallocation step.

tracer/src/Datadog.Trace/Debugger/SpanCodeOrigin/SpanCodeOrigin.cs
dudikeleti: fixed in b31b245ce0fdcd380fdbb36a6a4065db267f1e9b

tracer/src/Datadog.Trace/Debugger/SpanCodeOrigin/SpanCodeOrigin.cs
dudikeleti: fixed in b31b245ce0fdcd380fdbb36a6a4065db267f1e9b

tracer/src/Datadog.Trace/Tagging/TagsList.cs
andrewlock: Can we? 🤔 If we always start with the default capacity size of `4`, I can't actually see a scenario where we currently can skip a growth step 🤔 _if_ we allowed to add 9 tags at the same time, then I can see it, but otherwise I think this potentially just adds extra work? 🤔

dudikeleti: You’re absolutely right. At the moment, in the code I don’t initialize with 4 (I'll fix that), so 7 also saves one step. But even if I initialize with 4 from the start, we’ll run into the same situation at 9.
Do we want to optimize for this case?
It also creates an issue, for example, if there are already 7 existing tags and we try to add the same 7 again, we’ll end up with a list of 16 even though we didn’t actually add anything. Both scenarios are theoretical. Do you think it’s better to remove this optimization?

andrewlock: > But even if I initialize with 4 from the start, we’ll run into the same situation at 9.

Yeah, but by point is that we _can't_ add 9 tags at once currently, right? So it's a scenario that can't happen I think 😄  

> It also creates an issue, for example, if there are already 7 existing tags and we try to add the same 7 again, we’ll end up with a list of 16 even though we didn’t actually add anything

Oh yeah, that's a good point! 🤔 It's (presumably) going to be quite a rare case (hopefully!), but it does lend credence to it not necessarily being worth the update.

I'm still a bit torn on it tbh - the real question in my mind is whether it _is_ an optimization in practice. _Maybe_ it's safest to just remove it for now, and add it in later, seeing as it's currently blocking the codeorigin work? Adding the additional settag methods is definitely a win, as it avoids having to take and release the lock repeatedly, so that's good, and we can choose to optimize it further if we think it's worth it later. And when we add codeorigin by default, we're going to add those as properties to the ASpNetCore Tags as properties anyway for most of the tags, so this shouldn't be dircetly impacted?

dudikeleti: > but by point is that we can't add 9 tags at once currently, right?
&
> It's (presumably) going to be quite a rare case

Yes, both scenarios are theoretical.


> we can choose to optimize it further if we think it's worth it later

Makes sense. So for now I'll remove `EnsureAdditionalCapacity` method.

dudikeleti: 2f0c5eb7e05a9ba3a067bc8510d358c86e25989a

tracer/src/Datadog.Trace/Tagging/TagsList.cs
```diff
@@ -17,65 +17,195 @@ namespace Datadog.Trace.Tagging
...
-                        {
-                            tags[i] = new KeyValuePair<string, string>(key, value);
-                        }
-
-                        return;
+                        tags.RemoveAt(i);
                     }
+                    else
+                    {
+                        tags[i] = new(tag.Key, tag.Value);
```
andrewlock, line 178: nit: we can just set it directly, right?

EDIT: _ooooh_ the nullability difference, gah. OK, now avoiding the `KeyValuePair<string, string?>` copies makes more sense to me, as I was thinking we were going to need one _eventually_ anyway 😅

PR #8198, opened 2026-02-12
[FFE] Bump OpenFeature shim package to match OpenFeature

daniel-romano-DD:
## Summary of changes

- Updates the `OpenFeature` shim package to match the TFMs of the OpenFeature package directly
- Removes Newtonsoft dependency
- Fixes build process

## Reason for change

- Currently we're targeting a _lower_ TFM than the `OpenFeature` package, which can lead to build warnings and weird dependency behaviours. Given this is meant to be a thin wrapper for `OpenFeature`, we should match the TFMs they use
- We can avoid needing to add an additional dependency by using the vendored version that's already available in the tracer
- Ensure we publish the package when making a release

## Other details

### DIFF
--- a/tracer/src/Datadog.FeatureFlags.OpenFeature/Sdk/FeatureFlagsSdk.cs
+++ b/tracer/src/Datadog.FeatureFlags.OpenFeature/Sdk/FeatureFlagsSdk.cs
@@ -132,37 +132,16 @@ internal static class FeatureFlagsSdk
         }
     }
 
-    private static Value ConvertObject(object? obj)
+    private static Value ConvertObject(object? obj) => obj switch
     {
-        if (obj is Dictionary<string, object?> dic)
-        {
-            return ConvertStructure(dic);
-        }
-        else if (obj is object?[] arr)
-        {
-            return ConvertArray(arr);
-        }
-        else if (obj is long intVal)
-        {
-            return new Value(intVal);
-        }
-        else if (obj is double doubleVal)
-        {
-            return new Value(doubleVal);
-        }
-        else if (obj is string strVal)
-        {
-            return new Value(strVal);
-        }
-        else if (obj is bool boolVal)
-        {
-            return new Value(boolVal);
-        }
-        else
-        {
-            return new Value();
-        }
-    }
+        Dictionary<string, object?> dic => ConvertStructure(dic),
+        object?[] arr => ConvertArray(arr),
+        long intVal => new Value(intVal),
+        double doubleVal => new Value(doubleVal),
+        string strVal => new Value(strVal),
+        bool boolVal => new Value(boolVal),
+        _ => new Value()
+    };
 
     private static Value ConvertStructure(Dictionary<string, object?> structure)
     {


tracer/src/Datadog.FeatureFlags.OpenFeature/Sdk/FeatureFlagsSdk.cs
```diff
@@ -82,7 +81,7 @@ private static ResolutionDetails<T> GetResolutionDetails<T>(Datadog.Trace.Featur
                         null);
         }
 
-        var value = typeof(T) == typeof(Value) ? JsonToValue(evaluation.Value as string) : evaluation.Value!;
+        var value = typeof(T) == typeof(Value) ? JsonToValue(evaluation.Value) : evaluation.Value!;
```
leoromanovsky, line 84: 👍

tracer/src/Datadog.FeatureFlags.OpenFeature/Sdk/FeatureFlagsSdk.cs
```diff
@@ -116,60 +115,63 @@ private static ImmutableMetadata ToMetadata(IDictionary<string, string>? metadat
...
+        }
+        else if (obj is bool boolVal)
+        {
+            return new Value(boolVal);
+        }
+        else
+        {
+            return new Value();
+        }
+    }
```
andrewlock: nit: switch expressions make this much nicer😄 
```suggestion
    private static Value ConvertObject(object? obj) => obj switch
    {
        Dictionary<string, object?> dic => ConvertStructure(dic),
        object?[] arr => ConvertArray(arr),
        long intVal => new Value(intVal),
        double doubleVal => new Value(doubleVal),
        string strVal => new Value(strVal),
        bool boolVal => new Value(boolVal),
        _ => new Value()
    };
```

tracer/src/Datadog.FeatureFlags.OpenFeature/Datadog.FeatureFlags.OpenFeature.csproj
```diff
@@ -10,8 +10,9 @@
     <NoWarn>$(NoWarn);NU5100;NU5128;NU1901;NU1902;NU1903;NU1904</NoWarn>
     <PackageOutputPath Condition="'$(PackageOutputPath)' == '' ">bin\$(Configuration)\packages</PackageOutputPath>
     <SignAssembly>False</SignAssembly>
-    <Version>2.0.0</Version>
-    <AssemblyVersion>2.0.0</AssemblyVersion>
+    <Version>2.0.1</Version>
+    <AssemblyVersion>2.0.1</AssemblyVersion>
+    <TargetFrameworks>net462;netstandard2.0;net6.0;net8.0</TargetFrameworks>
```
andrewlock: Assembly version automatically uses Version, so [no need to duplicate it](https://andrewlock.net/version-vs-versionsuffix-vs-packageversion-what-do-they-all-mean/#assemblyversion)

Also, I think it makes to mention that we're matching the OpenFeature package version, and that these should be updated if the package is updated:

```suggestion
    <!-- These target frameworks should match the values exposed in the OpenFeature package referenced below-->
    <TargetFrameworks>net462;netstandard2.0;net6.0;net8.0</TargetFrameworks>
```

tracer/src/Datadog.FeatureFlags.OpenFeature/Datadog.FeatureFlags.OpenFeature.csproj
```diff
@@ -45,7 +46,6 @@
 
   <ItemGroup>
     <PackageReference Include="OpenFeature" Version="2.0.0" />
```
andrewlock, line 49: Just to reinforce it:

```suggestion
     <!-- If you update this version, be sure to update the TargetFrameworks element above-->
    <PackageReference Include="OpenFeature" Version="2.0.0" />
```

tracer/src/Datadog.Trace/FeatureFlags/FeatureFlagsEvaluator.cs
```diff
@@ -472,18 +473,60 @@ private static string GetMd5Hash(string input)
...
+                    case JTokenType.String:
+                        return token.ToString();
+                    case JTokenType.Boolean:
+                        return (bool)token;
+                    case JTokenType.Null:
+                        return null;
+                    default:
+                        return ConvertObject(null);
+                }
+            }
```
andrewlock: Moooar switch expressions 😄 
```suggestion
            static object? ConvertToken(JToken token) => token.Type switch
            {
                JTokenType.Object => ConvertObject((JObject)token),
                JTokenType.Array => ConvertArray((JArray)token),
                JTokenType.Integer => (long)token,
                JTokenType.Float => (double)token,
                JTokenType.String => token.ToString(),
                JTokenType.Boolean => (bool)token,
                JTokenType.Null => null,
                _ => ConvertObject(null)
            };
```

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/FeatureFlags/FeatureFlagsTests.cs
```diff
@@ -3,31 +3,24 @@
...
-using VerifyXunit;
 using Xunit;
 using Xunit.Abstractions;
 
 #pragma warning disable SA1402 // File may only contain a single type
 
 namespace Datadog.Trace.ClrProfiler.IntegrationTests.FeatureFlags;
 
+#if !NETCOREAPP2_1
+
```
andrewlock: Why? 🤔

andrewlock (commented):
Thanks! a few suggestions, and one question, but LGTM overall 👍

tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/FeatureFlags/FeatureFlagsTests.cs
dromanol: Leftover from tests. Removed

PR #8200, opened 2026-02-13
[Config registry] Add release step for v2

anna-git:
## Summary of changes

Add a step executed on release to update supported configuration ranges in the Feature Parity (FPD) registry, [cf libdatadog-build
](https://github.com/DataDog/libdatadog-build/blob/c9e1bbad77b8ec22018376babd1f2bed4cf0cc9d/scripts/config-inversion/config-inversion-update-supported-range.py)

Basicall it compares this release’s supported-config list to the previous release’s; then update FPD so each config’s supported tracer range (from/to) matches what this release added or dropped.

## Reason for change

Get versions from which configurations are supported in the fpd


anna-git:
> I'm guessing there's no way to test this ahead of a release? 😅

Yes there's a dry mode, I tested it locally, creating a "fake" v3.39.0 tag,  executing `LOCAL_CONFIG_PATH=tracer/src/Datadog.Trace/Configuration/supported-configurations.yaml \ DataDog/libdatadog-build/scripts/config-inversion/config-inversion-update-supported-range.py --dry-run` and it worked so theoretically shouldn't break the next release 😅.

genesor:
> I'm guessing there's no way to test this ahead of a release? 😅

Testing it is a bit of a pickle @andrewlock what we've done when writing the script actually running is run the dry mode and either use old tags to compare the output or disable the git part of the script.

PR #8202, opened 2026-02-13
Remove the `RedisObfuscationUtil.Obfuscate()` method as it's never used

andrewlock:
## Summary of changes

Remove dead code from the trace processor paths obfuscation 

## Reason for change

The code is never called, and it confused me thoroughly, embarrassing me in front of @pierotibou. NEVER AGAIN. 

This was originally part of the client-side-stats implementation, but was never enabled (and couldn't be), and in fact _shouldn't_ be, because the agent applies obfuscation anyway.

## Implementation details

Start deleting things and stop when I run out.

## Test coverage

Deleted some tests, this is just removal so meh


tracer/test/benchmarks/Benchmarks.Trace/TraceProcessorBenchmark.cs
```diff
@@ -22,7 +22,7 @@ public void GlobalSetup()
         {
             _normalizerTraceProcessor = new NormalizerTraceProcessor();
             _trucantorTraceProcessor = new TruncatorTraceProcessor();
-            _obfuscatorTraceProcessor = new ObfuscatorTraceProcessor(true);
+            _obfuscatorTraceProcessor = new ObfuscatorTraceProcessor();
```
pierotibou, line 25: Clever way of improving our benchmarks ;)

PR #8207, opened 2026-02-16
Add retry for IIS ASM tests

NachoEchevarria:
## Summary of changes

As a continuation of the work done in  https://github.com/DataDog/dd-trace-dotnet/pull/8011 and seeing the same flakes in integration_tests_windows_iis in the ASM tests, we should apply the same principle to these integration tests.


build/cmake/FindLibdatadog.cmake
```diff
@@ -4,18 +4,18 @@ endif()
 
 include(FetchContent)
 
-set(LIBDATADOG_VERSION "v25.0.0" CACHE STRING "libdatadog version")
```
andrewlock, line 7: I don't think these changes should be in this PR 😄

NachoEchevarria: Ups, wrong branch. Thanks for the warning!!!

NachoEchevarria: I have reverted the unintended changes

PR #8208, opened 2026-02-16
Disable Ryuk images

NachoEchevarria:
## Summary of changes

Disable Testcontainers Ryuk (Resource Reaper) in CI by setting `TESTCONTAINERS_RYUK_DISABLED=true` in the tests of the docker-compose service

## Reason for change

The Aerospike integration tests (and any other Testcontainers-based tests)[ intermittently fail in CI](https://app.datadoghq.com/ci/test/runs?query=test_level%3Atest%20%40test.service%3Add-trace-dotnet%20%40test.status%3Afail%20%40git.branch%3Amaster&agg_m=count&agg_m_source=base&agg_t=count&currentTab=overview&eventStack=AwAAAZxIR7mu03_XKQAAABhBWnhJVHFPM0FBQmdrOFo0elg3OTZVUnEAAAAkZjE5YzQ4NDctZWZmZi00YTljLTg2YTctMzBiOGI0MTI1MmY5AANnIg&fromUser=false&index=citest&testId=AwAAAZxIR7mu03_XKQAAABhBWnhJVHFPM0FBQmdrOFo0elg3OTZVUnEAAAAkZjE5YzQ4NDctZWZmZi00YTljLTg2YTctMzBiOGI0MTI1MmY5AANnIg&trace=AwAAAZxIR7mu03_XKQAAABhBWnhJVHFPM0FBQmdrOFo0elg3OTZVUnEAAAAkZjE5YzQ4NDctZWZmZi00YTljLTg2YTctMzBiOGI0MTI1MmY5AANnIg&start=1770716185071&end=1771320985071&paused=false) with:
```
Docker.DotNet.DockerImageNotFoundException: No such image: testcontainers/ryuk:0.5.1
```
Ryuk is a sidecar container that Testcontainers starts **by default** before every test container to handle resource cleanup on process exit. Since it needs to be pulled from Docker Hub, it is susceptible to transient failures (rate limiting, network issues), causing flaky test runs.

The CI pipeline already has its own Docker cleanup step (`.azure-pipelines/steps/clean-docker-containers.yml`) that runs `docker rm -f` on all containers after test execution, making Ryuk redundant.

Skipping the Ryuk image pull saves CI time on every Docker-based integration test run.

- [Testcontainers .NET - Custom Configuration](https://dotnet.testcontainers.org/custom_configuration/)
- [Testcontainers .NET - Resource Reaper](https://dotnet.testcontainers.org/api/resource_reaper/)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -502,7 +502,7 @@ services:
       - COSMOSDB_ENDPOINT=https://cosmosdb-emulator:8081
       - TEST_AGENT_HOST=test-agent
       - CONTAINER_HOSTNAME=http://integrationtests
-      - TESTCONTAINERS_RYUK_DISABLED=true
+      - TESTCONTAINERS_RYUK_DISABLED
       - DD_LOGGER_DD_API_KEY
       - DD_LOGGER_DD_TRACE_LOG_DIRECTORY=/project/artifacts/build_data/infra_logs
       - DD_LOGGER_DD_SERVICE
@@ -555,7 +555,7 @@ services:
       - enable_crash_dumps=${enable_crash_dumps:-true}
       - Verify_DisableClipboard=true
       - DiffEngine_Disabled=true
-      - TESTCONTAINERS_RYUK_DISABLED=true
+      - TESTCONTAINERS_RYUK_DISABLED
       - DD_LOGGER_DD_API_KEY
       - DD_LOGGER_DD_TRACE_LOG_DIRECTORY=/project/artifacts/build_data/infra_logs
       - DD_LOGGER_DD_SERVICE
@@ -611,7 +611,7 @@ services:
       - enable_crash_dumps=${enable_crash_dumps:-true}
       - Verify_DisableClipboard=true
       - DiffEngine_Disabled=true
-      - TESTCONTAINERS_RYUK_DISABLED=true
+      - TESTCONTAINERS_RYUK_DISABLED
       - CONTAINER_HOSTNAME=http://integrationtests
       - IncludeAllTestFrameworks
       - DD_LOGGER_DD_API_KEY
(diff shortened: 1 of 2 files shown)


andrewlock (approved):
Interesting 🤔 Makes me wonder if we should actually be pre-pulling and relying on that instead 😅 Maybe not though 🤷‍♂️ 

But regardless, I think we should probably make sure to apply this _everywhere_? 🤔 e.g. set it as a "top level" env var in ultimate-pipeline, and then pass it through to all of the docker-compose services? Otherwise I worry if/when we move to more test containers we'll end up running into the same issue again?

NachoEchevarria:
> Interesting 🤔 Makes me wonder if we should actually be pre-pulling and relying on that instead 😅 Maybe not though 🤷‍♂️
> 
> But regardless, I think we should probably make sure to apply this _everywhere_? 🤔 e.g. set it as a "top level" env var in ultimate-pipeline, and then pass it through to all of the docker-compose services? Otherwise I worry if/when we move to more test containers we'll end up running into the same issue again?

Thanks! I have added it to the other tests as well. Should we try with this setting before considering prepulling? Since we already take care of clean up, this should be ok? WDYT?

docker-compose.yml
```diff
@@ -502,6 +502,7 @@ services:
       - COSMOSDB_ENDPOINT=https://cosmosdb-emulator:8081
       - TEST_AGENT_HOST=test-agent
       - CONTAINER_HOSTNAME=http://integrationtests
+      - TESTCONTAINERS_RYUK_DISABLED=true
```
andrewlock: If we add this to the ultimate-pipeline.yml `variables`, then we should change it in this file to just pass through:

```suggestion
      - TESTCONTAINERS_RYUK_DISABLED
```

The advantage of that is we only have to toggle in one place 🙂

NachoEchevarria: Nice idea, thanks! Done!

PR #8209, opened 2026-02-17
Update `nuget.org` publishing to use trusted publishing

andrewlock:
## Summary of changes

Updates dd-trace-dotnet to publish NuGet packages using [trusted publishing](https://learn.microsoft.com/en-us/nuget/nuget-org/trusted-publishing)

## Reason for change

From [the documentation](https://learn.microsoft.com/en-us/nuget/nuget-org/trusted-publishing):

> Trusted Publishing is a better way to publish NuGet packages. You don’t need to manage long-lived API keys anymore. Instead, you use short-lived credentials issued by your CI/CD system, like GitHub Actions.
>
> This makes your publishing process safer by reducing the risk of leaked credentials. It also makes automation easier because you don’t need to rotate or store secrets. This approach is part of a broader industry shift toward secure, keyless publishing. If you're curious, check out the OpenSSF initiative: https://repos.openssf.org/trusted-publishers-for-all-package-repositories.

## Implementation details

Followed the steps in the documentation (and [this excellent blog post](https://andrewlock.net/easily-publishing-nuget-packages-from-github-actions-with-trusted-publishing/) - recommend reading that for a brief summary 😉). Having used this personally, I found it "just works" but there are some caveats to using it for orgs. 

Firstly, the trust policy is tied to a single account (i.e. mine), though it applies to `datadog`-owned packages. If I get removed from the org, someone else would need to set up a trust policy. The details I used to create the policy are shown below:

image

The trust policy is tied to my account via the `NUGET_TRUSTED_PUBLISHING_USERNAME` github secret. If we need to change to a different trust policy provided by a different nuget user, we should just need to update that secret to point to the new user.

Additionally, I left the legacy `create_draft_release.yml` workflow untouched (as opposed to the _new_ `create_normal_draft_release` and `create_hotfix_draft_release` workflows which use the `_create_draft_release` workflow), so that we have a fallback that keeps using the `NUGET_API_KEY` approach,  in case this doesn't work on the next release for some reason!

## Test coverage

This is all just YOLO, we can't _really_ test it very easily. That said, the current release tries to do a "noop" push to nuget.org of an existing package, to confirm permissions are set correctly, and that is still in place.

WE _could theoretically_ test this, by _heavily_ cutting out all the actual release bits out of `_create_draft_release.yml` and trying to trigger a release, but that's frankly too risky for my liking. Given we still have the fallback release path at the moment, that's good enough for me

### DIFF
--- a/.github/workflows/_create_draft_release.yml
+++ b/.github/workflows/_create_draft_release.yml
@@ -108,7 +108,6 @@ jobs:
            echo "  1. Create a new trusted policy using your nuget user. Follow the example in https://github.com/DataDog/dd-trace-dotnet/pull/8209" >&2
            echo "  2. Replace the NUGET_TRUSTED_PUBLISHING_USERNAME GitHub secret with your nuget.org username (not email) at https://github.com/DataDog/dd-trace-dotnet/settings/secrets/actions " >&2
            echo "  3. Try running this release again" >&2
-            echo "and then try running this release again" >&2
             exit 1
           fi
           


.github/workflows/_create_draft_release.yml
```diff
@@ -91,7 +100,7 @@ jobs:
           fi
 
           echo "Pushing package to test NuGet API key validity"
-          if ! dotnet nuget push "Datadog.Trace.Annotations.1.0.0.nupkg" --api-key ${{ secrets.NUGET_API_KEY }} --skip-duplicate --source https://api.nuget.org/v3/index.json; then
+          if ! dotnet nuget push "Datadog.Trace.Annotations.1.0.0.nupkg" --api-key "${{ steps.login.outputs.NUGET_API_KEY }}" --skip-duplicate --source https://api.nuget.org/v3/index.json; then
```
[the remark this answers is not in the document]
andrewlock: Oops
```suggestion
          if ! dotnet nuget push "Datadog.Trace.Annotations.1.0.0.nupkg" --api-key "${{ steps.nuget_login.outputs.NUGET_API_KEY }}" --skip-duplicate --source https://api.nuget.org/v3/index.json; then
```

.github/workflows/_create_draft_release.yml
```diff
@@ -166,7 +175,7 @@ jobs:
       - name: "Publish nuget packages to nuget.org"
         working-directory: ${{steps.assets.outputs.artifacts_path}}
         run: |
-          dotnet nuget push "*.${{steps.versions.outputs.full_version}}*.nupkg" --api-key ${{ secrets.NUGET_API_KEY }} --source https://api.nuget.org/v3/index.json
+          dotnet nuget push "*.${{steps.versions.outputs.full_version}}*.nupkg" --api-key "${{ steps.login.outputs.NUGET_API_KEY }}" --source https://api.nuget.org/v3/index.json
```
andrewlock:
```suggestion
          dotnet nuget push "*.${{steps.versions.outputs.full_version}}*.nupkg" --api-key "${{ steps.nuget_login.outputs.NUGET_API_KEY }}" --source https://api.nuget.org/v3/index.json
```

.github/workflows/_create_draft_release.yml
andrewlock:
```suggestion
      NUGET_TRUSTED_PUBLISHING_USERNAME:
```

.github/workflows/create_normal_draft_release.yml
```diff
@@ -33,13 +33,14 @@ jobs:
...
           issues: write # change milestones 
+          id-token: write  # enable GitHub OIDC token issuance for this job
         with:
             forced_commit_id: ${{ inputs.forced_commit_id }}
             ignore_gitlab_failures: ${{ inputs.ignore_gitlab_failures }}
             is_hotfix: false
         secrets:
             AZURE_DEVOPS_TOKEN: ${{ secrets.AZURE_DEVOPS_TOKEN }}
-            NUGET_API_KEY: ${{ secrets.NUGET_API_KEY }}
+            NUGET_TRUSTED_PUBLISHING_USERNAME: ${{ secrets.NUGET_TRUSTED_PUBLISHING_USERNAME }}
```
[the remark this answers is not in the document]
andrewlock: 👍 Fixed now, made `NUGET_TRUSTED_PUBLISHING_USERNAME` required instead

.github/workflows/_create_draft_release.yml
```diff
@@ -91,7 +100,7 @@ jobs:
           fi
 
           echo "Pushing package to test NuGet API key validity"
-          if ! dotnet nuget push "Datadog.Trace.Annotations.1.0.0.nupkg" --api-key ${{ secrets.NUGET_API_KEY }} --skip-duplicate --source https://api.nuget.org/v3/index.json; then
+          if ! dotnet nuget push "Datadog.Trace.Annotations.1.0.0.nupkg" --api-key "${{ steps.nuget_login.outputs.NUGET_API_KEY }}" --skip-duplicate --source https://api.nuget.org/v3/index.json; then
             echo "Push failed - the API key is likely not valid" >&2
             echo "Create a new NuGet api key at nuget.org, with glob permissions * to push to the Datadog org and" >&2
             echo "replace the NUGET_API_KEY GitHub secret with the new key at https://github.com/DataDog/dd-trace-dotnet/settings/secrets/actions " >&2
```
andrewlock:
```suggestion
           echo "Push failed - there is likely an issue with authentication." >&2
           echo "This workflow uses Trusted Publishing: https://learn.microsoft.com/en-us/nuget/nuget-org/trusted-publishing" >&2
           echo "The NUGET_TRUSTED_PUBLISHING_USERNAME contains the username that should have the trusted policy configured on nuget.org." >&2
           echo "If this user is no longer part of the datadog organisation on nuget.org: create a new trusted policy using your nuget user " >&2
           echo "  1. Create a new trusted policy using your nuget user. Follow the example in https://github.com/DataDog/dd-trace-dotnet/pull/8209" >&2
           echo "  2. Replace the NUGET_TRUSTED_PUBLISHING_USERNAME GitHub secret with your nuget.org username (not email) at https://github.com/DataDog/dd-trace-dotnet/settings/secrets/actions " >&2
           echo "  3. Try running this release again" >&2
```

.github/workflows/_create_draft_release.yml
```diff
@@ -91,10 +100,14 @@ jobs:
...
+          if ! dotnet nuget push "Datadog.Trace.Annotations.1.0.0.nupkg" --api-key "${{ steps.nuget_login.outputs.NUGET_API_KEY }}" --skip-duplicate --source https://api.nuget.org/v3/index.json; then
+           echo "Push failed - there is likely an issue with authentication." >&2
+           echo "This workflow uses Trusted Publishing: https://learn.microsoft.com/en-us/nuget/nuget-org/trusted-publishing" >&2
+           echo "The NUGET_TRUSTED_PUBLISHING_USERNAME contains the username that should have the trusted policy configured on nuget.org." >&2
+           echo "If this user is no longer part of the datadog organisation on nuget.org: create a new trusted policy using your nuget user " >&2
+           echo "  1. Create a new trusted policy using your nuget user. Follow the example in https://github.com/DataDog/dd-trace-dotnet/pull/8209" >&2
+           echo "  2. Replace the NUGET_TRUSTED_PUBLISHING_USERNAME GitHub secret with your nuget.org username (not email) at https://github.com/DataDog/dd-trace-dotnet/settings/secrets/actions " >&2
+           echo "  3. Try running this release again" >&2
             echo "and then try running this release again" >&2
```
andrewlock:
```suggestion
```

PR #8203, opened 2026-02-13
Reduce allocations in `HttpRequestUtils.GetUrl()`

andrewlock:
## Summary of changes

Reduces the allocations that occur in `HttpRequestUtils.GetUrl()`, particularly on .NET 6+

## Reason for change

While profiling a simple aspnetcore + httprequest app (our aspnetcore benchmark), this showed up as a _relatively_ high source of `string` allocations (6%) so wanted to try to reduce that. It was explicitly the version that takes a `Uri` that showed up, so optimized that somewhat.

## Implementation details

In newer runtimes (.NET 6+), we can use `GetComponents()` to avoid allocating each of the individual _pieces_ of url before formatting. We can also take advantage of the fact this whole path is only valid with absolute Uris, so we know we have a `Host`, and various other minor things.

> On older runtimes, `GetComponents()` _horrendously_ allocates, hence the split implementation 😅 

Otherwise, tried lots of things running various benchmarks 😅  `Uri` is a tricky beast to deal with, because it does a lot of internal caching too. In general though, we'll be dealing with a new `Uri` instance every time we call this method, so that's what I focused on. Benchmarked both approaches though, just to be sure, and we come out ahead in both cases.

<details><summary>Original benchmarks</summary>
<p>

| Method                 | Runtime            |      Mean |     Error | Allocated |
| ---------------------- | ------------------ | --------: | --------: | --------: |
| GetUrl_Original        | .NET 10.0          | 254.13 ns |  5.058 ns |     456 B |
| GetUrl_Updated         | .NET 10.0          | 215.42 ns |  4.340 ns |     288 B |
| GetUrl_Original        | .NET 6.0           | 384.78 ns |  4.289 ns |     440 B |
| GetUrl_Updated         | .NET 6.0           | 357.27 ns |  7.153 ns |     280 B |
| GetUrl_Original        | .NET Core 2.1      | 556.36 ns |  9.929 ns |     544 B |
| GetUrl_Updated         | .NET Core 2.1      | 495.81 ns |  7.232 ns |     472 B |
| GetUrl_Original        | .NET Core 3.1      | 546.39 ns |  7.438 ns |     528 B |
| GetUrl_Updated         | .NET Core 3.1      | 506.67 ns |  8.758 ns |     464 B |
| GetUrl_Original        | .NET Framework 4.8 | 727.18 ns | 10.738 ns |     786 B |
| GetUrl_Updated         | .NET Framework 4.8 | 783.27 ns | 43.809 ns |     706 B |
[...]

PR #8199, opened 2026-02-12
Improve performance for `UriHelpers.CleanUri`

andrewlock:
## Summary of changes

Reduce allocations for `UriHelpers.CleanUri()` that's called as part of HTTP requests (among others)

## Reason for change

The allocations showed up in some profiling so decided to dig in. And Oh Boy, this stuff is _terrible_ on < .NET 6 😅 And it's still terrible after this PR, just less terrible 😅  

## Implementation details

Two main improvements:
- If we need to use a `StringBuilder` to format the `Uri` _anyway_ (because we're stripping identifiers), then use one for the whole `string`, instead of partially building the string, and then doing another allocation.
- Use `GetComponents()`, specifying all the components we need, instead of accessing each of the `Uri` properties ad-hoc. It turns out these properties are actually very expensive, because they trigger a bunch of analysis. Doing it one-shot doesn't avoid that, but it still gives some improvements.

## Test coverage

I wrote a "characterisation" unit test, seeing as we didn't have any, to describe the current behaviour, to ensure I didn't change anything. Benchmarking was done (as shown below), comparing the original implementation with the new implementation, for three different scenarios:

- `CleanUri`: 
```csharp
_uri = new Uri("http://datadoghq.com/some-path");
UriHelpers.CleanUri(_uri, removeScheme: true, tryRemoveIds: true);
```

- `CleanUri_WithIds`:
```csharp
_uriWithIds= new Uri("http://datadoghq.com/some-path/123");
UriHelpers.CleanUri(_uriWithIds, removeScheme: true, tryRemoveIds: true);
```

- `CleanUri_IgnoreIds`:
```csharp
_uriWithIds= new Uri("http://datadoghq.com/some-path/123");
UriHelpers.CleanUri(_uriWithIds, removeScheme: true, tryRemoveIds: false);
```

The benchmark results heavily depend on the `Uri` provided, so it's only the relative changes that are interesting here, comparing original to updated.

> Also note that these took a long time to run, I was on a call for some of it, so the execution times may be a bit wonky, but allocations was what I was focusing on 😅 

<details><summary>Benchmarking code</summary>
<p>

```csharp
[MemoryDiagnoser, GroupBenchmarksBy(BenchmarkLogicalGroupRule.ByCategory), CategoriesColumn]
public class UriHelperBenchmarks
{
    private Uri _uri;
    private Uri _uriWithIds;
    private string _result;

    [GlobalSetup]
    public void GlobalSetup()
    {
[...]

PR #8180, opened 2026-02-09
Improve performance of `AspNetCoreResourceNameHelper.SimplifyRoutePattern`

andrewlock:
## Summary of changes

Switches to using `ValueStringBuilder` and other minor perf improvements

## Reason for change

Same as #8170, we want to improve performance where we can. Unfortunately, we can't easily avoid the enumeration allocation like we did in that PR, so most of the benefits here are simply from using `ValueStringBuilder` and other minor changes.

## Implementation details

Incorporated various changes based on the `SimplifyRoutePattern` that we use in the single-span aspnetcore observer and the changes made in #8170. The gains aren't as high here, because we can't reduce enumeration allocation.

## Test coverage

Covered by existing tests.

Ran benchmarks using the values. In general, there's a slight regression in duration for an improvement in Alloc Ratio. It's not very consistent though, particularly in <.NET Core 3.1. I think it's probably still worth the change, but open to thoughts

| Method   | Runtime       | Has Defaults | Expand Templates |      Mean | Allocated | Alloc Ratio |
| -------- | ------------- | ------------ | ---------------- | --------: | --------: | ----------: |
| Original | .NET 10.0     | False        | False            |  6.605 us |   6.51 KB |        1.00 |
| Updated  | .NET 10.0     | False        | False            |  8.559 us |    4.8 KB |        0.74 |
| Original | .NET 6.0      | False        | False            | 11.466 us |   6.51 KB |        1.00 |
| Updated  | .NET 6.0      | False        | False            | 11.654 us |    4.8 KB |        0.74 |
| Original | .NET Core 3.1 | False        | False            | 13.808 us |   6.51 KB |        1.00 |
| Updated  | .NET Core 3.1 | False        | False            | 15.266 us |    4.8 KB |        0.74 |
| Original | .NET Core 3.0 | False        | False            | 13.962 us |   6.51 KB |        1.00 |
| Updated  | .NET Core 3.0 | False        | False            | 19.757 us |   6.51 KB |        1.00 |
|          |               |              |                  |           |           |             |
| Original | .NET 10.0     | False        | True             |  6.453 us |    6.2 KB |        0.99 |
| Updated  | .NET 10.0     | False        | True             |  8.116 us |   4.65 KB |        0.75 |
| Original | .NET 6.0      | False        | True             | 11.121 us |   6.23 KB |        1.00 |
| Updated  | .NET 6.0      | False        | True             | 11.375 us |   4.65 KB |        0.75 |
[...]

PR #8210, opened 2026-02-17
Minor aspnetcore perf improvements

andrewlock:
## Summary of changes

A few minor improvements to the standard `AspNetCoreDiagnosticObserver`

## Reason for change

Looking into obvious perf improvements for ASP.NET Core, but only a few minor things stood out (apart from related PRs like #8199 and #8203).

## Implementation details

- Reduce size of MVC tags object by not deriving from `WebTags` (we never set those tags anyway)
- Delay creating spanlinks collection if we don't need it
- HttpRoute always matches AspNetCoreRoute, so can make it readonly

## Test coverage

Covered by existing tests, benchmarks show (tiny) allocation gains

## Other details

Relates to https://datadoghq.atlassian.net/browse/LANGPLAT-842

Related PRs:
- https://github.com/DataDog/dd-trace-dotnet/pull/8167
- https://github.com/DataDog/dd-trace-dotnet/pull/8170
- https://github.com/DataDog/dd-trace-dotnet/pull/8180
- https://github.com/DataDog/dd-trace-dotnet/pull/8196
- https://github.com/DataDog/dd-trace-dotnet/pull/8199
- https://github.com/DataDog/dd-trace-dotnet/pull/8203

### DIFF
--- a/Datadog.Trace.Minimal.slnf
+++ b/Datadog.Trace.Minimal.slnf
@@ -2,7 +2,9 @@
   "solution": {
     "path": "Datadog.Trace.sln",
     "projects": [
+      "tracer\\src\\Datadog.Trace.Annotations\\Datadog.Trace.Annotations.csproj",
       "tracer\\src\\Datadog.Trace.ClrProfiler.Managed.Loader\\Datadog.Trace.ClrProfiler.Managed.Loader.csproj",
+      "tracer\\src\\Datadog.Trace.Manual\\Datadog.Trace.Manual.csproj",
       "tracer\\src\\Datadog.Trace.MSBuild\\Datadog.Trace.MSBuild.csproj",
       "tracer\\src\\Datadog.Trace.SourceGenerators\\Datadog.Trace.SourceGenerators.csproj",
       "tracer\\src\\Datadog.Trace.Tools.Analyzers.CodeFixes\\Datadog.Trace.Tools.Analyzers.CodeFixes.csproj",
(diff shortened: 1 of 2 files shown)


tracer/src/Datadog.Trace/DiagnosticListeners/AspNetCoreDiagnosticObserver.cs
NachoEchevarria, line 382:
```suggestion
                var resourceName = request.PathBase.HasValue
                                       ? $"{rootSpanTags.HttpMethod} {request.PathBase.ToUriComponent()}{resourcePathName}"
                                       : $"{rootSpanTags.HttpMethod} {resourcePathName}";
```

Same as line 565

andrewlock: Haha, I literally had that exact commit originally, and removed it 😄 As the formatted components are all strings, and `request.PathBase.ToUriComponent()` already has `if (!request.PathBase.HasValue) { return string.Empty; }`, adding the extra `if` doesn't actually improve things 🙂

tracer/src/Datadog.Trace/Propagators/SpanContextPropagator.cs
NachoEchevarria, line 199:
```suggestion
                ExtractBehavior.Restart => new PropagationContext(default, cumulativeBaggage, null),
```
Instead an empty list, could we just pass null?

andrewlock: Good catch, thanks! 👍 I was going to look at the propagators in a separate PR though, as there's a bunch of things we can do there I think, I'll make sure to include this 🙂

tracer/src/Datadog.Trace/Tagging/AspNetCoreTags.cs
```diff
@@ -19,6 +19,6 @@ internal partial class AspNetCoreTags : WebTags
         public string AspNetCoreRoute { get; set; }
 
         [Tag(Tags.HttpRoute)]
-        public string HttpRoute { get; set; }
+        public string HttpRoute => AspNetCoreRoute;
```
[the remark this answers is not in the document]
andrewlock: Meh, generated code, and shouldn't be set by customer

tracer/src/Datadog.Trace/SpanLink.cs
```diff
@@ -17,8 +17,6 @@ namespace Datadog.Trace;
 /// </summary>
 internal sealed class SpanLink
 {
-    private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor<SpanLink>();
-
     /// <summary>
     /// Initializes a new instance of the <see cref="SpanLink"/> class.
```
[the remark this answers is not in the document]
andrewlock: meh

PR #8211, opened 2026-02-17
[Tracing] Add experimental support for exporting traces as OTLP (.NET 6+)

zacharycmontoya:
## Summary of changes
Adds experimental support for exporting traces using [OTLP](https://github.com/open-telemetry/opentelemetry-proto/blob/v1.9.0/opentelemetry/proto/trace/v1/trace.proto) rather than the Datadog MessagePack protocols. This allows the DD SDK to send traces to an OTel collector rather than a Datadog Trace Agent, with limited support for non-APM products.

This feature is enabled by setting `OTEL_TRACES_EXPORTER=otlp`.

Note: This feature is currently only supported for .NET 6+, and only the `http/json` OTLP protocol is supported at this time. Setting any other protocol value falls back to Datadog encoding with a startup warning.

### Configuration

| Configuration | Details |
|---------------|---------|
| `OTEL_TRACES_EXPORTER=otlp` | Enables the OTLP traces export |
| `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` | See the [OTLP Exporter Configuration docs](https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/) |
| `OTEL_EXPORTER_OTLP_TRACES_HEADERS` | See the [OTLP Exporter Configuration docs](https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/) |
| `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` | See the [OTLP Exporter Configuration docs](https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/) |
| `OTEL_EXPORTER_OTLP_TRACES_TIMEOUT` | See the [OTLP Exporter Configuration docs](https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/) |

## Reason for change
We are seeing an increasing number of scenarios where users have applications instrumented with the OTel SDK sending data to OTel collectors, and they would like to get additional features offered by the DD SDK without needing to update their OTel collector deployments. Although there will be follow-up work, this provides the ability for users to write vendor-neutral API instrumentation *and* emit vendor-neutral telemetry data so DD SDK users don't have to feel locked in when setting up Datadog APM.

### Configuration
All OTLP exporter configuration (Traces and Metrics) is read in `ExporterSettings` and exposed as properties (`OtlpTracesEndpoint`, `OtlpTracesProtocol`, `OtlpTracesHeaders`, `OtlpTracesTimeoutMs`, and the corresponding `OtlpMetrics*` properties). The `TracesEncoding` property (of type `Datadog.Trace.Agent.TracesEncoding`) determines the serialization format: `DatadogV0_4`, `OtlpProtobuf`, or `OtlpJson`. Only `OtlpJson` is currently functional; `OtlpProtobuf` is defined but not yet implemented.

OTLP endpoint resolution is handled separately from Datadog trace transport settings to avoid comingling the two URL calculation paths. The OTLP endpoint logic considers `DD_AGENT_HOST` before falling back to the OTLP default host of `localhost`.

### Serialization
The `SpanBuffer` class was refactored to use a pluggable `ISpanBufferSerializer` interface (methods: `HeaderSize`, `SerializeSpans`, `WriteHeader`, `FinishBody`). Two implementations exist:
- **`SpanBufferMessagePackSerializer`**: The existing Datadog MessagePack serialization, extracted into its own class.
- **`OtlpTracesJsonSerializer`**: Serializes `TraceChunkModel` and `Span` objects into the OTLP `ExportTraceServiceRequest` JSON structure. Key behaviors:
  - Resource attributes (service.name, service.version, deployment.environment.name, telemetry SDK info, git metadata, runtime ID) are emitted via `OtlpMapper.EmitResourceAttributesFromTraceChunk()`
  - Span attributes are capped at 128 per span
  - Span events (128 limit), span links (128 limit), span status, and span kind are all mapped
  - Field names use lowerCamelCase and enums use integer values per the OTLP http/json encoding spec

### Datadog-to-OTLP Mapping
The `OtlpMapper` static class (under `OpenTelemetry/`) handles the translation between Datadog span semantics and OTLP attributes. It determines which tags are resource-level vs. span-level, manages attribute limits, and processes string/double/byte[] tag types via an inner `TagWriter` struct.

### Sampling & Stats Aggregation
When OTLP export is enabled:
- **Sampling**: Only the `PrioritySampler` is used in `ShouldKeepTrace()`, aligning with the OpenTelemetry SDK behavior of exporting based solely on the sampling decision.
- **Default sampling rate**: A global sampling rate of 1.0 is set by default (can be overridden by user configuration), and the `AgentSamplingRule` is omitted since there is no Datadog Agent to communicate sampling rules.
- **`StatsAggregationKey`**: Extended with `IsError` and `IsTopLevel` fields for OTLP mode, allowing distinct histogram timeseries. The Datadog mode constructor continues to set these to false for backwards compatibility.

### Export
- **`ApiOtlp`** (NET6_0_OR_GREATER only): Implements `IApi` for OTLP endpoints with retry logic (up to 5 attempts with exponential backoff starting at 100ms).
- **`ManagedApiOtlp`**: Thread-safe wrapper that enables atomic swapping of `ApiOtlp` instances (for configuration refresh).
- **Traces**: Serialized by `OtlpTracesJsonSerializer` into `SpanBuffer`, then sent by `ApiOtlp.SendTracesAsync()` to the OTLP traces endpoint with custom headers.
- **Metrics (trace stats)**: `SendStatsAsync` is defined but currently returns success immediately — DDSketch-to-OTLP histogram conversion was removed from this PR and will be introduced in a follow-up PR alongside new unit tests.

### Wiring
In `TracerManagerFactory.GetAgentWriter()`, when `TracesEncoding` is `OtlpProtobuf` or `OtlpJson`, the method creates a `ManagedApiOtlp` and a `StatsAggregator` with `isOtlp: true`, then passes both to `AgentWriter`.

## Test coverage
- **Integration test**: `OpenTelemetrySdkTests.SubmitsOtlpTraces` — A parameterized test exercising different package versions, protocols, and configurations. Sends OTLP traces to the dd-apm-test-agent, retrieves payloads, validates trace/span ID formats (32-char/16-char hex), checks resource attribute consistency across requests, normalizes dynamic values, and performs snapshot testing.
- **Unit tests**: Tests for OTLP-specific behaviors including: stats aggregator enablement when exporting OTLP, exporting only sampled spans, and omission of the AgentSamplingRule.
- **Docker**: The docker-compose configuration was updated to wait for test-agent ports 4317 (gRPC) and 4318 (HTTP) for OTLP traffic.

## Follow-up work
- Export trace metrics (APM stats) via OTLP metrics with DDSketch-to-OTLP histogram conversion (removed from this PR — `SendStatsAsync` currently no-ops)
- Assert against trace metrics in the `SubmitsOtlpTraces` integration test
- Stop reading OTLP metrics settings in `TracerSettings` now that they are being read in `ExporterSettings`
- Implement `http/protobuf` OTLP protocol support (enum value exists but not yet functional)

### DIFF
--- a/tracer/build/_build/Build.Steps.cs
+++ b/tracer/build/_build/Build.Steps.cs
@@ -2534,6 +2534,8 @@ partial class Build
                new(@".*An error occurred while sending data to the agent at \\\\\.\\pipe\\trace-.*The operation has timed out.*", RegexOptions.Compiled),
                new(@".*An error occurred while sending data to the agent at \\\\\.\\pipe\\metrics-.*The operation has timed out.*", RegexOptions.Compiled),
                new(@".*Error detecting and reconfiguring git repository for shallow clone. System.IO.FileLoadException.*", RegexOptions.Compiled),
+               // Known errors in OpenTelemetrySdkTests
+               new(@".*An error occurred while sending data to the agent at http://test-agent.*", RegexOptions.Compiled),
                // These are thrown by the CallTargetNativeTests
                new(@".*Exception occurred when calling the CallTarget integration continuation. Datadog.Trace.DuckTyping.DuckTypeException: Throwing a ducktype exception.*"),
                new(@".*Exception occurred when calling the CallTarget integration continuation. System.MissingMethodException: Throwing a missing method exception.*"),


tracer/src/Datadog.Trace/Configuration/supported-configurations-docs.yaml
```diff
@@ -1225,6 +1225,28 @@ OTEL_EXPORTER_OTLP_TIMEOUT: |
...
+
+OTEL_EXPORTER_OTLP_TRACES_HEADERS: |
+  Configuration key to set custom headers for OTLP traces export.
+  Takes precedence over <see cref="ExporterOtlpHeaders"/>.
+  Format: api-key=key,other=value.
+
+OTEL_EXPORTER_OTLP_TRACES_PROTOCOL: |
+  Configuration key to set the OTLP protocol for traces export.
+  Takes precedence over <see cref="ExporterOtlpProtocol"/>.
+  Valid values: http/json, defaults to http/json.
```
ida613: should the valid values be http/json, http/protobuf and grpc?

zacharycmontoya: Eventually yes, but right now I've only implemented `http/json` so I haven't included `http/protobuf` or `grpc`

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -88,17 +89,27 @@ internal ExporterSettings(Raw rawSettings, Func<string, bool> fileExists, IConfi
...
+                        tracesUnixDomainSocketPath: rawSettings.TracesUnixDomainSocketPath);
 
+            TracesEncoding = traceSettings.Encoding;
             TracesTransport = traceSettings.Transport;
             TracesPipeName = traceSettings.PipeName;
             TracesUnixDomainSocketPath = traceSettings.UdsPath;
             AgentUri = traceSettings.AgentUri;
 
+            var otlpTraceSettings = GetOtlpTracesTransport(signalEndpoint: rawSettings.OtlpTracesEndpoint, generalEndpoint: rawSettings.OtlpEndpoint, signalProtocol: rawSettings.OtlpTracesProtocol, generalProtocol: rawSettings.OtlpProtocol);
+            TracesEncoding = otlpTraceSettings.Encoding;
```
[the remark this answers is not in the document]
zacharycmontoya: Fixed in d5a2750abbc14f83fbd5b8b275baf6981af90c85

tracer/src/Datadog.Trace/Agent/ApiOtlp.cs
```diff
@@ -0,0 +1,342 @@
...
+            // optionally injecting a log instance in here for testing purposes
+            _log = log ?? StaticLog;
+            _log.Debug("Creating new ApiOtlp");
+            _sendStats = SendStatsAsyncImpl;
+            _sendTraces = SendTracesAsyncImpl;
+
+            _apiRequestFactory = apiRequestFactory;
+            _tracesEncoding = exporterSettings.TracesEncoding;
+            _tracesEndpoint = exporterSettings.OtlpTracesEndpoint;
+            _statsEndpoint = exporterSettings.OtlpMetricsEndpoint;
```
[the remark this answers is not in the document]
zacharycmontoya: This comment is out-of-date. `OTEL_EXPORTER_OTLP_TRACES_HEADERS` is now used in the OTLP Traces exporter. `OTEL_EXPORTER_OTLP_TRACES_TIMEOUT` is not used yet, but this can be added at a later time.

tracer/src/Datadog.Trace/Configuration/ExporterSettings.Shared.cs
```diff
@@ -111,22 +123,79 @@ private TraceTransportSettings GetTraceTransport(string? agentUri, string? trace
...
+
+            if (!string.IsNullOrEmpty(generalEndpoint))
+            {
+                if (protocol == OtlpProtocol.Grpc && TryGetAgentUriAndTransport(generalEndpoint!, origin, encoding, protocol, out var grpcSettings))
+                {
+                    return grpcSettings;
+                }
+                else if (protocol != OtlpProtocol.Grpc)
+                {
+                    if (signalEndpoint!.EndsWith("/") && TryGetAgentUriAndTransport($"{generalEndpoint!}{defaultHttpRelativePath}", origin, encoding, protocol, out var httpSettings))
```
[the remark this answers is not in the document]
zacharycmontoya: Fixed in 5901718fed72e7a715960a144389cf79a8d8e440

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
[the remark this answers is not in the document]
zacharycmontoya: This is fixed in d5a2750abbc14f83fbd5b8b275baf6981af90c85

tracer/src/Datadog.Trace/Agent/AgentWriter.cs
```diff
@@ -95,12 +96,16 @@ internal AgentWriter(IApi api, IStatsAggregator? statsAggregator, IStatsdManager
             _batchInterval = batchInterval;
             _traceKeepRateCalculator = traceKeepRateCalculator;
 
-            var formatterResolver = SpanFormatterResolver.Instance;
+            ISpanBufferSerializer spanBufferSerializer = api.TracesEncoding switch
+            {
+                TracesEncoding.OtlpJson => new OtlpTracesJsonSerializer(),
+                _ => new SpanBufferMessagePackSerializer(SpanFormatterResolver.Instance),
```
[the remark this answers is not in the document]
zacharycmontoya: This is true, and a good call-out. However, we don't implement support for `grpc` or `http/protobuf` in this PR so the default logic is intended to fallback to regular Datadog MessagePack payloads.

tracer/src/Datadog.Trace/Configuration/ExporterSettings.Shared.cs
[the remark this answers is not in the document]
zacharycmontoya: Fixed in 5901718fed72e7a715960a144389cf79a8d8e440

tracer/src/Datadog.Trace/OpenTelemetry/Metrics/OtlpMetricsSerializer.cs
```diff
@@ -510,20 +497,19 @@ private byte[] SerializeHistogramDataPoint(MetricPoint metric)
...
-            for (int i = 0; i < bounds.Length; i++)
+            for (int i = 0; i < metric.SnapshotBucketBounds.Length; i++)
             {
                 WriteTag(writer, FieldNumbers.HistogramDataPointExplicitBounds, Fixed64);
-                writer.Write(bounds[i]);
+                writer.Write(metric.SnapshotBucketBounds[i]);
             }
 
-            if (metric.SnapshotCount > 0)
+            if (metric.SnapshotCount > 0 & metric.SnapshotMin != double.NaN)
```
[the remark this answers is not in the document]
zacharycmontoya: This comment highlighted an issue with comparing `double.NaN` to itself, but it didn't recommend a useful fix. The right way to express this comparison is using the method [Double.IsNan(double d)](https://learn.microsoft.com/en-us/dotnet/api/system.double.isnan?view=net-10.0#system-double-isnan(system-double))

tracer/src/Datadog.Trace/Agent/ApiOtlp.cs
```diff
@@ -0,0 +1,349 @@
...
+                        // TODO: Should back-propagate retryAfter value
+                        string retryAfter = response.StatusCode == 429 || response.StatusCode == 503 ? response.GetHeader("Retry-After") : null;
+                        _log.Debug<int, string>("Failed to submit {Count} traces. OTLP response header Retry-After={RetryAfter}", numberOfTraces, retryAfter ?? "unspecified");
+                    }
+
+                    // TODO: Telemetry - Record OTLP Traces API error with status code
+                    return finalTry ? SendResult.Failed_DontRetry : SendResult.Failed_CanRetry;
+                }
+
+                if (response.StatusCode == 429 || response.StatusCode == 413 || response.StatusCode == 408)
```
link04: It seems this statement is unreachable and maybe left over from before?

tracer/src/Datadog.Trace/OpenTelemetry/Traces/OtlpTracesJsonSerializer.cs
```diff
@@ -0,0 +1,715 @@
+// <copyright file="OtlpTracesJsonSerializer.cs" company="Datadog">
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2 License.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.Drawing;
```
link04: Seems like this one is not been used so could be taken out 👢

tracer/src/Datadog.Trace/Agent/SpanBufferMessagePackSerializer.cs
```diff
@@ -0,0 +1,58 @@
+// <copyright file="SpanBufferMessagePackSerializer.cs" company="Datadog">
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache 2 License.
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
+// </copyright>
+
+#nullable enable
+
+using System;
+using System.Collections.Generic;
+using System.Drawing;
```
link04: Same as the other class here

tracer/src/Datadog.Trace/OpenTelemetry/Traces/OtlpTracesJsonSerializer.cs
```diff
@@ -0,0 +1,715 @@
...
+            return 0;
+        }
+
+        // jsonWriter.WriteEndArray(); // close repeated Span spans field
+        // jsonWriter.WriteEndObject(); // Close ScopeSpans message
+        // jsonWriter.WriteEndArray(); // close repeated ScopeSpans field
+        // jsonWriter.WriteEndObject(); // Close ResourceSpans message
+        // jsonWriter.WriteEndArray(); // Close repeated ResourceSpans field
+        // jsonWriter.WriteEndObject(); // Close ExportTraceServiceRequest message
+        var buffer = Encoding.UTF8.GetBytes("]}]}]}");
```
link04: This is the same 6 bytes every time, do you think it could be a static readonly byte[] field to avoid the allocation on each flush since is during its path?

tracer/src/Datadog.Trace/OpenTelemetry/Traces/OtlpTracesJsonSerializer.cs
```diff
@@ -0,0 +1,715 @@
...
+    /// <param name="maxSize">Maximum allowed size of the trace chunk</param>
+    /// <returns>UTF-8 encoded JSON byte array</returns>
+    internal int SerializeSpans(ref byte[] bytes, int offset, in TraceChunkModel traceChunk, int spanBufferOffset, int maxSize)
+    {
+        using var memoryStream = new MemoryStream();
+        using var streamWriter = new StreamWriter(memoryStream, EncodingHelpers.Utf8NoBom, bufferSize: 4096, leaveOpen: true);
+        using var jsonWriter = new JsonTextWriter(streamWriter)
+        {
+            CloseOutput = false
+        };
```
link04, line 650: Just adding a **Note** for later: Found this one thanks to the AI help TBH, not blocking: 

This creates a new MemoryStream + StreamWriter + JsonTextWriter per trace chunk on the serialization thread. The MessagePack path avoids this by writing directly into the reused _temporaryBuffer. Since the serializer is only called from one thread at a time, these could be instance fields that get reset between calls instead of recreated.

I Was curious to see what I did before and noticed I do the same pattern on `OtlpMetricsSerializer` and also is done in `Utf8CountingPooledTextWriter`, so I thinking of creating a follow up to clean up there and update it here as well later on.

tracer/src/Datadog.Trace/OpenTelemetry/Traces/OtlpTracesJsonSerializer.cs
zacharycmontoya: Dang, don't know how that got there 😆 I'll remove that

tracer/src/Datadog.Trace/Agent/ApiOtlp.cs
```diff
@@ -0,0 +1,349 @@
...
+        private enum SendResult
+        {
+            Success,
+            Failed_CanRetry,
+            Failed_DontRetry,
+        }
+
+        public TracesEncoding TracesEncoding => _tracesEncoding;
+
+        public Task<bool> Ping() => Task.FromResult(true);
```
bouwkast, line 75: Is this always just supposed to be `true`?

tracer/src/Datadog.Trace/Agent/ApiOtlp.cs
```diff
@@ -0,0 +1,349 @@
...
+            {
+                _log.Debug("Successfully sent APM trace stats to the OTLP metrics endpoint.");
+            }
+            else
+            {
+                // TODO: Telemetry - Record OTLP Metrics API errors for APM trace stats
+            }
+
+            response?.Dispose();
+            return success ? SendResult.Success : SendResult.Failed_DontRetry;
```
bouwkast:
```suggestion
            return success ? SendResult.Success : SendResult.Failed_CanRetry;
```

Looking through this I think this will currently cause the `SendWithRetry` to not work unless it throws above?

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
```diff
@@ -88,17 +89,36 @@ internal ExporterSettings(Raw rawSettings, Func<string, bool> fileExists, IConfi
...
+            OtlpTracesHeaders = tracesHeaders?.Where(kvp => !string.IsNullOrWhiteSpace(kvp.Key) && !string.IsNullOrWhiteSpace(kvp.Value)).ToArray() ?? [];
+            OtlpTracesTimeoutMs = rawSettings.OtlpTracesTimeoutMs;
+
+            if (rawSettings.OtelTracesExporter == "otlp")
+            {
+                TracesEncoding = otlpTraceSettings.OtlpProtocol switch
+                {
+                    OtlpProtocol.HttpProtobuf => TracesEncoding.OtlpProtobuf,
+                    OtlpProtocol.HttpJson => TracesEncoding.OtlpJson,
+                    _ => TracesEncoding.DatadogV0_4,
```
bouwkast, line 119: If I understand this correctly if there is something set (such as grpc) this will fallback to using our Datadog encoding and send it to the Otlp endpoint? 

Should we log a warning or something that an unsupported value was set?

tracer/src/Datadog.Trace/OpenTelemetry/Traces/OtlpTracesJsonSerializer.cs
```diff
@@ -0,0 +1,715 @@
...
+        }
+
+        jsonWriter.Flush();
+        streamWriter.Flush();
+        memoryStream.Flush();
+
+        // Get the length of the written JSON
+        var length = (int)memoryStream.Position;
+
+        if (length - offset >= maxSize)
```
bouwkast: Up above in FinishBody there is this

```csharp
        if (length + offset - HeaderSize >= maxSize)
        {
            // We've already reached the maximum size, give up
            return 0;
        }
```

Is that the correct one? 

But maybe the `EnsureCapacity` below handles this too?

bouwkast: Oh wait, maybe it should be 

```
if (spanBufferOffset + length >= maxSize)
```

tracer/src/Datadog.Trace/OpenTelemetry/OtlpMapper.cs
```diff
@@ -0,0 +1,355 @@
...
+                {
+                    // gap detected, insert zero count for boundary introduced by getBoundaries
+                    counts.Add(0);
+                }
+
+                counts.Add((long)bin.Count);
+                lastBinIndex = binIndex;
+            }
+
+            return counts;
```
bouwkast: I didn't understand `GetBoundaries` / `GetCounts` really so I asked claude to help and it says that it is missing an extra overflow bucket that the spec expects? https://github.com/open-telemetry/opentelemetry-proto/blob/ca839c51f706f5d53bfb46f06c3e90c3af3a52c6/opentelemetry/proto/metrics/v1/metrics.proto#L477-L481

I don't really know if this is accurate, but I figured I'd call it out just in case

```
      // Overflow bucket for values above the last explicit bound (required by OTLP spec:
      // bucket_counts must have exactly explicit_bounds.length + 1 elements)
      counts.Add(0);
```

tracer/src/Datadog.Trace/OpenTelemetry/OtlpMapper.cs
```diff
@@ -0,0 +1,355 @@
...
+            SnapshotBucketBounds = GetBoundaries(bucket.OkSummary).ToArray(),
+        };
+
+        static List<double> GetBoundaries(DDSketch sketch)
+        {
+            var boundaries = new List<double>();
+
+            // OTel boundaries are defined as a sequence of upper bounds. When our sketch contains gaps we
+            // must introduce extra bounds to represent the lower bound of any bins after a gap; otherwise
+            // it would look like that bin covers not just its original range, but the gap as well.
```
bouwkast: Probably a me issue as I haven't worked with this very much but I am having a hard time understanding exactly what this means

tracer/src/Datadog.Trace/OpenTelemetry/Traces/OtlpTracesJsonSerializer.cs
zacharycmontoya: Good call, will do!

tracer/src/Datadog.Trace/OpenTelemetry/Traces/OtlpTracesJsonSerializer.cs
zacharycmontoya: Good insight here. This sounds like an improvement we can make for a future PR 👍🏼

tracer/src/Datadog.Trace/Agent/ApiOtlp.cs
zacharycmontoya: Yes that's correct. This loop for sending stats calls into  Datadog.Trace.OpenTelemetry.Metrics.OtlpExporter.ExportAsync which has already implemented its own retry logic, so once we get here it has either succeeded or attempted all its own retries. I'll add a comment that this is intended.

tracer/src/Datadog.Trace/Agent/ApiOtlp.cs
zacharycmontoya: Yes, since Ping() has just been used as a liveness check on the Datadog Trace Agent so in general this test doesn't make sense for the OTLP trace submission. I'll add a comment in line saying as much

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
zacharycmontoya: So the main switch that sets up our OTLP pipeline is [here](https://github.com/DataDog/dd-trace-dotnet/pull/8211/changes#diff-36064892d3f17aab3edada32d897bbdbd1f69f5cd7b73d89e3b58ea9d239ba3fR291), which only sets up the `ManagedApiOtlp` if `settings.Manager.InitialExporterSettings.TracesEncoding is TracesEncoding.OtlpProtobuf or TracesEncoding.OtlpJson`, so we don't accidentally send Datadog MessagePack to an OTLP endpoint.

However, I can do a better job at enforcing/logging misconfigurations. For now since I'm only exporting JSON, I can also turn HttpProtobuf into Datadog and log this behavior.

zacharycmontoya: Done in 3221cd66d16836341748ff549439bf4850e50e9c. Please resolve if you think this is sufficient

tracer/src/Datadog.Trace/Agent/ApiOtlp.cs
zacharycmontoya: Added comments in b90174fa83cd8a682131fb0c3ed52c06dff3e6fb

tracer/src/Datadog.Trace/Agent/ApiOtlp.cs
zacharycmontoya: Added comments in b90174fa83cd8a682131fb0c3ed52c06dff3e6fb

tracer/src/Datadog.Trace/OpenTelemetry/Traces/OtlpTracesJsonSerializer.cs
zacharycmontoya: Added a static readonly field in b90174fa83cd8a682131fb0c3ed52c06dff3e6fb

tracer/src/Datadog.Trace/Agent/SpanBufferMessagePackSerializer.cs
zacharycmontoya: Fixed in b90174fa83cd8a682131fb0c3ed52c06dff3e6fb

tracer/src/Datadog.Trace/OpenTelemetry/Traces/OtlpTracesJsonSerializer.cs
zacharycmontoya: Fixed in b90174fa83cd8a682131fb0c3ed52c06dff3e6fb

tracer/src/Datadog.Trace/Agent/ApiOtlp.cs
zacharycmontoya: Fixed in b90174fa83cd8a682131fb0c3ed52c06dff3e6fb

tracer/src/Datadog.Trace/OpenTelemetry/Traces/OtlpTracesJsonSerializer.cs
zacharycmontoya: I'll add a new comment, but spanBufferOffset is only used to calculate whether spans have been added. All of the writing is done in a temporary buffer, whose position is indicated by the `offset` variable. I'll rename `offset` though so it's more descriptive

tracer/src/Datadog.Trace/Agent/ApiOtlp.cs
bouwkast: Thanks!

tracer/src/Datadog.Trace/Configuration/ExporterSettings.cs
bouwkast: LGTM thanks!

tracer/src/Datadog.Trace/OpenTelemetry/OtlpMapper.cs
zacharycmontoya: For now, I've removed the DD Sketch to OTLP metrics conversion because we don't need this functionality for our near-term goals. This functionality can easily be reintroduced later with better testing along with it.

Removed in f32eda0edbfacfb97b16bb5eeca0466cc2281fce

tracer/src/Datadog.Trace/OpenTelemetry/OtlpMapper.cs
zacharycmontoya: For now, I've removed the DD Sketch to OTLP metrics conversion because we don't need this functionality for our near-term goals. This functionality can easily be reintroduced later with better testing along with it.

Removed in f32eda0edbfacfb97b16bb5eeca0466cc2281fce

PR #8214, opened 2026-02-18
[CI] Upload system-tests results to Test Optimization

nccatoni:
## Summary of changes

Adds `datadog-ci junit upload` steps to the `system_tests` stage in the Azure pipeline to push JUnit XML test results to Datadog Test Optimization after each run.

## Reason for change

Enable tracking of system-tests results in Datadog Test Optimization for visibility into test health and regressions.

## Implementation details

After the tests run and logs are grouped, the pipeline:
1. Installs `@datadog/datadog-ci` via npm
2. Uploads `logs*/reportJunit.xml` to Test Optimization using the `DD_API_KEY_PROD` secret

Follows the same pattern used in [system-tests push_to_test_optim action](https://github.com/DataDog/system-tests/blob/main/.github/actions/push_to_test_optim/action.yml).

## Test coverage

- [ ] Verify the Azure pipeline runs successfully with the new upload steps

## Other details

This is a CI-only change. No code changes.


.azure-pipelines/ultimate-pipeline.yml
```diff
@@ -4965,6 +4965,24 @@ stages:
         condition: always()
         displayName: Group logs
 
+      - script: npm install -g @datadog/datadog-ci || (sleep 60 && npm install -g @datadog/datadog-ci)
```
andrewlock: `sleep 60`, really? 😅

.azure-pipelines/ultimate-pipeline.yml
```diff
@@ -4965,6 +4965,24 @@ stages:
...
+        condition: always()
+        displayName: Install datadog-ci
```
andrewlock: We should expect this to fail:
```suggestion
        displayName: Install datadog-ci
        retryCountOnTaskFailure: 3
```

.azure-pipelines/ultimate-pipeline.yml
```diff
@@ -4965,6 +4965,24 @@ stages:
...
+
+      - script: |
+          datadog-ci junit upload \
+            ./logs*/reportJunit.xml \
+            --service system-tests \
+            --env ci \
+            --verbose \
+            --xpath-tag "test.codeowners=/testcase/properties/property[@name='test.codeowners']"
+        workingDirectory: system-tests
+        condition: always()
```
andrewlock, line 4977: We should retry if it fails, but don't fail the whole job if it does
```suggestion
        condition: always()
        retryCountOnTaskFailure: 3
        continueOnError: true
```

.azure-pipelines/ultimate-pipeline.yml
```diff
@@ -4965,6 +4965,24 @@ stages:
         condition: always()
         displayName: Group logs
 
+      - script: npm install -g @datadog/datadog-ci || (sleep 60 && npm install -g @datadog/datadog-ci)
+        condition: always()
+        displayName: Install datadog-ci
+
+      - script: |
```
andrewlock, line 4968: Is there a reason to have these as separate steps? If we combined them, we could add `continueOnError: true` to the single job. Right now, an npm issue will cause the system-tests to all fail when it's non-critical 🤔

.azure-pipelines/ultimate-pipeline.yml
```diff
@@ -4965,6 +4965,24 @@ stages:
...
+            --service system-tests \
+            --env ci \
+            --verbose \
+            --xpath-tag "test.codeowners=/testcase/properties/property[@name='test.codeowners']"
+        workingDirectory: system-tests
+        condition: always()
+        displayName: Upload test results to Test Optimization
+        env:
+          DATADOG_SITE: datadoghq.com
+          DATADOG_API_KEY: $(DD_API_KEY_PROD)
```
andrewlock: nit (it evaluates the same, just doesn't tie us to the specific secret name)
```suggestion
          DATADOG_API_KEY: $(ddApiKey)
```

andrewlock (approved):
Thanks!

PR #8216, opened 2026-02-18
NUnit: emit test.final_status on terminal executions

tonyredondo:
## Summary of changes
- Make NUnit emit `test.final_status` when that execution is the **final one** for the test.
- Explicitly set `test.final_status=skip` for pre-execution skips and disabled tests.
- Keep retry behavior consistent by setting `test.final_status` on terminal retry executions.
- Update CI assertion helpers to avoid local false failures caused by leftover optional metadata keys.

## Reason for change
People use `test.final_status` to answer simple questions like “which tests ultimately failed?”

NUnit still had a few paths where that tag could be missing or inconsistent (especially around skip/disabled and non-retry paths). This PR makes the behavior consistent with the intended rule: **set the tag when no more retries will run**.

## Implementation details
- `tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Testing/NUnit/NUnitIntegration.cs`
  - In pre-execution skip path, set `final_status=skip` before closing the test span.

- `tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Testing/NUnit/TestOptimizationTestCommand.cs`
  - Set `final_status=skip` for disabled tests.
  - On non-retry executions, check whether EFD/ATR/ATF will schedule retries.
  - If no retry will happen, set `final_status` from the current execution outcome.
  - Keep existing terminal-retry final_status behavior.

- `tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/CI/TestingFrameworkEvpTest.cs`
- `tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/CI/TestingFrameworkTest.cs`
  - Always remove checked metadata keys after assertions to prevent local `Assert.Single()` noise.

## Test coverage
- Targeted NUnit coverage for changed behavior was re-run:
  - `NUnitTests.SubmitTraces`
  - `NUnitEvpTests.SubmitTraces`
  - `NUnitEvpTests.EarlyFlakeDetection`
  - `NUnitEvpTests.QuarantinedTests`
  - `NUnitEvpTests.DisabledTests`
  - `NUnitEvpTests.AttemptToFixTests`
- Result: **13/13 passed**.

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Testing/NUnit/TestOptimizationTestCommand.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Testing/NUnit/TestOptimizationTestCommand.cs
@@ -221,9 +221,8 @@ internal sealed class TestOptimizationTestCommand
 
         if (test is not null && testTags is not null)
         {
-            // Emit final_status for any execution that won't be retried anymore:
-            // - tests with no retries
-            // - final retry execution
+            // Non-retry execution: emit final_status only when no retry will be scheduled.
+            // Final retry executions are handled in the isFinalExecution block below.
             if (!retryState.IsARetry && testTags.FinalStatus is null)
             {
                 var isSkippedOrInconclusive = resultStatus is TestStatus.Skipped or TestStatus.Inconclusive;


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Testing/NUnit/TestOptimizationTestCommand.cs
```diff
@@ -237,6 +221,28 @@ private ITestResult ExecuteTest(ITestExecutionContext context, int executionNumb
 
         if (test is not null && testTags is not null)
         {
+            // Emit final_status for any execution that won't be retried anymore:
+            // - tests with no retries
+            // - final retry execution
+            if (!retryState.IsARetry && testTags.FinalStatus is null)
```
andrewlock, line 226: Is this correct? If this was the final retry, won't `IsARetry` be true? 🤔

.gitignore
```diff
@@ -401,3 +401,6 @@ tools/dumps/
 !tools/dumps/README_DEBIAN_ARM64.md
 !tools/dumps/start_debian-arm64.cmd
 !tools/dumps/start_debian-arm64.sh
+
+# test optimization temp/run folder
+.dd/
```
andrewlock, line 406: ooooh, so _that's_ where that was coming from 😅 

I'm kinda surprised this isn't a problem for customers tbh 🤔

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Testing/NUnit/TestOptimizationTestCommand.cs
tonyredondo: I've updated the comment

PR #8217, opened 2026-02-18
Transient shutdown timeout

NachoEchevarria:
## Summary of changes

Add Error running shutdown hooks System.TimeoutException to the CheckBuildLogsForErrors known patterns allowlist to prevent flaky CI failures.

## Reason for change

 The CheckBuildLogsForErrors post-test step scans tracer log files for [Error] entries and fails the build if it finds unrecognized ones. Under CI resource pressure, LifetimeManager.RunShutdownTasks can exceed its 30s timeout during
  process shutdown, logging a TimeoutException. This causes builds to fail even though all tests passed (e.g., Test debian_net8.0 in https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=196033).
  

>   10:42:51 [ERR] 02/18/2026 10:12:46 +00:00 [Error] Error running shutdown hooks System.TimeoutException: The operation has timed out.

## Implementation details

 Added one regex to the knownPatterns list in Build.Steps.cs, next to the existing ThreadAbortException pattern since both are transient shutdown errors.

### DIFF
--- a/tracer/build/_build/Build.Steps.cs
+++ b/tracer/build/_build/Build.Steps.cs
@@ -2506,6 +2506,8 @@ partial class Build
                new(@".*Noop\dArgumentsVoidIntegration\.OnMethodBegin.*CallTargetNativeTest.*", RegexOptions.Compiled | RegexOptions.Singleline),
                new(@".*Noop\dArgumentsVoidIntegration\.OnMethodEnd.*CallTargetNativeTest.*", RegexOptions.Compiled | RegexOptions.Singleline),
                new(@".*System.Threading.ThreadAbortException: Thread was being aborted\.", RegexOptions.Compiled),
+               // Transient shutdown timeout - LifetimeManager hooks may not complete within 30s under CI resource pressure
+               new(@".*Error running shutdown hooks System\.TimeoutException.*", RegexOptions.Compiled),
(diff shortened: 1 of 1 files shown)


tracer/build/_build/Build.Steps.cs
```diff
@@ -2506,6 +2506,8 @@ string NormalizedPath(AbsolutePath ap)
                new(@".*Noop\dArgumentsVoidIntegration\.OnMethodBegin.*CallTargetNativeTest.*", RegexOptions.Compiled | RegexOptions.Singleline),
                new(@".*Noop\dArgumentsVoidIntegration\.OnMethodEnd.*CallTargetNativeTest.*", RegexOptions.Compiled | RegexOptions.Singleline),
                new(@".*System.Threading.ThreadAbortException: Thread was being aborted\.", RegexOptions.Compiled),
+               // Transient shutdown timeout - LifetimeManager hooks may not complete within 30s under CI resource pressure
```
andrewlock, line 2509: I am a little hesitant about this one tbh... I understand why we might want to add this, but if this started happening all the time, we would _really_ want to know about it, so I'm not sure we _should_ skip it 🤔

NachoEchevarria: I see. Maybe increasing the timeout to 60 secs woud be a better option? This way, if its a matter of time, we could fix it, but if it's something else, it will keep failing. WDYT?

PR #8227, opened 2026-02-20
Implement `app-extended-heartbeat` telemetry event

andrewlock:
## Summary of changes

Implements the [telemetry v2 extended heartbeat](https://github.com/DataDog/instrumentation-telemetry-api-docs/blob/main/GeneratedDocumentation/ApiDocs/v2/SchemaDocumentation/Schemas/app_extended_heartbeat.md) spec

## Reason for change

We've been asked to implement the extended heartbeat as the backend needs it for data consistency reasons.

## Implementation details

Today, after reporting config keys to the telemetry backend we drop them. After this PR, we keep each poll's `List<T>` in a `List<List<T>>`, and use that to send all the data in the extended heartbeat. We already keep the product and dependency data by design, so it's only configs that needed to change.

In addition, having the configs available means that we can now also include this in the telemetry flare data, which should be very useful for support cases. 

_Unfortunately_, the fact that the tracer flare can happen concurrently with the extended heartbeat complicates things slightly, and means we have to add some synchronization around the `GetData()`/`GetFullData()` calls. I don't expect any contention there in practice though, so it shouldn't be a big issue.

## Test coverage

Existing tests pass, and added some additional unit tests for both the tracer flare and the extended heartbeat. System tests will also be added, which will use the configuration key `DD_TELEMETRY_EXTENDED_HEARTBEAT_INTERVAL` to control the interval.

## Other details

Initially had 🤖 look at it, but I didn't like what it did at all, so went rogue and hand coded instead (ok it wrote one of the tests)

### DIFF
--- a/tracer/test/Datadog.Trace.Tests/Telemetry/Transports/TelemetryTransportFactoryTests.cs
+++ b/tracer/test/Datadog.Trace.Tests/Telemetry/Transports/TelemetryTransportFactoryTests.cs
@@ -27,7 +27,7 @@ public class TelemetryTransportFactoryTests
             agentlessSettings: agentlessEnabled ? new TelemetrySettings.AgentlessSettings(new Uri("http://localhost"), "SOME_API_KEY", null) : null,
             agentProxyEnabled: agentProxyEnabled,
             heartbeatInterval: TimeSpan.FromSeconds(1),
-            extendedHeartbeatInterval: TimeSpan.FromHours(24),
+            extendedHeartbeatInterval: TimeSpan.FromDays(1),
             dependencyCollectionEnabled: true,
             metricsEnabled: true,
             debugEnabled: false,


lucaspimentel:
In the PR description:

> the tracer flare can happen concurrently with the tracer flare 

one of those was meant to be "heartbeat"?

tracer/test/Datadog.Trace.Tests/Telemetry/Transports/TelemetryTransportFactoryTests.cs
```diff
@@ -27,6 +27,7 @@ public void UsesCorrectTransports(bool agentProxyEnabled, bool agentlessEnabled)
             agentlessSettings: agentlessEnabled ? new TelemetrySettings.AgentlessSettings(new Uri("http://localhost"), "SOME_API_KEY", null) : null,
             agentProxyEnabled: agentProxyEnabled,
             heartbeatInterval: TimeSpan.FromSeconds(1),
+            extendedHeartbeatInterval: TimeSpan.FromHours(24),
```
lucaspimentel: but `TimeSpan.FromDays(1)` is so much better!

andrewlock: Is it the same thing though, what happens if the app starts on daylight savings 😉

andrewlock: [Turns out it's _Defined_ in terms of hours](https://github.com/dotnet/dotnet/blob/b0f34d51fccc69fd334253924abd8d6853fad7aa/src/runtime/src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs#L85), so fair enough 😉

tracer/src/Datadog.Trace/Telemetry/Collectors/ConfigurationTelemetry.Collector.cs
```diff
@@ -43,31 +47,48 @@ public void CopyTo(IConfigurationTelemetry destination)
...
-            {
-                return data;
+                // we have multiple collections to report
+                _reportedCount = _allData.Count;
+                return new ListOfListOfConfigurationKeyValue(_allData, _reportedCount);
             }
+        }
 
-            GetData(config, data);
+        public ICollection<ConfigurationKeyValue>? GetFullData()
```
anna-git, line 78: nit: seems like return type can never be null?

tracer/src/Datadog.Trace/Telemetry/Collectors/ConfigurationTelemetry.Collector.cs
```diff
@@ -111,6 +163,103 @@ public void Clear()
             while (config.TryDequeue(out _))
             {
             }
+
+            // clears any saved data
+            _allData.Clear();
```
anna-git: does it need to be locked?

tracer/src/Datadog.Trace/Telemetry/Collectors/ConfigurationTelemetry.Collector.cs
```diff
@@ -43,31 +47,48 @@ public void CopyTo(IConfigurationTelemetry destination)
         /// <returns>Null if there are no changes, or the collector is not yet initialized</returns>
         public ICollection<ConfigurationKeyValue>? GetData()
```
anna-git, line 48: should this be renamed, as it took me🐢  some time to understand one is the diff one is a full snapshot? maybe `GetDataDiff` or `GetIncrementalData`?

tracer/src/Datadog.Trace/Telemetry/Collectors/ConfigurationTelemetry.Collector.cs
```diff
@@ -43,31 +47,48 @@ public void CopyTo(IConfigurationTelemetry destination)
...
+                    return null;
+                }
 
-            var config = Interlocked.Exchange(ref _entries, _backBuffer);
-            _backBuffer = config;
+                if (_reportedCount == _allData.Count - 1)
+                {
+                    // we just need to report the latest results
+                    _reportedCount++;
+                    return latestData;
```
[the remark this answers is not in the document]
anna-git: isnt it expected behavior? if extended heartbeat is true, we report everything and the next normal GetData call should return null?

tracer/src/Datadog.Trace/Telemetry/Collectors/ConfigurationTelemetry.Collector.cs
```diff
@@ -43,31 +47,48 @@ public void CopyTo(IConfigurationTelemetry destination)
...
+                }
 
-            if (config.IsEmpty)
-            {
-                return data;
+                // we have multiple collections to report
+                _reportedCount = _allData.Count;
+                return new ListOfListOfConfigurationKeyValue(_allData, _reportedCount);
```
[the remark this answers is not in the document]
anna-git: should `_reportedCount = _allData.Count;` be set after returning `ListOfListOfConfigurationKeyValue` ? as otherwise everything gets skipped?

tracer/src/Datadog.Trace/Telemetry/Collectors/ConfigurationTelemetry.Collector.cs
andrewlock: Correct, but in the interface it implements it can be currently

tracer/src/Datadog.Trace/Telemetry/Collectors/ConfigurationTelemetry.Collector.cs
andrewlock: `GetIncrementalData` makes sense 👍 We use this same naming across _all_ the collectors though, so I'll do a follow up PR to rename all those if that's ok?

anna-git: oh didn't realize that, sounds good for a follow up PR!

tracer/src/Datadog.Trace/Telemetry/Collectors/ConfigurationTelemetry.Collector.cs
andrewlock: Yeah, theoretically, it's only used in tests, so not a big deal, but makes sense 👍

tracer/src/Datadog.Trace/Telemetry/Collectors/ConfigurationTelemetry.Collector.cs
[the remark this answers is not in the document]
andrewlock: Yep, 100%. Makes me wonder why tests didn't catch this though, so will add one to make sure it does! 😄

tracer/src/Datadog.Trace/Telemetry/Collectors/ConfigurationTelemetry.Collector.cs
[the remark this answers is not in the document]
andrewlock: Yeah, as far as I can tell the AI is just wrong here, and it should get reported next itteration, because we only modify _reportedCount and _allData under the lock 🤔

tracer/src/Datadog.Trace/Telemetry/Collectors/ConfigurationTelemetry.Collector.cs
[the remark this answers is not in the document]
andrewlock: Added a test to confirm it's broken, and fixed!

tracer/src/Datadog.Trace/Telemetry/Collectors/ConfigurationTelemetry.Collector.cs
andrewlock: Fixed

tracer/src/Datadog.Trace/Telemetry/Collectors/ConfigurationTelemetry.Collector.cs
andrewlock: Done here: https://github.com/DataDog/dd-trace-dotnet/pull/8269

PR #8228, opened 2026-02-20
Add an `IArrayPool<char>` implementation for vendored Newtonsoft.JSON

andrewlock:
## Summary of changes

Adds a simple `IArrayPool<char>` for use by Newtonsoft.JSON, and uses it everywhere we can

## Reason for change

Newtonsoft.JSON fundamentally works with .NET's `char` type (UTF-16), (as opposed to System.Text.Json which works with UTF-8 where it can). To do so, it needs to create a bunch of `char[]` instances to use as buffers.

The `JsonTextReader` and `JsonTextWriter` abstractions allow plugging in an `IArrayPool<char>` implementation, and luckily this matches (pretty much exactly) the API exposed by the `ArrayPool` in corelib (+ vendored), so it's easy to implement.

This should help alleviate some GC pressure, as we currently do a fair amount of serializing and deserializing.

## Implementation details

Pretty simple:
- Implement `IArrayPool<char>` using `ArrayPool<char>.Shared`
- Get 🤖 to find everywhere that we could use it (`JsonTextReader` and `JsonTextWriter`) and initialize
- Fix some cases where these weren't being disposed.

> [!WARNING]
> It's important that we _do_ dispose these, so that the arrays are correctly returned to the pool, so that we don't leak memory

There are actually other places we can update too, as this PR doesn't cover the common `JsonConvert.Serialize()` etc, but I'll follow up with those in a separate PR.

## Test coverage

All the existing tests should pass. I worked on this as part of general perf work on remote config, so the results are a bit fuzzy (as I can't remember if it includes the savings from #8226 as well), but the results are pretty conclusive, especially for big payloads 😅 

| Method                       | size  |        Mean |         Error |  Allocated |
| ---------------------------- | ----- | ----------: | ------------: | ---------: |
| DeserializeResponse_Original | Small |    22.71 ns |  2,076.771 ns |   23.83 KB |
| DeserializeResponse_Updated  | Small |    13.70 us |      0.186 us |   17.23 KB |
|                              |       |             |               |            |
| DeserializeResponse_Original | Big   | 1,953.04 us | 58,665.219 ns |2,343.26 KB |
| DeserializeResponse_Updated  | Big   |   614.46 us |     11.988 us |  252.37 KB |

## Other details

Discovered this while exploring remote config optimizations

### DIFF
--- a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Kinesis/ContextPropagation.cs
+++ b/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Kinesis/ContextPropagation.cs
@@ -8,11 +8,15 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Text;
 using Datadog.Trace.DataStreamsMonitoring;
 using Datadog.Trace.DuckTyping;
 using Datadog.Trace.Headers;
 using Datadog.Trace.Logging;
 using Datadog.Trace.Propagators;
+using Datadog.Trace.SourceGenerators;
+using Datadog.Trace.Util;
+using Datadog.Trace.Util.Json;
 using Datadog.Trace.Vendors.Newtonsoft.Json;
 
 namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.AWS.Kinesis
@@ -109,6 +113,7 @@ namespace Datadog.Trace.ClrProfiler.AutoInstrumentation.AWS.Kinesis
             }
         }
 
+        [TestingAndPrivateOnly]
         internal static Dictionary<string, object>? ParseDataObject(MemoryStream dataStream)
         {
             try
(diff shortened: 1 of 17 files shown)


tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/Kinesis/ContextPropagation.cs
```diff
@@ -123,11 +128,13 @@ public static void InjectTraceIntoData<TRecordRequest>(Tracer tracer, TRecordReq
...
         }
 
+        [TestingAndPrivateOnly]
         public static Dictionary<string, object>? MemoryStreamToDictionary(MemoryStream stream)
         {
             // Convert the MemoryStream to a string
-            var streamReader = new StreamReader(stream);
-            var reader = new JsonTextReader(streamReader);
+            // Default values for StreamReader, but with leaveOpen:true
+            using var streamReader = new StreamReader(stream, Encoding.UTF8, detectEncodingFromByteOrderMarks: true,  bufferSize: 1024, leaveOpen: true);
```
andrewlock, line 136: Previously this wasn't disposing the `StreamReader`, _presumably_ because that closes the `MemoryStream` by default. I think _technically_ that was ok, but seemed better to be explicit here, especially as we now _have_ to dispose the `JsonTextReader`, and don't want there to be any ambiguity around it

tracer/src/Datadog.Trace/Debugger/Expressions/ProbeExpressionParser.cs
```diff
@@ -558,7 +560,7 @@ private ExpressionBodyAndParameters ParseProbeExpression(string expressionJson,
         var result = Expression.Variable(typeof(T), "$dd_el_result");
         scopeMembers.Add(result);
 
-        var reader = new JsonTextReader(new StringReader(expressionJson));
+        using var reader = new JsonTextReader(new StringReader(expressionJson)) { ArrayPool = JsonArrayPool.Shared };
```
andrewlock, line 563: I'm pretty sure this was just an accidental miss of the `using`

tracer/src/Datadog.Trace/Logging/DirectSubmission/Formatting/LogFormatter.cs
```diff
@@ -276,11 +277,11 @@ internal static void WriteValue(JsonWriter writer, object? value)
         }
 
         internal static JsonTextWriter GetJsonWriter(StringBuilder builder)
```
andrewlock, line 279: The callers "own" and handle disposing the returned object

dudikeleti (approved):

PR #8231, opened 2026-02-20
[Serverless] Cache serverless platform detection results

lucaspimentel:
## Summary of changes

Extract serverless platform detection logic from `EnvironmentHelpers` into dedicated cached detection classes under `Datadog.Trace.Serverless`:
- `AzureInfo` — Azure App Services, Azure Functions (including isolated host/worker detection)
- `AwsInfo` — AWS Lambda
- `GcpInfo` — Google Cloud Functions

Each class caches environment variable lookups in instance fields to avoid repeated reads (which incur heap allocations) on hot paths. An `Instance` singleton is used in production; tests create fresh instances for isolation.

## Reason for change

Platform detection methods (e.g., `IsAzureAppService()`, `IsAwsLambda()`) were called repeatedly from hot paths like span creation and peer service derivation. Each call re-read environment variables, adding unnecessary overhead. Caching the results eliminates redundant lookups.

This change also moves serverless-specific code out of `EnvironmentHelpers` and into the `Datadog.Trace.Serverless` namespace.

## Implementation details

- Add `AzureInfo`, `AwsInfo`, and `GcpInfo` classes under `Datadog.Trace.Serverless` namespace
- Each class exposes cached `bool` properties (e.g., `AzureInfo.Instance.IsAzureAppService`, `AwsInfo.Instance.IsAwsLambda`, `GcpInfo.Instance.IsCloudFunction`)
- `AzureInfo` also caches `AzureFunctionsWorkerRuntime`, `AzureFunctionsExtensionVersion`, and derived properties like `IsIsolatedFunction`, `IsIsolatedFunctionHostProcess`, and `IsIsolatedFunctionWorkerProcess`
- Non-static design with `Instance` singleton enables test isolation without `Reset()` methods
- Where possible, callers pass `bool` values from the detection classes rather than taking a dependency on the platform objects themselves (e.g., `ProfilerAvailabilityHelper`)
- Promote `EnvironmentHelpers.EnvironmentVariableExists` to `internal` visibility for use by the new classes
- Update all callers across the tracer to use the new types instead of `EnvironmentHelpers.IsXxx()` methods
- Remove the now-unused serverless detection methods from `EnvironmentHelpers`

## Test coverage

- Add `AwsInfoTests`, `AzureInfoTests`, and `GcpInfoTests` under `Serverless/`
- Update `ProfilerAvailabilityHelperTests` to pass platform detection bools explicitly
- Remove old `EnvironmentHelpersTests` that tested the now-removed methods

## Other details

n/a

> *"Why did the environment variable go to therapy? It was tired of being read over and over without anyone remembering what it said."* — Claude 🤖

### DIFF
--- a/tracer/src/Datadog.Trace/Serverless/AzureInfo.cs
+++ b/tracer/src/Datadog.Trace/Serverless/AzureInfo.cs
@@ -24,8 +24,6 @@ internal sealed class AzureInfo
     private bool? _isIsolatedFunction;
     private bool? _isIsolatedFunctionHostProcess;
     private bool? _isIsolatedFunctionWorkerProcess;
-    private bool _functionWorkerRuntimeCached;
-    private bool _functionExtensionVersionCached;
 
     /// <summary>
     /// Gets the shared singleton instance. Cached after first evaluation.
@@ -82,35 +80,13 @@ internal sealed class AzureInfo
     /// Gets the cached value of the "FUNCTIONS_WORKER_RUNTIME" environment variable.
     /// </summary>
     internal string? AzureFunctionsWorkerRuntime
-    {
-        get
-        {
-            if (!_functionWorkerRuntimeCached)
-            {
-                field = EnvironmentHelpers.GetEnvironmentVariable(PlatformKeys.AzureFunctions.FunctionsWorkerRuntime, defaultValue: string.Empty);
-                _functionWorkerRuntimeCached = true;
-            }
-
-            return field;
-        }
-    }
+        => field ??= EnvironmentHelpers.GetEnvironmentVariable(PlatformKeys.AzureFunctions.FunctionsWorkerRuntime, defaultValue: string.Empty);
 
     /// <summary>
     /// Gets the cached value of the "FUNCTIONS_EXTENSION_VERSION" environment variable.
     /// </summary>
     internal string? AzureFunctionsExtensionVersion
-    {
-        get
-        {
-            if (!_functionExtensionVersionCached)
-            {
-                field = EnvironmentHelpers.GetEnvironmentVariable(PlatformKeys.AzureFunctions.FunctionsExtensionVersion, defaultValue: string.Empty);
-                _functionExtensionVersionCached = true;
-            }
-
-            return field;
-        }
-    }
+        => field ??= EnvironmentHelpers.GetEnvironmentVariable(PlatformKeys.AzureFunctions.FunctionsExtensionVersion, defaultValue: string.Empty);
 
     /// <summary>
     /// Gets a value indicating whether the current environment is the Azure Functions host process


lucaspimentel:
@codex review

tracer/test/Datadog.Trace.Tests/ContinuousProfiler/ProfilerAvailabilityHelperTests.cs
```diff
@@ -34,46 +34,43 @@ public void IsContinuousProfilerAvailable_OnUnsupportedPlatforms_ReturnsFalse()
         SkipOn.PlatformAndArchitecture(SkipOn.PlatformValue.Windows, SkipOn.ArchitectureValue.X86);
         SkipOn.PlatformAndArchitecture(SkipOn.PlatformValue.Linux, SkipOn.ArchitectureValue.X64);
 
-        ProfilerAvailabilityHelper.IsContinuousProfilerAvailable_TestingOnly(ClrProfilerIsAttached).Should().BeFalse();
+        ProfilerAvailabilityHelper.IsContinuousProfilerAvailable_TestingOnly(ClrProfilerIsAttached, new Aws(), new Azure()).Should().BeFalse();
```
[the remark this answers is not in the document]
lucaspimentel: obvious compiler error. fixed!

lucaspimentel:
@codex review

tracer/src/Datadog.Trace/Serverless/Aws.cs
```diff
@@ -0,0 +1,34 @@
...
+using Datadog.Trace.Util;
+
+namespace Datadog.Trace.Serverless;
+
+/// <summary>
+/// Cached helpers for detecting AWS serverless platforms.
+/// Delegates to <see cref="EnvironmentHelpers"/> for env-var reads.
+/// Create a new instance for test isolation; use <see cref="Default"/> for production code.
+/// </summary>
+internal sealed class Aws
```
andrewlock: Not a big fan of calling these simply Aws/Azure/Gcp, partly due to collisions with those as namespaces, How do you feel like `AwsInfo` etc or something instead 🤔

tracer/src/Datadog.Trace/Serverless/Aws.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    private bool? _isLambda;
+
+    /// <summary>
+    /// Gets the shared singleton instance. Cached after first evaluation.
+    /// </summary>
+    internal static Aws Default { get; } = new();
```
andrewlock: non-blocking comment: we generally use `Instance` elsewhere, should we use that instead? Or even `Current`?

tracer/src/Datadog.Trace/Serverless/AzureInfo.cs
```diff
@@ -0,0 +1,171 @@
...
+        {
+            if (!_functionWorkerRuntimeCached)
+            {
+                field = EnvironmentHelpers.GetEnvironmentVariable(PlatformKeys.AzureFunctions.FunctionsWorkerRuntime, defaultValue: string.Empty);
+                _functionWorkerRuntimeCached = true;
+            }
+
+            return field;
+        }
+    }
```
andrewlock: Why the alternative format for these? 🤔 AFAICT, this is the same, as it will always return a non-null `string`?

```suggestion
    internal string FunctionWorkerRuntime
        => field ??= EnvironmentHelpers.GetEnvironmentVariable(PlatformKeys.AzureFunctions.FunctionsWorkerRuntime, defaultValue: string.Empty);
```

Also, I think you can replace all the dedicated fields in these types with `field` 😄

tracer/src/Datadog.Trace/Serverless/Aws.cs
lucaspimentel: Yeah, I went back and forth a few times. Originally it was something like `AwsPlatformDetection.IsAwsLambda` but it looked too long to me. Something short `Aws.IsLambda` looked nicer in comparison.

But you're right about naming conflicts later on, thanks.

tracer/src/Datadog.Trace/Serverless/AzureInfo.cs
lucaspimentel: > I think you can replace all the dedicated fields in these types with `field`

The other ones are `bool` properties with `bool?` backing fields, so they can't use `field` (type mismatch).

The types match in this one (`string?`). The logic is different because unlike with `bool?`, a `null` string is not enough to tell if we have a cached value or not. We can cache a `null` string if the env var is not set.

tracer/src/Datadog.Trace/Serverless/Aws.cs
lucaspimentel: renamed to `Instance`, thanks

tracer/src/Datadog.Trace/Serverless/Aws.cs
lucaspimentel: renamed all 3 to `<cloud>Info`

tracer/src/Datadog.Trace/Serverless/AzureInfo.cs
andrewlock: > properties with bool? backing fields, so they can't use field (type mismatch).

Good point 👍 

> We can cache a null string if the env var is not set.

But you specifically set the `defaultValue` to `string.Empty` here, so I don't think you ever _will_ cache a `null` value? 😄

lucaspimentel: That's right, thanks. Simplified.

PR #8232, opened 2026-02-21
[AAP] API10 downstream request analysis

daniel-romano-DD:
## Summary of changes
Implement needed changes to handle API10 (downstream request) vulnerability in RASP

### DIFF
--- a/tracer/test/Datadog.Trace.Security.IntegrationTests/AspNetBase.cs
+++ b/tracer/test/Datadog.Trace.Security.IntegrationTests/AspNetBase.cs
@@ -144,7 +144,7 @@ namespace Datadog.Trace.Security.IntegrationTests
             settings.AddRegexScrubber(AppSecFingerPrintNetwork, "_dd.appsec.fp.http.network: <NetworkPrint>");
         }
 
-        public async Task VerifySpans(IImmutableList<MockSpan> spans, VerifySettings settings, bool testInit = false, string methodNameOverride = null, string testName = null, string fileNameOverride = null, bool showRulesVersion = false, bool scrubCookiesFingerprint = false)
+        public async Task VerifySpans(IImmutableList<MockSpan> spans, VerifySettings settings, bool testInit = false, string methodNameOverride = null, string testName = null, string fileNameOverride = null, bool showRulesVersion = false, bool scrubCookiesFingerprint = false, Func<IReadOnlyCollection<MockSpan>, IOrderedEnumerable<MockSpan>> orderSpans = null)
         {
             settings.ModifySerialization(
                 serializationSettings =>
@@ -231,13 +231,17 @@ namespace Datadog.Trace.Security.IntegrationTests
             {
                 // Overriding the type name here as we have multiple test classes in the file
                 // Ensures that we get nice file nesting in Solution Explorer
-                await Verifier.Verify(spans, settings)
-                              .UseMethodName(methodNameOverride ?? "_")
-                              .UseTypeName(testName ?? GetTestName());
+                // The span ordering is kept identical to the passed in order,
+                // but that means that it must be deterministic. We should consider using the default
+                // ordering instead
+                orderSpans ??= x => x.OrderBy(_ => 0);
+                await VerifyHelper.VerifySpans(spans, settings, orderSpans)
+                                      .UseMethodName(methodNameOverride ?? "_")
+                                      .UseTypeName(testName ?? GetTestName());
             }
             else
             {
-                await VerifyHelper.VerifySpans(spans, settings)
+                await VerifyHelper.VerifySpans(spans, settings, orderSpans)
                                   .UseFileName(fileNameOverride)
                                   .DisableRequireUniquePrefix();
             }
--- a/tracer/test/Datadog.Trace.Security.IntegrationTests/RASP/AspNetCore5RaspDownstream.cs
+++ b/tracer/test/Datadog.Trace.Security.IntegrationTests/RASP/AspNetCore5RaspDownstream.cs
@@ -5,6 +5,7 @@
 
 #if NET5_0_OR_GREATER
 
+using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
 using System.Threading.Tasks;
@@ -61,12 +62,20 @@ public class AspNetCore5RaspDownstream : AspNetBase, IClassFixture<AspNetCoreTes
         var agent = Fixture.Agent;
 
         var spans = await SendRequestsAsync(agent, url, body, 1, 5, string.Empty, "application/json", null);
-        var spansFiltered = spans.Where(x => x.Type == SpanTypes.Web).ToList();
+        var spansFiltered = spans.Where(x => x.Type == SpanTypes.Web).ToImmutableList();
 
         var settings = VerifyHelper.GetSpanVerifierSettings();
         settings.UseParameters(testName, string.Empty, string.Empty);
-        await VerifySpans(spansFiltered.ToImmutableList(), settings);
+        await VerifySpans(spansFiltered, settings, orderSpans: OrderSpans);
     }
+
+    private static IOrderedEnumerable<MockSpan> OrderSpans(IReadOnlyCollection<MockSpan> spans)
+        => spans
+              .OrderBy(x => VerifyHelper.GetRootSpanResourceName(x, spans))
+              .ThenBy(x => VerifyHelper.GetSpanDepth(x, spans))
+              .ThenBy(x => x.Resource)
+              .ThenBy(x => x.Start)
+              .ThenBy(x => x.Duration);
 }
 
 #endif
--- a/tracer/test/snapshots/Security.AspNetCore5.RaspDownstream.__testName=WithBodyQueryParam_CapturesBody_url=_body=.verified.txt
+++ b/tracer/test/snapshots/Security.AspNetCore5.RaspDownstream.__testName=WithBodyQueryParam_CapturesBody_url=_body=.verified.txt
@@ -1,76 +1,7 @@
-﻿[
+[
   {
     TraceId: Id_1,
     SpanId: Id_2,
-    Name: aspnet_core_mvc.request,
-    Resource: POST /rasp/echo,
-    Service: Samples.Security.AspNetCore5,
-    Type: web,
-    ParentId: Id_3,
-    Tags: {
-      aspnet_core.action: echopost,
-      aspnet_core.controller: rasp,
-      aspnet_core.route: rasp/echo,
-      component: aspnet_core,
-      env: integration_tests,
-      language: dotnet,
-      span.kind: server
-    }
-  },
-  {
-    TraceId: Id_1,
-    SpanId: Id_3,
-    Name: aspnet_core.request,
-    Resource: POST /rasp/echo,
-    Service: Samples.Security.AspNetCore5,
-    Type: web,
-    ParentId: Id_4,
-    Tags: {
-      aspnet_core.endpoint: Samples.Security.AspNetCore5.Controllers.RaspController.EchoPost (Samples.Security.AspNetCore5),
-      aspnet_core.route: rasp/echo,
-      component: aspnet_core,
-      env: integration_tests,
-      http.client_ip: 127.0.0.1,
-      http.method: POST,
-      http.request.headers.host: localhost:00000,
-      http.route: rasp/echo,
-      http.status_code: 200,
-      http.url: http://localhost:00000/Rasp/Echo,
-      language: dotnet,
-      network.client.ip: 127.0.0.1,
-      runtime-id: Guid_1,
-      span.kind: server,
-      _dd.runtime_family: dotnet
-    },
-    Metrics: {
-      process_id: 0,
-      _dd.appsec.enabled: 1.0,
-      _dd.top_level: 1.0,
-      _dd.tracer_kr: 1.0,
-      _sampling_priority_v1: 1.0
-    }
-  },
-  {
-    TraceId: Id_1,
-    SpanId: Id_5,
-    Name: aspnet_core_mvc.request,
-    Resource: GET /rasp/downstreamtoself,
-    Service: Samples.Security.AspNetCore5,
-    Type: web,
-    ParentId: Id_6,
-    Tags: {
-      aspnet_core.action: downstreamtoself,
-      aspnet_core.controller: rasp,
-      aspnet_core.route: rasp/downstreamtoself,
-      component: aspnet_core,
-      env: integration_tests,
-      language: dotnet,
-      span.kind: server
-    }
-  },
-  {
-    TraceId: Id_1,
-    SpanId: Id_6,
     Name: aspnet_core.request,
     Resource: GET /rasp/downstreamtoself,
     Service: Samples.Security.AspNetCore5,
@@ -102,6 +33,75 @@
       _dd.top_level: 1.0,
       _dd.tracer_kr: 1.0,
       _sampling_priority_v1: 1.0
+    }
+  },
+  {
+    TraceId: Id_1,
+    SpanId: Id_3,
+    Name: aspnet_core_mvc.request,
+    Resource: GET /rasp/downstreamtoself,
+    Service: Samples.Security.AspNetCore5,
+    Type: web,
+    ParentId: Id_2,
+    Tags: {
+      aspnet_core.action: downstreamtoself,
+      aspnet_core.controller: rasp,
+      aspnet_core.route: rasp/downstreamtoself,
+      component: aspnet_core,
+      env: integration_tests,
+      language: dotnet,
+      span.kind: server
+    }
+  },
+  {
+    TraceId: Id_1,
+    SpanId: Id_4,
+    Name: aspnet_core.request,
+    Resource: POST /rasp/echo,
+    Service: Samples.Security.AspNetCore5,
+    Type: web,
+    ParentId: Id_5,
+    Tags: {
+      aspnet_core.endpoint: Samples.Security.AspNetCore5.Controllers.RaspController.EchoPost (Samples.Security.AspNetCore5),
+      aspnet_core.route: rasp/echo,
+      component: aspnet_core,
+      env: integration_tests,
+      http.client_ip: 127.0.0.1,
+      http.method: POST,
+      http.request.headers.host: localhost:00000,
+      http.route: rasp/echo,
+      http.status_code: 200,
+      http.url: http://localhost:00000/Rasp/Echo,
+      language: dotnet,
+      network.client.ip: 127.0.0.1,
+      runtime-id: Guid_1,
+      span.kind: server,
+      _dd.runtime_family: dotnet
+    },
+    Metrics: {
+      process_id: 0,
+      _dd.appsec.enabled: 1.0,
+      _dd.top_level: 1.0,
+      _dd.tracer_kr: 1.0,
+      _sampling_priority_v1: 1.0
+    }
+  },
+  {
+    TraceId: Id_1,
+    SpanId: Id_6,
+    Name: aspnet_core_mvc.request,
+    Resource: POST /rasp/echo,
+    Service: Samples.Security.AspNetCore5,
+    Type: web,
+    ParentId: Id_4,
+    Tags: {
+      aspnet_core.action: echopost,
+      aspnet_core.controller: rasp,
+      aspnet_core.route: rasp/echo,
+      component: aspnet_core,
+      env: integration_tests,
+      language: dotnet,
+      span.kind: server
     }
   }
 ]
--- a/tracer/test/snapshots/Security.AspNetCore5.RaspDownstream.__testName=WithBody_ParsesAndCapturesBody_url=_body=.verified.txt
+++ b/tracer/test/snapshots/Security.AspNetCore5.RaspDownstream.__testName=WithBody_ParsesAndCapturesBody_url=_body=.verified.txt
@@ -1,76 +1,7 @@
-﻿[
+[
   {
     TraceId: Id_1,
     SpanId: Id_2,
-    Name: aspnet_core_mvc.request,
-    Resource: POST /rasp/echo,
-    Service: Samples.Security.AspNetCore5,
-    Type: web,
-    ParentId: Id_3,
-    Tags: {
-      aspnet_core.action: echopost,
-      aspnet_core.controller: rasp,
-      aspnet_core.route: rasp/echo,
-      component: aspnet_core,
-      env: integration_tests,
-      language: dotnet,
-      span.kind: server
-    }
-  },
-  {
-    TraceId: Id_1,
-    SpanId: Id_3,
-    Name: aspnet_core.request,
-    Resource: POST /rasp/echo,
-    Service: Samples.Security.AspNetCore5,
-    Type: web,
-    ParentId: Id_4,
-    Tags: {
-      aspnet_core.endpoint: Samples.Security.AspNetCore5.Controllers.RaspController.EchoPost (Samples.Security.AspNetCore5),
-      aspnet_core.route: rasp/echo,
-      component: aspnet_core,
-      env: integration_tests,
-      http.client_ip: 127.0.0.1,
-      http.method: POST,
-      http.request.headers.host: localhost:00000,
-      http.route: rasp/echo,
-      http.status_code: 200,
-      http.url: http://localhost:00000/Rasp/Echo,
-      language: dotnet,
-      network.client.ip: 127.0.0.1,
-      runtime-id: Guid_1,
-      span.kind: server,
-      _dd.runtime_family: dotnet
-    },
-    Metrics: {
-      process_id: 0,
-      _dd.appsec.enabled: 1.0,
-      _dd.top_level: 1.0,
-      _dd.tracer_kr: 1.0,
-      _sampling_priority_v1: 1.0
-    }
-  },
-  {
-    TraceId: Id_1,
-    SpanId: Id_5,
-    Name: aspnet_core_mvc.request,
-    Resource: POST /rasp/downstreamtoself,
-    Service: Samples.Security.AspNetCore5,
-    Type: web,
-    ParentId: Id_6,
-    Tags: {
-      aspnet_core.action: downstreamtoselfpost,
-      aspnet_core.controller: rasp,
-      aspnet_core.route: rasp/downstreamtoself,
-      component: aspnet_core,
-      env: integration_tests,
-      language: dotnet,
-      span.kind: server
-    }
-  },
-  {
-    TraceId: Id_1,
-    SpanId: Id_6,
     Name: aspnet_core.request,
     Resource: POST /rasp/downstreamtoself,
     Service: Samples.Security.AspNetCore5,
@@ -102,6 +33,75 @@
       _dd.top_level: 1.0,
       _dd.tracer_kr: 1.0,
       _sampling_priority_v1: 1.0
+    }
+  },
+  {
+    TraceId: Id_1,
+    SpanId: Id_3,
+    Name: aspnet_core_mvc.request,
+    Resource: POST /rasp/downstreamtoself,
+    Service: Samples.Security.AspNetCore5,
+    Type: web,
+    ParentId: Id_2,
+    Tags: {
+      aspnet_core.action: downstreamtoselfpost,
+      aspnet_core.controller: rasp,
+      aspnet_core.route: rasp/downstreamtoself,
+      component: aspnet_core,
+      env: integration_tests,
+      language: dotnet,
+      span.kind: server
+    }
+  },
+  {
+    TraceId: Id_1,
+    SpanId: Id_4,
+    Name: aspnet_core.request,
+    Resource: POST /rasp/echo,
+    Service: Samples.Security.AspNetCore5,
+    Type: web,
+    ParentId: Id_5,
+    Tags: {
+      aspnet_core.endpoint: Samples.Security.AspNetCore5.Controllers.RaspController.EchoPost (Samples.Security.AspNetCore5),
+      aspnet_core.route: rasp/echo,
+      component: aspnet_core,
+      env: integration_tests,
+      http.client_ip: 127.0.0.1,
+      http.method: POST,
+      http.request.headers.host: localhost:00000,
+      http.route: rasp/echo,
+      http.status_code: 200,
+      http.url: http://localhost:00000/Rasp/Echo,
+      language: dotnet,
+      network.client.ip: 127.0.0.1,
+      runtime-id: Guid_1,
+      span.kind: server,
+      _dd.runtime_family: dotnet
+    },
+    Metrics: {
+      process_id: 0,
+      _dd.appsec.enabled: 1.0,
+      _dd.top_level: 1.0,
+      _dd.tracer_kr: 1.0,
+      _sampling_priority_v1: 1.0
+    }
+  },
+  {
+    TraceId: Id_1,
+    SpanId: Id_6,
+    Name: aspnet_core_mvc.request,
+    Resource: POST /rasp/echo,
+    Service: Samples.Security.AspNetCore5,
+    Type: web,
+    ParentId: Id_4,
+    Tags: {
+      aspnet_core.action: echopost,
+      aspnet_core.controller: rasp,
+      aspnet_core.route: rasp/echo,
+      component: aspnet_core,
+      env: integration_tests,
+      language: dotnet,
+      span.kind: server
     }
   }
 ]
--- a/tracer/test/snapshots/Security.AspNetCore5.RaspDownstream.__testName=WithoutBody_CapturesRequestData_url=_body=.verified.txt
+++ b/tracer/test/snapshots/Security.AspNetCore5.RaspDownstream.__testName=WithoutBody_CapturesRequestData_url=_body=.verified.txt
@@ -1,76 +1,7 @@
-﻿[
+[
   {
     TraceId: Id_1,
     SpanId: Id_2,
-    Name: aspnet_core_mvc.request,
-    Resource: GET /rasp/echo,
-    Service: Samples.Security.AspNetCore5,
-    Type: web,
-    ParentId: Id_3,
-    Tags: {
-      aspnet_core.action: echoget,
-      aspnet_core.controller: rasp,
-      aspnet_core.route: rasp/echo,
-      component: aspnet_core,
-      env: integration_tests,
-      language: dotnet,
-      span.kind: server
-    }
-  },
-  {
-    TraceId: Id_1,
-    SpanId: Id_3,
-    Name: aspnet_core.request,
-    Resource: GET /rasp/echo,
-    Service: Samples.Security.AspNetCore5,
-    Type: web,
-    ParentId: Id_4,
-    Tags: {
-      aspnet_core.endpoint: Samples.Security.AspNetCore5.Controllers.RaspController.EchoGet (Samples.Security.AspNetCore5),
-      aspnet_core.route: rasp/echo,
-      component: aspnet_core,
-      env: integration_tests,
-      http.client_ip: 127.0.0.1,
-      http.method: GET,
-      http.request.headers.host: localhost:00000,
-      http.route: rasp/echo,
-      http.status_code: 200,
-      http.url: http://localhost:00000/Rasp/Echo?body=%7B%20%22content%22:%20%22defaultBody%22,%20%22aws%22:%20%22credentials%22%20%7D,
-      language: dotnet,
-      network.client.ip: 127.0.0.1,
-      runtime-id: Guid_1,
-      span.kind: server,
-      _dd.runtime_family: dotnet
-    },
-    Metrics: {
-      process_id: 0,
-      _dd.appsec.enabled: 1.0,
-      _dd.top_level: 1.0,
-      _dd.tracer_kr: 1.0,
-      _sampling_priority_v1: 1.0
-    }
-  },
-  {
-    TraceId: Id_1,
-    SpanId: Id_5,
-    Name: aspnet_core_mvc.request,
-    Resource: GET /rasp/downstreamtoself,
-    Service: Samples.Security.AspNetCore5,
-    Type: web,
-    ParentId: Id_6,
-    Tags: {
-      aspnet_core.action: downstreamtoself,
-      aspnet_core.controller: rasp,
-      aspnet_core.route: rasp/downstreamtoself,
-      component: aspnet_core,
-      env: integration_tests,
-      language: dotnet,
-      span.kind: server
-    }
-  },
-  {
-    TraceId: Id_1,
-    SpanId: Id_6,
     Name: aspnet_core.request,
     Resource: GET /rasp/downstreamtoself,
     Service: Samples.Security.AspNetCore5,
@@ -102,6 +33,75 @@
       _dd.top_level: 1.0,
       _dd.tracer_kr: 1.0,
       _sampling_priority_v1: 1.0
+    }
+  },
+  {
+    TraceId: Id_1,
+    SpanId: Id_3,
+    Name: aspnet_core_mvc.request,
+    Resource: GET /rasp/downstreamtoself,
+    Service: Samples.Security.AspNetCore5,
+    Type: web,
+    ParentId: Id_2,
+    Tags: {
+      aspnet_core.action: downstreamtoself,
+      aspnet_core.controller: rasp,
+      aspnet_core.route: rasp/downstreamtoself,
+      component: aspnet_core,
+      env: integration_tests,
+      language: dotnet,
+      span.kind: server
+    }
+  },
+  {
+    TraceId: Id_1,
+    SpanId: Id_4,
+    Name: aspnet_core.request,
+    Resource: GET /rasp/echo,
+    Service: Samples.Security.AspNetCore5,
+    Type: web,
+    ParentId: Id_5,
+    Tags: {
+      aspnet_core.endpoint: Samples.Security.AspNetCore5.Controllers.RaspController.EchoGet (Samples.Security.AspNetCore5),
+      aspnet_core.route: rasp/echo,
+      component: aspnet_core,
+      env: integration_tests,
+      http.client_ip: 127.0.0.1,
+      http.method: GET,
+      http.request.headers.host: localhost:00000,
+      http.route: rasp/echo,
+      http.status_code: 200,
+      http.url: http://localhost:00000/Rasp/Echo?body=%7B%20%22content%22:%20%22defaultBody%22,%20%22aws%22:%20%22credentials%22%20%7D,
+      language: dotnet,
+      network.client.ip: 127.0.0.1,
+      runtime-id: Guid_1,
+      span.kind: server,
+      _dd.runtime_family: dotnet
+    },
+    Metrics: {
+      process_id: 0,
+      _dd.appsec.enabled: 1.0,
+      _dd.top_level: 1.0,
+      _dd.tracer_kr: 1.0,
+      _sampling_priority_v1: 1.0
+    }
+  },
+  {
+    TraceId: Id_1,
+    SpanId: Id_6,
+    Name: aspnet_core_mvc.request,
+    Resource: GET /rasp/echo,
+    Service: Samples.Security.AspNetCore5,
+    Type: web,
+    ParentId: Id_4,
+    Tags: {
+      aspnet_core.action: echoget,
+      aspnet_core.controller: rasp,
+      aspnet_core.route: rasp/echo,
+      component: aspnet_core,
+      env: integration_tests,
+      language: dotnet,
+      span.kind: server
     }
   }
 ]


tracer/src/Datadog.Trace/Datadog.Trace.csproj
```diff
@@ -111,8 +111,7 @@
     <None Include="$(DotNetRuntimeFiles)" />
 
     <!-- We use Unsafe.IsNullRef() which was added in .NET 5 -->
-    <Compile Include="Vendors/System.Runtime.CompilerServices.Unsafe/Unsafe.cs"
-        Condition="!$([MSBuild]::IsTargetFrameworkCompatible($(TargetFramework), 'net5.0'))" />
+    <Compile Include="Vendors/System.Runtime.CompilerServices.Unsafe/Unsafe.cs" Condition="!$([MSBuild]::IsTargetFrameworkCompatible($(TargetFramework), 'net5.0'))" />
```
andrewlock: Looks like an unrelated, change right?

tracer/test/Datadog.Trace.Security.IntegrationTests/RASP/AspNetCore5Rasp.cs
```diff
@@ -108,6 +108,10 @@ public async Task TryStartApp()
     [Trait("RunOnWindows", "True")]
     public async Task TestRaspRequest(string url, string exploit)
     {
+#pragma warning disable CS0618 // Type or member is obsolete
+        EnableDebugMode();
+#pragma warning restore CS0618 // Type or member is obsolete
+
```
andrewlock: You shouldn't add this, we do a separate Debug run
```suggestion
```

tracer/src/Datadog.Trace/AppSec/Rasp/HttpClient/IHttpContent.cs
```diff
@@ -0,0 +1,27 @@
...
+
+#nullable enable
+using System.IO;
+using System.Threading.Tasks;
+using Datadog.Trace.DuckTyping;
+
+namespace Datadog.Trace.AppSec.Rasp.HttpClient;
+
+/// <summary>
+/// HttpResponseMessage interface for ducktyping
```
andrewlock: I feel like we probably shouldn't need this duck type on .NET Core 🤔 The `HttpContent` type should be available anyway. We would need it for .NET FX .NET Standard, but in that case we should potentially guard it.... raises the question of why we're duck typing the `IHttpRequestMessage` on .NET Core too... I've asked on slack and we can go from there 😄 

```suggestion
/// HttpContent interface for ducktyping
```

tracer/src/Datadog.Trace/AppSec/Rasp/HttpClient/IHttpRequestMessage.cs
```diff
@@ -0,0 +1,25 @@
...
+using System;
+using Datadog.Trace.ClrProfiler.AutoInstrumentation.Http.HttpClient;
+using Datadog.Trace.DuckTyping;
+
+namespace Datadog.Trace.AppSec.Rasp.HttpClient;
+
+/// <summary>
+/// HttpRequestMessage interface for ducktyping
+/// </summary>
+internal interface IHttpRequestMessage : IDuckType
```
andrewlock: I have a _bit_ of a concern that this will cause confusion, as they're equally named types, which duck type the _same_ underlying type, just in different namespaces 😅 

If these types are always going to be valid, it might make sense to consolidate them, and just add the extra properties to the existing. I'm not sure which is best though tbh, and we can always revisit if we want to

tracer/src/Datadog.Trace/AppSec/Rasp/HttpClient/IMediaTypeHeaderValue.cs
```diff
@@ -0,0 +1,15 @@
...
+#nullable enable
+namespace Datadog.Trace.AppSec.Rasp.HttpClient;
+
+/// <summary>
+/// MediaTypeHeaderValue interface for ducktyping
+/// </summary>
+internal interface IMediaTypeHeaderValue
+{
+    string MediaType { get; }
+}
```
andrewlock: We probably could/should use a `[DuckCopy]` struct here instead. We have [a document here](https://github.com/DataDog/dd-trace-dotnet/blob/master/docs/development/DuckTyping.md#choosing-a-proxy-implementation) with a flow chart on how to choose between them.

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
```diff
@@ -8,12 +8,15 @@
...
-using System.Diagnostics;
+using System.Threading;
 using Datadog.Trace.AppSec.Coordinator;
+using Datadog.Trace.AppSec.Rasp.HttpClient;
 using Datadog.Trace.AppSec.Waf;
 using Datadog.Trace.Configuration;
+using Datadog.Trace.DuckTyping;
 using Datadog.Trace.Logging;
 using Datadog.Trace.Telemetry;
+using Datadog.Trace.VendoredMicrosoftCode.System.Reflection.Internal;
```
andrewlock: Do you really need this? In general we should never add references to VendoredMicrosoftCode directly

```suggestion
```

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
```diff
@@ -122,14 +134,14 @@ private static void CheckVulnerability(Dictionary<string, object> arguments, str
             return;
         }
 
-        var rootSpan = Tracer.Instance.InternalActiveScope?.Root?.Span;
+        rootSpan??= Tracer.Instance.InternalActiveScope?.Root?.Span;
```
andrewlock, line 135: Do we even need this line any more? Can't we make passing in the root span compulsory in the method instead?

(context: I hate that we have static `Tracer.Instance` calls scattered around, it makes understanding dependencies a bit of a nightmare, but it's an existing problem, so not a blocker)

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
```diff
@@ -218,6 +230,8 @@ private static void RunWafRasp(Dictionary<string, object> arguments, Span rootSp
                 var matchSuccesCode = result.ReturnCode == WafReturnCode.Match && result.ShouldBlock ?
                     BlockType.Success : BlockType.Irrelevant;
 
+                if (matchSuccesCode == BlockType.Success) { onBlock?.Invoke(); }
```
andrewlock: nit: style requirements, sorry 😉 I'm kinda surprised it compiles tbh 😅 
```suggestion
                if (matchSuccesCode == BlockType.Success)
                {
                    onBlock?.Invoke();
                }
```

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
```diff
@@ -294,4 +308,177 @@ internal static void OnCommandInjection(string fileName, string argumentLine, Co
             Log.Error(ex, "RASP: Error while checking command injection.");
         }
     }
+
+#if NETCOREAPP3_0_OR_GREATER
```
andrewlock: If this is only going to be .NET Core 3.0+ then we can almost certainly get rid of the duck types 🤔 That will likely improve performance around a bunch of things a bit

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
```diff
@@ -294,4 +308,177 @@ internal static void OnCommandInjection(string fileName, string argumentLine, Co
...
+    {
+        var enumerator = headers.GetEnumerator();
+        Dictionary<string, object>? headersDic = null;
+        while (enumerator.MoveNext())
+        {
+            var key = enumerator.Current.Key;
+            if (!key.Equals("cookie", StringComparison.OrdinalIgnoreCase))
+            {
+                headersDic ??= new Dictionary<string, object>();
+                var currentKey = key.ToLowerInvariant();
```
andrewlock, line 319: Do you _have_ to call ToLower on all the keys, or can we just use a case-insensitive comparer
```suggestion
                headersDic ??= new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
                var currentKey = key;
```

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
```diff
@@ -294,4 +308,177 @@ internal static void OnCommandInjection(string fileName, string argumentLine, Co
...
+                var value = enumerator.Current.Value;
+                if (!headersDic.TryAdd(currentKey, value))
+                {
+                    Log.Warning("Header {Key} couldn't be added as argument to the waf", currentKey);
```
andrewlock: I don't think we should log this warning. Including the same header multiple times is legal HTTP I believe, so this could cause a bunch of noise and log spam

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
```diff
@@ -23,6 +26,11 @@ internal static class RaspModule
     private static readonly IDatadogLogger Log = DatadogLogging.GetLoggerFor(typeof(RaspModule));
     private static bool _nullContextReported = false;
 
+#if NETCOREAPP3_0_OR_GREATER
+    [ThreadStatic]
+    private static bool _processDownstreamRequest = false;
+#endif
```
andrewlock, line 30: This seems like a super fragile way to trigger a downstream request 🤔 Surely we should add something to the context instead, as otherwise any `async` code is likely going to completely break this, and we'll end up processing the wrong request entirely 🤔

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
```diff
@@ -294,4 +308,177 @@ internal static void OnCommandInjection(string fileName, string argumentLine, Co
...
+                if (context is null)
+                {
+                    return false;
+                }
+
+                if (!requestMessageInstance.TryDuckCast<IHttpRequestMessage>(out var requestMessage))
+                {
+                    Log.Error("DuckCast to IHttpRequestMessage failed");
+                    return false;
+                }
```
andrewlock: Generally we shouldn't use `TryDuckCast` if there's no good reason for it to ever fail, as otherwise it fails "silently". But as mentioned, I think we could/should just use `HttpRequestMessage` here directly anyway

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
```diff
@@ -294,4 +308,177 @@ internal static void OnCommandInjection(string fileName, string argumentLine, Co
...
+
+                var wafArgs = new Dictionary<string, object>();
+                wafArgs[AddressesConstants.DownstreamUrl] = requestMessage.RequestUri.ToString();
```
andrewlock: Wondering out loud, do you want the "normalized" version, or the "original string"? 🤔 `OriginalString` has the benefit of not allocating again, but might not be what you want

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
```diff
@@ -294,4 +308,177 @@ internal static void OnCommandInjection(string fileName, string argumentLine, Co
...
+                wafArgs[AddressesConstants.DownstreamRequestMethod] = requestMessage.Method.Method;
+                if (requestMessage is { Headers: var headers } && headers is not null)
```
andrewlock: nit: this is the same thing, just shorter:

```suggestion
                if (requestMessage is { Headers: { } headers })
```

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
```diff
@@ -294,4 +308,177 @@ internal static void OnCommandInjection(string fileName, string argumentLine, Co
...
+                    }
+                }
+
+                if (context.IsHttpClientRequestSampled(requestSpanId))
+                {
+                    AddBody(requestMessage.Content, wafArgs, AddressesConstants.DownstreamRequestBody, security.AppSecBodyParsingSizeLimit);
+                }
+
+                // If a block is issued we must stop current child outbound request span, as the call is going to be interrupted
+                CheckVulnerability(wafArgs, AddressesConstants.DownstreamUrl, rootSpan, () => Tracer.Instance.InternalActiveScope?.Dispose());
```
andrewlock: This looks dangerous, there's no guarantee that the active scope you're disposing here is the one you _want_ to. Plus, it _should_ get disposed automatically by the integration anyway, right? 🤔

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
```diff
@@ -294,4 +308,177 @@ internal static void OnCommandInjection(string fileName, string argumentLine, Co
...
+        }
+
+        return false;
+    }
+
+    internal static void OnDownstreamResponse(object responseMessageInstance, ulong requestSpanId)
+    {
+        try
+        {
+            _processDownstreamRequest = false;
```
andrewlock, line 390: You're _always_ running this, even if `_processDownstreamRequest==false` already, is that correct? 🤔

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
```diff
@@ -294,4 +308,177 @@ internal static void OnCommandInjection(string fileName, string argumentLine, Co
...
+                return;
+            }
+
+            var context = rootSpan.Context.TraceContext.AppSecRequestContext;
+            if (context is null)
+            {
+                return;
+            }
+
+            if (!responseMessageInstance.TryDuckCast<IHttpResponseMessage>(out var requestMessage))
```
andrewlock: As mentioned above, I think we can use `HttpResponseMessage` directly.

nit: also, fix the naming, `requestMessage` -> `responseMessage`

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
```diff
@@ -294,4 +308,177 @@ internal static void OnCommandInjection(string fileName, string argumentLine, Co
...
+    private static void AddBody(IHttpContent content, Dictionary<string, object> wafArgs, string wafAddress, long bodySizeLimit)
+    {
+        try
+        {
+            if (content?.Instance is not null)
+            {
+                var contentType = content.Headers?.ContentType?.MediaType;
+                if (contentType is "application/json" or "text/plain")
+                {
+                    content.LoadIntoBufferAsync().SafeWait();
```
andrewlock: This is nasty... is there no way to make this code (and callsite) async? 🤔 Given we're (presumably) making async requests, it seems like we _Should_ be able to slot our own async code in there? 🤔

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
```diff
@@ -294,4 +308,177 @@ internal static void OnCommandInjection(string fileName, string argumentLine, Co
...
+                        var body = content.ReadAsStringAsync().SafeGetResult();
+                        if (BodyParser.Parse(body) is { } parsedBody)
+                        {
+                            wafArgs[AddressesConstants.DownstreamRequestBody] = parsedBody;
+                        }
+                    }
+                }
+            }
+        }
+        catch (Exception ex) when (ex is not BlockException)
```
andrewlock: How can a `BlockException` be thrown here?

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
```diff
@@ -294,4 +308,177 @@ internal static void OnCommandInjection(string fileName, string argumentLine, Co
...
+                    var len = 0L;
+                    var stream = content.ReadAsStreamAsync().SafeGetResult();
+                    if (stream != null)
+                    {
+                        len = stream.Length;
+                    }
+
+                    if (len > 0 && len < bodySizeLimit)
+                    {
+                        var body = content.ReadAsStringAsync().SafeGetResult();
```
andrewlock: This code doesn't feel right at all... 🤔 If we want to know how large the content is, we should check the headers - that way we can potentially no upfront whether it's too large to bother with. 

As for reading as a string, we _definitely_ shouldn't do that, because we end up reading the string, then converting it to bytes, and then the consumer is going to try reading it all again... that just feels horribly allocatey...

I assume the `LoadIntoBufferAsync()` is to try to ensure the consumer can still read the stuff you've already written... it's a nice idea if it works 😅 I haven't checked the tests, but I hope we've tested with all sorts of things, because LoadIntoBufferAsync() could easily just get a single character read etc...

tracer/src/Datadog.Trace/AppSec/Rasp/BodyParser.cs
```diff
@@ -0,0 +1,134 @@
...
+        if (!jsonReader.Read())
+        {
+            return null;
+        }
+
+        try
+        {
+            State state = new(MaxElements);
+            return ReadValue(jsonReader, ref state, 0);
+        }
```
andrewlock, line 42: It's very likely we're only going to load partial data given how we're loading the stream, which means we're very likely to throw, and this is very likely to always return null after doing a bunch of expensive work... not sure there's a good solution to that currently, other than you buffering the whole request

tracer/src/Datadog.Trace/AppSec/Rasp/HttpClient/IHttpRequestMessage.cs
dromanol: That was my first option, but then chose to separate them in order not to affect normal APM witho no ASM behavior (as I guess more properties would add a bit of overhead)

tracer/src/Datadog.Trace/AppSec/Rasp/HttpClient/IMediaTypeHeaderValue.cs
dromanol: Agree

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
dromanol: Agree (it must have been added automathically by the IDE)

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
dromanol: Agree, but in this case this is always executed in the same thread, as it's a callsite just before the call instruction. But we can discuss this live

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
dromanol: Agree

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
dromanol: You are probably right. I'll give it a shot

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
dromanol: The thing is waf wants all the header keys in lower case, so...

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
dromanol: Agree

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
dromanol: I added this because the scope never got disposed if we blocked (the AfterCall is never invoked)

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
dromanol: Just to be sure, but I guess this could trigger some sort of memory barrier. I'll add a check

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
dromanol: This has been the main choke point of the PR. I've found no way of retrieving the content length (no headers, no TryComputeLen, no nothing...)

dromanol: `LoadIntoBufferAsync loads the buffer into memory, which is not ideal, but also lets the response to be re-read`

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
dromanol: Copy pasta. Fixing it

tracer/src/Datadog.Trace/Datadog.Trace.csproj
dromanol: Right. Reverting it

tracer/test/Datadog.Trace.Security.IntegrationTests/RASP/AspNetCore5Rasp.cs
dromanol: It slipped before removing it

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
dromanol: Done

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
dromanol: Done

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
dromanol: :+1:

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
andrewlock: That seems like a problem 🤔  _Why_ didn't it get disposed? Even if we throw, we should currently catch that in our integrations and still dispose the scope. And regardless, this approach still seems dangerous, as you're "refetching" the current active scope in the lambda, and then disposing it regardless 🤔 

I would suggest we first look at _why_ the existing integrations aren't disposing the scope, and go from there...

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
andrewlock: I guess I just don't understand the expected flow here tbh

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
andrewlock: >  I've found no way of retrieving the content length (no headers, no TryComputeLen, no nothing...)

You should look for the `Content-Length` header. If it's a fixed size response (and not a `Transfer-Encoding` streamed response) then it _will_ be there (it has to be)

tracer/src/Datadog.Trace/AppSec/AppSecRequestContext.cs
```diff
@@ -26,6 +26,7 @@ internal sealed partial class AppSecRequestContext
     private readonly object _sync = new();
     private readonly RaspMetricsHelper? _raspMetricsHelper = Security.Instance.RaspEnabled ? new RaspMetricsHelper() : null;
     private readonly List<object> _wafSecurityEvents = new();
+    private readonly ConcurrentDictionary<ulong, bool> _sampledHttpClientRequests = new();
```
andrewlock, line 29: This is an unbounded collection that is just going to grown indefinitely which seems... bad 😅

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
andrewlock: To be clear, you need to look on the Headers on the `HttpContent` object, not the HttpResponseMessage object IIRC

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
```diff
@@ -294,4 +305,177 @@ internal static void OnCommandInjection(string fileName, string argumentLine, Co
...
+    private static async Task AddBody(System.Net.Http.HttpContent? content, Dictionary<string, object> wafArgs, string wafAddress, long bodySizeLimit)
+    {
+        try
+        {
+            if (content is not null)
+            {
+                var contentType = content.Headers?.ContentType?.MediaType;
+                if (contentType is "application/json")
+                {
+                    await content.LoadIntoBufferAsync().ConfigureAwait(false);
```
andrewlock: Looking at the framework code, this scares me. By default, it looks like `LoadIntoBufferAsync()` loads `int.MaxValue` bytes when you don't provide a max buffer size. Would need to check properly, but if so, that's a hell of an allocation. We should probably pass in the `bodySizeLimit` here at the very least (which is _still_ huge, but still).

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
```diff
@@ -294,4 +305,177 @@ internal static void OnCommandInjection(string fileName, string argumentLine, Co
...
+                    var len = 0L;
+                    var stream = content.ReadAsStreamAsync().Result;
```
andrewlock: You've already got `async` in this method, so make _everything_ async in here 😉

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Http/HttpClient/HttpMessageHandlerCommon.cs
```diff
@@ -43,7 +45,17 @@ public static CallTargetState OnMethodBegin<TTarget, TRequest>(TTarget instance,
...
 
                     tracer.TracerManager.Telemetry.IntegrationGeneratedSpan(implementationIntegrationId ?? integrationId);
-                    return new CallTargetState(scope);
+
+                    bool executeOnDownstreamResponse = false;
+#if NETCOREAPP3_0_OR_GREATER
+                    if (requestMessage.Instance is { } request)
+                    {
+                        var rootSpan = tracer.InternalActiveScope?.Root?.Span;
+                        executeOnDownstreamResponse = RaspModule.OnDownstreamRequest(request, scope.Span.SpanId, rootSpan).Result;
```
andrewlock: `Result` and `SafeWait` in these paths is probably not going to be ok 😟

tracer/src/Datadog.Trace/AppSec/AppSecRequestContext.cs
dromanol: This is just for the context of the request and it's limited by `ApiSecurityMaxDownstreamRequestBodyAnalysis`

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
```diff
@@ -89,7 +94,11 @@ internal static void OnLfi(string file)
 
     internal static void OnSSRF(string url)
     {
-        CheckVulnerability(new Dictionary<string, object> { [AddressesConstants.UrlAccess] = url }, AddressesConstants.UrlAccess);
+#if !NETCOREAPP3_0_OR_GREATER
+        CheckVulnerability(new Dictionary<string, object> { [AddressesConstants.DownstreamUrl] = url }, AddressesConstants.DownstreamUrl);
+#else
+        _processDownstreamRequest = true;
```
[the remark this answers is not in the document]
dromanol: On NetCore the SSRF is performed in the downstream request check

tracer/src/Datadog.Trace/AppSec/Security.cs
```diff
@@ -120,6 +123,10 @@ public static Security Instance
 
         internal bool RaspEnabled => _settings.RaspEnabled && AppsecEnabled;
 
+        internal long ApiSecurityMaxDownstreamRequestBodyAnalysis => _settings.ApiSecurityMaxDownstreamRequestBodyAnalysis;
```
e-n-0, line 126: Not sure to understand why is it declared as a `long` but `_settings.ApiSecurityMaxDownstreamRequestBodyAnalysis` is an `int`.

tracer/src/Datadog.Trace/AppSec/Security.cs
```diff
@@ -71,6 +73,7 @@ public Security(SecuritySettings? settings = null, IWaf? waf = null, IRcmSubscri
                 _waf = waf;
                 _configurationState = configurationState ?? new ConfigurationState(_settings, telemetry, _waf is null);
                 LifetimeManager.Instance.AddShutdownTask(RunShutdown);
+                _downstreamSampler = downstreamSampler ?? new DownstreamSampler(_settings.ApiSecurityMaxDownstreamRequestBodyAnalysis);
```
e-n-0: DownstreamSampler is taking a `double` and clamps the value for the sample rate. Is it expected that  `_settings.ApiSecurityMaxDownstreamRequestBodyAnalysis` (an `int`) is passed as argument here?

tracer/src/Datadog.Trace/AppSec/Rasp/BodyParser.cs
andrewlock: Discussed offline - because we bail if we have a "full" buffer, then this _shouldn't_ be the case

tracer/src/Datadog.Trace/AppSec/Rasp/DownstreamSampler.cs
```diff
@@ -0,0 +1,70 @@
...
+        const double max = 18446744073709551615.0;
+
+        if (rate < 0.5)
+        {
+            return (long)(rate * max) + long.MinValue;
+        }
+        else if (rate < 1.0)
+        {
+            return (long)((rate * max) + long.MinValue);
+        }
```
andrewlock: AFAICT, both of these branches are the same? 🤔 

Also, are we considering overflow correctly here? 🤔 Should this be checked or unchecked?

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
andrewlock: Agreed, I think it's fine here because it's literally one instrumentation wrapping the other. Makes me think we should potentially have a first class way of doing this though (unrelated musing...)

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
andrewlock: I guess we can/should fix it in a follow up PR?

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
andrewlock: We should probably remove it before merging

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
andrewlock: My point is, you're _always_ running this method, even if we're _not_ in a callsite scenario, right? Because in those cases `_processDownstreamRequest==false`

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
```diff
@@ -294,4 +303,178 @@ internal static void OnCommandInjection(string fileName, string argumentLine, Co
...
+                var contentType = content.Headers?.ContentType?.MediaType;
+                if (contentType is "application/json")
+                {
+                    var len = content.Headers?.ContentLength ?? 0;
+                    if (len > bodySizeLimit)
+                    {
+                        return;
+                    }
+
+                    await content.LoadIntoBufferAsync(len > 0 ? len : bodySizeLimit).ConfigureAwait(false);
```
andrewlock: I _think_ you only actually need to pass in `bodySizeLimit` here, but this is fine 🙂

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
```diff
@@ -294,4 +303,178 @@ internal static void OnCommandInjection(string fileName, string argumentLine, Co
...
+    private static async Task AddBody(System.Net.Http.HttpContent? content, Dictionary<string, object> wafArgs, string wafAddress, long bodySizeLimit)
+    {
+        try
+        {
+            if (content is not null)
+            {
+                var contentType = content.Headers?.ContentType?.MediaType;
+                if (contentType is "application/json")
+                {
+                    var len = content.Headers?.ContentLength ?? 0;
```
andrewlock, line 447: I think we should add some comments to this code in general, because it's super confusing and relies on internal behaviours, and we're definitely not going to remember the details in the future😅 
```suggestion
                    // This attempts to read the content length from the Content-Length header
                    // if provided. That tells us if the content is too large
                    // before we do anything expensive, and also ensures that we can safely 
                    // load the data into the buffer (so that it cab be re-read later)
                    var len = content.Headers?.ContentLength ?? 0;
```

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
andrewlock:
```suggestion
                    // This attempts to fill the buffer of the stream, without loading 
                    // the _entire_ content of the stream into memory
                    await content.LoadIntoBufferAsync(len > 0 ? len : bodySizeLimit).ConfigureAwait(false);
```

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
```diff
@@ -294,4 +303,178 @@ internal static void OnCommandInjection(string fileName, string argumentLine, Co
...
+                    if (len > bodySizeLimit)
+                    {
+                        return;
+                    }
+
+                    await content.LoadIntoBufferAsync(len > 0 ? len : bodySizeLimit).ConfigureAwait(false);
+                    len = content.Headers?.ContentLength ?? 0;
```
andrewlock:
```suggestion
                    // LoadIntoBufferAsync sets the content-length to the 
                    // length of the buffered content (unless we know the real value)
                    len = content.Headers?.ContentLength ?? 0;
```

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
```diff
@@ -294,4 +303,178 @@ internal static void OnCommandInjection(string fileName, string argumentLine, Co
...
+
+                    if (len > 0 && len <= bodySizeLimit)
```
andrewlock: We shouldn't do this if `len == bodySizeLimit`, because that means we are likely too big (because there's more data to come). It's also still not clear to me that `ReadAsStreamAsync()` _won't_ buffer everything in to memory if there's a very large response. Do we have any tests for that?

tracer/src/Datadog.Trace/AppSec/SecuritySettings.cs
```diff
@@ -280,6 +292,22 @@ public SecuritySettings(IConfigurationSource? source, IConfigurationTelemetry te
...
 
+        /// <summary>
+        /// Gets the number of downstream requests (per request) on which the request and response body are analysed.
+        /// </summary>
+        public int ApiSecurityMaxDownstreamRequestBodyAnalysis { get; }
+
+        /// <summary>
+        /// Gets the max size of the request body for being parsed.
+        /// </summary>
+        public int AppSecBodyParsingSizeLimit { get; }
```
andrewlock, line 303: nit: can we either add "in bytes" to the comment, or add the suffix `Bytes`

tracer/src/Datadog.Trace/Configuration/supported-configurations.yaml
```diff
@@ -150,6 +164,13 @@ supportedConfigurations:
     product: AppSec
     const_name: UserEventsAutoInstrumentationMode
     documentation: Automatic instrumentation of user event mode. Values can be ident, disabled, anon.
+  DD_APPSEC_BODY_PARSING_SIZE_LIMIT:
+  - implementation: A
+    type: int
+    default: '10000000'
+    product: AppSec
+    const_name: AppSecBodyParsingSizeLimit
+    documentation: Configuration key for limiting the size of the request body being parsed. Default value is 10_000_000.
```
andrewlock, line 173:
```suggestion
  DD_APPSEC_BODY_PARSING_SIZE_LIMIT:
  - implementation: A
    type: int
    default: '84000'
    product: AppSec
    const_name: AppSecBodyParsingSizeLimit
    documentation: Configuration key for limiting the size of the request body being parsed in Bytes. Default value is 84_000.
```

tracer/test/Datadog.Trace.Security.Unit.Tests/RASP/BodyParserTests.cs
```diff
@@ -0,0 +1,349 @@
...
+
+        resultTrue.Should().Be(true);
+        resultFalse.Should().Be(false);
+    }
+
+    [Fact]
+    public void Parse_PrimitiveNull_ReturnsNull()
+    {
+        var json = @"null";
+        var result = BodyParser.Parse(json);
```
andrewlock: AFACIT these tests all test the "string json"  path, whereas AFAICT we only use the stream version. Maybe we should just delete the string API entirely, and always call the stream version here?

tracer/test/Datadog.Trace.Security.Unit.Tests/RASP/BodyParserTests.cs
```diff
@@ -0,0 +1,349 @@
...
+using Datadog.Trace.Security.Unit.Tests.Utils;
+using FluentAssertions;
+using Xunit;
+
+namespace Datadog.Trace.Security.Unit.Tests.RASP;
+
+public class BodyParserTests
+{
+    [Fact]
+    public void Parse_NullString_ReturnsNull()
```
andrewlock, line 20: I think you should have some tests for invalid cases, e.g. invalid JSON of various forms

tracer/test/Datadog.Trace.Security.Unit.Tests/RASP/RaspModuleDownstreamTests.cs
```diff
@@ -0,0 +1,217 @@
...
+
+    [Fact]
+    public void AddBody_OversizedContent_SkipsBodyParsing()
+    {
+        var largeBody = new string('a', 100_000);
+        var mockContent = HttpMocks.CreateMockContent(largeBody, "application/json", 100_000);
+        var wafArgs = new Dictionary<string, object>();
+
+        // Use reflection to call the private AddBody method
+        var method = typeof(RaspModule).GetMethod("AddBody", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Static);
```
andrewlock: Just make the method public and mark it with `[TestingOnly]` - this reflection is horribly fragile 😅

tracer/test/Datadog.Trace.Security.Unit.Tests/RASP/RaspModuleDownstreamTests.cs
```diff
@@ -0,0 +1,217 @@
...
+        method.Should().NotBeNull();
+
+        method!.Invoke(null, [mockContent, wafArgs, AddressesConstants.DownstreamRequestBody, 10_000_000L]);
+
+        wafArgs.Should().NotContainKey(AddressesConstants.DownstreamRequestBody);
+    }
+
+    [Fact]
+    public void AddBody_InvalidJson_DoesNotAddBody()
```
andrewlock: I think there are a couple of key cases missing here, namely proper "streaming" responses, e.g. chunked encoding. I think we really should add those test cases to be confident it's always behaving as expected (where the content length can't be detected up front)

tracer/test/Datadog.Trace.Security.Unit.Tests/Utils/HttpMocks.cs
```diff
@@ -0,0 +1,82 @@
...
+        var request = new HttpRequestMessage(new HttpMethod(method), new Uri(url));
+
+        foreach (var kvp in headers ?? [])
+        {
+            request.Headers.TryAddWithoutValidation(kvp.Key, kvp.Value);
+        }
+
+        if (body is not null)
+        {
+            request.Content = new StringContent(body, Encoding.UTF8, contentType ?? "application/json");
```
andrewlock: Using `StringContent` is unfortunately not a good test for real life - it _knows_ the content size up front, which will not be the case for many responses.

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
dromanol: It's what the HttpContent does under the hood

tracer/test/Datadog.Trace.Security.Unit.Tests/RASP/BodyParserTests.cs
dromanol: Fixed. Removed the overload and migrated tests to use the Stream version

tracer/test/Datadog.Trace.Security.Unit.Tests/RASP/BodyParserTests.cs
dromanol: Added some

tracer/test/Datadog.Trace.Security.Unit.Tests/RASP/RaspModuleDownstreamTests.cs
dromanol: Oh, did not know that trick

tracer/test/Datadog.Trace.Security.Unit.Tests/RASP/RaspModuleDownstreamTests.cs
dromanol: Added some chunked HttpContent tests

tracer/src/Datadog.Trace/AppSec/Rasp/DownstreamSampler.cs
dromanol: Same thing. Fixing it

tracer/src/Datadog.Trace/AppSec/Security.cs
dromanol: Well seen. It's `ApiSecurityDownstreamBodyAnalysisSampleRate`

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Http/HttpClient/HttpMessageHandlerCommon.cs
```diff
@@ -78,6 +98,7 @@ public static TResponse OnMethodEnd<TTarget, TResponse>(TTarget instance, TRespo
                 if (responseMessage is System.Net.Http.HttpResponseMessage response)
                 {
                     var statusCode = (int)response.StatusCode;
+                    RaspModule.OnDownstreamResponse(response, scope.Span.SpanId);
```
andrewlock: You're passing `executeOnDownstreamResponse` in the state, but never checking it, I think it should be this?


```suggestion
                    if (state.State is true)
                    {
                        RaspModule.OnDownstreamResponse(response, scope.Span.SpanId);
                    }
```

tracer/test/Datadog.Trace.Security.Unit.Tests/RASP/BodyParserTests.cs
andrewlock: You're still not parsing any _invalid_ JSON though right? 🤔  I think you could do something as simple as 
```suggestion
    [Theory]
    [InlineData("""{"name":"test","age":30""")]
    [InlineData("""{"name":"test","age":""")]
    [InlineData("""{"name":"test","age" """)]
    [InlineData("""{"name":"test",""")]
    public void Parse_TruncatedString_ReturnsNull(string json)
    {
        BodyParser.Parse(json).Should.BeNull();
        ParseBody(json).Should().BeNull();
    }

    [Fact]
    public void Parse_NullString_ReturnsNull()
```

tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Http/HttpClient/HttpMessageHandlerCommon.cs
dromanol: Agree. Solved

tracer/src/Datadog.Trace/AppSec/SecuritySettings.cs
```diff
@@ -280,6 +292,22 @@ public SecuritySettings(IConfigurationSource? source, IConfigurationTelemetry te
         /// </summary>
         public bool ApiSecurityEnabled { get; }
 
+        /// <summary>
+        /// Gets the number of downstream requests (per request) on which the request and response body are analysed.
+        /// </summary>
+        public int ApiSecurityMaxDownstreamRequestBodyAnalysis { get; }
+
+        /// <summary>
+        /// Gets the max size of the request body for being parsed.
```
andrewlock:
```suggestion
        /// Gets the max size of the request body in bytes for being parsed.
```

tracer/test/Datadog.Trace.Security.Unit.Tests/RASP/RaspModuleDownstreamTests.cs
```diff
@@ -0,0 +1,235 @@
...
+        var chunkedContent = HttpMocks.CreateLargeChunkedContent(sizeInBytes: 10_000, "application/json");
+        var wafArgs = new Dictionary<string, object>();
+
+        await RaspModule.AddBody(chunkedContent, wafArgs, AddressesConstants.DownstreamResponseBody, bodySizeLimit);
+
+        // LoadIntoBufferAsync throws when content exceeds bodySizeLimit; AddBody catches and logs,
+        // so the body must not reach the WAF.
+        wafArgs.Should().NotContainKey(AddressesConstants.DownstreamResponseBody);
+    }
+
```
andrewlock:
```suggestion
    [Theory]
    [InlineData(500, 1_000L)
    [InlineData(1_000, 1_000L)
    [InlineData(10_000, 1_000L)
    public async Task AddBody_ChunkedEncoding_SkipsBody(sizeInBytes int, long bodySizeLimit, )
    {
        var chunkedContent = HttpMocks.CreateLargeChunkedContent(sizeInBytes: sizeInBytes, "application/json");
        var wafArgs = new Dictionary<string, object>();

        await RaspModule.AddBody(chunkedContent, wafArgs, AddressesConstants.DownstreamResponseBody, bodySizeLimit);

        wafArgs.Should().NotContainKey(AddressesConstants.DownstreamResponseBody);
    }

```

tracer/test/Datadog.Trace.Security.Unit.Tests/RASP/RaspModuleDownstreamTests.cs
```diff
@@ -0,0 +1,230 @@
...
+        // Build a JSON payload where the first bytes are syntactically incomplete
+        // (simulate a large array whose closing bracket is beyond the limit).
+        const long bodySizeLimit = 1_000L;
+        var chunkedContent = HttpMocks.CreateLargeChunkedContent(sizeInBytes: 900, "application/json", incomplete: true);
+        var wafArgs = new Dictionary<string, object>();
+
+        await RaspModule.AddBody(chunkedContent, wafArgs, AddressesConstants.DownstreamResponseBody, bodySizeLimit);
+
+        // The WAF must never receive a truncated/partial JSON body.
+        wafArgs.Should().NotContainKey(AddressesConstants.DownstreamResponseBody);
```
andrewlock: I don't think this test adds anything now

```suggestion
```

tracer/src/Datadog.Trace/AppSec/Rasp/RaspModule.cs
```diff
@@ -294,4 +304,165 @@ internal static void OnCommandInjection(string fileName, string argumentLine, Co
...
+        {
+            if (content is not null)
+            {
+                var contentType = content.Headers?.ContentType?.MediaType;
+                if (contentType is "application/json")
+                {
+                    // This attempts to read the content length from the Content-Length header
+                    // if provided. That tells us if the content is too large
+                    // before we do anything expensive, and also ensures that we can safely 
+                    // load the data into the buffer (so that it cab be re-read later)
```
andrewlock, line 446:
```suggestion
                    // This attempts to read the content length from the Content-Length header
                    // if provided. That tells us if the content is too large
                    // before we do anything expensive, and also ensures that we can safely
                    // load the data into the buffer (so that it cab be re-read later)
```

tracer/test/Datadog.Trace.Security.Unit.Tests/RASP/RaspModuleDownstreamTests.cs
```diff
@@ -0,0 +1,212 @@
...
+
+        await RaspModule.AddBody(mockContent, wafArgs, AddressesConstants.DownstreamRequestBody, 10_000_000L);
+
+        wafArgs.Should().NotContainKey(AddressesConstants.DownstreamRequestBody);
+    }
+
+    [Theory]
+    [InlineData(500, 1_000L)
+    [InlineData(1_000, 1_000L)
+    [InlineData(10_000, 1_000L)
```
andrewlock:
```suggestion
    [InlineData(500, 1_000L)]
    [InlineData(1_000, 1_000L)]
    [InlineData(10_000, 1_000L)]
```

tracer/test/Datadog.Trace.Security.Unit.Tests/RASP/RaspModuleDownstreamTests.cs
```diff
@@ -0,0 +1,212 @@
...
+        await RaspModule.AddBody(mockContent, wafArgs, AddressesConstants.DownstreamRequestBody, 10_000_000L);
+
+        wafArgs.Should().NotContainKey(AddressesConstants.DownstreamRequestBody);
+    }
+
+    [Theory]
+    [InlineData(500, 1_000L)]
+    [InlineData(1_000, 1_000L)]
+    [InlineData(10_000, 1_000L)]
+    public async Task AddBody_ChunkedEncoding_SkipsBody(sizeInBytes int, long bodySizeLimit, )
```
andrewlock:
```suggestion
    public async Task AddBody_ChunkedEncoding_SkipsBody(sizeInBytes int, long bodySizeLimit)
```

tracer/test/Datadog.Trace.Security.Unit.Tests/RASP/RaspModuleDownstreamTests.cs
```diff
@@ -0,0 +1,212 @@
...
+        await RaspModule.AddBody(mockContent, wafArgs, AddressesConstants.DownstreamRequestBody, 10_000_000L);
+
+        wafArgs.Should().NotContainKey(AddressesConstants.DownstreamRequestBody);
+    }
+
+    [Theory]
+    [InlineData(500, 1_000L)]
+    [InlineData(1_000, 1_000L)]
+    [InlineData(10_000, 1_000L)]
+    public async Task AddBody_ChunkedEncoding_SkipsBody(sizeInBytes int, long bodySizeLimit)
```
andrewlock:
```suggestion
    public async Task AddBody_ChunkedEncoding_SkipsBody(int sizeInBytes, long bodySizeLimit)
```

PR #8237, opened 2026-02-23
Remote configuration tests and performance improvements

andrewlock:
## Summary of changes

- Add tests for `RcmSubscriptionManager` and `RemoteConfigurationPath`
- Replace regex with string comparison in `RemoteConfigurationPath`

## Reason for change

- We were missing unit tests for remote config stuff, and I want to improve it without breaking things
- The `RemoteConfigurationPath` is running a `Regex` on every file listed in the remote config response (which happens every 5s), but it's a simple pattern that can be easily directly implemented

## Implementation details

- Used 🤖 to generate bunch of tests, and verified they are really how we want things to work
  - High level tests for `RcmSubscriptionManager`
  - Tests for `RemoteConfigurationPath` covering changes in this PR
- More 🤖 in the conversion, but it's relatively simple, once you decode the allowed patterns from the Regex 😄 

## Test coverage

Unit tests in this PR cover compatibility with the existing implementation.

Simple benchmarking for the regex improvements:

| Method                                   | Runtime            |     Mean |   Error | Allocated |
| ---------------------------------------- | ------------------ | -------: | ------: | --------: |
| RemoteConfigurationPathFromPath_Original | .NET 10.0          | 181.5 ns | 2.51 ns |     768 B |
| RemoteConfigurationPathFromPath_Updated  | .NET 10.0          |  54.8 ns | 3.90 ns |     152 B |
|                                          |                    |          |         |           |
| RemoteConfigurationPathFromPath_Original | .NET 6.0           | 204.0 ns | 2.64 ns |     768 B |
| RemoteConfigurationPathFromPath_Updated  | .NET 6.0           |  66.4 ns | 1.13 ns |     152 B |
|                                          |                    |          |         |           |
| RemoteConfigurationPathFromPath_Original | .NET Core 2.1      | 296.9 ns | 4.09 ns |     872 B |
| RemoteConfigurationPathFromPath_Updated  | .NET Core 2.1      |  82.2 ns | 2.21 ns |     160 B |
|                                          |                    |          |         |           |
| RemoteConfigurationPathFromPath_Original | .NET Core 3.1      | 281.0 ns | 3.79 ns |     768 B |
| RemoteConfigurationPathFromPath_Updated  | .NET Core 3.1      |  72.8 ns | 1.90 ns |     152 B |
|                                          |                    |          |         |           |
| RemoteConfigurationPathFromPath_Original | .NET Framework 4.8 | 326.7 ns | 2.11 ns |     875 B |
| RemoteConfigurationPathFromPath_Updated  | .NET Framework 4.8 | 110.0 ns | 1.76 ns |     160 B |

<details><summary>Benchmarking code</summary>
<p>

```csharp
[MemoryDiagnoser, GroupBenchmarksBy(BenchmarkLogicalGroupRule.ByCategory), CategoriesColumn]
public class RemoteConfigBenchmark
{
    private string _pathToTest;
    private string _result;

    [GlobalSetup]
    public void GlobalSetup()
    {
        _pathToTest = "datadog/2/ASM_FEATURES/ASM_FEATURES-third/testname";
    }

    [Benchmark]
    public string RemoteConfigurationPathFromPath_Original()
    {
        var result = OriginalRemoteConfigurationPath.FromPath(_pathToTest);
        _result = result.Id;
        return result.Path;
    }

    [Benchmark]
    public string RemoteConfigurationPathFromPath_Updated()
    {
        var result = RemoteConfigurationPath.FromPath(_pathToTest);
        _result = result.Id;
        return result.Path;
    }
}
```

</p>
</details> 

## Other details

All part of the Remote Config perf stack

### DIFF
--- a/tracer/src/Datadog.Trace/RemoteConfigurationManagement/RemoteConfigurationPath.cs
+++ b/tracer/src/Datadog.Trace/RemoteConfigurationManagement/RemoteConfigurationPath.cs
@@ -3,14 +3,15 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2017 Datadog, Inc.
 // </copyright>
 
-using System.Text.RegularExpressions;
+using System;
 using Datadog.Trace.Util;
 
 namespace Datadog.Trace.RemoteConfigurationManagement
 {
     internal sealed record RemoteConfigurationPath
     {
-        private static readonly Regex PathRegex = new Regex("^(datadog/\\d+|employee)/([^/]+)/([^/]+)/[^/]+$", RegexOptions.Compiled);
+        private const string DatadogPrefix = "datadog/";
+        private const string EmployeePrefix = "employee/";
 
         private RemoteConfigurationPath(string path, string product, string id)
         {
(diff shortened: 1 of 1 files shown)


tracer/src/Datadog.Trace/RemoteConfigurationManagement/RemoteConfigurationPath.cs
```diff
@@ -27,14 +28,72 @@ private RemoteConfigurationPath(string path, string product, string id)
...
+                var secondSlash = path.IndexOf('/', digitStart);
+                if (secondSlash < 0 || secondSlash == digitStart)
+                {
+                    ThrowHelper.ThrowException($"Error parsing path: {path}");
+                }
+
+                // Verify all characters between are digits
+                for (var i = digitStart; i < secondSlash; i++)
+                {
+                    if (path[i] is < '0' or > '9')
```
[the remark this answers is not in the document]
andrewlock: I'm like 99% sure that was probably a bug, but we should check...

tracer/src/Datadog.Trace/RemoteConfigurationManagement/RemoteConfigurationPath.cs
```diff
@@ -27,14 +28,72 @@ private RemoteConfigurationPath(string path, string product, string id)
...
+            {
+                ThrowHelper.ThrowException($"Error parsing path: {path}");
+            }
+
+            // Filename must be non-empty and there must be no more slashes
+            var filenameStart = fourthSlash + 1;
+            if (filenameStart >= path.Length || path.IndexOf('/', filenameStart) >= 0)
             {
                 ThrowHelper.ThrowException($"Error parsing path: {path}");
             }
```
lucaspimentel, line 93: @codex Could this code be implemented with our `SpanCharSplitter` helper from `tracer/src/Datadog.Trace/Util/StringExtensions.cs`?

(@andrewlock, not a blocker, I'm just curious)

lucaspimentel: I didn't ask for a review, silly.

lucaspimentel: @copilot what do you think?

lucaspimentel: I didn't ask you to start a new PR, either 😭

andrewlock: ## Summary of changes

- Replied to review comment explaining the trade-off: `SpanCharSplitter` only splits from position 0 of a `string`, so using it from `productStart` would require `path.Substring(productStart)` (an extra heap allocation), or a full method refactor to iterate all segments from the start (more complexity, no clear benefit).
- No code changes; the existing `IndexOf`-based approach is kept as-is.

## Reason for change

The current parsing is already allocation-free through the scanning phase (only allocating at the final two `Substring` calls for `product` and `id`). Using `SpanCharSplitter` here would not improve performance or readability.

daniel-romano-DD (approved):
Nice

tracer/src/Datadog.Trace/RemoteConfigurationManagement/RemoteConfigurationPath.cs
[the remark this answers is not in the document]
andrewlock: So AFAICT from the agent code [here](https://github.com/DataDog/datadog-agent/blob/5fb1364caaf88074e8a0e514df6de5116368329f/pkg/config/remote/data/file.go#L17-L25) and [here](https://github.com/DataDog/datadog-agent/blob/5fb1364caaf88074e8a0e514df6de5116368329f/pkg/remoteconfig/state/path.go#L16-L24) - the id really is an `org_id`, so only ASCII digits are valid (even though these cases _also_ use `\d` so technically would allow non-ascii decimal digits)

PR #8238, opened 2026-02-23
Reduce payload size sent by agent in remote config

andrewlock:
## Summary of changes

Stop sending the fixed `RootVersion = 1` with every remote config request

## Reason for change

Currently we're sending a fixed value of `RootVersion = 1` for all our remote config requests, but doing so causes the agent to repeatedly send us all the root certificates, significantly increasing the payload size, because it thinks we haven't seen them. Sending the "final" version, acknowledges it, and stops all the extra data, saving ~35kB per call.

## Implementation details

- Deserialize the roots in `GetRcmResponse`, leaving them as base64 encoded `string`s (which is how they are sent in the payload)
- When processing the response, de-encode the _last_ root object, and deserialize it
  - Using the `Base64DecodingStream` introduced in #8226 to avoid extra allocations
  - Using the `IArrayPool` from #8228
  - Introduces a `MinimalTufRoot` (in contrast to `TufRoot`) so that we only materialize what we need (the `roots.signed.version` key)

This implementation avoids ~35kB per call for subsequent remote config requests.

## Test coverage

Added unit test, and did manual test with the real agent, to confirm the expected behaviour (reduction in data sent)

## Other details

All part of the Remote Config perf stack

### DIFF
--- a/tracer/test/Datadog.Trace.Tests/RemoteConfigurationManagement/RcmSubscriptionManagerTests.cs
+++ b/tracer/test/Datadog.Trace.Tests/RemoteConfigurationManagement/RcmSubscriptionManagerTests.cs
@@ -756,6 +756,45 @@ public class RcmSubscriptionManagerTests
         req4.Client.State.BackendClientState.Should().Be("state-v3");
     }
 
+    [Fact]
+    public async Task SendRequest_RootVersion_DefaultsToOneAndUpdatesFromLastRootsEntry()
+    {
+        var manager = new RcmSubscriptionManager();
+        manager.SubscribeToChanges(new Subscription((_, _) => [], RcmProducts.AsmFeatures));
+
+        // First request: root version should default to 1
+        GetRcmRequest captured = null;
+        await manager.SendRequest(
+            CreateTracer(),
+            request =>
+            {
+                captured = request;
+                return Task.FromResult<GetRcmResponse>(null);
+            });
+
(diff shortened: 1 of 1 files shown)


tracer/src/Datadog.Trace/RemoteConfigurationManagement/RcmSubscriptionManager.cs
```diff
@@ -357,5 +363,33 @@ private async Task ProcessResponse(GetRcmResponse response)
...
+        // and extract the version
+
+        var lastRoot = roots[roots.Count - 1];
+        using var stream = new Base64DecodingStream(lastRoot);
+        using var streamReader = new StreamReader(stream);
+        using var jsonReader = new JsonTextReader(streamReader)
+        {
+            ArrayPool = JsonArrayPool.Shared,
+        };
+        var tufRoot = new JsonSerializer().Deserialize<MinimalTufRoot>(jsonReader);
```
lucaspimentel, line 387: Is this `new JsonSerializer().Deserialize()` called every 5 secs? Should we cache and reuse the `JsonSerializer` instance?

andrewlock: Yes it is 👍  It's a good question - the JsonConvert methods all create a throw-away one, so I'm not sure if we actually _need_ a new one each time or not 🤔 I'll look into it, and see if we can/should cache them in general

PR #8248, opened 2026-02-27
Add support for ducktyping private fields and properties  from a base type

andrewlock:
## Summary of changes

Add support for ducktyping fields and properties that are defined on a base type

## Reason for change

While attempting to support the latest NUnit package (#8246) we realised that we can't ducktype fields/properties if they're defined on a base type. E.g.:

```csharp
public class TypeToDuckType : SomeBase {}
public class SomeBase
{
    private int _thisIsTheFieldWeWant;
}

[DuckType]
public struct MyDuckType
{
    [DuckField(Name = "_thisIsTheFieldWeWant")]
    public int MyField;
}
```

If we attempt to duck type `TypeToDuckType` in the above case, we'll fail, and there's no `BindingFlags` we can use to work around it, so we have to manually look up the type hierarchy.

## Implementation details

Add an extra property to `[DuckAttribute]`:

```csharp

[DuckType]
public struct MyDuckType
{
    [DuckField(Name = "_thisIsTheFieldWeWant", FallbackToBaseTypes = true)]
    public int MyField;
}
```

When true, we check up the hierarchy:

```csharp
if (duckAttribute.FallbackToBaseTypes)
{
    var currentType = targetType;
    while (targetProperty is null && currentType is { IsValueType: false, BaseType: not null } && currentType.BaseType != typeof(object))
    {
        currentType = currentType.BaseType;
        targetProperty = GetTargetProperty(currentType, duckAttribute.Name, duckAttribute.BindingFlags);
    }
}
```

We added a flag and didn't just do it automatically for safety reasons.

## Test coverage

Added a bunch of additional unit tests

## Other details

Required for #8246

### DIFF
--- a/tracer/test/Datadog.Trace.DuckTyping.Tests/Fields/ReferenceType/ProxiesDefinitions/IObscureDuckType.cs
+++ b/tracer/test/Datadog.Trace.DuckTyping.Tests/Fields/ReferenceType/ProxiesDefinitions/IObscureDuckType.cs
@@ -7,63 +7,63 @@ namespace Datadog.Trace.DuckTyping.Tests.Fields.ReferenceType.ProxiesDefinitions
 {
     public interface IObscureDuckType
     {
-        [DuckField(Name = "_publicStaticReadonlyReferenceTypeField")]
+        [DuckField(Name = "_publicStaticReadonlyReferenceTypeField", FallbackToBaseTypes = true)]
         string PublicStaticReadonlyReferenceTypeField { get; }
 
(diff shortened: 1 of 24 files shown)


andrewlock:
> LGTM. Could you add a small blurb in `docs/development/DuckTyping.md` about the new DuckAttribute property?

Yep, to avoid triggering CI again unnecessarily, added it in the stacked PR that uses the flag: https://github.com/DataDog/dd-trace-dotnet/pull/8246

PR #4328, opened 2023-06-29
Replace the spinlock by a monitor in dogstatsd

kevingosse:
## Summary of changes

Pulling https://github.com/DataDog/dogstatsd-csharp-client/pull/192 from upstream.

Spinlock shouldn't be used to protect an I/O operation. With the default timeout, the operation can take up to 2 seconds, which ends up using a lot of CPU. This PR replaces it with a Monitor.

PR #8265, opened 2026-03-03
[Metrics] Fixing sync-over-async patterns in DogStatsD Client Fork

link04:
## Summary of changes

Fix sync-over-async patterns in the vendored DogStatsD client and add async disposal for shutdown.

## Reason for change

`NamedPipeTransport.SendBuffer()` calls `WriteAsync().Wait()` on `NamedPipeClientStream` which leads to sync-over-async that can deadlock under thread pool starvation. `AsynchronousWorker.Dispose()` also blocks with `worker.Wait()` on LongRunning tasks, making shutdown slow when the agent is unreachable.

Part of [Enabling .NET Runtime Metrics by Default](https://docs.google.com/document/d/1tekvCvlOkn12pU3jLK-ePZ0nrynSW7p8IWgG-Dh7zwQ/edit?pli=1&tab=t.0#heading=h.7u0xyjk7jxq5). This should land before we enable Runtime Metrics by default since it hardens the shared DogStatsD client that runtime metrics uses.

## Implementation details

- `NamedPipeTransport`: Replace `WriteAsync().Wait()` with synchronous `Write()`. This is a worker thread sending metrics, so sync is the right fit.
- Add `DisposeAsync()` through the disposal chain: `AsynchronousWorker` -> `StatsBufferize` -> `StatsdData` -> `DogStatsdService` -> `StatsdManager`.
- `StatsdClientHolder` uses a `TaskCompletionSource` to signal when async disposal finishes, bridging the sync `Release()` path and the async shutdown path.
- `TracerManager.RunShutdownTasksAsync` now awaits `StatsdManager.DisposeAsync()` instead of synchronous `Dispose()`.

## Test coverage

Existing tests still pass 🤞.

### DIFF
--- a/tracer/src/Datadog.Trace/DogStatsd/StatsdManager.cs
+++ b/tracer/src/Datadog.Trace/DogStatsd/StatsdManager.cs
@@ -316,6 +316,9 @@ internal sealed class StatsdManager : IStatsdManager
             if (Interlocked.Exchange(ref _disposed, 1) == 0)
             {
                 Log.Debug("Disposing DogStatsdService");
+                // We push this to a background thread to avoid disposals running in-line
+                // in consuming code. This flushes any outstanding data too
+                // so could take a relatively long time
                 _ = Task.Run(DisposeClientAsync);
             }
         }
--- a/tracer/src/Datadog.Trace/TracerManager.cs
+++ b/tracer/src/Datadog.Trace/TracerManager.cs
@@ -779,6 +779,10 @@ namespace Datadog.Trace
 
                     instance.RuntimeMetrics?.Dispose();
 
+                    // Fire-and-forget: on master the old sync Dispose() was already
+                    // fire-and-forget internally (Task.Run). DisposeAsync flushes buffers
+                    // and drains worker threads which can take several seconds, longer than
+                    // the window between repeated termination signals on .NET 10.
                     instance.Statsd?.DisposeAsync().ContinueWith(t => Log.Error(t.Exception, "Error waiting for StatsD disposal"), TaskContinuationOptions.OnlyOnFaulted);
 
                     Log.Debug("Finished waiting for disposals.");
--- a/tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
+++ b/tracer/src/Datadog.Trace/Debugger/DynamicInstrumentation.cs
@@ -657,6 +657,10 @@ namespace Datadog.Trace.Debugger
                         .Add(_diagnosticsUploader)
                         .Add(_probeStatusPoller)
                         .DisposeAll();
+
+            // Cannot await here because Dispose() is synchronous and callers hold locks.
+            // On master, _dogStats was disposed via SafeDisposal.Add() which called sync
+            // Dispose() — itself fire-and-forget internally via Task.Run().
             _dogStats?.DisposeAsync().ContinueWith(t => Log.Error(t.Exception, "Error waiting for StatsD disposal"), TaskContinuationOptions.OnlyOnFaulted);
         }
     }
--- a/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/LifetimeManager/TerminationSignalTests.cs
+++ b/tracer/test/Datadog.Trace.ClrProfiler.IntegrationTests/LifetimeManager/TerminationSignalTests.cs
@@ -28,7 +28,7 @@ public class TerminationSignalTests : TestHelper
     {
     }
 
-    [SkippableFact]
+    [SkippableFact(Skip = "There are timing issues in the shutdown after adding DisposeAsync")]
     public async Task SigtermTriggersShutdownOnce_WhenRepeated()
     {
         await RunSigtermTestAsync(signalCount: 2, usePublishWithRid: false);


tracer/src/Datadog.Trace/Vendors/StatsdClient/Transport/NamedPipeTransport.cs
```diff
@@ -19,7 +19,7 @@ internal class NamedPipeTransport : ITransport
 
         public NamedPipeTransport(string pipeName, TimeSpan? timeout = null)
         {
-            _namedPipe = new NamedPipeClientStream(".", pipeName, PipeDirection.Out, PipeOptions.Asynchronous);
```
andrewlock, line 22: I don't think we should be making this change 🤔 I don't see why we should be converting from async to sync operations

tracer/src/Datadog.Trace/Vendors/StatsdClient/Transport/NamedPipeTransport.cs
```diff
@@ -65,16 +65,12 @@ private bool SendBuffer(byte[] buffer, int length, bool allowRetry)
...
+                _namedPipe.Write(buffer, 0, length);
+                return true;
             }
             catch (IOException)
             {
             }
-            catch (AggregateException e) when (e.InnerException is IOException)
-            {
-                // dotnet6.0 raises AggregateException when an IOException occurs.
-            }
```
andrewlock, line 77: I don't see why this should be deleted

tracer/src/Datadog.Trace/Vendors/StatsdClient/Transport/NamedPipeTransport.cs
```diff
@@ -65,16 +65,12 @@ private bool SendBuffer(byte[] buffer, int length, bool allowRetry)
 
             try
             {
-                // WriteAsync overload with a CancellationToken instance seems to not work.
-                return _namedPipe.WriteAsync(buffer, 0, length).Wait(_timeout);
+                _namedPipe.Write(buffer, 0, length);
```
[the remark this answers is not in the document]
andrewlock: Yeah, what the AI says seems reasonable, we potentially need to investigate the source of why CancellationToken doesn't work 🤔

tracer/src/Datadog.Trace/DogStatsd/StatsdManager.cs
andrewlock: If `DisposeAsync` is the new "correct" API, then we should remove `Dispose` entirely, and update the `IStatsdManager` interface

tracer/src/Datadog.Trace/DogStatsd/StatsdManager.cs
```diff
@@ -164,6 +164,24 @@ public void Dispose()
...
+        {
+            previous = _current;
+            _current = null;
+        }
+
+        if (previous is not null)
+        {
+            previous.MarkClosing();
+            await previous.WaitForDisposalAsync().ConfigureAwait(false);
+        }
```
andrewlock: Most of this work is duplicating the `EnsureClient` API, which feels a _little_ bit gross 🤔 I think we could remove the duplication with a small tweak to `EnsureClient()`:

```diff
-   private void EnsureClient(bool ensureCreated, bool forceRecreate)
+   private void EnsureClient(bool ensureCreated, bool forceRecreate, out StatsdClientHolder? previous)
    {
-       StatsdClientHolder? previous;
        Log.Debug("Recreating statsdClient: Create new client: {CreateClient}, Force recreate: {ForceRecreate}", ensureCreated, forceRecreate);
```

(I suggested `out` param instead of a return value, so that it's very clear that what you're getting back is _not_ the client you "ensured", it's the _previous_ client (if any))

Then the `DisposeAsync()` method becomes (very similar to sync original):

```csharp
public async Task DisposeAsync()
{
    _settingSubscription.Dispose();
    // We swap out the client to trigger flushing and disposal
    EnsureClient(ensureCreated: false, forceRecreate: true, out var previous);

    if (previous is not null)
    {
        await previous.WaitForDisposalAsync().ConfigureAwait(false);
    }
}
```

tracer/src/Datadog.Trace/DogStatsd/StatsdManager.cs
```diff
@@ -300,23 +321,37 @@ public void MarkClosing()
...
         }
 
+        public Task WaitForDisposalAsync() => _disposalComplete.Task;
+
         private void Dispose()
         {
             if (Interlocked.Exchange(ref _disposed, 1) == 0)
             {
                 Log.Debug("Disposing DogStatsdService");
+                _ = DisposeClientAsync();
```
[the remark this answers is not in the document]
andrewlock: Yeah, I think this is going to be an issue, we likely still need to push this to a background thread, even though we're doing async:

```suggestion
                _ = Task.Run(DisposeClientAsync);
```

tracer/src/Datadog.Trace/Vendors/StatsdClient/Transport/NamedPipeTransport.cs
link04: So the original attempt was to remove the **sync-over-async** `WriteAsync(...).Wait(_timeout)` pattern and using AI I got to this point as a good approach for this change(trust me I tried learning why do this instead of WriteAsync and end up going with it), but now looking after your comment I learned this call runs on a LongRunning task (dedicated thread), so thread pool starvation deadlocks don't apply here like they do in our Dipose() methods, so I'll change it back.

tracer/src/Datadog.Trace/Vendors/StatsdClient/Transport/NamedPipeTransport.cs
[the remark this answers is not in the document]
link04: For this one I took a look and did some research with Claude but some context.
- The comment comes from StatsD vendored code(they replaced how they handle this as well.)
- It was more applicable to how it looked before Kevin's change on the PR we discussed before: [4328](https://github.com/DataDog/dd-trace-dotnet/pull/4328)
```
var cts = new CancellationTokenSource(_timeout);
_namedPipe.WriteAsync(buffer, 0, length).Wait(cts.Token);
```
The reason why it didn't work(We vendored this in in 2020, the issue below opened November of 2019 and solved in 2022 for .NET 7):

On .NET 6 and below, PipeStream.WriteAsync(byte[], int, int, CancellationToken) inherits from Stream, and the base implementation only checks the CancellationToken upfront — once the I/O operation starts, the token is ignored. As Stephen Toub (the .NET runtime maintainer) explained in dotnet/runtime#31390:

> If you don't pass the Asynchronous flag, then the operations can't be done with async I/O (at the Windows level). That also means we don't have an overlapped with which to call CancelIoEx. So when Asynchronous isn't used, these async methods just check the cancellation token upfront, and then queue a work item that does the synchronous operation.

Even with PipeOptions.Asynchronous, the byte array overload of WriteAsync in older runtimes didn't properly wire up cancellation to the underlying overlapped I/O.

The .NET 7 fix:
PR dotnet/runtime#72503 by Stephen Toub added proper cancellation support using CancelSynchronousIo. On .NET 7+, WriteAsync(buffer, offset, count, cancellationToken) should actually work.

**What this means for a follow-up:**
Since the tracer targets net6.0, a proper fix could conditionally use the CancellationToken overload on .NET 7+ (where it works) and keep the current .Wait(token) approach on .NET 6/Framework. But that's added complexity — the current approach works fine for now I imagine and can be revisit in a different PR for improved performance, WDYT?

tracer/src/Datadog.Trace/Vendors/StatsdClient/Transport/NamedPipeTransport.cs
link04: So I removed it because this one shouldn't be thrown based on my research about `Write()` vs `WriteAsync()` which I am adding back.

> Synchronous Write() throws IOException directly — it can't throw AggregateException since there's no Task involved. So the catch became dead code.

tracer/src/Datadog.Trace/DogStatsd/StatsdManager.cs
link04: Thank you! I tried applying that change on my last commit.

tracer/src/Datadog.Trace/DogStatsd/StatsdManager.cs
link04: Makes sense, I also did the same change for IDogStatsd and related.

tracer/src/Datadog.Trace/Vendors/StatsdClient/Transport/NamedPipeTransport.cs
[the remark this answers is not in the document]
andrewlock: I still don't entirely follow this tbh 😅 

> If you don't pass the Asynchronous flag

We _do_, pass that flag today. Note that the "asynchronous" vs "synchornous" in the named pipe implementations is _different_ to `async` in C# (and no, I don't understand all that stuff either currently 😅) But interestingly we still _don't_ pass the cancellation token to the `WriteAsync` call which seems like the "correct" approach, given, like you say, this is sync over async 🤔 Is that API not available?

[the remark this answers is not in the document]
link04: You’re right, we do pass PipeOptions.Asynchronous. The CancellationToken overload for WriteAsync is available, but the original code has that comment saying it “seems to not work”, which sent me down a bit of a rabbit hole.

My understanding is that this is due to how `PipeStream.WriteAsync(byte[], int, int, CancellationToken)` was implemented on .NET < 7: it inherits from Stream and only checks the token up front. Once the I/O is in flight, the token is effectively ignored. In .NET 7 this was fixed in dotnet/runtime#72503 using CancelSynchronousIo, so cancellation can actually interrupt the I/O, lol this is why I got confused with our sync-over-async:

> This change fixes that by using CancelSynchronousIo to cancel the synchronous I/O performed as part of these async operations, which are implemented as async-over-sync (queueing to the thread pool a work item that performs the synchronous I/O).

That’s why the existing workaround passes the timeout to `.Wait()`(Kevin's fix) instead of to `WriteAsync` itself so it doesn’t cancel the underlying I/O, but it does bound how long the caller waits. Because the tracer still targets net6.0, we can’t rely on the .NET 7 behavior for everyone. I am thinking we could conditionally use the CancellationToken overload on .NET 7+ as a follow-up, but I figured that could be out of scope for this PR.

tracer/src/Datadog.Trace/DogStatsd/StatsdManager.cs
```diff
@@ -300,23 +309,30 @@ public void MarkClosing()
...
         }
 
+        public Task WaitForDisposalAsync() => _disposalComplete.Task;
+
         private void Dispose()
         {
             if (Interlocked.Exchange(ref _disposed, 1) == 0)
             {
                 Log.Debug("Disposing DogStatsdService");
+                _ = Task.Run(DisposeClientAsync);
```
andrewlock, line 322:
```suggestion
                // We push this to a background thread to avoid disposals running in-line
                // in consuming code. This flushes any outstanding data too
                // so could take a relatively long time
                _ = Task.Run(DisposeClientAsync);
```

tracer/src/Datadog.Trace/Vendors/StatsdClient/DogStatsdService.cs
```diff
@@ -255,13 +256,16 @@ public void Flush()
         }
 
         /// <summary>
-        /// Disposes an instance of DogStatsdService.
-        /// Flushes all metrics.
+        /// Asynchronously disposes an instance of DogStatsdService.
+        /// Flushes all metrics without blocking on worker tasks.
         /// </summary>
-        public void Dispose()
+        public async Task DisposeAsync()
```
andrewlock, line 262: Just to clarify, the vendored statsdclient code is _already_ not idempotent right? i.e. if we'd re-run the vendoring tool prior to this PR, we would _already_ have got different code. If so, then this change is probably fine, but it would be nice to have a follow up so that these changes are persisted across re-renderings 🤔 Not a blocker IMO though

tracer/src/Datadog.Trace/Vendors/StatsdClient/DogStatsdService.cs
```diff
@@ -255,13 +256,16 @@ public void Flush()
...
         {
-            _statsdData?.Dispose();
-            _statsdData = null;
+            if (_statsdData is not null)
+            {
+                await _statsdData.DisposeAsync().ConfigureAwait(false);
+                _statsdData = null;
+            }
```
andrewlock, line 268: Super nit, not sure that it's really necessary but hey, why not be safe 😅 
```suggestion
            if (Interlocked.Exchange(ref _statsdData, null!) is { } statsd)
            {
                await statsd.DisposeAsync().ConfigureAwait(false);
            }
```

tracer/src/Datadog.Trace/TracerManager.cs
```diff
@@ -775,7 +778,8 @@ private static async Task RunShutdownTasksAsync(TracerManager instance, Timer he
                     }
 
                     instance.RuntimeMetrics?.Dispose();
-                    instance.Statsd?.Dispose();
+
+                    instance.Statsd?.DisposeAsync().ContinueWith(t => Log.Error(t.Exception, "Error waiting for StatsD disposal"), TaskContinuationOptions.OnlyOnFaulted);
```
andrewlock, line 786: Might be worth adding a comment about _why_ we're doing this here

tracer/test/Datadog.Trace.Tests/Configuration/DynamicConfigurationTests.cs
```diff
@@ -71,10 +72,12 @@ public void ApplyTagsToDirectLogs()
             DynamicConfigurationManager.OnlyForTests_ApplyConfiguration(configBuilder, tracerSettings);
 
             tracerManager.DirectLogSubmission.Formatter.Tags.Should().Be("key2:value2");
+
+            await tracerManager.ShutdownAsync();
```
andrewlock, line 76: oooh, good catch on this, I wonder if we have anywhere else that's doing this 🤔

tracer/test/Datadog.Trace.Tests/DogStatsd/StatsdManagerTests.cs
```diff
@@ -108,10 +108,10 @@ public void HasImpactingChanges_WhenMutableChangesGlobalTags()
     }
 
     [Fact]
-    public void InitialState_ClientNotCreated()
+    public async Task InitialState_ClientNotCreated()
     {
         var clientCount = 0;
-        using var manager = new StatsdManager(new TracerSettings(), (_, _, _) =>
+        var manager = new StatsdManager(new TracerSettings(), (_, _, _) =>
```
andrewlock, line 114: _Ideally_ we should do this, as it ensures we cleanup even if the test fails. But I'm actually not sure if it works in all TFMs 😅 

```suggestion
        await using var manager = new StatsdManager(new TracerSettings(), (_, _, _) =>
```

tracer/src/Datadog.Trace/Vendors/StatsdClient/DogStatsdService.cs
andrewlock: EDIT: looks like a _bunch_ of other statsd dispose pattern is like this so meh

tracer/test/Datadog.Trace.Tests/DogStatsd/StatsdManagerTests.cs
link04: Looked into this and found that StatsdManager doesn't implement IAsyncDisposable so it wouldn't work on all TFMs, I can try adding a note to look in a follow up PR.

tracer/src/Datadog.Trace/DogStatsd/IStatsdManager.cs
```diff
@@ -5,11 +5,12 @@
 
 #nullable enable
 using System;
+using System.Threading.Tasks;
 using Datadog.Trace.Vendors.StatsdClient;
 
 namespace Datadog.Trace.DogStatsd;
 
-internal interface IStatsdManager : IDisposable
+internal interface IStatsdManager
```
bouwkast, line 13: Just wondering why we don't implement `IAsyncDisposable`? Maybe a TFM issue (can't search it right now 😛)

Not a  issue from what I see just curious

tracer/src/Datadog.Trace/Vendors/StatsdClient/IDogStatsd.cs
```diff
@@ -3,13 +3,14 @@
 // This file was automatically generated by the UpdateVendors tool.
```
bouwkast, line 3: I think I saw it somewhere but we should probably do something in a followup PR that disables re-vendoring of this to make sure we don't break it

bouwkast (approved):
Thanks!

Just a couple of follow ups but they may have already been addressed/answered

PR #8266, opened 2026-03-03
[Metrics] Send Diagnostics GC Pause Time as Counter instead of Timer

link04:
## Summary of changes

Send Diagnostics GC pause time as a Counter instead of a Timer.

## Reason for change

The Diagnostics listener calls `GC.GetTotalPauseDuration()` which returns cumulative pause for the whole 10s interval. Sending that as a single `Timer()` inflates `.median`/`.max` compared to EventListener, which sends one `Timer()` per individual GC event. There's no public .NET API for per-GC pause duration without EventPipe.

After consulting with the Agent Metrics team (#agent-metrics), the recommended approach is a separate Counter for total pause, with average computed at query time. See [RFC: Enabling .NET Runtime Metrics by Default — GC Pause Time](https://docs.google.com/document/d/1tekvCvlOkn12pU3jLK-ePZ0nrynSW7p8IWgG-Dh7zwQ/edit?pli=1&tab=t.0#heading=h.hz8udk7cv9zf).

## Implementation details

- Add `GcPauseTimeTotal` constant (`runtime.dotnet.gc.pause_time.total`) to `MetricsNames`.
- Replace `statsd.Timer(MetricsNames.GcPauseTime, ...)` with `statsd.Counter(MetricsNames.GcPauseTimeTotal, (long)totalPauseDelta)` in `DiagnosticsMetricsRuntimeMetricsListener`.
- Average per-GC pause can be computed at query time: `pause_time.total / (gc.count.gen0 + gen1 + gen2)` in OOTB dashboard updated in a separate PR: 
https://github.com/DataDog/integrations-internal-core/pull/2823

## Test coverage

Updated `DiagnosticMetricsRuntimeMetricsListenerTests.MonitorGarbageCollections` to assert `Counter(GcPauseTimeTotal)` instead of `Timer(GcPauseTime)`.

## Other details

Part of [Enabling .NET Runtime Metrics by Default](https://docs.google.com/document/d/1tekvCvlOkn12pU3jLK-ePZ0nrynSW7p8IWgG-Dh7zwQ/edit?pli=1&tab=t.0#heading=h.7u0xyjk7jxq5).


tracer/src/Datadog.Trace/RuntimeMetrics/DiagnosticsMetricsRuntimeMetricsListener.cs
```diff
@@ -170,7 +170,14 @@ public void Refresh()
         if (gcPauseTimeMilliSeconds.HasValue && _previousGcPauseTime.HasValue
                                              && gcPauseTimeMilliSeconds.Value != _previousGcPauseTime.Value)
         {
-            statsd.Timer(MetricsNames.GcPauseTime, gcPauseTimeMilliSeconds.Value - _previousGcPauseTime.Value);
+            var totalPauseDelta = gcPauseTimeMilliSeconds.Value - _previousGcPauseTime.Value;
+
+            if (totalPauseDelta > 0)
+            {
```
andrewlock: I think we can actually ditch all of this "not 0-length pauses" stuff now. That was basically a misunderstanding in the original implementation, so instead we can just report whatever we have (as long as it's `>= 0`, which it _always_ should be). That's probably especially worthwhile, because the `long` truncation can essentially potentially send `0` _anyway_, so we're not guaranteeing sending non-zero values as it is😄

tracer/src/Datadog.Trace/RuntimeMetrics/DiagnosticsMetricsRuntimeMetricsListener.cs
```diff
@@ -170,7 +170,14 @@ public void Refresh()
...
+                // Send total pause as a Counter. Avg per-GC pause can be computed at query time:
+                // pause_time.total / (gc.count.gen0 + gc.count.gen1 + gc.count.gen2)
+                statsd.Counter(MetricsNames.GcPauseTimeTotal, (long)totalPauseDelta);
```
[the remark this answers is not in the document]
andrewlock: In general I think this is fine, as it's what we did before, but it _does_ mean we would _always_ under report (because we wuold _always_ chop off the decimal points. A (potentially better) alternative would be to explicitly round the value first, which will round up/down as appropriate, and reduce bias:
```suggestion
                statsd.Counter(MetricsNames.GcPauseTimeTotal, (long)Math.Round(totalPauseDelta));
```

tracer/src/Datadog.Trace/RuntimeMetrics/DiagnosticsMetricsRuntimeMetricsListener.cs
link04: I went ahead and did the changes including removing totalPauseDelta > 0 since the Delta should not be 0 unless there is some bug somewhere.

